PR #164, opened 2019-11-15
Add decoration to focused item

asger-semmle:
When following a link from the result viewer, we highlighted the target range simply by selecting it. 

When this range is an identifier, it triggers vscode's occurrence-match highlighting, which makes it hard to see at a glance what was highlighted. This PR introduces decorators to highlight it a bit more, using the same colors as in search/replace.

The decoration is cleared on any selection change event in that editor.

Before
-
Screenshot 2019-11-15 at 12 40 14

After
-
Screenshot 2019-11-15 at 12 40 08

### DIFF
--- a/extensions/ql-vscode/src/interface.ts
+++ b/extensions/ql-vscode/src/interface.ts
@@ -98,6 +98,7 @@ export class InterfaceManager extends DisposableObject {
 
     super();
     this.push(this._diagnosticCollection);
+    this.push(vscode.window.onDidChangeTextEditorSelection(this.handleSelectionChange.bind(this)));
   }
 
   // Returns the webview panel, creating it if it doesn't already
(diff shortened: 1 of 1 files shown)


jcreedcmu:
Code lgtm, but I still find the simultaneous decoration and selection of code to be a bit jarring, since for some reason I still have not completely understood, selection includes one character's worth of whitespace at every newline, but decorations with `isWholeLine: false` do not, resulting in things like the following screenshot:
image

Do you suppose we could just rely on decoration only, and skip the selection-setting on https://github.com/github/vscode-codeql/pull/164/files#diff-73bb24db58f098e75356f2e05edb5e62R432 ?

asgerf:
That's a good point.

I've experimented quite a bit with where to place the cursor, and found that for _small_ ranges, bracket-highlighting is more attention-grabbing than our decorator which is why I didn't like placing the cursor at the beginning. But multi-line selections don't have that problem (as they're large) so I decided to just treat them differently.

WDYT?

PR #173, opened 2019-11-20
Implement QL Test support

dbartol:
The PR contains the initial implementing of QL Test support in CodeQL for Visual Studio Code (https://github.com/github/pe-security-codeql/issues/28). ~~Because QL Test support isn't quite ready in the CLI yet, this PR uses `odasa` to run the tests for now. As CLI support comes online, it should be straightforward to swap out the implementation to use the CLI.~~ This is now implemented using `codeql tests run`.

The treeview UI for the tests is implemented via the `hbenl.vscode-test-explorer` extension. This extension is open source, and appears to be actively maintained. It's used by a couple dozen existing extensions for tests for various languages. The extension doesn't really do anything on its own, so taking it as a dependency isn't introducing any unwanted UI clutter. Note that I did have to remove the `--disable-extensions` argument from `launch.json`, because otherwise the test explorer extension gets disabled, preventing our own extension from loading.

The UI will display a root node for each QL pack that contains tests, with the actual test directories and files as descendants of that root node. We consider only those QL packs in the workspace; QL packs on the default CodeQL search path are ignored. We use `codeql resolve qlpacks` to find the packs, and then watch all `qlpack.yml` files in the workspace for changes in order to refresh the pack discovery when necessary. Ideally, we'd have the CLI return a set of path patterns to watch, but for now the current implementation works fine.

To discover the tests within a given pack, ~~we walk the pack's directory tree manually for now, until the relevant CLI command is available~~we use `codeql resolve tests`. Because we do not yet have a mechanism in `qlpack.yml` to specify whether or not the pack contains tests, we assume that any pack whose name ends with "-tests" to contain nothing but tests, and any other pack to contain no tests. This is sufficient for the tests in the QL repo. As with QL pack discovery, we watch the file system for changes in `.ql` and `.qlref` files in order to refresh the tree of tests if anything changes.

To actually run the tests, we just invoke `codeql test run` with the appropriate arguments.

The `test-ui.ts` file implements a couple of additional commands for the context menu of the test treeview. You can accept the output of a failing test (copying the `.actual` file to the `.expected` file), and you can bring up a diff view of the `.expected` and `.actual` files).

This PR includes a couple of related utility classes. `UIService` makes it a little easier to implement a service that handles VS Code commands. `Discovery` is a base class that handles most of the work that is shared between the different kinds of discovery that we do, like avoiding running multiple discovery operations simultaneously if we get a storm of file change notifications.


dbartol:
I've added a much better error message when trying to run tests without Semmle Core configured. Since we may be waiting a while for CLI test support, I think we should merge this PR now rather than let it rot while we wait.

.vscode/launch.json
```diff
@@ -8,8 +8,7 @@
       "request": "launch",
       "runtimeExecutable": "${execPath}",
       "args": [
-        "--extensionDevelopmentPath=${workspaceRoot}/dist/vscode-codeql",
-        "--disable-extensions"
```
jcreedcmu: I don't quite understand the reason for removing `--disable-extensions`? I thought we had it in here to ensure that integration test failures couldn't be because some other extension was confusing things.

dbartol: Our extension now depends on the "VS Code Test Explorer" extension. If we launch VS Code with `--disable-extensions`, it disables the "VS Code Test Explorer" extension, and therefore refuses to load our own extension due to the disabled dependency.

extensions/ql-vscode/package.json
```diff
@@ -18,10 +18,14 @@
...
     "Programming Languages"
   ],
+  "extensionDependencies": [
+    "hbenl.vscode-test-explorer"
+  ],
   "activationEvents": [
     "onLanguage:ql",
     "onView:codeQLDatabases",
     "onView:codeQLQueryHistory",
+    "onView:test-explorer",
```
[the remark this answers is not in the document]
dbartol: I think we need it for the same reason we need the other `onView:` events: If I open a workspace that contains CodeQL tests, but have not yet opened a .ql file, I should still be able to browse and run my tests.

extensions/ql-vscode/package.json
```diff
@@ -111,6 +121,22 @@
           "type": "boolean",
           "default": false,
           "description": "Enable debug logging and tuple counting when running CodeQL queries. This information is useful for debugging query performance."
+        },
+        "codeQL.tests.semmleCoreDistributionPath": {
+          "scope": "window",
```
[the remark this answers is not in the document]
dbartol: No longer relevant, since we don't need odasa anymore.

extensions/ql-vscode/src/qltest-discovery.ts
```diff
@@ -0,0 +1,282 @@
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import { QLPackDiscovery } from './qlpack-discovery';
+import { Discovery } from './discovery';
+import { EventEmitter, Event, Uri, GlobPattern, workspace, RelativePattern } from 'vscode';
+import { DisposableObject } from 'semmle-vscode-utils';
+
+/**
+ * A node in the tree of tests. This will be either a `QLTestDirector` or a `QLTestFile`.
```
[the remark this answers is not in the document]
dbartol: Fixed

extensions/ql-vscode/package.json
```diff
@@ -111,6 +121,22 @@
...
+          "type": "string",
+          "default": "",
+          "description": "Location of the Semmle Core distribution"
+        },
+        "codeQL.tests.semmleCoreLicensePath": {
+          "scope": "window",
```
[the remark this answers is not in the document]
dbartol: No longer relevant.

extensions/ql-vscode/package.json
```diff
@@ -111,6 +121,22 @@
...
+          "type": "string",
+          "description": "Location of the directory containing the Semmle Core license file"
+        },
+        "codeQL.tests.numberOfThreads": {
+          "scope": "window",
+          "type": "integer",
```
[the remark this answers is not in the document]
dbartol: Fixed to be 1, allowed range [1, 1024], just like the number of threads for queries.

extensions/ql-vscode/src/cli.ts
```diff
@@ -66,13 +66,6 @@ export interface SourceInfo {
   sourceLocationPrefix: string;
 }
 
-/**
- * The expected output of `codeql resolve qlpacks`.
- */
-export interface ResolvedQLPacks {
-  [index: string]: string[];
-}
-
```
jcreedcmu: Seems to be used in `src/qlpack-discovery.ts`, actually

dbartol: Oops. Fixed.

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -1,5 +1,7 @@
 import * as cp from 'child_process';
-import { DisposableObject } from 'semmle-vscode-utils';
+// Import from the specific module within `semmle-vscode-utils`, rather than via `index.ts`, because
+// we avoid taking an accidental runtime dependency on `vscode` this way.
+import { DisposableObject } from 'semmle-vscode-utils/out/disposable-object';
```
jcreedcmu: I'm a little confused by this. Depending on `/out/` seems abstraction-boundary-breaking in a way that makes me feel worried. If we really wanted to cut this dependency, could we instead just refactor `disposable-object` to refer to a pared down interface we define ourselves like
```
interface SimplyDisposable {
  dispose(): void;
}
```
?
Separately, on the very next line we depend on `{ Disposable } from 'vscode'`, so what's being saved?

extensions/ql-vscode/src/qltest-discovery.ts
```diff
@@ -0,0 +1,206 @@
...
+/**
+ * A directory containing one or more QL tests or other test directories.
+ */
+export class QLTestDirectory extends QLTestNode {
+  constructor(_path: string, _name: string, private _children: QLTestNode[]) {
+    super(_path, _name);
+  }
+
+  public get children(): readonly QLTestNode[] {
+    return this._children;
```
jcreedcmu:
```suggestion
    return this._children.sort((a, b) => a.path.localeCompare(b.path));
```
Or perhaps there's a better place to do sorting, but we should ensure a sorted order rather than relying on the filesystem to do it for us --- for indeed on linux I see tests in an arbitrary unsorted order.

dbartol: Fixed

extensions/ql-vscode/src/queryserver-client.ts
dbartol: There is a unit test that tries to run code from `queryserver-cli.ts` outside of VSCode. If we just do `import { DisposableObject } from 'semmle-vscode-utils'`, that imports the `index.ts` of `semmle-vscode-utils`, which in turn imports not just the definition of `DisposableObject`, but also other code. `DisposableObject` only depends on `Disposable`, which is merely a TypeScript type that gets erased when transpiled to JavaScript. The other code in `semmle-vscode-utils`, however, has actual implementation dependencies on executable code exported from `vscode`, and therefore forces a runtime dependency on `vscode`. By importing only that single file from within the module, we avoid importing the other code from `semmle-vscode-utils`, and thus avoid the runtime dependency.

A cleaner solution would probably be to move `DisposableObject` into a package that has only a dev dependency on `vscode`, or simply defines its own `Disposable`. I'm happy to do this, either in this PR or in a follow up.

dbartol: Here's the failure this causes: https://github.com/dbartol/vscode-codeql/runs/490822526

jcreedcmu: Ah, sounds good. didn't realize it was actually *causing* test failures, rather than just being a cleanliness issue. Filed https://github.com/github/vscode-codeql/issues/268.

PR #175, opened 2019-11-21
Add commands for navigating the steps on a path

asgerf:
This adds the command `CodeQL: Show Next Step on Path` and `CodeQL: Show Previous Step on Path` for navigating the steps on the currently shown path.

The idea is that you can add keybindings for these commands to navigate the path without the mouse. Doing this with the mouse required quite a bit of zig-zag eye tracking between the code view and result view.

You still have to click one of the steps on a path to "choose" that path. There are currently no commands for switching between results or paths. It would be nice to have a fully keyboard-controllable result view, but I think I'll leave that for a future PR.

It's a bit of a hassle to make the result view remember which path node was previously clicked. For this we need a way to reference a specific part of the result set, that is, a result index, path index, and a path step index. I introduced the types `Keys.Result`, `Keys.Path`, and `Keys.PathNode` to represent such indices.

Commit-by-commit review recommended.

### DIFF
--- a/extensions/ql-vscode/src/view/alert-table.tsx
+++ b/extensions/ql-vscode/src/view/alert-table.tsx
@@ -4,8 +4,8 @@ import * as Sarif from 'sarif';
 import * as Keys from '../result-keys';
 import { LocationStyle, ResolvableLocationValue } from 'semmle-bqrs';
 import * as octicons from './octicons';
-import { className, renderLocation, ResultTableProps, zebraStripe, selectableZebraStripe } from './result-table-utils';
-import { PathTableResultSet } from './results';
+import { className, renderLocation, ResultTableProps, zebraStripe, selectableZebraStripe, jumpToLocation } from './result-table-utils';
+import { PathTableResultSet, onNavigation, NavigationEvent } from './results';
 
 export type PathTableProps = ResultTableProps & { resultSet: PathTableResultSet };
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/view/alert-table.tsx
```diff
@@ -72,7 +74,8 @@ export function getPathRelativeToSourceLocationPrefix(sourceLocationPrefix: stri
 export class PathTable extends React.Component<PathTableProps, PathTableState> {
   constructor(props: PathTableProps) {
     super(props);
-    this.state = { expanded: {} };
+    this.state = { expanded: {}, selectedPathNode: undefined };
+    this.handleNavigationEvent = this.handleNavigationEvent.bind(this);
```
jcreedcmu: I almost prefer to have the `.bind(this)`s at the callsite, or else give the property the `.bind()`ed method a different name from the original method name. Do you have a strong opinion here? I haven't seen this pattern before but am willing to defer if you feel like it's idiomatic.

asgerf: I'd say it's quite common. It's recommended [on reactjs.org](https://reactjs.org/docs/handling-events.html), with the other alternative being to make the method into a field instead.

> We generally recommend binding in the constructor or using the class fields syntax, to avoid this sort of performance problem.

PR #176, opened 2019-11-21
Improvements to Query History Labels

jcreedcmu:
- there's now a command in the context menu for setting the label to some custom text
- the default format string can be configured in settings

Fixes https://github.com/github/pe-security-codeql/issues/198


felicitymay:
This improvement seems to be missing from the change log as of: https://github.com/github/vscode-codeql/runs/342551391

jcreedcmu:
Good point: https://github.com/github/vscode-codeql/pull/195.

PR #181, opened 2019-11-25
Store the current counter value and use it for the sorted results path.

alexet:
This way they don't clash when switching between sorted queries. As the filename wasn't assigned in the constructor, it was using the value for the next run query rather than the id for the current query.

### DIFF
--- a/extensions/ql-vscode/src/queries.ts
+++ b/extensions/ql-vscode/src/queries.ts
@@ -48,7 +48,7 @@ export class QueryInfo {
    */
   sortedResultsInfo: Map<string, SortedResultSetInfo>;
   dataset: vscode.Uri; // guarantee the existence of a well-defined dataset dir at this point
-
+  queryId: number;
   constructor(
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/queries.ts
```diff
@@ -48,25 +48,25 @@ export class QueryInfo {
...
+  queryId: number;
   constructor(
     public program: messages.QlProgram,
     public dbItem: DatabaseItem,
     public queryDbscheme: string, // the dbscheme file the query expects, based on library path resolution
     public quickEvalPosition?: messages.Position,
     public metadata?: cli.QueryMetadata,
   ) {
-    this.compiledQueryPath = path.join(tmpDir.name, `compiledQuery${queryCounter}.qlo`);
+    this.queryId = queryCounter++;
```
henrymercer: Could potentially rename `queryCounter` to `nextQueryId` to make the distinction between these variables clearer.

PR #183, opened 2019-11-25
Line 3

jymok12:
This extension is written in TypeScript for VS Code used to find problem in code bases.

I've converted 3 sentences into 1 sentence


adityasharad:
Hi @jymok12, thanks for your interest.

The description is intended to convey the message that this extension adds language support for CodeQL to VS Code. (CodeQL is both a programming language and code analysis engine maintained by GitHub.) For this reason, I would prefer to keep the existing description as it includes all the relevant information.

Is there something about this description that you find unclear, which we can improve?

PR #184, opened 2019-11-25
Create diagnostics messages using sarif.

alexet:
This covers a few edge cases not handled in the current support and removes a dependency from the typescript parsing code.

### DIFF
--- a/extensions/ql-vscode/src/interface-types.ts
+++ b/extensions/ql-vscode/src/interface-types.ts
@@ -16,6 +16,14 @@ export interface DatabaseInfo {
   databaseUri: string;
 }
 
+/** Arbitrary query metadata */
+export interface QueryMetadata {	
+  name?: string,	
+  description?: string,	
+  id?: string,	
+  kind?: string	
+}
+
 export interface PreviousExecution {
   queryName: string;
   time: string;
@@ -31,7 +39,6 @@ export interface Interpretation {
 
 export interface ResultsInfo {
   resultsPath: string;
-  interpretedResultsPath: string;
 }
 
 export interface SortedResultSetInfo {
@@ -56,7 +63,7 @@ export interface SetStateMsg {
   sortedResultsMap: SortedResultsMap;
   interpretation: undefined | Interpretation;
   database: DatabaseInfo;
-  kind?: string;
+  metadata?: QueryMetadata
   /**
    * Whether to keep displaying the old results while rendering the new results.
    *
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/interface-types.ts
```diff
@@ -16,6 +16,9 @@ export interface DatabaseInfo {
   databaseUri: string;
 }
 
+/** Arbitrary query metadata */
+export type QueryMetadata = {[key: string]: string};
```
jcreedcmu: I don't really understand the motivation behind the change to this type. What do we gain from *not* expressing some understanding of the structure of the metadata? Right now the only ones we use are `kind` and `id`, but it seems useful as documentation in case we want to use other fields in the future.

extensions/ql-vscode/src/interface.ts
```diff
@@ -262,10 +263,31 @@ export class InterfaceManager extends DisposableObject {
       sortedResultsMap,
       database: info.database,
       shouldKeepOldResultsWhileRendering,
-      kind: info.query.metadata ? info.query.metadata.kind : undefined
+      metadata: info.query.metadata
```
jcreedcmu: Passing along all of the metadata here instead of just `kind` seems like a good idea, though.

extensions/ql-vscode/src/queries.ts
```diff
@@ -173,11 +172,12 @@ export class QueryInfo {
 /**
  * Call cli command to interpret results.
  */
-export async function interpretResults(server: cli.CodeQLCliServer, queryInfo: QueryInfo, resultsInfo: ResultsInfo, sourceInfo?: cli.SourceInfo): Promise<sarif.Log> {
-  if (await fs.pathExists(resultsInfo.interpretedResultsPath)) {
-    return JSON.parse(await fs.readFile(resultsInfo.interpretedResultsPath, 'utf8'));
+export async function interpretResults(server: cli.CodeQLCliServer, metadata: QueryMetadata | undefined, resultsPath: string, sourceInfo?: cli.SourceInfo): Promise<sarif.Log> {
+  const interpretedResultsPath = resultsPath + ".interpreted.sarif"
```
jcreedcmu: Was there something wrong about computing all the requisite paths for a query up front? Is there a situation in which we need the flexibility of multiple different results files for a single `queryInfo`, necessitating passing the `resultsPath` down by itself? I was trying to isolate all the filename generation to a single place.

alexet: Ok I have switched back to generating them in one place. I had to change some things around to pass the data through the UI. So this may need a fairly large re-review.

jcreedcmu:
This code effectively landed in #222.

PR #185, opened 2019-11-28
Split queries

alexet:
Split up query running, upgrade running, and operations on queries post run. The main point is clearly separating which state is post query run state (such as the locations of the sorted results) compared to the immutable results of running the query.


jcreedcmu:
This code effectively landed in #222.

PR #195, opened 2019-12-11
Update changelog

jcreedcmu:
Add some items off the top of my head. Let me know if there's anything that should be here that I neglected.


felicitymay:
Thanks Jason. All LGTM, but I'll leave Shati to merge this if she's happy with the updates.

shati-patel:
Is this change worth mentioning too? I don't know how detailed we need the changelog to be.

- Add commands for navigating the steps on a path 
(see https://github.com/github/vscode-codeql/pull/175)

alexet:
Yes that should also be added.

PR #204, opened 2020-01-02
Fix double-encoding of "#" in webview URI conversion

henrymercer:
This fixes sorting for result sets with a "#" in their name.  Add an integration test to verify this behaviour.


alexet:
LGTM

PR #205, opened 2020-01-03
Increase timeout for query unit tests

henrymercer:
This pull increases the timeout for the query unit tests to avoid transient test failures like https://github.com/github/vscode-codeql/commit/e966c339d35766550b91a85eda1a93d8cb822112/checks?check_suite_id=383124497 resulting from the query server not starting up in time.

### DIFF
--- a/extensions/ql-vscode/test/pure-tests/query-test.ts
+++ b/extensions/ql-vscode/test/pure-tests/query-test.ts
@@ -83,6 +83,10 @@ describe('using the query server', function () {
     }
   });
 
+  // Note this does not work with arrow functions as the test case bodies:
+  // ensure they are all written with standard anonymous functions.
+  this.timeout(10000);
+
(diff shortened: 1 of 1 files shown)


shati-patel:
(Drive-by comment)

Should this (and previous changes like #202 and #204) be added to the [changelog](https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/CHANGELOG.md)? If so, could we do that soon? A bulk update before a release (like last time) is more likely to miss stuff!

alexet:
This is just a fix for tests so it doesn't need to be added. The other two PRs you mentioned probably should have a change-log entry.

henrymercer:
> A bulk update before a release (like last time) is more likely to miss stuff!

Good point, I've made a pull to update the changelog: https://github.com/github/vscode-codeql/pull/206

PR #210, opened 2020-01-13
Teach extension to do Quick Query

jcreedcmu:
Fixes https://github.com/github/codeql-coreql-team/issues/27.


extensions/ql-vscode/src/quick-query.ts
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * `getBaseText` heuristically returns an appropriate import
+ * statement prelude based on the filename of the dbscheme file
+ * given. This information might be more appropriately provided by
+ * the qlpack itself.
+ */
+function getBaseText(dbschemeBase: string) {
+  if (dbschemeBase == 'semmlecode.javascript.dbscheme') return 'import javascript\n\nselect ""';
+  if (dbschemeBase == 'semmlecode.cpp.dbscheme') return 'import cpp\n\nselect ""';
+  if (dbschemeBase == 'semmlecode.java.dbscheme') return 'import java\n\nselect ""';
```
p0: This is called [`semmlecode.dbscheme`](https://github.com/Semmle/ql/blob/master/java/ql/src/config/semmlecode.dbscheme), not `semmlecode.java.dbscheme`.

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -0,0 +1,143 @@
...
+  if (dbschemeBase == 'semmlecode.python.dbscheme') return 'import python\n\nselect ""';
+  return '';
```
p0: Should this be `'select ""'`?

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -0,0 +1,143 @@
...
+ * `getBaseText` heuristically returns an appropriate import
+ * statement prelude based on the filename of the dbscheme file
+ * given. This information might be more appropriately provided by
+ * the qlpack itself.
+ */
+function getBaseText(dbschemeBase: string) {
+  if (dbschemeBase == 'semmlecode.javascript.dbscheme') return 'import javascript\n\nselect ""';
+  if (dbschemeBase == 'semmlecode.cpp.dbscheme') return 'import cpp\n\nselect ""';
+  if (dbschemeBase == 'semmlecode.java.dbscheme') return 'import java\n\nselect ""';
+  if (dbschemeBase == 'semmlecode.python.dbscheme') return 'import python\n\nselect ""';
```
p0: We should also handle `semmlecode.csharp.dbscheme` and `go.dbscheme`.

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -0,0 +1,143 @@
...
+  throw new Error(`Could not find qlpack file for dbscheme ${dbschemePath}`);
+}
+
+/**
+ * `getBaseText` heuristically returns an appropriate import
+ * statement prelude based on the filename of the dbscheme file
+ * given. This information might be more appropriately provided by
+ * the qlpack itself.
+ */
+function getBaseText(dbschemeBase: string) {
```
p0: It's a little annoying to hard-code these here. I guess the long-term solution is to allow QL packs to indicate a "default import" in their metadata, and then use the default import from the QL pack with a matching dbscheme?

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Show a buffer the user can enter a simple query into.
+ */
+export async function displayQuickQuery(ctx: ExtensionContext, cliServer: CodeQLCliServer, databaseUI: DatabaseUI) {
+  try {
+
+    // If there is already a quick query open, don't clobber it, just
+    // show it.
+    const existing = workspace.textDocuments.find(doc => path.basename(doc.uri.fsPath) === QUICK_QUERY_QUERY_NAME);
```
p0: Is checking on the `path.basename` enough? `QUICK_QUERY_QUER_NAME` isn't currently a particularly unique string, and if someone does have a QL file with a clashing name open they'll get odd behaviour.

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -0,0 +1,143 @@
...
+    const datasetFolder = await dbItem.getDatasetFolder(cliServer);
+    const dbschemes = await glob(path.join(datasetFolder, '*.dbscheme'))
+
+    if (dbschemes.length < 1) {
+      throw new Error(`Can't find dbscheme for current database in ${datasetFolder}`);
+    }
+
+    const dbscheme = dbschemes[0];
+    if (dbschemes.length > 1) {
+      logger.log(`Found multiple dbschemes in ${datasetFolder}; arbitrarily choosing the first, ${dbscheme}`);
```
p0: Does `glob` return files in filesystem-iteration-order or in alphabetic order? It'd probably be safer to sort the dbscheme names before taking the first, otherwise filesystem alterations can change which one we resolve to.

Alternatively, make this a loud error for now -- at the point where we have multiple dbschemes in a database, we'll have to seriously consider what the behaviour of quick queries should be...

jcreedcmu: Done; sorted for determinism and made error louder.

extensions/ql-vscode/src/quick-query.ts
jcreedcmu: I tried sticking a sentinel in the `vscode.Uri.fragment`, which works on the vscode side (and I think is pretty dependable from the point of view of the user not being likely to be able to collide with it) but the language server doesn't like filenames with fragments in them at the moment:
```
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:67)
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.notify(GenericEndpoint.java:152)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleNotification(RemoteEndpoint.java:220)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:187)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:194)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
	at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:113)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:65)
	... 11 more
Caused by: java.lang.IllegalArgumentException: URI has a fragment component
	at java.base/sun.nio.fs.UnixUriUtils.fromUri(Unknown Source)
	at java.base/sun.nio.fs.UnixFileSystemProvider.getPath(Unknown Source)
	at java.base/java.nio.file.Path.of(Unknown Source)
	at java.base/java.nio.file.Paths.get(Unknown Source)
	at com.semmle.frontend.server.LanguageServerUtils.tryConvertUriToPath(LanguageServerUtils.java:88)
	at com.semmle.frontend.server.LanguageServerUtils.uriToPath(LanguageServerUtils.java:70)
	at com.semmle.frontend.server.QLLanguageServer$QLTextDocumentService.documentUpdate(QLLanguageServer.java:375)
	at com.semmle.frontend.server.QLLanguageServer$QLTextDocumentService.didOpen(QLLanguageServer.java:391)
	... 16 more
```
Could we change `ideserver` to ignore that? Any downsides?

extensions/ql-vscode/src/quick-query.ts
jcreedcmu: Yes, that is what the comment was supposed to be referring to. Made the comment a little clearer.

extensions/ql-vscode/src/quick-query.ts
jcreedcmu: Done.

extensions/ql-vscode/src/quick-query.ts
jcreedcmu: Good catch, done.

extensions/ql-vscode/src/quick-query.ts
jcreedcmu: Oops, thanks, done.

extensions/ql-vscode/src/quick-query.ts
alexet: I think a better solution in the end will be to add quick eval support to the ide server however I think this solution is OK for now. I think the correct solution will be to use a custom uri scheme for quick-query and a method for setting the qlpacks for quick-query. That way we also don't need add the workspace either.

extensions/ql-vscode/package.json
```diff
@@ -134,6 +134,10 @@
         "command": "codeQL.quickEval",
         "title": "CodeQL: Quick Evaluation"
       },
+      {
```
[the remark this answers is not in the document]
jcreedcmu: Good catch.

PR #211, opened 2020-01-17
Use codeql.exe instead of codeql.cmd on Windows

nickrolfe:
We just merged a change so that the main CodeQL executable on Windows is called `codeql.exe`. For now, `codeql.cmd` still exists as a wrapper, so we have some leeway in making this change.

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -85,7 +85,7 @@
           "scope": "machine",
           "type": "string",
           "default": "",
-          "description": "Path to the CodeQL executable that should be used by the CodeQL extension. The executable is named `codeql` on Linux/Mac and `codeql.cmd` on Windows. This overrides all other CodeQL CLI settings."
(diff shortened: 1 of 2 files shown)


shati-patel:
@nickrolfe I happened to spot this PR, but would you mind opening an issue so we remember to update the documentation?
-> https://github.com/github/codeql-coreql-team/issues/new/ (**Improve existing CodeQL docs**)

adityasharad:
Checking my understanding about compatibility: this will only work with newer versions of the CLI that include the `codeql.exe` file, so should we either:
- hold off on making an extension release until a suitable CLI release is made first, or
- update the [version constraint](https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/src/distribution.ts#L41)?

nickrolfe:
I'll leave it up to you, but there's a third option:
* Revert this change and only un-revert after a suitable CLI release has been made, since the new CLI release will still have a wrapper `codeql.cmd` for compatibility.

PR #212, opened 2020-01-17
Reuse existing editor if file already open

RasmusWL:


RasmusWL:
Failed _ Run integration tests (Linux)_ with 

```
    ✓ should correctly round trip from filesystem to webview and back

    ✓ does not double-encode # in URIs

Server response: <?xml version="1.0" encoding="UTF-8"

Exit code:   null
Done

Failed to run tests
```

No clue why is happening :confused:

PR #214, opened 2020-01-21
Revert "Use codeql.exe instead of codeql.cmd on Windows"

henrymercer:
Reverts github/vscode-codeql#211

We need to (1) make a new CLI release and (2) update the extension's CLI version constraint before launching CodeQL via `codeql.exe`.

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -85,7 +85,7 @@
           "scope": "machine",
           "type": "string",
           "default": "",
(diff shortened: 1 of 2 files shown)


jcreedcmu:
Ah, my bad, didn't realize this was gated on that.

jcreedcmu:
Waiting for flaky test to re-run...

henrymercer:
> Waiting for flaky test to re-run...
Do you know which test this is?

jcreedcmu:
I don't understand some of these test failures, and have grown paranoid that they are due to the actions infra rather than the tests themselves? I've seen ubuntu and windows tests fail with no more informative error message than some very large integer exit code. Any other insight extremely welcome.

PR #215, opened 2020-01-21
Release action: Fetch master branch in checkout

henrymercer:
This PR fetches the master branch so the `base` branch exists for the "Create version bump PR" step.  It also updates the version of https://github.com/peter-evans/create-pull-request to fix failing dependency resolution in the current version of the action.

Example run: https://github.com/henrymercer/vscode-codeql/runs/401342217
Example PR: https://github.com/henrymercer/vscode-codeql/pull/4


.github/workflows/release.yml
```diff
@@ -29,6 +29,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@master
```
adityasharad: As discussed separately, our understanding is that for tag pushes this action will check out the tag. Please add a comment here to remind us of this fact.

.github/workflows/release.yml
```diff
@@ -29,6 +29,10 @@ jobs:
...
 
+      - name: Fetch master branch
+        run: |
```
adityasharad: Comment: `master` is needed as the base branch of the generated PR

.github/workflows/release.yml
```diff
@@ -100,13 +104,12 @@ jobs:
           echo "::set-output name=next_version::$NEXT_VERSION"
 
       - name: Create version bump PR
-        uses: peter-evans/create-pull-request@7531167f24e3914996c8d5110b5e08478ddadff9  # v1.8.0
+        uses: peter-evans/create-pull-request@c202684c928d4c9f18394b2ad11df905c5d8b40c # v2.1.2
```
adityasharad: Good spot, thank you. I didn't realise this was affecting all repos using the action.

Issue #216, opened 2020-01-23
Autocomplete improvements


calumgrant:
If I'm typing some QL, and the last characters I wrote were `and` or `or`, then the autocomplete pops up with a list of candidate predicates and classes containing the respective characters.

This is quite tiring. A few suggested fixes:

- Ensure that `and` or `or` is the top autocomplete item
- Suppress the autocomplete box if the token is `and` or `or` (or indeed any other keyword?)
- Make autocomplete context aware. For example, 
  - if the completion requires a logic operation, do not populate autocomplete with predicates or classes
  - if the completion requires a predicate or variable, only populate autocomplete with variables and predicates
  - if the completion requires a class name, only populate autocomplete with class names or types.

A few suggestions for making autocomplete context aware:
- Use the parse tree. This probably won't work since the QL parser is dreadful at error recovery.
- If the preceding token is a `(` or a `,`, only populate type names
- If the preceding token is `{`, `|`, `=`, `and` or `or`, only populate predicate names or quantifiers (`exists` etc).
- If the preceding token is a variable name, or a `)`, only populate logic keywords
- If the preceding token is `.`, ideally use the type to guide autocomplete. If not, autocomplete using predicate names.
- Otherwise, autocomplete using all things (predicates & types etc).

I'm sure these rules are incomplete, and will need further refinement, but I hope you get the idea.

alexet:
This sounds like you are running into the issue of the vscode word based suggestions. We already do much the above by using the parse tree. However if we don't give a suggestion vscode automatically tries to give suggestions instead. 

Does disabling "word based suggestions" in you user preferences solve most of your problems. We are looking into improving this by adding keywords to the list of things the we suggest. We are also looking into disabling this automatically for ql but changing the users settings feels quite invasive.

jcreedcmu:
The just-released v1.0.4 incorporates #217, which should at least restore the Eclipse status quo with respect to disabling word-based autocomplete.

aeisenberg:
Closing this issue because it has been partially addressed. We will open an issue elsewhere for the remaining work.

PR #217, opened 2020-01-23
Default to disabling word based auto-complete when editing QL.

jcreedcmu:
Addresses at least one point of https://github.com/github/vscode-codeql/issues/216.


alexet:
Can you add this to the changelog?

p0:
Do we want this for the `dbscheme` language too?

jcreedcmu:
Done and done.

PR #220, opened 2020-01-28
Truncate long database upgrade messages in native dialog box

jcreedcmu:
Fixes https://github.com/github/codeql-coreql-team/issues/252.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,10 @@ jobs:
         with:
           fetch-depth: 1
 
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '10.18.1'
+
       - name: Build
         run: |
           cd build
@@ -45,6 +49,10 @@ jobs:
         uses: actions/checkout@v1
         with:
           fetch-depth: 1
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '10.18.1'
 
       # We have to build the dependencies in `lib` before running any tests.
       - name: Build
--- a/extensions/ql-vscode/src/queries.ts
+++ b/extensions/ql-vscode/src/queries.ts
@@ -16,11 +16,11 @@ import { QueryHistoryItemOptions } from './query-history';
 import { isQuickQueryPath } from './quick-query';
 
 /**
- * Maximum number of lines to put in a binary choice dialog message,
+ * Maximum number of lines to include from database upgrade message,
  * to work around the fact that we can't guarantee a scrollable text
- * box for it.
+ * box for it when displaying in dialog boxes.
  */
-const MAX_MESSAGE_LINES = 20;
+const MAX_UPGRADE_MESSAGE_LINES = 10;
 
 /**
  * queries.ts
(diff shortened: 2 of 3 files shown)


jcreedcmu:
The "Pin nodejs version during Actions build" commit
solves the problem of whatever node/npm ubuntu-latest happens to
have in /usr/local/bin/{node,npm} producing Error: Cannot find module 'semver' errors when it is misconfigured in the image.

(cf. https://askubuntu.com/questions/1152570/npm-cant-find-module-semver-error-in-ubuntu-19-04)

extensions/ql-vscode/src/queries.ts
```diff
@@ -273,7 +280,13 @@ async function checkAndConfirmDatabaseUpgrade(qs: qsClient.QueryServerClient, db
 
   logger.log(descriptionMessage);
   // Ask the user to confirm the upgrade.
-  const shouldUpgrade = await helpers.showBinaryChoiceDialog(`Should the database ${db.databaseUri.fsPath} be upgraded?\n\n${descriptionMessage}`);
+
+  let messageLines = descriptionMessage.split('\n');
+  if (messageLines.length > MAX_MESSAGE_LINES) {
+    messageLines = messageLines.slice(0, MAX_MESSAGE_LINES);
```
henrymercer: Nit: similarly I think this should be `MAX_MESSAGE_LINES - 3`.

extensions/ql-vscode/src/queries.ts
```diff
@@ -273,7 +280,13 @@ async function checkAndConfirmDatabaseUpgrade(qs: qsClient.QueryServerClient, db
...
+    messageLines.push("... [full list of changes can be found in Output > CodeQL Extension Log]");
```
henrymercer: It might be helpful here to explain a little more about how to get to the CodeQL Extension Log, for example like we do on compilation error messages.  Another option could be adding a button that focuses the log.

extensions/ql-vscode/src/queries.ts
```diff
@@ -273,7 +280,13 @@ async function checkAndConfirmDatabaseUpgrade(qs: qsClient.QueryServerClient, db
 
   logger.log(descriptionMessage);
   // Ask the user to confirm the upgrade.
-  const shouldUpgrade = await helpers.showBinaryChoiceDialog(`Should the database ${db.databaseUri.fsPath} be upgraded?\n\n${descriptionMessage}`);
+
+  let messageLines = descriptionMessage.split('\n');
+  if (messageLines.length > MAX_MESSAGE_LINES) {
```
henrymercer: Nit: per the JSDoc, I think this should be `MAX_MESSAGE_LINES - 2` since we add two extra lines in line 289.

extensions/ql-vscode/src/queries.ts
```diff
@@ -15,6 +15,13 @@ import { promisify } from 'util';
 import { QueryHistoryItemOptions } from './query-history';
 import { isQuickQueryPath } from './quick-query';
 
+/**
+ * Maximum number of lines to put in a binary choice dialog message,
+ * to work around the fact that we can't guarantee a scrollable text
+ * box for it.
+ */
+const MAX_MESSAGE_LINES = 20;
```
henrymercer: Do we know that 20 lines is going to fit on small netbook-style screens?  I think if necessary we could afford to be more conservative here, since we provide the full output in the extension log.

.github/workflows/main.yml
```diff
@@ -14,6 +14,10 @@ jobs:
         with:
           fetch-depth: 1
 
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '10.18.1'
```
henrymercer: OOI how did you choose this version?

extensions/ql-vscode/src/queries.ts
jcreedcmu: I was just thinking of it as a coarse-grained limit on the size of the information coming from the database upgrade script. Due to line wrapping this isn't going to be exact anyway.

extensions/ql-vscode/src/queries.ts
jcreedcmu: Yeah, a button to cancel-but-show-log sounds like a good idea.

extensions/ql-vscode/src/queries.ts
jcreedcmu: I'm fine with being more conservative, like 10.

.github/workflows/main.yml
jcreedcmu: The current rush configuration appeared to want 10.x, so I declined to use the latest version at the time of writing 12.4.1. This is the currently latest available 10.x, according to running `actions/setup-node` and looking at `node -v`

extensions/ql-vscode/src/queries.ts
henrymercer: Makes sense, in that case could we rename / document that this limit is coarse-grained / a heuristic?

jcreedcmu: :+1:

extensions/ql-vscode/src/queries.ts
```diff
@@ -281,12 +281,26 @@ async function checkAndConfirmDatabaseUpgrade(qs: qsClient.QueryServerClient, db
...
+  const noItem = { title: 'No', isCloseAffordance: true }
+  let dialogOptions: vscode.MessageItem[] = [yesItem, noItem];
+
   let messageLines = descriptionMessage.split('\n');
-  if (messageLines.length > MAX_MESSAGE_LINES) {
-    messageLines = messageLines.slice(0, MAX_MESSAGE_LINES);
-    messageLines.push("... [full list of changes can be found in Output > CodeQL Extension Log]");
+  if (messageLines.length > MAX_UPGRADE_MESSAGE_LINES) {
+    messageLines = messageLines.slice(0, MAX_UPGRADE_MESSAGE_LINES);
+    messageLines.push("... [truncating list of upgrades, click 'No, Show Changes' see full list]");
```
henrymercer:
```suggestion
    messageLines.push("The list of upgrades was truncated, click "No, Show Changes" to see the full list.");
```

extensions/ql-vscode/src/queries.ts
```diff
@@ -281,12 +281,26 @@ async function checkAndConfirmDatabaseUpgrade(qs: qsClient.QueryServerClient, db
...
+
+  const message = `Should the database ${db.databaseUri.fsPath} be upgraded?\n\n${messageLines.join("\n")}`;
+  const chosenItem = await vscode.window.showInformationMessage(message, { modal: true }, ...dialogOptions);
+
+  if (chosenItem === showLogItem) {
+    logger.outputChannel.show();
+  }
+
+  const shouldUpgrade = chosenItem === yesItem;
   if (shouldUpgrade) {
```
henrymercer: Nit: inline `shouldUpgrade`

PR #229, opened 2020-02-10
Exclude documentation from CI workflow (WIP don't merge)

jcreedcmu:
Actions appears to be computing the diff with respect to the pull request base, so that a commit will trigger the workflow if the pull request as a whole contains any changes that modify non-markdown files, which is slightly unfortunate, but still an improvement for tiny documentation PRs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,13 @@
 name: Build Extension
-on: [push, pull_request]
+on:
+  push:
+    paths:
+    - '**'
+    - '!**.md'
+  pull_request:
+    paths:
+    - '**'
+    - '!**.md'
 
 jobs:
   build:


.github/workflows/main.yml
```diff
@@ -1,5 +1,13 @@
 name: Build Extension
-on: [push, pull_request]
+on:
+  push:
+    paths:
+    - '**'
+    - '!**.md'
```
henrymercer: This could be simplified to a single line with `paths-ignore`.

jcreedcmu: Could, but I prefer this mechanism which explicitly gives the condition for running CI as "any of $these, but not $these" rather than a simple ignore --- for, empirically, I found that just doing `!**.extension` without a positive '**' before it had the consequence of CI *never* running, and even though `paths-ignore` probably does the right thing by itself, it's less obvious to me as a hypothetical reader-of-this-file-6-months-from-now that it does, if that makes sense?

PR #230, opened 2020-02-11
Fix weird behavior of Quick Query in non-multi-root workspaces

jcreedcmu:
Starting from a single-root workspace, Quick Query tries to add a workspace folder which points into workspace state, which then causes an extension host restart, and a different workspace state, leading to two Quick Query workspace folders.

The restart upon switching from single-root to multi-root appears to be unavoidable, and interrupts extension code execution, see https://github.com/microsoft/vscode/issues/69335#issuecomment-521105295 for details.


aeisenberg:
I feel confident enough now to approve this request. It looks fine,

PR #241, opened 2020-02-13
Trim prefix from git ref name during release

jcreedcmu:
This was tested by pushing a dummy tag `v9.8.7` and seeing that the correct artifact name was produced.


.github/workflows/release.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           echo "::set-output name=vsix_path::$VSIX_PATH"
           # Transform the GitHub ref so it can be used in a filename.
           # This is mainly needed for testing branches that modify this workflow.
```
henrymercer:
```suggestion
          # The last sed invocation is used for testing branches that modify this workflow.
```

PR #246, opened 2020-02-20
Add query history items at query start time

### LINKED ISSUE #277: Use pagination for raw and interpreted results
With this we can (hopefully) remove the limits on result counts.

Migrated from https://github.com/github/codeql-coreql-team/issues/195

jcreedcmu:
Add query history items right when the user initiates a query, rather
than waiting until query compilation/execution has finished. This
leads to a more predictable ordering of query history items.

This doesn't affect user-facing behavior other than timing.

Fixes https://github.com/github/codeql-coreql-team/issues/277

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
         os: [ubuntu-latest, windows-latest]
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
           fetch-depth: 1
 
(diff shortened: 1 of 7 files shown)


jcreedcmu:
The statement
> There is no visual indication that a query history item is in-progress
isn't *quite* true --- I did add the string "Running" in the query history view itself, though it might not be visible depending on the size the user has chosen for that side panel. Otherwise you're absolutely correct about the results view getting out of sync.

I think it might be simpler if the results view actually presents some visual that conveys 'this query terminated with an error' (maybe even showing a useful error message) or a spinner animation or something to indicate that the query is still running, if it is? Selecting the item for the just-started query (before it has finished) makes sense to me, *if* the results view changes from 'still running spinner' to the actual results when the query finishes.

henrymercer:
For the use case of progressively refining a QL query, I think continuing to show the previous results until the new results are available (instead of displaying a spinner in the results view as soon as the new query run is started) would be valuable.

jcreedcmu:
Closing for now, need to rethink approach.

PR #248, opened 2020-02-21
Fix version bump PR creation during release workflow

jcreedcmu:
No user-facing changes.

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CodeQL for Visual Studio Code: Changelog
 
+## 1.0.6
+
+- Add command to restart query server.
+
 ## 1.0.5 - 13 February 2020
 
 - Add an icon next to any failed query runs in the query history
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -206,6 +206,10 @@
       {
         "command": "codeQLQueryHistory.setLabel",
         "title": "Set Label"
+      },
+      {
+        "command": "codeQL.restartQueryServer",
+        "title": "CodeQL: Restart Query Server"
       }
     ],
     "menus": {
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -28,6 +28,7 @@
     "onCommand:codeQLDatabases.chooseDatabase",
     "onCommand:codeQLDatabases.setCurrentDatabase",
     "onCommand:codeQL.quickQuery",
+    "onCommand:codeQL.restartQueryServer",
     "onWebviewPanel:resultsView",
     "onFileSystem:codeql-zip-archive"
   ],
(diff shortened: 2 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -38,6 +38,7 @@ jobs:
       - name: Fetch master branch
         run: |
           git fetch --depth=1 origin master:master
+          git checkout master
```
jcreedcmu: Apparently we simply need to actually checkout master to avoid that error about the PR base being invalid.

jcreedcmu:
The other version bumps are required, I think, to keep up with underlying API changes in github actions? I get errors without them that I don't fully understand.

.github/workflows/release.yml
henrymercer: I think this step should be just before the "Bump patch version" step, since we want to build and release the extension based on the commit with the tag, whereas we want the version bump PR to be based on `master`.

henrymercer:
> The other version bumps are required, I think, to keep up with underlying API changes in github actions? I get errors without them that I don't fully understand.

Bumping `checkout` from v1 to v2 and bumping `peter-evans/create-pull-request` seems completely reasonable.  I'm surprised downgrading `checkout` from master to v2 in `.github/workflows/release.yml` was required, but it seems like a reasonable thing to do anyway.

.github/workflows/release.yml
jcreedcmu: Good call - those are extensionally the same as I've been doing the release, but better to match intension as well.

jcreedcmu:
> I'm surprised downgrading checkout from master to v2 in .github/workflows/release.yml was required, but it seems like a reasonable thing to do anyway.

Indeed, it wasn't required afaik, but settling on a single, released version of the action everywhere in the config seemed like a reasonable thing to do.

.github/workflows/release.yml
```diff
@@ -99,6 +93,13 @@ jobs:
           asset_name: ${{ format('vscode-codeql-{0}.vsix', steps.prepare-artifacts.outputs.ref_name) }}
           asset_content_type: application/zip
 
+      # The checkout action does not fetch the master branch.
+      # Fetch the master branch so that we can base the version bump PR against master.
+      - name: Fetch master branch
```
henrymercer:
```suggestion
      - name: Checkout master branch
```

Nit

.github/workflows/main.yml
```diff
@@ -10,7 +10,7 @@ jobs:
         os: [ubuntu-latest, windows-latest]
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
```
adityasharad: Use the SHA corresponding to the version tag (here and later).

jcreedcmu: I think I'm persuaded that for `actions/...` actions, it's better to effectively pinning to `2.*.*` as we do here, as I judge it more likely that we will thereby *get* security fixes, rather than the less likely event that github's `actions` account is compromised.

PR #252, opened 2020-02-27
Relax CLI version constraint to allow minor version increases

jcreedcmu:
No changes to UI.

Fixes https://github.com/github/vscode-codeql/issues/253.


extensions/ql-vscode/src/distribution.ts
```diff
@@ -39,9 +39,9 @@ const DEFAULT_DISTRIBUTION_REPOSITORY_NAME = "codeql-cli-binaries";
  * This applies to both extension-managed and CLI distributions.
  */
 export const DEFAULT_DISTRIBUTION_VERSION_CONSTRAINT: VersionConstraint = {
-  description: "2.0.*",
+  description: "2.*.*",
   isVersionCompatible: (v: Version) => {
-    return v.majorVersion === 2 && v.minorVersion === 0
+    return v.majorVersion === 2 && v.minorVersion >= 0
```
jcreedcmu: Just leaving the trivial `>=0` in in case we want to require a certain minor version in the future.

PR #260, opened 2020-03-04
Add compiled grammar to top level for use in linguist

alexet:
Linguist needs a texmate grammar in a repository. The grammar is just a slightly modified version of the vscode one (the markdown grammar is different and doesn't have the same subrules). It seems easier to put it here where in future we can make it track the vscode grammar changes.

I plan to add a case to the action to fail if running the script changes the file but we want to get linguist support done soon.

### DIFF
--- a/syntaxes/README.md
+++ b/syntaxes/README.md
@@ -1,0 +1,3 @@
+This folder contains a compiled version of the textmate grammar for use with systems that need a compiled verison of the grammar in the repository such as linguist. It also contains a patvh for the grammar to make it work with linguist.
+
+To update the grammar, first build the extension, then run "./updateSyntax".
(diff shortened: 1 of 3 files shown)


syntaxes/README.md
```diff
@@ -0,0 +1,3 @@
+This folder contains a compiled version of the textmate grammar for use with systems that need a compiled verison of the grammar in the repository such as linguist. It also contains a patvh for the grammar to make it work with linguist.
```
jcreedcmu:
```suggestion
This folder contains a compiled version of the textmate grammar for use with systems that need a compiled verison of the grammar in the repository such as linguist. It also contains a patch for the grammar to make it work with linguist.
```

alexet:
I have accepted your suggestion.

jcreedcmu:
Pity we can't expect people have `jq`, but the perl script is fine by me for now.

jcreedcmu:
(with jq 1.6, one could do
```
jq 'walk(if type == "array" then map(select((type == "object" and has("include") and (.include | test("^text.html.markdown#" )) ) | not)) else . end)' /tmp/repo.json
```
to remove all objects with `"include": "text.html.markdown#..."`)

PR #262, opened 2020-03-05
Retry integration tests when they fail

jcreedcmu:
Integration tests have been flaky. I tracked down the problem to the vscode electron binary sometimes segfaulting *after* all tests have been successfully run. The information the test runner gets from `vscode-test` unfortunately does not distinguish this segfault from test failure. See comments for what we could do about this.

I tried bisecting which integration test, if any, was causing the problem, without much success --- locally I get a segfault rate of about 0.5%-5% depending on which tests are present. It seems like the fewer tests are included in the suite, the fewer segfaults, but artificially duplicating tests doesn't make the segfaults tractably common. A minimal commandline to (possibly) trigger the segfault, after `npm run integration` has been run at least once, is
```
$PATH_UP_TO/vscode-codeql/.vscode-test/vscode-1.42.1/VSCode-linux-x64/code  \
  --disable-extensions \
  --extensionDevelopmentPath=$PATH_UP_TO/vscode-codeql/extensions/ql-vscode \
  --extensionTestsPath=$PATH_UP_TO/vscode-codeql//extensions/ql-vscode/out/vscode-tests/no-workspace/index
```
I wrapped this in a bash while loop waiting for `$?` to be 139. The `.vscode-test/vscode-1.42.1/VSCode-linux-x64/code` binary downloaded by `vscode-test` is stripped of debugging symbols so I wasn't able to get any insight from the coredump.

### DIFF
--- a/extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
+++ b/extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
@@ -29,7 +29,7 @@ async function runTestsWithRetry(suite: Suite, tries: number): Promise<void> {
     } catch (err) {
       console.error(`Exception raised while running tests: ${err}`);
       if (t < tries - 1)
-        console.error('Retrying...');
+        console.log('Retrying...');
     }
   }
   console.error(`Tried running suite ${tries} time(s), still failed, giving up.`);
@@ -67,10 +67,10 @@ async function main() {
     ];
 
     for (const integrationTestSuite of integrationTestSuites) {
-      await runTestsWithRetry(integrationTestSuite, 2);
+      await runTestsWithRetry(integrationTestSuite, 3);
     }
   } catch (err) {
-    console.error('Unexpected exception while running tests');
+    console.error(`Unexpected exception while running tests: ${err}`);
     process.exit(1);
   }
 }
--- a/extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
+++ b/extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
@@ -1,6 +1,41 @@
 import * as path from 'path';
 import { runTests } from 'vscode-test';
 
+// A subset of the fields in TestOptions from vscode-test, which we
+// would simply use instead, but for the fact that it doesn't export
+// it.
+type Suite = {
+  extensionDevelopmentPath: string,
+  extensionTestsPath: string,
+  launchArgs: string[]
+};
+
+/**
+ * Run an integration test suite `suite` at most `tries` times, or
+ * until it succeeds, whichever comes first.
+ *
+ * TODO: Presently there is no way to distinguish a legitimately
+ * failed test run from the test runner being terminated by a signal.
+ * If in the future there arises a way to distinguish these cases
+ * (e.g. https://github.com/microsoft/vscode-test/pull/56) only retry
+ * in the terminated-by-signal case.
+ */
+async function runTestsWithRetry(suite: Suite, tries: number): Promise<void> {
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
```diff
@@ -1,6 +1,41 @@
...
+  extensionDevelopmentPath: string,
+  extensionTestsPath: string,
+  launchArgs: string[]
+};
+
+/**
+ * Run an integration test suite `suite` at most `tries` times, or
+ * until it succeeds, whichever comes first.
+ *
+ * TODO: Presently there is no way to distinguish a legitimately
```
henrymercer: Suggest we add an issue to track this

extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
```diff
@@ -32,11 +67,10 @@ async function main() {
     ];
 
     for (const integrationTestSuite of integrationTestSuites) {
-      // Download and unzip VS Code if necessary, and run the integration test suite.
-      await runTests(integrationTestSuite);
+      await runTestsWithRetry(integrationTestSuite, 2);
```
henrymercer: Does running with 2 tries work reliably?  Anecdotally, the integration tests seem to have a surprisingly high probability of failing in the GitHub Actions environment.

extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
```diff
@@ -32,11 +67,10 @@ async function main() {
...
     }
   } catch (err) {
-    console.error('Failed to run tests');
+    console.error('Unexpected exception while running tests');
```
henrymercer: Worth logging the error too here?

extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
```diff
@@ -1,6 +1,41 @@
...
+async function runTestsWithRetry(suite: Suite, tries: number): Promise<void> {
+  for (let t = 0; t < tries; t++) {
+    try {
+      // Download and unzip VS Code if necessary, and run the integration test suite.
+      await runTests(suite);
+      return;
+    } catch (err) {
+      console.error(`Exception raised while running tests: ${err}`);
+      if (t < tries - 1)
+        console.error('Retrying...');
```
henrymercer:
```suggestion
        console.log('Retrying...');
```

extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
```diff
@@ -1,6 +1,41 @@
 import * as path from 'path';
 import { runTests } from 'vscode-test';
 
+// A subset of the fields in TestOptions from vscode-test, which we
+// would simply use instead, but for the fact that it doesn't export
+// it.
+type Suite = {
```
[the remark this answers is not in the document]
jcreedcmu: All the fields of `TestOptions` happen to already be optional; the issue is that https://github.com/microsoft/vscode-test/blob/master/lib/index.ts doesn't export it.

extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
jcreedcmu: I mean, it's a tradeoff between how much pointless work we want to do when a test is actually failing. I'm fine bumping this to 3. I have seen it fail twice in a row exactly once locally, but yeah, the rate of failure seems a little higher on CI.

henrymercer:
No docs impact

PR #265, opened 2020-03-05
Change vscode settings to hide generated files

aeisenberg:
This change updates the vscode settings to hide generated files. If this is not the way you want to work or that this is not appropriate for project settings, I can just move this to my global settings. However, I think these changes are generally useful to all contributors.


aeisenberg:
Found a few more files we should ignore and going to update this.

PR #266, opened 2020-03-06
Display timeout warning and display canceled queries

### LINKED ISSUE #250: Timeout during query compilation should produce failed query history item
**Describe the bug**
When query compilation times out, the message `The request (id: 6, method: 'compilation/compileQuery') has been cancelled` is added to the extension log, but a query history item corresponding to the timed out query is _not_ added to the query history view.

**To Reproduce**
1. Set `codeQL.runningQueries.timeout` to 1 second
1. Run a long-running query e.g. a path problem query
1. Focus the query history view
1. Observe that the query history view doesn't contain an item corresponding to the timed-out compilation.

**Expected behavior**
The query history view should contain an item corresponding to the timed-out compilation with a failed icon and the status "timed out".

### LINKED ISSUE #251: Timeout during query compilation or query evaluation should produce notification
**Describe the bug**
If a query compilation or a query evaluation times out, the extension should produce a notification to inform the user of the timeout.

**To Reproduce**
1. Set `codeQL.runningQueries.timeout` to 1 second
1. Run a long-running query e.g. a path problem query
1. Observe that the query run notification disappears, indicating that the query run has terminated, however there is no notification informing the user of the timeout.

**Expected behavior**
When the query run notification disappears, an error notification should be displayed indicating that the query run has timed out.

aeisenberg:
fix: Display queries that are canceled during compilation

This change converts a cancelled query into a synthetic query result
that is displayed in query history.

Also includes some light refactoring.

fix: Display message when query times out

Also, add the message to the log.

Closes #250 
Closes #251


jcreedcmu:
Oops, don't believe this approval yet, I clicked in the wrong tab.

extensions/ql-vscode/src/query-results.ts
```diff
@@ -68,7 +68,7 @@ export class CompletedQuery implements QueryWithResults {
         return `timed out after ${this.result.evaluationTime / 1000} seconds`;
       case messages.QueryResultType.OTHER_ERROR:
       default:
-        return `failed`;
+        return `failed: ${this.result.message || ''}`;
```
jcreedcmu: Either replace empty string with some nonce like 'unknown error' or 'no error message' or something, or put the colon inside the conditional.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -448,20 +469,30 @@ export async function compileAndRunQueryAgainstDatabase(
...
+  db: DatabaseItem,
+  historyItemOptions: QueryHistoryItemOptions,
+  message: string
+) {
+
+  return {
+    query,
+    result: {
+      evaluationTime: 0,
+      resultType: messages.QueryResultType.CANCELLATION,
```
jcreedcmu: I think you want `resultType` to be a parameter of `createSyntheticResult`? This should still be `messages.QueryResultType.OTHER_ERROR` if we arrive here from the non-cancellation fallthrough (your diff's line 472).

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -411,10 +418,24 @@ export async function compileAndRunQueryAgainstDatabase(
...
+  } catch (e) {
+    if (e instanceof ResponseError && e.code == ErrorCodes.RequestCancelled) {
+      return createSyntheticResult(query, db, historyItemOptions, 'Query cancelled');
+    } else {
+      throw e;
+    }
+  }
 
   if (errors.length == 0) {
     const result = await query.run(qs);
```
jcreedcmu: Do you want to take care of cancelled running queries right here too?

aeisenberg: Yes!

aeisenberg: Hmmm... `QueryResultType` is a namespace, which means that we can't pass it as a parameter type. Seems to me we should change it to an enum. Is there any reason why it's not?

aeisenberg: Done. Will include the change in my next push.

PR #270, opened 2020-03-09
Convert namespaces to enums in messages.ts

jcreedcmu:
So we can refer to their types if necessary.


aeisenberg:
This all looks reasonable to me, but someone with more experience should approve.

PR #271, opened 2020-03-09
Revert "Convert namespaces to enums in messages.ts"

henrymercer:
The queryserver protocol only commits to the error being a number, so that a new error type can be added to the CodeQL CLI without a major version bump.  Therefore these values should be namespaces not enums to reflect the fact they are not exhaustive.

Reverts github/vscode-codeql#270

### DIFF
--- a/extensions/ql-vscode/src/messages.ts
+++ b/extensions/ql-vscode/src/messages.ts
@@ -299,15 +299,15 @@ export interface CompilationMessage {
 /**
  * Severity of different messages
  */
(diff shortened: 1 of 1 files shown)


henrymercer:
@jcreedcmu It might be an idea to reflect the non-exhaustive nature of these in the docstrings, what do you think?

aeisenberg:
And a reason why these shouldn't be enums so we don't need to have this conversation again.

Issue #273, opened 2020-03-09
Link to query server log when query compilation fails


aeisenberg:
From https://github.com/github/codeql-coreql-team/issues/287

We could save the user a bit of time by adding a button to this error message dialog that opens the query server log.

aeisenberg:
This would be nice, especially for users not familiar with the vscode interface.

jcreedcmu:
@shati-patel is this necessary to document as well?

aeisenberg:
Is this sufficient? 

https://github.com/github/vscode-codeql/pull/274/files#diff-8c3199aebb6f791ef5520c0ae3976ec9R5

shati-patel:
Thanks for the mention! The changelog should be sufficient in this case since it's a nice, self-explanatory change 😊

PR #274, opened 2020-03-09
Add "Show log" button for all messages

### LINKED ISSUE #287: Deprecate `codeql.cmd` on windows
The extension has moved away from calling `codeql.cmd` in the cli. Instead, it calls `codeql.exe` by default. Eventually, the `cmd` file will be removed from the cli. However, there are some caveats we still need to handle:

1. User has an old version of cli installed that does not have `codeql.exe` available.
2. User has explicitly specified `codeql.cmd` as their executable path.

For 1, we should display a warning and suggestion to upgrade and fall back to the `cmd` file. For 2, we again display a warning, but use it anyway.

aeisenberg:
This change ensures that "Show log" is available on all messages
from the extension. It's important to note that the only place that
was specifying an "item" before was doing it incorrectly. That's
been fixed.

Closes #287

## Checklist

- [x] [CHANGELOG.md](../extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.

### DIFF
--- a/extensions/ql-vscode/src/databases.ts
+++ b/extensions/ql-vscode/src/databases.ts
@@ -571,7 +571,7 @@ export class DatabaseManager extends DisposableObject {
       }
     } catch (e) {
       // database list had an unexpected type - nothing to be done?
-      showAndLogErrorMessage('Database list loading failed: ${}', e.message);
+      showAndLogErrorMessage(`Database list loading failed: ${e.message}`);
     }
   }
 
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -201,7 +201,9 @@ export async function activate(ctx: ExtensionContext): Promise<void> {
     } else if (distributionResult.kind === FindDistributionResultKind.NoDistribution) {
       registerErrorStubs([checkForUpdatesCommand], command => async () => {
         const installActionName = "Install CodeQL CLI";
-        const chosenAction = await helpers.showAndLogErrorMessage(`Can't execute ${command}: missing CodeQL CLI.`, installActionName);
+        const chosenAction = await helpers.showAndLogErrorMessage(`Can't execute ${command}: missing CodeQL CLI.`, {
+          items: [installActionName]
+        });
         if (chosenAction === installActionName) {
           installOrUpdateThenTryActivate({
             isUserInitiated: true,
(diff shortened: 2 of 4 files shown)


aeisenberg:
Added the changelog update and this includes adding a few lines for changes in previous PRs that I forgot to add.

extensions/ql-vscode/src/databases.ts
```diff
@@ -571,7 +571,7 @@ export class DatabaseManager extends DisposableObject {
       }
     } catch (e) {
       // database list had an unexpected type - nothing to be done?
-      showAndLogErrorMessage('Database list loading failed: ${}', e.message);
+      showAndLogErrorMessage(`Database list loading failed: ${e.message}`);
```
jcreedcmu: Oh, good catch.

extensions/ql-vscode/src/extension.ts
```diff
@@ -319,10 +319,7 @@ async function activateWithInstalledDistribution(ctx: ExtensionContext, distribu
   ctx.subscriptions.push(commands.registerCommand('codeQL.quickQuery', async () => displayQuickQuery(ctx, cliServer, databaseUI)));
   ctx.subscriptions.push(commands.registerCommand('codeQL.restartQueryServer', async () => {
     await qs.restartQueryServer();
-    const response = await Window.showInformationMessage('CodeQL Query Server restarted.', 'Show Log');
-    if (response === 'Show Log') {
-      qs.showLog();
-    }
+    helpers.showAndLogInformationMessage('CodeQL Query Server restarted.');
```
jcreedcmu: Sorry, my fault for not being clear; the helper just needs an extra (optional? defaulting to `logging.logger`?) `OutputChannelArgument` argument or something, since the log stream that this existing 'Show Log' button shows (which I think is what we want) is the one named 'CodeQL Query Server', rather than the default 'CodeQL Extension Log', which is what `helpers.ts` is getting from `logging.logger`.

aeisenberg: Darn. You're right. I wasn't reading carefully enough. I'll fix. Though, it's a little more complex since the last argument is varargs, but I should be able to do something.

jcreedcmu: If we're making the wrappers around the underlying varargs methods on `vscode.window`, I wouldn't mind at all refactoring to make them take an explicit array argument instead.

aeisenberg: What about if it takes a `message` arg and an `options` arg. This will make it more flexible and future-proof since I'm sure we'll want to augment with different behaviour later.

jcreedcmu: sure, sounds fine

extensions/ql-vscode/src/helpers.ts
```diff
@@ -47,42 +47,45 @@ export function withProgress<R>(
...
-export async function showAndLogInformationMessage(message: string, ...items: string[]): Promise<string | undefined> {
-  return internalShowAndLog(message, Window.showInformationMessage, ...items);
+export async function showAndLogInformationMessage(message: string, outputLogger = logger, ...items: string[]): Promise<string | undefined> {
+  return internalShowAndLog(message, items, outputLogger, Window.showInformationMessage);
 }
 
-async function internalShowAndLog(message: string, fn: Function, ...items: string[]): Promise<string | undefined> {
-  logger.log(message);
-  const label = 'Show log';
+async function internalShowAndLog(message: string, items: string[], outputLogger = logger, fn: Function): Promise<string | undefined> {
```
jcreedcmu: Totally a nit, but I think I'd prefer the specific type `(message: string, label: string, ...items: string[]) => Thenable<string | undefined>` to just `Function`.

aeisenberg:
Also, Closes #273

PR #281, opened 2020-03-12
chore: Introduce eslint

### LINKED ISSUE #238: Set up a linter to standardise code formatting
To save time when reviewing PRs by reducing formatting changes and discrepancies, we should set up a linter or similar to automatically standardise the formatting of our code.

aeisenberg:
Adds eslint support and fixes linting problems in a few files.

This change adds an npm task, but does not enforce linting for builds.

The idea is to slowly fix linting problems over time.

Closes #238.


aeisenberg:
> If possible, I would prefer to allow anonymous => functions to not explicitly specify their return type (but requiring return types for named functions and methods seems good)

I wasn't too fond of this either. Simple functions _should_ have their types inferred. What I really want to avoid is implicit _any_ functions. I think it's a compiler setting. I'll play around and see what I can find. 

> I wanted to say that I prefer ,s instead of ;s in record types {a: number, b: string} since it lines up with the syntax of values, but I see the typescript language server I'm using seems to canonicalize that to

Using `;` is in the specs. Using `,` is tolerated.

PR #282, opened 2020-03-12
fix: Readable error message for invalid file

aeisenberg:
Display an understandable error message when trying to Run Query on
a qll file. Closes https://github.com/github/codeql-coreql-team/issues/201

### DIFF
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -337,7 +337,7 @@ async function determineSelectedQuery(selectedResourceUri: vscode.Uri | undefine
   const queryPath = queryUri.fsPath || '';
 
   if (!queryPath.endsWith('.ql')) {
-    throw new Error('The selected resource is not a QL file.');
+    throw new Error('The selected resource is not a CodeQL query file; It should have the extension ".ql".');
   }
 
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/run-queries.ts
```diff
@@ -334,7 +334,11 @@ async function determineSelectedQuery(selectedResourceUri: vscode.Uri | undefine
   if (queryUri.scheme !== 'file') {
     throw new Error('Can only run queries that are on disk.');
   }
-  const queryPath = queryUri.fsPath;
+  const queryPath = queryUri.fsPath || '';
+
+  if (!queryPath.endsWith('.ql')) {
+    throw new Error('The selected resource is not a QL file.');
```
jcreedcmu:
```suggestion
    throw new Error('The selected resource is not a CodeQL query file; It should have the extension ".ql".');
```

jcreedcmu: Or... maybe some variant of that. 
`The selected resource is not a CodeQL query file. (extension ".ql")`
`The selected resource is not a CodeQL query file. CodeQL queries have the extension ".ql".`
I mostly just want to avoid the confusion that `.qll` files are arguably, at least to a newcomer, a kind of QL file, in that they are QL libraries.

PR #284, opened 2020-03-16
feat: Save log files per query

aeisenberg:
fix for https://github.com/github/codeql-coreql-team/issues/236
fix for https://github.com/github/codeql-coreql-team/issues/234

This feature adds logging per-query. Each query will be logged in its
own location in either workspace or globally shared location in
vscode.

There are limitations here. We are only guessing when one query ends
and another begins. We assume that queries don't occur in parallel.
If they do, the previous query will have its results intermingled
with the current query's results.

To fix that, we will need to update how the query-server emits log
messages so that each query message is attached to a tag that
specifies the query that emitted it.

This PR is a partial fix for the two issues above. A complete fix will also allow the side logs to be available from the UI.


.github/workflows/main.yml
```diff
@@ -80,7 +80,7 @@ jobs:
         if: matrix.os == 'windows-latest'
         run: |
           cd extensions/ql-vscode
-          $env:CODEQL_PATH=$(Join-Path $env:GITHUB_WORKSPACE -ChildPath 'codeql-home/codeql/codeql.cmd')
```
jcreedcmu: Although this seems like the right thing to do for now, I'm just a little worried at the divergence of the test environment from the actual user experience.

jcreedcmu: Maybe a comment/issue documenting why CI appears to be quarantining the executable, and what else we speculate we could do about that if something changes about Actions's VM setup in the future, or something?

aeisenberg: I think we are trying to move away from using the `codeql.cmd` file. If you look at the contents of the file, it is just a thing wrapper around calling the executable. The failing build always had to do with the `exe` file, so moving back to calling `cmd` should have no effect.

We should consider explicitly deprecating (and later removing) the `cmd` file. I'll bring that up as a discussion point.

jcreedcmu: Ah, ok, I'd misunderstood. This is totally fine then.

extensions/ql-vscode/src/logging.ts
```diff
@@ -1,42 +1,124 @@
...
 
-  logWithoutTrailingNewline(message: string) {
-    this.outputChannel.append(message);
+  /**
+   * This function is asynchronous and will only resolve once the message is written
+   * to the side log (if required). It is not necessary to await the results of this
+   * function if you don't need to guarantee that the log writing is complete before
+   * continuing.
+   */
+  async log(message: string, options = { } as LogOptions): Promise<void> {
```
jcreedcmu: This idiom makes sense to me, but I think we may have a LGTM-based CI hook that complains about calls to async functions in void context that don't await... so we might want a wrapper function `logAsync` or some such around doing exactly that which includes whatever magic comment makes LGTM ignore it. Not a request for a change, just a heads-up.

jcreedcmu: (at least I think that was the mechanism that caught an actual missing-`await` bug of mine a while ago...)

extensions/ql-vscode/src/logging.ts
```diff
@@ -1,42 +1,124 @@
...
+      options.trailingNewline = true;
+    }
+
+    if (options.trailingNewline) {
+      this.outputChannel.appendLine(message);
+    } else {
+      this.outputChannel.append(message);
+    }
+
+    if (this.additionalLogLocationPath && options.additionalLogLocation) {
```
jcreedcmu: nit: should these be tests for `!== undefined` rather than relying on falsiness?

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -157,8 +165,10 @@ export class QueryServerClient extends DisposableObject {
   }
 
   async sendRequest<P, R, E, RO>(type: RequestType<WithProgressId<P>, R, E, RO>, parameter: P, token?: CancellationToken, progress?: (res: ProgressMessage) => void): Promise<R> {
-    let id = this.nextProgress++;
-    this.progressCallbacks[id] = progress;
+    const id = this.nextProgress++;
+    this.progressCallbacks[id] = progress; // parameter.queryToCheck.queryPath (last segment)
```
jcreedcmu: Can you expand this comment a bit?

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -168,4 +178,24 @@ export class QueryServerClient extends DisposableObject {
...
+  /**
+   * Updates the active query every time there is a new request to compile.
+   * The active query is used to specify the side log.
+   *
+   * This isn't ideal because in situations where there are queries running
+   * in parallel, their output is interleaved, but because the activeQuery
+   * is fixed, the previous query will have results mixed in with the active
+   * query's side log.
+   *
+   * The ideal solution would be to extract the active query from the log
```
jcreedcmu: I think can simplify and update this comment to reflect that in situations where there are queries running in parallel, the logs that are emitted are fundamentally not easily assignable to individual queries.

extensions/ql-vscode/test/pure-tests/logging.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+const proxyquire = pq.noPreserveCache().noCallThru();
+chai.should();
+chai.use(sinonChai);
+const expect = chai.expect;
+
+describe('OutputChannelLogger tests', () => {
+  let OutputChannelLogger;
+  const tempFolders: Record<string, tmp.DirResult> = {};
+  // let globalStoragePathTmp: tmp.DirResult;
```
jcreedcmu: Are these commented lines supposed to be here?

extensions/ql-vscode/src/logging.ts
aeisenberg: Doesn't seem to be causing any errors right now. There is a eslint config for this. We can enable it. Might be prudent...

extensions/ql-vscode/src/logging.ts
aeisenberg: Hmmm...the way they are, they also catch empty strings, which seems like what we want.

extensions/ql-vscode/src/queryserver-client.ts
aeisenberg: Oops. Should not be there.

extensions/ql-vscode/src/logging.ts
jcreedcmu: Ok, just wanted to be sure it was on purpose.

extensions/ql-vscode/src/queryserver-client.ts
aeisenberg: Oops. Should not be there.

extensions/ql-vscode/test/pure-tests/logging.test.ts
aeisenberg: oops.

extensions/ql-vscode/src/queryserver-client.ts
aeisenberg: Will do.

PR #294, opened 2020-03-18
Fix quick-eval in .qll error

jcreedcmu:


extensions/ql-vscode/src/vscode-tests/minimal-workspace/determining-selected-query-test.ts
```diff
@@ -0,0 +1,44 @@
...
+  const document = await vscode.workspace.openTextDocument(documentPath);
+  await vscode.window.showTextDocument(document!);
+  return document;
+}
+
+export function run() {
+  describe('Determining selected query', async () => {
+    it('should allow ql files to be queried', async () => {
+      const q = await determineSelectedQuery(Uri.parse('file:///tmp/queryname.ql'), false);
+      expect(q.queryPath).to.equal('/tmp/queryname.ql');
```
aeisenberg: This fails on windows. Use `path.separator` instead of `/`.

jcreedcmu: Oof, thanks for noticing.

jcreedcmu:
> One comment, that suggestion be moved to another PR is that we move workspace setup into some sort of chai plugin. And we use the plugin before each test that requires a workspace. And `use`ing the plugin will ensure that the workspace either has been created, or it creates the workspace right then. This way I think we can avoid the requirement on test ordering.

Agree this sounds useful. I'm not sure it entirely avoids keeping test ordering in mind for those tests that depend on "negative" information (`describe(something) { it('should throw an exception if the extension isn't initialized yet')`) but certainly lazily loading the extension if it isn't loaded yet makes sense for tests that depend positively on it being loaded.

extensions/ql-vscode/src/vscode-tests/minimal-workspace/determining-selected-query-test.ts
```diff
@@ -16,20 +16,20 @@ export function run() {
   describe('Determining selected query', async () => {
     it('should allow ql files to be queried', async () => {
       const q = await determineSelectedQuery(Uri.parse('file:///tmp/queryname.ql'), false);
-      expect(q.queryPath).to.equal('/tmp/queryname.ql');
+      expect(q.queryPath).to.equal(path.join('/', 'tmp', 'queryname.ql'));
```
aeisenberg: Is using `/` really platform agnostic? I guess if tests pass, then it's fine.

jcreedcmu: Yeah, that counts as the root directory even in `path.win32`.

PR #295, opened 2020-03-18
lint: Ran the auto-fix command for the linter

aeisenberg:
Just some code cleanup. Nothing big here.


aeisenberg:
Note that this PR is based off of #284 and so it looks like a lot of commits, but it's really only one.

PR #299, opened 2020-03-19
chore: Remove unnecessary package lock

aeisenberg:
I think this file is not needed. It's causing a security warning. Let's get rid of it.


jcreedcmu:
Changed to updating the package lock instead of removing it, for controlled reproducibility of CI installation of rush.

PR #302, opened 2020-03-19
feat: Allow large log files to be opened externally

### LINKED ISSUE #301: Showing log associated to query should work for large query logs
@tausbn reported being unable to view ~448M query log file from the query history, with the only message being that arising from
```
helpers.showAndLogErrorMessage(`Could not open log file ${queryHistoryItem.logFileLocation}`);
```
in `query-history.ts`. He says when he tried to open the same file manually, it worked. We should probably at least log the exception being handled, and if possible, show the log in some way, even if it's without syntax highlighting.

aeisenberg:
If the user tries to open a log file that is too large for vscode's
extension mechanism to handle, reveal the file in the finder/explorer
and let the user open in an external program.

Fixes #301

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -5,6 +5,7 @@ import { CompletedQuery } from './query-results';
 import { QueryHistoryConfig } from './config';
 import { QueryWithResults } from './run-queries';
 import * as helpers from './helpers';
+import { logger } from './logging';
 
 /**
  * query-history.ts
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/query-history.ts
```diff
@@ -191,12 +192,26 @@ export class QueryHistoryManager {
...
     if (queryHistoryItem.logFileLocation) {
+      const uri = vscode.Uri.parse(queryHistoryItem.logFileLocation);
       try {
-        await vscode.window.showTextDocument(vscode.Uri.parse(queryHistoryItem.logFileLocation), {
-          viewColumn: vscode.ViewColumn.Beside
+        await vscode.window.showTextDocument(uri, {
         });
       } catch (e) {
-        helpers.showAndLogErrorMessage(`Could not open log file ${queryHistoryItem.logFileLocation}`);
+        if (e.message.includes('Files above 50MB cannot be synchronized with extensions')) {
```
jcreedcmu: I was going to ask if there was any other signal the exception is the appropriate one to react to other than scraping its message, but it looks like `src/vs/workbench/api/browser/mainThreadDocuments.ts` is throwing an `Error`, not anything more specific :(

extensions/ql-vscode/src/query-history.ts
```diff
@@ -191,12 +192,26 @@ export class QueryHistoryManager {
...
+          const res = await helpers.showBinaryChoiceDialog('File is too large to open in the editor, do you want to open exterally?');
+          if (res) {
+            try {
+              await vscode.commands.executeCommand('revealFileInOS', uri);
+            } catch (e) {
+              logger.log(e.message);
```
jcreedcmu: Maybe a `showAndLogErrorMessage` here, too, so the user isn't surprised by silence when the `revealFileInOS` fails.

extensions/ql-vscode/src/query-history.ts
aeisenberg: Yeah, I was looking for something better, but couldn't find it. And the worst thing is that it looks like it's not _actually_ 50Mb. There are some other triggers for this (like length of token stream), even though the message is the same AFAICT.

extensions/ql-vscode/src/query-history.ts
aeisenberg: Will do.

jcreedcmu:
@github/product-docs-dsp does this need docs?

shati-patel:
Thanks for the ping! I've created a [documentation issue](https://github.com/github/product-documentation/issues/2319) to figure out if/where to document this 😃

PR #308, opened 2020-03-23
fix: Alignment of alerts row in query view

### LINKED ISSUE #237: Text in interpreted results view header appears to have slightly different vertical alignment
**Describe the bug**
The various bits of text in the interpreted results view header appear to have very slightly different vertical alignments:

image

Here, the text on the right hand side is the lowest, followed by the text on the right hand side, followed by the text in the centre.

**To Reproduce**
Run a path query, view the query results, and select "alerts" from the dropdown to view the interpreted results. 

**Expected behavior**
The text in the screenshot should have the same vertical alignment.

aeisenberg:
fixes #237.

### DIFF
--- a/extensions/ql-vscode/src/view/resultsView.css
+++ b/extensions/ql-vscode/src/view/resultsView.css
@@ -27,12 +27,19 @@
 .vscode-codeql__result-table-toggle-diagnostics label,
 .vscode-codeql__result-table-toggle-diagnostics input {
   display: inline-block;
-  vertical-align: middle;
+  vertical-align: baseline;
 }
 .vscode-codeql__result-table-toggle-diagnostics input {
-  margin: 3px 3px 1px 13px;
+  margin: 0 3px 1px 13px;
+}
+.vscode-codeql__result-table-toggle-diagnostics label {
+  margin-top: 1px;
 }
 
+.number-of-results {
+  padding-left: 3em;
+  padding-top: 1px;
+}
 
 .vscode-codeql__result-table th {
   border-top: 1px solid rgba(88,96,105,0.25);
@@ -150,8 +157,4 @@ td.vscode-codeql__path-index-cell {
 
 .octicon-light {
   opacity: 0.6;
-}
-
-.number-of-results {
-  padding-left: 3em;
 }


jcreedcmu:
Yeah, I lean towards thinking that fixing the underlying choice of alignment is preferable to pixel hacking in this case?

aeisenberg:
>  fixing the underlying choice of alignment

Not sure how I would do that without adding some padding. The problem is that this row uses three dom elements that all layout in different ways: a button, span, and select box. I'll take another look, but I'm not confident that this would be easy to fix. I also don't think this is a big enough issue to spend significant time on.

> I'm still seeing a slightly higher baseline for the results predicate dropdown after this PR

I'll look at that too. Looks like it's less than a pixel.

aeisenberg:
The problem is that there is a misalignment is between the select and the span (the leftmost two elements on the row). The classes that you mention, @henrymercer only apply to the input and the label on the right. Even if I change the `vertical-align` property, I'll still need to make a change to either the select or the span. 

So, I came up with something that seems to work ok. It's more complex than before

henrymercer:
@aeisenberg Are you able to repro locally?  I still seem to be getting misalignment when I check out your PR:

image

I wonder whether what we really want is all the text above the results table to be vertically aligned with the middle of the checkbox.  I don't know whether that's achievable nicely (one option could just be reverting to your original implementation) but does that sound like the right goal?

aeisenberg:
Hmmm...I'm seeing something different with the latest of this PR. As far as I can tell, the row is completely aligned.

Not sure what the difference is. Are you on Linux? I'm on mac.

Is this issue a big concern for you? I've already spent more than an hour on this and I'm not sure if it is worth my time to spend much more.

_Extension_Development_Host__-_CodeQL_Query_Results_—_vscode-codeql-starter__Workspace__and_query-history_ts_—_vscode-codeql

henrymercer:
> Not sure what the difference is. Are you on Linux? I'm on mac.

I'm on Mac as well

> Is this issue a big concern for you? I've already spent more than an hour on this and I'm not sure if it is worth my time to spend much more.

No definitely not, it's a nit.

aeisenberg:
@jcreedcmu would you be able to try this out on your linux machine and see whether you can reproduce? Though, I also recommend not spending much more time on this since there are other things to work on.

aeisenberg:
Going to close this unless we can come to some sort of consensus or conclusion here.

PR #309, opened 2020-03-23
feat: Add keybindings

aeisenberg:
Keybindings for show query view and run query.

Closes https://github.com/github/codeql-coreql-team/issues/74

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CodeQL for Visual Studio Code: Changelog
 
+## 1.1.2
+
+- Add key binding `CTRL+K R` to run query on Mac and Windows
+- Add key binding `CTRL+Shift+Q` to open the query view on Mac and Windows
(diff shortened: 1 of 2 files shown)


jcreedcmu:
- are these consistent with internal user expectations about running queries in eclipse? Clearly users can customize but it might be nice if the defaults are already what people expect.
 - per https://code.visualstudio.com/api/references/contribution-points I suspect we want to have a `"mac": "cmd+..."` version (or are you finding on mac that it automatically converts ctrl to cmd?)

aeisenberg:
`cmd+shift+Q` shuts down the computer, so we want to avoid that.

>  are you finding on mac that it automatically converts ctrl to cmd?

No. I just wanted to avoid shutting down someone's machine. :)

I have no biases or expectations as to what the keybindings should be, so I'll just rely on suggestions.

aeisenberg:
There's no consensus as to what the keybindings should be and I don't think this PR adds much. So, I am going to avoid adding keybindings for now.

PR #312, opened 2020-03-24
lint: Add linting to the build

aeisenberg:
This PR ensures all files are linted properly and (I think) adds linting to the build process. The build should fail if there are linting errors.


extensions/ql-vscode/src/databases.ts
```diff
@@ -430,21 +430,23 @@ class DatabaseItemImpl implements DatabaseItem {
...
       logger.log(`Waiting for event ${event} timed out after ${timeoutMs}ms`);
       res(undefined);
       dispose();
     }, timeoutMs);
+    const disposable = event(e => {
+      res(e);
+      dispose();
+    });
+    function dispose() {
+      if (timeout !== undefined) {
```
jcreedcmu: Huh, somehow I didn't think typescript let me forward-reference this function recursively. If it does, these checks for `!== undefined` aren't necessary anymore, are they?

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -70,6 +70,16 @@ function getQuickQueriesDir(ctx: ExtensionContext): string {
   return queriesPath;
 }
 
+
+function updateQuickQueryDir(queriesDir: string, index: number, len: number) {
```
jcreedcmu: Why is this required to be hoisted? I'm not really against it, just curious.

aeisenberg: There was a linting warning to avoid nested functions. It's not really a big deal and I don't feel strongly about it. It was easier for me to change the code than to change the rule.

extensions/ql-vscode/src/databases.ts
aeisenberg: `timeout !== undefined` is never needed since `clearTimeout` is a noop for non-timers. And it looks like the other if statement is not necessary either.

extensions/ql-vscode/src/quick-query.ts
aeisenberg: Actually, now that I look at the rule more closely, it specifically targets named functions that are not in the top-level of another function or global. I think this is generally good since it makes things harder to read if there are lots of nested functions scattered inside of if statements and elsewhere. Made a little fix here.

jcreedcmu:
Curious why it didn't complain about the nested `function dispose()` then. Not terrifically important, although I'd be in favor of allowing nested functions generally.

aeisenberg:
Functions that are top level inside another one are ok. https://eslint.org/docs/rules/no-inner-declarations

```
function a() {
  function b()  {}
}
```
is ok, but

```
function a() {
  try {
    function b()  {}
  } catch (e) {}
}
```
is not.

PR #316, opened 2020-03-25
fix: Avoid warning user if no new launcher exists

aeisenberg:
Allow extension to remain on the deprecated launcher.

There's no point in issuing a warning for using a deprecated launcher if the correct launcher doesn't exist.

Also, adds a bunch of unit tests around here.

### DIFF
--- a/extensions/ql-vscode/src/distribution.ts
+++ b/extensions/ql-vscode/src/distribution.ts
@@ -112,7 +112,13 @@ export class DistributionManager implements DistributionProvider {
           "that a CodeQL executable exists at the specified path or remove the setting.");
         return undefined;
       }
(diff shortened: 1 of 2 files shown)


aeisenberg:
Based on discussion in #303

extensions/ql-vscode/src/distribution.ts
```diff
@@ -173,6 +179,21 @@ export class DistributionManager implements DistributionProvider {
     return this._onDidChangeDistribution;
   }
 
+  /**
+   * @return true iff the non-deprecated launcher name exists on the file system
+   * in the same directory as the specified launcher if an external installation
```
hmakholm:
```suggestion
   * in the same directory as the specified launcher and an external installation
```

aeisenberg: Not quite right. I'll reword.

PR #324, opened 2020-04-02
Add 'Always Save' to running unsaved query dialog

### LINKED ISSUE #323: Running an unsaved query should have 'always save' choice
Presently when you run a query, it gives you the option to save the file before running. That dialog should have an option to permanently save a preference for autosaving queries before running.

jcreedcmu:
Fixes #323.

cc: @github/product-docs-dsp in case this needs docs.

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -132,6 +132,11 @@
           "default": false,
           "description": "Enable debug logging and tuple counting when running CodeQL queries. This information is useful for debugging query performance."
         },
+        "codeQL.runningQueries.autoSave": {
+          "type": "boolean",
+          "default": false,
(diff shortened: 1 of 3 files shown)


shati-patel:
> cc: @github/product-docs-dsp in case this needs docs.

Thank you for the mention! Just a changelog entry should be fine in this case.

jcreedcmu:
Not sure why unit tests are timing out on windows only.

jcreedcmu:
Tests weren't running on windows because `codeql` the shell script (or possibly `codeql.cmd`) was overwriting `codeql.exe` the binary during unzipping --- implying, I think, correct me if I'm wrong, that it's specficially `unzip` which is making this call as to which files are morally the same as one another, not the underlying filesystem.

Issue #325, opened 2020-04-05
Provide basic XML syntax highlighting for qhelp files


p-:
**Problem**
As a developer it would be nice to have basic syntax highlighting support for qhelp files.
Currently qhelp files are displayed as plain text and look rather dull ;)
Also there's no help for finding out which tags belong together.

**Proposed solution**
Since qhelp files are basically XML files interpreting them as XML files would be a low hanging fruit.
This would not automatically XML validation support as VS code does not support XML validation out of the box.

**Describe alternatives you've considered**
An alternative would be to use a different VSCode grammar set that restricts the set of tags that can be used. However, I think that would be a better candidate for XML validation.

**Additional context**
I will create a small pull request for this issue.
The Pull request instructions instructed me that issues need to be created for any UI or other user-facing changes, so I created this one.

aeisenberg:
Closed by #326.

PR #326, opened 2020-04-05
feat: XML syntax highlighting for qhelp files

p-:
Provides basic XML syntax highlighting for qhelp files as described in issue #325.

I noticed the need for qhelp file features after annoying @aschackmull with bad XML in my qhelp file 😉

This change tells vscode to interpret qhelp files as XML as seen in this screenshot:

qhelp-syntax-highlighting

## Open questions
* It is not clear to me if name is qhelp or Qhelp? As both variants are used in the [Query help style guide](https://github.com/Semmle/ql/blob/master/docs/query-help-style-guide.md)
* I was not sure whether I (as an external contributor) should add this small feature to the CHANGELOG and cc product-docs-dsp?

## Checklist

- [ ] [CHANGELOG.md](../extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [x] Issues have been created for any UI or other user-facing changes made by this pull request.  (#325)

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -77,6 +77,15 @@
           ".dbscheme"
         ],
         "configuration": "./language-configuration.json"
+      },
+      {
+        "id": "xml",
+        "aliases": [
+          "qhelp"
+        ],
+        "extensions": [
+          ".qhelp"
+        ]
       }
     ],
     "grammars": [


jcreedcmu:
Thanks for the contribution! Looks good to me. The windows test regression should be fixed by https://github.com/github/vscode-codeql/commit/ba3381fbf9a977c530aaeede8ac65824e48903e5 which is already in master.

jcreedcmu:
@product-docs-dsp any thoughts on whether it should be referred to as "Qhelp" vs. "qhelp" vs. something else?

mchammer01:
Good question @jcreedcmu 👍 
I think we need to discuss this as a team. Do you know if we are keeping these files as they are on GitHub? I have been helping with documenting a new feature on GitHub, and on one of the pages, we see the content of a .`qhelp` file. I'd like to know too, how we should refer to such files, externally (.qhelp files?) and for the end-user documentation (query help).

jf205:
Hi @jcreedcmu and @p-!

> any thoughts on whether it should be referred to as "Qhelp" vs. "qhelp" vs. something else?

The convention the docs team go with is that we always refer to them as 'query help files', with the extension `.qhelp`. I can see that the 'Query help style guide' doesn't follow that convention though, so I'll open up a PR to sort that out 👍

p-:
Hi @jf205 
Thanks for clarifying!

@jcreedcmu We/I should probably change the name that is displayed on the bottom right to "Query Help" (similar to "Plain Text"). What do you think?

Issue #263, opened 2020-03-05
Only retry integration tests when tests are terminated


jcreedcmu:
In https://github.com/github/vscode-codeql/pull/262, we retry integration tests whenever either they fail, or they're terminated by a signal coming from a flaky segfault in the underlying test runner.

If in the future there arises a way to distinguish these cases (e.g. https://github.com/microsoft/vscode-test/pull/56) only retry in the terminated-by-signal case.

PR #328, opened 2020-04-13
Retry integration tests only on segfault

jcreedcmu:
Whenever https://github.com/microsoft/vscode-test releases 1.4.0, which includes 
https://github.com/microsoft/vscode-test/pull/56, we can merge this PR.

Fixes https://github.com/github/vscode-codeql/issues/263.

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -84,7 +84,7 @@ dependencies:
   vsce: 1.74.0
   vscode-jsonrpc: 5.0.1
   vscode-languageclient: 6.1.3
-  vscode-test: 1.3.0
+  vscode-test: 1.4.0
   vscode-test-adapter-api: 1.7.0
   vscode-test-adapter-util: 0.7.0
   webpack: 4.42.0_webpack@4.42.0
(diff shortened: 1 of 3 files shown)


aeisenberg:
Nice. It looks safe to merge now. On current VSCode, the retry will just never be triggered, so the behaviour would be the same.

aeisenberg:
This looks good. Feel free to merge when 1.4.0 is out.

jcreedcmu:
Right, I do want to wait until the change to `vscode-test` lands, because otherwise the behavior on our end is that we *won't* retry when we get a SIGSEGV.

jcreedcmu:
(Also, this PR needs to include a rush-managed change to `package.json` that actually pins to version 1.4.0 of `vscode-test`, and I am waiting until that lands, rather than manually changing the lockfile, in case their dependencies change)

jcreedcmu:
Looks like `vscode-test` might actually be releasing 1.4.0 soon, and https://github.com/microsoft/vscode-test/issues/63 shows someone else is also interested in https://github.com/microsoft/vscode-test/pull/56 landing.

jcreedcmu:
I tested by pushing ddd9f734 above and confirming that tests correctly failed without retries.

PR #337, opened 2020-04-21
Add experimental support for Jump-to-def and Find-references

jcreedcmu:
This adds the ability to use vscode command palette commands `Go To Definition` and `Go To References` in source archive files for which there are appropriately tagged queries in the appropriate qlpack. The qlpack is inferred from the metadata of the database the source archive came from, in more or less the same way that quick query does.

This ability is gated for the time being by a hidden feature flag in the user's vscode settings. Internal users can add 
```
 "codeQL.experimentalFeatures": true,
```
to their settings json file.

An example of "appropriately tagged queries" for C++ is added by https://github.com/Semmle/ql/pull/3308.

## Checklist

- [x] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/CHANGELOG.md) has not been updated yet, because this is not yet generally user visible.
- [x] `@github/product-docs-dsp` now is a great time to test-drive this, but it can be shipped as-is without documentation, because users won't be able to accidentally stumble on it undocumented.


extensions/ql-vscode/src/cli.ts
```diff
@@ -614,6 +614,26 @@ export class CodeQLCliServer implements Disposable {
...
   }
+
+  /**
+   * Gets information about queries in a query suite.
+   * @param additionalPacks A list of directories to search for qlpacks before searching in `searchPath`.
+   * @param searchPath A list of directories to search for packs not found in `additionalPacks`. If undefined,
+   *   the default CLI search path is used.
+   * @returns A list of query files found
+   */
+  resolveQueriesInSuite(suite: string, additionalPacks: string[], searchPath?: string[]): Promise<string[]> {
```
aeisenberg, line 625: Missing the `suite` param in the tsdoc

extensions/ql-vscode/src/cli.ts
```diff
@@ -614,6 +614,26 @@ export class CodeQLCliServer implements Disposable {
...
+    const args = ['--additional-packs', additionalPacks.join(path.delimiter)];
+    if (searchPath !== undefined) {
+      args.push('--search-path', path.join(...searchPath));
+    }
+    args.push(suite);
+    return this.runJsonCodeQlCliCommand<string[]>(
```
aeisenberg, line 631: Should you mark this function as `async`?

extensions/ql-vscode/src/config.ts
```diff
@@ -39,6 +39,17 @@ class Setting {
...
+// Enable experimental features
+
+/**
+ * This setting is deliberately not in package.json so that it does
+ * not appear in the settings ui in vscode itself. If users want to
+ * enable experimental features, they can add
+ * "codeQl.experimentalFeatures" directly in their vscode settings
+ * json file.
+ */
+export const EXPERIMENTAL_FEATURES_SETTING = new Setting('experimentalFeatures', ROOT_SETTING);
```
aeisenberg, line 51: nice. I like Easter eggs. :)

extensions/ql-vscode/src/definitions.ts
```diff
@@ -0,0 +1,202 @@
...
+ * source-language files. We may eventually want to find a way to
+ * generalize this to other custom queries, e.g. showing dataflow to
+ * or from a selected identifier.
+ */
+
+const TEMPLATE_NAME = "selectedSourceFile";
+const SELECT_QUERY_NAME = "#select";
+
+enum KeyType {
+  DefinitionQuery, ReferenceQuery
```
aeisenberg, line 26: For easier debugging, I like to do:

```
DefinitionQuery = 'DefinitionQuery',
ReferenceQuery = 'ReferenceQuery'
```

extensions/ql-vscode/src/cli.ts
jcreedcmu: It's a fair question, and one which applies to a bunch of other `Promise`-returning functions in this file, but I think that I like the existing convention that we mark a function as `async` if it needs to be, i.e., actually uses `await`, and otherwise let its return type being `Promise` being a sufficient signal that what it returns isn't immediately a value but a future.

extensions/ql-vscode/src/definitions.ts
jcreedcmu: Sure, can do. Iirc with ts enums, you do have maps in both directions (e.g. `KeyType.ReferenceQuery = 1` and `KeyType[1] = "ReferenceQuery"`) but I can understand not wanting to have the extra step.

Issue #339, opened 2020-04-22
Minor bug: XML files are labeled as "qhelp"


shati-patel:
**Describe the bug**

![image](https://user-images.githubusercontent.com/42641846/80005359-5ea76180-84bb-11ea-9b2a-e745efae6af2.png)

**To reproduce**
Open an XML file in a workspace that has the CodeQL extension enabled. 

**Expected behavior**
- Query help (`.qhelp`) files are highlighted as XML
- XML files are still labeled as "XML"

**Additional context**
`@p-` recently introduced XML highlighting for Query help files (see https://github.com/github/vscode-codeql/pull/326) ❤️  
That update probably got VS Code confused about "normal" XML files. 

(Tangential note: I have now added a [file association](https://code.visualstudio.com/docs/languages/overview#_adding-a-file-extension-to-a-language) in my user settings, which tells VS Code to treat `.qhelp` files as XML, without overriding the default XML behavior: 
```
"files.associations": {
    "*.qhelp": "xml"
}
```
I don't know if it's possible to impose that from within the extension though, so that might not be a helpful approach!)

PR #340, opened 2020-04-23
Remove 'qhelp' as global alias for 'xml' filetype

jcreedcmu:
Fixes https://github.com/github/vscode-codeql/issues/339.

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -80,9 +80,6 @@
       },
       {
         "id": "xml",
-        "aliases": [
-          "qhelp"
-        ],
         "extensions": [
           ".qhelp"
         ]


jcreedcmu:
I think this is the right thing to do; I thought the aliases were scoped to the file extension being added, but apparently not.

aeisenberg:
Odd. but when I try your PR, my `qhelp` files are displayed as `html`. And I don't even see `qhelp` as an option.

jcreedcmu:
Huh, that's strange and I don't know how to explain that yet. I wouldn't expect to see qhelp as an option, precisely because I removed it as an alias --- the intent of this PR is to restore the situation to there only being a notion of a file "being" xml, and removing the possibility of a file being of type qhelp, but preserving that the `.qhelp` extension should cause them to be viewed as xml. Can you repro the behavior you're seeing with this PR but with all other extensions disabled?

aeisenberg:
I can't explain what happened before. I tried this again and this seems to be working as expected. This just associates qhelp as an xml file, which I think is perfectly fine. In the future, we can consider creating a proper qhelp language type.

PR #344, opened 2020-04-28
Changelog: Add release date

shati-patel:
The latest changelog entry (for 1.1.2) didn't include the date. Is it possible to include this retroactively?


jcreedcmu:
Good catch!

PR #350, opened 2020-05-05
Experimental: Enable parsing bqrs with the cli instead of in the webview

jcreedcmu:


dbartol:
I agree that results parsing shouldn't happen in the webview. Before committing to all BQRS parsing being done in the CLI, though, I'd like to understand the performance impact. For interactive viewing of results in the results view, I expect that it will be fine. However, I'd also like to enable lots of new features that depend on running custom queries with custom output schemas and parsing the results. Think "live AST view", or code exploration features like "find all callers/callees". I don't want to have to add new interpretation code to the CLI itself for each new query, but I don't know if expanding everything to JSON and sending it across process boundaries is going to be quick enough. I'd be less worried if I believed that no query like that would ever have to parse a large amount of output at once, but I just don't know enough about the kinds of queries we'll eventually want to write to be sure.

jcreedcmu:
Your concern sounds valid, and I'm definitely not planning to merge anything that deletes the existing path without careful thought.

I think that specifically relying on doing any `fetch(...)` calls in the webview is something we need to be moving away from, though; afaik, it's a nonstarter for containerized vscode, and it complicates reasoning about which part of the extension has access to the filesystem or not.

The question of whether bqrs parsing is done in one of the cli/server pieces or in the main (i.e. nonwebview) part of the extension is a completely reasonable question to keep open for debate. I still feel like moving as much as we can to the cli as possible is quite desirable if only for ensuring that the bqrs format itself can evolve without having multiple implementations that need to be changed.

extensions/ql-vscode/src/adapt.ts
```diff
@@ -0,0 +1,103 @@
...
+export type ResultValue = ResultElement | ResultUri | string;
+
+export interface RawResultSet {
+  readonly schema: AdaptedSchema;
+  readonly rows: readonly ResultRow[];
+}
+
+function adaptKind(kind: ColumnKind): ColumnType {
+  // XXX what about 'u'?
+  if (kind == 'e') {
```
aeisenberg, line 36: nit: `===`

extensions/ql-vscode/src/adapt.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export function adaptValue(val: ColumnValue): ResultValue {
+  // XXX taking a lot of incorrect shortcuts here
+
+  if (typeof val == 'string') {
+    return val;
+  }
+
+  if (typeof val == 'number' || typeof val == 'boolean') {
```
aeisenberg, line 64: nit: `===` here and above/below.

extensions/ql-vscode/src/interface.ts
```diff
@@ -363,13 +363,28 @@ export class InterfaceManager extends DisposableObject {
...
     }
 
+    let resultSets: RawResultSet[] | undefined;
+
+    if (EXPERIMENTAL_BQRS_SETTING.getValue()) {
+      resultSets = [];
+      const pageSize = 100;
+      const schemas = await this.cliServer.bqrsInfo(results.query.resultsPaths.resultsPath, pageSize);
+      for (const schema of schemas["result-sets"]) {
+        const chunk = await this.cliServer.bqrsDecode(results.query.resultsPaths.resultsPath, schema.name, pageSize, 0)
```
aeisenberg, line 373: Does this get only the first 100 results for each resultSet? Maybe ok for a PoC. Though, I'd be leaning towards getting all of the results for the PoC just so we can see how much juice these codespace vms really have.

jcreedcmu: Ah, sure, good point.

PR #352, opened 2020-05-07
Update versions of json-rpc dependencies.

jcreedcmu:


extensions/ql-vscode/package.json
```diff
@@ -464,8 +464,8 @@
     "tmp": "^0.1.0",
     "tree-kill": "~1.2.2",
     "unzipper": "~0.10.5",
-    "vscode-jsonrpc": "^4.0.0",
-    "vscode-languageclient": "^5.2.1",
+    "vscode-jsonrpc": "~5.0",
```
dbartol, line 467: OOI, why `~` and not `^`? I think `^` is the npm default now, and allows for minor version upgrades, whereas `~` only allows patch upgrades.

jcreedcmu: Ah, you're right. Used ~ from muscle memory, but seems better to be consistent with other package specs anyhow.

PR #353, opened 2020-05-07
Add source locations for string results

### LINKED ISSUE #348: Support for source locations provided by `getURL`
The help topic on [Locations and strings for QL entities](https://help.semmle.com/QL/learn-ql/locations.html#providing-urls) describes how a `getURL` predicate can be used to provide a source location for a QL entity.

As a contrived example, the following query selects all addition expressions in a JavaScript database, but with their source locations shifted to the right by one character:

```ql
import javascript

class ShiftedAddExpr extends AddExpr {
  string getURL() {
    exists(string filepath, int startline, int startcolumn, int endline, int endcolumn |
      getLocation().hasLocationInfo(filepath, startline, startcolumn, endline, endcolumn) and
      result =
        "file://" + filepath + ":" + startline + ":" + (startcolumn + 1) + ":" + endline + ":" +
          (endcolumn + 1)
    )
  }
}

select any(ShiftedAddExpr a)
```

This works in the legacy QL-for-Eclipse plugin, but in the VSCode extension the results end up not being clickable.

In the interest of feature completeness, it would be great if source locations provided by `getURL` work.

aeisenberg:
Fixes #348.

This PR also includes 2 small, non-related changes for better editing experience.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -32,5 +32,7 @@
   "eslint.options": {
     // This is necessary so that eslint can properly resolve its plugins
     "resolvePluginsRelativeTo": "./extensions/ql-vscode"
-  }
+  },
+  // Force this to false since this will cause too many changes on each commit
+  "editor.formatOnSave": false
 }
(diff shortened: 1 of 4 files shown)


jcreedcmu:
I'm to understand that using the cli bqrs parser will also naturally fix this issue, but this seems good as a temporary fix.

aeisenberg:
Eek. I think I need to change something. I think we're going to be off-by-one for these.

aeisenberg:
Received clarification on https://github.com/github/vscode-codeql/issues/348#issuecomment-625957878 that the existing implementation is correct.

PR #358, opened 2020-05-11
Add command to download, unzip, and open databases

### LINKED ISSUE #357: Add ability to download a database from a url and install into the extension
One way to make it easier for users on codespaces, is to ensure a simple way of installing a database into a running instance of the extension.

This task is about installing a database from a zip file located at a URL.

aeisenberg:
New command that requests a URL and allows a user to install a
database from that url.

Closes #357

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databaseFetcher.ts
+++ b/extensions/ql-vscode/src/databaseFetcher.ts
@@ -1,0 +1,112 @@
+import * as fetch from "node-fetch";
+import * as unzipper from "unzipper";
+import { ExtensionContext, Uri, ProgressOptions, ProgressLocation, commands, window } from "vscode";
+import * as fs from "fs-extra";
+import * as path from "path";
+import { DatabaseManager } from "./databases";
+import { ProgressCallback, showAndLogErrorMessage, withProgress } from "./helpers";
+
+export default async function promptFetchDatabase(dbm: DatabaseManager, ctx: ExtensionContext) {
+  try {
+    const databaseUrl = await window.showInputBox({
+      prompt: 'Enter URL of zipfile of database to download'
+    });
+
+    if (databaseUrl) {
+      validateUrl(databaseUrl);
+
+      const progressOptions: ProgressOptions = {
+        location: ProgressLocation.Notification,
+        title: 'Adding database from URL',
+        cancellable: false,
+      };
+      await withProgress(progressOptions, async progress => await databaseFetcher(databaseUrl, dbm, ctx, progress));
+      commands.executeCommand('codeQLDatabases.focus');
+    }
+  } catch (e) {
+    showAndLogErrorMessage(e.message);
+  }
+}
+
+async function databaseFetcher(
+  databaseUrl: string,
+  databasesManager: DatabaseManager,
+  ctx: ExtensionContext,
+  progressCallback: ProgressCallback
+): Promise<void> {
+  progressCallback({
+    maxStep: 3,
+    message: 'Downloading database',
+    step: 1
+  });
+  const storagePath = ctx.storagePath || ctx.globalStoragePath;
+  if (!storagePath) {
+    throw new Error("No storage path specified.");
+  }
+  const unzipPath = await getStorageFolder(storagePath, databaseUrl);
+
+  const response = await fetch.default(databaseUrl);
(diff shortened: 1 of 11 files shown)


extensions/ql-vscode/src/databases.ts
```diff
@@ -680,6 +688,11 @@ export class DatabaseManager extends DisposableObject {
   private updatePersistedDatabaseList(): void {
     this.ctx.workspaceState.update(DB_LIST, this._databaseItems.map(item => item.getPersistedState()));
   }
+
+  private isExtensionControlledLocation(uri: vscode.Uri) {
+    const storagePath = this.ctx.storagePath || this.ctx.globalStoragePath;
+    return uri.path.startsWith(storagePath);
```
adityasharad, line 694: This might need symlink resolution, e.g. call `realpath` on the two paths before checking the prefix.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -0,0 +1,97 @@
...
+    message: 'Downloading database',
+    step: 1
+  });
+  const storagePath = ctx.storagePath || ctx.globalStoragePath;
+  if (!storagePath) {
+    throw new Error("No storage path specified.");
+  }
+  const unzipPath = await getStorageFolder(storagePath, databaseUrl);
+
+  const response = await fetch.default(databaseUrl);
```
adityasharad, line 48: Although the URL is user-provided and we are on the user's filesystem, this concerns me a little. Could we enforce that the URL is HTTPS, and possibly compare the downloaded artifact against a checksum, like we do when updating the CLI?

aeisenberg: HTTPS is something we can enforce. Though, not sure how we can get a checksum. This is just a zip file and there's no guarantee that a checksum exists, nor is there a way to discover it if one does.

extensions/ql-vscode/src/databases.ts
aeisenberg: sure.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -0,0 +1,114 @@
...
+  databasesManager: DatabaseManager,
+  ctx: ExtensionContext,
+  progressCallback: ProgressCallback
+): Promise<void> {
+  progressCallback({
+    maxStep: 3,
+    message: 'Downloading database',
+    step: 1
+  });
+  const storagePath = ctx.storagePath || ctx.globalStoragePath;
```
jcreedcmu, line 44: maybe factor out this `this.ctx.storagePath || this.ctx.globalStoragePath`, and the one that occurs in `databaseFetcher.ts`, into a function in `databaseFetcher.ts` that takes an extension context and returns that, just to reduce the probability that those ever get out of sync in the future.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -0,0 +1,97 @@
...
+import * as unzipper from "unzipper";
+import { ExtensionContext, Uri, ProgressOptions, ProgressLocation, commands, window } from "vscode";
+import * as fs from "fs-extra";
+import * as path from "path";
+import { DatabaseManager } from "./databases";
+import { ProgressCallback, showAndLogErrorMessage, withProgress } from "./helpers";
+
+
+
+export default async function runCommand(dbm: DatabaseManager, ctx: ExtensionContext) {
```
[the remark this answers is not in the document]
aeisenberg: Yes.

aeisenberg:

This is a new feature for the extension. It's primarily helpful for getting databases into a codespace, though it's also useful for local development.

The new `CodeQL: Download database` command will bring up an input box where a user enters a URL (https is supported, only) that points to a zip file containing a code ql database. The database will be downloaded and unzipped to a location managed by the extension. Then it will be imported into the workspace and set as the current database.

When the database is removed from the workspace, it will also be removed from disk (this is not true of databases that were already on disk and are managed by the user).

shati-patel:
Thanks for the ping! We'll look at the docs impact in [this issue](https://github.com/github/semmle-docs/issues/107) 📝

PR #360, opened 2020-05-12
Add format and lint on commit

aeisenberg:
This PR adds husky to the project. This will force linting and formatting to run on each commit.

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -465,7 +465,8 @@
     "update-vscode": "node ./node_modules/vscode/bin/install",
     "postinstall": "node ./node_modules/vscode/bin/install",
     "format": "tsfmt -r",
-    "lint": "eslint src test --ext .ts,.tsx"
+    "lint": "eslint src test --ext .ts,.tsx",
+    "lint-staged": "lint-staged"
   },
   "dependencies": {
     "child-process-promise": "^2.2.1",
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/package.json
```diff
@@ -542,6 +543,22 @@
...
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run lint-staged"
+    }
+  },
+  "lint-staged": {
+    "./**/*.{json,css,scss,md}": [
+      "prettier --write"
+    ],
+    "./**/*.{js,jsx,ts,tsx}": [
```
aeisenberg, line 559: Oops. Need to only do ts files here.

aeisenberg:
@jcreedcmu when you try this out, can you verify that the precommit hooks are installed for you automatically when you install husky. I had to run `npm rebuild` as described in [their docs](https://www.npmjs.com/package/husky#reinstall). I don't know if this is because we are in a monorepo situation.

extensions/ql-vscode/package.json
```diff
@@ -542,6 +543,23 @@
...
     "@types/proxyquire": "~1.3.28",
-    "eslint-plugin-react": "~7.19.0"
+    "eslint-plugin-react": "~7.19.0",
+    "husky": "~4.2.5",
+    "lint-staged": "~10.2.2",
+    "prettier": "~2.0.5"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run lint-staged"
```
jcreedcmu, line 553: this seems fine for now, but I am leaving a note here that husky appears to also have `pre-push` in case this turns out to feel to slow.

aeisenberg: Yep..I'm open to that if that's the direction you want to go.

jcreedcmu:
> @jcreedcmu when you try this out, can you verify that the precommit hooks are installed for you automatically when you install husky. I had to run `npm rebuild` as described in [their docs](https://www.npmjs.com/package/husky#reinstall). I don't know if this is because we are in a monorepo situation.

Looks like it worked; I ran `rush update && rush install && rush build` and then
```
$ cat .git/hooks/pre-commit
#!/bin/sh
# husky

# Created by Husky v4.2.5 (https://github.com/typicode/husky#readme)
#   At: 5/12/2020, 1:53:35 PM
#   From: /home/jcreed/git/ql-vscode/common/temp/node_modules/.pnpm/registry.npmjs.org/husky/4.2.5/node_modules/husky (https://github.com/typicode/husky#readme)

. "$(dirname "$0")/husky.sh"
```

jcreedcmu:
Hm, but running that script doesn't actually format yet.

jcreedcmu:
Ok, the `npm rebuild` was necessary for me too. And it is a bit of a bummer that it adds 3-4s of latency per commit. But I can live with that.

PR #366, opened 2020-05-13
Augments the add database command to handle zip files

aeisenberg:
The add database command can now add databases by zip file. When a
file is selected, the zip file is attempted to be extracted into a
directory managed by the extension.

Once extracted, a database is searched for, by looking for a .dbinfo
file.

Crucially, we are using the same infrastructure to download a database
as we are to add a database by zip file.

## Checklist


extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -45,36 +45,27 @@ async function databaseFetcher(
...
-  // if there is a single directory inside, then assume that's what we want to import
-  const dirs = await fs.readdir(unzipPath);
-  const dbPath = dirs?.length === 1 && (await fs.stat(path.join(unzipPath, dirs[0]))).isDirectory
-    ? path.join(unzipPath, dirs[0])
-    : unzipPath;
-
-  // might need to upgrade before importing...
-  const item = await databasesManager.openDatabase(Uri.parse(dbPath));
-  databasesManager.setCurrentDatabaseItem(item);
+  const dbPath = await findDirWithFile(unzipPath, '.dbinfo');
```
adityasharad, line 60: Note that this is `codeql-database.yml` for newer databases.

aeisenberg: Yikes. Is there a stable file to look for?

jcreedcmu: Seems like we want to check for the presence of either `codeql-database.yml` or `.dbinfo`, yeah?

jcreedcmu: Checking for the presence of either of those two seems appropriate to me.

jcreedcmu:
Unfortunately, this implementation doesn't work for me on Linux, and I suspect it won't work on Windows, based on docs in https://code.visualstudio.com/api/references/vscode-api#OpenDialogOptions --- the dialog lets me select only folders, not files, if `canSelectFolders` is set, even if `canSelectFiles` is true. We might have to rethink how this works.

On alternative might be to enable the 'set current database' context menu command in the vscode file explorer for .zip files. I think this would work well in the case of codespaces with snapshots you downloaded locally, since then the flow could be just drag-and-drop, right-click and set as current database.

jcreedcmu:
I think doing something like
```
        {
          "command": "codeQL.setCurrentDatabase",
          "group": "9_qlCommands",
          "when": "resourceScheme == codeql-zip-archive || explorerResourceIsFolder || resourceExtname == .zip"
        },
```
will work.

aeisenberg:
That's interesting. I will try your suggestion out.

aeisenberg:

Some new features in this PR

* We've changed the icons on the databases view
* We've added two new commands on the databases view: 
  - Choose Database from Internet (the cloud icon), which is the same action I implemented in #358.
  - Choose Database from Archive (the file icon), which will take a zip file, and unzip it into a extension-controlled location and import the database into the workspace.
  - Choose Database from Folder (the folder icon), which is existing functionality to add a database from a folder, except that the icon has changed.

aeisenberg:
@jcreedcmu, I'll do your suggestion in a separate PR.

aeisenberg:
Fixes #369.

PR #367, opened 2020-05-14
Make links variable-width without underline

jbj:
These are two independent changes.
- The monospace font used in links made the text larger so that less text could fit on the screen. It also suggested that all link text was an code snippet, which it isn't. The advantage of a fixed-width font, vertical alignment, was not put to any use.
- Underlining of links made it almost impossible to distinguish a space from an underscore.

Before:
vscode-underline

After:
vscode-no-underline

Notice how it's practically impossible to tell underscores from spaces in the underlined link for `unpack_index_entry output argument [cache_nr]` (row 11).

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CodeQL for Visual Studio Code: Changelog
 
+## 1.1.5
+
+- Links in results are no longer underlined and monospaced.
+
 ## 1.1.4 - 13 May 2020
 
 - Add the ability to download and install databases archives from the internet.
--- a/extensions/ql-vscode/src/view/resultsView.css
+++ b/extensions/ql-vscode/src/view/resultsView.css
@@ -75,7 +75,7 @@
 }
 
 .vscode-codeql__result-table-location-link {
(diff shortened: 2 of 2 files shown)


jcreedcmu:
I mostly like this idea, but I feel like monospaced font is good simply for visually signalling that a piece of text _is_ code --- how would you feel about making the result monospaced, but decreasing the font size a bit?

@aeisenberg suggested it would be good if we could match the exact font the user currently has configured for text in buffers in vscode, though it's not immediately obvious to me whether we can get at that across the webview boundary easily.

jbj:
> I mostly like this idea, but I feel like monospaced font is good simply for visually signalling that a piece of text _is_ code

The problem is when the text _isn't_ code. It's only sometimes code. In my screenshots, the non-code phrases are: "user input", "call to", "output argument", and "BufferReadSideEffect". That last one needs to be changed into something human-readable.

aeisenberg:
I like removing the underline. Although I agree that using monospace font for code is more intuitive, I also see how it's not all code, so I don't really feel strongly about using monospace or not here. I'm fine letting this go through as-is.

jcreedcmu:
Ah, ok, good point. I don't feel too strongly then either.

PR #374, opened 2020-05-15
Fix paths on windows when opening archive databases

### LINKED ISSUE #373: Cannot open a local archive database on windows
Using the new `Choose database from archive` command on a windows machine, the database fails to open due to a path error.

Discovered by @shati-patel

aeisenberg:
Fixes #373.

### DIFF
--- a/extensions/ql-vscode/src/databaseFetcher.ts
+++ b/extensions/ql-vscode/src/databaseFetcher.ts
@@ -12,7 +12,7 @@ import { ProgressCallback, showAndLogErrorMessage, withProgress, showAndLogInfor
  * @param databasesManager the DatabaseManager
  * @param storagePath where to store the unzipped database.
  */
-export async function promptImportInternetDatabase(databasesManager: DatabaseManager, storagePath: string): Promise<DatabaseItem | undefined>  {
+export async function promptImportInternetDatabase(databasesManager: DatabaseManager, storagePath: string): Promise<DatabaseItem | undefined> {
   let item: DatabaseItem | undefined = undefined;
 
   try {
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -115,14 +115,17 @@ async function databaseArchiveFetcher(
 }
 
 async function getStorageFolder(storagePath: string, urlStr: string) {
+  // we need some approximation for a folder name
   const url = Uri.parse(urlStr);
-  let lastName = path.basename(url.path).substring(0, 255);
+  let lastName = path.basename(url.path).substring(0, 250);
```
jcreedcmu, line 120: What's going on with this length limit? Add a comment maybe?

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -115,14 +115,17 @@ async function databaseArchiveFetcher(
 }
 
 async function getStorageFolder(storagePath: string, urlStr: string) {
+  // we need some approximation for a folder name
```
jcreedcmu, line 118: Not clear to me what's being approximated?

aeisenberg: I'll make that clearer. The max file name length is 260 on windows and 255 on macs. But the zip files tend to have very long names. Also, we add a few extra chars at the end in case there are duplicates.

I'm partially thinking that we should just avoid this and use a random name.

aeisenberg: Wait...we can't just use a random name since we also use the folder name as the database name when we import the database.

PR #377, opened 2020-05-19
Fix naming and availability in command palette of various commands

jcreedcmu:


shati-patel:
(Sorry, a few drive-by questions!)
- Is there a reason for there to be two "CodeQL: Download database" commands? (`codeQL.downloadDatabase` and `codeQL.chooseDatabaseInternet`) They both show up in the Command Palette 🤔 
- Would it be okay to change that command to "CodeQL: Download Database", to keep the capitalization consistent with other commands?

jcreedcmu:
Oh, good catch! I had misunderstood the problem. These unprefixed commands were not supposed to be in the command palette at all, but just hooked up to those database view buttons. The command `codeQL.downloadDatabase` is the palette command, and (what is now named, for consistency) `codeQLDatabases.chooseDatabaseInternet` is the command behind the corresponding button, and will no longer show up in the command palette as of the above commit.

extensions/ql-vscode/package.json
```diff
@@ -27,9 +27,9 @@
     "onView:codeQLQueryHistory",
     "onView:test-explorer",
     "onCommand:codeQL.checkForUpdatesToCLI",
-    "onCommand:codeQL.chooseDatabaseFolder",
-    "onCommand:codeQL.chooseDatabaseArchive",
-    "onCommand:codeQL.chooseDatabaseInternet",
+    "onCommand:codeQLDatabases.chooseDatabaseFolder",
```
aeisenberg, line 30: I had thought the difference between the `codeQL` namespace and the `codeQLDatabases` namespace was that the latter accepted an existing database as an argument (and so was only applicable in the databases view). Whereas the former could be invoked in any context.

I see you're disabling them as global commands, so maybe that's what we want.

extensions/ql-vscode/package.json
```diff
@@ -442,6 +442,18 @@
           "command": "codeQLDatabases.removeDatabase",
           "when": "false"
         },
+        {
+          "command": "codeQLDatabases.chooseDatabaseFolder",
```
aeisenberg, line 446: Do we want to disable these as global commands?

aeisenberg:
Yes, this is inconsistent. Maybe the best solution is to simply remove `codeQL.downloadDatabase`. Is there any reason _not_ to make the other commands globally available?

extensions/ql-vscode/package.json
jcreedcmu: Happy to be persuaded about choice of convention, but my intent so far about the `codeQL$FOO.$BAR` name structure (where `$FOO` is nonempty) was just about scoping to portions of the interface. Consider for example `codeQLDatabases.sortByName` which does not take a database as an argument, and things like `codeQLQueryHistory.openQuery` which are relevant to user events taking place in the query history view, but don't imply any particular constraints about the types of those handlers.

extensions/ql-vscode/package.json
jcreedcmu: We definitely *do* want to disable these commands if we want to consistently maintain the pattern that the hover text over the buttons is not be prefixed, and the command palette command names is prefixed with `CodeQL: `. And I think both of those patterns are desirable.

However, we totally could *also* have command-palette-visible commands (which would be named, like, `codeQL.chooseDatabaseFolder`) which do the same thing.

jcreedcmu:
> Yes, this is inconsistent. Maybe the best solution is to simply remove `codeQL.downloadDatabase`. Is there any reason _not_ to make the other commands globally available?

As commented above, I don't think we need to remove it, but I could make the naming of the internal command more consistent, and add command palette versions of the other two, just in case people are exploring around the command palette and/or prefer to do things with the keyboard instead of clicking on icons with the mouse. Commit incoming shortly...

extensions/ql-vscode/package.json
aeisenberg: That makes sense. Maybe we _don't_ want to pollute the global scope here with commands that are so tightly related to the databases view (I'm contradicting what I said earlier). So, what you have right now is fine.

aeisenberg:
Either way is fine: removing the palette command, or adding the 2 extra palette commands available consistently.

PR #381, opened 2020-05-19
Avoid need for `build` directory

dbartol:
It looks like we added a `build` directory a while back to make it possible to run Rush without installing it globally on the build agent. Rush already has a mechanism for handling this: `common/scripts/install-run-rush.js`. You run that script with the same command line you'd pass to `rush`, and it automatically installs (locally) whatever version of Rush is in your `rush.json` and then runs the command. We were already using that in our `.vscode/tasks.json`. This PR uses it in our Actions workflows as well.

Note that in addition to letting us delete a whole directory, this also prevents our Actions from installing the `rush` package and all of its dependencies twice. Before, we would install whatever version of `rush` was specified in `build/package.json`, and then that would install whatever version of `rush` was specified in `rush.json`. With `install-run-rush.js`, it just directly installs the version specified in `rush.json`.

## Checklist


jcreedcmu:
Nice!

PR #389, opened 2020-05-21
Add error message when there are no definitions/references queries available.

### LINKED ISSUE #387: Better error message when there are no definitions/references queries.
When a user tries to do source archive jump-to-definition when there are no queries available (possibly due merely to an older version of https://github.com/github/codeql) the result they'll see is that there are 'no definitions' or 'no references' found.

What we should do is detect that there are in fact no available queries, possibly explain that that's the case, and suggest that they upgrade.

jcreedcmu:
Turns out we were already raising an exception, but the `DefinitionProvider`/`ReferencesProvider` helpfully silently swallows raised exceptions. (well, they get logged to console, but that's as good as silent from the average user's perspective). I don't see any way to suppress the 'zero definitions/references found' popup, but at least we can clarify *why* zero were found with this message.

I included the specific tag that's being looked for by the extension, since I think it might potentially help advanced users debug what's going on in the event that the issue is *not* merely upgrading, but the queries being missing or improperly located or something because of some future refactoring. 

Fixes #387.

### DIFF
--- a/extensions/ql-vscode/src/definitions.ts
+++ b/extensions/ql-vscode/src/definitions.ts
@@ -34,6 +34,13 @@ function tagOfKeyType(keyType: KeyType): string {
   }
 }
 
+function nameOfKeyType(keyType: KeyType): string {
+  switch (keyType) {
+    case KeyType.DefinitionQuery: return "definitions";
+    case KeyType.ReferenceQuery: return "references";
+  }
+}
+
 async function resolveQueries(cli: CodeQLCliServer, qlpack: string, keyType: KeyType): Promise<string[]> {
   const suiteFile = tmp.fileSync({ postfix: '.qls' }).name;
   const suiteYaml = { qlpack, include: { kind: 'definitions', 'tags contain': tagOfKeyType(keyType) } };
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/definitions.ts
```diff
@@ -34,14 +34,24 @@ function tagOfKeyType(keyType: KeyType): string {
...
 async function resolveQueries(cli: CodeQLCliServer, qlpack: string, keyType: KeyType): Promise<string[]> {
   const suiteFile = tmp.fileSync({ postfix: '.qls' }).name;
   const suiteYaml = { qlpack, include: { kind: 'definitions', 'tags contain': tagOfKeyType(keyType) } };
   await fs.writeFile(suiteFile, yaml.safeDump(suiteYaml), 'utf8');
 
   const queries = await cli.resolveQueriesInSuite(suiteFile, helpers.getOnDiskWorkspaceFolders());
   if (queries.length === 0) {
-    throw new Error("Couldn't find any queries for qlpack");
+    vscode.window.showErrorMessage(
+      `No ${nameOfKeyType(keyType)} queries (tagged "${tagOfKeyType(keyType)}") could be found in the current library path. It might be necessary to upgrade the CodeQL libraries.`
```
aeisenberg, line 52: This is fine, though I wonder if the message could link out to some instructions on how to do that. Not really necessary for this to go through.

PR #390, opened 2020-05-21
Add onEnterRules for language config

### LINKED ISSUE #171: Feature request: QLDoc continuation on new line
Hi,

The QL plugin for eclipse has the nice feature that when I write a QLdoc block comment,
it adds the * and indents correctly whenever I press enter on a newline.
The VSCode plugin doesn't provide this.
For other languages, https://marketplace.visualstudio.com/items?itemName=kevinkyang.auto-comment-blocks is available, which seems to achieve what I want.

aeisenberg:
This change provides proper indent/outdent for block comments. Through
onEnterRules. Because onEnterRules are not exactly API, I had to use
a back door to implement them.

Also, it tweaks the language-configuration.json by adding more support
for things like word boundaries and auto-closing pairs.

Since QL has similar syntactical items as JavaScriot, I started with
the JS lang config and removed single quotes and back ticks.

Fixes #171 

## Checklist

### DIFF
--- a/extensions/ql-vscode/language-configuration.json
+++ b/extensions/ql-vscode/language-configuration.json
@@ -1,72 +1,38 @@
 {
-    "comments": {
-        // symbol used for single line comment. Remove this entry if your language does not support line comments
-        "lineComment": "//",
-        // symbols used for start and end a block comment. Remove this entry if your language does not support block comments
-        "blockComment": [
-            "/*",
-            "*/"
-        ]
-    },
-    // symbols used as brackets
-    "brackets": [
-        [
-            "{",
-            "}"
-        ],
-        [
-            "[",
-            "]"
-        ],
-        [
-            "(",
-            ")"
-        ]
-    ],
-    // symbols that are auto closed when typing
-    "autoClosingPairs": [
-        [
-            "{",
-            "}"
-        ],
-        [
-            "[",
-            "]"
-        ],
-        [
-            "(",
-            ")"
-        ],
-        [
-            "\"",
-            "\""
-        ],
-        [
-            "'",
-            "'"
-        ]
-    ],
-    // symbols that that can be used to surround a selection
-    "surroundingPairs": [
-        [
-            "{",
-            "}"
-        ],
-        [
-            "[",
-            "]"
-        ],
-        [
-            "(",
-            ")"
-        ],
-        [
-            "\"",
-            "\""
-        ],
-        [
-            "'",
-            "'"
-        ]
-    ]
+  "comments": {
+    "lineComment": "//",
+    "blockComment": ["/*", "*/"]
+  },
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/languageSupport.ts
```diff
@@ -0,0 +1,52 @@
...
+export function install() {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const langConfig = require('../language-configuration.json');
+  // setLanguageConfiguration requires a regexp for the wordpattern, not a string
+  langConfig.wordPattern = new RegExp(langConfig.wordPattern);
+  langConfig.onEnterRules = onEnterRules;
+
+  languages.setLanguageConfiguration('ql', langConfig);
+  languages.setLanguageConfiguration('qll', langConfig);
+  languages.setLanguageConfiguration('dbscheme', langConfig);
```
jcreedcmu, line 23: Is this right? The dbscheme files have a syntax that's entirely unlike ql and qll

extensions/ql-vscode/language-configuration.json
```diff
@@ -1,72 +1,38 @@
...
-            "\""
-        ],
-        [
-            "'",
-            "'"
-        ]
-    ]
-}
\ No newline at end of file
+  "comments": {
```
jcreedcmu, line 2: Do we have to get rid of comments? Is this now being consumed as a strict `.json` file as opposed to a lenient `.json`-maybe-with-comments file?

extensions/ql-vscode/src/languageSupport.ts
aeisenberg: In the package.json, dbscheme uses the same language config as ql and qll. So, this is not a regression, though we could consider implementing a different language config in the future.

extensions/ql-vscode/language-configuration.json
aeisenberg: Unfortunately, yes. Because I am calling `require` on this later, we can't have comments in the file. :(

There might be a way to require a json with comments file, but not sure how.

aeisenberg:
Ping @github/product-docs-ds! 

This PR introduces a small UI improvement where typing a QL doc comment will auto close. Also, pressing `enter` inside of a qldoc comment will automatically indent and prefix the line with a `*`.

shati-patel:
These look like handy updates! 🎉 Just the changelog entry is enough here (we don't go into too much detail about autocomplete/query writing [features](https://help.semmle.com/codeql/codeql-for-vscode/procedures/about-codeql-for-vscode.html#features) in the docs).

aeisenberg:
No problem. Just making sure you're aware of the changes. Ideally, these are changes that don't need documentation because they "just work" and users don't need to think about them.

PR #399, opened 2020-05-25
Allow users more flexibility when opening a DB

### LINKED ISSUE #383: Improve ease of selecting a database on the filesystem
https://github.com/github/codeql-coreql-team/issues/70

Slightly confusing that a database is a folder that can be navigated deeper into.

We could allow the user to select the (suitably renamed) `.dbinfo` file instead.

aeisenberg:
Closes #383.

See the heuristics in the issue.

### DIFF
--- a/extensions/ql-vscode/src/databases-ui.ts
+++ b/extensions/ql-vscode/src/databases-ui.ts
@@ -9,6 +9,7 @@ import { clearCacheInDatabase, UserCancellationException } from './run-queries';
 import * as qsClient from './queryserver-client';
 import { upgradeDatabase } from './upgrades';
 import { importArchiveDatabase, promptImportInternetDatabase } from './databaseFetcher';
+import * as fs from 'fs-extra';
 
 type ThemableIconPath = { light: string; dark: string } | string;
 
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -361,13 +362,34 @@ export class DatabaseUI extends DisposableObject {
...
+   *
+   * 1. If the selected URI to add is a file, choose the containing directory
+   * 2. If the selected URI is a directory matching db-*, choose the containing directory
+   * 3. choose the current directory
+   *
+   * @param uri
+   */
+  private async fixDbUri(uri: Uri): Promise<Uri> {
+    let dbPath = uri.fsPath;
+    if ((await fs.stat(dbPath)).isFile()) {
```
jcreedcmu, line 387: Does this break selecting a `.zip` archive of a db? If not, maybe refine the comment slightly to emphasize that we're only talking about the codepath that is concerned with intentionally selecting a database *directory*.

jcreedcmu: Yeah, oops, that definitely is the case here. Just requesting slightly more context in the doc comment then.

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -361,13 +362,34 @@ export class DatabaseUI extends DisposableObject {
...
+      dbPath = path.dirname(dbPath);
+    }
+    if (path.basename(dbPath).startsWith('db-')) {
+      dbPath = path.dirname(dbPath);
+    }
+    return Uri.parse(`file:${dbPath}`, true);
```
jcreedcmu, line 393: I think there's a `Uri.file` method to construct a `Uri` out of a path

jcreedcmu: (Same applies to all the 
```
Uri.parse(`string interpolation`)
``` 
going on below)

extensions/ql-vscode/test/data/library.qll
```diff
@@ -1,3 +1,3 @@
 predicate foo() {
- 1 == 1
-}
\ No newline at end of file
+ 1 = 1
```
jcreedcmu, line 2: hah, how long was this broken, I wonder

aeisenberg: Oh...didn't even mean to commit this. I'll extract to a separate commit.

aeisenberg:
Added a changelog update, but I don't think this is worth pinging the documentation team.

PR #401, opened 2020-05-26
Implement pagination for BQRS results.

jcreedcmu:
Make pagination work for raw tables, leaving alert tables alone for now.

This is worth reviewing, although I'm aware of a couple of improvements I want to make imminently. There is some code duplication between the first render and subsequent page fetches that I want to refactor, and the visual appearance of the pagination widget is definitely worth cleaning up a bit.

This should be a no-op if the experimental flag is off, but that deserves either some unit tests or slightly more strenous manual testing.

### DIFF
--- a/extensions/ql-vscode/src/interface.ts
+++ b/extensions/ql-vscode/src/interface.ts
@@ -450,19 +450,14 @@ export class InterfaceManager extends DisposableObject {
     const adaptedSchema = adaptSchema(schema);
     const resultSet = adaptBqrs(adaptedSchema, chunk);
 
-    let parsedResultSets: ParsedResultSets;
-    if (resultSet !== undefined) {
-      parsedResultSets = {
-        t: 'ExtensionParsed',
-        pageNumber,
-        resultSet,
-        numPages: numPagesOfResultSet(resultSet),
-        selectedTable: selectedTable,
-        resultSetNames
-      };
-    }
-    else
-      parsedResultSets = { t: 'WebviewParsed', selectedTable };
+    const parsedResultSets: ParsedResultSets = {
+      t: 'ExtensionParsed',
+      pageNumber,
+      resultSet,
+      numPages: numPagesOfResultSet(resultSet),
+      selectedTable: selectedTable,
+      resultSetNames
+    };
 
     await this.postMessage({
       t: "setState",
--- a/extensions/ql-vscode/src/interface.ts
+++ b/extensions/ql-vscode/src/interface.ts
@@ -10,14 +10,15 @@ import { CodeQLCliServer } from './cli';
 import { DatabaseItem, DatabaseManager } from './databases';
 import { showAndLogErrorMessage } from './helpers';
 import { assertNever } from './helpers-pure';
-import { FromResultsViewMsg, Interpretation, INTERPRETED_RESULTS_PER_RUN_LIMIT, IntoResultsViewMsg, QueryMetadata, ResultsPaths, SortedResultSetInfo, SortedResultsMap, InterpretedResultsSortState, SortDirection } from './interface-types';
+import { FromResultsViewMsg, Interpretation, INTERPRETED_RESULTS_PER_RUN_LIMIT, IntoResultsViewMsg, QueryMetadata, ResultsPaths, SortedResultSetInfo, SortedResultsMap, InterpretedResultsSortState, SortDirection, RAW_RESULTS_PAGE_SIZE } from './interface-types';
 import { Logger } from './logging';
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/interface.ts
```diff
@@ -364,18 +374,37 @@ export class InterfaceManager extends DisposableObject {
...
-        const chunk = await this.cliServer.bqrsDecode(results.query.resultsPaths.resultsPath, schema.name);
+        const chunk = await this.cliServer.bqrsDecode(results.query.resultsPaths.resultsPath, schema.name, RAW_RESULTS_PAGE_SIZE, schema.pagination?.offsets[0]);
         const adaptedSchema = adaptSchema(schema);
         const resultSet = adaptBqrs(adaptedSchema, chunk);
-        resultSets.push(resultSet);
+
+        return {
+          t: 'ExtensionParsed',
+          pageNumber: 0,
+          numPages: numPagesOfResultSet(resultSet),
```
aeisenberg, line 398: We might want to change this to `numResults` and `pageSize`. This gives us the option of configurable page sizes.

extensions/ql-vscode/src/interface.ts
```diff
@@ -384,14 +413,67 @@ export class InterfaceManager extends DisposableObject {
...
       database: results.database,
       shouldKeepOldResultsWhileRendering,
       metadata: results.query.metadata
     });
   }
 
+  /**
+   * Show a page of raw results from the chosen table.
+   */
+  public async showPageOfResults(selectedTable: string, pageNumber: number): Promise<void> {
```
aeisenberg, line 427: I'd rather see `pageNumber` changed to `pageSize` and `offset`. For now, we can keep a single, default page size, but we should allow users to choose this in the future. It shouldn't be hard.

extensions/ql-vscode/src/view/result-tables.tsx
```diff
@@ -75,23 +77,66 @@ export class ResultTables
     return resultSets;
   }
 
+  private getResultSetNames(resultSets: ResultSet[]): string[] {
+    if (this.props.parsedResultSets.t === 'ExtensionParsed') {
+      return this.props.parsedResultSets.resultSetNames.concat([ALERTS_TABLE_NAME]);
+    }
+    else {
```
aeisenberg, line 84: And this branch is deprecated, right? To be removed when we only send paginated results? If so, can you mark that?

extensions/ql-vscode/src/view/result-tables.tsx
```diff
@@ -75,23 +77,66 @@ export class ResultTables
...
+    }
+  }
+
+  /**
+   * Holds if we have a result set obtained from the extension that came
+   * from the ExtensionParsed branch of ParsedResultSets. This is evidence
+   * that the user has the experimental flag turned on that allows extension-side
+   * bqrs parsing.
+   */
+  paginationAllowed(): boolean {
```
aeisenberg, line 95: I wonder if it's even worth it to allow for backwards compatibility. I think it would make things simpler to just get rid of the old way of doing things. What do you think?

extensions/ql-vscode/src/view/result-tables.tsx
```diff
@@ -118,24 +163,81 @@ export class ResultTables
...
+        selectedTable,
+      });
+    };
+    return <span>
+      <button onClick={prevPage} >&lt;</button>
+      <input value={this.state.selectedPage} onChange={onChange}
+        onBlur={e => choosePage(e.target.value)}
+        onKeyDown={e => { if (e.keyCode === 13) choosePage((e.target as HTMLInputElement).value); }}
+      />
+      <button value=">" onClick={nextPage} >&gt;</button>
```
aeisenberg, line 213: Do we need a `value` attribute?

extensions/ql-vscode/src/view/result-tables.tsx
```diff
@@ -118,24 +163,81 @@ export class ResultTables
...
+      vscode.postMessage({
+        t: 'changePage',
+        pageNumber: Math.min(resultSets.pageNumber + 1, resultSets.numPages - 1),
+        selectedTable,
+      });
+    };
+    return <span>
+      <button onClick={prevPage} >&lt;</button>
+      <input value={this.state.selectedPage} onChange={onChange}
+        onBlur={e => choosePage(e.target.value)}
```
aeisenberg, line 210: Do we need `onBlur` and `onKeyDown` if we are already doing `onChange`?

aeisenberg:
All right. Let’s get this in.

PR #403, opened 2020-05-26
Introduce codeql code scanning action

aeisenberg:
Adds a new workflow to run codescanning on every commit.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,22 @@
+name: "Code Scanning - CodeQL"
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 0 * * 0'
+
+jobs:
+  codeql:
+    strategy:
+      fail-fast: false
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


aeisenberg:
Closes #402

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,21 @@
+name: "Code Scanning - CodeQL"
+
+on:
```
jcreedcmu, line 3: Does this want to be `on: [push, pull_request]` instead?

adityasharad:
`pull_request` is not yet supported and might not work correctly - `push` is what you want (and it will handle PRs created from this repo, though not yet from upstream).

jcreedcmu:
Ok, in that case I merely misdiagnosed the failure of codescanning to immediately run on the PR itself; subsequent PRs seem to be working great with it.

aeisenberg:
@adityasharad so that means in order to run codeql on forks, we need to keep the LGTM integration? If that's the case, I wonder if this code scanning workflow is useful at all.

PR #404, opened 2020-05-26
Remove pull_request from actions

jcreedcmu:


.github/workflows/main.yml
```diff
@@ -1,5 +1,5 @@
 name: Build Extension
-on: [push, pull_request]
+on: push
```
aeisenberg, line 2: Are we sure we want to do this? And if we do, we need this:

```
  on: [push]
```

or 

```
  on:
    push:
```

jcreedcmu:
The motivation for doing it was @adityasharad saying pull_request wasn't supported. You're correct about my yaml error, though, will fix.

alexet:
It's only not supported for the codeql action. It is fine for the checks action and ensures that the tests are run on PRs from users who haven't enabled actions on their fork.

jcreedcmu:
Ah, ok, so I misunderstood that, then. Closing as this is no longer necessary.

alexet:
You may still want to delete it from `codeql.yml`

jcreedcmu:
Correct; but https://github.com/github/vscode-codeql/pull/405 is already doing that.

PR #405, opened 2020-05-26
Allow download from lgtm

aeisenberg:
Adds a new command to download databases from lgtm. It's working,
though need to create better error messages on failure.

## Checklist


extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -1,43 +1,121 @@
...
+export async function promptImportLgtmDatabase(
+  databasesManager: DatabaseManager,
+  storagePath: string
+): Promise<DatabaseItem | undefined> {
+  let item: DatabaseItem | undefined = undefined;
+
+  try {
+    const lgtmUrl = await window.showInputBox({
+      prompt:
+        "Enter the project URL on LGTM (e.g., https://lgtm.com/projects/g/github/codeql)",
```
jcreedcmu, line 83: Any thoughts on why not to require simply the owner/project string, in this example, just `github/codeql`? I feel like that would save a bit of validation complexity.

jcreedcmu: Or do we not have confidence the prefix is always going to be `https://lgtm.com/projects/g/` in the future?

aeisenberg: I was thinking it's just easier for a user to copy the full URL without thinking about which substring to copy and paste it here. And the other thing is that I don't really know what the `g` is.

Maybe I should first get some clarification on what that`g` means.

shati-patel: The `g` means that it's a GitHub project (LGTM also has some [GitLab](https://lgtm.com/projects/gl/sj_vs/lgtm-gitlab-integration-test/) and [BitBucket](https://lgtm.com/projects/b/bitbucket/django_features/) ones.)

jcreedcmu: Hm, that's fair. In that case I might suggest heuristically accepting
`https://lgtm.com/projects/g/github/codeql/?mode=list`
and
`https://lgtm.com/projects/g/github/codeql/alerts`
since the former is what `https://lgtm.com/projects/g/github/codeql/` seems to redirect to that, and `.../alerts` urls are suggested by the lgtm interface in places?

Dunno, not important for this PR necessarily, just a nicety since it's possible to DWIM in these cases.

aeisenberg: I was thinking it's just easier for a user to copy the full URL without thinking about which substring to copy and paste it here. And the other thing is that I don't really know what the `g` is.

Maybe I should first get some clarification on what that`g` means.

jcreedcmu: Ok, so how about just checking that we're at `(www.)lgtm.com`, and there are at least (instead of exactly) 4 components to the path, and the first is `projects`, and we pull out the third and fourth?

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -218,3 +322,72 @@ async function findDirWithFile(dir: string, ...toFind: string[]): Promise<string
...
+    if (uri.scheme !== "https") {
+      return false;
+    }
+
+    if (uri.authority !== "lgtm.com" && uri.authority !== "www.lgtm.com") {
+      return false;
+    }
+
+    const paths = uri.path.split("/").filter((segment) => segment);
+    return paths.length === 4 && paths[0] === "projects" && paths[1] === "g";
```
aibaars, line 342:
```suggestion
    return paths.length >= 4 && paths[0] === "projects";
```
The URL pattern is `https://lgtm.com/projects/{provider}/{org}/{name}` . In addition to GitHub.com(`g`), LGTM currently hosts projects from Bitbucket (`b`), GitLab (`gl`) and plain git (`git`). 

After the `{provider}/{org}/{name}` path components, there may be the components related to sub pages. I think it would be best to accept those too.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -218,3 +322,72 @@ async function findDirWithFile(dir: string, ...toFind: string[]): Promise<string
...
+    }
+    return `https://lgtm.com/${[
+      "api",
+      "v1.0",
+      "snapshots",
+      projectJson.id,
+      language,
+    ].join("/")}`;
+  } catch (e) {
+    throw new Error(`Invalid LGTM URL: ${lgtmUrl}`);
```
aibaars, line 374: You should probably include the error message that LGTM returns on error. If a user provides a valid LGTM URL but the wrong language, we would wrongly complain that the LGTM URL is invalid.

extensions/ql-vscode/src/databaseFetcher.ts
aibaars: > Ok, so how about just checking that we're at `(www.)lgtm.com`, and there are at least (instead of exactly) 4 components to the path, and the first is `projects`, and we pull out the third and fourth?

Yes, indeed. Except that you should pull out the second, third and fourth. The URL pattern is `/projects/{provider}/{org}/{name}/sub_pages...`

extensions/ql-vscode/src/databaseFetcher.ts
aeisenberg: Makes sense.

extensions/ql-vscode/src/databaseFetcher.ts
aeisenberg: Sure. I can do something, but just note that a user can't request an invalid language since we only allow a user to select from the languages exposed in the project metadata.

aibaars: That's true. Although more precise error message are likely to be helpful in case we ever need to debug a failure.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -339,7 +342,7 @@ function looksLikeLgtmUrl(lgtmUrl: string | undefined): lgtmUrl is string {
     }
 
     const paths = uri.path.split("/").filter((segment) => segment);
-    return paths.length === 4 && paths[0] === "projects" && paths[1] === "g";
+    return paths.length >= 4 && paths[0] === "projects";
```
jcreedcmu, line 345: Maybe just leave a comment in here explaining the URL munging going on --- basically a copy-paste of what @aibaars told us would be good. Otherwise this whole PR lgtm.

aeisenberg: Sure

extensions/ql-vscode/src/databaseFetcher.ts
aeisenberg: Sure. I'll send these messages to the log files.

shati-patel:
Minor confusion: The extension lets me select a Python database for https://lgtm.com/projects/g/mrdoob/three.js, but then gives the error `invalid signature: 0x6f63227b`.

Turns out there isn't actually a Python database available: 
image

Is it possible to make this clearer in the logs?

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Better formatting and autoindentation when adding QLDoc comments to `.ql` and `.qll` files.
 - Allow for more flexibility when opening a database in the workspace. A user can now choose the actual database folder, or the nested `db-*` folder.
+- Add ability for users to download databases directly from LGTM.
```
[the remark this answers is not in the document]
aibaars: Indeed, LGTM.com only.

It would actually work for public projects on LGTM enterprise installations that have anonymous access enabled. This is a rather rare combination, so let's ignore that to avoid confusion ;-)

aibaars:
> Minor confusion: The extension lets me select a Python database for https://lgtm.com/projects/g/mrdoob/three.js, but then gives the error `invalid signature: 0x6f63227b`.
> 
> Turns out there isn't actually a Python database available:
> image
> 
> Is it possible to make this clearer in the logs?

Good point. That project does have python code, but the latest snapshot build failed for python. . As a result the database is not available. Probably best to present the failure returned by the LGTM API. On error there is a JSON response like this
```
{"code":404,"
error":"The specified language could not be found for the project"
}
```

aeisenberg:
I think this addresses all comments in the PR so far. Unfortunately, the current download mechanism does not support tar.gz files or other kinds of compression. I'll open a new issue for that.

aeisenberg:
See #410.

Issue #342, opened 2020-04-28
Feature request: Export query results (as SARIF/CSV)


jcreedcmu:
After a query is run, we should allow exporting the results as a SARIF file --- in the implementation, this will actually be nearly trivial, as we already use the codeql cli to create a SARIF file during interpretation, and store it in a workspace-local state directory. A solution could simply be to provide a way of conveniently opening this file within vscode (from the query history view, or from the results panel) after which point the user can save it wherever they like.

Also consider exporting as CSV for parity with the eclipse plugin.

- [x] SARIF
- [x] CSV

rajatdiptabiswas:
Would you kindly let me know what exactly needs to be done so that I can try working on this?
Please mention anything I should read on or know about before starting this. It will help out a lot.

Thanks in advance!

jcreedcmu:
Thanks for offering to contribute! I'll try to explain in a bit more detail.

The function [interpretBqrs](https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/src/cli.ts#L513) reads in a bqrs (binary query result set) file and produces a SARIF file. We want the user to be able to view and save this file elsewhere if they want, when a query is finished running. The place in the filesystem where that SARIF file is stored is already chosen and passed down there, if we look up the stack at callers of interpretBqrs, which are [interpretResults](https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/src/query-results.ts#L151) and [getTruncatedResults](https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/src/interface.ts#L387) and in [interpretResultsInfo](https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/src/interface.ts#L443). So there we see `query.resultsPaths` has the appropriate paths for a `query` object of type `QueryInfo`.

One thing that could be done is, for an item in the query history (a thing of type [CompletedQuery](https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/src/query-results.ts#L12)) add a context menu command "Show SARIF Results" which requests vscode open the file that corresponds to the interpreted results path of the query of the query history item that you had selected. If you need any pointers to vscode api internals, or to examples of other context commands in the extension source, lmk!

rajatdiptabiswas:
Seems a bit complicated but I'll try my best.
I have an exam coming up, but I shall try to get it done as soon as possible.

I'll let you know if I run into any issues!

rajatdiptabiswas:
[...]

PR #407, opened 2020-05-27
Allow viewing SARIF from query history view

jcreedcmu:
Addresses the SARIF part of https://github.com/github/vscode-codeql/issues/342. Could still use CSV export.

`@github/product-docs-dsp` any need for docs for this?

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -74,7 +74,7 @@ class HistoryTreeDataProvider implements vscode.TreeDataProvider<CompletedQuery>
   constructor(private ctx: ExtensionContext) {
   }
 
-  getTreeItem(element: CompletedQuery): vscode.TreeItem {
+  async getTreeItem(element: CompletedQuery): Promise<vscode.TreeItem> {
     const it = new vscode.TreeItem(element.toString());
 
     it.command = {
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/query-history.ts
```diff
@@ -83,6 +83,11 @@ class HistoryTreeDataProvider implements vscode.TreeDataProvider<CompletedQuery>
       arguments: [element],
     };
 
+    // Mark this query history item according to whether it has a
+    // SARIF file so that we can make context menu items conditionally
+    // available.
+    it.contextValue = element.query.hasInterpretedResults() ? 'interpretedResultsItem' : 'rawResultsItem';
```
aeisenberg, line 89: Neat.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -157,15 +157,22 @@ export class QueryInfo {
...
       logger.log("Cannot produce interpreted results since the database does not have a .dbinfo or codeql-database.yml file.");
     }
     return hasMetadataFile;
   }
+
+  /**
+   * Holds if this query actually has produced interpreted results.
+   */
+  hasInterpretedResults(): boolean {
+    return fs.existsSync(this.resultsPaths.interpretedResultsPath);
```
aeisenberg, line 174: Can we make this function async? Otherwise, this is blocking.

aeisenberg:
And looks like the unit tests caught a legitimate issue that would have been hard to find otherwise. Nice. :)

extensions/ql-vscode/src/run-queries.ts
jcreedcmu: Ah, I was afraid the `getTreeItem` had to by synchronous, but it looks like vscode permits it to return a `Promise`. Happy to make it `async` then.

jcreedcmu:
> And looks like the unit tests caught a legitimate issue that would have been hard to find otherwise. Nice. :)

Ah, right, I did leave out the command palette exclusion to see if it would trigger that, then promptly forgot that I had done so. Thankfully, unit tests don't forget :)

PR #421, opened 2020-06-01
Download platform-specific releases if they are available.

jcreedcmu:
Choose platform-specific release asset based on platform, falling back
to `codeql.zip` if platform is unknown, or platform-specific release
asset is unavailable.

### DIFF
--- a/extensions/ql-vscode/src/distribution.ts
+++ b/extensions/ql-vscode/src/distribution.ts
@@ -377,7 +377,7 @@ class ExtensionSpecificDistributionManager {
       release => {
         const matchingAssets = release.assets.filter(asset => asset.name === requiredAssetName);
         if (matchingAssets.length === 0) {
-          // For example, this could be a release with only platform-specific assets.
+          // For example, this could be a release with no platform-specific assets.
           logger.log(`INFO: Ignoring a release with no assets named ${requiredAssetName}`);
           return false;
         }
--- a/extensions/ql-vscode/src/distribution.ts
+++ b/extensions/ql-vscode/src/distribution.ts
@@ -378,7 +378,7 @@ class ExtensionSpecificDistributionManager {
   private async getLatestRelease(): Promise<Release> {
     const requiredAssetName = this.getRequiredAssetName();
     logger.log(`Searching for latest release including ${requiredAssetName}. `);
-    const release = await this.createReleasesApiConsumer().getLatestRelease(
+    return this.createReleasesApiConsumer().getLatestRelease(
       this._versionRange,
       this._config.includePrerelease,
       release => {
--- a/extensions/ql-vscode/src/distribution.ts
+++ b/extensions/ql-vscode/src/distribution.ts
@@ -297,6 +297,10 @@ class ExtensionSpecificDistributionManager {
         `but encountered an error: ${e}.`);
     }
 
+    // NOTE: We are tacitly assuming that at this point
+    // release.assets.length === 1. This is ensured by
+    // getLatestRelease for now. Reconsider this if a cli release ever
+    // has more than one asset.
     const assetStream = await this.createReleasesApiConsumer().streamBinaryContentOfAsset(release.assets[0]);
     const tmpDirectory = await fs.mkdtemp(path.join(os.tmpdir(), "vscode-codeql"));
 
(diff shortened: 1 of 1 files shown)


hmakholm:
I'm not expecting a need to _download_ more than one asset from a release.

How painful would it be to look one release back for a platform-specific download (such that in the transition period where we have a fake one-asset "newest" release for the benefit of the previous extension, users of the updated extension would still get the benefit of a smaller download?)

jcreedcmu:
Okay, can try that; I do want (something like) this code in the extension as soon as possible, though, so that the eventual clean mechanism of having multiple assets in one release can be introduced.

extensions/ql-vscode/src/distribution.ts
```diff
@@ -297,6 +297,10 @@ class ExtensionSpecificDistributionManager {
         `but encountered an error: ${e}.`);
     }
 
+    // NOTE: We are tacitly assuming that at this point
```
aeisenberg, line 300: I thought this is no longer true?

extensions/ql-vscode/src/distribution.ts
```diff
@@ -354,26 +358,46 @@ class ExtensionSpecificDistributionManager {
...
+    const release = await this.createReleasesApiConsumer().getLatestRelease(
       this._versionRange,
       this._config.includePrerelease,
       release => {
-        // FIXME: Look for platform-specific codeql distribution if available
-        // https://github.com/github/vscode-codeql/issues/417
-        const matchingAssets = release.assets.filter(asset => asset.name === 'codeql.zip');
+        const matchingAssets = release.assets.filter(asset => asset.name === requiredAssetName);
         if (matchingAssets.length === 0) {
           // For example, this could be a release with only platform-specific assets.
```
aeisenberg, line 380:
```suggestion
          // For example, this could be a release with no platform-specific assets.
```

extensions/ql-vscode/src/distribution.ts
```diff
@@ -354,26 +358,46 @@ class ExtensionSpecificDistributionManager {
...
         if (matchingAssets.length > 1) {
-          logger.log("WARNING: Ignoring a release with more than one asset named codeql.zip");
+          logger.log(`WARNING: Ignoring a release with more than one asset named ${requiredAssetName}`);
           return false;
         }
         return true;
       }
     );
+    // Actually filter assets to the unique one that we require.
+    release.assets = release.assets.filter(asset => asset.name === requiredAssetName);
```
aeisenberg, line 392: Do we want to first look for the platform specific asset, and only if that one isn't found, we then look for the generic asset?

extensions/ql-vscode/src/distribution.ts
jcreedcmu: After some consideration, I felt persuaded that your argument that downloading assets serially was bad enough to warrant not attempting to go so far out of my way to be future proof here. This PR is trying to be the minimal step forward that at least accounts for the change in cli distribution that we *do* expect to make.

extensions/ql-vscode/src/distribution.ts
jcreedcmu: I believed (correct me if I'm wrong @hmakholm) the plan was to, for the time being, make releases in pairs, one (call it `v3.4.5`) which contains the all-platform `codeql.zip`, and another (named something like `v3.4.5-platform`) which contains separate platform-specific releases. I believe @henrymercer's latest revision to `ReleasesApiConsumer.getLatestRelease` will find the latest release _which satisfies `additionalCompatibilityCheck`_. So when `v3.4.5` and `v3.4.5-platform` have been released, this code, on for example linux, will download `v3.4.5-platform` because it will be the latest release _that contains `codeql-linux64.zip`_.

However, now that I think about it, I'm still not sure I'm convinced this plan is watertight. With the version of `distribution.ts` that we've had for a while, namely https://github.com/github/vscode-codeql/blob/7d23a833b178690eede71aa5e292c3c4b369ccab/extensions/ql-vscode/src/distribution.ts is it not the case that the extension will take whichever of `v3.4.5-platform` and `v3.4.5` has been released chronologically most recently? And if that is true, won't there be some moment, no matter how we order their release, when `v3.4.5-platform` is the most recent, and the extension has a chance of downloading a release with multiple release assets?

hmakholm: @jcreedcmu, I must addmt I have tuned a bit out of the discussion, so personally my plan for the time being is to wait until the next release draws near and then ask #codeql-extensions what a timing and naming strategy would be that (a) works with all supported version of the extension, and (b) under that constraint, gives the best experience for users of the _latest_ version of the extension.

jcreedcmu: (Or at least it is if we are committed to a cli release schedule that always delivers platform specific zips in some form - merging this PR amounts to trusting that commitment has been made)

jcreedcmu: @hmakholm that is entirely reasonable; my earlier worry was that a naming strategy that satisfied both (a) and "involves any platform-specific assets at all" _didn't exist_, but I think now you've convinced me that it does.

extensions/ql-vscode/src/distribution.ts
```diff
@@ -297,6 +297,17 @@ class ExtensionSpecificDistributionManager {
...
+    const assets = release.assets.filter(asset => asset.name === requiredAssetName);
+    if (assets.length === 0) {
+      throw new Error(`Invariant violation: chose a release to install that didn't have ${requiredAssetName}`);
+    }
+    if (assets.length > 1) {
+      logger.log('WARNING: chose a release with more than one asset to install, found ' +
+        assets.map(asset => asset.name).join(', '));
+    }
+
     const assetStream = await this.createReleasesApiConsumer().streamBinaryContentOfAsset(release.assets[0]);
```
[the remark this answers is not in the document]
jcreedcmu: Ack, good catch.

extensions/ql-vscode/src/distribution.ts
```diff
@@ -354,26 +365,37 @@ class ExtensionSpecificDistributionManager {
...
+    if (os.platform() === 'linux') return 'codeql-linux64.zip';
+    if (os.platform() === 'darwin') return 'codeql-osx64.zip';
+    if (os.platform() === 'win32') return 'codeql-win64.zip';
+    return 'codeql.zip';
+  }
+
   private async getLatestRelease(): Promise<Release> {
-    return await this.createReleasesApiConsumer().getLatestRelease(
+    const requiredAssetName = this.getRequiredAssetName();
+    logger.log(`Searching for latest release including ${requiredAssetName}. `);
```
henrymercer, line 380:
```suggestion
    logger.log(`Searching for latest release including ${requiredAssetName}.`);
```

extensions/ql-vscode/src/distribution.ts
```diff
@@ -354,26 +365,37 @@ class ExtensionSpecificDistributionManager {
...
+    const release = await this.createReleasesApiConsumer().getLatestRelease(
```
henrymercer, line 381:
```suggestion
    return this.createReleasesApiConsumer().getLatestRelease(
```

PR #422, opened 2020-06-01
Chore: Retry tests more aggressively on windows

jcreedcmu:
There are some flaky CI test failures that manifest only as a message
like

    [main 2020-06-01T16:09:47.671Z] [VS Code]: render process crashed!

(and only afaict on windows) which I am not sure how to detect at the
moment. If that message is occurring in the exception caught at this
stage, we can check for it.


jcreedcmu:
An example: https://github.com/github/vscode-codeql/runs/727473886

jcreedcmu:
I would too, but don't know how best to investigate further. I see some hints on https://stackoverflow.com/questions/3847745/systrace-for-windows for "things that might be analogous to strace on windows".

PR #423, opened 2020-06-01
Use semver instead of homegrown version data.

jcreedcmu:
Two assumptions that were true before that are no longer true:
 - That a prelease string may contain '.'. The `semver` module parses
   this as a delimiter between multiple prerelease specifiers.

       > semver.parse('v1.0.0-a.0')
       SemVer {
         options: { loose: false, includePrerelease: false },
         loose: false,
         includePrerelease: false,
         raw: 'v1.0.0-a.0',
         major: 1,
         minor: 0,
         patch: 0,
         prerelease: [ 'a', 0 ],
         build: [],
         version: '1.0.0-a.0' }

 - That a missing prerelease or build specifier is represented
   as `undefined`. It's now represented as `[]`.

I don't think either of these cause problems, but I'm noting them here
in case anyone can think of a reason why they might.

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -26,6 +26,7 @@ dependencies:
   '@types/react': 16.9.23
   '@types/react-dom': 16.9.5
   '@types/sarif': 2.1.2
+  '@types/semver': 7.2.0
   '@types/sinon': 7.5.2
   '@types/sinon-chai': 3.2.3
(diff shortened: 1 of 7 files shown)


aeisenberg:
I don't think either of those would be a problem. But, just to call it out, could you add a test for each of them and maybe add a comment, too? I see that you updated the tests to account for the changes, but maybe both the old and new syntaxes should be tested.

jcreedcmu:
Superceded by #425.

Issue #428, opened 2020-06-02
Removing leading whitespace when pasting code


RasmusWL:
**Describe the bug**
My CodeQL VSCode extension is removing leading whitespace when pasting code into ql/qll files :confused: Pressing CTRL-Z brings back the leading-whitespace to my pasted code.

**To Reproduce**
Paste some QL code into a `.ql` or `.qll` file, for example.

```
predicate foo() {
  none()
}
```

What will actually be pasted is

```
predicate foo() {
none()
}
```

**Expected behavior**
My code gets pasted as is

**Additional context**

- I'm using extension version 1.2.1 with locally compiled codeql binary (which I just rebuild from newest master). I disabled all other extensions in VSCode (just to be sure), and restarted VSCode as well. 
- Manually changing my installed version to 1.2.0 resolves this problem. 
- I'm running Ubuntu 18.04 if that helps any.

jcreedcmu:
I can repro that there is strange indentation behavior, except that I get _extra_ indentation:
```
predicate foo() {
    none()
  }
```

RasmusWL:
Oh, that might be because I'm using a standard tab-size of 4 spaces for QL code (since we still haven't made the move for Python).

aeisenberg:
I can reproduce this. When I change the following setting to `advanced` (default is `full`), the paste behaves as I expect it to.

    "editor.autoIndent": "advanced"

The difference between advanced and full is that full will "honor indentation rules defined by the language". This implies that the indentation rules are being applied incorrectly.

PR #429, opened 2020-06-02
Fix bad indentation on paste

aeisenberg:
I don't fully understand why this is working differently, but these changes enable proper behavior on pasting ql into the editor.

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/master/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.


aeisenberg:
Fixes #428

extensions/ql-vscode/src/languageSupport.ts
```diff
@@ -17,11 +17,14 @@ export function install() {
   // setLanguageConfiguration requires a regexp for the wordpattern, not a string
   langConfig.wordPattern = new RegExp(langConfig.wordPattern);
   langConfig.onEnterRules = onEnterRules;
+  langConfig.indentationRules = {
+		increaseIndentPattern: /^((?!.*?\/\*).*\*\/)?\s*[\}\]].*$/,
+		decreaseIndentPattern: /^((?!\/\/).)*(\{[^}"'`]*|\([^)"'`]*|\[[^\]"'`]*)$/
+	};
```
alexet, line 23: This still looks to me like these are back to front. You seem to have `}` and `]` increasing the indentation and `{` and [`  increasing it.

jcreedcmu: These are the same patterns suggested by https://code.visualstudio.com/api/language-extensions/language-configuration-guide so as long as they are swapped as suggested this lgtm.

aeisenberg:
It does make sense to swap them, but that begs the question _why_ things seem to work even when they are mixed up.

jcreedcmu:
For me, they don't; I get incorrect indentation when pasting with them swapped, e.g.
```
predicate foo() {
exists(int x | x = x
)
}
```

aeisenberg:
This is still not perfect, and I think fundamentally there are limitations in how vscode works (that are not present in Eclipse), so we are never going to be able to get a pasting and formatting experience in vscode.

jcreedcmu:
Hm, as of this commit now I can reproduce the bad behavior
```
predicate foo() {
    exists(int x | x = x
    )
    }
```
again, which was what I saw earlier.

jcreedcmu:
Ok --- I think I realize that the failure I'm seeing is (now) simply vscode *not* formatting pasted code in a way that it *could* format it, rather than actively *mis*formatting it. In that case this lgtm.

PR #431, opened 2020-06-05
Add compare view

### LINKED ISSUE #394: Add ability to compare and diff pairs of result sets
## Feature

A user should be able to determine which entries have been added, deleted, or changed between a table of one query and the same table in another query.

## Implementation

1. User right-clicks on a query in the query view
1. User selects the _Compare with..._ command
1. A quick options panel appears with the other available queries to compare, sorted by database and with queries from the current database on top. (Alternatively, we can restict to show only queries in the current database...)
1. User selects the desired query
1. A new query compare web view opens for these queries.
   - There is a dropdown where users can select a table to compare (only tables with shared names are available)
   - There is a short description of what is being compared, including stats (# num added, num removed, etc)
   - A table with two columns in the main section.
   - Each table row corresponds to one of: an addition, removal, or deletion of a query row between the two queries. (Highlighted appropriately, see QL4E for ideas)
   - Use a similar comparison algorithm to what QL4E has.
   - Hyperlinks should be appropriately available.

## Open questions

* Should we limit the max number of rows in the view?
* Should we only allow comparisons between queries in the same database?
* Should we allow the ignoring of one or more columns (eg- the id column may be different between queries, but still correspond to the same element)?
* Should there be multiple comparisons view openable at once?
* Can we compare path queries or other similar kinds of queries with nested tables?

## Messages sent from extension to webview
This message is sent to the webview in order to compare two queries

```
{
  t: 'compare'
  stats: {
    added: <num rows added>,
    removed: <num rows removed>,
    changed: <num rows changed>,
    fromQuery: { ...details about the query we are comparing from... }, id, description, row count, time to run, date ran, etc...
    toQuery: { ...details about the query we are comparing to... },
  },
  rows: [...] // array of rows in order: added, removed, or changed. I don't yet know what this looks like
  tablesToCompare: <list of tables available for comparison>
  currentTable: <table currently being compared>
}
```

## Messages sent from webview to extension

This message is sent to the extension when a user changes the table to view

```
{
  t: 'change-compare',
  newTable: <new table to compare>,
  fromId: <id of the query to compare from>
  toId: <id of the query to compare to>
}
```

aeisenberg:
This PR adds the compare view. See #394.

To compare 2 queries:

1. Run a query
2. Run another query that has the same number of columns, but slightly different results 
3. In the query history view, select a query and right click.
4. Select _Compare with..._
5. Choose the second query
6. The compare view should show up.

Fixes #394

## Checklist


dbartol:
To compare text files, you right click on one, select “Select for Compare”, then right click on another and select “Compare with Selected”. We should probably use the same approach to invoke our diffs.

aeisenberg:
@dbartol I never even knew that feature existed until you mentioned it here. I could try to do something similar. I'd like to first get the basic feature in and then work on ways to improve and fill it out completely.

aeisenberg:
The reason why there are so many commits right now is that I rebased on top of #437. When that change gets merged, this one should shrink.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -277,10 +301,10 @@ export class QueryHistoryManager {
   constructor(
     ctx: ExtensionContext,
     private queryHistoryConfigListener: QueryHistoryConfig,
-    selectedCallback?: (item: CompletedQuery) => Promise<void>
+    private selectedCallback: (item: CompletedQuery) => Promise<void>,
+    private doCompareCallback: (from: CompletedQuery, to: CompletedQuery) => Promise<void>,
   ) {
     this.ctx = ctx;
```
jcreedcmu, line 307: any reason not to replace `this.ctx = ctx` with an argument list field declaration also?

extensions/ql-vscode/src/compare/compare-interface.ts
```diff
@@ -0,0 +1,270 @@
...
+      from,
+      to,
+      selectedResultSetName
+    );
+    if (currentResultSetName) {
+      await this.postMessage({
+        t: 'setComparisons',
+        stats: {
+          fromQuery: {
+            // since we split the description into several rows
```
jcreedcmu, line 73: It's not clear to me that the user-defined label case works well here. When I set the label of the query-being-compared to to `[%t] custom`, what I see is just the query file name and not the string 'custom' appear anywhere in the view.

extensions/ql-vscode/src/compare/resultsDiff.ts
```diff
@@ -0,0 +1,58 @@
...
+ * 2. Queries have same number and order of columns
+ * 3. Rows are not changed or re-ordered, they are only added or removed
+ *
+ * @param fromResults the source query
+ * @param toResults the target query
+ *
+ * @throws Error when:
+ *  1. number of columns do not match
+ *  2. If either query is empty
+ *  3. If the queries are 100% disjoint
```
jcreedcmu, line 20: When I do a comparison of disjoint queries, the comparison view is forced visible, and I do get an error message saying the results are disjoint, but the comparison view still contains whichever comparison it did before, which I think is slightly misleading. I think I'd prefer --- but would be happy deferring to a follow-up PR if this is at all complicated --- having one of the possible states of the comparison view actually being "results disjoint" as opposed to `window.showErrorMessage`.

extensions/ql-vscode/src/query-history.ts
aeisenberg: Makes sense. Must have missed that one.

extensions/ql-vscode/src/languageSupport.ts
```diff
@@ -18,38 +18,38 @@ export function install() {
...
-    // e.g.  */|
-    beforeText: /^(\t|[ ])*[ ]\*\/\s*$/,
-    action: { indentAction: IndentAction.None, removeText: 1 }
-  },
-  {
-    // e.g.  *-----*/|
-    beforeText: /^(\t|[ ])*[ ]\*[^/]*\*\/\s*$/,
-    action: { indentAction: IndentAction.None, removeText: 1 }
-  }
+const onEnterRules: string[] = [
```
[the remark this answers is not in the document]
aeisenberg: That's a mistake.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -135,10 +135,19 @@ class HistoryTreeDataProvider implements vscode.TreeDataProvider<CompletedQuery>
...
+
+  }
+
+  get allHistory(): CompletedQuery[] {
+    return this.history;
   }
 
   refresh() {
-    this._onDidChangeTreeData.fire();
+    this._onDidChangeTreeData.fire(undefined);
```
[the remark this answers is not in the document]
aeisenberg: The next version of the vscode API converts this into a required argument. This is just avoiding a compile error for when we upgrade.

extensions/ql-vscode/src/compare/resultsDiff.ts
aeisenberg: Hmm...right. I should fix that.

extensions/ql-vscode/src/query-history.ts
aeisenberg: Actually...don't need it to be a class property at all. removed the field.

extensions/ql-vscode/src/view/result-tables.tsx
```diff
@@ -274,6 +286,14 @@ class ResultTable extends React.Component<ResultTableProps, {}> {
         {...this.props} resultSet={resultSet} />;
       case 'SarifResultSet': return <PathTable
         {...this.props} resultSet={resultSet} />;
+      default:
```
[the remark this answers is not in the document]
aeisenberg: Will remove.

extensions/ql-vscode/src/compare/compare-interface.ts
aeisenberg: Hmmm...let me see.

aeisenberg: Ugh...silly mistake. Fixed now.

extensions/ql-vscode/src/compare/resultsDiff.ts
aeisenberg: There are a few other error states. What I will do is ensure that the compare view will hide the columns and display the error message in this case.

jcreedcmu: Yeah, that sounds great.

jcreedcmu:
@dbartol about different ways to invoke comparison, I tend to agree with you that the most natural to me is select the two things and then invoke comparison --- but I think it's fine to get this PR in and then add that functionality afterwards, since multiselect brings a couple other complications with it. Showing an appropriate error message when any commands (including those other than comparison) are invoked with an inappropriate number of items selected is the main one that comes to mind.

jcreedcmu:
Looking good --- just one more suggestion, in `CompareTable.tsx`, how about adding a `style={{verticalAlign: "top"}}` to the first two `<td>`s in the `table.vscode-codeql__compare-body` ? When I compare two queries where the added rows greatly outnumber the deleted or vice-versa, the smaller one gets kinda visually lost by vertical centering.

aeisenberg:
Ping @github/product-docs-ds!

This is a fairly large change. I can describe here how it will work, or we can sync up over zoom. Up to you.

shati-patel:
Thanks @aeisenberg, exciting update! Feel free to drop any information into the docs issue if you want, though we'll also let you know if we need more help when we get round to documentation 😃

PR #432, opened 2020-06-06
Make a few internal packages publishable

dbartol:
No user-visible changes.

Months ago, I split some of our shared utilities out into separate npm packages to make it easier to reuse them in other repos. I finally got around to actually consuming them from my own repo, and there were a few changes required to make the process work. For now, the only packages I'm sharing are `semmle-vscode-utils` and `build-tasks`.

I renamed `semmle-vscode-utils` -> `@github/codeql-vscode-utils`, and `build-tasks` -> `@github/codeql-build-tasks`. The `@github/` scope is necessary because the GitHub package registry requires that all published packages be scoped to your organization. The `codeql-` prefix was added because the package namespace is shared with the entire `github` org, so it seemed polite to give a hint as to the packages' origin. Note that the source code directories have not been renamed. This seemed like unnecessary churn, and we can always do it later if anyone cares.

I added a Rush "version policy" for these two packages. This is how Rush determines when a package needs to be published, and how it should be versioned. `individualVersion` just means that each package in the group is versioned and published individually whenever it has changes.

The actual publishing process uses `rush change` to create change notes and bump the version numbers, followed by `rush publish` to actually publish to the GitHub package registry. The change note files you see added to the repo were generated this way. I didn't bother to automate the publishing process yet. For now, I'm the only one consuming these packages, and there's no need to burden anyone else who might make changes to these packages with extra work. I'll just publish them manually when I need to pick up new changes. The packages are published to our private repo for now, but should we decide to make them public in the future, we could always publish them somewhere else.

## Checklist


lib/semmle-vscode-utils/package.json
```diff
@@ -1,26 +1,27 @@
...
   "author": "GitHub",
-  "private": true,
-  "version": "0.0.1",
+  "private": false,
+  "version": "0.0.3",
   "publisher": "GitHub",
   "repository": {
     "type": "git",
-    "url": "https://github.com/github/vscode-codeql"
+    "url": "https://github.com/github/codeql-coreql-team.git"
```
aeisenberg, line 10: Should we be pointing to a private repo for a public package?

dbartol: Good question. I believe that this URL is necessary in order for GitHub Package Registry to route the published package to the right repo for permissions. The easiest thing to do is leave the private repo URL in the `package.json` as I currently have it, but I could also do the private publishing via a script that tweaks the repo URL before publishing.

dbartol:
> 1. What did tuyas for auto generating the changelog? Probably good for the rest of the project to do the same.

I used `rush change` to detect what packages have changed and prompt for change notes. If you look at [this commit](https://github.com/github/vscode-codeql/pull/432/commits/d3caf77f90478d7116463fb76d7e48c6e80160a7), you can see that it generates individual change note JSON files recording the description and SemVer severity of each change to each package. `rush publish` then aggregates all of those into a single `CHANGELOG.md` and determines how to bump the version number based on the "most-breaking" change note since the last version bump.

dbartol:
> 2. Should we add an action to deploy when the repo is tagged? (Could happen in a different PR.)

Since these packages are only being published privately, and since I'm the only consumer of them so far, I don't think it's worth the effort to automate yet. I that I'll just occasionally decide to pick up a new version, I'll go run `rush change` and summarize the changes myself, and then publish. If we decide to publish these packages publicly for some reason, we would want to automate for sure.

dbartol:
> 3. What process are using for deployment? Should probably update the readme for that (again, a separate PR would be fine.)

I'm not sure how much of that belongs in the public repo (related to your other comment about the private repo URL). If we're fine with having that in the public repo, I can add a `CONTRIBUTING-INTERNAL.md` or something with more info, and maybe even a VS Code task to do it. Otherwise, I can record the process somewhere else. Either way, how about I wait until the next time I need to publish to write that all up?

lib/semmle-vscode-utils/package.json
aeisenberg: Ah...I see. It's probably fine to keep it this way. If we ever get to the point where we want to make this public, we can change. We could even publish to this repo's GPR.

adityasharad:
Passing suggestion before you publish the packages regularly: may I suggest `codeql-build-tasks` become `codeql-vscode-build-tasks` or `vscode-codeql-build-tasks`, just so it's not confused with more general tools for CodeQL builds and analysis?

dbartol:
@adityasharad `codeql-build-tasks` isn't just for VS Code, although I suppose there are a couple tasks in there that are only useful for VS Code extensions. How about `codeql-gulp-tasks`?

PR #437, opened 2020-06-09
Force use of single quotes in formatter settings

aeisenberg:
This PR enforces double quotes in the formatting rules. Turns out this was never enforced before, even though that was (mostly) the de facto style. 

This PR also will fail linting (and hence any `git push`) if there are any formatting warnings.

Since reformatting is now happening on commit, it means that running `git commit` will take longer for `eslint` to run. I'm ok to remove this, and to run `eslint` on push instead (ie- a push will fail if formatting is off).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -33,6 +33,6 @@
     // This is necessary so that eslint can properly resolve its plugins
     "resolvePluginsRelativeTo": "./extensions/ql-vscode"
   },
-  // Force this to false since this will cause too many changes on each commit
-  "editor.formatOnSave": false
+  "editor.formatOnSave": false,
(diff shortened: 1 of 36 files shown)


dbartol:
The single quotes may all be my fault. Bad habit carried over from PowerShell of all places. My bad.

aeisenberg:
I actually prefer single quotes as well. But, this project was sort of using half and half. I'm just trying to enforce consistency.

jcreedcmu:
If we all prefer single quotes (as I do as well) is there any way to enforce that instead?

aeisenberg:
Let me change this PR to single quotes. I'll wait a little bit in order to get my other changes in first.

PR #442, opened 2020-06-10
Bump version of rush and pnpm

aeisenberg:
Chore: use latest versions of both build tools.

### DIFF
--- a/tools/build-tasks/src/rush.ts
+++ b/tools/build-tasks/src/rush.ts
@@ -26,23 +26,32 @@ export class RushContext {
   private readonly packageStore: string;
 
   constructor(public readonly rushConfig: RushConfiguration) {
-    this.packageStore = path.join(rushConfig.pnpmStoreFolder, '2');
+    this.packageStore = path.normalize(path.join(
+      rushConfig.pnpmOptions.pnpmStorePath,
+      "..",
+      "node_modules",
+      ".pnpm"
+    ));
   }
 
   private async findPackageInRepository(name: string, version: string): Promise<string> {
-    // Packages may be pulled from multiple registries, each of which has its own directory in the
-    // pnpm store. Search for the package name in any of these directories. We use `*.*` to match
-    // the directory name to avoid searching the `local` directory, which does not represent a
-    // package registry.
-    const results = await glob(`*.*/${name}/${version}/package`, {
+    // See https://pnpm.js.org/en/symlinked-node-modules-structure for how packages are laid out
+    // note that this will fail if we ever get into a situation where we have multiple versions
+    // of the same package due to peer dependencies.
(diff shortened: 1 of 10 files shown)


aeisenberg:
Wow. This really opened a can of worms with the configuration. Looks like we cannot update typescript past 3.8.x since in 3.9, a new check was introduced that added some compiler errors. 

Also, bumping the rush version forced us to remove the reference to `pnpmStoreFolder` and reference it a different way, but unfortunately, we are still having some resolution errors.

aeisenberg:
@dbartol If you have a simple solution here for a fix, I'll do it, but otherwise, I'll just abandon this PR.

tools/build-tasks/src/rush.ts
```diff
@@ -26,7 +26,7 @@ export class RushContext {
   private readonly packageStore: string;
 
   constructor(public readonly rushConfig: RushConfiguration) {
-    this.packageStore = path.join(rushConfig.pnpmStoreFolder, '2');
+    this.packageStore = path.join(rushConfig.pnpmOptions.pnpmStorePath, '2');
```
aeisenberg, line 29: @dbartol the problem is here, I believe. `pnpmStorePath` is not correct. I haven't looked too deeply here, but I just think the `packageStore` path is not correct. Is `2` a version number>

dbartol: The existing deployment code depends on the layout of the pnpm store and shrinkwrap file. The root of the problem is that there seems to be no good way to get from the maze of symlinks created by pnpm to the npm-style nested `node_modules` layout expected by the `vsce` tool that packages the extension. We currently try to find the right packages in the pnpm package store and copy them into a `node_modules` tree, allowing duplicates that would normally by symlinked by pnpm. Then we run `npm dedupe` to try to hoist out the duplicates before we run `vsce package`.

There seems to be a new `rush deploy` command in the most recent releases of Rush. This does part of what we need, by copying the package and its dependencies to a deployment directory, skipping any dev dependencies. However, even this deployment directory uses pnpm's symlinking strategy, because to do otherwise would potentially break the semantics in subtle ways.

I wonder if we'd be better off just switching package managers. Rush supports pnpm, npm, and Yarn. When I first compared them, pnpm sounded like the best solution, but if we lose one of its primary benefits when packaging our extension anyway, maybe it's not worth the trouble.

`vsce package` does have a `--yarn` switch now. I wonder if we could just switch to Yarn and get some of this for free.

aeisenberg: `pnpm` is the package manager best supported by rush. But, yeah...I see the benefits of pnpm aren't relevant for packaging the extension. I'll work on this some more and try swapping out for yarn.  Is it too late to move to lerna or yarn workspaces?

aeisenberg:
Putting this on hold. We may work on removing rush from this project completely.

PR #444, opened 2020-06-11
Explicitly use the workspace-configured locale

### LINKED ISSUE #440: Use locale-aware date formatting for date labels
Date labels in the query history not respect the user’s system locale.

aeisenberg:
When the `env.language` value is not passed as the locale, OS-default
locale is used. This change ensures that we use the workspace locale
where we want and explicitly calls out where we should continue to use
the OS-default.

Fixes #440.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -226,7 +226,9 @@ async function convertToQlPath(filePath: string): Promise<string> {
       const fileName = path.basename(filePath);
       const fileNames = await promisify<string, string[]>(fs.readdir)(dir);
       for (const name of fileNames) {
-        // leave the locale argument empty so that the default OS locale is used.
+        // Leave the locale argument empty so that the default OS locale is used.
+        // We do this because this operation works on filesystem entities, which
+        // use the os locale, regardless of the locale of the running VS Code instance.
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/run-queries.ts
```diff
@@ -226,6 +226,7 @@ async function convertToQlPath(filePath: string): Promise<string> {
       const fileName = path.basename(filePath);
       const fileNames = await promisify<string, string[]>(fs.readdir)(dir);
       for (const name of fileNames) {
+        // leave the locale argument empty so that the default OS locale is used.
```
jcreedcmu, line 229: I don't think this is blocking, but I am curious: why is it more appropriate to use OS default locale instead of `env.language` here?

aeisenberg: Because we're looking at the filesystem, the locale of these files are always coming from the os, regardless of the locale of the running vscode instance.

I'll clarify that.

aeisenberg, line 229:
```suggestion
        // Leave the locale argument empty so that the default OS locale is used.
        // We do this because this operation works on filesystem entities, which 
        // use the os locale, regardless of the locale of the running VS Code instance.
```

PR #449, opened 2020-06-11
Force undefined version if version command fails

aeisenberg:
The assumption is that the cli is old or corrupted. In either case we want to upgrade.

### DIFF
--- a/extensions/ql-vscode/src/cli-version.ts
+++ b/extensions/ql-vscode/src/cli-version.ts
@@ -6,12 +6,19 @@ import { Logger } from "./logging";
  * Get the version of a CodeQL CLI.
  */
 export async function getCodeQlCliVersion(codeQlPath: string, logger: Logger): Promise<semver.SemVer | undefined> {
-  const output: string = await runCodeQlCliCommand(
-    codeQlPath,
-    ["version"],
-    ["--format=terse"],
-    "Checking CodeQL version",
-    logger
-  );
-  return semver.parse(output.trim()) || undefined;
+  try {
+    const output: string = await runCodeQlCliCommand(
+      codeQlPath,
+      ["version"],
+      ["--format=terse"],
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/cli-version.ts
```diff
@@ -6,12 +6,18 @@ import { Logger } from "./logging";
...
+      ["version"],
+      ["--format=terse"],
+      "Checking CodeQL version",
+      logger
+    );
+    return semver.parse(output.trim()) || undefined;
+  } catch (e) {
+    // failed to run the version command. This would happen if the cli version is _really_ old, or it is corrupted.
+    // in either we can't determine compatibility.
+    return undefined;
```
henrymercer, line 21: Suggest also logging the exception

extensions/ql-vscode/src/cli-version.ts
```diff
@@ -6,12 +6,18 @@ import { Logger } from "./logging";
...
+      codeQlPath,
+      ["version"],
+      ["--format=terse"],
+      "Checking CodeQL version",
+      logger
+    );
+    return semver.parse(output.trim()) || undefined;
+  } catch (e) {
+    // failed to run the version command. This would happen if the cli version is _really_ old, or it is corrupted.
+    // in either we can't determine compatibility.
```
henrymercer, line 20: Nit:

```suggestion
    // Failed to run the version command. This might happen if the cli version is _really_ old, or it is corrupted.
    // Either way, we can't determine compatibility.
```

extensions/ql-vscode/src/cli-version.ts
aeisenberg: Sure.

aeisenberg:
Thanks!

I don't think this warrants a changelog entry.

PR #452, opened 2020-06-16
Change references to master branch to main branch

jcreedcmu:
I have already pushed `main` upstream pointing at the same commit as
master, so this PR should be safe to merge. To the best of my
knowledge all that's necessary after that is to change the
default branch
in https://github.com/github/vscode-codeql/settings/branches.


jcreedcmu:
Fixes #453.

jcreedcmu:
@shati-patel anywhere else in docs we refer to specific urls starting from https://github.com/github/vscode-codeql/blob/master/ ? If so, could change them to https://github.com/github/vscode-codeql/blob/main/ . No great harm will come from not changing them yet --- not planning to delete the master branch yet, just deprecate.

aeisenberg:
Note that we are also going to need to change the settings and add `main` as the default branch and add protections to it.

jcreedcmu:
Yes, that's what I meant with the link in the PR description

PR #456, opened 2020-06-16
Add Multi-select support to query history view

### LINKED ISSUE #454: Allow multi-select in query history view
Some commands in the query history view make sense to support multiple items:

* codeQLQueryHistory.removeHistoryItem --- we should be able to remove multiple items at once
* codeQLQueryHistory.compareWith --- if exactly two items are selected, the compare view should focus on the two of them (if they are compatible queries)

I would like to ensure that when there is multi-select, only the appropriate commands are active, but due to https://github.com/microsoft/vscode/issues/99767, this feature has not been implemented in vscode. So, we will just have to throw an error if the wrong number of items are selected for a query.

aeisenberg:
This is not quite ideal due to
https://github.com/microsoft/vscode/issues/99767

Allow multiselection in the query-history view. For commands
that shouldn't accept multiple options, show a user message
to that effect.

For remove query, allow multiple removals at once.

For compare query, allow selecting of exactly two queries.
Otherwise, throw an error. Also, verify that the selected queries
are compatible to compare.

Fixes #454 

## Checklist


aeisenberg:
Ping @github/product-docs-dsp! A small, but useful feature.

shati-patel:
> Ping @github/product-docs-dsp! A small, but useful feature.

Noted! Thanks for the ping.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -162,120 +168,284 @@ export class QueryHistoryManager {
...
     }
   }
 
-  async handleItemClicked(queryHistoryItem: CompletedQuery) {
-    this.treeDataProvider.setCurrentItem(queryHistoryItem);
+  async handleItemClicked(
+    singleItem: CompletedQuery,
+    multiSelect: CompletedQuery[]
+  ) {
+    if (!this.assertSingleQuery(multiSelect)) {
```
jcreedcmu, line 367: I'm not sure if this requires any change to your code here, but I observe that the `command` of a `TreeItem` seems to only be invoked when the selection is a single item, so I think this assertion will never fail. Maybe log an additional invariant violation message if it does?

extensions/ql-vscode/src/query-history.ts
```diff
@@ -162,120 +168,284 @@ export class QueryHistoryManager {
...
       await this.invokeCallbackOn(current);
     }
   }
 
-  async handleSetLabel(queryHistoryItem: CompletedQuery) {
+  async handleSetLabel(
+    singleItem: CompletedQuery,
+    multiSelect: CompletedQuery[]
+  ): Promise<void> {
+    if (!this.assertSingleQuery(multiSelect)) {
```
jcreedcmu, line 324: Since labels can be `sprintf`-like patterns, I think we should support setting group of query history items' label at once.

extensions/ql-vscode/src/query-history.ts
aeisenberg: I'm not sure what you mean here? The behaviour is that when there is a single selection, the command is invoked with `singleItem` being the one that was clicked on and `multiSelect` being undefined.

When there are multiple selections, `singleItem` is still the one being clicked on and `multiSelect` contains the list of items selected.

extensions/ql-vscode/src/query-history.ts
aeisenberg: I feel like that's a bit confusing. If someone creates a label without a sprintf pattern, all of the labels will be the same and you can no longer distinguish between them.

I guess, I _could_ special case things where if you multi-select a label change, the second one is suffixed with `-2` and then `-3`, etc.

extensions/ql-vscode/src/query-history.ts
jcreedcmu: Correct. I think (but I'm not 100% confident) the only time vscode will ever invoke this particular event handler, it will be the case that `multiSelect === undefined`.

extensions/ql-vscode/src/query-history.ts
jcreedcmu: We discussed this out of band and I backed off from thinking it's important to have this. I'll make a PR if I think I have a good solution.

PR #457, opened 2020-06-16
Reorganize ts files and group similar files together

aeisenberg:
This is a quick reorg of our files. There has been no other changes here other than moving files around and anything required to get the project compiling again.

We can discuss a better organization as well as whether or not we should do this refactoring at all.


aeisenberg:
I don't think this is worth doing right now.

PR #458, opened 2020-06-17
Remove all code from the master branch

aeisenberg:
Add a deprecation message and instructions on how to proceed.

This PR should obviously be merged into master only, so we need to do some magic to get there (probably just temporarily remove the master branch protection and push directly).

All files have been deleted and the README.md has been replaced.


aeisenberg:
Oh...right and the actions are going to fail since they no longer exist.

jcreedcmu:
Ah, nice, this seems like a good way to communicate the need to migrate to anybody still locally tracking `master`. I will disable branch protection temporarily and manually push this.

PR #461, opened 2020-06-17
Add multi-select support for db view

### LINKED ISSUE #455: Allow multi-select in database view
Similar to #454, but for the databases view. Here, we want to support multi-select for:

* remove database
* upgrade database
* show database directory (possibly)

aeisenberg:
These commands now work on multiple databases:

* remove databases
* open external folder
* upgrade databases## Checklist

Fixes #455.

### DIFF
--- a/extensions/ql-vscode/src/databases-ui.ts
+++ b/extensions/ql-vscode/src/databases-ui.ts
@@ -1,16 +1,35 @@
 import * as path from 'path';
 import { DisposableObject } from '@github/codeql-vscode-utils';
-import { commands, Event, EventEmitter, ExtensionContext, ProviderResult, TreeDataProvider, TreeItem, Uri, window, env } from 'vscode';
+import {
+  commands,
+  Event,
+  EventEmitter,
+  ExtensionContext,
+  ProviderResult,
+  TreeDataProvider,
+  TreeItem,
+  Uri,
+  window,
+  env,
+} from 'vscode';
 import * as fs from 'fs-extra';
 
 import * as cli from './cli';
-import { DatabaseItem, DatabaseManager, getUpgradesDirectories } from './databases';
+import {
+  DatabaseItem,
+  DatabaseManager,
+  getUpgradesDirectories,
+} from './databases';
 import { getOnDiskWorkspaceFolders, showAndLogErrorMessage } from './helpers';
 import { logger } from './logging';
 import { clearCacheInDatabase, UserCancellationException } from './run-queries';
 import * as qsClient from './queryserver-client';
 import { upgradeDatabase } from './upgrades';
-import { importArchiveDatabase, promptImportInternetDatabase, promptImportLgtmDatabase } from './databaseFetcher';
+import {
+  importArchiveDatabase,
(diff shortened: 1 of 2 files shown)


aeisenberg:

Similar to #456, but for the databases view.

shati-patel:
Thanks for the ping! I've added this to our existing docs issue for "multi-selecting things" and we'll check if anything needs updating 😃

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -213,113 +330,174 @@ export class DatabaseUI extends DisposableObject {
...
     }
-  }
-
-  handleChooseDatabaseInternet = async (): Promise<DatabaseItem | undefined> => {
-    return await promptImportInternetDatabase(this.databaseManager, this.storagePath);
-  }
+  };
+
+  handleChooseDatabaseInternet = async (): Promise<
+    DatabaseItem | undefined
```
jcreedcmu, line 336: I've been trying to ignore these reformattings that (to my eye) unnecessarily split lines, but splitting up this type is really weird, given that `Promise<DatabaseItem | undefined>` is kept together on the next function. Is there a setting that's setting a column limit to something significantly less than 80 columns?

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -213,113 +330,174 @@ export class DatabaseUI extends DisposableObject {
...
-      logger.log('Received request to upgrade database, but database contents could not be found.');
-      return;
-    }
-    if (databaseItem.contents.dbSchemeUri === undefined) {
-      logger.log('Received request to upgrade database, but database has no schema.');
-      return;
-    }
+    await this.handleUpgradeDatabase(
+      this.databaseManager.currentDatabaseItem,
+      []
```
jcreedcmu, line 370: For consistency with how multi-query commands were handled, shouldn't this empty list be either `undefined` or a singleton list of `this.databaseManager.currentDatabaseItem`?

jcreedcmu: I guess based on the way it's implemented it doesn't really matter. I'm very much not a fan of this pattern of `(singleThing: T, multipleThings: T[])` where allegedly one or the other can be undefined, and usually the `multipleThings` includes the `singleThing`, but vscode's API imposes it on us at the command callback interface, so I guess it's fine.

extensions/ql-vscode/src/databases-ui.ts
aeisenberg: This is the standard prettier reformatting (recognizing some flags in our settings). And it is aggressive about short lines. I know we're not using prettier in the build. It is currently set to 80 columns. I can make it 100 from now on.

extensions/ql-vscode/src/databases-ui.ts
aeisenberg: I'm disappointed with how under-specified this part of the API is. There's no clear location where it's described how the multiselection passes arguments as far as I can tell. I had to figure this out trial and error.

PR #469, opened 2020-06-22
Use `Open` instead of `Extract` to open zip files

### LINKED ISSUE #468: Not all zipped databases can be imported into the workspace
Due to https://github.com/ZJONSSON/node-unzipper/issues/202, there are some zip files that cannot be extracted. This is due to the way we are extracting files.

aeisenberg:
This allows opening zip files whose local headers are not correct.

Fixes #468.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CodeQL for Visual Studio Code: Changelog
 
+## [Unreleased]
+
+- Fix unzipping of large files.
+
 ## 1.3.0 - 22 June 2020
 
 - Report error when selecting invalid database.
--- a/extensions/ql-vscode/src/databaseFetcher.ts
+++ b/extensions/ql-vscode/src/databaseFetcher.ts
@@ -9,6 +9,7 @@ import {
 } from 'vscode';
 import * as fs from 'fs-extra';
 import * as path from 'path';
+
 import { DatabaseManager, DatabaseItem } from './databases';
 import {
   ProgressCallback,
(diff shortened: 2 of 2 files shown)


aeisenberg:
Note that we are still using extract for the databases downloaded from lgtm or elsewhere on the internet. These downloads seem to be extractable with no problem. And furthermore, the method implemented here requires reading the end of the archive before starting to extract any files. This makes it difficult to stream the database extraction since we need to read the end before we can read the beginning.

The `unzipper.Open.url` method might work, but it requires a dependency on the `request` package, which is deprecated and our own dependency on `node-fetch` is not supported. I'm reluctant to implement this until I understand more. See also https://github.com/ZJONSSON/node-unzipper/issues/204

jcreedcmu:
Yeah, since the issue is that only the data at the very end of the file is trusted, it's hard to imagine an implementation path for remote files that's an improvement over "just download the whole file and then do local extraction via `unzipper.Open.file`" unless we know the HTTP endpoint we're getting it from supports range requests. (I don't know off the top of my head if that's common, or if that's what `unzipper.Open.url` tries to do)

aeisenberg:
I was considering that workflow to first download and then `Open`. `Open.url` does not try to do that. Instead, it does something funky. 

1. Issue a `GET` request for the archive
2. When headers come in, look at the size header
3. Cancel request
4. Issue a second request, but only for the end of the archive that has the directory, using a `range` header
5. Well, I didn't get further than that in my debugging. Not sure if it then issues requests using `range` to request individual files, or it just gets everything.

The problem is that this process is broken because it requires us to provide a `request` library, but it is unclear which libraries are supported.

PR #473, opened 2020-06-23
Ensure query compare order matches expectation

### LINKED ISSUE #470: Query compare order can be reversed in some cases
**Describe the bug**
When selecting two queries to compare, the user expects that the first query selected is the _from_ query and the second one is _to_. However, sometimes these are reversed. The _from_ query is always the one on top in the tree view, and _to_ is below.

**To Reproduce**

1. Execute 2 queries
2. Select the bottom query first
3. Use multiselection to select the top one second
4. Right click and _compare with..._

**Expected behavior**
The bottom query should be on the right and top on the left, but the actual behavior is that the queries are reversed.

aeisenberg:
A user typically expects that the first selection would be
the query that they are comparing _from_ and the second query
is being compared _to_.

This commit ensures that something like this expectation will
always hold.

So, when there are two queries selected, the first one selected
will always be _from_ and appear on the left side of the compare
view. The one selected later will be _to_ and appear on the right.

There is a corner case when there are 3 or more selected queries
and a user *unselects* a query. We do not track the selection
order of the remaining two queries.

Fixes #470.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Fix unzipping of large files.
+- Ensure compare order is consistent when selecting two queries to compare. The first query selected is always the _from_ query and the query selected later is always the _to_ query.
 
 ## 1.3.0 - 22 June 2020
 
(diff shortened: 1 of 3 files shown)


aeisenberg:
Ping @github/product-docs-dsp. I don't think it is worth updating the docs for this, but letting you know just in case.

shati-patel:
> Ping @github/product-docs-dsp. I don't think it is worth updating the docs for this, but letting you know just in case.

You're right this doesn't affect the docs, but thank you for letting us know 😃

PR #475, opened 2020-06-24
Avoid using a synchronous file system command

aeisenberg:
Small refactoring to avoid a synchronous call.


jcreedcmu:
I think I'd prefer ditching `tmp` for https://github.com/benjamingr/tmp-promise if we want to not have a sync method here.

aeisenberg:
That's fine.

PR #478, opened 2020-06-25
Remove unused code path

aeisenberg:
`resolveRawDataset` can never be called.

Had a discussion with @jcreedcmu and it seems like we don't need to use `resolveRawDataset`.


aeisenberg:
None of the checklist is relevant. This is internal code cleanup.

PR #480, opened 2020-06-26
Ensure source folders are zipped

### LINKED ISSUE #479: Auto-zip the src location when installing databases from LGTM
Unzipped source locations are deprecated. The codeql cli produces source locations in a `src.zip` file. The deprecated odasa tool does not zip and creates an `src` folder. The extension has a few code paths that only work in the case of a zipped source location.

When adding a database from a zip file (either locally, LGTM, or anywhere on the internet), the extension should automatically zip the source location so that the database is more compatible with the extension.

Note that if a database is added from a local directory, the extension should _not_ perform the zipping. This is because this action is destructive, databases are meant to be read-only, and we don't know what other purposes the database is being used for.

aeisenberg:
Zips source folders of databases when they are added. Only if the databases are fully controlled by VS Code.

Fixes #479

Also, avoids the error specified in https://github.com/github/codeql-coreql-team/issues/452

## Checklist

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -92,6 +92,7 @@ dependencies:
   vscode-test-adapter-util: 0.7.0
   webpack: 4.42.0_webpack@4.42.0
   webpack-cli: 3.3.11_webpack@4.42.0
+  zip-a-folder: 0.0.12
 lockfileVersion: 5.1
 packages:
   /@babel/code-frame/7.8.3:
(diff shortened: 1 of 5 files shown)


aeisenberg:
Ping! @github/product-docs-dsp

This is an internal change that most users won't notice directly. This change ensures that all newly added databases from LGTM will work with Go To Definition/Reference, as well as any other contextual queries we add in the future. 

Note that this will only update newly added databases. Existing databases will still have their sources unzipped.

PR #481, opened 2020-06-30
Allow pagination for interpreted results

jcreedcmu:
Enable pagination for interpreted as well as raw results.

This is still under feature flag, so no user-facing changes quite yet, but the flag can possibly be enabled for everyone once this lands. There will be a number of code cleanups available at that point.

### DIFF
--- a/extensions/ql-vscode/src/interface.ts
+++ b/extensions/ql-vscode/src/interface.ts
@@ -407,7 +407,7 @@ export class InterfaceManager extends DisposableObject {
       database: this._displayedQuery.database,
       metadata: this._displayedQuery.query.metadata,
       pageNumber,
-      totalPages: Math.ceil(this._interpretation.sarif.runs[0].results.length / INTERPRETED_RESULTS_PAGE_SIZE),
+      numPages: Math.ceil(this._interpretation.sarif.runs[0].results.length / INTERPRETED_RESULTS_PAGE_SIZE),
     });
   }
 
--- a/extensions/ql-vscode/src/interface.ts
+++ b/extensions/ql-vscode/src/interface.ts
@@ -392,6 +392,9 @@ export class InterfaceManager extends DisposableObject {
   public async showPageOfInterpretedResults(
     pageNumber: number
   ): Promise<void> {
+    if (this._displayedQuery === undefined) {
+      throw new Error(`Trying to show interpreted results but displayed query was undefined`);
+    }
     if (this._interpretation === undefined) {
       throw new Error(`Trying to show interpreted results but interpretation was undefined`);
     }
@@ -401,6 +404,8 @@ export class InterfaceManager extends DisposableObject {
     await this.postMessage({
       t: 'showInterpretedPage',
       interpretation: this.getPageOfInterpretedResults(pageNumber),
+      database: this._displayedQuery.database,
+      metadata: this._displayedQuery.query.metadata,
       pageNumber,
       totalPages: Math.ceil(this._interpretation.sarif.runs[0].results.length / INTERPRETED_RESULTS_PAGE_SIZE),
     });
--- a/extensions/ql-vscode/src/interface.ts
+++ b/extensions/ql-vscode/src/interface.ts
@@ -362,7 +362,7 @@ export class InterfaceManager extends DisposableObject {
           t: 'ExtensionParsed',
           pageNumber: 0,
           numPages: numPagesOfResultSet(resultSet),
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/interface.ts
```diff
@@ -357,7 +357,7 @@ export class InterfaceManager extends DisposableObject {
           t: 'ExtensionParsed',
           pageNumber: 0,
           numPages: numPagesOfResultSet(resultSet),
-          resultSet,
+          resultSet: { t: 'RawResultSet', ...resultSet },
```
aeisenberg, line 360: Has the type `RawResultSet` changed? In the current main branch, it looks like there is no `t` property.

extensions/ql-vscode/src/view/result-tables.tsx
```diff
@@ -70,13 +68,13 @@ function renderResultCountString(resultSet: ResultSet): JSX.Element {
 export class ResultTables
   extends React.Component<ResultTablesProps, ResultTablesState> {
 
-  private getResultSets(): ResultSet[] {
+  private static _getResultSetsOfProps(props: ResultTablesProps): ResultSet[] {
```
aeisenberg, line 71: Why convert to static?

extensions/ql-vscode/src/view/result-tables.tsx
```diff
@@ -85,12 +83,16 @@ export class ResultTables
...
         schema: { name: ALERTS_TABLE_NAME, version: 0, columns: [], tupleCount: 1 },
         name: ALERTS_TABLE_NAME,
-        ...this.props.interpretation,
+        ...props.interpretation,
       });
     }
     return resultSets;
   }
 
+  private getResultSets(): ResultSet[] {
```
aeisenberg, line 92: Looks like this isn't used.

Wait...it is. Though, not sure what the difference is between this function and the static `_getResultSetsOfProps`.

extensions/ql-vscode/src/view/result-tables.tsx
```diff
@@ -112,20 +114,22 @@ export class ResultTables
 
   /**
    * Holds if we actually should show pagination interface right now. This is
-   * still false for the time being when we're viewing alerts.
+   * true as long as the result sets we're given are marked to permit it.
```
aeisenberg, line 117: should probably just get rid of this function since it duplicates `paginationAllowed` without adding anything new. And `paginationAllowed` will be removed when we remove the pagination flag, right?

aeisenberg:
It looks like this adds pagination for alerts table, but not paths tables. Is that part coming later?

extensions/ql-vscode/src/view/result-tables.tsx
jcreedcmu: Sure, could just inline that now.

jcreedcmu:
> It looks like this adds pagination for alerts table, but not paths tables. Is that part coming later?

No, it works for me on path tables when I test it? This may be a worthwhile cleanup in how things are named if that's confusing.

extensions/ql-vscode/src/view/result-tables.tsx
jcreedcmu: So that it can be called from the static lifecycle method `getDerivedStateFromProps`.

extensions/ql-vscode/src/interface.ts
jcreedcmu: To accomodate sarif results, I changed the type of the `.resultSet` field of `ExtensionParsedResultSets` from being always only a `RawResultSet` to being a `ResultSet` (which is either a `RawResultSet` or a `SarifResultSet`, tagged appropriately)

extensions/ql-vscode/src/view/result-tables.tsx
jcreedcmu: I just factored out the body of that method into a static method that takes a props argument, so that I could call it from `getDerivedStateFromProps`.

aeisenberg:
OK...tried this out on a proper path query. Looks great. A couple of things, though:

1. We should have some better styling for the pagination section (I said that before, and this can still happen later).
2. it's not possible to edit the pagination text box. I haven't looked at the reason why, but I suspect that this is a react problem. Should be easy to fix.

aeisenberg:
I think for 2, the problem has to do wih `getDerivedStateFromProps` returning the old value of the selected page.

PR #492, opened 2020-07-09
Add the AST Viewer

aeisenberg:
This commit adds the AST Viewer for viewing the QL AST of a file in a
database.

The different components are as follows:

1. There is a new view `codeQLAstViewer`, which displays the AST
2. This view is backed by the `AstViewerDataProvider` and `AstViewer` classes in astView.ts
3. To generate an AST, we use contextual queries, similar to how Find references/declarations are implemented. In particular, in `definitions.ts` there is `TemplatePrintAstProvider` which provides an AST for a given source buffer.
  - Similar to the other queries, we first determine which database the buffer belongs to.
  - Based on that, we generate a synthetic qlpack and run the templatized `printAst.ql` query
  - We plug in the archive-relative path name of the source file.
  - After the query is run, we wrap the results in an `AstBuilder` instance.
  - When requested, the `AstBuilder` will generate the full AST of the file from the BQRS results.
  - The AST roots (all top-level elements, functions, variable declarations, etc, are roots) are passed to the `AstViewer` instance, which handles the display lifecycle and other VS Code-specific functions.

There are a few unrelated pieces here, which can be pulled out to another PR if required:

- The `codeQLQueryHistory` view now has a _welcome_ message to make it more obvious to users how to start.
- `definitions.ts` is moved to the `contextual` subfolder.
- `fileRangeFromURI` is extracted from `definitions.ts` to its own file so it can be reused.

Also, note that this relies on https://github.com/github/codeql/pull/3931 for the C/C++ query to be available in the QL sources. Other languages will need similar queries.

## Checklist


aeisenberg:
Note: Now rebaased on #494. Let's get that one in first.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## [Unreleased]
 
+- Add the AST Viewer to inspect the QL AST of a source file in a database. Currently, only available for C/C++ sources.
 - Fix error with choosing qlpack search path.
+- Add the AST Viewer to inspect the QL AST of a source file in a database. Currently, only available for C/C++ sources.
```
Marcono1234, line 7: This line exists twice

aeisenberg: Thanks for pointing that out. Must have been a bad merge.

extensions/ql-vscode/src/contextual/astBuilder.ts
```diff
@@ -0,0 +1,121 @@
...
+          if (child) {
+            child.parent = item;
+            item.children.push(child);
+          }
+        });
+      }
+    });
+
+    // find the roots and add the order
+    for(const [, item] of idToItem) {
```
jcreedcmu, line 105: I didn't know you didn't have to write an explicit `undefined` to avoid binding a tuple deconstructor field! Nice.

extensions/ql-vscode/src/contextual/definitions.ts
```diff
@@ -25,27 +29,46 @@ const SELECT_QUERY_NAME = '#select';
...
   DefinitionQuery = 'DefinitionQuery',
   ReferenceQuery = 'ReferenceQuery',
+  PrintAstQuery = 'PrintAstQuery',
 }
 
 function tagOfKeyType(keyType: KeyType): string {
   switch (keyType) {
-    case KeyType.DefinitionQuery: return 'ide-contextual-queries/local-definitions';
-    case KeyType.ReferenceQuery: return 'ide-contextual-queries/local-references';
+    case KeyType.DefinitionQuery:
```
jcreedcmu, line 37: It seems like a bunch of these definitions should be hoisted out of `definitions.ts` into some more general file in `contextual/` but that needn't happen in this PR.

extensions/ql-vscode/src/cli.ts
```diff
@@ -499,7 +499,7 @@ export class CodeQLCliServer implements Disposable {
   */
   async bqrsDecode(bqrsPath: string, resultSet: string, pageSize?: number, offset?: number): Promise<DecodedBqrsChunk> {
     const subcommandArgs = [
-      '--entities=url,string',
+      '--entities=id,url,string',
```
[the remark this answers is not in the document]
aeisenberg: I originally implemented this as optionally including the `id` field. I could add it back.

extensions/ql-vscode/src/contextual/astBuilder.ts
```diff
@@ -0,0 +1,121 @@
...
+import { CodeQLCliServer } from '../cli';
+import { DecodedBqrsChunk, BqrsId, EntityValue } from '../bqrs-cli-types';
+import { DatabaseItem } from '../databases';
+import { AstItem, RootAstItem } from '../astViewer';
+
+/**
+ * A class that wraps a tree of QL results from a query that
+ * has an @kind of graph
+ */
+// RENAME to ASTParser / ASTCreator
```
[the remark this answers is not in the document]
aeisenberg: Oops. That was a comment to myself. I must have missed that one. Will do it now.

extensions/ql-vscode/src/contextual/astBuilder.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const idToItem = new Map<BqrsId, AstItem | RootAstItem>();
+    const parentToChildren = new Map<BqrsId, BqrsId[]>();
+    const childToParent = new Map<BqrsId, BqrsId>();
+    const astOrder = new Map<BqrsId, number>();
+    const roots = [];
+
+    // Build up the parent-child relationships
+    edgeTuples.tuples.forEach(tuple => {
+      const from = tuple[0] as EntityValue;
```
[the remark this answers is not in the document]
aeisenberg: Fair point. I will clarify.

extensions/ql-vscode/src/contextual/definitions.ts
aeisenberg: Sure. I agree that it's getting a bit messy. I'll work on that next.

extensions/ql-vscode/src/cli.ts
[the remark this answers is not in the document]
jcreedcmu: Do whatever you think is reasonable; I can see the side of the argument for it being awkward to make it optional too, since that makes the type vary.

extensions/ql-vscode/src/contextual/definitions.ts
```diff
@@ -145,8 +261,8 @@ async function getLinksFromResults(results: QueryWithResults, cli: CodeQLCliServ
       const dest = tuple[1] as EntityValue;
       const srcFile = src.url && fileRangeFromURI(src.url, db);
       const destFile = dest.url && fileRangeFromURI(dest.url, db);
-      if (srcFile && destFile && filter(srcFile.file.toString(), destFile.file.toString())) {
-        localLinks.push({ targetRange: destFile.range, targetUri: destFile.file, originSelectionRange: srcFile.range, originUri: srcFile.file });
+      if (srcFile && destFile && filter(srcFile.uri.toString(), destFile.uri.toString())) {
```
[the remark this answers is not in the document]
aeisenberg: Ahhh...I had forgotten that I did this. We are comparing 2 strings to see if they are a match and if so, then we mark the tuple as a match. Earlier on, we call `getDefinitions` and `getReferences` with `document.uri.toString()`. So, I am not sure how this worked in the past. I'll go try this out now and see.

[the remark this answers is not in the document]
aeisenberg: Right. I extracted `fileRangeFromURI` to its own file and I made it return a `vscode.Location` object. It's the same as before, except it is returning a `vscode` api type. `file` and `uri` are both `vscode.Uri` instances.

extensions/ql-vscode/src/contextual/astBuilder.ts
```diff
@@ -48,61 +48,75 @@ export default class AstBuilder {
...
+          if (!children) {
+            parentToChildren.set(sourceId, children = []);
+          }
+          children.push(targetId);
+          break;
         }
-        children.push(toId);
+
+        default:
+        // ignore other tupleTypes since they are not needed by the ast viewer
```
jcreedcmu, line 71: Ah, ok, great --- I didn't know that other values of `tupleType` existed, and it's great to clarify here that we're explicitly ignoring them.

aeisenberg: Thanks for suggesting that change. I think it makes the code clearer.

Issue #493, opened 2020-07-09
Integration tests fail with vscode 1.47.0


jcreedcmu:
We're newly seeing
```
 1 failing

  1) interface-utils
       webview uri conversion
         should correctly round trip from filesystem to webview and back:

      AssertionError: expected 'file:///file///tmp/uri_test__23687aaaBKTl7W50L.bqrs' to equal 'file:///tmp/uri_test__23687aaaBKTl7W50L.bqrs'
      + expected - actual

      -file:///file///tmp/uri_test__23687aaaBKTl7W50L.bqrs
      +file:///tmp/uri_test__23687aaaBKTl7W50L.bqrs
```
when running integration tests against vscode 1.47.0. The behavior of `webview.asWebviewUri` may have changed.

PR #494, opened 2020-07-09
Remove failing integration test

jcreedcmu:
This works around but does not fundamentally address https://github.com/github/vscode-codeql/issues/493.

### DIFF
--- a/extensions/ql-vscode/src/interface-utils.ts
+++ b/extensions/ql-vscode/src/interface-utils.ts
@@ -50,14 +50,6 @@ export function fileUriToWebviewUri(
   return panel.webview.asWebviewUri(fileUriOnDisk).toString();
 }
 
-/** Converts a URI string received from a webview into a local filesystem URI for the same resource. */
-export function webviewUriToFileUri(webviewUri: string): Uri {
-  // Webview URIs used the vscode-resource scheme. The filesystem path of the resource can be obtained from the path component of the webview URI.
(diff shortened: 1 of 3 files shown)


aeisenberg:
Yep. As soon as I updated my VS Code, I'm getting the same problem locally.

adityasharad:
Can we test against both a stable version and the latest? It looks like useful information that the tests started failing when VS Code made a breaking change, and I don't want to lose the ability to discover such problems early.

aeisenberg:
I think before we merge this, we need to figure out where the problem came from. I just looked at the vs code sources and neither [`uri.ts`](https://github.com/microsoft/vscode/blob/master/src/vs/base/common/uri.ts) nor https://github.com/microsoft/vscode/blob/master/src/vs/workbench/api/common/shared/webview.ts have any relevant changes. There could be other related changes that I haven't seen, though.

jcreedcmu:
I think I've determined that the underlying problem is that we *were* dangerously relying on the underlying representation of vscode resource URIs, (see https://github.com/microsoft/vscode/pull/97777 for some breadcrumbs into the process of that change) and that underlying representation changed out from under us.

However, thankfully, we *are* no longer relying on that representation --- it was only used to get back the file uri for purposes of `fetch`ing directly from the webview. Because of progress towards cli result parsing, this is dead code and that failing test is no longer relevant.

PR #496, opened 2020-07-13
Reduce dependencies on internal modules

jcreedcmu:
Moves what currently remains of the extension's dependencies on `semmle-bqrs`, `semmle-io`, `semmle-io-node` into some smaller files local to the extension project.


extensions/ql-vscode/src/bqrs-types.ts
```diff
@@ -0,0 +1,102 @@
+/**
```
aeisenberg, line 1: There is some duplicates between this file and `bqrs-util-types.ts`. This doesn't need to be addressed now, but we should add a TODO or create an issue for that.

extensions/ql-vscode/src/compare/view/tsconfig.json
```diff
@@ -1,24 +0,0 @@
-{
```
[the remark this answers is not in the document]
aeisenberg: Nevermind...I see you fixed this in a later commit.

aeisenberg:
I think this looks great. What still needs to be done before we can remove the extra packages? 

And could you make a note somewhere about consolidating `bqrs-util-types.ts` and `bqrs-types.ts`?

extensions/ql-vscode/src/bqrs-types.ts
jcreedcmu: It's worse than that --- the overlap between the two files is not exact duplication, but a bunch of pairs of types that sometimes share the same name and purpose, and which aren't exactly the same. This is exactly the migration work that remains to be done, but which can be done independently of removing the other build infrastructure.

I added the word `TODO` to the comment at the top of `bqrs-types.ts`, at least, if that helps :)

jcreedcmu:
(at least I assume that where you say `bqrs-util-types.ts` you mean `bqrs-cli-types.ts`)

Issue #498, opened 2020-07-14
Simplify Build Process


jcreedcmu:
Following https://github.com/github/codeql-coreql-team/issues/194, we will not need many of the separate modules in the `lib` directory. @aeisenberg and I have experienced some friction using `rush` to manage dependencies and builds of separate modules relative to just `gulp`-based builds of the extension itself. @dbartol says that he still does find being able to build and/or update dependencies with a single command useful. We should try to find a solution that makes everybody happy.

dbartol:
I'd like to understand the sources of the friction you've encountered here. In my work on the extension, I pretty much just use `rush update`, `rush build`, `rush format`, and occasionally `rush add` to add a new dependency. When I do that, the build just works, the dependency updating just works, and formatting of the all of source files just works. Is the friction that one or more of those commands doesn't really just work? Or are there other operations that Rush is making more difficult?

There is a place in the build process where the way we use Rush _is_ overcomplicating things. When setting it up, I chose pnpm as the underlying package manager, based on its ability to solve some of the uglier consequences of npm, like "phantom dependencies" and "doppelgangers". What I did not realize at the time was that when we actually package the extension into a .vsix, the `vsce` tool pretty much expects the `node_modules` layout to conform to the typical npm layout, forcing us to have a "deploy" step in the build that more or less manually copies the dependencies into the form that npm would have in the first place. I believe that if we just used npm (or maybe Yarn) as the underlying package manager, we could avoid this step, but I haven't tried it yet.

To summarize a much longer discussion I had in Slack with @jcreedcmu, here's my thought process for why the system we have today is our best option:

**Do we really need multiple packages?**
Yes, but not as many as we have. We need one package for the extension, and we should have one package for the API that the extension exposes to other extensions, to enable other extensions to run queries (including context-sensitive ones) and help examine results. The API package would be small, consisting mostly of TypeScript declarations for the interfaces that would actually be implemented by the extension package itself. For an example of how other extensions do this, see the `vscode-test-explorer` extension that our extension depends on at runtime, and its corresponding `vscode-test-adapter-api` package that our extension depends on at build time for the API declaration. 
I also believe that we should keep `semmle-vscode-utils` separate to make development of internal extensions a bit easier, but that's not as compelling of an argument on its own.

**How should we deal with having multiple packages?**
[...]

PR #503, opened 2020-07-16
Try moving build to just gulp

jcreedcmu:
This PR is a proposal for how to fulfill https://github.com/github/vscode-codeql/issues/498. All build configuration is moved into `extensions/ql-vscode` and now only one buildable module exists, so we don't use rush anymore. The Actions workflow files call gulp directly. Building the package with `vsce` seems to remove `devDependencies` just fine by itself, so we no longer have to do any traversal of our dependencies, just copy `node_modules` to the `dist` directory and let `vsce` do its thing. 

In fact, the only reason we have to create a distribution directory and copy anything at all is because we fiddle with the version number in `package.json`. But this seems fine --- it lets us have exact control over which files go in the package. But maybe `vsce` pays attention to the `"files"` section in `package.json`, I haven't checked that.


jcreedcmu:
(for a sense of scale, this diff is +196 loc, -3116 loc excluding the shrinkwrap/lock files)

jcreedcmu:
absolutely; that seems like the immediate next step. I only postponed it because this is plenty big and it seemed less deeply consequential (although it requires changing some relative paths in a few files iirc)

.vscode/tasks.json
```diff
@@ -10,7 +10,10 @@
         "kind": "build",
         "isDefault": true
       },
-      "command": "node common/scripts/install-run-rush.js build --verbose",
+      "command": "npx gulp buildWithoutPackage --verbose",
```
aeisenberg, line 13: I dont think we need `npx` here. We can assume users have already run `npm install` before running in vs code. So, just do the same, but remove `npx`. You may need to add a full path here, so `${workspaceRoot}/extensions/ql-vscode/node_modules/.bin/gulp`.

.github/workflows/main.yml
```diff
@@ -19,11 +19,15 @@ jobs:
...
+        run: |
+          cd extensions/ql-vscode
+          npm install
         shell: bash
 
       - name: Build
-        run: node common/scripts/install-run-rush.js build
+        run: |
+          cd extensions/ql-vscode
+          gulp
```
aeisenberg, line 30: Can we change this to `npm run build`? This will ensure `$PATH` is correct.

.github/workflows/main.yml
```diff
@@ -57,11 +61,15 @@ jobs:
...
```
aeisenberg, line 72: Here, too.

.github/workflows/release.yml
```diff
@@ -34,11 +34,15 @@ jobs:
...
+        run: |
+          cd extensions/ql-vscode
+          npm install
         shell: bash
 
       - name: Build
-        run: node common/scripts/install-run-rush.js build --release
+        run: |
+          cd extensions/ql-vscode
+          gulp --release
```
aeisenberg, line 45: Can we add a `npm run release:build` script?

aeisenberg: Or, `npm run build -- --release` should work.

extensions/ql-vscode/gulpfile.ts/package.ts
```diff
@@ -9,7 +9,7 @@ export async function packageExtension(): Promise<void> {
     'package',
     '--out', path.resolve(deployedPackage.distPath, '..', `${deployedPackage.name}-${deployedPackage.version}.vsix`)
   ];
-  const proc = child_process.spawn('vsce', args, {
+  const proc = childProcess.spawn('npx', ['vsce'].concat(args), {
```
aeisenberg, line 12: I'd like to avoid `npx` and use the package-installed version of `vsce`.

extensions/ql-vscode/package.json
```diff
@@ -586,92 +586,92 @@
     "preintegration": "rm -rf ./out/vscode-tests && gulp",
     "integration": "node ./out/vscode-tests/run-integration-tests.js",
     "update-vscode": "node ./node_modules/vscode/bin/install",
-    "postinstall": "npm rebuild && node ./node_modules/vscode/bin/install",
     "format": "tsfmt -r && eslint src test --ext .ts,.tsx --fix",
     "lint": "eslint src test --ext .ts,.tsx --max-warnings=0",
     "format-staged": "lint-staged"
   },
   "dependencies": {
+    "@types/semver": "~7.2.0",
```
aeisenberg, line 594: Can you move this to devDependencies?

extensions/ql-vscode/tsconfig.json
```diff
@@ -1,3 +1,32 @@
 {
-  "extends": "./node_modules/typescript-config/extension.tsconfig.json"
+  "$schema": "http://json.schemastore.org/tsconfig",
```
aeisenberg, line 2: Is this line important?

.vscode/tasks.json
jcreedcmu: I sort of like `npx` as a terse idiom here, but I don't mind being more explicit if you think it's better to avoid the magic of `npx` auto-downloading packages. I'm pretty sure we already need to be in the correct `cwd` to have gulp work, so a path relative to `extensions/ql-vscode` should work.

.github/workflows/main.yml
jcreedcmu: Sure, and I agree with similar comments below that confining any call to gulp into the `package.json` scripts is generally a better approach.

extensions/ql-vscode/gulpfile.ts/package.ts
jcreedcmu: Agree as above.

extensions/ql-vscode/package.json
jcreedcmu: Oops, good catch.

extensions/ql-vscode/tsconfig.json
jcreedcmu: I don't presently know of anything that requires it but it occurred in other tsconfig files in the previous state of this repo, and it seemed like fine documentation. I have no overwhelmingly strong opinion about including or eliminating it.

.vscode/tasks.json
aeisenberg: My concern with npx is with versioning. I've had situations before where people ran into different behaviour because `npx` had installed different versions. If we are running from the `node_modules` directory, this won't happen.

extensions/ql-vscode/tsconfig.json
aeisenberg: I'm fine with keeping it. I've just never seen it explicitly before. I think vscode adds it automatically, but other tools may not.

aeisenberg:
NIce.

jcreedcmu:
Still need to fix `CONTRIBUTING.md` now.

Issue #510, opened 2020-07-23
Database downloaded message shown when "Download Database" from URL action cancelled


ceh:
**Describe the bug**

"Download Database" from URL action shows a message that a database has been downloaded even when the action is cancelled.

![Screenshot from 2020-07-23 06-24-05](https://user-images.githubusercontent.com/248952/88252108-507d0300-ccad-11ea-85e5-db087864af97.png)

PR #511, opened 2020-07-23
Suppress database downloaded message when action canceled

ceh:
Fixes https://github.com/github/vscode-codeql/issues/510

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databaseFetcher.ts
+++ b/extensions/ql-vscode/src/databaseFetcher.ts
@@ -54,12 +54,10 @@ export async function promptImportInternetDatabase(
             progress
           ))
       );
-      commands.executeCommand('codeQLDatabases.focus');
-    }
-    if (item) {
-      showAndLogInformationMessage(
-        'Database downloaded and imported successfully.'
-      );
+      if (item) {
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -56,9 +56,11 @@ export async function promptImportInternetDatabase(
       );
       commands.executeCommand('codeQLDatabases.focus');
```
jcreedcmu, line 57: Thanks for the PR!

I notice this line right above that focuses the databases panel is also unconditionally being called regardless of whether we actually have a non-`undefined` value of `item` as well. Could you move the `showAndLogInformationMessage` up here and include both it and the `codeQLDatabases.focus` in the conditional?

ceh: :+1: Fixed. Please take another look.

Looks like there's opportunity for a refactor since there's some duplicated code, but I think we can leave that for another PR.

jcreedcmu:
Yeah, that's great! The fact that focusing was happening even on cancellation in those other functions escaped my attention, good catch.

PR #513, opened 2020-07-23
Fix broken launch config

jcreedcmu:
We need to provide the `--extensionDevelopmentPath` flag in these
launch configurations.

It appears to be unnecessary to include
`${workspaceRoot}/extensions/ql-vscode/out/vscode-tests/**/*.js`
in addition to the strictly more general pattern
${workspaceRoot}/extensions/ql-vscode/out/**/*.js

An unfortunate UI fact appears to be that the log of the gulp build is
focused whenever these tasks are run, even though the log you actually
care about seeing is in the `Debug Console` section. Not sure how to
fix that.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -50,14 +50,13 @@
       "request": "launch",
       "runtimeExecutable": "${execPath}",
       "args": [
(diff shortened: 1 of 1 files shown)


aeisenberg:
That last part has always been true. Thanks for catching this.

PR #516, opened 2020-07-23
Refactor contextual queries

aeisenberg:
Break the `definitions.ts` file into logically contained smaller files. And add unit tests for `resolveQueries`.


aeisenberg:
Note that this PR should be targeted on #514.

PR #518, opened 2020-07-24
Filters qltest-discovery

### LINKED ISSUE #517: Duplicated tests in test explorer view
**Describe the bug**

When opening the test explorer in a mult-root workspace, each root contains a copy of all tests available in the workspace.

**To Reproduce**

1. Open vscode starter workspace
1. Open test explorer view
1. Refresh
1. Expand any two root folders

Notice that the two roots have duplicated sub-folders.

_Extension_Development_Host__-_LocalVariableHidesGlobalVariable_qlref_—_vscode-codeql-starter__Workspace_

**Expected behavior**

Sub-folders should not be duplicated across workspace roots.

**Additional context**

There might be a way to think of this as a feature, not a bug. It might be the case that tests run in the context of different workspace roots will behave differently, so _maybe_ we want to allow this duplication.

aeisenberg:
qlpack tests that are not contained within the current workspace folder
will be filtered from the test runner view.

This also fixes a test that should have been failing but wasn't.

Fixes #517.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Add the AST Viewer to inspect the QL AST of a source file in a database. Currently, only available for C/C++ sources.
 - Fix pagination when there are no results.
 - Suppress database downloaded from URL message when action canceled.
+- Fix QL test discovery to avoid showing duplicate tests in the test explorer.
 
 ## 1.3.1 - 7 July 2020
 
--- a/extensions/ql-vscode/src/qlpack-discovery.ts
+++ b/extensions/ql-vscode/src/qlpack-discovery.ts
@@ -20,7 +20,7 @@ export class QLPackDiscovery extends Discovery<QlpacksInfo> {
     private readonly workspaceFolder: WorkspaceFolder,
     private readonly cliServer: CodeQLCliServer
   ) {
-    super();
+    super('QL Pack Discovery');
 
     // Watch for any changes to `qlpack.yml` files in this workspace folder.
     // TODO: The CLI server should tell us what paths to watch for.
--- a/extensions/ql-vscode/src/qlpack-discovery.ts
+++ b/extensions/ql-vscode/src/qlpack-discovery.ts
@@ -16,9 +16,10 @@ export class QLPackDiscovery extends Discovery<QlpacksInfo> {
   private readonly watcher = this.push(new MultiFileSystemWatcher());
   private _qlPacks: readonly QLPack[] = [];
 
-  constructor(private readonly workspaceFolder: WorkspaceFolder,
-    private readonly cliServer: CodeQLCliServer) {
-
+  constructor(
+    private readonly workspaceFolder: WorkspaceFolder,
+    private readonly cliServer: CodeQLCliServer
+  ) {
     super();
 
     // Watch for any changes to `qlpack.yml` files in this workspace folder.
@@ -26,8 +27,6 @@ export class QLPackDiscovery extends Discovery<QlpacksInfo> {
     this.watcher.addWatch(new RelativePattern(this.workspaceFolder, '**/qlpack.yml'));
     this.watcher.addWatch(new RelativePattern(this.workspaceFolder, '**/.codeqlmanifest.json'));
     this.push(this.watcher.onDidChange(this.handleQLPackFileChanged, this));
-
-    this.refresh();
   }
 
   public get onDidChangeQLPacks(): Event<void> { return this._onDidChangeQLPacks.event; }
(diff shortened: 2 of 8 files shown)


extensions/ql-vscode/src/qltest-discovery.ts
```diff
@@ -177,6 +174,15 @@ export class QLTestDiscovery extends Discovery<QLTestDiscoveryResults> {
     this._onDidChangeTests.fire();
   }
 
+  /**
+   * Only include qlpacks suffixed with '-test' that are contained
+   * within the provided workspace folder.
+   */
+  private isRelevantQlPack(qlPack: QLPack): boolean {
+    return qlPack.name.endsWith('-tests')
+      && qlPack.uri.fsPath.startsWith(this.workspaceFolder.uri.fsPath);
```
adityasharad, line 183: Just making sure: Are the URI paths normalised the same way / always on the same filesystem? Just want to be sure there's no weirdness with the pack URI and the folder URI not matching up, e.g. on Windows.

extensions/ql-vscode/src/qltest-discovery.ts
```diff
@@ -177,6 +174,15 @@ export class QLTestDiscovery extends Discovery<QLTestDiscoveryResults> {
     this._onDidChangeTests.fire();
   }
 
+  /**
+   * Only include qlpacks suffixed with '-test' that are contained
```
adityasharad, line 178:
```suggestion
   * Only include qlpacks suffixed with '-tests' that are contained
```

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Add the AST Viewer to inspect the QL AST of a source file in a database. Currently, only available for C/C++ sources.
 - Fix pagination when there are no results.
 - Suppress database downloaded from URL message when action canceled.
+- Fix qltest discovery to avoid showing duplicate tests in the test explorer.
```
adityasharad, line 10:
```suggestion
- Fix QL test discovery to avoid showing duplicate tests in the test explorer.
```

extensions/ql-vscode/src/qltest-discovery.ts
aeisenberg: The Uri `fsPath` property will normalize the path according to the OS path separator. This should not be a problem. See https://code.visualstudio.com/api/references/vscode-api#Uri

dbartol:
```suggestion
      && (workspace.getWorkspaceFolder(qlpack.uri) === this.workspaceFolder);
```
Even easier and cleaner.

aeisenberg: Unfortunately, that won't work because there's no guarantee that two workspace folders are the same object. Though, something like this should work and is almost as clean.

    

```suggestion
      && workspace.getWorkspaceFolder(qlPack.uri)?.index === this.workspaceFolder.index;
```

aeisenberg: It's a better solution, but makes the tests slightly uglier.

jcreedcmu: I think comparing `.index` should be correct, but I thought it was the case that vscode guaranteed that a workspace folder was uniquely characterized by its url --- so in that case wouldn't it also suffice to compare equality of
`workspace.getWorkspaceFolder(qlpack.uri).uri.toString()`
and
`this.workspaceFolder.uri.toString()`
?

If that makes testing easier since you don't need to control index order, I suggest that, otherwise, `.index` is fine.

extensions/ql-vscode/src/qlpack-discovery.ts
```diff
@@ -16,18 +16,17 @@ export class QLPackDiscovery extends Discovery<QlpacksInfo> {
...
+    private readonly cliServer: CodeQLCliServer
+  ) {
     super();
 
     // Watch for any changes to `qlpack.yml` files in this workspace folder.
     // TODO: The CLI server should tell us what paths to watch for.
     this.watcher.addWatch(new RelativePattern(this.workspaceFolder, '**/qlpack.yml'));
     this.watcher.addWatch(new RelativePattern(this.workspaceFolder, '**/.codeqlmanifest.json'));
     this.push(this.watcher.onDidChange(this.handleQLPackFileChanged, this));
-
```
jcreedcmu, line 29: Can you explain why factoring out these `.refresh()` calls out of the constructor is ok? Do we not need to call refresh when we first construct these for real, outside of tests?

aeisenberg: They are only constructed in one place and we refresh both of them there/ https://github.com/github/vscode-codeql/pull/518/files#diff-070c9f57c79a24599bedea1127e05a06R102-R103

If you prefer, I could stub out the `refresh` method for the tests and move it back to the way it was.

PR #529, opened 2020-08-12
Fix sorting of raw results

jcreedcmu:


extensions/ql-vscode/src/interface.ts
```diff
@@ -235,18 +261,10 @@ export class InterfaceManager extends DisposableObject {
...
         break;
       case 'changeSort':
-        await this.changeSortState(query =>
-          query.updateSortState(
-            this.cliServer,
-            msg.resultSetName,
-            msg.sortState
-          )
-        );
+        await this.changeRawSortState(this.cliServer, msg.resultSetName, msg.sortState);
```
aeisenberg, line 264: Do we need to pass in `this.cliServer` since it is already a property on `this`?

extensions/ql-vscode/src/interface.ts
```diff
@@ -201,10 +202,35 @@ export class InterfaceManager extends DisposableObject {
...
+    await this.postMessage({ t: 'resultsUpdating' });
+    await this._displayedQuery.updateSortState(
+      server,
+      resultSetName,
+      sortState
+    );
+    // Sorting resets to first page, as there is arguably no particular
+    // correlation between the results on the nth page that the user
+    // was previously viewing and the contents of the nth page in a
+    // new sorted order.
```
aeisenberg, line 230: This makes sense.

extensions/ql-vscode/src/interface.ts
jcreedcmu: No, good catch, will just grab it directly

PR #533, opened 2020-08-12
Update changelog after release

aeisenberg:
Add a simple perl script that will augment the CHANGELOG with an [UNRELEASED] section when creating the PR after a release.


jcreedcmu:
Nit: You don't need the parens around postfix `if`'s condition

PR #536, opened 2020-08-13
Fix #535

### LINKED ISSUE #535: Non-primitive columns with a toString but not a url are displayed as "none"
**Describe the bug**

Non-primitive columns with a toString but not a url show as "none" in the results.

**To Reproduce**
Run `class Zero extends int { Zero() { this = 0 }} select 0, 0.(Zero)`.
Alternatively run `newtype TZero = MkZero() class Zero extends TZero { string toString() { result = "foo" }} select MkZero().(Zero)`

**Expected behavior**

You see the number zero/the string "foo"

**Additional context**
I think the issue is in `adaptValue` in `adapt.ts` as that is somewhere where the string "none" appears.

jcreedcmu:


extensions/ql-vscode/src/adapt.ts
```diff
@@ -73,7 +73,7 @@ export function adaptValue(val: ColumnValue): ResultValue {
   }
 
   if (url === undefined) {
-    return 'none';
+    return val.label || 'no label';
```
alexet, line 76: `val.label` can legitimately be the empty string.

jcreedcmu: True, but elsewhere in the extension we also made an effort to display *some* non-empty string to signal clearly that this situation occurred. Would you be ok with, like `[empty label]` or something? In principle it would collide with the label literally being the string `"[empty label]"` but that seems unlikely.

alexet: I think you should be doing the same thing here as an empty string, which is to return it here. I don't mind if it becomes non-empty but we should be consistent. I also think that label is not the right word as I don't think is used as the name of the string representation in other parts of the UI.

jcreedcmu: Hm, ok, I'm convinced. Changing it to turn `undefined` into empty string, and passing empty string up at this stage. If we do want to impose that empty string is shown in some other concrete way in the UI, we can do it farther downstream.

However: the dictionary key `label` is coming directly from the json output by the cli, isn't it? Should we change that in the cli?

jcreedcmu: Or do you mean that it merely isn't the right word to actually show to the user --- in that case I can agree.

PR #537, opened 2020-08-14
Fix changing page forgetting about sorting

jcreedcmu:


extensions/ql-vscode/src/interface.ts
```diff
@@ -270,7 +270,15 @@ export class InterfaceManager extends DisposableObject {
...
         else {
-          await this.showPageOfRawResults(msg.selectedTable, msg.pageNumber);
+          await this.showPageOfRawResults(
+            msg.selectedTable,
+            msg.pageNumber,
+            // When we are in an unsorted state, we guarantee that
+            // sortedResultsInfo doesn't have an entry for the current
+            // result set. Use this to determine whether or not we use
+            // the sorted bqrs file.
+            this._displayedQuery?.sortedResultsInfo.has(msg.selectedTable) || false
```
aeisenberg, line 280: I prefer using `!!` to cast to a Boolean in these cases, but this is fine, too.

jcreedcmu: Somehow I think of `!!` as being for converting something that completely isn't a boolean into the boolean of its truthiness, and `|| false` seems like the right thing for `boolean | undefined`, but six of one half dozen of the other.

PR #544, opened 2020-08-28
Fix highlighting after disembodied IPA branch

### LINKED ISSUE #543: Incorrect syntax highlighting of `private` following a newtype.
The `private` keyword gets highlighted incorrectly in the following case:

```codeql
newtype TA = TB()

private predicate foo() { any() }
```

Specifically, on my machine, it gets highlighted as:

Screenshot 2020-08-28 at 18 26 37

If I add `{ any() }` to the branch of `TB()` it gets the correct highlighting. @alexet wrote the following on Slack: 

> Vscode uses textmate which is regex based so it very hard to do anything fancy. I think that we highlight private as a keyword only after } or ; and imports. I am not sure it is actually possible to syntax highlight QL in vscode using the normal system.

dbartol:
Fixes #543
```ql
newtype TA = TB()

private predicate foo() { any() }
```
Our TextMate grammar didn't realize that the newtype declaration ended after the closing paren of the branch's parameter list, so the `private` modifier was highlighted incorrectly.

It's surprisingly tricky to get TextMate to handle this correctly, so I wound up just treating the IPA declaration head (`newtype TA`), the branch head (`= TB`), the branch parameter list, and the branch body as directly children of the module body. This is kind of hacky, but it does fix the bug without introducing any new cases where we have incorrect highlighting of valid code.

## Checklist

### DIFF
--- a/extensions/ql-vscode/syntaxes/ql.tmLanguage.yml
+++ b/extensions/ql-vscode/syntaxes/ql.tmLanguage.yml
@@ -99,6 +99,12 @@ repository:
   comment-start:
     match: '// | /\*'
 
+  # A pattern that can start a run of whitespace or a comment.
+  # Commonly used as a negative lookahead in the `end` regex of a nonterminal, when searching for
+  # tokens that can't start a child of that nonterminal.
+  whitespace-or-comment-start:
+    match: '\s | $ | (?#comment-start)'
+
   # All tokens that can appear in any context.
   non-context-sensitive:
     patterns:
@@ -113,7 +119,7 @@ repository:
     name: keyword.operator.relational.ql
   
   comparison-operator:
-    match: '=|\!-'
+    match: '=|\!\='
     name: keyword.operator.comparison.ql
   
   arithmetic-operator:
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/syntaxes/ql.tmLanguage.yml
```diff
@@ -113,7 +119,7 @@ repository:
     name: keyword.operator.relational.ql
   
   comparison-operator:
-    match: '=|\!-'
+    match: '=|\!\='
```
aeisenberg, line 122: Was this a bug?

dbartol: Yes, that's a trivial bug I noticed while I was in the code. With the default color settings, it doesn't affect anything, because operators don't get a special color. In custom color schemes, though, we would highlight `!-` instead of `!=`.

aeisenberg: Ah. Makes sense. Good catch.

PR #574, opened 2020-09-09
Refactor: Remove duplicated BQRS types

### LINKED ISSUE #520: Merge duplicated BQRS-related types
Now that we have moved all BQRS parsing from the extension into the cli, the `bqrs-types.ts` module is superseded by `bqrs-cli-types.ts`. This issue is about removing all references to  `bqrs-types.ts` and replacing it with references to `bqrs-cli-types.ts`

aeisenberg:
This refactoring combines the types in `bqrs-types.ts` and
`bqrs-cli-types.ts`. Historically, the former was used for BQRS files
parsed by the extension and the latter for BQRS files parsed by the cli.
They describe the same file types, but using different property and type
names.

We have moved to parsing all BQRS files by the cli. This refactoring
removes the `bqrs-types.ts` file and replaces all BQRS references to
use types in `bqrs-cli-types.ts`.

Additionally, the `adapt.ts` file has been deleted since its purpose
was to convert between extension and cli BQRS types. Some one type and
one function from `adapt.ts` has been moved from `adapt.ts` to 
`bqrs-types.ts`. It's possible that we want to do a further refactoring
to simply remove them both.

This is a purely internal refactoring and does not require any CHANGELOG entry.

Fixes #520.

## Checklist

- [ n/a ] Issues have been created for any UI or other user-facing changes made by this pull request.


extensions/ql-vscode/src/bqrs-cli-types.ts
```diff
@@ -87,15 +75,28 @@ export interface WholeFileLocation {
...
 export type ColumnValue = EntityValue | number | string | boolean;
 
+export type ResultRow = ColumnValue[];
+
+export interface RawResultSet {
+  readonly schema: ResultSetSchema;
+  readonly rows: readonly ResultRow[];
+}
+
+export function adaptBqrs(schema: ResultSetSchema, page: DecodedBqrsChunk): RawResultSet {
```
jcreedcmu, line 93: We still want a comment here, do we not, that this function is similarly deprecated to how all of `adapt.ts` was before? I still am of the belief that we could still push further and have code downstream consume `scheme` and `page` more or less directly, and avoid this repackaging. (not that I am asking for that refactoring to be in this PR) 

Alternatively if you think this function should stay, maybe find it a more meaningful name...

extensions/ql-vscode/src/sarif-utils.ts
```diff
@@ -1,21 +1,23 @@
...
 import * as path from 'path';
-import { LocationStyle, ResolvableLocationValue } from './bqrs-types';
+import { ResolvableLocationValue } from './bqrs-cli-types';
 
 export interface SarifLink {
   dest: number;
   text: string;
 }
 
+interface NoLocation {
```
jcreedcmu, line 10: A comment here would be nice, just something like "The type of a result that has no associated location. `hint` is a string intended for display to the user that explains why there is no location."

extensions/ql-vscode/src/sarif-utils.ts
```diff
@@ -57,68 +59,78 @@ export function parseSarifPlainTextMessage(message: string): SarifMessageCompone
...
+    return { hint: 'no artifact location' };
   if (physicalLocation.artifactLocation.uri === undefined)
-    return { t: 'NoLocation', hint: 'artifact location has no uri' };
+    return { hint: 'artifact location has no uri' };
 
   // This is not necessarily really an absolute uri; it could either be a
   // file uri or a relative uri.
   const uri = physicalLocation.artifactLocation.uri;
 
+  // FIXME: This is probably wrong
```
jcreedcmu, line 88: What's going on here? Why are you changing it to return the full uri instead of the path?

extensions/ql-vscode/src/view/RawTableValue.tsx
```diff
@@ -1,22 +1,30 @@
...
+    || typeof v === 'number'
+    || typeof v === 'boolean') {
     return <span>{v}</span>;
   }
-  else if ('uri' in v) {
-    return <a href={v.uri}>{v.uri}</a>;
-  }
-  else {
-    return renderLocation(v.location, v.label, props.databaseUri);
+
```
jcreedcmu, line 19: (1) I fear that the fallthrough to silently rendering `<span />` when the location is not one of the known forms might lead to confusing debugging later. At least log what `loc` was, maybe?
(2) Can this logic be factored out as its own function in some way that can be shared with `renderSarifLocationWithText` and `renderSarifLocation`?

extensions/ql-vscode/src/view/results.tsx
```diff
@@ -113,7 +111,11 @@ class App extends React.Component<{}, ResultsViewState> {
             resultSet: {
               t: 'SarifResultSet',
               name: ALERTS_TABLE_NAME,
-              schema: { name: ALERTS_TABLE_NAME, version: 0, columns: [], tupleCount: 1 },
+              schema: {
```
jcreedcmu, line 114: Is there a good reason for dropping the `version: 0`? Is that field just not in this type now and I am confused?

extensions/ql-vscode/src/view/results.tsx
```diff
@@ -174,7 +176,10 @@ class App extends React.Component<{}, ResultsViewState> {
     resultsInfo: ResultsInfo
   ): Promise<readonly ResultSet[]> {
     const parsedResultSets = resultsInfo.parsedResultSets;
-    return [{ t: 'RawResultSet', ...parsedResultSets.resultSet }];
+    return [{
+      ...parsedResultSets.resultSet,
+      t: (parsedResultSets.resultSet.t ?? 'RawResultSet') as any
```
jcreedcmu, line 181: forgot that null-coalescing actually made it into the language!

can you leave a comment that explains the situation in which `parsedResultSets.resultSet.t` can either be or not be undefined, though? I'm actually confused why the types allow that. If this could be rephrased to avoid the `as any` that would be even better, but I'm not religiously opposed to it if it makes sense.

extensions/ql-vscode/src/vscode-tests/no-workspace/interface-utils.test.ts
```diff
@@ -73,20 +72,17 @@ describe('interface-utils', () => {
...
-          },
+          'file://hucairz:1:1:1:1',
           mockDatabaseItem
         )
       ).to.deep.equal(
         new vscode.Location(
-          vscode.Uri.parse('abc'),
-          new vscode.Range(0, 0, 0, 0)
+          vscode.Uri.file('abc'),
+          new vscode.Range(0, 0, 0, 1)
```
jcreedcmu, line 85: I'm slightly suspicious at the test output changing. We're sure this is ok? Was it wrong before?

jcreedcmu (changes requested):

extensions/ql-vscode/src/bqrs-cli-types.ts
aeisenberg: Yeah. I wasn't happy about keeping this function. It makes the object shape slightly more intuitive, but it is just an extra step that is unnecessary. I'll add a comment here and perhaps we can remove in the future.

aeisenberg: I'll keep it for now since it makes things slightly easier to work with, though I'm not entirely happy with it. I'll mark it with a TODO that we may want to change in the future.

extensions/ql-vscode/src/view/RawTableValue.tsx
aeisenberg: I am trying to reproduce the logic of #546. When the location is "empty", we should render as an empty span. I can refactor to make this more explicit.

extensions/ql-vscode/src/sarif-utils.ts
aeisenberg: I should have removed that comment, since I have tested and it is working.

The old style used a path, but the new (sarif style) uses a uri. It's annoyingly complex, but at least we're only using one style here. I can work on this and make it easier to understand. Probably a unit test or two would be good as well.

extensions/ql-vscode/src/view/results.tsx
aeisenberg: I honestly can't remember. I'll check on this.

aeisenberg: Right...this is was always the behaviour. (ie- use `RawResultSet` if no `t` is specifed, otherwise use whatever is in `t`. The reason why I changed is that the latest version of tsc started screaming about this.

I honestly don't know which situations we'll have an empty `t` and if we do, it's probably a bug. I'm partially thinking that we delete it and log any errors.

jcreedcmu: Hm, ok, then I consider the way the code is now to be a sufficient red flag that this should ideally be fixed. Thanks, tsc :+1:

jcreedcmu (approved):

PR #590, opened 2020-09-25
Add the commandRunner

### LINKED ISSUE #464: Add ability for user to cancel all simultaneously running queries
#462 introduced multi-running queries, where a user can select multiple queries and run them all at once. However, once initiated, a user can't cancel all queries at once, only individually. This is particularly problematic when a user kicks off lots of queries.

This issue is about enabling the user to cancel all queries in a single action.

An open question is, do we want to enable the user to cancel _all_ running queries at once, or only the queries for the particular simultaneous run?

----

 ## Update 

Here's a summary and rationale behind the approach.

Right now the way we register vscode commands is ad hoc and spread throughout the code. (A command is just something that runs a function whenever you perform a keystroke combination or some other task). Because of this, each command is handling it's own errors (in a somewhat consistent way). Also, progress monitors are not applied consistently, so not everything is able to be cancelled. This refactoring is about extracting all this to a single location so every command is registered consistently and this refactoring will make it possible to use a single progress monitor when running multiple queries and allow all of them to be cancelled with one click.

### LINKED ISSUE #534: Run queries in folders on an old database opens many upgrade dialogs
**Describe the bug**
When using "Run queries in selected files" on a CodeQL db that needs an upgrade you get the dialog once per query.

**To Reproduce**
Try to run multiple queries at once on an outdated CodeQL db

**Expected behavior**
One upgrade prompt between all queries

aeisenberg:
The commandRunner wraps all vscode command registrations. It provides
uniform error handling and an optional progress monitor.

In general, progress monitors should only be created by the
commandRunner and passed through to the locations that use it.

Resolves #464.
Resolves #534.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Max number of simultaneous queries launchable by runQueries command is now configurable by changing the `codeQL.runningQueries.maxQueries` setting.
 - Allow simultaneously run queries to be canceled in a single-click.
 - Prevent multiple upgrade dialogs from appearing when running simultaneous queries on upgradeable databases.
+- Allow simultaneously run queries to be canceled in a single-click.
 - Max number of simultaneous queries launchable by runQueries command is now configurable by changing the codeQL.runningQueries.maxQueries setting.
 - Fix sorting of results. Some pages of results would have the wrong sort order and columns.
 - Remember previous sort order when reloading query results.
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Add friendly welcome message when the databases view is empty.
 - Add open query, open results, and remove query commands in the query history view title bar.
 - Max number of simultaneous queries launchable by runQueries command is now configurable by changing the `codeQL.runningQueries.maxQueries` setting.
+- Allow simultaneously run queries to be canceled in a single-click.
+- Prevent multiple upgrade dialogs from appearing when running simultaneous queries on upgradeable databases.
+- Max number of simultaneous queries launchable by runQueries command is now configurable by changing the codeQL.runningQueries.maxQueries setting.
 - Fix sorting of results. Some pages of results would have the wrong sort order and columns.
 - Remember previous sort order when reloading query results.
 - Fix proper escaping of backslashes in SARIF message strings.
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -4,11 +4,9 @@
 
 - Add friendly welcome message when the databases view is empty.
 - Add open query, open results, and remove query commands in the query history view title bar.
-- Max number of simultaneous queries launchable by runQueries command is now configurable by changing the `codeQL.runningQueries.maxQueries` setting.
+- The maximum number of simultaneous queries launchable by the `CodeQL: Run Queries in Selected Files` command is now configurable by changing the `codeQL.runningQueries.maxQueries` setting.
 - Allow simultaneously run queries to be canceled in a single-click.
 - Prevent multiple upgrade dialogs from appearing when running simultaneous queries on upgradeable databases.
-- Allow simultaneously run queries to be canceled in a single-click.
-- Max number of simultaneous queries launchable by runQueries command is now configurable by changing the codeQL.runningQueries.maxQueries setting.
 - Fix sorting of results. Some pages of results would have the wrong sort order and columns.
 - Remember previous sort order when reloading query results.
 - Fix proper escaping of backslashes in SARIF message strings.
--- a/extensions/ql-vscode/src/contextual/locationFinder.ts
+++ b/extensions/ql-vscode/src/contextual/locationFinder.ts
@@ -1,13 +1,14 @@
 import * as vscode from 'vscode';
 
 import { decodeSourceArchiveUri, zipArchiveScheme } from '../archive-filesystem-provider';
-import { ColumnKindCode, EntityValue, getResultSetSchema } from '../bqrs-cli-types';
+import { ColumnKindCode, EntityValue, getResultSetSchema, ResultSetSchema } from '../bqrs-cli-types';
 import { CodeQLCliServer } from '../cli';
 import { DatabaseManager, DatabaseItem } from '../databases';
 import fileRangeFromURI from './fileRangeFromURI';
 import * as messages from '../messages';
 import { QueryServerClient } from '../queryserver-client';
 import { QueryWithResults, compileAndRunQueryAgainstDatabase } from '../run-queries';
+import { ProgressCallback } from '../helpers';
 import { KeyType } from './keyType';
 import { qlpackOfDatabase, resolveQueries } from './queryResolver';
 
@@ -28,6 +29,8 @@ export interface FullLocationLink extends vscode.LocationLink {
  * @param dbm The database manager
  * @param uriString The selected source file and location
  * @param keyType The contextual query type to run
+ * @param progress A progress callback
+ * @param token A CancellationToken
  * @param filter A function that will filter extraneous results
  */
 export async function getLocationsForUriString(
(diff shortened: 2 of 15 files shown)


aeisenberg:
This is still a draft, but it is good enough to have a look and get some feedback. I need to do some more manual testing as well as add some unit tests.

extensions/ql-vscode/src/contextual/locationFinder.ts
```diff
@@ -58,8 +59,25 @@ export async function getLocationsForUriString(
           }
         }
       };
-      const results = await compileAndRunQueryAgainstDatabase(cli, qs, db, false, vscode.Uri.file(query), templates);
-      if (results.result.resultType == messages.QueryResultType.SUCCESS) {
+
+      const results: QueryWithResults = await withProgress({
```
aeisenberg, line 63: Move this call so that it is wrapped around the largest area possible.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -3,8 +3,7 @@ import * as unzipper from 'unzipper';
 import { zip } from 'zip-a-folder';
 import {
   Uri,
-  ProgressOptions,
-  ProgressLocation,
+  CancellationToken,
```
aeisenberg, line 6: This file is mostly about disentangling all the error handling and progress reporting.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -28,42 +25,32 @@ import { logger } from './logging';
...
+    return;
   }
 
+  validateHttpsUrl(databaseUrl);
+
+  const item = await databaseArchiveFetcher(
+    databaseUrl,
+    databasesManager,
+    storagePath,
+    progress
```
aeisenberg, line 45: Can we use the cancel token here to make the operation cancellable?

extensions/ql-vscode/src/extension.ts
```diff
@@ -11,14 +11,19 @@ import {
   env,
   window
 } from 'vscode';
-import { LanguageClient } from 'vscode-languageclient';
+import { CancellationToken, LanguageClient } from 'vscode-languageclient';
```
aeisenberg, line 14: Wrong cancellation token being used here.

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -2,12 +2,11 @@ import * as fs from 'fs-extra';
 import * as yaml from 'js-yaml';
 import * as path from 'path';
 import { ExtensionContext, window as Window, workspace, Uri } from 'vscode';
-import { ErrorCodes, ResponseError } from 'vscode-languageclient';
+import { CancellationToken, ErrorCodes, ResponseError } from 'vscode-languageclient';
```
aeisenberg, line 5: Wrong `CancellationToken`.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -3,7 +3,7 @@ import * as fs from 'fs-extra';
 import * as path from 'path';
 import * as tmp from 'tmp';
 import * as vscode from 'vscode';
-import { ErrorCodes, ResponseError } from 'vscode-languageclient';
+import { CancellationToken, ErrorCodes, ResponseError } from 'vscode-languageclient';
```
aeisenberg, line 6: Wrong `CancellationToken`

extensions/ql-vscode/src/databaseFetcher.ts
aeisenberg: Not so easy. The download and unzip functions are from third parties and do not easily accept cancellations.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [UNRELEASED]
 
+- Allow simltaneously run queries to be canceled in a single-click.
```
shati-patel, line 5:
```suggestion
- Allow simultaneously run queries to be canceled in a single-click.
```
(teeny tiny typo 📝)

aeisenberg: Thanks!

extensions/ql-vscode/src/contextual/templateProvider.ts
```diff
@@ -45,14 +45,22 @@ export class TemplateQueryDefinitionProvider implements vscode.DefinitionProvide
...
   private async getDefinitions(uriString: string): Promise<vscode.LocationLink[]> {
-    return getLocationsForUriString(
-      this.cli,
-      this.qs,
-      this.dbm,
-      uriString,
-      KeyType.DefinitionQuery,
-      (src, _dest) => src === uriString
-    );
+    return await withProgress({
```
adityasharad, line 48: Here and elsewhere: This `return await` can just be a `return` right? Otherwise we'll await the outcome of the promise, then wrap up the result in a second promise anyway, and return that.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -28,42 +25,32 @@ import { logger } from './logging';
...
-  } catch (e) {
-    showAndLogErrorMessage(e.message);
+  const databaseUrl = await window.showInputBox({
+    prompt: 'Enter URL of zipfile of database to download',
+  });
+  if (!databaseUrl) {
+    return;
   }
 
+  validateHttpsUrl(databaseUrl);
```
adityasharad, line 39: Do you still need the `try-catch` for when this throws on invalid URLs? Or are you fine with letting that reach the top level?

extensions/ql-vscode/src/helpers.ts
```diff
@@ -50,10 +78,58 @@ export function withProgress<R>(
...
+ * Otherwise, no progress monitor will be used.
+ *
+ * If a task is run with a progress monitor, the first two arguments to the task are always
+ * the progress callback, and the cancellation token. And this is followed by any extra command arguments
+ * (eg- selection, multiselection, ...).
+ *
+ * If there is no progress monitor, then only extra command arguments are passed in.
+ *
+ * @param commandId The ID of the command to register.
+ * @param task The task to run. If passing taskOptions, then this task must be a `ProgressTask`.
```
adityasharad, line 100:
```suggestion
 * @param task The task to run. If passing `progressOptions`, then this task must be a `ProgressTask`.
```

extensions/ql-vscode/src/helpers.ts
```diff
@@ -50,10 +78,58 @@ export function withProgress<R>(
...
+        // User has cancelled this action manually
+        if (e.silent) {
+          logger.log(e.message);
+        } else {
+          showAndLogWarningMessage(e.message);
+        }
+      } else if (e instanceof Error) {
+        showAndLogErrorMessage(e.message);
+      } else {
+        throw e;
```
adityasharad, line 127: Should we always handle this (show and log) to avoid uncaught promise rejections at the top level?

extensions/ql-vscode/src/helpers.ts
```diff
@@ -50,10 +78,58 @@ export function withProgress<R>(
...
+ *
+ * @param commandId The ID of the command to register.
+ * @param task The task to run. If passing taskOptions, then this task must be a `ProgressTask`.
+ * @param progressOptions Optional argument. If present, then the task is run with a progress monitor
+ *                    and cancellation token, otherwise it is run with no arguments.
+ */
+export function commandRunner<R>(
+  commandId: string,
+  task: ProgressTask<R> | NoProgressTask,
+  progressOptions?: ProgressOptions
```
adityasharad, line 107: I wonder whether it would be safer to split this into two variants:
- accept `task: ProgressTask<R>` and `progressOptions`
- accept `task: NoProgressTask` and no `progressOptions`.

That way it's obvious whenever we attempt to write new code that uses this.

extensions/ql-vscode/src/databaseFetcher.ts
adityasharad: Reading further I think this reaches the top-level error handler in `commandRunner`, so that's fine.

aeisenberg:
Thanks for the review. I rebased and didn't address any of your comments yet. I will tomorrow.

extensions/ql-vscode/src/contextual/templateProvider.ts
aeisenberg: It's redundant. I can remove. Though, I don't think the behaviour is any different due to hoisting of promises. It is essentially this snippet:

```js
async function a() {
  throw new Error('1')
}

async function b1() {
  return await a()
}

async function b2() {
  return a()
}

async function c() {
  try {
    await b1()
  } catch (e) {
    console.log(e)
  }
  try {
    await b2()
  } catch (e) {
    console.log(e)
  }
}

c()
```

Both errors are properly captured.

extensions/ql-vscode/src/databaseFetcher.ts
aeisenberg: Yes.

extensions/ql-vscode/src/helpers.ts
aeisenberg: Our code should _always_ be throwing errors. If a non-error is thrown, then that is just...strange and something unexpected is happening. I don't feel strongly about this since it _shouldn't_ happen. I can remove.

aeisenberg: Need to handle the case where `e.message` is undefined.

extensions/ql-vscode/src/helpers.ts
aeisenberg: So, `commandRunner` and `commandRunnerWithProgress`?  I guess this would allow us to ensure we send the proper number of arguments for each style.

adityasharad: That sounds fine. I don't know whether TS overloading rules will let you have the same name with different types and arities.

aeisenberg: It's actually a good thing that I changed to different names. It uncovered one instance when I was calling a command handler with the wrong arguments.

extensions/ql-vscode/src/helpers.ts
```diff
@@ -50,10 +79,94 @@ export function withProgress<R>(
...
+ *
+ * There are two ways to invoke the command task: with or without a progress monitor
+ * If progressOptions are passed in, then the command task will run with a progress monitor
+ * Otherwise, no progress monitor will be used.
+ *
+ * If a task is run with a progress monitor, the first two arguments to the task are always
+ * the progress callback, and the cancellation token. And this is followed by any extra command arguments
+ * (eg- selection, multiselection, ...).
+ *
+ * If there is no progress monitor, then only extra command arguments are passed in.
```
adityasharad, line 98: Split this across the two methods, so each documents the specific case it handles.

adityasharad (approved):

PR #596, opened 2020-09-28
Fix Query sort issues

### LINKED ISSUE #586: Sorting in the result viewer doesn't work with pagination.
**Describe the bug**
If I click a column in the result viewer to sort and then click to the next page of results, then the sorting is no longer applied.

**To Reproduce**
I can currently reproduce by running
```ql
from int x, int y
where x in [1..15] and y in [1..15]
select x, y
```
and sorting on `y`. Then the results are no longer sorted when I go to page 2.

**Expected behavior**
Once results are sorted by a specific column I expect that sorting to be respected across several pages of results.

### LINKED ISSUE #587: Sorting in the result viewer is not preserved when reloading the results.
**Describe the bug**
If I sort a set of results by clicking a column, and then reload the results by clicking in the Query History (either directly or after having viewed some other result set), then the column that I sorted on is remembered and displayed, but the results are no longer sorted as indicated by the triangle next to the column name.

**To Reproduce**
Execute a query. For example the following:
```ql
from int x, int y
where x in [1..2] and y in [1..2]
select x, y
```
Click a column to sort the results (the sorting must differ from the default ordering in order for the problem to be observable), in this case clicking `y` should do the trick. Click the line for the query in the Query History, such that the results reload. Observe that `y` is still highlighted as the sorted column, but the tuples are back in the default order.

**Expected behavior**
I expect either that the triangle would be cleared when reloading results to indicate that the sorting was forgotten, or alternatively, since the sorted column is clearly remembered and redisplayed with the triangle icon, that the sorting is reapplied. The latter option seems to be the more useful option.

aeisenberg:
Fix Query sort issues

## Fix invalid sort after reloading query results

With this change, we use the stored sort order if it exists
after reloading a query results page.

## Fix incorrect call to bqrs info when retrieving paginated results

When retrieving paginated results, need to make sure we are getting
page offsets from the correct results file.

Previously, we were incorrectly extracting page offsets from the default
(unsorted) file. With this change, we ensure that we get offsets from
the proper results file when there is a request for a page of results.

Resolves #586
Resolves #587

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/interface.ts
+++ b/extensions/ql-vscode/src/interface.ts
@@ -364,8 +364,13 @@ export class InterfaceManager extends DisposableObject {
         (resultSet) => resultSet.name == selectedTable
       )!;
 
+      // Use sorted results path if it exists. This may happen if we are
+      // reloading the results view after it has been sorted in the past.
+      const resultsPath = results.sortedResultsInfo.get(selectedTable)?.resultsPath
+        || results.query.resultsPaths.resultsPath;
+
       const chunk = await this.cliServer.bqrsDecode(
-        results.query.resultsPaths.resultsPath,
+        resultsPath,
         schema.name,
         {
           offset: schema.pagination?.offsets[0],
--- a/extensions/ql-vscode/src/interface.ts
+++ b/extensions/ql-vscode/src/interface.ts
@@ -365,14 +365,16 @@ export class InterfaceManager extends DisposableObject {
 
     // Use sorted results path if it exists. This may happen if we are
     // reloading the results view after it has been sorted in the past.
-    const resultsPath = results.sortedResultsInfo.get(selectedTable)?.resultsPath
-      || results.query.resultsPaths.resultsPath;
+    const resultsPath = results.getResultsPath(selectedTable);
 
     const chunk = await this.cliServer.bqrsDecode(
       resultsPath,
       schema.name,
       {
-        // always use the first page. –
+        // Always send the first page.
+        // It may not wind up being the page we actually show,
+        // if there are interpreted results, but speculatively
+        // send anyway.
         offset: schema.pagination?.offsets[0],
         pageSize: RAW_RESULTS_PAGE_SIZE
       }
@@ -433,8 +435,7 @@ export class InterfaceManager extends DisposableObject {
   }
 
   private async getResultSetSchemas(results: CompletedQuery, selectedTable = ''): Promise<ResultSetSchema[]> {
-    const resultsPath = results.sortedResultsInfo.get(selectedTable)?.resultsPath
-        || results.query.resultsPaths.resultsPath;
+    const resultsPath = results.getResultsPath(selectedTable);
     const schemas = await this.cliServer.bqrsInfo(
       resultsPath,
       RAW_RESULTS_PAGE_SIZE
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/interface.ts
```diff
@@ -352,35 +352,39 @@ export class InterfaceManager extends DisposableObject {
...
 
-      // This may not wind up being the page we actually show, if there are interpreted results,
-      // but speculatively send it anyway.
-      const selectedTable = getDefaultResultSetName(resultSetNames);
-      const schema = resultSetSchemas.find(
-        (resultSet) => resultSet.name == selectedTable
-      )!;
+    // Use sorted results path if it exists. This may happen if we are
+    // reloading the results view after it has been sorted in the past.
+    const resultsPath = results.sortedResultsInfo.get(selectedTable)?.resultsPath
```
adityasharad, line 368: Would it be useful to have this as a helper function on `CompletedQuery`, so we can ensure we always use this logic when getting the results path?

extensions/ql-vscode/src/interface.ts
```diff
@@ -352,35 +352,38 @@ export class InterfaceManager extends DisposableObject {
...
-        numInterpretedPages: numInterpretedPages(this._interpretation),
-        resultSet: { t: 'RawResultSet', ...resultSet },
-        selectedTable: undefined,
-        resultSetNames,
-      };
+    const chunk = await this.cliServer.bqrsDecode(
+      resultsPath,
+      schema.name,
+      {
+        // always use the first page. –
```
jcreedcmu, line 374: Please preserve the `// This may not wind up ... ` comment and put it here.

extensions/ql-vscode/src/interface.ts
```diff
@@ -390,7 +393,7 @@ export class InterfaceManager extends DisposableObject {
       resultsPath: this.convertPathToWebviewUri(
         results.query.resultsPaths.resultsPath
       ),
-      parsedResultSets: await getParsedResultSets(),
+      parsedResultSets,
```
jcreedcmu, line 396: I was going to ask why you inlined this function, since obviously it was called in multiple places... but I see that it's not. Maybe it was in the past. Agree that inlining it now makes sense.

extensions/ql-vscode/src/query-results.ts
```diff
@@ -72,6 +71,10 @@ export class CompletedQuery implements QueryWithResults {
     }
   }
 
+  getResultsPath(selectedTable: string, useSorted = true): string {
+    return (useSorted && this.sortedResultsInfo.get(selectedTable)?.resultsPath)
```
jcreedcmu, line 75: Any reason not to `useSorted ? this.sortedResultsInfo.get(selectedTable)?.resultsPath) : this.query.resultsPaths.resultsPath`? Am I missing something about the use of the short-circuiting boolean operators?

extensions/ql-vscode/src/query-results.ts
```diff
@@ -89,9 +92,8 @@ export class CompletedQuery implements QueryWithResults {
   }
 
   getLabel(): string {
-    if (this.options.label !== undefined)
-      return this.options.label;
-    return this.config.format;
+      return this.options?.label
+        || this.config.format;
```
jcreedcmu, line 96: Is changing behavior on `this.options?.label == ""` desirable? (before it would have returned it, now it falls through to `this.config.format`) Kind of a corner case, and maybe your change is for the better; if the user explicitly tries to set the label to empty string, maybe we should behave as if they want the default back. Just wanted to make sure we're both aware.

jcreedcmu (approved):
LGTM modulo nits

extensions/ql-vscode/src/query-results.ts
aeisenberg: I was just trying to be concise, but it's actually confusing to read. I think I'll make it longer and simpler.

This is not correct:

```js
useSorted ? this.sortedResultsInfo.get(selectedTable)?.resultsPath) : this.query.resultsPaths.resultsPath
```

because if `this.sortedResultsInfo.get(selectedTable)?.resultsPath)` is falsy then we need to fall back on `this.query.resultsPaths.resultsPath` hence the short-circuiting.

extensions/ql-vscode/src/query-results.ts
aeisenberg: Yes. That's what I was thinking. I don't think people would want an empty label.

extensions/ql-vscode/src/interface.ts
aeisenberg: Yes. Done.

adityasharad (approved):
Looks sensible, and happy to see more tests!

jcreedcmu (approved):

extensions/ql-vscode/src/query-results.ts
jcreedcmu: Ah, yeah, less concise but more clear would be preferred, since in fact I did misread it by accident.

PR #598, opened 2020-09-29
Fix "View SARIF" context menu race.

jcreedcmu:
Fixes #598.

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -50,6 +50,15 @@ const SHOW_QUERY_TEXT_QUICK_EVAL_MSG = `\
  */
 const FAILED_QUERY_HISTORY_ITEM_ICON = 'media/red-x.svg';
 
+export async function updateTreeItemContextValue(element: CompletedQuery): Promise<void> {
+  // Mark this query history item according to whether it has a
+  // SARIF file so that we can make context menu items conditionally
+  // available.
+  element.treeItem!.contextValue = (await element.query.hasInterpretedResults())
+    ? 'interpretedResultsItem'
+    : 'rawResultsItem';
+}
+
 /**
  * Tree data provider for the query history view.
  */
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/query-history.ts
```diff
@@ -77,6 +86,9 @@ class HistoryTreeDataProvider
   constructor(private ctx: ExtensionContext) { }
 
   async getTreeItem(element: CompletedQuery): Promise<vscode.TreeItem> {
+    if (element.treeItem !== undefined)
```
aeisenberg, line 89: I think it would be nicer to convert `treeItem` to a getter on `CompletedQuery`. And move all this code into the getter. It avoids a type assertion on `updateTreeItemContextValue`.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -50,6 +50,15 @@ const SHOW_QUERY_TEXT_QUICK_EVAL_MSG = `\
  */
 const FAILED_QUERY_HISTORY_ITEM_ICON = 'media/red-x.svg';
 
+export async function updateTreeItemContextValue(element: CompletedQuery): Promise<void> {
```
aeisenberg, line 53: Would this make more sense as a member function on `CompletedQuery`?

aeisenberg (commented):
I think your implementation is fine, though I think it can be cleaned up somewhat.

jcreedcmu:
I tried testing this and it isn't working yet for me.

extensions/ql-vscode/src/query-history.ts
jcreedcmu: Good idea.

jcreedcmu: Wait, no, this function needs to have the context of the treeview. The `CompletedQuery` doesn't have that information, or at least doesn't plausibly have it until the treeitem is created.

extensions/ql-vscode/src/query-history.ts
jcreedcmu: I think it makes sense as a member function on `QueryHistoryManager`, because of my attempting to decouple `CompletedQuery` from having to know too much about the whole `QueryHistoryManager` and the associated tree data provider. Already I don't love having to put the tree item itself as a field in `CompletedQuery`, but this is the simplest thing that I can see works.

jcreedcmu:
I had forgotten to actually `.refresh()` the tree data provider. This now seems to fix the observed bug correctly.

aeisenberg (approved):
Code is good. OK to merge if it is working.

PR #601, opened 2020-09-29
Ensure all fields have labels

### LINKED ISSUE #535: Non-primitive columns with a toString but not a url are displayed as "none"

aeisenberg:
Never show an empty string in the results view. This fixes #535 (again).

Thanks @shati-patel for noticing this. It was indeed a regression with #574.

## Checklist

- [ n/a ] Issues have been created for any UI or other user-facing changes made by this pull request.

### DIFF
--- a/extensions/ql-vscode/src/view/result-table-utils.tsx
+++ b/extensions/ql-vscode/src/view/result-table-utils.tsx
@@ -68,18 +68,18 @@ export function renderLocation(
   callback?: () => void
 ): JSX.Element {
 
-  if (loc === undefined) {
-    return <span />;
-  } else if (isStringLoc(loc)) {
-    return <a href={loc}>{loc}</a>;
-  }
-
   // If the label was empty, use a placeholder instead, so the link is still clickable.
   let displayLabel = label;
   if (!label) {
     displayLabel = '[empty string]';
   } else if (label.match(/^\s+$/)) {
     displayLabel = `[whitespace: "${label}"]`;
+  }
+
+  if (loc === undefined) {
+    return <span>{displayLabel}</span>;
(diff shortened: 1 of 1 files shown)


shati-patel (approved):
Thanks for the fix! This works for me locally 😺

jcreedcmu (approved):
This LGTM as a fix, but does this or does this not mean that something's already going wrong earlier that causes `loc` to be undefined when it shouldn't be?

aeisenberg:
Here is the sample query that triggers the issue. I think that the fact there is no location is correct.

```
class Zero extends int {
  Zero() {
    this = 0
  }
}

select 0, 0.(Zero)
```

If I change that query to this and add a location explicitly, the link works (assuming you have a `check_cli_conventions.go` file at that path in your source archive).

```
class Zero extends int {
  Zero() {
    this = 0
  }

  predicate hasLocationInfo(string path, int sl, int sc, int el, int ec) {
    path = "/opt/src/cmd/clicheck/check_cli_conventions.go" and
    sl = 0 and
    sc = 0 and
    el = 0 and
    ec = 0
  }
}

select 0, 0.(Zero)
```

PR #602, opened 2020-09-30
Clear problems view when a database is removed

### LINKED ISSUE #525: User should be able to configure page size in results viewer
Currently, the page size in the results viewer is fixed at 100. The user should be able to configure this for other page sizes.

My thought is that on the results page, we include a drop down next to the pagination section where the user can pick different page sizes. Say: 50, 100, 200, 500, 1000, 2000. And perhaps this value should be a regular setting as well so it can be saved across restarts.

aeisenberg:
This commit fixes the problem whereby a database is removed and the
problems associated with queries run from that database stick around
in the problems view.

Also, once problems are cleared, we need to make sure that we uncheck
the checkbox in the results view.

This commit has several limitations:

1. There is duplicated code for message handling in both results.tsx and
result-tables.tsx.
2. Problems are cleared whenever there is *any* change to any database.
Ideally we should only clear problems when a database is removed and
only problems associated with that database. I'll fix part of this in
a future commit.

Resolves #525

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -4,12 +4,11 @@
 
 - Add friendly welcome message when the databases view is empty.
 - Add open query, open results, and remove query commands in the query history view title bar.
-- Max number of simultaneous queries launchable by runQueries command is now configurable by changing the codeQL.runningQueries.maxQueries setting.
+- Max number of simultaneous queries launchable by runQueries command is now configurable by changing the `codeQL.runningQueries.maxQueries` setting.
 - Fix sorting of results. Some pages of results would have the wrong sort order and columns.
 - Remember previous sort order when reloading query results.
 - Fix proper escaping of backslashes in SARIF message strings.
 - Allow setting `codeQL.runningQueries.numberOfThreads` and `codeQL.runningTests.numberOfThreads` to 0, (which is interpreted as 'use one thread per core on the machine').
-- Clear the problems view of all Code QL problems when a database is removed.
 - Clear the problems view of all CodeQL query results when a database is removed.
 
 ## 1.3.3 - 16 September 2020
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fix proper escaping of backslashes in SARIF message strings.
 - Allow setting `codeQL.runningQueries.numberOfThreads` and `codeQL.runningTests.numberOfThreads` to 0, (which is interpreted as 'use one thread per core on the machine').
 - Clear the problems view of all Code QL problems when a database is removed.
+- Clear the problems view of all CodeQL query results when a database is removed.
 
 ## 1.3.3 - 16 September 2020
 
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Remember previous sort order when reloading query results.
 - Fix proper escaping of backslashes in SARIF message strings.
 - Allow setting `codeQL.runningQueries.numberOfThreads` and `codeQL.runningTests.numberOfThreads` to 0, (which is interpreted as 'use one thread per core on the machine').
+- Clear the problems view of all Code QL problems when a database is removed.
 
 ## 1.3.3 - 16 September 2020
 
--- a/extensions/ql-vscode/src/databases-ui.ts
+++ b/extensions/ql-vscode/src/databases-ui.ts
@@ -16,6 +16,7 @@ import * as fs from 'fs-extra';
 
 import * as cli from './cli';
 import {
+  DatabaseChangedEvent,
   DatabaseItem,
   DatabaseManager,
   getUpgradesDirectories,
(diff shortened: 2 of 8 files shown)


jcreedcmu (dismissed):

aeisenberg:
There are a few changes I want to make here. Please don't merge yet.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Fix sorting of results. Some pages of results would have the wrong sort order and columns.
 - Remember previous sort order when reloading query results.
 - Fix proper escaping of backslashes in SARIF message strings.
+- Clear the problems view of all Code QL problems when a database is removed.
```
adityasharad, line 9:
```suggestion
- Clear the problems view of all CodeQL query results when a database is removed.
```

extensions/ql-vscode/src/view/result-tables.tsx
```diff
@@ -101,7 +103,17 @@ export class ResultTables
...
     const selectedPage = (props.parsedResultSets.pageNumber + 1) + '';
-    this.state = { selectedTable, selectedPage };
+    this.state = {
+      selectedTable,
+      selectedPage,
+      problemsViewSelected: false
+    };
+  }
+
+  untoggleProbemsView() {
```
adityasharad, line 113:
```suggestion
  untoggleProblemsView() {
```

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -101,19 +100,19 @@ class DatabaseTreeDataProvider extends DisposableObject
     return this._onDidChangeTreeData.event;
   }
 
-  private handleDidChangeDatabaseItem = (
-    databaseItem: DatabaseItem | undefined
-  ): void => {
-    this._onDidChangeTreeData.fire(databaseItem);
+  private handleDidChangeDatabaseItem = (event: DatabaseChangedEvent): void => {
+    // Note that events from the databse manager are instances of DatabaseChangedEvent
+    // and events fired by the UI are instances of DatabaseItem
```
adityasharad, line 105: Will we ever receive a `DatabaseItem` here, and so need to handle `event.item || event as DatabaseItem`?

extensions/ql-vscode/src/view/result-tables.tsx
```diff
@@ -244,6 +269,40 @@ export class ResultTables
...
+          problemsViewSelected: false
+        });
+        break;
+
+      default:
+        // noop
+    }
+  }
+
+  // TODO: Duplicated from results.tsx consider a way to
```
adityasharad, line 286: Can we access the parent's `vscodeMessageHandler`?

extensions/ql-vscode/src/interface-types.ts
```diff
@@ -127,11 +127,16 @@ export interface NavigatePathMsg {
   direction: number;
 }
 
+export interface UntoggleShowProblemsMsg {
```
adityasharad, line 130: There aren't many comments in this file currently, but would be nice to at least add one to explain the meaning of each new message, as it relates to user behaviour.

adityasharad (commented):
Looks sensible. Minor changelog conflict to resolve and some clarifying questions.

You mentioned wanting to add tests - is that feasible now or would you like to do that as a follow-up?

extensions/ql-vscode/src/view/result-tables.tsx
aeisenberg: done

extensions/ql-vscode/src/databases-ui.ts
aeisenberg: `item` is an optional field on the event. The `onDidChangeTreeData` method is from the super class. When that method is called, this triggers the tree view to update. If an item is passed, then only that item (and children, but db view items have no children) is updated. If no item is passed, then the entire tree is updated. 

When we add or remove an item, the entire tree needs to be updated and when an item is renamed, we only need to update that single item.

Hopefully, that clears things up a little. I'll add a comment on that matter.

extensions/ql-vscode/src/view/result-tables.tsx
aeisenberg: Not directly. We would need to pass it through the properties. And we'd still need to make calls on (un-)mount. So, any solution I could come up with where we don't duplicate code includes more boilerplate than we have now.

aeisenberg: We could just extract the message handler into a react context that is shared throughout the react tree.

It would still be more code/boilerplate than we have now. But, at least it would be using proper react idioms. Though, I'd still argue it's more complex than what we have now. Can we consider this a future enhancement in case we need to share the message handler code in more locations?

adityasharad: Absolutely. Just curious how this might work, no reason to block this PR.

extensions/ql-vscode/src/databases-ui.ts
adityasharad: Oh got it - `event.item` will just be `undefined` if it's not actually a `DatabaseItem`. Was thinking too much in Java!

adityasharad (dismissed):

adityasharad (approved):

adityasharad (approved):

PR #610, opened 2020-10-04
Select the appropriate node in the AST viewer when the editor text selection changes

### LINKED ISSUE #578: User should be able to navigate to the associated AST class from a node in the AST Viewer
This issue is about adding a command to allow a user to navigate to the CodeQL declaring class of an ast node in the AST Viewer.

aeisenberg:
Working locally, but need to create more tests.

Resolves #578

## Checklist

No need to add this to the changelog since this feature is not yet released.


extensions/ql-vscode/src/astViewer.ts
```diff
@@ -83,33 +96,83 @@ class AstViewerDataProvider implements vscode.TreeDataProvider<AstItem | RootAst
...
+  private updateTreeSelection(e: TextEditorSelectionChangeEvent) {
+    function isInside(selectedRange: Range, astRange?: Range): boolean {
+      return !!astRange?.contains(selectedRange);
+    }
+
+    // Recursively iterate all children until we find the node with the smallest
+    // range that contains the selection.
+    // Some nodes do not have a location, but their children might, so must
+    // recurse though location-less AST nodes to see if children are correct.
+    function findBest(selectedRange: Range, items?: RootAstItem[]): RootAstItem | undefined {
```
adityasharad, line 135: I was concerned about performance but this is pretty snappy on ~1000 LOC. Let's keep an eye on how it performs for much larger files.

extensions/ql-vscode/src/contextual/astBuilder.ts
```diff
@@ -95,8 +97,8 @@ export default class AstBuilder {
           const parent = idToItem.get(childToParent.has(id) ? childToParent.get(id)! : -1);
 
           if (parent) {
-            const astItem = item as AstItem;
-            astItem.parent = parent;
+            const astItem = item as unknown as AstItem;
```
adityasharad, line 100: Bit sad, but guess we're missing the `parent` property. Perhaps there's room to relate `AstItem` and `RootAstItem` so that you can easily convert between them (with/without parent) without having to cast to unknown, but not blocking.

adityasharad (approved):
Tested and works well. We don't seem to have file top-levels in the AST, so a selection that spans multiple top-level nodes will select the one closest to the cursor in the AST viewer, which I think is a reasonable fallback.

extensions/ql-vscode/src/astViewer.ts
aeisenberg: Yes, I considered adding a debouncing parameter and delaying navigation for that reason, but I didn't notice any issues doing it the simple way.

extensions/ql-vscode/src/contextual/astBuilder.ts
aeisenberg: Yeah, I'm not happy that I sometimes need to explicitly cast to `RootAstItem` also even when it is not necessarily a root.

Maybe I should rename `AstItem` --> `ChildAstItem` and `RootAstItem` --> `AstItem`.

adityasharad: That can work. Or make `parent` optional on `AstItem`, and `RootAstItem` override it to `undefined`.

extensions/ql-vscode/src/astViewer.ts
```diff
@@ -20,21 +20,23 @@ import { UrlValue, BqrsId } from './bqrs-cli-types';
...
   order: number;
 }
 
-export type RootAstItem = Omit<AstItem, 'parent'>;
+export interface ChildAstItem extends AstItem {
+  parent: ChildAstItem | AstItem;
+}
 
-class AstViewerDataProvider implements TreeDataProvider<AstItem | RootAstItem> {
+class AstViewerDataProvider implements TreeDataProvider<ChildAstItem | AstItem> {
```
adityasharad, line 37: Assuming this is covariant or invariant with respect to its type parameter, just `AstItem` would probably work, unless you're trying to enforce that everything may be a `ChildAstItem`.

adityasharad (approved):

extensions/ql-vscode/src/astViewer.ts
aeisenberg: Sure. I'll fix that. Will need another 👀 .

adityasharad (approved):

PR #612, opened 2020-10-08
Add command to view the DIL of a query

### LINKED ISSUE #395: View and Compare DIL in VS Code
Ensure that VSCode can view the Datalog Intermediary Language (DIL) of a query with reasonable syntax highlighting and formatting. It's likely that showing this readonly will be fine for our purposes.

Also, would be nice to see a comparison view as well, though likely, this can be delegated to VS Code's standard comparison editor.

aeisenberg:
Simple command that generates the DIL for a query if it doesn't exist and then opens it in an editor.

Fixes #395.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Fix proper escaping of backslashes in SARIF message strings.
 - Allow setting `codeQL.runningQueries.numberOfThreads` and `codeQL.runningTests.numberOfThreads` to 0, (which is interpreted as 'use one thread per core on the machine').
 - Clear the problems view of all CodeQL query results when a database is removed.
+- Add a `View DIL` command on query history items. This opens a text editor containing the Datalog Intermediary Language representation of the compiled query.
 
 ## 1.3.3 - 16 September 2020
 
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -650,6 +650,14 @@ export class CodeQLCliServer implements Disposable {
       'Resolving queries',
     );
   }
+
+  async generateDil(qloFile: string, outFile: string): Promise<void> {
+    await this.runCodeQlCliCommand(
+      ['query', 'decompile'],
+      ['--kind', 'dil', '-o', outFile, qloFile],
+      'Generating DIL',
+    );
+  }
 }
 
 /**
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -7,6 +7,7 @@ import { QueryWithResults } from './run-queries';
 import * as helpers from './helpers';
 import { logger } from './logging';
 import { URLSearchParams } from 'url';
+import { QueryServerClient } from './queryserver-client';
 
 /**
  * query-history.ts
@@ -180,6 +181,7 @@ export class QueryHistoryManager {
 
   constructor(
     ctx: ExtensionContext,
+    private qs: QueryServerClient,
     private queryHistoryConfigListener: QueryHistoryConfig,
     private selectedCallback: (item: CompletedQuery) => Promise<void>,
     private doCompareCallback: (
@@ -248,6 +250,12 @@ export class QueryHistoryManager {
       helpers.commandRunner(
         'codeQLQueryHistory.viewSarif',
         this.handleViewSarif.bind(this)
+      )
+    );
+    ctx.subscriptions.push(
+      helpers.commandRunner(
+        'codeQLQueryHistory.viewDil',
+        this.handleViewDil.bind(this)
       )
     );
     ctx.subscriptions.push(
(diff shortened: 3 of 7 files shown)


extensions/ql-vscode/src/query-history.ts
```diff
@@ -511,7 +532,9 @@ export class QueryHistoryManager {
   private async tryOpenExternalFile(fileLocation: string) {
     const uri = vscode.Uri.file(fileLocation);
     try {
-      await vscode.window.showTextDocument(uri);
+      await vscode.window.showTextDocument(uri, {
+        preview: false
```
adityasharad, line 536: Is this also used by the source archive? If so it's a behaviour change worth documenting.

extensions/ql-vscode/src/cli.ts
```diff
@@ -650,6 +650,14 @@ export class CodeQLCliServer implements Disposable {
       'Resolving queries',
     );
   }
+
+  async generateDil(qloFile: string, outFile: string): Promise<void> {
+    await this.runCodeQlCliCommand(
+      ['query', 'decompile'],
+      ['-o', outFile, qloFile],
```
adityasharad, line 657: I suggest explicitly passing `--kind dil`. We do always set `includeDilInQlo` in the compilation request so this is the default, but good to be obvious.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Fix proper escaping of backslashes in SARIF message strings.
 - Allow setting `codeQL.runningQueries.numberOfThreads` and `codeQL.runningTests.numberOfThreads` to 0, (which is interpreted as 'use one thread per core on the machine').
 - Clear the problems view of all CodeQL query results when a database is removed.
+- Add the view DIL command on a query history item that opens the Datalog Intermediary Language text in an editor.
```
adityasharad, line 18:
```suggestion
- Add the `View DIL` command on a query history item. This opens a text editor containing the Datalog Intermediary Language representation of the compiled query.
```

Keeping the wording consistent with https://help.semmle.com/codeql/codeql-cli/commands/query-decompile.html.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -169,6 +171,29 @@ export class QueryInfo {
...
   }
+
+  /**
+   * Holds if this query already has DIL produced
+   */
+  async hasDil(): Promise<boolean> {
+    return fs.pathExists(this.dilPath);
+  }
+
+  async ensureDilPath(qs: qsClient.QueryServerClient): Promise<string> {
```
adityasharad, line 182: Can we unit test this? And also whether we always send `includeDilInQlo: true` in the compilation request (otherwise this command will not be very useful)?

adityasharad (commented):
Reviewed only the final commit. 
@github/docs-content-dsp this is user-facing but aimed at advanced users. Is it worth doing a help update for this new command?

extensions/ql-vscode/src/query-history.ts
aeisenberg: This is a minor change, but my thinking is that if someone went to the trouble of opening a log file or DIL file, then they probably don't want it to be stuck in preview mode and disappear when navigating around. 

Is this a change that users would want to know about? My feeling is that this change just makes things _work_ better.  Knowing about it won't change users' behaviours. Mentioning it in docs or changelogs might just be noise. Only reason to do it IMO is if we want recognition that there is a snazzy new feature.

extensions/ql-vscode/src/cli.ts
aeisenberg: Sure

extensions/ql-vscode/CHANGELOG.md
aeisenberg: Made slight changes, but used your suggestion.

extensions/ql-vscode/src/run-queries.ts
aeisenberg: ok.

extensions/ql-vscode/src/query-history.ts
adityasharad: It makes complete sense for opening DIL (or even opening queries). I was curious about whether this handler is shared by the code that opens source archive files -- there it might become quite cluttered if we open every source link in a permanent tab. (I personally would not mind, but it's easy to click on a lot of result links.)

aeisenberg: Oh...I see what you're asking. No. This code path is only used for commands coming from query history. So, user needs to right-click and actively try to see this file.

aeisenberg:
Ping @github/docs-content-dsp! This is a new feature in the vscode extension. It allows users to view the DIL of a compiled query. It is accessible from the query history view and meant for advanced users.

felicitymay:
👋🏻 - thanks for the pings. Not sure why only the last one was picked up by the docs content team's bot.

I expect that other member of the DSP writing team might know the answer, but do we have any guidance on how to use the DIL view to debug queries?

Even if we don't say much about this new option in the main docs, it might be worth adding a brief mention to the glossary entry for DIL: https://help.semmle.com/codeql/advanced-glossary.html?#dil

felicitymay:
Let me know what you think about this suggestion ☝🏻 and I'll raise an issue, if this is appropriate.

adityasharad:
Thanks @felicitymay. We do not have a detailed public guide on debugging with DIL. I think the following should do for now:
- update the glossary entry you linked for DIL, to mention the View DIL command in the VS Code extension
- add a note towards the end of https://help.semmle.com/codeql/codeql-for-vscode/procedures/using-extension.html, mentioning the View DIL command on the query history view for debugging.

felicitymay:
https://github.com/github/docs-content/issues/3079 created for the docs update.

Do you have an idea of when you plan to ship this change?

aeisenberg:
> Do you have an idea of when you plan to ship this change?

I'd like to ship it some time mid-week, but I'm a bit flexible if you need more time.

adityasharad (approved):

PR #617, opened 2020-10-14
Add open type support from AST Viewer

### LINKED ISSUE #614: Add command to open AST node type declaration in AST Viewer
When a user selects an AST Node in the AST Viewer, there should be a way to open the underlying QL type declaration.

aeisenberg:
**Note** This is still a draft. The UX here is off. When you invoke the Open AST Type command, you are navigated to the search bar and you lose your place in the AST Viewer. I'm not a fan of this. However, it might be the case that even with this poor UX, we are better served having the feature than not.

We make the assumption that AST items whose label start with a bracketed 
name refer to a Code QL AST type.

We would like to use the language server's open type command, but that 
requires a file URI and a text location in that file.

Instead, we use a file search. This is an experimental feature. Let's 
see if it is useful.

Fixes #614. This is an alternative implementation that gets us part-way to the desired goal and maybe it is sufficient for now.

## Checklist

- [ ] `@github/docs-content-dsp` has been cc'd in all issues for UI or other user-facing changes made by this pull request.

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -366,6 +366,10 @@
         "title": "Go To Code"
       },
       {
+        "command": "codeQLAstViewer.openType",
+        "title": "Find AST Type"
+      },
+      {
         "command": "codeQLAstViewer.clear",
         "title": "Clear AST",
         "icon": {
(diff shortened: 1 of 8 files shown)


aeisenberg:
This approach is not going to work. Closing this off until we get a better idea. Likely we are going to have to update the Query server to get this working.

PR #624, opened 2020-10-19
Restructure the tree in the Test Explorer View

### LINKED ISSUE #595: QL Test Explorer should discover tests via `codeql resolve tests`
The current implementation of Test Explorer finds QL tests by looking for packs from the current workspace that are available on the CLI's library path (usually via `.codeqlmanifest.json`), then looking for tests within those packs. This misses tests in packs that aren't specified in `.codeqlmanifest.json`, even though packs containing only queries and tests, or even packs containing only tests, have no other reason to be specified in the manifest.

Instead of searching for registered packs, the Test Explorer should just invoke `codeql resolve tests` from the root of the workspace.

**Additional context**
A related concern exists about how the Test Explorer tree displays the tests. It currently has one top-level node for each pack, which will then expand to a directory tree containing the individual tests and test directories. It has been suggested that it would be better to have the Test Explorer tree show the actual directory tree, similar to the file Explorer tree. Directories containing no tests would not be shown, and any directory containing only a single other test directory would be combined with the contained test directory:

Original file system:
```
root
+ A
  + A1
    - test_a1.ql
  + A2
    - test_a2.ql
+ B // No tests under here
  + B1
    + foo.txt
+ C  // Only one subdirectory contains tests
  + C1  // Only one subdirectory contains tests
    + D
      - test_d1.ql
      - test_d2.ql
```

Test Explorer tree:
```
root
+ A
  + A1
    - test_a1
  + A2
    - test_a2
+ C/C1/D
  - test_d1.ql
  - test_d2.ql
```

If the above would be a better way to display the tests, it might be worth switching to that when implementing the `codeql resolve tests` code, rather than having to figure out how to map the output of `codeql resolve tests` back to packs.

aeisenberg:
With this change we display the tree based on the file system not based
on ql-packs. We also merge test folders whose only child is another
test folder.

Resolves #595

## Checklist


extensions/ql-vscode/src/qltest-discovery.ts
```diff
@@ -55,10 +54,23 @@ export class QLTestDirectory extends QLTestNode {
...
       child.finish();
-    }
+      if (child.children?.length === 1 && child.children[0] instanceof QLTestDirectory) {
+        // collapse children
+        const replacement = new QLTestDirectory(
+          child.children[0].path,
+          child.name + ' / ' + child.children[0].name,
+          Array.from(child.children[0].children)
+        );
+        this._children[i] = replacement;
```
adityasharad, line 71: This kind of concurrent modification is fine because we're only modifying the current item?

extensions/ql-vscode/src/qltest-discovery.ts
```diff
@@ -98,12 +110,12 @@ interface QLTestDiscoveryResults {
   /**
    * The root test directory for each QL pack that contains tests.
```
adityasharad, line 111: Update comment please :)

extensions/ql-vscode/src/qltest-discovery.ts
```diff
@@ -98,12 +110,12 @@ interface QLTestDiscoveryResults {
...
    */
-  testDirectories: QLTestDirectory[];
+  testDirectory: QLTestDirectory | undefined;
   /**
    * The list of file system paths to watch. If any of these paths changes, the discovery results
```
adityasharad, line 115: Update comment please :)

extensions/ql-vscode/src/qltest-discovery.ts
```diff
@@ -128,12 +140,16 @@ export class QLTestDiscovery extends Discovery<QLTestDiscoveryResults> {
   /**
    * Event to be fired when the set of discovered tests may have changed.
    */
-  public get onDidChangeTests(): Event<void> { return this._onDidChangeTests.event; }
+  public get onDidChangeTests(): Event<void> {
+    return this._onDidChangeTests.event;
+  }
 
   /**
    * The root test directory for each QL pack that contains tests.
```
adityasharad, line 148: Update - I assume this is just the workspace folder, viewed as a `QLTestDirectory`.

extensions/ql-vscode/src/qltest-discovery.ts
```diff
@@ -146,70 +162,43 @@ export class QLTestDiscovery extends Discovery<QLTestDiscoveryResults> {
...
-        const testPackage = await this.discoverTests(qlPack.uri.fsPath, qlPack.name);
-        if (testPackage !== undefined) {
-          testDirectories.push(testPackage);
-        }
-      }
-    }
-
-    return { testDirectories, watchPaths };
+    return {
+      testDirectory: await this.discoverTests(),
```
adityasharad, line 166: We now unconditionally run test discovery on all workspace folders. How does this scale/perform when we have a large workspace with many folders, or a single large repo (like our internal codebase)? Does it also attempt to search the source archives that we add to the workspace in the extension, and if so can we skip them?

aeisenberg (commented):
Oh...never submitted these comments from days ago.

adityasharad:
Some outstanding doc comments, and changelog conflicts.

aeisenberg:
Update here. I found a bug in how discovery works. QLTest discovery runs twice at startup. The reason is this:

1. QLPack discovery starts
2. QLTest discovery starts
3. When QLPack discovery finishes, it triggers a new discovery for QLTest. Even though QLTest discovery just finished.

This is the case with the `main` branch right now. In the new setup we don't rely on QLPacks at all. It can be completely removed. And code is significantly simplified.

Now the bad news. Each initial discovery operation in the new way (from the root of the workspace folder takes between 500-1500ms. Under the current way, where we do discovery from the roots of the test qlpacks, initial discovery takes between 100-200ms. 

For whatever reason, subsequent discoveries under the new way are quicker (700-800ms) and slower under the old way (400-500ms). This data was not captured very scientifically.

I personally think that this difference in time is not a deal breaker. Initial discovery being slow is less important since that happens during other initialization so, a little slowness for tests is not very noticeable. And the fact that it takes about 2x longer for subsequent discoveries is not noticeable either since this happens during a save, while editing.

With this in mind, I will refactor to remove the double discovery operations. This will mean that qlpack discovery will be removed.

aeisenberg:
Ping! @github/docs-content-dsp This PR changes the layout structure of the QL test tree. Before this PR, the test tree was scoped by workspace folders, then by ql packs in those folders. 

The new layout that the ql test tree matches the filesystem. This makes it easier to map tests back to their location in the filesystem.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 - Add a `View DIL` command on query history items. This opens a text editor containing the Datalog Intermediary Language representation of the compiled query.
 - Remove feature flag for the AST Viewer. For more information on how to use the AST Viewer, [see the documentation](https://help.semmle.com/codeql/codeql-for-vscode/procedures/exploring-the-structure-of-your-source-code.html).
 - The `codeQL.runningTests.numberOfThreads` setting is now used correctly when running tests.
+- Alter structure of the _Test Explorer_ tree. It now follows the structure of the filesystem instead of the qlpacks.
```
shati-patel, line 31:
```suggestion
- Alter structure of the _Test Explorer_ tree. It now follows the structure of the filesystem instead of the QL packs.
```

shati-patel (commented):
> Ping! @github/docs-content-dsp This PR changes the layout structure of the QL test tree. Before this PR, the test tree was scoped by workspace folders, then by ql packs in those folders.
> 
> The new layout that the ql test tree matches the filesystem. This makes it easier to map tests back to their location in the filesystem.

Thanks, makes sense! We don't go into detail about the structure of the Test Explorer view in the [docs](https://help.semmle.com/codeql/codeql-for-vscode/procedures/testing-queries.html), so this doesn't need explicit documentation 😃

adityasharad (approved):

extensions/ql-vscode/CHANGELOG.md
adityasharad: Whoops I think this suggestion got omitted.

PR #626, opened 2020-10-20
Add more structured output for tests

### LINKED ISSUE #582: Debugging broken QL unit tests is difficult
The VSCode console does not show you where an error is located. The error location is only available in the log file. 

The error location should be printed to the output console.

aeisenberg:
The diff and the errors were always available, but they were not being sent to the output.

Additionally, make sure to send output to both the test explorer log and the codeql test log.

Resolves #582.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/test-adapter.ts
+++ b/extensions/ql-vscode/src/test-adapter.ts
@@ -217,10 +217,26 @@ export class QLTestAdapter extends DisposableObject implements TestAdapter {
       cancellationToken: cancellationToken,
       logger: testLogger
     })) {
+      const state = event.pass
+        ? 'passed'
+        : event.messages?.length
+        ? 'errored'
+        : 'failed';
+      let message: string | undefined;
+      if (event.diff?.length) {
+        message = ['', `${state}: ${event.test}`, ...event.diff, ''].join('\n');
+        testLogger.log(message);
+      }
+      (event.diff || []).join('\n');
       this._testStates.fire({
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/test-adapter.ts
```diff
@@ -221,10 +221,26 @@ export class QLTestAdapter extends DisposableObject implements TestAdapter {
       cancellationToken: cancellationToken,
       logger: testLogger
     })) {
+      const state = event.pass
+        ? 'passed'
+        : event.messages?.length
+        ? 'errored'
```
adityasharad, line 227: Would `failed due to errors` or `failed to run` be more meaningful for the user? I had to think about the difference between 'errored' and 'failed' for a moment.

adityasharad (commented):
Should be good to go after a rebase.

extensions/ql-vscode/src/test-adapter.ts
aeisenberg: That won't work. `state` is a property of the `TestEvent` interface and only allows `'running' | 'passed' | 'failed' | 'skipped' | 'errored'`. The hope is that the messages will be more explicit as to what happened.

adityasharad: Oh fair enough. We can consider having more detail in the way `message` is constructed below, but no need to block this PR.

adityasharad (approved):

PR #638, opened 2020-10-22
Remove the --kind dil options from decompile command

aeisenberg:
This option is not recognized. I had added this option earlier and not sufficiently tested it.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -673,7 +673,7 @@ export class CodeQLCliServer implements Disposable {
   async generateDil(qloFile: string, outFile: string): Promise<void> {
     await this.runCodeQlCliCommand(
       ['query', 'decompile'],
-      ['--kind', 'dil', '-o', outFile, qloFile],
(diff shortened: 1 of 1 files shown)


hmakholm:
> This option is not recognized

Do you mean that the CLI fails to accept it?

Note that we'll only be able to produce the DIL if it has been saved during the compilation process (which unfortunately means existing compilation cache entries cannot be reused).

adityasharad:
Recording discussion from Slack: the `--kind` flag was added in CLI 2.3.0, and Andrew was on 2.2.5.
We can bump the extension's expected version number, but do need to be more robust at handling old CLI versions, since customers may be using them for LGTM Enterprise compatibility. So for now I suggest we conditionally add the `--kind` flag based on the detected CLI version.

adityasharad:
Okay as discussed, let's get this in for now (since we always compile with DIL included in the extension), and we'll make the version handling smarter in future.

adityasharad (approved):

aeisenberg:
+1 and this is a good example of where we need more integration tests.

PR #639, opened 2020-10-22
Use single file upgrades where possible.

alexet:
Where possible use the new single file form of upgrades. Older versions of `codeql` simply ignore the `singleFileUpgrades` property and give multifile upgrades. Hence we can just set the new flag and see which sort of upgrade we get pack. If a version of codeql returns a multi file upgrade then it certainly supports running them so there are no issues there.

### DIFF
--- a/extensions/ql-vscode/src/pure/messages.ts
+++ b/extensions/ql-vscode/src/pure/messages.ts
@@ -401,6 +401,11 @@ export interface CompileUpgradeParams {
    * A directory to store parts of the compiled upgrade
    */
   upgradeTempDir: string;
+  /** 
+   * Enable single file upgrades, set to true to allow
+   * using single file upgrades.
+   */
+  singleFileUpgrades?: boolean;
 }
 
 /**
@@ -487,10 +492,42 @@ export interface UpgradeDescription {
   newSha: string;
 }
 
+
+export type CompiledUpgrades = MultiFileCompiledUpgrades | SingleFileCompiledUpgrade
+
 /**
- * A compiled upgrade.
+ * A compiled upgrade. 
+ * The upgrade is spread among multiple files.
  */
-export interface CompiledUpgrades {
+export interface MultiFileCompiledUpgrades {
+  /**
+  * The initial sha of the dbscheme to upgrade from
+  */
+  initialSha: string;
+  /**
+   * The path to the new dataset statistics
+   */
+  newStatsPath: string;
+  /**
+   * The path to the new dataset dbscheme
+   */
+  newDbscheme: string;
+  /**
+   * The steps in the upgrade path
+   */
+  scripts: CompiledUpgradeScript[];
+  /**
+   * Will never exist in an old result
+   */
+  compiledUpgradeFile?: never;
+  /**
+   * The sha of the target dataset.
+   */
+  targetSha: string;
+}
+
+
+export interface SingleFileCompiledUpgrade {
   /**
   * The initial sha of the dbscheme to upgrade from
   */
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/upgrades.ts
```diff
@@ -150,7 +150,11 @@ export async function upgradeDatabase(
 
   try {
     qs.logger.log('Running the following database upgrade:');
-    qs.logger.log(compileUpgradeResult.compiledUpgrades.scripts.map(s => s.description.description).join('\n'));
+    if (compileUpgradeResult.compiledUpgrades.compiledUpgradeFile === undefined) {
+      qs.logger.log(compileUpgradeResult.compiledUpgrades.scripts.map(s => s.description.description).join('\n'));
+    } else {
+      qs.logger.log(compileUpgradeResult.compiledUpgrades.descriptions.map(s => s.description).join('\n'));
+    }
```
aeisenberg, line 157: nit: Can you restructure the if expression? I find it hard to reason about. Something like this I find easier to read

```suggestion
    if (compileUpgradeResult.compiledUpgrades.compiledUpgradeFile) {
      // Single file upgrade
      qs.logger.log(compileUpgradeResult.compiledUpgrades.descriptions.map(s => s.description).join('\n'));
    } else {
      qs.logger.log(compileUpgradeResult.compiledUpgrades.scripts.map(s => s.description.description).join('\n'));
    }
```

aeisenberg (approved):
My suggestion is not a blocker for going in, but I think it would make it easier to read.

extensions/ql-vscode/src/upgrades.ts
alexet: Is you suggestion mostly about switching the sense of the test or about the `=== undefined`.

It has to have the `=== undefined` part to get the types correct (according to the types we could have an empty `compiledUpgradeFile`). However I can flip the sense of the test if that makes more sense to you.

aeisenberg: I see why typing is broken if you avoid the `undefined`. I think just adding a comment would be sufficient for me.

extensions/ql-vscode/src/messages.ts
```diff
@@ -487,10 +492,14 @@ export interface UpgradeDescription {
   newSha: string;
 }
 
+
+export type CompiledUpgrades = MultiFileCompiledUpgrades | SingleFileCompiledUpgrade
```
aeisenberg, line 496: And looking at this again, it's not easily clear that these two types differ in only two properties. In this case, I wonder if using proper inheritance would make things clearer. So, something like this:

```js
export interface CompiledUpgrades {
  // add all common properties here
}

interface MultiFileCompiledUpgrades extends CompiledUpgrades {
  newDbscheme: string;
  scripts: CompiledUpgradeScript[];
  compiledUpgradeFile?: never;
}

interface SingleFileCompiledUpgrade extends CompiledUpgrades {
  descriptions: UpgradeDescription[];
  compiledUpgradeFile: string;
}
```

Also. note that I don't think you need to export these types.

aeisenberg:
Converting this to a draft. Please mark as "ready for review" when you want me to look at it.

alexet:
I think this is ready at the moment.

aeisenberg (commented):
Looks good. Possible small improvement below.

aeisenberg:
I've tested it with both an old codeql cli and a new one. Behaviour works as expected. :shipit:

extensions/ql-vscode/src/pure/messages.ts
```diff
@@ -401,6 +401,11 @@ export interface CompileUpgradeParams {
    * A directory to store parts of the compiled upgrade
    */
   upgradeTempDir: string;
+  /** 
+   * Enable single file upgrades, set to true to allow
+   * using single file upgrades.
+   */
+  singleFileUpgrades?: boolean;
```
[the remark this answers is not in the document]
aeisenberg: Made this change.

aeisenberg:
After this build passes, I will merge.

PR #644, opened 2020-10-23
Add version info to cli

aeisenberg:
And also only add the `--kind=DIL` to `generateDil` if version is
`>= 2.3.0`.

Integration tests will come next.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -4,15 +4,16 @@ import * as child_process from 'child_process';
 import * as fs from 'fs-extra';
 import * as path from 'path';
 import * as sarif from 'sarif';
+import { SemVer } from 'semver';
 import { Readable } from 'stream';
 import { StringDecoder } from 'string_decoder';
 import * as tk from 'tree-kill';
-import * as util from 'util';
+import { promisify } from 'util';
 import { CancellationToken, Disposable } from 'vscode';
 
 import { BQRSInfo, DecodedBqrsChunk } from './bqrs-cli-types';
 import { CliConfig } from './config';
-import { DistributionProvider } from './distribution';
+import { DistributionProvider, FindDistributionResultKind } from './distribution';
 import { assertNever } from './helpers-pure';
 import { QueryMetadata, SortDirection } from './interface-types';
 import { Logger, ProgressReporter } from './logging';
@@ -115,6 +116,11 @@ interface BqrsDecodeOptions {
  */
 export class CodeQLCliServer implements Disposable {
 
+  /**
+   * CLI version where --kind=DIL was introduced
+   */
+  private static CLI_VERSION_WITH_DECOMPILE_KIND_DIL = new SemVer('2.3.0');
+
   /** The process for the cli server, or undefined if one doesn't exist yet */
   process?: child_process.ChildProcessWithoutNullStreams;
   /** Queue of future commands*/
@@ -123,6 +129,12 @@ export class CodeQLCliServer implements Disposable {
   commandInProcess: boolean;
   /**  A buffer with a single null byte. */
   nullBuffer: Buffer;
+
+  /** Version of current cli, lazily computed by the `getVersion()` method */
+  _version: SemVer | undefined;
+
+  /** Path to current codeQL executable, or undefined if not running yet. */
+  codeQlPath: string | undefined;
 
   constructor(
     private distributionProvider: DistributionProvider,
@@ -140,10 +152,10 @@ export class CodeQLCliServer implements Disposable {
     if (this.cliConfig.onDidChangeConfiguration) {
       this.cliConfig.onDidChangeConfiguration(() => {
         this.restartCliServer();
+        this._version = undefined;
       });
     }
   }
-
 
   dispose(): void {
     this.killProcessIfRunning();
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -115,6 +111,11 @@ interface BqrsDecodeOptions {
  */
 export class CodeQLCliServer implements Disposable {
 
+  /**
+   * CLI version where --format=DIL was introduced
+   */
+  private static DIL_VERSION = new SemVer('2.3.0');
```
adityasharad, line 117: Nit: This made me think of the optimiser version and confused me. Perhaps something more boring and explicit, like `CLI_VERSION_WITH_DECOMPILE_KIND_DIL`? Also update comment to mention `--kind` instead of `--format`.

extensions/ql-vscode/src/cli.ts
```diff
@@ -210,15 +216,23 @@ export class CodeQLCliServer implements Disposable {
...
       ['execute', 'cli-server'],
       [],
       this.logger,
       _data => { /**/ }
     );
+
+    if (codeQlPath !== this.codeQlPath) {
+      // path has changed. Update the version info
+      this.codeQlPath = codeQlPath;
+      this.version = await getCodeQlCliVersion(codeQlPath, this.logger);
```
adityasharad, line 232: I see why this works, but my instinct is that we already have this information, and we don't need to recompute it:

Specifically, `this.getCodeQlPath()` could be made more general, so that it returns the entire `Distribution` object instead of just the path string. The `DistributionProvider` here already knows the CLI version, and whether it is compatible with the extension, it just needs to expose that information as part of the interface instead of being hidden in `DistributionManager`.
Then, instead of making another CLI call to get the version here, we just compare the property.

extensions/ql-vscode/src/cli-version.ts
```diff
@@ -22,3 +30,39 @@ export async function getCodeQlCliVersion(codeQlPath: string, logger: Logger): P
...
+ * Runs a CodeQL CLI command without invoking the CLI server, returning the output as a string.
+ * @param codeQlPath The path to the CLI.
+ * @param command The `codeql` command to be run, provided as an array of command/subcommand names.
+ * @param commandArgs The arguments to pass to the `codeql` command.
+ * @param description Description of the action being run, to be shown in log and error messages.
+ * @param logger Logger to write command log messages, e.g. to an output channel.
+ * @param progressReporter Used to output progress messages, e.g. to the status bar.
+ * @returns The contents of the command's stdout, if the command succeeded.
+ */
+export async function runCodeQlCliCommand(
```
adityasharad, line 45: Nit: I see that this and `LOGGING_FLAGS` are single-use, but keeping them in `cli` might still make sense since they aren't really specific to CLI versioning. That said, the name is confusing because there is an identical one within `CliServer`, so we could change it to make it more explicit that this variation is standalone / without a server.

adityasharad (changes requested):

extensions/ql-vscode/src/cli.ts
aeisenberg: That's not always true. In `ExtensionSpecificDistributionManager.getCodeQlPathWithoutVersionCheck`, the path is grabbed by joining a few strings and directory names. There's no version info there that I can find. Unless the version is hidden in the directory somewhere, I don't know how we can do it.

extensions/ql-vscode/src/cli-version.ts
aeisenberg: I moved this to `cli-version` in order to avoid a circular dependency. `getCodeQlCliVersion` is now used inside of `cli`.

We could rename this file to `serverless-cli.ts`. Maybe that would make more sense here.

extensions/ql-vscode/src/cli.ts
aeisenberg: That's a mouthful, but is fine.

extensions/ql-vscode/src/cli.ts
adityasharad: I see, since we've deliberately omitted the version check in that case. Any chance we can call `DistributionManager.getDistribution() instead`? That already knows how to look up the version.

aeisenberg: We can do that. Though, internally, `DistributionManager.getDistribution()` uses `getCodeQlCliVersion`. It's maybe a bit cleaner this way. Implemented, tested locally and pushed.

extensions/ql-vscode/src/cli-version.ts
aeisenberg: In this next commit, I undid that move since I am now using the `DistributionManager` to get the version, and the circular dependency is not longer happening.

adityasharad (approved):
Neat, thank you!

extensions/ql-vscode/test/pure-tests/query-test.ts
```diff
@@ -10,6 +10,7 @@ import * as qsClient from '../../src/queryserver-client';
 import * as cli from '../../src/cli';
 import { ProgressReporter, Logger } from '../../src/logging';
 import { ColumnValue } from '../../src/bqrs-cli-types';
+import { FindDistributionResultKind } from '../../src/distribution';
```
adityasharad, line 13: Oh I think this change introduces a test dependency on `vscode`, which `pure-tests` have to avoid.

aeisenberg: I'm wondering if this distinction is still worth it. If it's not easy to remove this dependency, then I may move `query-test.ts` to the integration test.

aeisenberg: And thanks for tracking that down. I was about to look into this myself.

PR #645, opened 2020-10-26
Add integration tests with the CLI

### LINKED ISSUE #642: Run integration tests against multiple versions of CodeQL
Two areas to consider:
- We currently test against only the `latest` CLI release. This should be expanded to either all previous supported CLI versions, or at least to all CLI versions applicable to LGTM Enterprise releases (since we expect some users to be still on those versions).
- We currently test against only the full `codeql.zip`, but the extension itself actually downloads and uses the platform-specific artifacts. I think we should test against all possible artifacts.

We can do this in parallel by including the CLI version in the Actions matrix, and updating the GitHub API calls.

aeisenberg:
Note that even though this PR is labeled as `Complexity: High`, it is not particularly dangerous in that most of the complexity is around how the tests run.

This commit adds integration tests that run commands using the CLI. This
change introduces a number of enhancements in order to get there.

1. Augments the index-template.ts file so that it downloads an appropriate cli version if requested.
2. Adds the ensureCli.ts that performs the download if a a suitable version is not already installed. See the comments in the file for how this is done.
3. Changes how run-integration-tests is done so that the directories run are specified through a cli argument.
4. Updates the main.yml workflow so that it also runs the cli-integration tests.
5. Takes advantage of the return value of the call to `activate` on the extension. This allows the integration tests to have access to internal variables of the extension like the context, cli, and query server.
5. And of course, adds a handful of simple tests that ensure we have a cli installed of the correct version.

Resolves #642.

## Checklist


aeisenberg:
OK...I'm making progress here. Tests are still failing, but now it is a _different_ failure. Uncovered by fixing the first failure.

.github/workflows/main.yml
```diff
@@ -20,17 +20,17 @@ jobs:
 
       - uses: actions/setup-node@v1
         with:
-          node-version: '10.18.1'
+          node-version: '14.14.0'
```
aeisenberg, line 23: Move to latest version of node. I thought earlier this was causing errors in the builds, but it wasn't. Though, may as well move to a later version.

.github/workflows/main.yml
```diff
@@ -91,27 +90,71 @@ jobs:
...
-          cd extensions/ql-vscode
           npm run integration
+
+  cli-test:
+    name: CLI Test
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        version: ['v2.3.1', 'v2.3.0']
```
aeisenberg, line 124: We can choose as many versions as we like. Though, my suggestion is to use the latest as well as one for each supported version of LGTM.

extensions/ql-vscode/package.json
```diff
@@ -687,7 +687,8 @@
     "watch:extension": "tsc --watch",
     "test": "mocha --exit -r ts-node/register test/pure-tests/**/*.ts",
     "preintegration": "rm -rf ./out/vscode-tests && gulp",
-    "integration": "node ./out/vscode-tests/run-integration-tests.js",
+    "integration": "node ./out/vscode-tests/run-integration-tests.js no-workspace,minimal-workspace",
```
aeisenberg, line 690: Note: must pass in the directories to test as part of the argument.

extensions/ql-vscode/src/distribution.ts
```diff
@@ -49,16 +49,38 @@ export interface DistributionProvider {
...
 
 export class DistributionManager implements DistributionProvider {
-  constructor(extensionContext: ExtensionContext, config: DistributionConfig, versionRange: semver.Range) {
-    this._config = config;
-    this._extensionSpecificDistributionManager = new ExtensionSpecificDistributionManager(extensionContext, config, versionRange);
+
+  /**
+   * Get the name of the codeql cli installation we prefer to install, based on our current platform.
+   */
+  public static getRequiredAssetName(): string {
```
aeisenberg, line 56: Made static so we can use this in our integration tests.

extensions/ql-vscode/src/distribution.ts
```diff
@@ -49,16 +49,38 @@ export interface DistributionProvider {
...
+      case 'win32':
+        return 'codeql-win64.zip';
+      default:
+        return 'codeql.zip';
+    }
+  }
+
+  constructor(
+    public readonly config: DistributionConfig,
+    private readonly versionRange: semver.Range,
```
aeisenberg, line 71: Remove the leading `_` from these variables just because it is not necessary.

extensions/ql-vscode/src/distribution.ts
```diff
@@ -181,8 +203,8 @@ export class DistributionManager implements DistributionProvider {
   public async checkForUpdatesToExtensionManagedDistribution(
     minSecondsSinceLastUpdateCheck: number): Promise<DistributionUpdateCheckResult> {
     const distribution = await this.getDistributionWithoutVersionCheck();
-    const extensionManagedCodeQlPath = await this._extensionSpecificDistributionManager.getCodeQlPathWithoutVersionCheck();
-    if (distribution?.codeQlPath !== extensionManagedCodeQlPath) {
+    const extensionManagedCodeQlPath = await this._extensionSpecificDistributionManager!.getCodeQlPathWithoutVersionCheck();
```
aeisenberg, line 206: This type assertion may not be necessary any more. Let me try to remove it.

extensions/ql-vscode/src/distribution.ts
```diff
@@ -202,7 +224,7 @@ export class DistributionManager implements DistributionProvider {
    */
   public installExtensionManagedDistributionRelease(release: Release,
     progressCallback?: helpers.ProgressCallback): Promise<void> {
-    return this._extensionSpecificDistributionManager.installDistributionRelease(release, progressCallback);
+    return this._extensionSpecificDistributionManager!.installDistributionRelease(release, progressCallback);
```
aeisenberg, line 227: And here...

extensions/ql-vscode/src/distribution.ts
```diff
@@ -576,7 +584,7 @@ export async function extractZipArchive(archivePath: string, outPath: string): P
   }));
 }
 
-function codeQlLauncherName(): string {
+export function codeQlLauncherName(): string {
```
aeisenberg, line 587: Exported so we can use in integration tests.

extensions/ql-vscode/src/extension.ts
```diff
@@ -10,7 +10,7 @@ import {
   Uri,
   window as Window,
   env,
-  window
+  window,
```
aeisenberg, line 13: remove

extensions/ql-vscode/src/extension.ts
```diff
@@ -253,29 +274,30 @@ export async function activate(ctx: ExtensionContext): Promise<void> {
...
-  async function installOrUpdateThenTryActivate(config: DistributionUpdateConfig): Promise<void> {
+  async function installOrUpdateThenTryActivate(config: DistributionUpdateConfig): Promise<CodeQLExtensionInterface | {}> {
     await installOrUpdateDistribution(config);
 
     // Display the warnings even if the extension has already activated.
     const distributionResult = await getDistributionDisplayingDistributionWarnings();
 
     if (!beganMainExtensionActivation && distributionResult.kind !== FindDistributionResultKind.NoDistribution) {
-      await activateWithInstalledDistribution(ctx, distributionManager);
+      return await activateWithInstalledDistribution(ctx, distributionManager);
```
aeisenberg, line 284: Not a big fan of having return statements hidden in the middle of the function. Try to only have one return statement at the end.

extensions/ql-vscode/src/helpers.ts
```diff
@@ -121,7 +121,8 @@ export function commandRunner(
 ): Disposable {
   return commands.registerCommand(commandId, async (...args: any[]) => {
     try {
-      await task(...args);
+      const res = await task(...args);
+      return res;
```
aeisenberg, line 125: Combine

extensions/ql-vscode/src/query-history.ts
```diff
@@ -59,16 +59,15 @@ interface QueryHistoryDataProvider extends vscode.TreeDataProvider<CompletedQuer
 /**
  * Tree data provider for the query history view.
  */
-class HistoryTreeDataProvider implements QueryHistoryDataProvider {
+class HistoryTreeDataProvider extends DisposableObject implements QueryHistoryDataProvider {
```
aeisenberg, line 62: One issue we were having during integration testing was the fact that some things were not properly being disposed. I went through and made sure more things had proper disposal. So, you'll see that spread throughout this commit.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -59,16 +59,15 @@ interface QueryHistoryDataProvider extends vscode.TreeDataProvider<CompletedQuer
...
   /**
    * XXX: This idiom for how to get a `.fire()`-able event emitter was
```
aeisenberg, line 64: Get rid of this comment. It is a fairly standard idiom throughout all of vscode.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -135,7 +135,7 @@ class HistoryTreeDataProvider implements QueryHistoryDataProvider {
     return this.current;
   }
 
-  push(item: CompletedQuery): void {
+  pushQuery(item: CompletedQuery): void {
```
aeisenberg, line 138: Need to rename so as not to override the `super.push()` method which is entirely different.

extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  afterEach(() => {
+    cli.dispose();
+  });
+
+  it('should have the correct version of the cli', async () => {
+    expect(
+      (await cli.getVersion()).toString()
+    ).to.eq(
+      new SemVer(process.env.CLI_VERSION || '').toString()
```
aeisenberg, line 40: This env variable is set in the `ensureCli.ts` module below.

extensions/ql-vscode/src/vscode-tests/no-workspace/astViewer.test.ts
```diff
@@ -15,19 +15,25 @@ const expect = chai.expect;
 
 describe('AstViewer', () => {
   let astRoots: AstItem[];
-  let viewer: AstViewer;
+  let viewer: AstViewer | undefined;
+  let sandbox: sinon.SinonSandbox;
   beforeEach(async () => {
+    sandbox = sinon.createSandbox();
```
aeisenberg, line 21: Use `sinon.sandbox` where possible. This ensures all stubs and mocks are cleaned up when `sandbox.restore()` is called. One problem with the integration tests earlier on was that mocks were not being properly restored and we were seeing confusing errors across tests.

extensions/ql-vscode/src/vscode-tests/no-workspace/distribution.test.ts
```diff
@@ -252,8 +252,12 @@ describe('Launcher path', () => {
     expect(result).to.equal(undefined);
   });
 
-  it('should not warn when deprecated launcher is used, but no new launcher is available', async () => {
-    const manager = new (createModule().DistributionManager)(undefined as any, { customCodeQlPath: pathToCmd } as any, undefined as any);
+  it('should not warn when deprecated launcher is used, but no new launcher is available', async function() {
+    const manager = new (createModule().DistributionManager)(
+      { customCodeQlPath: pathToCmd } as any,
```
aeisenberg, line 257: `DistributionManager` has changed. So, need to update these tests.

extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
```diff
@@ -58,34 +69,33 @@ async function runTestsWithRetryOnSegfault(suite: Suite, tries: number): Promise
...
-          path.resolve(__dirname, '../../test/data'),
-          '--disable-extensions',
-        ]
-      }
-    ];
 
-    for (const integrationTestSuite of integrationTestSuites) {
-      await runTestsWithRetryOnSegfault(integrationTestSuite, 3);
+    // Which tests to run. Use a comma-separated list of directories.
+    const testDirsString = process.argv[2];
```
aeisenberg, line 77: See `package.json` for where this is used.

extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
```diff
@@ -94,3 +104,25 @@ async function main() {
 }
 
 main();
+
+
+function getLaunchArgs(dir: TestDir) {
```
aeisenberg, line 109: This was uglier than I had hoped. `'--disable-extensions'` needs to be called for minimal and no workspace runs otherwise ambient extensions get in the way of the run. This _may_ not be a problem on the build server, but it is definitely a problem locally.

extensions/ql-vscode/test/pure-tests/command-lint.test.ts
```diff
@@ -37,17 +37,20 @@ describe('commands declared in package.json', function() {
...
       command.match(/^codeQLDatabases\./)
       || command.match(/^codeQLQueryHistory\./)
       || command.match(/^codeQLAstViewer\./)
     ) {
       scopedCmds.add(command);
       expect(title).not.to.be.undefined;
       commandTitles[command] = title!;
-    }
-    else {
+    } else if (
```
aeisenberg, line 49: This change is no longer necessary.

aeisenberg (commented):
Added a bunch of comments that help explain things in the commit. Some of the comments are reminders for myself to change things later.

extensions/ql-vscode/src/extension.ts
```diff
@@ -113,16 +113,39 @@ function registerErrorStubs(excludedCommands: string[], stubGenerator: (command:
...
+ * Returns the CodeQLExtensionInterface, or an empty object if the interface is not
+ * available afer activation is complete. This will happen if there is no cli
+ * installed when the extension starts. Downloading and installing the cli
+ * will happen at a later time.
+ *
+ * @param ctx The extension context
+ *
+ * @returns CodeQLExtensionInterface
+ */
+export async function activate(ctx: ExtensionContext): Promise<CodeQLExtensionInterface | {}> {
```
adityasharad, line 139: Just making sure: extension initialisation in VS Code doesn't require a `Promise<void>` return type?

extensions/ql-vscode/src/query-history.ts
adityasharad: Still planning to do this?

extensions/ql-vscode/src/vscode-tests/cli-integration/global.helper.ts
```diff
@@ -0,0 +1,67 @@
...
+        });
+      } catch (e) {
+        fail('Failed to download test database: ' + e);
+      }
+    }
+  },
+
+  async () => {
+    // Set the CLI version here before activation to ensure we don't accidentally try to download a cli
+    await workspace.getConfiguration().update('codeQL.cli.executablePath', process.env.CLI_PATH, ConfigurationTarget.Global);
```
adityasharad, line 49: Good call. Would it make sense to also test picking up a CLI from the PATH? (Perhaps we have that tested elsewhere.)

extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
```diff
@@ -1,16 +1,17 @@
...
+import * as cp from 'child_process';
+import {
+  runTests,
+  downloadAndUnzipVSCode,
+  resolveCliPathFromVSCodeExecutablePath
+} from 'vscode-test';
+import { assertNever } from '../pure/helpers-pure';
+
+// For some reason, `TestOptions` is not exported directly from `vscode-test`,
+// but we can be tricky and import directly from the out file.
```
adityasharad, line 12: crafty

extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
```diff
@@ -22,11 +23,21 @@ type Suite = {
 // testing against old versions if necessary.
 const VSCODE_VERSION = 'stable';
 
+// List if test dirs
+//   - no-workspace - Tests with no workspace selected upon launch.
+//   - minimal-workspace - Tests with a simple workspace selected upon launch.
+//   - cli-integration - Tests that require a cli to invoke actual commands
+enum TestDir {
+  NoWorskspace = 'no-workspace',
+  MinimalWorskspace = 'minimal-workspace',
```
adityasharad, line 32: minor typos: Workspace

extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
adityasharad: Annoying but fine for now, since I see you manually install the Test Explorer.

.github/workflows/main.yml
adityasharad: We've had a few recent releases, so let's go with 2.2.6, 2.3.3, 2.4.0, latest.
https://github.com/github/codeql-cli-binaries/releases/latest will always point you to the latest release, so we should deliberately test against that even if we are slow to update the hardcoded list.

extensions/ql-vscode/src/vscode-tests/ensureCli.ts
```diff
@@ -0,0 +1,129 @@
...
+  console.error(e);
+  process.exit(-1);
+});
+
+const _1MB = 1024 * 1024;
+const _10MB = _1MB * 10;
+
+// CLI version to test. Hard code the latest as default. And be sure
+// to update the env if it is not otherwise set.
+const CLI_VERSION = process.env.CLI_VERSION || 'v2.3.1';
```
adityasharad, line 38: I think you can hardcode `latest`, though you should check the API docs to make sure the asset download URL still works the same.

adityasharad (commented):
This is great! Didn't go into the details of the test harness execution from Node (if it runs, I trust you) but the refactoring and tests themselves look good.

extensions/ql-vscode/src/query-history.ts
aeisenberg: done

extensions/ql-vscode/src/extension.ts
aeisenberg: Correct. The return value is the public api of the extension.

extensions/ql-vscode/src/vscode-tests/ensureCli.ts
aeisenberg: Unfortunately,

This does not work:
https://github.com/github/codeql-cli-binaries/releases/download/latest/codeql-osx64.zip

But this does:
https://github.com/github/codeql-cli-binaries/releases/download/v2.3.1/codeql-osx64.zip

aeisenberg: Also, a downside of using `latest` locally is that once you download `latest` once, you will never re-download. We check whether or not to re-download based on the presence of the directory name. The directory name is generated based on the version number. I could probably be more sophisticated about this, but that would add complexity.

adityasharad (approved):

PR #648, opened 2020-10-27
Ensure decoded archive fs paths are never empty

### LINKED ISSUE #647: Database sources unavailable after adding database on windows
The database source folder is not available on windows after upgrading to v1.3.4 of the extension and adding a new database.

There is a workaround to manually edit the workspace configuration folder and add a `/` as the last character in the path.

aeisenberg:
Empty paths should be replaced as '/'. This is a fix for a bug
introduced in 899f988df871c4923af01508b7efd06169888b90.

Resolves #647.

## Checklist


aeisenberg:
Confirmed that this works both in mac and windows. Not tested on linux, but I have no reason to believe it is different from mac. The problem was that our zip file extractor was failing to find the root directory when there is no slash. 899f988 removed the requirement for source archive root paths to have a `/`. This commit ensures that they always do.

alexet (approved):

Issue #652, opened 2020-10-28
Encode URI is being passed as a source location into the archive-filesystem-provider


aeisenberg:
**Describe the bug**

We are sometimes passing encoded uris into the archive-filesystem-provider for the path inside of the archive file. The file paths are all unencoded. And so, when we have a source location URI with special characters in it, the locations cannot be found inside of the archive.

**To Reproduce**

1. Create a database whose source is inside of a directory path with a special character in it (eg- create a space in one of the directory names)
2. Import database into the extension
3. Run a non-interpreted query that includes a source location
4. BUG: When clicking on the source location you are _not_ navigated to the source.

**Expected behavior**
You _are_ navigated to the source

**Additional context**
Interpreted queries (ie- problems and path problems) do not have this issue. The sarif values already have unencoded paths.

aeisenberg:
We need to unencode only those locations that are from non-interpreted results. This _should_ happen in the extension side, not the webview side.

RasmusWL:
I just experienced this from running a `path-problem` in VS Code.

I used the Python database from https://lgtm.com/projects/g/SuryaPradeepM/web-design with the new [CodeInjection query](https://github.com/github/codeql/blob/92494441a72cd1c04c9359dc8c0d35e2a4525f6d/python/ql/src/Security/CWE-094/CodeInjection.ql). The result location has spaces in it, and from the logs I see:

```
Unable to handleMsgFromView: cannot open codeql-zip-archive://.../SuryaPradeepM_web-design_d1625d1/src.zip/opt/src/web%2520design%2520assignment/form%2520project/intform.py. Detail: Unable to read file 'codeql-zip-archive://.../SuryaPradeepM_web-design_d1625d1/src.zip/opt/src/web%20design%20assignment/form%20project/intform.py' (EntryNotFound (FileSystemError): uri 'codeql-zip-archive://.../SuryaPradeepM_web-design_d1625d1/src.zip/opt/src/web%2520design%2520assignment/form%2520project/intform.py', interpreted as '/opt/src/web%20design%20assignment/form%20project/intform.py' in archive '.../SuryaPradeepM_web-design_d1625d1/src.zip')
```

aeisenberg:
@RasmusWL, if you're feeling courageous, you can install the version of the extension built in this PR (see the checks page) and install that. I've verified that it is working on windows, but further verification would be nice.

aeisenberg:
It's this PR #653. Forgot to include the link.

[...]

PR #653, opened 2020-10-28
Ensure uris are using encoded strings

### LINKED ISSUE #562: Add navigate to super/overrides declaration
CodeQL makes heavy use of overridden predicates on classes. There is no simple navigation step to go from the current predicate to the super-predicate or sub-predicate. It would be useful to implement something like this.

aeisenberg:
Fixes #562

This fixes a bug where if there are special characters in a database
path, it is not possible to navigate to that file from the results view.

Note that the results from our BQRS returned properly encoded URIs, but
our paths coming from sarif were unencoded. Our path parsing handled
the latter correctly (even though these are not correct URIs) and the
former incorrectly.

The fix here is to first ensure all uris are properly encoded. We do
this by running `encodeURI` in sarif-utils (can't run encodeURIComponent
or else the path separators `/` will also be encoded).

Then, we ensure that when we resolve locations, we decode all file
paths.

This works in all cases I have tried. I still have an issue with running
View AST on some of these databases, but that I believe is a separate
issue.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/contextual/fileRangeFromURI.ts
+++ b/extensions/ql-vscode/src/contextual/fileRangeFromURI.ts
@@ -20,9 +20,8 @@ export default function fileRangeFromURI(uri: UrlValue | undefined, db: Database
       Math.max(0, (loc.endLine || 0) - 1),
       Math.max(0, (loc.endColumn || 0)));
     try {
-      const parsed = vscode.Uri.parse(uri.uri, true);
-      if (parsed.scheme === 'file') {
-        return new vscode.Location(db.resolveSourceFile(parsed.fsPath), range);
+      if (uri.uri.startsWith('file:')) {
+        return new vscode.Location(db.resolveSourceFile(uri.uri), range);
       }
       return undefined;
     } catch (e) {
--- a/extensions/ql-vscode/src/databases.ts
+++ b/extensions/ql-vscode/src/databases.ts
@@ -341,28 +341,29 @@ export class DatabaseItemImpl implements DatabaseItem {
     }
   }
 
-  public resolveSourceFile(uri: string | undefined): vscode.Uri {
+  public resolveSourceFile(uriStr: string | undefined): vscode.Uri {
     const sourceArchive = this.sourceArchive;
-    // Sometimes, we are passed a path, sometimes a file URI.
-    // We need to convert this to a file path that is probably inside of a zip file.
-    const file = uri?.replace(/file:/, '');
+    const uri = uriStr ? vscode.Uri.parse(uriStr, true) : undefined;
+    if (uri && uri.scheme !== 'file') {
+      throw new Error(`Invalid uri scheme in ${uriStr}. Only 'file' is allowed.`);
+    }
     if (!sourceArchive) {
-      if (file) {
-        // Treat it as an absolute path.
-        return vscode.Uri.file(file);
+      if (uri) {
+        return uri;
       } else {
         return this.databaseUri;
       }
     }
 
-    if (file) {
-      const absoluteFilePath = file.replace(':', '_');
-      // Strip any leading slashes from the file path, and replace `:` with `_`.
-      const relativeFilePath = absoluteFilePath.replace(/^\/*/, '').replace(':', '_');
+    if (uri) {
+      const relativeFilePath = decodeURI(uri.path).replace(':', '_').replace(/^\/*/, '');
       if (sourceArchive.scheme === zipArchiveScheme) {
         const zipRef = decodeSourceArchiveUri(sourceArchive);
+        const pathWithinSourceArchive = zipRef.pathWithinSourceArchive === '/'
+          ? relativeFilePath
+          : zipRef.pathWithinSourceArchive + '/' + relativeFilePath;
         return encodeSourceArchiveUri({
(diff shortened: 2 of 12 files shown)


aeisenberg:
Confirmed that this fix works on windows, too.

extensions/ql-vscode/src/contextual/fileRangeFromURI.ts
```diff
@@ -20,9 +20,8 @@ export default function fileRangeFromURI(uri: UrlValue | undefined, db: Database
       Math.max(0, (loc.endLine || 0) - 1),
       Math.max(0, (loc.endColumn || 0)));
     try {
-      const parsed = vscode.Uri.parse(uri.uri, true);
-      if (parsed.scheme === 'file') {
-        return new vscode.Location(db.resolveSourceFile(parsed.fsPath), range);
+      if (uri.uri.startsWith('file')) {
```
alexet, line 23: Can we make this `file:` to prevent silly uris like `filefoo:` from crashing.

extensions/ql-vscode/src/databases.ts
```diff
@@ -341,28 +341,30 @@ export class DatabaseItemImpl implements DatabaseItem {
...
       }
     }
 
-    if (file) {
-      const absoluteFilePath = file.replace(':', '_');
-      // Strip any leading slashes from the file path, and replace `:` with `_`.
-      const relativeFilePath = absoluteFilePath.replace(/^\/*/, '').replace(':', '_');
+    if (uri) {
+      // path or fsPath
+      const relativeFilePath = decodeURI(uri.path.replace(':', '_').replace(/^\/*/, ''));
```
alexet, line 360: Replacement should happen after decoding here.

extensions/ql-vscode/test/pure-tests/sarif-utils.test.ts
```diff
@@ -49,6 +49,10 @@ describe('parsing sarif', () => {
       .to.eq('file:C:/a/b/?x=test');
     expect(getPathRelativeToSourceLocationPrefix('C:\\a\\b', '%3Fx%3Dtest'))
       .to.eq('file:C:/a/b/%3Fx%3Dtest');
```
alexet, line 51: These tests are wrong in that file uri paths must always start with a `/`.

extensions/ql-vscode/src/sarif-utils.ts
```diff
@@ -72,7 +72,18 @@ export function getPathRelativeToSourceLocationPrefix(
   sourceLocationPrefix: string,
   sarifRelativeUri: string
 ) {
-  const normalizedSourceLocationPrefix = sourceLocationPrefix.replace(/\\/g, '/');
+  // convert a platform specific path into encoded path uri segments
+  // need to be careful about drive letters
+  let prefix = '';
+  if (sourceLocationPrefix[1] === ':') {
```
alexet, line 78: This ins't correct as windows paths should have a `/` before the drive letter. 
I think the following is correct (it is what we do in java and seems to mostly match vscode-uri). I also notice that vscode-uri is availiable in npm standalone. Is it possible to use it directly and therefore allow it to be used in tests. Then we avoid some of this awkwardness of implementing uri operations standalone.

```
const normalisedPath = path.replace('/\\/g', '/');
const pathComponent = path[0] ==  '/' ? path : "/" + path;
const encodedPathComponent = pathComponent .split('/').map(encodeURIComponent).join('/');
return `file:${encodedPathComponent}/${sarifRelativeUri}`
```

alexet (changes requested):

extensions/ql-vscode/src/contextual/fileRangeFromURI.ts
aeisenberg: sure.

extensions/ql-vscode/src/databases.ts
aeisenberg: Yes

extensions/ql-vscode/test/pure-tests/sarif-utils.test.ts
aeisenberg: The first argument to `getPathRelativeToSourceLocationPrefix` is not a URI component. It is a file path that comes from the sarif file. Or at least that is my understanding of how sarif works (and how this extension has been implemented since I arrived).

When I look at sarif files, all the locations are json blobs like this:

```
"location" : {
        "uri" : "include/virtio.h",
        "uriBaseId" : "%SRCROOT%",
        "index" : 1
}
```

The `uri` properties seem to be relative paths.

It's possible that we don't need to do the check for the drive letter (though, not sure about that). Also, maybe should add some more documentation here.

extensions/ql-vscode/src/sarif-utils.ts
aeisenberg: OK. I see what you're saying. I think we've been lucky that the URI implementation is flexible enough to handle these uris. I can change this.

aeisenberg:
Changes addressed, but there are conflicts with main. Let me fix those up too.

alexet (approved):

alexet:
The linked issue was wrong and I have reopened it, but I don't know where the correct issue is.

aeisenberg:
Should have been #652.

PR #655, opened 2020-10-31
Add the assert-pure query

aeisenberg:
This query ensures that all of our files marked as "pure" remain that
way. In this case "pure" means that it does not depend on vscode and
can therefore be run in tests outside of a runtime environment.

This commit also explicitly moves all of our "pure" files to the
`src/pure` directory.

This PR is expected to have a failing query. I purposely used `src` instead of `pure` as the directory name. Before merging, we should change to `pure`.


.github/codeql/codeql-config.yml
```diff
@@ -0,0 +1,10 @@
+name: "CodeQL config"
+queries:
+  - name: Run custom javascript queries
+    uses: ./.github/codeql/queries
```
adityasharad, line 4: Make sure we keep running the standard queries too (and we might as well run the security+quality set instead of just security).
```suggestion
  - name: Run standard queries
    uses: security-and-quality
  - name: Run custom javascript queries
    uses: ./.github/codeql/queries
```

.github/codeql/queries/assert-pure.ql
```diff
@@ -0,0 +1,29 @@
+/**
+ * @name Assert Pure
```
adityasharad, line 2:
```suggestion
 * @name Unwanted dependency on vscode API
```

.github/codeql/queries/assert-pure.ql
```diff
@@ -0,0 +1,29 @@
...
+ * @kind problem
+ * @problem.severity error
+ * @id vscode-codeql/assert-pure
+ * @description This query asserts that the modules tested in the pure-tests are "pure".
```
adityasharad, line 6:
```suggestion
 * @description The modules stored under `pure` and tested in the `pure-tests` are intended to be "pure".
```

.github/codeql/queries/assert-pure.ql
```diff
@@ -0,0 +1,29 @@
...
+ImportDeclaration transitiveImports(Module m) {
+  exists (ImportDeclaration i |
+    (result = i and i = m.getAnImport()) or
+    (result = transitiveImports(m.getAnImport().resolveImportedPath()))
+  )
+}
+
+from Module m, VSCodeImportDeclaration v
+where
+  // For now, let's use `src` and force this to fail. Should change to `pure`.
```
adityasharad, line 26:
```suggestion
```

.github/codeql/queries/assert-pure.ql
```diff
@@ -0,0 +1,29 @@
...
+  m.getFile().getParentContainer().getBaseName() = "pure" and
```
adityasharad, line 27: This won't handle subfolders within `pure`. Should we consider those too?

.github/codeql/queries/assert-pure.ql
```diff
@@ -0,0 +1,29 @@
...
+  v = transitiveImports(m)
+select m.getFile(), "This module is not pure"
```
adityasharad, line 29: The message will show up in the alert, so let's make it obvious what the problem is.
```suggestion
select m, "This module is not pure: it has a transitive dependency on the vscode API imported $@", v, "here"
```

.github/codeql/queries/assert-pure.ql
```diff
@@ -0,0 +1,29 @@
...
+  VSCodeImportDeclaration() {
+    this.getImportedPath().getValue() = "vscode"
+  }
+}
+
+ImportDeclaration transitiveImports(Module m) {
+  exists (ImportDeclaration i |
+    (result = i and i = m.getAnImport()) or
+    (result = transitiveImports(m.getAnImport().resolveImportedPath()))
+  )
```
adityasharad, line 21: As written, the second disjunct is a Cartesian product, because it doesn't talk about `i`. We can pull out the base definition of `i` and use it in both disjuncts.

```
  exists (ImportDeclaration i | i = m.getAnImport() |
    result = i or
    result = transitiveImports(i.resolveImportedPath())
  )
```

As a more general approach, perhaps we can get away with just the following, assuming there is a `Module` entity for `vscode`:
```
where m.getAnImportedModule+().getName() = "vscode"
```

adityasharad (commented):
Neat! Some suggestions on the query and code scanning configuration.

.github/codeql/queries/assert-pure.ql
aeisenberg: We don't have any at the moment, but we probably should.

.github/codeql/queries/assert-pure.ql
aeisenberg: The second suggestion does not allow for using `VSCodeImportDeclaration` in the results. So, I think your first suggestion will work better.

aeisenberg:
All comments addressed.

.github/codeql/queries/assert-pure.ql
adityasharad: Good point! It's looking good already now, but for further refinement we can:
- use `Import` everywhere instead of `ImportDeclaration` (so it handles all forms of import, even though in practice we generally stick to ES2015 imports)
- simplify the recursive logic to `where m.getAnImportedModule*().getAnImport() = v`

aeisenberg: Done.

adityasharad (approved):

.github/codeql/queries/assert-pure.ql
adityasharad: Sorry...tried to be too clever. `Import` is an abstract class so extending it won't work here.
Two options:
- More general but wordy: `class VSCodeImport extends ASTNode { VSCodeImport() { this.(Import).getImportedPath()... }`
- Less general but neater: `class VSCodeImport extends ImportDeclaration { VSCodeImport() { this.getImportedPath()... }`

aeisenberg: I didn't even know that syntax works.

.github/codeql/queries/assert-pure.ql
aeisenberg: Pushed a change that should get this working on sub-directories.

.github/codeql/queries/assert-pure.ql
```diff
@@ -0,0 +1,21 @@
...
+
+class VSCodeImport extends ASTNode {
+  VSCodeImport() {
+    this.(Import).getImportedPath().getValue() = "vscode"
+  }
+}
+
+from Module m, VSCodeImport v
+where
+  m.getFile().getParentContainer().getRelativePath().regexpMatch(".*src/pure/.*")  and
```
adityasharad, line 19:
```suggestion
  m.getFile().getRelativePath().regexpMatch(".*src/pure/.*")  and
```

aeisenberg:
This query is verified in the query console.

PR #657, opened 2020-11-03
Results View: Fix display of booleans

alexet:
React nodes containing booleans aren't rendered. We want a node with a string. 

I don't why this has only just stopped working but this seems to fix the problem.


dbartol (approved):
LGTM

PR #666, opened 2020-11-03
Use the value not the label for the print ast node

### LINKED ISSUE #659: AST viewer uses `node.toString` for labels and ignores `value`
Here's a simple, language-agnostic print-AST query implementation:

```codeql
/**
 * @id foo/print-ast
 * @kind graph
 * @tags ide-contextual-queries/print-ast
 */
class Node extends string {
  Node() { this in ["foo", "bar", "baz"] }

  Node getChild() { this = "foo" and result in ["bar", "baz"] }
}

query predicate nodes(Node node, string key, string value) {
  key = "semmle.label" and
  value = node.toUpperCase()
}

query predicate edges(Node source, Node target, string key, string value) {
  source.getChild() = target and
  key = "semmle.label" and
  value = "edge"
}

query predicate graphProperties(string key, string value) {
  key = "semmle.graphKind" and value = "tree"
}
```

And here's how it looks in the AST viewer:
image

My understanding after a discussion with @dbartol is that the AST viewer is expected to use the `value` argument as the label for nodes, but instead it uses `node.toString()`. You can see in the screenshot that the labels are lower-case, so `value` is not being used.

aeisenberg:
Fixes #659

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/contextual/astBuilder.ts
+++ b/extensions/ql-vscode/src/contextual/astBuilder.ts
@@ -75,18 +75,18 @@ export default class AstBuilder {
 
     // populate parents and children
     nodeTuples.tuples.forEach(tuple => {
-      const [entity, tupleType, orderValue] = tuple as [EntityValue, string, string];
+      const [entity, tupleType, value] = tuple as [EntityValue, string, string];
       const id = entity.id!;
 
       switch (tupleType) {
         case 'semmle.order':
-          astOrder.set(id, Number(orderValue));
+          astOrder.set(id, Number(value));
           break;
 
         case 'semmle.label': {
           const item = {
             id,
-            label: entity.label,
+            label: value ?? entity.label,
             location: entity.url,
             fileLocation: fileRangeFromURI(entity.url, this.db),
             children: [] as ChildAstItem[],
--- a/extensions/ql-vscode/src/vscode-tests/no-workspace/data/astBuilder.json
+++ b/extensions/ql-vscode/src/vscode-tests/no-workspace/data/astBuilder.json
@@ -18,7 +18,7 @@
       [
         {
           "id": 26359,
-          "label": "",
+          "label": "SHOULD NOT USE ",
           "url": {
             "uri": "file:/opt/src/arch/sandbox/lib/interrupts.c",
             "startLine": 19,
@@ -33,7 +33,7 @@
       [
         {
           "id": 26360,
-          "label": "",
+          "label": "SHOULD NOT USE ",
           "url": {
             "uri": "file:/opt/src/arch/sandbox/lib/interrupts.c",
             "startLine": 15,
@@ -48,7 +48,7 @@
       [
         {
           "id": 26361,
-          "label": "",
+          "label": "SHOULD NOT USE ",
           "url": {
             "uri": "file:/opt/src/arch/sandbox/lib/interrupts.c",
             "startLine": 10,
@@ -63,7 +63,7 @@
       [
         {
           "id": 0,
-          "label": "[TopLevelFunction] int disable_interrupts()",
+          "label": "SHOULD NOT USE [TopLevelFunction] int disable_interrupts()",
           "url": {
             "uri": "file:/opt/src/arch/sandbox/lib/interrupts.c",
             "startLine": 19,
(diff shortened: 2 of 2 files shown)


extensions/ql-vscode/src/contextual/astBuilder.ts
```diff
@@ -75,18 +75,18 @@ export default class AstBuilder {
...
         case 'semmle.order':
-          astOrder.set(id, Number(orderValue));
+          astOrder.set(id, Number(value));
           break;
 
         case 'semmle.label': {
           const item = {
             id,
-            label: entity.label,
+            label: value,
```
adityasharad, line 89: What happens if we have an AST query that doesn't set `semmle.label` on its nodes? For example
```
query predicate nodes(Node node) { ... }
```
In that case I would expect `value` here to be falsy, and I suggest we fall back to `entity.label` in that default case.

extensions/ql-vscode/src/vscode-tests/no-workspace/data/astBuilder.json
```diff
@@ -18,7 +18,7 @@
       [
         {
           "id": 26359,
-          "label": "",
+          "label": "SHOULD NOT USE ",
```
adityasharad, line 21: I like having a fixed test like this, but we have to keep it up to date. Is it feasible to have an integration test that runs a real AST query and checks we can handle the results?

adityasharad (commented):
Generally looks good, some questions about fallback behaviour and testing.

extensions/ql-vscode/src/contextual/astBuilder.ts
aeisenberg: Sounds reasonable.

extensions/ql-vscode/src/vscode-tests/no-workspace/data/astBuilder.json
aeisenberg: We can do that. I think we should create a new issue that lists a bunch of integration tests that we should add once the integration test PR is merged. I'll do that right now.

extensions/ql-vscode/src/contextual/astBuilder.ts
adityasharad: To clarify: I think we should handle that in the `default:` case too, in case `semmle.label` is not set at all.

extensions/ql-vscode/src/vscode-tests/no-workspace/data/astBuilder.json
aeisenberg: #676 is the start. Feel free to add more.

extensions/ql-vscode/src/contextual/astBuilder.ts
aeisenberg: I don't think that's the behaviour we want. Right now, I don't know what the other kind of labels there are, but we don't want them to overwrite whatever is in the `semmle.label`. And more importantly, we don't even know what _would_ be in the value field and if it makes any sense to display.

adityasharad: If we can assume all AST queries follow this pattern:
```
query predicate nodes(Node node, string key, string value) {
  key = "semmle.label" and ...
}
```
then the `nodes` table will have tuples with `semmle.label` set, and this PR is good to go.

I am trying to be cautious about the possibility that _no_ such label is set, which will happen if you write:
```
query predicate nodes(Node node) { ... }
```
In that case I think we'd have to create an item under the `default:` case of the `switch` in this code, which uses `entity.label`. Perhaps that's not something we need to handle in this PR if we know it won't arise from any of the implementations.

aeisenberg: If an entity does not have a `semmle.label` `tupleType`, then I think this is a bug. As far as I know existing languages all borrow from the cpp implementation which always creates a `semmle.label` for each node. This is not mentioned in the specs and maybe we should clarify.

aeisenberg: https://github.com/github/codeql-c-analysis-team/pull/187

adityasharad (approved):
I'm convinced! Ship it.

PR #667, opened 2020-11-03
Disable codeql test commands from the command palette

### LINKED ISSUE #658: CodeQL: Show Test Output Differences fails from the Command Palette
**Describe the bug**
Choosing the `CodeQL: Show Test Output Differences` option from the Command Palette gives the following error:
```
Cannot read property 'info' of undefined
```
The command works fine when selected from the right-click context menu for the test in the Test Explorer

**To Reproduce**

 1. Run a CodeQL query test that fails
 2. Select the `CodeQL: Show Test Output Differences` option from the Command Palette.

**Expected behavior**
Ideally, it would show me the test output differences for the last test. Otherwise, can we hide it from the Command Palette? Or at least not throw an error?

aeisenberg:
These commands are not applicable from the global context. They require
an argument to be passed in. So, they should be hidden in the command
palette.

Fixes #658.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -355,11 +355,11 @@
       },
       {
         "command": "codeQLTests.showOutputDifferences",
-        "title": "CodeQL: Show Test Output Differences"
+        "title": "Show Test Output Differences"
       },
       {
         "command": "codeQLTests.acceptOutput",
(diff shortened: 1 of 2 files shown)


adityasharad (approved):
I was going to suggest a query, but it looks like you've modelled the pattern in the unit tests instead.

PR #668, opened 2020-11-03
Avoid recursive selection changes in ast viewer

### LINKED ISSUE #656: AST Viewer selection bounces around
When clicking on an entry in the AST viewer, if there is a child entry whose source location is identical to the parent's location, the selected entry will change to the child entry.

The reason is that we do a depth first search of the best tree selection based on current text selection.

Possible solution: when determining which viewer node to select, if two nodes have the same source location, but one is already selected don't change the selection.

aeisenberg:
This will prevent selections jumping around when an ast entry is
selected and its child has the same source location as the current
selection.

Fixes #656.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/astViewer.ts
+++ b/extensions/ql-vscode/src/astViewer.ts
@@ -8,6 +8,7 @@ import {
   TreeItem,
   TreeView,
   TextEditorSelectionChangeEvent,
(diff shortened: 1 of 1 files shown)


aeisenberg:
Turned out to be a lot simpler than I expected. There's an API for this.

p0 (approved):

PR #669, opened 2020-11-03
Creates an empty `.expected` file when running test output compare

### LINKED ISSUE #338: In test diffs, compare the result to an empty file if no .expected file exists
After creating a new test (with source and query, but no `.expected` file) and running a test through the test explorer, the "show test output differences" menu item fails with `Error: unable to resolve non-existing file`.

The "Show test output differences" menu item opens a diff window in which the `.actual` file is compared to an empty file. It might also display a warning message.

**Describe alternatives you've considered**
The menu item could instead open the `.actual` file itself, possibly while displaying a warning message.

aeisenberg:
If the expected file does not already exists. This helps with test
creation and allows users to create tests more quickly.

Fixes #338.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Editors opened by navigating from the results view are no longer opened in _preview mode_. Now they are opened as a persistent editor. [#630](https://github.com/github/vscode-codeql/pull/630)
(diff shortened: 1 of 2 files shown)


adityasharad (approved):
Sensible.
@rdmarsh2 or @dbartol do you know if we have test cases without an `.expected` file (rather than having an empty file)? Do consistency checks use this pattern? Just want to make sure we're not spuriously creating empty `.expected` files during extension use.

rdmarsh2:
To my knowledge, we don't. All the C++ consistency checks are done with multiple query predicates imported into a single `.qll`, so they have predicate names in the `.expected` files even if they pass.

adityasharad:
Thanks! In that case, LGTM once conflicts are resolved.

PR #675, opened 2020-11-06
Clean databases folder on startup

### LINKED ISSUE #674: Periodically and on demand clean up old databases in the extension
Most of the time when a user removes a database, it is also cleared from disk, but sometimes this doesn't happen. I haven't figured out the precise situations where this can occur, but I think it has something to do with workspaces that aren't shut down properly (ie- vs code crashes). And it may be more likely on windows which more aggressively locks folders.

The proposal here is to create a command that will look in workspace storage for any abandoned databases. This command can be run on demand, and also in the background after startup. (Arguably, we may not want to expose this command directly to users since it really is a plumbing command.)

It will work like this:

1. get all directories in workspace storage
2. check each directory to see if it is a database
3. check if this database is currently imported in the workspace
4. if not delete the entire directory

aeisenberg:
Cleans orphan databases on startup. This commit also bumps the fs-extra
dependency to get readdir with dirent objects.

Adds the `asyncFilter` to filter arrays asynchronously.

Implemented as a command so a saavy user could assign a keyboard shortcut to it, but it is not accessible through the command palette.

Fixes #674.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -198,6 +198,10 @@
         }
       },
       {
+        "command": "codeQLDatabases.removeOrphanedDatabases",
+        "title": "Delete unused databases"
+      },
+      {
         "command": "codeQLDatabases.chooseDatabaseArchive",
         "title": "Choose Database from Archive",
         "icon": {
@@ -574,6 +578,10 @@
           "when": "false"
         },
         {
+          "command": "codeQLDatabases.removeOrphanedDatabases",
+          "when": "false"
+        },
+        {
           "command": "codeQLDatabases.chooseDatabaseInternet",
           "when": "false"
         },
@@ -704,7 +712,7 @@
   "dependencies": {
     "child-process-promise": "^2.2.1",
     "classnames": "~2.2.6",
-    "fs-extra": "^8.1.0",
+    "fs-extra": "^9.0.1",
     "glob-promise": "^3.4.0",
     "js-yaml": "^3.14.0",
     "minimist": "~1.2.5",
@@ -727,7 +735,7 @@
     "@types/chai-as-promised": "~7.1.2",
     "@types/child-process-promise": "^2.2.1",
     "@types/classnames": "~2.2.9",
-    "@types/fs-extra": "^8.0.0",
+    "@types/fs-extra": "^9.0.3",
     "@types/glob": "^7.1.1",
     "@types/google-protobuf": "^3.2.7",
     "@types/gulp": "^4.0.6",
@@ -735,7 +743,7 @@
     "@types/js-yaml": "^3.12.5",
     "@types/jszip": "~3.1.6",
     "@types/mocha": "~8.0.3",
-    "@types/node": "^12.0.8",
+    "@types/node": "^12.14.1",
     "@types/node-fetch": "~2.5.2",
     "@types/proxyquire": "~1.3.28",
     "@types/react": "^16.8.17",
--- a/extensions/ql-vscode/src/databases-ui.ts
+++ b/extensions/ql-vscode/src/databases-ui.ts
@@ -8,7 +8,7 @@ import {
   TreeItem,
   Uri,
   window,
-  env
+  env,
 } from 'vscode';
 import * as fs from 'fs-extra';
 
@@ -18,6 +18,8 @@ import {
   DatabaseItem,
   DatabaseManager,
   getUpgradesDirectories,
+  isLikelyDatabaseRoot,
+  isLikelyDbLanguageFolder,
 } from './databases';
 import {
   commandRunner,
@@ -36,6 +38,7 @@ import {
   promptImportLgtmDatabase,
 } from './databaseFetcher';
 import { CancellationToken } from 'vscode-jsonrpc';
+import { asyncFilter } from './pure/helpers-pure';
 
 type ThemableIconPath = { light: string; dark: string } | string;
 
@@ -229,7 +232,9 @@ export class DatabaseUI extends DisposableObject {
         canSelectMany: true,
       })
     );
+  }
 
+  init() {
     logger.log('Registering database panel commands.');
     this.push(
       commandRunnerWithProgress(
@@ -340,6 +345,12 @@ export class DatabaseUI extends DisposableObject {
         this.handleOpenFolder
       )
     );
+    this.push(
+      commandRunner(
+        'codeQLDatabases.removeOrphanedDatabases',
+        this.handleRemoveOrphanedDatabases
+      )
+    );
   }
 
   private handleMakeCurrentDatabase = async (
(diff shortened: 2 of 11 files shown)


aeisenberg:
Ping! @github/docs-content-dsp This is not entirely a user facing feature and perhaps mentioning in the changelog is good enough. But some users may want to be aware of this. 

From the changelog: 

> Whenever the extension restarts, orphaned databases will be cleaned up. These are databases whose files are located inside of the extension's storage area, but are not imported into the workspace.

shati-patel:
> Ping! @github/docs-content-dsp This is not entirely a user facing feature and perhaps mentioning in the changelog is good enough. But some users may want to be aware of this.
> 
> From the changelog:
> 
> > Whenever the extension restarts, orphaned databases will be cleaned up. These are databases whose files are located inside of the extension's storage area, but are not imported into the workspace.

Thanks for letting us know 👍🏽 What does "cleaning up" the database actually mean in this case? As long as it doesn't break the database or do anything unexpected, we're probably fine with just a changelog entry 😊

aeisenberg:
Cleaning up means deleting unused databases.

In this case, the user has first imported a database as a zip file (or from a url or LGTM) and it was placed inside of the extension's storage area (a place that is not user facing). Then the user has removed this database from the extension. Most of the time, this means that the database is deleted from the file system as well, but sometimes (typically because windows has not released the file system lock), the database cannot immediately be removed from disk. Thus the database is _orphaned_. It exists, but is no longer used anywhere.

This change will ensure that these orphaned databases are removed eventually.

Note that this change _does not_ affect databases added as a filesystem folder. We assume these databases are user controlled even after being removed from the extension.

shati-patel:
> Cleaning up means deleting unused databases.
> 
> In this case, the user has first imported a database as a zip file (or from a url or LGTM) and it was placed inside of the extension's storage area (a place that is not user facing). Then the user has removed this database from the extension. Most of the time, this means that the database is deleted from the file system as well, but sometimes (typically because windows has not released the file system lock), the database cannot immediately be removed from disk. Thus the database is _orphaned_. It exists, but is no longer used anywhere.
> 
> This change will ensure that these orphaned databases are removed eventually.
> 
> Note that this change _does not_ affect databases added as a filesystem folder. We assume these databases are user controlled even after being removed from the extension.

Great, thanks for clarifying! That all sounds sensible and pretty harmless. I don't think we need to document it.

extensions/ql-vscode/test/pure-tests/helpers-pures.test.ts
```diff
@@ -0,0 +1,22 @@
+import { fail } from 'assert';
```
adityasharad, line 1: Typo in filename: pures -> pure

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -670,7 +718,11 @@ export class DatabaseUI extends DisposableObject {
...
 // TODO: Get the list of supported languages from a list that will be auto-updated.
-const dbRegeEx = /^db-(javascript|go|cpp|java|python|csharp)$/;
+const dbRegeEx = /^db-(javascript|go|cpp|java|python|csharp|ruby)$/;
 function isLikelyDbFolder(dbPath: string) {
   return path.basename(dbPath).match(dbRegeEx);
 }
+
+async function isDatabaseDirectory(dir: string) {
+  return (await fs.readdir(dir)).some(isLikelyDbFolder);
+}
```
adityasharad, line 728: Not very fond of having to maintain a list of language slugs.
Elsewhere, I expect we are checking for the existence of `codeql-database.yml` (or `.dbinfo` as a fallback) to determine whether a directory is a CodeQL database. Can we continue to use that here instead of relying on the dataset folder name?

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -360,6 +369,45 @@ export class DatabaseUI extends DisposableObject {
...
+      dbDirs.map(async dbDir => {
+        try {
+          logger.log(`Deleting orphaned database '${dbDir}'.`);
+          await fs.rmdir(dbDir, { recursive: true } as any);  // typings doesn't recognize the options argument
+        } catch (e) {
+          failures.push(`${e.message}: ${dbDir}`);
+        }
+      })
+    );
+    showAndLogErrorMessage(`Failed to delete orphaned databases:\n  ${failures.join('  \n')}'. Must delete manually.`);
```
adityasharad, line 407: Nit: why leading space before the newline? And since this is user-facing, perhaps give them an action, e.g. 'To delete unused databases, please remove them manually from the workspace storage folder.'

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -360,6 +369,45 @@ export class DatabaseUI extends DisposableObject {
...
+  handleRemoveOrphanedDatabases = async (): Promise<void> => {
+    logger.log('Removing orphaned databases.');
+    let dbDirs =
+      // read directory
+      (await fs.readdir(this.storagePath, { withFileTypes: true }))
+        // remove non-directories
+        .filter(dirent => dirent.isDirectory())
+        // get the full path
+        .map(dirent => path.join(this.storagePath, dirent.name))
+        // filter databases still in workspace
```
adityasharad, line 381:
```suggestion
        // filter out databases still in workspace
```

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -360,6 +369,45 @@ export class DatabaseUI extends DisposableObject {
...
+        .filter(dbDir => {
+          const dbUri = Uri.file(dbDir);
+          return this.databaseManager.databaseItems.every(item => item.databaseUri.fsPath !== dbUri.fsPath);
```
adityasharad, line 384: I don't think we have to do this search for every identified directory in storage.
I suggest creating a `Set` up front containing all the `fsPath` values from `this.databaseManager.databaseItems`, and then this filter just becomes a set lookup.
My assumption here is that when this function is awaited, there is no way for the user to add databases to the workspace after the set is constructed but before cleanup completes (otherwise we have a race condition).

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -360,6 +369,45 @@ export class DatabaseUI extends DisposableObject {
     }
   };
 
+  handleRemoveOrphanedDatabases = async (): Promise<void> => {
+    logger.log('Removing orphaned databases.');
```
adityasharad, line 373:
```suggestion
    logger.log('Removing orphaned databases from workspace storage.');
```

extensions/ql-vscode/src/databases-ui.ts
adityasharad, line 407: This should only be called when `failures` is non-empty.

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -360,6 +369,45 @@ export class DatabaseUI extends DisposableObject {
...
+    dbDirs = await asyncFilter(dbDirs, isDatabaseDirectory);
+
+    if (!dbDirs.length) {
+      logger.log('No orphaned databases found.');
+      return;
+    }
+
+    // delete
+    const failures = [] as string[];
+    await Promise.all(
```
adityasharad, line 397: I think [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) gives you similar result collection behaviour without having to insert your own `catch`. You would have to check for `status == 'rejected'` to collect the failures.

extensions/ql-vscode/src/extension.ts
```diff
@@ -349,6 +349,7 @@ async function activateWithInstalledDistribution(
     getContextStoragePath(ctx),
     ctx.extensionPath
   );
+  databaseUI.init();
```
adityasharad, line 352: Can this call be part of the database UI constructor?

extensions/ql-vscode/package.json
```diff
@@ -197,6 +197,10 @@
           "dark": "media/dark/folder-opened-plus.svg"
         }
       },
+      {
+        "command": "codeQLDatabases.removeOrphanedDatabases",
+        "title": "Remove databases no longer imported into VS Code"
```
adityasharad, line 202: Clean up unused databases?

extensions/ql-vscode/src/databases-ui.ts
aeisenberg: I agree about the hard-coded languages, but I'm not quite sure what to do about them. Maybe there is some kind of way we can discover this by introspecting which standard libraries are installed, but that seems complex for now. 

I'll be more precise and check for one of those two files.

extensions/ql-vscode/src/databases-ui.ts
aeisenberg: Leading space is for indentation, to indent paths we could not remove. Though, the space should come _after_ the newline.

Also, the failures should contain the full path to the database. Maybe I will use only the basename and include the storage folder elsewhere.

extensions/ql-vscode/src/databases-ui.ts
aeisenberg: `allSettled` was introduced in node 12.9. I _think_ that's fine to use in vscode (my instance uses v12.14.1).

aeisenberg: Having trouble getting tsc to recognize `allSettled` in the editor. I think I need to change the tsconfig somehow, but not exactly sure.

Though, I don't think the results would be much cleaner since I still need to keep filter the results and keep track of paths.

extensions/ql-vscode/src/extension.ts
aeisenberg: No. The tests will fail since it registers commands. And if we create an instance more than once and re-register commands, there is an error. 

I guess I could alternatively add an optional flag to signal the avoidance of command registration. But, that seems uglier to me.

extensions/ql-vscode/package.json
aeisenberg: This is not really a user facing command. But, sure.

extensions/ql-vscode/src/databases-ui.ts
aeisenberg: yes

extensions/ql-vscode/src/databases-ui.ts
adityasharad: No worries. What you've got is pretty clear :)

alexet (approved):

PR #679, opened 2020-11-13
Non destructive upgrades in vscode

alexet:
This depends on a rebased copy of #639 and also on some unreleased cli features and shouldn't be merged until we know which version they will be released in.

This PR makes us use silent non-destructive upgrades. It also refactors the current upgrade mechanism in places so we have slightly fewer code paths.

There are a few identation changes so you may want to add `?w=1` to the url when reviewing to ignore those changes.

### DIFF
--- a/extensions/ql-vscode/src/pure/messages.ts
+++ b/extensions/ql-vscode/src/pure/messages.ts
@@ -475,7 +475,7 @@ export interface CompileUpgradeResult {
   error?: string;
 }
 
-export interface CompiledUpgradeSequence {
+export interface CompileUpgradeSequenceResult {
   /**
    * The compiled upgrades as a single file.
    */
@@ -1006,7 +1006,7 @@ export const compileUpgrade = new rpc.RequestType<WithProgressId<CompileUpgradeP
 /**
  * Compile an upgrade script to upgrade a dataset.
  */
-export const compileUpgradeSequence = new rpc.RequestType<WithProgressId<CompileUpgradeSequenceParams>, CompiledUpgradeSequence, void, void>('compilation/compileUpgradeSequence');
+export const compileUpgradeSequence = new rpc.RequestType<WithProgressId<CompileUpgradeSequenceParams>, CompileUpgradeSequenceResult, void, void>('compilation/compileUpgradeSequence');
 
 /**
  * Clear the cache of a dataset
--- a/extensions/ql-vscode/src/pure/messages.ts
+++ b/extensions/ql-vscode/src/pure/messages.ts
@@ -425,14 +425,6 @@ export interface CompileUpgradeSequenceParams {
    * A directory to store parts of the compiled upgrade
    */
   upgradeTempDir: string;
-  /** 
-   * The first dbscheme in the sequence.
-   */
-  initialDbscheme: string;
-  /** 
-   * The last dbscheme in the sequence.
-   */
-  finalDbscheme: string;
 }
 
 /**
@@ -483,12 +475,17 @@ export interface CompileUpgradeResult {
   error?: string;
 }
 
-export interface SingleFileCompiledUpgradeResult extends CompileUpgradeResult {
+export interface CompiledUpgradeSequence {
   /**
-   * The compiled upgrade.
+   * The compiled upgrades as a single file.
    */
-  compiledUpgrades?: SingleFileCompiledUpgrade;
+  compiledUpgrades?: string;
+  /**
+   * Any errors that occurred when checking the scripts.
+   */
+  error?: string;
 }
+
 
 /**
  * A description of a upgrade process
@@ -528,7 +525,7 @@ export interface UpgradeDescription {
 }
 
 
-export type CompiledUpgrades = MultiFileCompiledUpgrades | SingleFileCompiledUpgrade
+export type CompiledUpgrades = MultiFileCompiledUpgrades | SingleFileCompiledUpgrades
 
 /**
  * The parts shared by all compiled upgrades
@@ -573,7 +570,7 @@ interface MultiFileCompiledUpgrades extends CompiledUpgradesBase {
  * A compiled upgrade.
  * The upgrade is in a single file.
  */
-export interface SingleFileCompiledUpgrade extends CompiledUpgradesBase {
+export interface SingleFileCompiledUpgrades extends CompiledUpgradesBase {
   /**
    * The steps in the upgrade path
    */
@@ -1009,8 +1006,7 @@ export const compileUpgrade = new rpc.RequestType<WithProgressId<CompileUpgradeP
 /**
  * Compile an upgrade script to upgrade a dataset.
  */
-export const compileUpgradeSequence = new rpc.RequestType<WithProgressId<CompileUpgradeSequenceParams>, SingleFileCompiledUpgradeResult, void, void>('compilation/compileUpgradeSequence');
-
+export const compileUpgradeSequence = new rpc.RequestType<WithProgressId<CompileUpgradeSequenceParams>, CompiledUpgradeSequence, void, void>('compilation/compileUpgradeSequence');
 
 /**
  * Clear the cache of a dataset
--- a/extensions/ql-vscode/src/pure/messages.ts
+++ b/extensions/ql-vscode/src/pure/messages.ts
@@ -414,6 +414,28 @@ export interface CompileUpgradeParams {
 }
 
 /**
+ * Parameters for compiling an upgrade.
+ */
+export interface CompileUpgradeSequenceParams {
+  /**
+   * The sequence of upogrades to compile
+   */
+  upgradePaths: string[];
+  /**
+   * A directory to store parts of the compiled upgrade
+   */
+  upgradeTempDir: string;
+  /** 
+   * The first dbscheme in the sequence.
+   */
+  initialDbscheme: string;
+  /** 
+   * The last dbscheme in the sequence.
+   */
+  finalDbscheme: string;
+}
+
+/**
  * Parameters describing an upgrade
  */
 export interface UpgradeParams {
@@ -460,6 +482,14 @@ export interface CompileUpgradeResult {
    */
   error?: string;
 }
+
+export interface SingleFileCompiledUpgradeResult extends CompileUpgradeResult {
+  /**
+   * The compiled upgrade.
+   */
+  compiledUpgrades?: SingleFileCompiledUpgrade;
+}
+
 /**
  * A description of a upgrade process
  */
@@ -696,6 +726,10 @@ export interface QueryToRun {
    * A uri pointing to the qlo to run.
    */
   qlo: string;
+  /**
+   * A uri pointing to the compiled upgrade file.
+   */
+  compiledUpgrade?: string;
   /**
    * The path where we should save this queries results
    */
@@ -972,6 +1006,10 @@ export const checkUpgrade = new rpc.RequestType<WithProgressId<UpgradeParams>, C
  * Compile an upgrade script to upgrade a dataset.
  */
 export const compileUpgrade = new rpc.RequestType<WithProgressId<CompileUpgradeParams>, CompileUpgradeResult, void, void>('compilation/compileUpgrade');
+/**
+ * Compile an upgrade script to upgrade a dataset.
+ */
+export const compileUpgradeSequence = new rpc.RequestType<WithProgressId<CompileUpgradeSequenceParams>, SingleFileCompiledUpgradeResult, void, void>('compilation/compileUpgradeSequence');
 
 
 /**
--- a/extensions/ql-vscode/src/pure/messages.ts
+++ b/extensions/ql-vscode/src/pure/messages.ts
@@ -418,7 +418,7 @@ export interface CompileUpgradeParams {
  */
 export interface CompileUpgradeSequenceParams {
   /**
-   * The sequence of upogrades to compile
+   * The sequence of upgrades to compile
    */
   upgradePaths: string[];
   /**
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -17,7 +17,6 @@ import * as config from './config';
 import { DatabaseItem } from './databases';
 import { getOnDiskWorkspaceFolders, showAndLogErrorMessage } from './helpers';
 import { ProgressCallback, UserCancellationException } from './commandRunner';
-import * as helpers from './helpers';
 import { DatabaseInfo, QueryMetadata, ResultsPaths } from './pure/interface-types';
 import { logger } from './logging';
 import * as messages from './pure/messages';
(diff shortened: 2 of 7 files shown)


extensions/ql-vscode/src/pure/messages.ts
```diff
@@ -499,14 +537,46 @@ export interface CompiledUpgrades {
...
+ * A compiled upgrade. 
+ * The upgrade is in a single file.
+ */
+export interface SingleFileCompiledUpgrade {
+  /**
+   * The steps in the upgrade path
+   */
+  descriptions: UpgradeDescription[];
+  /**
+   * The file containing the upgrade
```
aeisenberg, line 577: For consistency, keep the comments the same, so we know they are the same thing. If they aren't the same thing, then be explicit about it.

```suggestion
   * A uri pointing to the compiled upgrade file.
```

extensions/ql-vscode/src/pure/messages.ts
```diff
@@ -499,14 +537,46 @@ export interface CompiledUpgrades {
...
-  targetSha: string;
+  compiledUpgradeFile?: never;
+}
+
+
+/**
+ * A compiled upgrade. 
+ * The upgrade is in a single file.
+ */
+export interface SingleFileCompiledUpgrade {
```
aeisenberg, line 571: Does this need to extend `CompiledUpgradesBase`?

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -309,25 +311,60 @@ async function checkDbschemeCompatibility(
...
         progress,
         token
       );
     }
   }
 }
 
+
+function reportNoUpgradePath(query: QueryInfo) {
+  logger.log(`Query ${query.program.queryPath} expects database scheme ${query.queryDbscheme}, but database has scheme ${query.program.dbschemePath}, and no upgrade path found`);
```
aeisenberg, line 332: If this is run through a workbench command, then all errors are logged to the console, so I don't think this line needs to be here.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -309,25 +311,60 @@ async function checkDbschemeCompatibility(
...
+async function compileNonDestructiveUpgrade(
+  qs: qsClient.QueryServerClient,
+  upgradeTemp: tmp.DirResult,
+  query: QueryInfo,
+  progress: helpers.ProgressCallback,
+  token: CancellationToken,
+): Promise<string> {
+  const searchPath = helpers.getOnDiskWorkspaceFolders();
+
+  if (query.dbItem.contents === undefined || query.dbItem.contents.dbSchemeUri === undefined) {
```
aeisenberg, line 348: nit:

```suggestion
  if (!query.dbItem?.contents?.dbSchemeUri) {
```

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -510,64 +547,74 @@ export async function compileAndRunQueryAgainstDatabase(
...
-    for (const error of errors) {
-      const message = error.message || '[no error message available]';
-      const formatted = `ERROR: ${message} (${error.position.fileName}:${error.position.line}:${error.position.column}:${error.position.endLine}:${error.position.endColumn})`;
-      formattedMessages.push(formatted);
-      qs.logger.log(formatted);
     }
-    if (quickEval && formattedMessages.length <= 3) {
-      helpers.showAndLogErrorMessage('Quick evaluation compilation failed: \n' + formattedMessages.join('\n'));
+
+    if (errors.length == 0) {
```
aeisenberg, line 570: This _should_ be a linting error.

```suggestion
    if (errors.length === 0) {
```

aeisenberg (commented):
This is a partial review. Looks good. Just some small things. I'll finish this off on Monday.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -510,64 +547,74 @@ export async function compileAndRunQueryAgainstDatabase(
...
+        formattedMessages.push(formatted);
+        qs.logger.log(formatted);
+      }
+      if (quickEval && formattedMessages.length <= 3) {
+        helpers.showAndLogErrorMessage('Quick evaluation compilation failed: \n' + formattedMessages.join('\n'));
+      } else {
+        helpers.showAndLogErrorMessage((quickEval ? 'Quick evaluation' : 'Query') +
+          ' compilation failed. Please make sure there are no errors in the query, the database is up to date,' +
+          ' and the query and database use the same target language. For more details on the error, go to View > Output,' +
+          ' and choose CodeQL Query Server from the dropdown.');
```
aeisenberg, line 611: Nit: it will make things easier to read if you extract this to a constant.

extensions/ql-vscode/src/upgrades.ts
```diff
@@ -14,73 +17,94 @@ import { upgradesTmpDir } from './run-queries';
...
+ * This requires 3 features. The ability to compile an upgrade sequence; The ability to 
+ * run a non-desturcitve upgrades as a query; the ability to specify a target when 
+ * resolving upgrades.
  */
-async function checkAndConfirmDatabaseUpgrade(
-  qs: qsClient.QueryServerClient,
+export async function hasNondestructiveUpgradeCapabilities(qs: qsClient.QueryServerClient): Promise<boolean> {
+  // TODO change to actual version when known
+  // Note it is probably something 2.4.something
+  return (await qs.cliServer.getVersion()).compare('2.3.2') >= 0;
```
aeisenberg, line 29: I find this easier to read because I can never remember exactly what `compare` returns.

```suggestion
  return semver.satisfies((await qs.cliServer.getVersion()), '2.3.2');
```

extensions/ql-vscode/src/upgrades.ts
```diff
@@ -14,73 +17,94 @@ import { upgradesTmpDir } from './run-queries';
...
+  targetDbScheme: string,
+  resolvedSequence: string[],
+  currentUpgradeTmp: tmp.DirResult,
+  progress: helpers.ProgressCallback,
+  token: vscode.CancellationToken
+): Promise<messages.CompileUpgradeResult> {
+  if (await hasNondestructiveUpgradeCapabilities(qs)) {
+    return await compileDatabaseUpgradeSequence(qs, db, targetDbScheme, resolvedSequence, currentUpgradeTmp, progress, token);
+  } else {
+    if (db.contents === undefined || db.contents.dbSchemeUri === undefined) {
```
aeisenberg, line 68: nit: 

```suggestion
    if (db.contents?.dbSchemeUri) {
```

extensions/ql-vscode/src/upgrades.ts
```diff
@@ -96,107 +120,99 @@ async function checkAndConfirmDatabaseUpgrade(
     dialogOptions.push(showLogItem);
   }
 
-  const message = `Should the database ${db.databaseUri.fsPath} be upgraded?\n\n${messageLines.join('\n')}`;
+  const message = `Should the database ${db} be upgraded?\n\n${messageLines.join('\n')}`;
```
aeisenberg, line 123: Do you mean `db.databaseUri.fsPath`?

extensions/ql-vscode/src/upgrades.ts
```diff
@@ -96,107 +120,99 @@ async function checkAndConfirmDatabaseUpgrade(
...
-  try {
-    compileUpgradeResult = await compileDatabaseUpgrade(qs, upgradeParams, progress, token);
-  }
-  catch (e) {
-    helpers.showAndLogErrorMessage(`Compilation of database upgrades failed: ${e}`);
-    return;
-  }
-  finally {
-    qs.logger.log('Done compiling database upgrade.');
+  if (db.contents === undefined || db.contents.dbSchemeUri === undefined) {
```
aeisenberg, line 164: Same as above

extensions/ql-vscode/src/upgrades.ts
```diff
@@ -14,79 +18,101 @@ import { upgradesTmpDir } from './run-queries';
...
+ * 
+ * This requires 3 features. The ability to compile an upgrade sequence; The ability to 
+ * run a non-destructive upgrades as a query; the ability to specify a target when 
+ * resolving upgrades.
  */
-async function checkAndConfirmDatabaseUpgrade(
-  qs: qsClient.QueryServerClient,
+export async function hasNondestructiveUpgradeCapabilities(qs: qsClient.QueryServerClient): Promise<boolean> {
+  // TODO change to actual version when known
+  // Note it is probably something 2.4.something
```
aeisenberg, line 29: Can this comment be removed now?

extensions/ql-vscode/src/upgrades.ts
```diff
@@ -14,79 +18,101 @@ import { upgradesTmpDir } from './run-queries';
...
 
 /**
- * Checks whether the given database can be upgraded to the given target DB scheme,
- * and whether the user wants to proceed with the upgrade.
- * Reports errors to both the user and the console.
- * @returns the `UpgradeParams` needed to start the upgrade, if the upgrade is possible and was confirmed by the user, or `undefined` otherwise.
+ * Check that we support non-destructive upgrades.
+ * 
+ * This requires 3 features. The ability to compile an upgrade sequence; The ability to 
+ * run a non-destructive upgrades as a query; the ability to specify a target when
```
aeisenberg, line 24: Three features are mentioned here, but this function only checks for one of them. Can you clarify in this comment what is actually being checked here.

extensions/ql-vscode/src/upgrades.ts
```diff
@@ -14,79 +18,101 @@ import { upgradesTmpDir } from './run-queries';
...
+export async function hasNondestructiveUpgradeCapabilities(qs: qsClient.QueryServerClient): Promise<boolean> {
+  // TODO change to actual version when known
+  // Note it is probably something 2.4.something
+  return semver.gte(await qs.cliServer.getVersion(), '2.4.1');
+}
+
+
+/**
+ * Compile a database upgrade sequence.
+ * Callers must check that this is valid with the current queryserver first.
```
aeisenberg, line 36: Any reason not to do this check inside this function? The version info is retrieved once and cached, so it is usually a fast operation.

extensions/ql-vscode/src/pure/messages.ts
```diff
@@ -460,6 +482,14 @@ export interface CompileUpgradeResult {
    */
   error?: string;
 }
+
+export interface SingleFileCompiledUpgradeResult extends CompileUpgradeResult {
+  /**
+   * The compiled upgrade.
+   */
+  compiledUpgrades?: SingleFileCompiledUpgrade;
```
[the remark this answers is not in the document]
alexet: This name has to be consistent with the java code. As it is a subtype of the old structure it is easier to not bother changing things.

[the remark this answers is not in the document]
aeisenberg: That's fine.

alexet:
You comment have made me realise that we can simplify the queryserver api, but it will mean that we need to wait for another codeql release to get this in.

alexet:
Ok this should be ready for re-review. Sorry I was so slow at coming back to this.

extensions/ql-vscode/src/pure/messages.ts
```diff
@@ -972,7 +1003,10 @@ export const checkUpgrade = new rpc.RequestType<WithProgressId<UpgradeParams>, C
  * Compile an upgrade script to upgrade a dataset.
  */
 export const compileUpgrade = new rpc.RequestType<WithProgressId<CompileUpgradeParams>, CompileUpgradeResult, void, void>('compilation/compileUpgrade');
-
+/**
+ * Compile an upgrade script to upgrade a dataset.
+ */
+export const compileUpgradeSequence = new rpc.RequestType<WithProgressId<CompileUpgradeSequenceParams>, CompiledUpgradeSequence, void, void>('compilation/compileUpgradeSequence');
```
aeisenberg, line 1009: Our naming convention is usually `XxxParams` and `XxxResult`.  Here it's `CompileUpgradeSequenceParams` and `CompiledUpgradeSequence`. Can you follow the same convention here?

aeisenberg: Hmmm...and I see below that `EvaluateQueriesParams` and `EvaluationComplete` does not follow the convention either. I still think it's useful to do so in order to know how these two interfaces are related.

I may fix that in a later PR.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -311,25 +314,58 @@ async function checkDbschemeCompatibility(
...
-        getUpgradesDirectories(scripts),
         progress,
         token
       );
     }
   }
 }
 
+function reportNoUpgradePath(query: QueryInfo) {
+  throw new Error(`Query ${query.program.queryPath} expects database scheme ${query.queryDbscheme}, but the current database has a different scheme, and no database upgrades are available. The current database scheme may be newer than the CodeQL query libraries in your workspace. Please try using a newer version of the query libraries.`);
```
aeisenberg, line 334: Nit: Consider emphasizing the proposed solution so it's not hidden in a wall of text.

```suggestion
  throw new Error(`Query ${query.program.queryPath} expects database scheme ${query.queryDbscheme}, but the current database has a different scheme, and no database upgrades are available. The current database scheme may be newer than the CodeQL query libraries in your workspace.\n\nPlease try using a newer version of the query libraries.`);
```

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -516,67 +552,78 @@ export async function compileAndRunQueryAgainstDatabase(
   }
 
   const query = new QueryInfo(qlProgram, db, packConfig.dbscheme, quickEvalPosition, metadata, templates);
-  await checkDbschemeCompatibility(cliServer, qs, query, progress, token);
 
-  let errors;
+  const upgradeDir = tmp.dirSync({ dir: upgradesTmpDir.name });
```
aeisenberg, line 556: Avoid using sync functions in user code. You can use the `tmp-promise` package to make an async call to creating a  tmp directory. After changing the import statement, all you need to do is this:

```suggestion
  const upgradeDir = await tmp.dir({ dir: upgradesTmpDir.name });
```

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -516,67 +552,78 @@ export async function compileAndRunQueryAgainstDatabase(
...
     }
-    if (quickEval && formattedMessages.length <= 3) {
-      showAndLogErrorMessage('Quick evaluation compilation failed: \n' + formattedMessages.join('\n'));
+
+    if (errors.length === 0) {
+      const result = await query.run(qs, upgradeQlo, progress, token);
+      if (result.resultType !== messages.QueryResultType.SUCCESS) {
+        const message = result.message || 'Failed to run query';
+        logger.log(message);
+        helpers.showAndLogErrorMessage(message);
```
aeisenberg, line 580: Nit: should already be imported directly.

```suggestion
        showAndLogErrorMessage(message);
```

extensions/ql-vscode/src/upgrades.ts
```diff
@@ -15,79 +19,95 @@ import { upgradesTmpDir } from './run-queries';
...
-  try {
-    qs.logger.log('Checking database upgrade...');
-    checkUpgradeResult = await checkDatabaseUpgrade(qs, params, progress, token);
-  }
-  catch (e) {
-    throw new Error(`Database cannot be upgraded: ${e}`);
-  }
-  finally {
-    qs.logger.log('Done checking database upgrade.');
+  if (!hasNondestructiveUpgradeCapabilities(qs)) {
```
aeisenberg, line 46: Nice...code scanning worked here. This expression needs an `await`.

extensions/ql-vscode/src/upgrades.ts
```diff
@@ -15,79 +19,95 @@ import { upgradesTmpDir } from './run-queries';
...
+    throw new Error('The version of codeql is to old to run non-destructive upgrades.');
```
aeisenberg, line 47:
```suggestion
    throw new Error('The version of codeql is to old too run non-destructive upgrades.');
```

aeisenberg (changes requested):
I think the only major issue is the missing `await` clause. Everything else is just wording and naming changes.

aeisenberg (approved):

PR #689, opened 2020-11-23
Add workflow_dispatch to release workflow

aeisenberg:
This will allow us to create easily accessible artifacts for any branch.

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,19 +6,15 @@
 
 name: Release
 on:
-  push:
-    # Path filters are not evaluated for pushes to tags.
-    # (source: https://help.github.com/en/github/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onpushpull_requestpaths)
-    # So this workflow is triggered in the following events:
-    # - Release event: a SemVer tag, e.g. v1.0.0 or v1.0.0-alpha, is pushed
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+*'
-    # OR
-    # - Test event: this file is modified on a branch in the main repo containing `/actions/` in the name.
-    branches:
-      - '**/actions/**'
+  pull_request:
     paths:
       - '**/workflows/release.yml'
+
+  workflow_dispatch:
+
+  push:
+    tags:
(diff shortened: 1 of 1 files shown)


adityasharad (approved):
Good idea.
Shall we also remove the `push: branches: '**/actions/**'`? That was a hack I put in to work around the absence of workflow dispatch at the time.

aeisenberg:
Sure. I can remove that.

.github/workflows/release.yml
```diff
@@ -6,17 +6,14 @@
...
     # So this workflow is triggered in the following events:
     # - Release event: a SemVer tag, e.g. v1.0.0 or v1.0.0-alpha, is pushed
     tags:
       - 'v[0-9]+.[0-9]+.[0-9]+*'
-    # OR
-    # - Test event: this file is modified on a branch in the main repo containing `/actions/` in the name.
-    branches:
-      - '**/actions/**'
     paths:
       - '**/workflows/release.yml'
```
adityasharad, line 18: Sorry, realised this `paths` needs to be removed along with the `branches`, OR even better, we turn it into a `pull_request: paths: '**/workflows/release.yml'`. Then we can test PRs that change the workflow itself without running on every push.

aeisenberg: Yep. That makes sense.

aeisenberg:
Wait...pull request won't work since this is using a secret from the parent repo. Will running on push work at all? Unless it's running in the parent. Maybe we can use a different event...

adityasharad (approved):
Much nicer than where we started. Thanks for the patience!

aeisenberg:
`pull_request_target` is what we need, I think, but concerned about privacy implications of exfiltrating secrets. How does that work?

adityasharad:
`pull_request` is fine. I don't expect us to make that secret available and run the release workflow on PRs from forks. If we want to update the release workflow we should open a PR from a branch in the parent repo.

aeisenberg:
This workflow will fail if it is run on `pull_request`, so there's no point in running at all for that event.

aeisenberg:
I'm trying the `pull_request_target` right now. If this works, then

PR #692, opened 2020-11-26
Fix dubious index check

max-schaefer:
I'm pretty sure this is meant to check `folderIndex`, not `index`.

Can you advise on how to add a test for this? (Of course it's fine if you want to just take over the PR instead.)

## Checklist


aeisenberg:
Thanks for creating this PR. I don't think we have any unit tests for this part of the code. I don't mind taking this over and making one. The change looks correct.

aeisenberg:
This slipped my mind yesterday. I'll check today.

aeisenberg:
A bit more complex than I was hoping to create these unit tests, but as we go forward, it should be easier to write them since we can base new ones off of existing tests.

aeisenberg:
@max-schaefer thanks for getting me to create these unit tests. They uncovered a bug!

adityasharad (approved):
LGTM, and thanks for the detailed tests! One question about test organisation.

aeisenberg:
Did you have a question? I don't see one posted.

adityasharad:
I did write one inline, and it seems to have been lost when I submitted the review...

Some of the tests are independent of the workspace, specifically the ones for URI processing. I wonder whether those could remain as `no-workspace` tests where we don't have to mock the `DatabaseManager` or the filesystem.

aeisenberg:
We could do that. Though I wonder what the benefit would be. I do think it's nice to keep related tests together even if they don't all require a workspace.

adityasharad:
My thinking was that they are unit tests that shouldn't rely on any properties of the workspace, and I'd like them to continue being independent of the workspace.
However we cannot use the `vscode` API outside the VS Code integration test harness anyway, so a true separation between unit and integration tests with respect to VS Code APIs is impossible anyway. Let's stick with this.

aeisenberg:
Previously, they were part of the `no-workspace` tests, which still run in a vscode harness, so there's not much difference between before and now. The only difference is that the tests now run in a context where they have access to a workspace (not needed for all tests).

Issue #676, opened 2020-11-09
Integration tests to create for the extension


aeisenberg:
Once #645, we will have the option of adding integration tests to the extension. This issue will keep a list of all the integration tests we want to create. And over time, we will implement all:

* [ ] AST Viewer works and displays some results
* [ ] Query can run and create a results web view that displays results
* [ ] View DIL 
* [ ] View Logs
* databases
    * [ ] Add / Remove from folder
    * [ ] Add / Remove from archive
    * [ ] Add / Remove from LGTM
    * [ ] Add / Remove from URL
* [ ] Compare queries shows some results

aeisenberg:
Uploading a simple database for use in integration tests. This database zip file should not be removed.

[simple-db.zip](https://github.com/github/vscode-codeql/files/5586722/simple-db.zip)

PR #694, opened 2020-11-30
Pass computeDefaultStrings to query server when compiling queries

aeisenberg:
## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Fix bug when removing databases where sometimes the source folder would not be removed from the workspace or the database files would not be removed from the workspace storage location. [#692](https://github.com/github/vscode-codeql/pull/692)
+- Always pass the `computeDefaultStrings` flag to query compilation commands. This will change the output of some queries. [#694](https://github.com/github/vscode-codeql/pull/694)
 
 ## 1.3.7 - 24 November 2020
 
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [UNRELEASED]
 
 - Fix bug when removing databases where sometimes the source folder would not be removed from the workspace or the database files would not be removed from the workspace storage location. [#692](https://github.com/github/vscode-codeql/pull/692)
-- Always pass the `computeDefaultStrings` flag to query compilation commands. This will change the output of some queries. [#694](https://github.com/github/vscode-codeql/pull/694)
+- Query results with no string representation will now be displayed with placeholder text in query results. Previously, they were omitted. [#694](https://github.com/github/vscode-codeql/pull/694)
 
 ## 1.3.7 - 24 November 2020
 
--- a/extensions/ql-vscode/src/vscode-tests/no-workspace/query-test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/no-workspace/query-test.ts
@@ -168,6 +168,7 @@ describe('using the query server', function() {
             localChecking: false,
             noComputeGetUrl: false,
             noComputeToString: false,
+            computeDefaultStrings: true
           },
           queryToCheck: qlProgram,
           resultPath: COMPILED_QUERY_PATH,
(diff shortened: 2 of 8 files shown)


extensions/ql-vscode/src/vscode-tests/no-workspace/query-test.ts
```diff
@@ -168,6 +168,7 @@ describe('using the query server', function() {
             localChecking: false,
             noComputeGetUrl: false,
             noComputeToString: false,
+            computeDefaultStrings: true
```
adityasharad, line 171: Can you add a new `QueryTestCase` to exercise this?
The simplest I can think of is a query that looks like this:

```ql
newtype TUnit = MkUnit()

class Unit extends TUnit {
  Unit() { this = MkUnit() }

  string toString() { none() }
}

from Unit u
select u
```

Existing behaviour: this produces 0 results.
New behaviour: this produces 1 result, whose text is `(no string representation)`.

adityasharad (commented):
Nice. I suggest adding a query integration test.

adityasharad:
Worth adding a short changelog entry too, since it's actually user-visible. Something like "Query results without a string representation were previously not computed. They will now be computed and displayed with a default placeholder string."

extensions/ql-vscode/src/vscode-tests/no-workspace/query-test.ts
aeisenberg: I added that integration test to the list. #676

Currently, even the basic query test is failing. Need to work on that.

adityasharad: Sure. For this particular test, we already have tests in `query-test` to run a simple query and get the raw results out, so I suggest adding a case in this PR if possible without needing the new integration test machinery.

aeisenberg: Ah...I see. OK. I will add that tomorrow.

aeisenberg: So, those tests have never been running in CI. Our scripts match on `*.test.ts`, but that file is `query-test.ts`. I will fix.

adityasharad: 😱

aeisenberg:
Test is added and and query tests are now running in CI.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Fix bug when removing databases where sometimes the source folder would not be removed from the workspace or the database files would not be removed from the workspace storage location. [#692](https://github.com/github/vscode-codeql/pull/692)
+- Always pass the `computeDefaultStrings` flag to query compilation commands. This will change the output of some queries. [#694](https://github.com/github/vscode-codeql/pull/694)
```
adityasharad, line 6: I think I put this into an earlier comment I can't find - I'd prefer to mention what the user-facing change here is, i.e. query results with no string representation were previously omitted, and will now be displayed with placeholder text.

extensions/ql-vscode/src/vscode-tests/no-workspace/query.test.ts
```diff
@@ -61,21 +59,27 @@ type QueryTestCase = {
...
   {
-    queryPath: path.join(__dirname, '../data/query.ql'),
+    queryPath: path.join(baseDir, 'query.ql'),
     expectedResultSets: {
       '#select': [[42, 3.14159, 'hello world', true]]
     }
   },
   {
-    queryPath: path.join(__dirname, '../data/multiple-result-sets.ql'),
+    queryPath: path.join(baseDir, 'compute-default-strings.ql'),
```
adityasharad, line 68: Looks great! The test failures are now from `files.test.ts`, which expected to find only 2 queries in this folder and now found the third one.

extensions/ql-vscode/CHANGELOG.md
aeisenberg: Sure. I didn't quite know how to say it. This sounds good.

extensions/ql-vscode/src/vscode-tests/no-workspace/query.test.ts
aeisenberg: Fixed now. Also, the sort was wrong also.

adityasharad (approved):

PR #697, opened 2020-12-03
Add a language label next to databases in the UI

### LINKED ISSUE #321: Add a marker of the target language of databases in the tree view
This is a subtask of #318.

aeisenberg:
This change will only work on databases created by cli >= 2.4.1. In that
version, a new `primaryLanguage` field in the `codeql-database.yml`
file. We use this property as the language.

This change also includes a refactoring of the logic around extracting
database information heuristically based on file location. As much
as possible, it is extracted to the `helpers` module. Also, the
initial quick query text is generated based on the language (if known)
otherwise it falls back to the old style of generation.

Fixes #321

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -50,6 +50,7 @@ export interface DbInfo {
   sourceArchiveRoot: string;
   datasetFolder: string;
   logsFolder: string;
+  primaryLanguage: string;
 }
 
 /**
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -125,7 +125,7 @@ export class CodeQLCliServer implements Disposable {
   private static CLI_VERSION_WITH_DECOMPILE_KIND_DIL = new SemVer('2.3.0');
 
   /**
-   * CLI version where --kind=DIL was introduced
+   * CLI version where languages are exposed during a `codeql resolve database` command.
    */
   private static CLI_VERSION_WITH_LANGUAGE = new SemVer('2.4.1');
 
@@ -716,7 +716,7 @@ export class CodeQLCliServer implements Disposable {
     return (await this.getVersion()).compare(CodeQLCliServer.CLI_VERSION_WITH_DECOMPILE_KIND_DIL) >= 0;
   }
 
-  public async supportsLangaugeName() {
+  public async supportsLanguageName() {
     return (await this.getVersion()).compare(CodeQLCliServer.CLI_VERSION_WITH_LANGUAGE) >= 0;
   }
 
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -50,7 +50,7 @@ export interface DbInfo {
   sourceArchiveRoot: string;
   datasetFolder: string;
   logsFolder: string;
-  primaryLanguage: string;
+  languages: string[];
 }
 
 /**
@@ -123,6 +123,11 @@ export class CodeQLCliServer implements Disposable {
    * CLI version where --kind=DIL was introduced
    */
   private static CLI_VERSION_WITH_DECOMPILE_KIND_DIL = new SemVer('2.3.0');
+
+  /**
+   * CLI version where --kind=DIL was introduced
+   */
+  private static CLI_VERSION_WITH_LANGUAGE = new SemVer('2.4.1');
 
   /** The process for the cli server, or undefined if one doesn't exist yet */
   process?: child_process.ChildProcessWithoutNullStreams;
@@ -690,7 +695,7 @@ export class CodeQLCliServer implements Disposable {
   }
 
   async generateDil(qloFile: string, outFile: string): Promise<void> {
-    const extraArgs = (await this.getVersion()).compare(CodeQLCliServer.CLI_VERSION_WITH_DECOMPILE_KIND_DIL) >= 0
+    const extraArgs = await this.supportsDecompileDil()
       ? ['--kind', 'dil', '-o', outFile, qloFile]
       : ['-o', outFile, qloFile];
     await this.runCodeQlCliCommand(
@@ -705,6 +710,14 @@ export class CodeQLCliServer implements Disposable {
       this._version = await this.refreshVersion();
     }
     return this._version;
+  }
+
+  private async supportsDecompileDil() {
+    return (await this.getVersion()).compare(CodeQLCliServer.CLI_VERSION_WITH_DECOMPILE_KIND_DIL) >= 0;
+  }
+
+  public async supportsLangaugeName() {
+    return (await this.getVersion()).compare(CodeQLCliServer.CLI_VERSION_WITH_LANGUAGE) >= 0;
   }
 
   private async refreshVersion() {
--- a/extensions/ql-vscode/src/databases.ts
+++ b/extensions/ql-vscode/src/databases.ts
@@ -831,8 +831,10 @@ export class DatabaseManager extends DisposableObject {
   }
 
   private async getPrimaryLanguage(dbPath: string) {
-    if (!(await this.cli.supportsLangaugeName())) {
-      // return undefined so that we continually recalculate until the cli version is bumped
+    if (!(await this.cli.supportsLanguageName())) {
+      // return undefined so that we recalculate on restart until the cli is at a version that
+      // supports this feature. This recalculation is cheap since we avoid calling into the cli
+      // unless we know it can return the langauges property.
       return undefined;
     }
     const dbInfo = await this.cli.resolveDatabase(dbPath);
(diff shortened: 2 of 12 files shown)


extensions/ql-vscode/src/helpers.ts
```diff
@@ -464,3 +457,72 @@ export class CachedOperation<U> {
...
+
+/**
+ * The following functions al heuristically determine metadata about databases.
+ */
+
+const dbSchemeToLanguage = {
+  'semmlecode.javascript.dbscheme': 'javascript',
+  'semmlecode.cpp.dbscheme': 'cpp',
+  'semmlecode.dbscheme': 'java',
+  'semmlecode.python.dbscheme': 'pyhton',
```
adityasharad, line 471:
```suggestion
  'semmlecode.python.dbscheme': 'python',
```

extensions/ql-vscode/src/helpers.ts
```diff
@@ -464,3 +457,72 @@ export class CachedOperation<U> {
...
+  ]));
+
+  return !!((a || b) && c);
+}
+
+export function isLikelyDbLanguageFolder(dbPath: string) {
+  return !!path.basename(dbPath).startsWith('db-');
+}
+
+export async function getPrimaryLanguage(root: string) {
```
adityasharad, line 517: Instead of parsing the config file, make a call to `codeql resolve database --format json` and get the first value in the `languages` field.

adityasharad (changes requested):
Refactoring looks good. I suggest getting the language from a CLI command instead of the YAML.

extensions/ql-vscode/src/helpers.ts
hmakholm: Agreed. Especially because the open-sourced VSCode extension is our showcase for the Right way to do things.

hmakholm: That will also mean that the extension will not need its own `dbSchemeToLanguage` lookup table, since the `codeql resolve database` now already tries a dbscheme-based heuristic for language detection.

aeisenberg: The reason I chose not to do that is that currently, there is no cli available on `openDatabase`. Also, making a filesystem check will be faster than starting up a jvm to get this information.

I could certainly stuff a reference to the cli into the constructor. But I'd prefer not to do that for the reasons above unless the approach I'm taking would produce incorrect results.

hmakholm: I would lean towards considering "people who write their own IDE integrations for CodeQL think this is the right thing to do, and proceed to bake half-understood internal knowledge of database internals into their own work" to be an _incorrect result_. :-)

On the other hand, I don't have the insight to appreciate the possible problems in dealing with "there is no cli available on `openDatabase`". Why is that?

aeisenberg: The `cli` is a singleton that gets passed around when needed. The `DatabaseManager` does not hold a reference to the `cli`. I would need to change that in order to call a cli command here. It's not a big change, but it is an architectural one.

Also, I still need `dbSchemeToLanguage` as a fallback in case the database was not created by a modern CLI. I will add a comment around using that so it's known that this is not recommended.

So, the question is, should I be getting the database label from the `codeql-database.yml` or from `database resolve`? I'm not sure what your comment was on this or about the `dbSchemeToLanguage`.

hmakholm:
> Also, I still need `dbSchemeToLanguage` as a fallback in case the database was not created by a modern CLI. I will add a comment around using that so it's known that this is not recommended.

If the CLI is new enough to return `languages` from `resolve database` at all, it will _already_ do its own dbscheme-based heuristics when it's asked about old databases.

> So, the question is, should I be getting the database label from the `codeql-database.yml` or from `database resolve`? I'm not sure what your comment was on this or about the `dbSchemeToLanguage`.

I think the extension should rely exclusively on `codeql resolve database`. I would really want to avoid exposing hacky heuristics in the open-sourced extension. That's why I suggested adding a feature to `resolve database` in the first place.

aeisenberg:
Two questions:

1. What happens if the CLI is _not_ new enough to return a languages field? That's one scenario I am trying to handle. I guess in this case we could just avoid returning anything and this would be an ecouragement to migrate to a newer version. But we still need to use `dbscheme` heuristics for generating the import statements of quick queries
2. Are you saying that even if there is no `primaryLanguage` field in the `codeql-database.yml`, the database resolve command will still return the languages list?

hmakholm:
Hmmyes, an existing quick-query feature of course needs to keep working.

I was thinking more in terms of the _new_ feature with showing language tags for databases in the UI. Here I think it is reasonably to expect that you need both a new extension and a CLI that supports the feature.

aeisenberg:
I think this addresses all the concerns. Look at the second commit for the changes. I rebased to remove merge conflicts.

extensions/ql-vscode/src/helpers.ts
```diff
@@ -464,3 +457,74 @@ export class CachedOperation<U> {
...
+
+/**
+ * The following functions al heuristically determine metadata about databases.
+ */
+
+/**
+ * Note that this heuristic is only being used for backwards compatibility with
+ * CLI versions before the langauge name was introduced to dbInfo. Implementations
+ * that do not require backwards compatibility should call
+ * `cli.CodeQLCliServer.resolveDatabase` and use the first entry in the
```
hmakholm, line 471: `Implementations` sounds like an "off" word to use here. Applications? Features?

hmakholm (approved):
I don't feel qualified to review the TS part, but I'm happy with the general approach to finding the language now.

extensions/ql-vscode/src/cli.ts
```diff
@@ -123,6 +124,11 @@ export class CodeQLCliServer implements Disposable {
    */
   private static CLI_VERSION_WITH_DECOMPILE_KIND_DIL = new SemVer('2.3.0');
 
+  /**
+   * CLI version where --kind=DIL was introduced
```
adityasharad, line 128: Copy-paste comment.

extensions/ql-vscode/src/cli.ts
```diff
@@ -706,6 +712,14 @@ export class CodeQLCliServer implements Disposable {
     return this._version;
   }
 
+  private async supportsDecompileDil() {
+    return (await this.getVersion()).compare(CodeQLCliServer.CLI_VERSION_WITH_DECOMPILE_KIND_DIL) >= 0;
+  }
+
+  public async supportsLangaugeName() {
```
adityasharad, line 719: typo

extensions/ql-vscode/src/cli.ts
```diff
@@ -123,6 +124,11 @@ export class CodeQLCliServer implements Disposable {
    */
   private static CLI_VERSION_WITH_DECOMPILE_KIND_DIL = new SemVer('2.3.0');
 
+  /**
+   * CLI version where --kind=DIL was introduced
+   */
+  private static CLI_VERSION_WITH_LANGUAGE = new SemVer('2.4.1');
```
adityasharad, line 130: Starting to reach the point where we might want an enum of `SemVer`s for different features :) (no need to do this now)

extensions/ql-vscode/src/databases.ts
```diff
@@ -804,6 +829,15 @@ export class DatabaseManager extends DisposableObject {
     }
     return false;
   }
+
+  private async getPrimaryLanguage(dbPath: string) {
+    if (!(await this.cli.supportsLangaugeName())) {
+      // return undefined so that we continually recalculate until the cli version is bumped
```
adityasharad, line 835: What do you mean by recalculate?

extensions/ql-vscode/src/helpers.ts
```diff
@@ -464,3 +457,74 @@ export class CachedOperation<U> {
...
+  'semmlecode.python.dbscheme': 'python',
+  'semmlecode.csharp.dbscheme': 'csharp',
+  'go.dbscheme': 'go'
+};
+
+/**
+ * Returns the initial contents for an empty query, based on the language of the selected
+ * databse.
+ *
+ * First try to get the contents text based on language. if that fails, try to get based on
```
adityasharad, line 490:
```suggestion
 * First try to use the given language name. If that doesn't exist, try to infer it based on
```

extensions/ql-vscode/src/helpers.ts
adityasharad: Features sounds good.

extensions/ql-vscode/src/databases.ts
```diff
@@ -502,18 +519,16 @@ export class DatabaseManager extends DisposableObject {
     progress: ProgressCallback,
     token: vscode.CancellationToken,
     uri: vscode.Uri,
-    options?: DatabaseOptions
   ): Promise<DatabaseItem> {
-
     const contents = await resolveDatabaseContents(uri);
-    const realOptions = options || {};
```
adityasharad, line 509: What happened to this? Were we never passing in `options` to begin with? Can't that happen when loading a persisted item?

adityasharad (commented):
A few minor questions.

extensions/ql-vscode/src/databases.ts
aeisenberg: No, we were never passing in `options`. We load persisted items from `loadPersistedState()`. So, it is an entirely different path.

adityasharad (approved):

PR #698, opened 2020-12-05
Re-enable the queries cli test

aeisenberg:
* Requires that QL_PATH environment variable is set and points to a
  checkout of github/codeql
* Adds the `quiet` flag to the cli. When set, this flag will prevent
  some modal dialogs from disrupting the flow. Currently, we only ensure
  that the upgrades dialog is avoided.
* Update the main.yml workflow to checkout the codeql repo

Added a second commit, which does the following:

Move query.test.ts to the cli integration tests

* Now query.test.ts runs on multiple cli versions
* Removed most `dispose` calls in cli tests because each test shares the
  same instance of the extension and all of its properties. So, we
  shouldn't be disposing until the last test completes. It's likely that
  we will need to be more careful about cleaning up state between test
  runs, but we haven't hit that yet and this can happen in a later
  commit.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -124,6 +124,7 @@ export interface CodeQLExtensionInterface {
   readonly distributionManager: DistributionManager;
   readonly databaseManager: DatabaseManager;
   readonly databaseUI: DatabaseUI;
+  readonly dispose: () => void;
 }
 
 /**
@@ -679,7 +680,10 @@ async function activateWithInstalledDistribution(
     qs,
     distributionManager,
     databaseManager: dbm,
-    databaseUI
+    databaseUI,
+    dispose: () => {
+      ctx.subscriptions.forEach(d => d.dispose());
+    }
   };
 }
 
--- a/extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
@@ -18,9 +18,7 @@ import { QueryServerClient } from '../../queryserver-client';
 /**
  * Integration tests for queries
  */
-// TODO: Not currently able to run this because we do not have access to the javascript qlpack.
-// Need to download ql libraries separately.
-xdescribe('Queries', function() {
+describe.only('Queries', function() {
   this.timeout(20000);
 
   let dbItem: DatabaseItem;
@@ -40,6 +38,7 @@ xdescribe('Queries', function() {
         databaseManager = extension.databaseManager;
         cli = extension.cliServer;
         qs = extension.qs;
+        cli.quiet = true;
       } else {
         throw new Error('Extension not initialized. Make sure cli is downloaded and installed properly.');
       }
--- a/extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
@@ -13,13 +13,18 @@ import { importArchiveDatabase } from '../../databaseFetcher';
 import { compileAndRunQueryAgainstDatabase } from '../../run-queries';
 import { CodeQLCliServer } from '../../cli';
 import { QueryServerClient } from '../../queryserver-client';
+import { skipIfNoCodeQL } from '../ensureCli';
 
 
 /**
  * Integration tests for queries
  */
-describe.only('Queries', function() {
+describe('Queries', function() {
   this.timeout(20000);
+
+  before(function() {
+    skipIfNoCodeQL(this);
+  });
 
   let dbItem: DatabaseItem;
   let databaseManager: DatabaseManager;
@@ -67,8 +72,6 @@ describe.only('Queries', function() {
 
   afterEach(() => {
     try {
-      databaseManager?.dispose();
-      qs?.dispose();
       sandbox.restore();
     } catch (e) {
       fail(e);
--- a/extensions/ql-vscode/src/vscode-tests/cli-integration/query.test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/cli-integration/query.test.ts
@@ -8,9 +8,11 @@ import { CancellationTokenSource } from 'vscode-jsonrpc';
 import * as messages from '../../pure/messages';
 import * as qsClient from '../../queryserver-client';
 import * as cli from '../../cli';
-import { ProgressReporter, Logger } from '../../logging';
 import { ColumnValue } from '../../pure/bqrs-cli-types';
-import { FindDistributionResultKind } from '../../distribution';
+import { extensions } from 'vscode';
+import { CodeQLExtensionInterface } from '../../extension';
+import { fail } from 'assert';
+import { skipIfNoCodeQL } from '../ensureCli';
 
 
 const baseDir = path.join(__dirname, '../../../test/data');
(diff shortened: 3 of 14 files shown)


extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
```diff
@@ -120,7 +122,10 @@ function getLaunchArgs(dir: TestDir) {
       ];
 
     case TestDir.CliIntegration:
-      break;
+      return [
+        path.resolve(__dirname, '../../test/data'),
+        process.env.TEST_CODEQL_PATH!
```
adityasharad, line 127: This is a list of workspace folders right?

extensions/ql-vscode/src/vscode-tests/ensureCli.ts
```diff
@@ -105,6 +106,15 @@ export async function ensureCli(useCli: boolean) {
   }
 }
 
+/**
+ * Heuristically determines if the codeql libraries are installed in this
+ * workspace. Looks for the existance of a folder whose path ends in `/codeql`
+ */
+export function hasCodeQL() {
+  const folders = workspace.workspaceFolders;
+  return !!folders?.some(folder => folder.uri.path.endsWith('/codeql'));
```
adityasharad, line 115: A more principled way would be to run `codeql resolve qlpacks` and make sure we've picked up the packs in the `codeql` repo. I think that is worth doing as an integration test, but perhaps not necessary here just to verify your test setup.

extensions/ql-vscode/src/vscode-tests/cli-integration/query.test.ts
```diff
@@ -81,65 +83,36 @@ const queryTestCases: QueryTestCase[] = [
 
 describe('using the query server', function() {
   before(function() {
-    if (process.env['CODEQL_PATH'] === undefined) {
-      console.log('The environment variable CODEQL_PATH is not set. The query server tests, which require the CodeQL CLI, will be skipped.');
+    if (!hasCodeQL()) {
+      console.log('The CodeQL libraries are not available as a folder in this workspace. To fix: checkout the github/codeql repository and set the TEST_CODEQL_PATH envirnment variable to the checked out directory.');
```
adityasharad, line 87: typo: environment

extensions/ql-vscode/src/vscode-tests/cli-integration/query.test.ts
```diff
@@ -81,65 +83,36 @@ const queryTestCases: QueryTestCase[] = [
...
   let cliServer: cli.CodeQLCliServer;
   const queryServerStarted = new Checkpoint<void>();
-  after(() => {
-    qs?.dispose();
-    cliServer?.dispose();
+
+  beforeEach(async () => {
+    try {
+      const extension = await extensions.getExtension<CodeQLExtensionInterface | {}>('GitHub.vscode-codeql')!.activate();
+      if ('databaseManager' in extension) {
```
adityasharad, line 103: Maybe check for `cliServer`? It's a bit odd you check for `databaseManager` and then don't use it.

extensions/ql-vscode/src/vscode-tests/cli-integration/query.test.ts
```diff
@@ -81,65 +83,36 @@ const queryTestCases: QueryTestCase[] = [
 
 describe('using the query server', function() {
   before(function() {
-    if (process.env['CODEQL_PATH'] === undefined) {
-      console.log('The environment variable CODEQL_PATH is not set. The query server tests, which require the CodeQL CLI, will be skipped.');
+    if (!hasCodeQL()) {
```
adityasharad, line 86: I agree it would be nice to separate the tests that need the standard libraries from those that don't, but I also agree this can be done later.

extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
```diff
@@ -13,16 +13,22 @@ import { importArchiveDatabase } from '../../databaseFetcher';
...
  */
-// TODO: Not currently able to run this because we do not have access to the javascript qlpack.
-// Need to download ql libraries separately.
-xdescribe('Queries', function() {
+describe('Queries', function() {
   this.timeout(20000);
 
+  before(function() {
+    if (!hasCodeQL()) {
+      console.log('The CodeQL libraries are not available as a folder in this workspace. To fix: checkout the github/codeql repository and set the TEST_CODEQL_PATH envirnment variable to the checked out directory.');
```
adityasharad, line 27: typo: environment

extensions/ql-vscode/src/vscode-tests/cli-integration/query.test.ts
adityasharad: Actually I think this particular set of tests doesn't need the standard libraries.

extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
```diff
@@ -68,8 +75,7 @@ xdescribe('Queries', function() {
 
   afterEach(() => {
     try {
-      databaseManager?.dispose();
-      qs?.dispose();
+      // dispose?.();
```
adityasharad, line 78: Remove?

extensions/ql-vscode/src/extension.ts
```diff
@@ -679,7 +680,11 @@ async function activateWithInstalledDistribution(
     qs,
     distributionManager,
     databaseManager: dbm,
-    databaseUI
+    databaseUI,
+    dispose: () => {
+      ctx.subscriptions.forEach(disposable => disposable.dispose());
+    }
```
adityasharad, line 686: Why do we need to explicitly write this? I expect VS Code to do something like this and dispose all `subscriptions` when the extension is deactivated. Can we make such a deactivate call during test teardown?

extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts
aeisenberg: Yes. I will add a comment here.

extensions/ql-vscode/src/vscode-tests/ensureCli.ts
aeisenberg: Sure. I'll add that.

extensions/ql-vscode/src/vscode-tests/cli-integration/query.test.ts
aeisenberg: Because this test now calls the cli in the same way as the rest of the extension, there is a qlpack in scope and the standard libraries are required.

I'm not sure if it's possible to run a query without a qlpack in scope inside of the extension.

extensions/ql-vscode/src/extension.ts
aeisenberg: Actually, we don't need this.

And, no, we cannot deactivate the extension during test teardown. We can only deactivate the extension a single time for all tests. This is _bad_ because we will be potentially sharing state across tests, but that appears to be the way it works.

This line above is left over from an attempt of mine to properly dispose after each test. But, when the disposal happened, the entire test process shut down early. So it's not possible to do proper cleanup after each test.

extensions/ql-vscode/src/vscode-tests/cli-integration/query.test.ts
adityasharad: These particular tests do not use the standard libraries though: there is no import statement. You do need a sample qlpack.yml file next to the test queries (I think we have one).
Happy to merge and see if we can weaken the requirement later.

adityasharad (approved):
Appreciate you pushing forward our integration testing capability!

extensions/ql-vscode/src/vscode-tests/cli-integration/query.test.ts
aeisenberg: Right...the qlpack being used here explicitly requires the javascript libs. We would need to use a different qlpack that doesn't require any dependent libraries and to do that, we'd need to put them in a different directory since this directory is shared with "real" queries that actually query a database.

adityasharad: Ohh yes this makes sense now.
- https://github.com/github/vscode-codeql/tree/main/extensions/ql-vscode/test/data is missing a `qlpack.yml`, and should have one that points to `test.dbscheme` and no other dependencies.
- https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/vscode-tests/cli-integration/data is missing a dbscheme file, and its `qlpack.yml` should point to that dbscheme file and no other dependencies.

aeisenberg: OK. Let me see if I can get that working. Also, not a fan that we have some tests in `ql-vscode/test` and others in `ql-vs-code/src/vscode-tests`.

PR #700, opened 2020-12-07
Save downloaded DB archives to disk before unzipping

### LINKED ISSUE #621: Importing a .zip database kills the extension
**Describe the bug**
Importing a .zip database kills the extension

**To Reproduce**
- Download the C/C++ database from https://lgtm.com/projects/g/torvalds/linux/ci/#ql
- Import it from vscode.

**Expected behavior**
The database is imported

### LINKED ISSUE #622: Unable to Download from LGTM for certain databases.
Certain databases cannot be imported directly from LGTM. They must first be saved to disk and then imported. An example is https://lgtm.com/projects/g/torvalds/linux/ci/#ql (the python database).

I discovered this when looking into #621.

I think this is because of how the zip file is constructed. Our current approach is to unzip the http request as it streams in, but this is more error prone if the central directory is not completely correct. See https://github.com/ZJONSSON/node-unzipper/issues/149

I may need to try another approach, where we first download the zip file to disk, and then unzip it into the extension.

aeisenberg:
This fixes two classes of DBs that can't be installed directly from
downloading:

1. DBs whose central directories do not align with their file headers.
   We need to download and save the entire archive  before we can read
   the central directory and use that to guide the unzipping.
2. Large DBs require too much memory so can't be downloaded and unzipped
   in a single stream.

We also add proper progress notifications to the download progress
monitor so users are aware of how many more MBs are left to download.

It's not yet possible to do the same for unzipping using the current
unzipper library, since unzipping using the central directory does not
expose a stream.

Fixes #621
Fixes #622

Integration tests are passing and added two more for progress monitoring.

Manually tested on the linux database from lgtm https://lgtm.com/projects/g/torvalds/linux/ci and it is working.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix bug when removing databases where sometimes the source folder would not be removed from the workspace or the database files would not be removed from the workspace storage location. [#692](https://github.com/github/vscode-codeql/pull/692)
 - Query results with no string representation will now be displayed with placeholder text in query results. Previously, they were omitted. [#694](https://github.com/github/vscode-codeql/pull/694)
 - Add a label for the language of a database in the databases view. This will only take effect for new databases created with the CodeQL CLI v2.4.1 or later. [#697](https://github.com/github/vscode-codeql/pull/697)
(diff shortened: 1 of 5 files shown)


alexet (approved):
LGTM. I tried to resolve the merge conflict in the ui but have realised it would have probably been cleaner for you to rebase instead. Feel free to remove my merge commit and rebase instead.

alexet:
Actually squash and merge will do the right thing.

aeisenberg:
Thanks for the review!

PR #703, opened 2020-12-11
Add missing call to `getPrimaryDbscheme` in `qlpackOfDatabase`

hvitved:
Without this line I get errors like `Could not find qlpack file for dbscheme ...` when using 'View AST'.


aeisenberg:
Thanks for creating this PR. I'll take a look on Monday. It's probably fine, but the tests need to be updated.

aeisenberg:
A small change and fixing the tests. After this it is good to merge.

aeisenberg (approved):

PR #705, opened 2020-12-14
Graph view for `@kind graph` queries

hvitved:
DO NOT MERGE

This is a prototype implementation of graph visualization in the extension (implemented in collaboration with @AlonaHlobina  and @aibaars). It works for queries marked as `@kind graph`, and we also implemented a 'View CFG' command that executes the query with ID `ide-contextual-queries/print-cfg` (if it exists).

This PR is not meant to be merged, but rather it can be used as inspiration for a "real" implementation. Until we have an official implementation, I plan to use the artifacts created via this PR.


aeisenberg:
@hvitved what is your plan for this PR? Are you using this yourself? It is a great idea, but I don't have the bandwidth to publicly release this. With a little bit of work, we can probably release this behind a feature flag.

hvitved:
> @hvitved what is your plan for this PR? Are you using this yourself? It is a great idea, but I don't have the bandwidth to publicly release this. With a little bit of work, we can probably release this behind a feature flag.

My plan was to keep this branch alive until we have proper support. Yes, I use it myself.

aeisenberg:
Thanks for the update. When I get a chance, I'll work on it to release it behind a feature flag.

hvitved:
Thanks @aeisenberg, that would be very useful. But please don't use the code in this branch as-is; there are many (or at least some) cases of "we have no idea what we are doing" :-)

nickrolfe:
This has been invaluable for me in my recent work debugging a QL regex parser, although it does get quite unwieldy with large graphs. Something more interactive that lets you collapse subtrees would be amazing.

If this does become the basis for a released feature, I should note that I've found a few issues:

- backslashes in node names are not rendered – it's possible some escaping is required
- the area of the window in which the graph is rendered does not seem to fill the available space – see screenshot below, where I zoomed in to demonstrate
- it seems to use the `semmle.order` property on the `nodes` predicate, rather than `edges`, to decide on the order in which child nodes are rendered, contrary to my expectations
- at one point I managed to produce a graph query that would cause the query results windows to be *completely empty* (i.e. not even showing the metadata header). After that, even running non-graph queries would not fix it, and I had to restart VS Code. I'm afraid I don't know how to reproduce this now.

image

aeisenberg:
I hope to get a chance to look at this next week. I'll be putting this feature behind the canary flag, so we won't advertise it widely. I'm planning on largely using what this PR as-is since it's shown to be useful already. Thanks again for putting this up and apologies it's taken over a year to look at this.

PR #708, opened 2020-12-15
Move some database commands to overflow menu

aeisenberg:
This commit moves the "Add *" and "Download *" database commands
from the buttons on top and into an overflow menu.

I believe this is easier to read since the icons were never fully
understandable and required hovering to read.

Also, renamed the commands to be more consistent. Now all begin
with "Add Database from".

Simple PR to change the layout of the view commands. I didn't know this was possible until I saw it in another extension. Will look like this:

_Extension_Development_Host__-_XXE_qlref_—_vscode-codeql-starter__Workspace_

I'm open to better wording than "Download Database from Internet".

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -196,7 +196,7 @@
       },
       {
         "command": "codeQLDatabases.chooseDatabaseFolder",
-        "title": "Choose Database from Folder",
+        "title": "Add Database from Folder",
         "icon": {
           "light": "media/light/folder-opened-plus.svg",
           "dark": "media/dark/folder-opened-plus.svg"
@@ -208,7 +208,7 @@
       },
       {
         "command": "codeQLDatabases.chooseDatabaseArchive",
-        "title": "Choose Database from Archive",
+        "title": "Add Database from Archive",
         "icon": {
           "light": "media/light/archive-plus.svg",
           "dark": "media/dark/archive-plus.svg"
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/package.json
```diff
@@ -208,23 +208,23 @@
...
+        "title": "Add Database from Archive",
         "icon": {
           "light": "media/light/archive-plus.svg",
           "dark": "media/dark/archive-plus.svg"
         }
       },
       {
         "command": "codeQLDatabases.chooseDatabaseInternet",
-        "title": "Download Database",
+        "title": "Add Database from Internet",
```
adityasharad, line 219: "from URL" is another option.

extensions/ql-vscode/package.json
```diff
@@ -398,22 +398,22 @@
         {
           "command": "codeQLDatabases.chooseDatabaseFolder",
           "when": "view == codeQLDatabases",
-          "group": "navigation"
+          "group": "1_open@1"
```
adityasharad, line 401: Is it possible to create your own group below an icon, rather than the overflow menu? I think decluttering the database view is a good idea, but I am concerned users will not know to look in the overflow menu for 'Add Database' commands, whereas an 'Add Database From...' submenu may be more obvious.

aeisenberg: I don't know if that's possible. Let me look.

aeisenberg: No. It's not. https://code.visualstudio.com/api/extension-guides/tree-view#view-actions

I could put the actions in both places, but that seems worse. I feel like I would want to talk to a UX engineer for this.

yoff:
It does declutter a bit, but it seems a lot less discoverable than before...

jbj:
As a user of the extension, I prefer the existing behaviour. I've only ever used two of those buttons, but I've used them a lot: Add Database from {Archive, Folder}.

aeisenberg:
No consensus on this. Not going to go ahead with this change.

PR #710, opened 2020-12-15
Fix set label command on history items

### LINKED ISSUE #709: Set Label on a query history item is broken
The "Set label" command does not affect the UI. The problem is that the tree items are being cached and even though we are updating the label on the query item, the tree item is not reflecting this change.

This bug was introduced in #598. I have a different way to address #598 that will also fix the label.

aeisenberg:
This removes the cached treeItem that is a property of the
completedQuery. We should not be caching them since they are cached by
the vscode api itself. Instead, we should recreate whenever requested.

Also, this change fixes #598 in a new way. Instead of adding the
context to the cached treeItem, we simply refresh only the item that has
changed. This is a fast operation.

Fixes #709.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -52,14 +52,10 @@ const SHOW_QUERY_TEXT_QUICK_EVAL_MSG = `\
  */
 const FAILED_QUERY_HISTORY_ITEM_ICON = 'media/red-x.svg';
 
-interface QueryHistoryDataProvider extends vscode.TreeDataProvider<CompletedQuery> {
-  updateTreeItemContextValue(element: CompletedQuery): Promise<void>;
-}
-
 /**
  * Tree data provider for the query history view.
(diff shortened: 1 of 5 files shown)


aeisenberg:
Converting to draft since I want to add a few unit tests.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -157,8 +144,8 @@ class HistoryTreeDataProvider extends DisposableObject implements QueryHistoryDa
     return this.history;
   }
 
-  refresh() {
-    this._onDidChangeTreeData.fire(undefined);
+  refresh(CompletedQuery?: CompletedQuery) {
```
adityasharad, line 147: Lowercase the variable to distinguish them.

adityasharad (commented):
Looks very reasonable. I'll wait for tests :)

adityasharad (approved):

PR #722, opened 2021-01-11
Fix Code Scanning warnings

aeisenberg:
Removes unnecessary code.

### DIFF
--- a/extensions/ql-vscode/src/test-adapter.ts
+++ b/extensions/ql-vscode/src/test-adapter.ts
@@ -220,14 +220,13 @@ export class QLTestAdapter extends DisposableObject implements TestAdapter {
       const state = event.pass
         ? 'passed'
         : event.messages?.length
-        ? 'errored'
-        : 'failed';
+          ? 'errored'
+          : 'failed';
       let message: string | undefined;
       if (event.diff?.length) {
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/test-adapter.ts
```diff
@@ -220,14 +220,13 @@ export class QLTestAdapter extends DisposableObject implements TestAdapter {
...
-        ? 'errored'
-        : 'failed';
+          ? 'errored'
+          : 'failed';
       let message: string | undefined;
       if (event.diff?.length) {
         message = ['', `${state}: ${event.test}`, ...event.diff, ''].join('\n');
         testLogger.log(message);
       }
-      (event.diff || []).join('\n');
```
adityasharad, line 230: Did we mean to use this above instead of `...event.diff`?

adityasharad (approved):

extensions/ql-vscode/src/test-adapter.ts
aeisenberg: No...in the previous line, it is testing that `event.diff` is truthy.

adityasharad: And it's putting it into `message`

aeisenberg:
Hmmm...flaky test. I should track down why.

PR #611, opened 2020-10-06
Add telemetry for commands

aeisenberg:
This commit adds telemetry capturing for command execution. The data captured is only the command id.

We also capture errors thrown by any command execution.

This adds telemetry to an Azure application insights instance. Talk to me if you want access to it.

We still need a way to notify users that telemetry is being collected and a way to disable it.

Also:

* Extract commandRunner to its own file
* Add an option to log all telemetry commands to the console
* Add a popup (text to be determined) that requires users to accept
  telemetry before using.
* Convert telemetry to opt-in instead of opt-out
* Add a "canary" setting that users can use to opt-in to unreleased
  features.

## Checklist

- [x] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [x] Issues have been created for any UI or other user-facing changes made by this pull request.
- [x] `@github/docs-content-dsp` has been cc'd in all issues for UI or other user-facing changes made by this pull request.

PR #723, opened 2021-01-18
Refactor: Move commandRunner to its own module

aeisenberg:
Also, extract related functions and types. There are no behavioral changes in this commit. Only refactorings.

This aligns with the same refactoring in #611.


adityasharad (approved):
LGTM. I noticed we have a mix of kebab-case and camelCase in module filenames (usually the former). We can consider making that consistent, but I currently don't feel strongly about it.

aeisenberg:
I'm guilty of using camel case for module names (which I slightly prefer). It would be good to standardize.

PR #729, opened 2021-01-21
Forward scored metadata for canary users

henrymercer:
Add a `canary` config setting to enable users to opt into unreleased features of the extension.  When canary features are turned on, forward the `@scored` tag if it's present in the query metadata to enable experimental features for internal users.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -12,6 +12,7 @@ import { promisify } from 'util';
 import { CancellationToken, Disposable } from 'vscode';
 
 import { BQRSInfo, DecodedBqrsChunk } from './pure/bqrs-cli-types';
+import * as config from './config';
 import { CliConfig } from './config';
 import { DistributionProvider, FindDistributionResultKind } from './distribution';
 import { assertNever } from './pure/helpers-pure';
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -585,6 +586,9 @@ export class CodeQLCliServer implements Disposable {
       // grouping client-side.
       '--no-group-results',
     ];
+    if (config.isCanary() && metadata.scored !== undefined) {
+      args.push(`-t=scored=${metadata.scored}`);
```
aeisenberg, line 590: So, we are really just adding the `scored` flag to the call to interpret bqrs? Seems straightforward enough, though some questions:

* Is there a problem if we pass this flag to older versions of the CLI?
* I thought that the `@scored` metadata was just a boolean flag, though it seems to be typed as a string. What are the possible values for `@scored`?

henrymercer: > Is there a problem if we pass this flag to older versions of the CLI?

Older versions of the CLI won't be problematic since the CLI interprets query metadata as a property bag and unrecognised properties are ignored.

> I thought that the `@scored` metadata was just a boolean flag, though it seems to be typed as a string. What are the possible values for `@scored`?

The query metadata in `QueryInfo` comes from the output of `codeql resolve metadata` where each metadata key-value pair has type `Entry<String, String>`.  So at this stage of the pipeline `metadata.scored` will be the everything right of the `@scored ` line of the query metadata.

aeisenberg: So, metadata like this:

```
/**
 * @scored
 */
```

Will have an empty string as a value? I guess in the Java code, you will just distinguish between having the flag and not having the flag, where the actual value will be ignored. Are you passing the full string just in case in the future we do want to distinguish different values?

henrymercer: That's correct.  While the string has no semantic meaning in terms of interpretation, there is a small benefit to passing the full string.  By passing the full string, we can produce a warning if the user writes something like `@scored no` saying that arguments to the `@scored` tag are ignored.

aeisenberg: Thanks for the clarification!

aeisenberg (approved):

PR #734, opened 2021-01-29
Ensure databases are re-registered when query server restarts

### LINKED ISSUE #733: "No result from server" after changing codeql path
If after running a query I change the path to codeql and then run another query I get `No result from server` printed into the extension log and as a message. The queryserver log shows  `- - - COMPILATION DONE - - - ` which shows that compilation seems to have happened.

Restarting vscode seems to fix this but restarting the query server doesn't seem to fix it.

aeisenberg:
This commit fixes #733. It does it by ensuring that the query server
emits an event when it restarts the query server. The database manager
listens for this even and properly re-registers its databases.

A few caveats though:

1. Convert query restarts to using a command that includes progress.
   This will ensure that errors on restart are logged properly.
2. Because we want to log errors, we cannot use the vscode standard
   EventEmitters. They run in the next tick and therefore any errors
   will not be associated with this command execution.
3. Update the default cli version to run integration tests against to
   2.4.2.
4. Add a new integration test that fails if databases are not
   re-registered.

## Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -121,7 +121,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        version: ['v2.2.6', 'v2.3.3', 'v2.4.0']
+        version: ['v2.2.6', 'v2.3.3', 'v2.4.2']
     env:
       CLI_VERSION: ${{ matrix.version }}
       TEST_CODEQL_PATH: '${{ github.workspace }}/codeql'
--- a/extensions/ql-vscode/src/databases.ts
+++ b/extensions/ql-vscode/src/databases.ts
@@ -514,7 +514,10 @@ export class DatabaseManager extends DisposableObject {
   ) {
     super();
 
-    this.loadPersistedState();  // Let this run async.
+    qs.onDidStartQueryServer(this.reregisterDatabases.bind(this));
+
+    // Let this run async.
+    this.loadPersistedState();
   }
 
   public async openDatabase(
@@ -540,6 +543,22 @@ export class DatabaseManager extends DisposableObject {
     await this.addDatabaseSourceArchiveFolder(databaseItem);
 
     return databaseItem;
+  }
+
+  private async reregisterDatabases(
+    progress: ProgressCallback,
+    token: vscode.CancellationToken
+  ) {
+    let completed = 0;
+    await Promise.all(this._databaseItems.map(async (databaseItem) => {
+      await this.registerDatabase(progress, token, databaseItem);
+      completed++;
+      progress({
+        maxStep: this._databaseItems.length,
+        step: completed,
+        message: 'Re-registering databases'
+      });
+    }));
   }
 
   private async addDatabaseSourceArchiveFolder(item: DatabaseItem) {
--- a/extensions/ql-vscode/src/queryserver-client.ts
+++ b/extensions/ql-vscode/src/queryserver-client.ts
@@ -1,14 +1,15 @@
 import * as cp from 'child_process';
 import * as path from 'path';
 import { DisposableObject } from './vscode-utils/disposable-object';
-import { Disposable } from 'vscode';
-import { CancellationToken, createMessageConnection, MessageConnection, RequestType } from 'vscode-jsonrpc';
+import { Disposable, CancellationToken, commands } from 'vscode';
+import { createMessageConnection, MessageConnection, RequestType } from 'vscode-jsonrpc';
 import * as cli from './cli';
 import { QueryServerConfig } from './config';
 import { Logger, ProgressReporter } from './logging';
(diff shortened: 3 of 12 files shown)


extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -60,6 +61,16 @@ export class QueryServerClient extends DisposableObject {
   nextCallback: number;
   nextProgress: number;
   withProgressReporting: WithProgressReporting;
+
+  private readonly _onDidStartQueryServer = [] as ProgressTask<void>[];
```
adityasharad, line 65: Maybe call this something like `queryServerStartListeners`

extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
```diff
@@ -100,4 +100,27 @@ describe('Queries', function() {
...
+        qs,
+        dbItem,
+        false,
+        Uri.file(queryPath),
+        progress,
+        token
+      );
+
+      // this message would indicate that the databases were not properly reregistered
+      expect(result.result.message).not.to.eq('No result from server');
```
adityasharad, line 119: Can you also check that the `result` status is successful? Otherwise we could change the message and break the behaviour without failing this test.

extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
```diff
@@ -100,4 +100,27 @@ describe('Queries', function() {
...
+      expect(result.options.queryText).to.eq(fs.readFileSync(queryPath, 'utf8'));
+    } catch (e) {
+      console.error('Test Failed');
+      fail(e);
+    }
```
adityasharad, line 124: Was this for debugging or do you need to keep it?

extensions/ql-vscode/src/databases.ts
```diff
@@ -542,6 +545,24 @@ export class DatabaseManager extends DisposableObject {
     return databaseItem;
   }
 
+  private onQueryServerRestarted = async (
```
adityasharad, line 548: Suggest calling this `reregisterDatabases` and adding an explanatory comment. (Once it gets passed to `qs`, that code knows it will be run on restart.)

extensions/ql-vscode/src/databases.ts
```diff
@@ -542,6 +545,24 @@ export class DatabaseManager extends DisposableObject {
...
+      await this.registerDatabase(progress, token, databaseItem);
+      completed++;
+      progress({
+        maxStep: this._databaseItems.length,
+        step: completed,
+        message: 'Re-registering databases'
+      });
+    }));
+
+    throw new Error('yikes');
```
adityasharad, line 563: Remove?

.github/workflows/main.yml
```diff
@@ -121,7 +121,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        version: ['v2.2.6', 'v2.3.3', 'v2.4.0']
+        version: ['v2.2.6', 'v2.3.3', 'v2.4.2']
```
adityasharad, line 124: I wonder if we could autoupdate this (either within the workflow or by a scheduled PR) to the latest patch versions of all released CLIs. Might try it later.

adityasharad (commented):
Appreciate the quick fix and the test! A few minor suggestions.

extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
aeisenberg: We should keep this. The tests aren't failed if an error is thrown. We need to explicitly throw it here.

extensions/ql-vscode/src/queryserver-client.ts
aeisenberg: I was being consistent with our other event fields. But, I'm not fond of the naming convention.

extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
aeisenberg: Sure...I can do something here.

extensions/ql-vscode/src/databases.ts
aeisenberg: Yikes! Forgot to remove that.

.github/workflows/main.yml
aeisenberg: I thought that, too. But seems too much for this PR.

adityasharad (approved):

aeisenberg:
The integration tests were failing because of a real bug. Fixed now.

adityasharad (approved):

adityasharad:
Looks good. Do you want a change note for the upgrade bugfix?

aeisenberg:
> Looks good. Do you want a change note for the upgrade bugfix?

Sure...why not...

PR #747, opened 2021-02-08
Avoid clobbering quick-query file when re-opened

aeisenberg:
Only recreate the qlpack.yml file.

Also, add an integration test for quick-query creation.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/quick-query.ts
+++ b/extensions/ql-vscode/src/quick-query.ts
@@ -1,16 +1,20 @@
 import * as fs from 'fs-extra';
 import * as yaml from 'js-yaml';
 import * as path from 'path';
-import { CancellationToken, ExtensionContext, window as Window, workspace, Uri } from 'vscode';
+import {
+  CancellationToken,
+  ExtensionContext,
+  window as Window,
+  workspace,
+  Uri
+} from 'vscode';
 import { ErrorCodes, ResponseError } from 'vscode-languageclient';
 import { CodeQLCliServer } from './cli';
 import { DatabaseUI } from './databases-ui';
-import { logger } from './logging';
 import {
   getInitialQueryContents,
   getPrimaryDbscheme,
   getQlPackForDbscheme,
-  showAndLogErrorMessage,
   showBinaryChoiceDialog,
 } from './helpers';
 import {
@@ -21,23 +25,35 @@ import {
 const QUICK_QUERIES_DIR_NAME = 'quick-queries';
 const QUICK_QUERY_QUERY_NAME = 'quick-query.ql';
 const QUICK_QUERY_WORKSPACE_FOLDER_NAME = 'Quick Queries';
+const QLPACK_FILE_HEADER = '# This is an automatically generated file.\n\n';
 
 export function isQuickQueryPath(queryPath: string): boolean {
   return path.basename(queryPath) === QUICK_QUERY_QUERY_NAME;
 }
 
-function getQuickQueriesDir(ctx: ExtensionContext): string {
+async function getQuickQueriesDir(ctx: ExtensionContext): Promise<string> {
   const storagePath = ctx.storagePath;
   if (storagePath === undefined) {
     throw new Error('Workspace storage path is undefined');
   }
   const queriesPath = path.join(storagePath, QUICK_QUERIES_DIR_NAME);
-  fs.ensureDir(queriesPath, { mode: 0o700 });
+  await fs.ensureDir(queriesPath, { mode: 0o700 });
   return queriesPath;
 }
 
+function updateQuickQueryDir(queriesDir: string, index: number, len: number) {
+  workspace.updateWorkspaceFolders(
+    index,
+    len,
+    { uri: Uri.file(queriesDir), name: QUICK_QUERY_WORKSPACE_FOLDER_NAME }
+  );
+}
 
-
+function findExistingQuickQueryEditor() {
+  return Window.visibleTextEditors.find(editor =>
+    path.basename(editor.document.uri.fsPath) === QUICK_QUERY_QUERY_NAME
+  );
+}
 
 /**
  * Show a buffer the user can enter a simple query into.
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
```diff
@@ -21,7 +22,7 @@ import { QueryResultType } from '../../pure/messages';
  * Integration tests for queries
  */
 describe('Queries', function() {
-  this.timeout(20000);
+  this.timeout(20000000);
```
adityasharad, line 25: Do we need this long?

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -109,24 +106,22 @@ export async function displayQuickQuery(
       libraryPathDependencies: [qlpack]
     };
 
+    // Only recreate the query file if it doesn't already exist.
+    // Always recreate the qlpack file because the file will change when database languages change
```
adityasharad, line 110: If the database language changes, then the existing query file may not be useful (especially if it contains `import <language>`). Can we detect whether the language has changed, and replace both query and qlpack files only in that case?

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -109,24 +106,22 @@ export async function displayQuickQuery(
...
-      logger.log(e.message);
-    }
-    else if (e instanceof Error)
-      showAndLogErrorMessage(e.message);
-    else
+  } catch (e) {
+    if (e instanceof ResponseError && e.code == ErrorCodes.RequestCancelled) {
+      throw new UserCancellationException(e.message);
+    } else {
       throw e;
```
adityasharad, line 124: This will go to the top-level handler that logs it before displaying, right?

aeisenberg: That's right. This is aligning the command with all the other commands.

I should probably also take `if (e instanceof ResponseError && e.code == ErrorCodes.RequestCancelled)` and put it into the global handler.

extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
aeisenberg: Forgot to remove that.

extensions/ql-vscode/src/quick-query.ts
aeisenberg: I was hoping to avoid the complexity, but maybe because of the import statement, we do want to overwrite the old file.

adityasharad: Simpler: always overwrite both query and qlpack files whenever you switch database. No need to check the language.

This is suboptimal if you're writing a quick query for the same language and want to run it on several different DBs. But it's an improvement on what we currently have. If the user needs the old query they can still get it from query history.

aeisenberg: The complexity is more that the user may change databases and keep the quick-query file open. When that happens, the contents will be wrong. The quick-query contents are only recreated when the command is called.

I guess this incorrect behaviour is currently in the extension. In order to fix this, I need to add a listener for database changes. We already have these listeners, so it's not too bad.

adityasharad: I think that's ok? If they have the query open we ignore it. If they ask for a fresh Quick Query we give them a fresh one if it's a new DB, or the existing one if the DB hasn't changed. Seems like a strict win without having to add more listeners, though if you think that's worthwhile I don't mind.

nickrolfe:
I downloaded the `.vsix` and had a quick play, and it's definitely a nice QoL improvement for me. Thanks!

extensions/ql-vscode/src/quick-query.ts
aeisenberg: No, that's getting messy. I'll change things so that the quick query is only updated when the database changes (if necessary).

aeisenberg:
Discovered a bug in quick-queries (exists in current implementation as well). I'll fix this bug in this PR.

The bug is that sometimes even after you close the quick query window, the query text document stays in memory. Our check for whether or not to recreate the quick query fails (ie- the extension _thinks_ the query window has remained open, even though it is closed, so the quick-queries are not recreated).

I'll make the fix in this PR, but in a different commit.

aeisenberg:
Discovered a bug in quick-queries (exists in current implementation as well). I'll fix this bug in this PR.

The bug is that sometimes even after you close the quick query window, the query text document stays in memory. Our check for whether or not to recreate the quick query fails (ie- the extension _thinks_ the query window has remained open, even though it is closed, so the quick-queries are not recreated).

I'll make the fix in this PR, but in a different commit.

aeisenberg:
Discovered a bug in quick-queries (exists in current implementation as well). I'll fix this bug in this PR.

The bug is that sometimes even after you close the quick query window, the query text document stays in memory. Our check for whether or not to recreate the quick query fails (ie- the extension _thinks_ the query window has remained open, even though it is closed, so the quick-queries are not recreated).

I'll make the fix in this PR, but in a different commit.

aeisenberg:
Hmmm...I need to figure out this flaky test failure. It is passing locally.

aeisenberg:
OK...this is the same problem that we originally had in the telemetry tests. Should be fixed by adding a short wait after updating the settings. In the future, I should refactor all `wait()` functions into a single, shared instance. That's for later, though.

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -102,31 +111,38 @@ export async function displayQuickQuery(
...
+    }
+  }
+}
+
+async function checkShouldRewrite(qlPackFile: string, newDepedency: string) {
```
adityasharad, line 142:
```suggestion
async function checkShouldRewrite(qlPackFile: string, newDependency: string) {
```

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -102,31 +111,38 @@ export async function displayQuickQuery(
...
+  if (!(await fs.pathExists(qlPackFile))) {
+    return true;
   }
+  const qlPackContents: any = yaml.safeLoad(await fs.readFile(qlPackFile, 'utf8'));
+  return qlPackContents?.libraryPathDependencies?.[0] !== newDepedency;
```
adityasharad, line 147:
```suggestion
  return qlPackContents?.libraryPathDependencies?.[0] !== newDependency;
```

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -102,31 +111,38 @@ export async function displayQuickQuery(
...
       throw e;
+    }
+  }
+}
+
+async function checkShouldRewrite(qlPackFile: string, newDepedency: string) {
+  if (!(await fs.pathExists(qlPackFile))) {
+    return true;
   }
+  const qlPackContents: any = yaml.safeLoad(await fs.readFile(qlPackFile, 'utf8'));
```
adityasharad, line 146: Worth having a simple type here instead of `any`?

adityasharad (approved):

extensions/ql-vscode/src/quick-query.ts
aeisenberg: I don't think it adds much value here. If we are going to start using qlpack files elsewhere in the code, then yes.

PR #752, opened 2021-02-11
Update typings for mocha

aeisenberg:
Internal change. Also, updates the lock file format for compatibility with npm v7.

Also, see the second commit:

Fix globalSetup/teardown for mocha

Updating the typings for mocha uncovered an error in how we were
registering global setups and teardowns.

When calling `mocha.globalSetup` or `mocha.globalTeardown`, any
previously registered globals are overwritten. The workaround
is to attach globals directly to the internal options object.

This is a requirement because we are now registering globals in 
multiple files.

Unfortunately, the typings for mocha do not permit this and I may need
to fix them again.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package-lock.json
+++ b/extensions/ql-vscode/package-lock.json
@@ -1,8 +1,12255 @@
 {
   "name": "vscode-codeql",
   "version": "1.4.3",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "vscode-codeql",
+      "version": "1.4.3",
+      "license": "MIT",
+      "dependencies": {
+        "child-process-promise": "^2.2.1",
+        "classnames": "~2.2.6",
+        "fs-extra": "^9.0.1",
+        "glob-promise": "^3.4.0",
+        "js-yaml": "^3.14.0",
+        "minimist": "~1.2.5",
+        "node-fetch": "~2.6.0",
+        "react": "^16.8.6",
+        "react-dom": "^16.8.6",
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/package-lock.json
```diff
@@ -1,8 +1,12255 @@
 {
   "name": "vscode-codeql",
   "version": "1.4.3",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
adityasharad, line 4: Does this mean we have to use Node 14 / NPM 7? If so worth updating the contributor instructions.

adityasharad (approved):

extensions/ql-vscode/package-lock.json
aeisenberg: this lock file format is backwards compatible with npm v5 and v6. And theoretically compatible with v12. Looking at the readme now, it shows some outdated dependencies. It will _probably_ work with them, but we're not testing on them any more, so I'll update it.

aeisenberg:
Concerning that the integration tests are failing here. I'll have to explore why.

aeisenberg:
Another review, please. The failing tests uncovered a problem with the way we are registering globals. It hasn't affected anything yet, but it's likely to do so in the future.

adityasharad (approved):

aeisenberg:
Thanks.  But, something's still weird and tests are stil failing. I'll have to inspect more.

adityasharad:
Do the downloaded CLI and the test DB need to be registered as globals?

aeisenberg:
It was just a bad merge. I missed a few bits. Should be working now.

PR #753, opened 2021-02-11
Fix error messages for ast viewers and update caching

aeisenberg:
This commit does two things:

1. Add more appropriate error messages when asts can't be viewed.
2. Make better use of cached operations for asts. In the past, we were
not actually using cached operations. Each time an ast view request
occurred, we created a new TemplatePrintAstProvider instance. With this
change, we reuse the TemplatePrintAstProvider between calls and ensure
that an AST that is called once is reused on subsequent calls.

I received complaints from users such as @sj that AST viewing was erroring out and it wasn't clear why it was happening.

And the second part of this change will make subsequent calls to view an AST of a file extremely fast.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/contextual/templateProvider.ts
+++ b/extensions/ql-vscode/src/contextual/templateProvider.ts
@@ -1,4 +1,15 @@
-import * as vscode from 'vscode';
+import {
+  CancellationToken,
+  DefinitionProvider,
+  Location,
+  LocationLink,
+  Position,
+  ProgressLocation,
+  ReferenceContext,
+  ReferenceProvider,
+  TextDocument,
+  Uri
+} from 'vscode';
 
 import { decodeSourceArchiveUri, encodeArchiveBasePath, zipArchiveScheme } from '../archive-filesystem-provider';
 import { CodeQLCliServer } from '../cli';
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/contextual/templateProvider.ts
```diff
@@ -112,40 +123,41 @@ export class TemplateQueryReferenceProvider implements vscode.ReferenceProvider
...
-      return;
-    }
-    const queryResults = await this.cache.get(document.uri.toString());
-    if (!queryResults) {
-      return;
+      throw new Error('Cannot view the AST. Please select a valid source file inside a CodeQL database.');
     }
+    const queryResults = await this.cache.get(document.uri.toString(), progress, token);
 
     return new AstBuilder(
```
adityasharad, line 146: Looks like we now unconditionally return an `AstBuilder`. Do you still need to return `undefined` if we didn't find any results?

aeisenberg: The cached operation always returns `QueryWithResults` or it throws. Since the call to get the cached item always has the same type as the operation itself, we are guaranteed that the value will never be `undefined`.

adityasharad (approved):

PR #765, opened 2021-02-25
Use downgrades when fixing dbscheme mismatches where possible.

alexet:
When resolving "upgrades" also resolve downgrades when doing things non-destructively. This requires codeql `2.4.4`. This won't do anything until we start shipping downgrades.

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -131,6 +131,12 @@ export class CodeQLCliServer implements Disposable {
    */
   private static CLI_VERSION_WITH_LANGUAGE = new SemVer('2.4.1');
 
+  /**
+   * CLI version where `codeql resolve upgrades` supports
+   * the `--allow-downgrades` flag
+   */
+  private static CLI_VERSION_WITH_DOWNGRADES = new SemVer('2.4.4');
+
   /** The process for the cli server, or undefined if one doesn't exist yet */
   process?: child_process.ChildProcessWithoutNullStreams;
   /** Queue of future commands*/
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -131,6 +131,12 @@ export class CodeQLCliServer implements Disposable {
    */
   private static CLI_VERSION_WITH_LANGUAGE = new SemVer('2.4.1');
 
+  /**
+   * CLI version where `codeql resolve upgrades` supports
+   * the `--allow-downgrades` flag
+   */
+  private static CLI_VERSION_WITH_DOWNGRADES = new SemVer('2.4.4');
```
aeisenberg, line 138: Not necessary for this PR, but we should think of a better way of dealing with feature flags since I can only imagine this is going to get worse.

extensions/ql-vscode/src/cli.ts
```diff
@@ -655,12 +661,15 @@ export class CodeQLCliServer implements Disposable {
    * @param targetDbScheme The dbscheme to try to upgrade to.
    * @returns A list of database upgrade script directories
    */
-  resolveUpgrades(dbScheme: string, searchPath: string[], targetDbScheme?: string): Promise<UpgradesInfo> {
+  async resolveUpgrades(dbScheme: string, searchPath: string[], allowDowngradesIfPossible: boolean, targetDbScheme?: string): Promise<UpgradesInfo> {
```
aeisenberg, line 664: Please update the jsdoc above.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -348,7 +348,7 @@ async function compileNonDestructiveUpgrade(
   if (!query.dbItem?.contents?.dbSchemeUri) {
     throw new Error('Database is invalid, and cannot be upgraded.');
   }
-  const { scripts, matchesTarget } = await qs.cliServer.resolveUpgrades(query.dbItem.contents.dbSchemeUri.fsPath, searchPath, query.queryDbscheme);
+  const { scripts, matchesTarget } = await qs.cliServer.resolveUpgrades(query.dbItem.contents.dbSchemeUri.fsPath, searchPath, true, query.queryDbscheme);
```
aeisenberg, line 351: It's unclear to me why this is the only place where we allow a downgrade. Can you clarify?

alexet: We only want to downgrade with non-destructive cases. The other uses are either in the manual upgrade which is destructive or for suggesting destructive upgrades when running with older cli versions.

aeisenberg (approved):
Looks fine by me. I have not run this code.

PR #770, opened 2021-03-02
Add error message when interpretation fails

### LINKED ISSUE #759: Improve discoverability of errors that break the alerts view
**Describe the bug**
In [a recent discussion](https://github.com/github/securitylab/discussions/274#discussioncomment-388467), @adriaanjacobs asked about a situation where the alerts view was silently unusable for a particular path query against a particular database. Ultimately, there was an error in the extension output pane, which indicated that we had [encountered an exception](https://github.com/github/vscode-codeql/blob/ab31d86a8d7e09926772e0b84744e8a45639be47/extensions/ql-vscode/src/interface.ts#L630) while processing interpreted results, and were falling back on raw display.

While that's certainly better than not displaying results at all, it would be nice to find a way of making it clear why the `alerts` view is non-functional (perhaps by replacing it with an error pane that shows details about the exception we caught?).

**To Reproduce**
I have a database and query that reproduce the problem -- ping me on Slack if it would be useful to have them.

aeisenberg:
One way it can fail is if the SARIF is too large. We explicitly call
out that error because the raw message received from the node runtime
is not very understandable.

Fixes #759

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Allow using raw LGTM project slugs for fetching LGTM databases. [#769](https://github.com/github/vscode-codeql/pull/769)
+- Better error messages when BQRS interpretation fails to produce SARIF. [#770](https://github.com/github/vscode-codeql/pull/770)
 
 ## 1.4.3 - 22 February 2021
 
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -603,8 +603,12 @@ export class CodeQLCliServer implements Disposable {
     let output: string;
     try {
       output = await fs.readFile(interpretedResultsPath, 'utf8');
-    } catch (err) {
-      throw new Error(`Reading output of interpretation failed: ${err.stderr || err}`);
+    } catch (e) {
+      const rawMessage = e.stderr || e.message;
+      const errorMessage = rawMessage.startsWith('Cannot create a string')
+        ? `SARIF too large. ${rawMessage}`
+        : rawMessage;
+      throw new Error(`Reading output of interpretation failed: ${errorMessage}`);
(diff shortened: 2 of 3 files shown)


extensions/ql-vscode/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Allow using raw LGTM project slugs for fetching LGTM databases. [#769](https://github.com/github/vscode-codeql/pull/769)
+- Better error messages when BQRS interpretation fails to produce SARIF. [#770](https://github.com/github/vscode-codeql/pull/770)
```
adityasharad, line 6:
```suggestion
- Better error messages when result interpretation fails to produce SARIF for the `alerts` view . [#770](https://github.com/github/vscode-codeql/pull/770)
```

extensions/ql-vscode/src/cli.ts
```diff
@@ -603,8 +603,12 @@ export class CodeQLCliServer implements Disposable {
     let output: string;
     try {
       output = await fs.readFile(interpretedResultsPath, 'utf8');
-    } catch (err) {
-      throw new Error(`Reading output of interpretation failed: ${err.stderr || err}`);
+    } catch (e) {
+      const rawMessage = e.stderr || e.message;
+      const errorMessage = rawMessage.startsWith('Cannot create a string')
+        ? `SARIF too large. ${rawMessage}`
```
adityasharad, line 609: Could we log the raw message but make the user-facing one a bit friendlier? Something like "SARIF results are too large to process for the alerts view. Will show raw results instead."

adityasharad (approved):
Generally looks good. Minor suggestion on whether we can make the user-facing message shorter and more understandable, so they understand the exact impact.

extensions/ql-vscode/src/cli.ts
aeisenberg: Hmmm...I think I can do that.

Issue #767, opened 2021-02-28
Support Sorting in Query History View


Ovi3:
**Problem description**
After run a lot of query, there're a lot of item in Query History View. And  I wanna choose those item that has at least one result. 
Is there a way to filter those items? If not, this issue is a feature request.

**Describe the solution I'd like**
Filter function in Query History View

aeisenberg:
There's no exact way to do this at the moment, but there are two features you can use that might help.

1. You can remove any history item (select and click the trash can icon).
2. You can rename any history item (right click and select "set label").

Rather than filtering, I think a more useful option would be to support sorting by name, date added, and by result count.

Ovi3:
Yeah, sorting is a better way. Thanks for the help.

aeisenberg:
This is a reasonable feature request. I don't know when we will get to it, but I'd like to keep it open so that we can in the future.

aeisenberg:
@edoardopirovano 👋🏼 Welcome!

First take a look at the way we are sorting databases. We have two options (date and name) there and you can use the same icons, etc. And as to answering your questions:

1. I think that's entirely reasonable. Go for it. 
2. This is trickier as you mention. The easiest thing to do would most likely be to fill in the result count when you first open the query results page (all queries are opened immediately after being completed). You might be able to do this here: https://github.com/aeisenberg/vscode-codeql/blob/be9084e83e39d87a82d1760085ee34de88571390/extensions/ql-vscode/src/extension.ts#L448-L453 Though, that will need some more exploration.

So, maybe it's best to focus on 1) first and get to 2) later.

PR #777, opened 2021-03-11
Implement sorting of query history by name, date, and result count

edoardopirovano:
This pull request addresses #767 by introducing three ways of sorting the query history view - by a query's label, by the date the query was run, and by the number of results the query produced.


extensions/ql-vscode/src/interface.ts
```diff
@@ -400,6 +400,7 @@ export class InterfaceManager extends DisposableObject {
       }
     );
     const resultSet = transformBqrsResultSet(schema, chunk);
+    results.setResultCount(resultSet.schema.rows);
```
aeisenberg, line 403: This is a little confusing. The `alerts` table has 33 results in this example, but the `#select` table has 3906 results. I think people would expect that the result count matches the number of rows in the default table (in this case, it's `alerts`).

I'm not entirely sure why there is a difference here. I'm just running the example query from the starter repo:

```ql
/**
 * @name Empty block
 * @kind problem
 * @problem.severity warning
 * @id javascript/example/empty-block
 */

import javascript

from BlockStmt b
where b.getNumStmt() = 0
select b, "This is an empty block."
```

It would be nice if we could get the counts for the default table, but I'm not sure how easy this is without loading the sarif.

_Extension_Development_Host__-_CodeQL_Query_Results_—_vscode-codeql-starter__Workspace_

aeisenberg (approved):
Looks good. And is working for me.

aeisenberg:
Can you squash and rebase your commits?

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Respect the `codeQL.runningQueries.numberOfThreads` setting when creating SARIF files during result interpretation. [#771](https://github.com/github/vscode-codeql/pull/771)
 - Allow using raw LGTM project slugs for fetching LGTM databases. [#769](https://github.com/github/vscode-codeql/pull/769)
 - Better error messages when BQRS interpretation fails to produce SARIF. [#770](https://github.com/github/vscode-codeql/pull/770)
+- Implement sorting of the query history view by name, date, and results count. [#767](https://github.com/github/vscode-codeql/pull/767)
```
aeisenberg, line 9: We usually link to the PRs, not the issues.

```suggestion
- Implement sorting of the query history view by name, date, and results count. [#777](https://github.com/github/vscode-codeql/pull/777)
```

extensions/ql-vscode/package.json
```diff
@@ -179,8 +179,8 @@
         },
         "codeQL.queryHistory.format": {
           "type": "string",
-          "default": "[%t] %q on %d - %s",
-          "description": "Default string for how to label query history items. %t is the time of the query, %q is the query name, %d is the database name, and %s is a status string."
+          "default": "%q on %d - %s, %r results [%t]",
```
aeisenberg, line 182: Avoid pluralization issues:

```suggestion
          "default": "%q on %d - %s, %r result count [%t]",
```

aeisenberg (changes requested):
Sorry....a few more changes. Just noticed them. If a query has an error, the result count is -1. That should not be displayed in the UI. 

_Extension_Development_Host__-_example2_ql_—_vscode-codeql-starter__Workspace_

Looks like `-1` is being set in the code. Can you change this to `0`?

aeisenberg (approved):
Thanks for the quick turnaround.

Issue #593, opened 2020-09-28
Add option to include the --save-cache option to the query server to more aggressively save the cache


Sn0rlax-lgtm:
Running a query, closing, reopening, and rerunning a query on VS Code yields no performance benefits. Is VS Code (or Eclipse/Visual Studio) configurable to use an on disk cache, similar to the CodeQL CLI?

p0:
The VS Code extension actually uses the CodeQL CLI under the hood, so it should be using a disk cache in exactly the same way. I expect at least every `cached` predicate to end up in the disk cache (which you can see under `db-*/default/cache` in a CodeQL database folder), meaning that after a VS Code restart re-running a query should only execute the final stage (and not any of the preceding stages), assuming the QL code is the same.

Are you observing a situation where this does not happen? Would you be able to provide reproduction instructions?

Snorlax-lgtm:
As far as I can tell, any large codebase and any query runs long the first time, 0s the second time. Upon closing and reopening VS Code we go back to the long timings of the first query.

p0:
This is expected -- if you keep running queries in a single VS Code session, the CodeQL engine stays up, and its memory cache is able to service repeat queries (or indeed slightly edited queries) very quickly.

If you shut down the QL engine (by restarting VS Code, or switching database), that memory cache is lost, and only the disk cache is available. We deliberately avoid caching the final query stage eagerly -- one of the main use cases for the VS Code extension is interactively refining analyses, where you run the same query with slight modifications to the final stage repeatedly. If we wrote each of those intermediate results to the disk cache, we'd use an unreasonable amount of disk space. The disk cache aims to contain predicates that are likely to be useful for future query evaluations, which in practice tends to mean `cached` predicates (aka non-final query stages).

Aside: The engine actually has a mode where it [uses the disk cache much more eagerly](https://help.semmle.com/codeql/codeql-cli/commands/query-run.html#cmdoption-codeql-query-run-no-save-cache). We could consider exposing that mode (ideally with a user-configurable disk cache threshold) in the VS Code extension, in order to make restarts less disruptive, though some experiments on whether it would actually save time overall or in fact spend longer writing single-use data to disk would be needed.

Snorlax-lgtm:
> This is expected -- if you keep running queries in a single VS Code session, the CodeQL engine stays up, and its memory cache is able to service repeat queries (or indeed slightly edited queries) very quickly.
> 
[...]

PR #778, opened 2021-03-11
Expand disk cache evaluator options

edoardopirovano:
This PR addresses #593 by expanding the options that can be passed to control how the evaluator handles the disk cache. In particular, it exposes to the user two settings - the first enables the saving of intermediate results to the cache, and the second allows limiting of the size of the cache. The default behaviour is unchanged from the current options - i.e. intermediate results are not saved to the disk cache, and the evaluator is allowed to choose its own limit on maximum cache size.

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [UNRELEASED]
 
+- Introduce evaluator options for saving intermediate results to the disk cache, and for limiting the size of this cache. [#593](https://github.com/github/vscode-codeql/pull/593)
 - Respect the `codeQL.runningQueries.numberOfThreads` setting when creating SARIF files during result interpretation. [#771](https://github.com/github/vscode-codeql/pull/771)
 - Allow using raw LGTM project slugs for fetching LGTM databases. [#769](https://github.com/github/vscode-codeql/pull/769)
 - Better error messages when BQRS interpretation fails to produce SARIF. [#770](https://github.com/github/vscode-codeql/pull/770)
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-- Introduce evaluator options for saving intermediate results to the disk cache, and for limiting the size of this cache. [#593](https://github.com/github/vscode-codeql/pull/593)
+- Introduce evaluator options for saving intermediate results to the disk cache (`codeQL.runningQueries.saveCache`) and for limiting the size of this cache (`codeQL.runningQueries.cacheSize`). [#593](https://github.com/github/vscode-codeql/pull/593)
 - Respect the `codeQL.runningQueries.numberOfThreads` setting when creating SARIF files during result interpretation. [#771](https://github.com/github/vscode-codeql/pull/771)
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/package.json
```diff
@@ -123,6 +123,20 @@
           "maximum": 1024,
           "description": "Number of threads for running queries."
         },
+        "codeQL.runningQueries.saveCache": {
+          "type": "boolean",
+          "default": false,
+          "description": "Aggressively save intermediate results to the disk cache. This advanced option is not recommended as it will greatly increase disk usage and evaluation time, but it may speed up subsequent queries if they are similar."
```
aeisenberg, line 129: I'd avoid the phrase "not recommended".

```suggestion
          "description": "Aggressively save intermediate results to the disk cache. This may speed up subsequent queries if they are similar. Be aware that using this option will greatly increase disk usage and initial evaluation time."
```

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [UNRELEASED]
 
+- Introduce evaluator options for saving intermediate results to the disk cache, and for limiting the size of this cache. [#593](https://github.com/github/vscode-codeql/pull/593)
```
aeisenberg, line 5:
```suggestion
- Introduce evaluator options for saving intermediate results to the disk cache (`codeQL.runningQueries.saveCache`) and for limiting the size of this cache (`codeQL.runningQueries.cacheSize`). [#593](https://github.com/github/vscode-codeql/pull/593)
```

extensions/ql-vscode/package.json
```diff
@@ -123,6 +123,20 @@
...
+        },
+        "codeQL.runningQueries.cacheSize": {
+          "type": [
+            "integer",
+            "null"
+          ],
+          "default": null,
+          "minimum": 1024,
+          "description": "Maximum size of the disk cache (in MB). Leave blank to allow the evaluator to automatically adjust the size of the disk cache based on the size of the codebase and the complexity of the queries being executed."
```
aeisenberg, line 138:
```suggestion
          "description": "Maximum size of the disk cache (in MB). Leave blank or set to 0 to allow the evaluator to automatically adjust the size of the disk cache based on the size of the codebase and the complexity of the queries being executed."
```

extensions/ql-vscode/package.json
```diff
@@ -123,6 +123,20 @@
           "maximum": 1024,
           "description": "Number of threads for running queries."
         },
+        "codeQL.runningQueries.saveCache": {
```
aeisenberg, line 126: Probably add a scope of `"scope": "window"` to ensure that downloaded workspaces don't accidentally blow up someone's hard drive.

extensions/ql-vscode/package.json
edoardopirovano: The value can't actually be set to `0` because of the `"minimum": 1024`

aeisenberg (approved):

aeisenberg:
@github/docs-content-dsp Do we need to update docs for this?

shati-patel:
> @github/docs-content-dsp Do we need to update docs for this?

👋🏽 Thanks for the ping! This one doesn't need a docs update 😃 (We have a general section about "[Configuring settings for running queries](https://codeql.github.com/docs/codeql-for-visual-studio-code/customizing-settings/#configuring-settings-for-running-queries)", and we can rely on the descriptions in the extension itself for more detail.)

PR #784, opened 2021-03-16
Implement viewing of query results as a CSV

edoardopirovano:
This pull request addresses the second part of #342 by introducing an option when right clicking a completed query in the query history view to see a CSV of the results. If the CSV results for the query have been viewed before, the file is simply opened. If they have not been viewed before, the CSV is generated (in the temporary directory for the query's execution) using a call to the CLI, and is then opened.

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -26,6 +26,11 @@ import { CompilationMessage } from './pure/messages';
 const SARIF_FORMAT = 'sarifv2.1.0';
 
 /**
+ * The string used to specify CSV format.
+ */
+const CSV_FORMAT = 'csv';
+
+/**
  * Flags to pass to all cli commands.
  */
 const LOGGING_FLAGS = ['-v', '--log-to-stderr'];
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -629,6 +640,9 @@ export class CodeQLCliServer implements Disposable {
     }
   }
 
+  async generateResultsCsv(metadata: { kind: string; id: string; scored?: string }, resultsPath: string, csvPath: string, sourceInfo?: SourceInfo): Promise<void> {
```
aeisenberg, line 643: Hmmm...the `scored` parameter is unused here. I'm not sure if it is even used when generating csv. @henrymercer is it worth passing in the `scored` value as we're doing for sarif interpretation?

henrymercer: We're still producing interpreted alerts here AIUI, so yes I think we'd need to forward the `scored` query metadata here so that queries with alert scores are interpreted correctly.

edoardopirovano: I think this is being forwarded correctly? It's one of the keys in the `metadata` parameter, which is passed on and used in line 605. Do you have an example query with the scored metadata I could use to test this? As far as I can tell none of the standard library queries use it unless I'm missing something...

aeisenberg: Ah...yes. My mistake. It _is_ being used correctly. The scored parameter is something that @henrymercer has been working on to implement ranked queries that order results by confidence. It's still experimental, and it is only enabled when you enter `canary` mode, which explicitly opts into experimental features.

What would be useful here is if you could replace `{ kind: string; id: string; scored?: string }` with `QueryMetadata` from `interface-types.ts` since it really is the same type.

Once that is changed, I think this is good to go in.

aeisenberg (approved):
Nice work!

Issue #606, opened 2020-10-02
Allow configuring arguments to qltest


jcreedcmu:
Could create a user-level setting for extra flags to pass on the commandline. A per-workspace setting might also work in principle, but I think that might arguably be a security risk if we passed it directly to shell execution.

PR #785, opened 2021-03-16
Add option to pass additional arguments when running tests

edoardopirovano:
This PR addresses #606 by adding a configuration option to pass additional arguments to the CLI when running tests. By default, this option is empty and no additional arguments are passed.

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -182,6 +182,12 @@
           "default": "%q on %d - %s, %r result count [%t]",
           "description": "Default string for how to label query history items. %t is the time of the query, %q is the query name, %d is the database name, %r is the number of results, and %s is a status string."
         },
+        "codeQL.runningTests.additionalTestArguments": {
+          "scope": "machine",
+          "type": "array",
+          "default": [],
+          "markdownDescription": "Additional command line arguments to pass to the CLI when [running tests](https://codeql.github.com/docs/codeql-cli/manual/test-run/). This setting should be an array of strings, each containing an argument to be passed."
+        },
         "codeQL.runningTests.numberOfThreads": {
           "scope": "window",
           "type": "integer",
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -509,12 +509,12 @@ export class CodeQLCliServer implements Disposable {
     testPaths: string[], workspaces: string[], options: TestRunOptions
   ): AsyncGenerator<TestCompleted, void, unknown> {
 
-    const subcommandArgs = [
+    const subcommandArgs = this.cliConfig.additionalTestArguments.concat([
       '--additional-packs', workspaces.join(path.delimiter),
       '--threads',
       this.cliConfig.numberTestThreads.toString(),
       ...testPaths
-    ];
+    ]);
 
     for await (const event of await this.runAsyncCodeQlCliCommand<TestCompleted>(['test', 'run'],
       subcommandArgs, 'Run CodeQL Tests', options.cancellationToken, options.logger)) {
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Allow using raw LGTM project slugs for fetching LGTM databases. [#769](https://github.com/github/vscode-codeql/pull/769)
 - Better error messages when BQRS interpretation fails to produce SARIF. [#770](https://github.com/github/vscode-codeql/pull/770)
 - Implement sorting of the query history view by name, date, and results count. [#777](https://github.com/github/vscode-codeql/pull/777)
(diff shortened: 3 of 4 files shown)


extensions/ql-vscode/package.json
```diff
@@ -182,6 +182,12 @@
           "default": "%q on %d - %s, %r result count [%t]",
           "description": "Default string for how to label query history items. %t is the time of the query, %q is the query name, %d is the database name, %r is the number of results, and %s is a status string."
         },
+        "codeQL.runningTests.additionalTestArguments": {
+          "scope": "window",
+          "type": "string",
+          "default": "",
+          "description": "Additional command line arguments to pass to the CLI when running tests."
```
aeisenberg, line 189: Please make sure the syntax is right and run this. I haven't verified that this is working correctly.

```suggestion
          "markdownDescription": "Additional command line arguments to pass to the CLI when [running tests](https://codeql.github.com/docs/codeql-cli/manual/test-run/)."
```

extensions/ql-vscode/src/cli.ts
```diff
@@ -509,12 +509,12 @@ export class CodeQLCliServer implements Disposable {
     testPaths: string[], workspaces: string[], options: TestRunOptions
   ): AsyncGenerator<TestCompleted, void, unknown> {
 
-    const subcommandArgs = [
+    const subcommandArgs = this.cliConfig.additionalTestArguments.toString().split(' ').concat([
```
aeisenberg, line 512: This will create empty arguments if there are duplicate spaces. Try using a regex instead. I think this will work better.

```suggestion
    const subcommandArgs = this.cliConfig.additionalTestArguments.toString().split(/\s+/).concat([
```

aeisenberg (changes requested):

edoardopirovano:
Tested after the changes and can confirm the option works as expected. In particular, I tried the `--slice` and `--show-extractor-output` options and observed the corresponding changes in output which confirm that the options are being picked up.

extensions/ql-vscode/package.json
```diff
@@ -182,6 +182,12 @@
           "default": "%q on %d - %s, %r result count [%t]",
           "description": "Default string for how to label query history items. %t is the time of the query, %q is the query name, %d is the database name, %r is the number of results, and %s is a status string."
         },
+        "codeQL.runningTests.additionalTestArguments": {
+          "scope": "window",
```
adityasharad, line 186: I suggest making this more restrictive: use `machine` just like we do for `codeql.cli.executablePath`. This ensure the setting can only be configured at the user level and prevents malicious workspaces from executing arbitrary code.

adityasharad (changes requested):

extensions/ql-vscode/package.json
aeisenberg: Is this really a danger in this case? Can the `run test` command invoke arbitrary commands? I guess it would if you can specify a build command for a test database, but not sure if that's possible.

extensions/ql-vscode/src/cli.ts
```diff
@@ -509,12 +509,12 @@ export class CodeQLCliServer implements Disposable {
     testPaths: string[], workspaces: string[], options: TestRunOptions
   ): AsyncGenerator<TestCompleted, void, unknown> {
 
-    const subcommandArgs = [
+    const subcommandArgs = this.cliConfig.additionalTestArguments.toString().split(/\s+/).concat([
```
aeisenberg, line 512: Comment from @henrymercer. This will fail things like `--string-arg "hello world"`. We will need some more sophisticated parsing here.

aeisenberg: You could change the type of the setting to `array` and each arg is one string element of the array, but then they can only be editted in the settings.json see: https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema

This may not be so bad as long as you have a proper description in the text.

aeisenberg (changes requested):

extensions/ql-vscode/package.json
adityasharad: Mainly being cautious. Saves us revisiting this in future if we introduce such an option. Right now, I think it's only potentially vulnerable through some indirection, e.g. if the controlled workspace points the test/extractor to a different folder it controls.

aeisenberg: That's fine by me, probably best to avoid. I just wanted to make sure there wasn't a specific flaw we need to prevent.

edoardopirovano:
Have addressed the above comments and squashed into one commit.

aeisenberg (approved):
Nice.

aeisenberg:
@github/docs-content-dsp We probably need to add some documentation for this new functionality.

shati-patel:
> @github/docs-content-dsp We probably need to add some documentation for this new functionality.

Thanks! I've opened an internal docs issue.

Issue #724, opened 2021-01-18
Ensure query error messages in popups are succinct


aeisenberg:
**Is your feature request related to a problem? Please describe.**

When the query server throws an exception while running a query, it is easy to ignore or be unaware of it. In many cases, the query run still returns results, just not in the expected format. Users can be confused by this.

Exceptions are sent to the output log, but there is no notification that an error has occurred at all. If the user is not monitoring the output log, then they will not be aware that an error has occurred at all.

For example, running with invalid query metadata can send an exception like this to the console:

```
A fatal error occurred: Could not process query metadata.
Error was: No query kind specified(NO_KIND_SPECIFIED).
[2021-01-18 14:08:15] Exception caught at top level: Could not process query metadata.
                      Error was: No query kind specified(NO_KIND_SPECIFIED).
                      com.semmle.cli2.bqrs.InterpretCommand.executeSubcommand(InterpretCommand.java:123)
                      com.semmle.cli2.picocli.SubcommandCommon.executeWithParent(SubcommandCommon.java:414)
                      com.semmle.cli2.execute.CliServerCommand.lambda$executeSubcommand$0(CliServerCommand.java:67)
                      com.semmle.cli2.picocli.SubcommandMaker.runMain(SubcommandMaker.java:201)
                      com.semmle.cli2.execute.CliServerCommand.executeSubcommand(CliServerCommand.java:67)
                      com.semmle.cli2.picocli.SubcommandCommon.call(SubcommandCommon.java:430)
                      com.semmle.cli2.picocli.SubcommandMaker.runMain(SubcommandMaker.java:201)
                      com.semmle.cli2.picocli.SubcommandMaker.runMain(SubcommandMaker.java:209)
                      com.semmle.cli2.CodeQL.main(CodeQL.java:91)
```

**Describe the solution you'd like**

It would be useful if some relevant part of this message was put in a message dialog so that users would be aware that some unexpected behaviour occurred.

I imagine that we'd want to do some regex matching on messages coming in to find things that look like errors. If the regex matches, then we place that message (or part of the message) into a popup.

edoardopirovano:
Hmm, on my install I already seem to get a popup advising me to look at the log when an error has occurred in the CLI. For instance, see the screenshot below for what happens when I run a query with invalid metadata (as in your example). Could this be something to do with my configuration? Or are there errors in the query server that do not result in a CLI error that we would like to be outputting? If so, could you give me an example of one and I'm happy to work on this issue then?

[...]

PR #786, opened 2021-03-17
Shorten error messages displayed in popups

edoardopirovano:
This addresses #724 by running a few regexes on error messages before they are displayed in pop-ups to try to shorten them by removing elements such as log lines and stack traces that will be too lengthy to show meaningfully in a popup.

### DIFF
--- a/extensions/ql-vscode/src/helpers.ts
+++ b/extensions/ql-vscode/src/helpers.ts
@@ -11,6 +11,7 @@ import {
 } from 'vscode';
 import { CodeQLCliServer } from './cli';
 import { logger } from './logging';
+import { shortenErrorMessage } from './pure/helpers-pure';
 
 /**
  * Show an error message and log it to the console
@@ -29,8 +30,9 @@ export async function showAndLogErrorMessage(message: string, {
   items = [] as string[],
   fullMessage = undefined as (string | undefined)
 } = {}): Promise<string | undefined> {
-  return internalShowAndLog(message, items, outputLogger, Window.showErrorMessage, fullMessage);
+  return internalShowAndLog(shortenErrorMessage(message), items, outputLogger, Window.showErrorMessage, fullMessage);
 }
+
 /**
  * Show a warning message and log it to the console
  *
--- a/extensions/ql-vscode/test/pure-tests/helpers-pure.test.ts
+++ b/extensions/ql-vscode/test/pure-tests/helpers-pure.test.ts
@@ -35,7 +35,7 @@ describe('helpers-pure', () => {
       runTest(input, expected);
     });
 
-    it('removes log lines lines', async () => {
+    it('removes log lines', async () => {
       const input = ['first line', '[2020-02-15 09:10:15] Some logging information', 'last line'];
       const expected = ['first line', 'last line'];
       runTest(input, expected);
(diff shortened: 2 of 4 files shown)


extensions/ql-vscode/src/helpers.ts
```diff
@@ -29,8 +29,34 @@ export async function showAndLogErrorMessage(message: string, {
...
-  return internalShowAndLog(message, items, outputLogger, Window.showErrorMessage, fullMessage);
+  return internalShowAndLog(shortenErrorMessage(message), items, outputLogger, Window.showErrorMessage, fullMessage);
 }
+
+/**
+ * Creates a succint version of an error message, e.g. for displaying in a pop-up.
+ * @param errorMessage The error message to shorten.
+ * @return A shortened version of the error message.
+ */
+function shortenErrorMessage(errorMessage: string): string {
```
aeisenberg, line 40: Could you add some unit tests for this?

extensions/ql-vscode/src/helpers.ts
```diff
@@ -29,8 +29,34 @@ export async function showAndLogErrorMessage(message: string, {
...
+  // Filter out lines corresponding to Java stack traces.
+  const stackTraceLine = /\n\s*(com|org|net)\..*\.java:[0-9]+\)(?=\n)/g;
```
aeisenberg, line 42: Would this be better so you're not relying on newlines, which may not work on windows.

```suggestion
  const stackTraceLine = /^\s*(com|org|net)\..*\.java:[0-9]+\)$/g;
```

extensions/ql-vscode/src/helpers.ts
```diff
@@ -29,8 +29,34 @@ export async function showAndLogErrorMessage(message: string, {
...
+  let succintMessage = errorMessage.toString().replace(stackTraceLine, '');
+
+  // Filter out lines corresponding to log lines.
+  const logLine = /\n\[[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9] [0-9][0-9]:[0-9][0-9]:[0-9][0-9]\].*/g;
```
aeisenberg, line 46:
```suggestion
  const logLine = /^\[[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9] [0-9][0-9]:[0-9][0-9]:[0-9][0-9]\].*/g;
```

extensions/ql-vscode/src/helpers.ts
```diff
@@ -29,8 +29,34 @@ export async function showAndLogErrorMessage(message: string, {
...
+  succintMessage = succintMessage.replace(logLine, '');
+
+  // Trim indentation at the start of lines.
+  const indentedLine = /\n\s+/g;
```
aeisenberg, line 50:
```suggestion
  const indentedLine = /^\s+/g;
```

extensions/ql-vscode/src/helpers.ts
```diff
@@ -29,8 +29,34 @@ export async function showAndLogErrorMessage(message: string, {
...
+  succintMessage = succintMessage.replace(indentedLine, '\n');
+
+  // Remove duplicated lines.
+  const duplicatedLine = /\n(.*)\n\1/g;
```
aeisenberg, line 54: Careful here. The `\n` won't work on windows. You should probably be working on this line-by-line instead of running the regex over each line.

aeisenberg (changes requested):
Searching for `\n` will not work on windows. It might be easier to split line-by-line and run the regex on each line separately.

Also, now that I think about it, maybe we only need to show the first line or two of the error messages (ie- only show up to, but excluding, where the stack trace starts). This would make the code much simpler.

edoardopirovano:
Hmm, I agree that in the specific case of the error message that started this PR looking at the first two lines would achieve the desired result. However, I think that trying to remove things we don't want from the message while preserving it as much as possible seems better in general, since there might be cases where an important bit of information on a later line. I've fixed the regexes to work on Windows line endings, and added unit tests that run with both types of line endings (they'll still break on the line endings from pre-2001 Mac OS, but I imagine we don't support this).

extensions/ql-vscode/test/pure-tests/helpers-pure.test.ts
```diff
@@ -19,4 +19,38 @@ describe('helpers-pure', () => {
...
+      expect(shortenErrorMessage(input.join(LINUX_LINE_END)).split(LINUX_LINE_END)).eql(expected);
+    }
+
+    it('removes stack traces', async () => {
+      const input = ['first line', '    com.a.java.class.command(Class.java:65)', 'last line'];
+      const expected = ['first line', 'last line'];
+      runTest(input, expected);
+    });
+
+    it('removes log lines lines', async () => {
```
aeisenberg, line 38:
```suggestion
    it('removes log lines', async () => {
```

extensions/ql-vscode/test/pure-tests/helpers-pure.test.ts
```diff
@@ -19,4 +19,38 @@ describe('helpers-pure', () => {
...
+    });
+
+    it('strips whitespace', async () => {
+      const input = ['    first line', '    last line'];
+      const expected = ['first line', 'last line'];
+      runTest(input, expected);
+    });
+
+    it('removes duplicate lines', async () => {
+      const input = ['first line', 'last line', 'last line'];
```
aeisenberg, line 51: What's the behaviour here:
```suggestion
      const input = ['first line', 'last line', 'last line', 'last line'];
```

Looking at the regex, I'm not sure. Could you add a unit test for that?

aeisenberg (changes requested):
Can you also add a test for the sample error message from #724?

I do have some concern that this solution is complex and fragile. All of our error messages have their important pieces in the first line or two. This solution is probably fine since it's unlikely that we will be changing formats. I'd just like to see some more tests as described.

edoardopirovano:
I've added additional tests per your comments, but on further reflection I think I actually do agree that this is maybe a little fragile and simply truncating the message after two lines is the easier way to address this. Happy for you to close this PR and I'll do that instead if that's still your thinking!

aeisenberg:
Sure. That sounds like a better plan.

Issue #774, opened 2021-03-08
Test failures are not properly reported


Marcono1234:
**Version**
Extension Version: 1.4.3
CodeQL CLI Version: 2.4.4

**Describe the bug**
It appears the VSCode extension invokes CodeQL CLI with `--format jsonz`. With that format only a small portion of the information is written to stdout, most of the information is inside the JSON data. With #626 some of the JSON data is read, but apparently not all. Most notably, the extension is apparently not reading `failureDescription`.

This makes understanding test failures pretty difficult, for example when no `.expected` file exists (see also https://github.com/github/codeql/issues/5343).

**To Reproduce**
1. Clone https://github.com/github/codeql
2. Delete [`cpp/ql/test/library-tests/dataflow/DefaultTaintTracking/annotate_sinks_only/remote-flow-sink.expected`](https://github.com/github/codeql/blob/main/cpp/ql/test/library-tests/dataflow/DefaultTaintTracking/annotate_sinks_only/remote-flow-sink.expected)
3. Run only the corresponding test file `remote-flow-sink.ql` from the "Test" tab of VSCode
:x: The output only says (roughly)
    > 0 tests passed; 1 tests failed:
    > FAILED: path/to/test.ql
    > failed: path/to/test.ql
    > Query produced no results.

**Expected behavior**
The VSCode extension should evaluate the `failureDescription`, if present, (and maybe more members of the JSON data) of the test result and inform the user.

aeisenberg:
Thanks for raising this issue. I see how this could have clarified things on https://github.com/github/codeql/issues/5343.

Marcono1234:
I was mostly writing this from the perspective of `InlineExpectationsTest` which has empty `.actual` and `.expected` files, but it appears for tests with a non-empty `.actual` file, even when the `.expected` file is missing (and a `failureDescription` is present), it still generates useful `diff` results. So maybe the `failureDescription` should be evaluated in addition to `diff`?

Marcono1234:
Has been fixed by #788; in version 1.4.4 of the extension the missing `.expected` file is now correctly reported.

PR #788, opened 2021-03-18
Report description for test failure when possible

edoardopirovano:
Addresses #774 by reading the `failureDescription` field for a test failure and when this is present reporting this instead of the diff. I believe this is the right choice since as I understand it when this field is present, the diff will not be useful as it means something went more catastrophically wrong (e.g. we couldn't find the expected output).

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -104,6 +104,7 @@ export interface TestCompleted {
   evaluationMs: number;
   expected: string;
   diff: string[] | undefined;
+  failureDescription?: string;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -104,6 +104,7 @@ export interface TestCompleted {
   evaluationMs: number;
   expected: string;
   diff: string[] | undefined;
+  failureDescription: string;
```
aeisenberg, line 107:
```suggestion
  failureDescription?: string;
```

aeisenberg (approved):

Marcono1234:
Thanks! Do you or should I now close #774? Or are you in general only closing issues once the changes have been released?

edoardopirovano:
This is in the 1.4.4 release so I think it should be okay to close the issue (I don't have the access to so, though). Ooh, thanks for heads up, I wasn't aware of those keywords.

aeisenberg:
If in your PR description, you put `Fixes #XXX`, then issue #XXX will be automatically closed when the PR is merged.

PR #792, opened 2021-03-19
Limit error messages shown in popups to 2 lines

edoardopirovano:
Addresses #724 by limiting error messages in popups to 2 lines. Note that this required changing around a couple of error messages that were likely to be more than 2 lines but we wanted to show all of to ensure that they do not get trimmed.

### DIFF
--- a/extensions/ql-vscode/src/helpers.ts
+++ b/extensions/ql-vscode/src/helpers.ts
@@ -29,8 +29,13 @@ export async function showAndLogErrorMessage(message: string, {
   items = [] as string[],
   fullMessage = undefined as (string | undefined)
 } = {}): Promise<string | undefined> {
-  return internalShowAndLog(message, items, outputLogger, Window.showErrorMessage, fullMessage);
+  return internalShowAndLog(dropLinesExceptInitial(message), items, outputLogger, Window.showErrorMessage, fullMessage);
 }
+
+function dropLinesExceptInitial(message: string, n = 2) {
+  return message.toString().split(/\r?\n/).slice(0, n).join('\n');
+}
+
 /**
  * Show a warning message and log it to the console
  *
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -640,8 +640,11 @@ export async function compileAndRunQueryAgainstDatabase(
         formattedMessages.push(formatted);
         qs.logger.log(formatted);
       }
-      if (quickEval && formattedMessages.length <= 3) {
-        showAndLogErrorMessage('Quick evaluation compilation failed: \n' + formattedMessages.join('\n'));
+      if (quickEval && formattedMessages.length <= 2) {
+        // If there are more than 2 error messages, they will not be displayed well in a popup
+        // and will be trimmed by the function displaying the error popup. Accordingly, we only
+        // try to show the errors if there are 2 or less, otherwise we direct the user to the log.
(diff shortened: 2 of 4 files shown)


extensions/ql-vscode/src/helpers.ts
```diff
@@ -29,8 +29,14 @@ export async function showAndLogErrorMessage(message: string, {
   items = [] as string[],
   fullMessage = undefined as (string | undefined)
 } = {}): Promise<string | undefined> {
-  return internalShowAndLog(message, items, outputLogger, Window.showErrorMessage, fullMessage);
+  const NUM_LINES_TO_KEEP = 2;
+  return internalShowAndLog(dropLinesExceptInitial(message, NUM_LINES_TO_KEEP), items, outputLogger, Window.showErrorMessage, fullMessage);
 }
+
+function dropLinesExceptInitial(message: string, n: number) {
```
aeisenberg, line 36: nit: We only ever use `2`, so this might be easier and you can avoid the constant. Not a big deal and either way is fine.

```suggestion
function dropLinesExceptInitial(message: string, n = 2) {
```

extensions/ql-vscode/src/helpers.ts
```diff
@@ -29,8 +29,14 @@ export async function showAndLogErrorMessage(message: string, {
   items = [] as string[],
   fullMessage = undefined as (string | undefined)
 } = {}): Promise<string | undefined> {
-  return internalShowAndLog(message, items, outputLogger, Window.showErrorMessage, fullMessage);
+  const NUM_LINES_TO_KEEP = 2;
+  return internalShowAndLog(dropLinesExceptInitial(message, NUM_LINES_TO_KEEP), items, outputLogger, Window.showErrorMessage, fullMessage);
```
aeisenberg, line 33: See other suggestion below.

```suggestion
  return internalShowAndLog(dropLinesExceptInitial(message), items, outputLogger, Window.showErrorMessage, fullMessage);
```

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -635,8 +635,8 @@ export async function compileAndRunQueryAgainstDatabase(
         formattedMessages.push(formatted);
         qs.logger.log(formatted);
       }
-      if (quickEval && formattedMessages.length <= 3) {
-        showAndLogErrorMessage('Quick evaluation compilation failed: \n' + formattedMessages.join('\n'));
+      if (quickEval && formattedMessages.length <= 2) {
+        showAndLogErrorMessage('Quick evaluation compilation failed: ' + formattedMessages.join('\n'));
```
aeisenberg, line 639: This section seems strange to me. Why `<= 2`? This isn't a problem with your code since it was `<= 3` and neither magic number is clear to me. If you understand this, can you add a comment here? If not, then can you add a `TODO` so that we can hopefully fix later.

aeisenberg (commented):
Some small suggestions and then ready to go in.

extensions/ql-vscode/src/run-queries.ts
edoardopirovano: My understanding is that this is simply a number of compilation errors that it would be reasonable to show in a pop-up, and if there's more errors than that we instead show the generic error in `compilationFailedErrorTail` which says to look at the full output in the log. I think changing it from `3` to `2` is reasonable if we are limiting other log messages to two lines - three seems like a lot of compilation messages to try and show in a pop-up anyways.

aeisenberg (approved):

PR #803, opened 2021-03-22
Fix running tests when ms-python is installed.

### LINKED ISSUE #802: No output while running tests when Python extension is also installed
**Describe the bug**
When both the CodeQL extension and the `ms-python` extension are installed, there is no output or logging when running tests. A while after starting a test, it goes back to the "grey stop symbol" icon (from "blue ellipsis").
(originally observed by @tausbn)

**To Reproduce**
- Install CodeQL for VS Code 1.4.3 or later.
- Install the VS Code Python extension.
- Open a CodeQL workspace (e.g. `github/codeql`) and run tests from the Test Explorer view.

**Expected behavior**
- The tests run, the Test Explorer shows the status, and the Output windows show the logs.

**Diagnosis**

The VS Code Python extension has a dependency that overrides `String.prototype.startsWith` for the entire VS Code Node process. This breaks code (in our extension or others) that depends on the 2-argument version of `startsWith`.
@alexet has filed upstream issues at https://github.com/microsoft/vscode-python/issues/15744 and https://github.com/simonepri/pidtree/issues/13.

**Workaround**
Affected users may uninstall or disable the VS Code Python extension to work around this problem.

alexet:
Fixes #802 by manually implementing `startsWith`. This is probably a slight performance hit but I don't think we can do much about it.

This is the only place where we use the two argument version whcih is the one that becomes broken.

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -907,6 +907,20 @@ class SplitBuffer {
   }
 
   /**
+   * A version of startsWith that isn't overriden by a broken version of ms-python.
+   * 
+   * The definition comes from
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
+   * which is CC0/public domain
+   * 
+   * See https://github.com/github/vscode-codeql/issues/802 for more context as to why we need it.
+   */
+  private static startsWith(s: string, searchString: string, position: number): boolean {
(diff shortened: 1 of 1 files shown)


adityasharad (approved):
Looks sensible. I first thought we should reset the prototype, but that creates similar downstream effects for other extensions.

adityasharad:
Forgot - can you add a changelog entry? This PR's title and link should be enough.

alexet:
If we change the prototype then it depends which extension gets there first.

PR #815, opened 2021-03-29
Add feature to jump to the .ql file referenced by a .qlref

### LINKED ISSUE #654: Feature request: jump to source from .qlref files
Navigating from a `.qlref` file in a test directory to the `.ql` file it's testing takes more steps than I'd like, especially when the source and test are in different qlpacks.

In the Explorer sidebar view, the right-click menu for a `.qlref` file includes an option to open the `.ql` file which the `.qlref` file refers to.

edoardopirovano:
This PR adds an option to jump from a `.qlref` file to the `.ql` file it references. Note that it should not be merged yet, as it relies on a new CLI call `resolve qlref` which is not released yet.

Closes #654

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -151,7 +151,7 @@ export class CodeQLCliServer implements Disposable {
   /**
    * CLI version where the `codeql resolve qlref` command is available.
    */
-  private static CLI_VERSION_WITH_RESOLVE_QLREF = new SemVer('2.5.1');
+  public static CLI_VERSION_WITH_RESOLVE_QLREF = new SemVer('2.5.1');
 
   /** The process for the cli server, or undefined if one doesn't exist yet */
   process?: child_process.ChildProcessWithoutNullStreams;
@@ -937,11 +937,11 @@ class SplitBuffer {
 
   /**
    * A version of startsWith that isn't overriden by a broken version of ms-python.
-   * 
+   *
    * The definition comes from
    * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
    * which is CC0/public domain
-   * 
+   *
    * See https://github.com/github/vscode-codeql/issues/802 for more context as to why we need it.
    */
   private static startsWith(s: string, searchString: string, position: number): boolean {
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -73,6 +73,11 @@ export interface UpgradesInfo {
  */
 export type QlpacksInfo = { [name: string]: string[] };
 
+/**
+ * The expected output of `codeql resolve qlref`.
+ */
+export type QlrefInfo = { resolvedPath: string };
+
 // `codeql bqrs interpret` requires both of these to be present or
 // both absent.
 export interface SourceInfo {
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/extension.ts
```diff
@@ -480,6 +480,23 @@ async function activateWithInstalledDistribution(
...
+    if (qs !== undefined) {
+      if (await cliServer.supportsResolveQlref()) {
+        const resolved = await cliServer.resolveQlref(selectedQuery.path);
+        const uri = Uri.file(resolved);
+        await window.showTextDocument(uri, { preview: false });
+      } else {
+        helpers.showAndLogErrorMessage(
+          'Jumping from a .qlref file to the .ql file it references is not '
+          + 'supported with the CLI version you are running.\n'
+          + 'Please upgrade your CLI to use this feature.');
```
aeisenberg, line 495: Not necessary, but it would be neat if you could add a link here that can automatically run the check for updates command when run. Not even sure if this is possible.

aeisenberg (commented):
This looks great. Let's get the related PR in semmle-code merged and then we can merge this one.

extensions/ql-vscode/src/cli.ts
```diff
@@ -143,6 +148,11 @@ export class CodeQLCliServer implements Disposable {
    */
   private static CLI_VERSION_WITH_DOWNGRADES = new SemVer('2.4.4');
 
+  /**
+   * CLI version where the `codeql resolve qlref` command is available.
+   */
+  private static CLI_VERSION_WITH_RESOLVE_QLREF = new SemVer('2.5.1');
+
```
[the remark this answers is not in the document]
edoardopirovano: Agree this needs better managing as we add more such checks. This is probably something for a future PR, however. Another thing to consider is how far back we want to support old CLI versions. I imagine we don't want to keep these checks in forever, but rather want to eventually remove them and force users to update their CLI?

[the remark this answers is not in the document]
aeisenberg: I've got some ideas here. Maybe I'll get to it this week.

extensions/ql-vscode/src/extension.ts
edoardopirovano: I agree this would be nice, but as far as I can tell it's a little tricky in the current setup since the functions we need to call to do this live in the code that activates the extension and are no longer in scope at this point. I think we should refactor so that there is a function for updating the CLI inside `DistributionManager` that we can call after the extension has been activated, but this seems out of the scope of this PR.

edoardopirovano:
Updated to account for the minor CLI changes in the internal PR, which has now been merged. This is now ready for merging, although doing so will mean we should not release a new version of the extension until the new CLI is released.

extensions/ql-vscode/src/extension.ts
```diff
@@ -480,6 +480,23 @@ async function activateWithInstalledDistribution(
...
+    if (qs !== undefined) {
+      if (await cliServer.supportsResolveQlref()) {
+        const resolved = await cliServer.resolveQlref(selectedQuery.path);
+        const uri = Uri.file(resolved.resolvedPath);
+        await window.showTextDocument(uri, { preview: false });
+      } else {
+        helpers.showAndLogErrorMessage(
+          'Jumping from a .qlref file to the .ql file it references is not '
+          + 'supported with the CLI version you are running.\n'
+          + 'Please upgrade your CLI to use this feature.');
```
aeisenberg, line 495: Would be nice to include the minimum cli version here. Something like:

```suggestion
          + `Please upgrade your CLI to version ${CLI_VERSION_WITH_RESOLVE_QLREF} or later to use this feature.`);
```

aeisenberg (approved):
This looks good and can is ready to be merged, but I think you're right that we should hold off for now until cli v2.5.1 is released. It would be annoying if you got this message and had no way of upgrading to a sufficient version.

PR #818, opened 2021-04-01
Add a canary setting to avoid caching AST viewer queries

aeisenberg:
When codeql library developers are working on PrintAST queries, it is
not easy to use the AST Viewer. The AST Viewer caches results so that
multiple calls to view the AST of the same file are nearly
instantaneous.

However, this breaks down if you are changing the actual queries that
perform AST viewing. In this case, you do not want the cache to be
active.

This commit adds an undocumented setting that prevents caching. To
enable, set:

```
"codeQL.isCanary": true,
"codeQL.astViewer.disableCache": true
```

Note that *both* settings must be true for this to work.

This behaviour and all canary behaviour should be documented somewhere.
I will add that later.

Any suggestions on where this should go would be appreciated. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/config.ts
+++ b/extensions/ql-vscode/src/config.ts
@@ -41,6 +41,7 @@ const ROOT_SETTING = new Setting('codeQL');
 
 // Global configuration
 const TELEMETRY_SETTING = new Setting('telemetry', ROOT_SETTING);
+const AST_VIEWER_SETTING = new Setting('astViewer', ROOT_SETTING);
 const GLOBAL_TELEMETRY_SETTING = new Setting('telemetry');
 
 export const LOG_TELEMETRY = new Setting('logTelemetry', TELEMETRY_SETTING);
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/config.ts
```diff
@@ -279,3 +279,8 @@ export const CANARY_FEATURES = new Setting('canary', ROOT_SETTING);
 export function isCanary() {
   return !!CANARY_FEATURES.getValue<boolean>();
 }
+
+/**
+ * Avoids caching in the AST viewer if the user is also a canary user.
+ */
+export const NO_CACHE_AST_VIEWER = new Setting('noCacheAstViewer', ROOT_SETTING);
```
adityasharad, line 286: Would it make sense to have an `astViewer` or `viewAst` parent setting, and then a `noCache` or `disableCache` underneath it? Not sure if we expect future AST viewer-related options.

adityasharad (approved):
Code LGTM, one suggestion on setting groups. We could add an "advanced users" section to the public docs.

extensions/ql-vscode/src/config.ts
aeisenberg: I like that better. Made the change.

aeisenberg:
I'll document it internally.

PR #829, opened 2021-04-14
Add option to disable animated stickers

Faer1-brent-gamble:
Add option to disable animated stickers 358


adityasharad:
Abuse PR.

PR #833, opened 2021-04-19
Avoid printing a stack trace when there is no resultsPath

aeisenberg:
I don't know exactly when this can happen, but a customer has just
shown me a stack trace like this:

```
TypeError: Cannot destructure property 'resultsPath' of 'resultsPaths' as it is undefined.
    at Object.interpretResults (/xxx/.vscode/extensions/github.vscode-codeql-1.4.5/out/query-results.js:120:13)
    at InterfaceManager._getInterpretedResults (/xxx/.vscode/extensions/github.vscode-codeql-1.4.5/out/interface.js:377:45)
    at InterfaceManager.showResultsAsDiagnostics (/xxx/.vscode/extensions/github.vscode-codeql-1.4.5/out/interface.js:447:43)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async InterfaceManager.handleMsgFromView (/xxx/.vscode/extensions/github.vscode-codeql-1.4.5/out/interface.js:151:29)
```

This commit will avoid printing this stack trace and instead print
a more descriptive message to the logs.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/interface.ts
+++ b/extensions/ql-vscode/src/interface.ts
@@ -561,24 +561,26 @@ export class InterfaceManager extends DisposableObject {
     sourceInfo: cli.SourceInfo | undefined,
     sourceLocationPrefix: string,
     sortState: InterpretedResultsSortState | undefined
-  ): Promise<Interpretation> {
+  ): Promise<Interpretation | undefined> {
+    if (!resultsPaths) {
+      this.logger.log('No results path. Cannot display interpreted results.');
+      return undefined;
+    }
+
     const sarif = await interpretResults(
       this.cliServer,
       metadata,
       resultsPaths,
       sourceInfo
     );
+
     sarif.runs.forEach(run => {
       if (run.results !== undefined) {
         sortInterpretedResults(run.results, sortState);
       }
     });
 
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/interface.ts
```diff
@@ -561,24 +561,26 @@ export class InterfaceManager extends DisposableObject {
...
-  ): Promise<Interpretation> {
+  ): Promise<Interpretation | undefined> {
     const sarif = await interpretResults(
       this.cliServer,
       metadata,
       resultsPaths,
       sourceInfo
     );
+
+    if (!resultsPaths) {
```
adityasharad, line 572: Should we do this before the call to `interpretResults` above? Or does that populate `resultsPaths`?

aeisenberg: Yes, I think this should be moved up.

adityasharad (approved):
Looks reasonable. I assume the caller is prepared to handle an `undefined` response?

aeisenberg:
> Looks reasonable. I assume the caller is prepared to handle an `undefined` response?

Yes.

PR #838, opened 2021-04-21
Forward all query metadata to the queryserver

henrymercer:
This PR forwards all the query metadata to the queryserver.

Currently, only `@kind` and `@id` are forwarded for most users.  This results in metadata errors when running some queries, even if the queries being run had the correct metadata.  For example, the `@kind metric` query requires either the `@tags summary` metadata or the `@metricType` metadata key to be defined.  However the VS Code extension doesn't currently forward the `@tags` and `@metricType` properties, so the queryserver reports errors whenever a user tries to run these queries.

With this PR, we forward all the query metadata to the queryserver, so queries like these will no longer report metadata errors.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Avoid showing an error popup when user runs a query with `@kind table` metadata. [#814](https://github.com/github/vscode-codeql/pull/814)
 - Add an option to jump from a .qlref file to the .ql file it references. [#815](https://github.com/github/vscode-codeql/pull/815)
 - Avoid opening the results panel when a database is deleted. [#831](https://github.com/github/vscode-codeql/pull/831)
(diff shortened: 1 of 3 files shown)


aeisenberg (commented):
This looks fine.  But just want to be sure that if a user accidentally includes invalid metadata, will the interpretation fail?

henrymercer:
That's correct.  Perhaps this change warrants an entry in the CHANGELOG?

henrymercer:
We currently generate metadata errors in (at least) two places: the metadata verifier, which is invoked during compilation, and the interpretation itself.  For VS Code, only the metadata errors during the interpretation itself are problematic.  I don't believe there are any cases where forwarding additional metadata would make the interpretation start failing.

aeisenberg (approved):
Thanks for the clarification. Yes, I think this needs a changelog entry.

PR #846, opened 2021-04-23
Watch for changes in directory structure

### LINKED ISSUE #799: Test Explorer consistently does not detect certain file actions
**Version**
CodeQL CLI v.2.4.6
Extension v1.4.4

OS: Windows 10

**Describe the bug**
It appears the Test Explorer consistently does not detect certain file actions, requiring a restart of VSCode for them to be detected (or adding new test files; this seems to update the view as well).

Affected:
- Extension never detects deletion of folders containing tests
- Extension never detects renaming folders containing tests
- Extension never detects moving folders containing tests, e.g. moving `/some-tests/test.qlref` inside of `/parent-of-tests`

It is probably not an issue with `codeql resolve tests`, because that seems to report the correct directories.

**To Reproduce**
Try performing any of the actions above and see whether the Test Explorer notices the change.

In case you cannot reproduce this, try opening https://github.com/Marcono1234/codeql-java-queries in VSCode as folder (have not tested as workspace), try the performing the actions described above with any of the folders under `codeql-custom-queries-java/tests` and see whether the Test Explorer notices the changes.

**Expected behavior**
The Test Explorer should notice the file actions described above and adjust the view accordingly.

**Additional context**
I think I have also experienced #310 a few times, however I cannot reproduce it.

aeisenberg:
This ensures that directories renamed, added or deleted are
properly checked to see if they contain tests. The test tree
will be correctly updated when any directory changes.s

Fixes #799.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/qltest-discovery.ts
+++ b/extensions/ql-vscode/src/qltest-discovery.ts
@@ -167,9 +167,11 @@ export class QLTestDiscovery extends Discovery<QLTestDiscoveryResults> {
   protected update(results: QLTestDiscoveryResults): void {
     this._testDirectory = results.testDirectory;
 
-    // Watch for changes to any `.ql` or `.qlref` file in any of the QL packs that contain tests.
     this.watcher.clear();
+    // Watch for changes to any `.ql` or `.qlref` file in any of the QL packs that contain tests.
     this.watcher.addWatch(new RelativePattern(results.watchPath, '**/*.{ql,qlref}'));
+    // need to explicitly watch for changes to directories themselves.
+    this.watcher.addWatch(new RelativePattern(results.watchPath, '**/'));
     this._onDidChangeTests.fire();
   }
 
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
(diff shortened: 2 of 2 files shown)


extensions/ql-vscode/src/qltest-discovery.ts
```diff
@@ -167,9 +167,11 @@ export class QLTestDiscovery extends Discovery<QLTestDiscoveryResults> {
   protected update(results: QLTestDiscoveryResults): void {
     this._testDirectory = results.testDirectory;
 
-    // Watch for changes to any `.ql` or `.qlref` file in any of the QL packs that contain tests.
     this.watcher.clear();
+    // Watch for changes to any `.ql` or `.qlref` file in any of the QL packs that contain tests.
     this.watcher.addWatch(new RelativePattern(results.watchPath, '**/*.{ql,qlref}'));
+    // need to explicitly watch for changes to directories themselves.
+    this.watcher.addWatch(new RelativePattern(results.watchPath, '**/'));
```
adityasharad, line 174: So this is all directories under a workspace folder? Any noticeable hit to performance on a large repo? Do we need to exclude database source archives?

aeisenberg: Nothing noticeable. I renamed the `ql` folder that is a submodule in the starter repo and there was no noticeable lag in the file explorer. The test explorer itself took a few seconds to update, but I think that's fair because it needs to re-discover the entire tree. 

We don't need to do any special handling for source archives since they are readonly and aren't watched for changes.

adityasharad (approved):

Issue #849, opened 2021-04-26
Integrate the workspace trust feature


aeisenberg:
This issue is about integrating the new workspace trust feature into vscode-codeql. See https://github.com/microsoft/vscode/issues/120251 for the general feature.

Implementing this feature properly would be an effective fix for #509. 

- [ ] Determine which settings require workspace trust. I found the following, but there may be more:
    - `codeQL.cli.executablePath`
    - `codeQL.runningTests.additionalTestArguments`
- [ ] Determine which commands require workspace trust
    - I don't think there are any, but worth taking a deeper look.

Issue #509, opened 2020-07-21
Allow CodeQL CLI version (not path) to be specified in workspace settings


dbartol:
**Is your feature request related to a problem? Please describe.**
The current settings schema for the extension lets you specify an arbitrary path to the CodeQL CLI. For security reasons, this setting is ignored when specified in a workspace settings file, because a workspace settings file is usually pulled down as part of the source code. The makes it difficult to specify which version of the CLI a workspace is compatible with. I have this problem to some extent locally, but for customers with multiple query authors working in a repo that depends on a particular version of the CLI and standard libraries, the lack of a way to ensure a consistent version of the CLI is even more frustrating.

**Describe the solution you'd like**
I'd like a new extension setting, `CodeQL > Cli: Version`, which allows you the specify the version tag of a public CLI release, and will ensure that the extension downloads and uses that specific release of the CLI. Hopefully, restricting the source of the CLI to our own release feed is sufficiently secure that we can automatically download from there even if it's in the workspace settings. If that is still too scary, though, we could at least put up a prompt saying "this workspace requests version `v2.3.6` of the CodeQL CLI. Do you want to download and use it?".

**Describe alternatives you've considered**
I've considered making my User-level CodeQL CLI path point to a script that tries to figure out the workspace and associated version, but that seems cumbersome for actual customers.

aeisenberg:
I think this makes sense, especially if this helps our users to migrate away from unsupported products. Our downloading and distribution logic is already quite complex, so I'd want to implement this carefully.

Some questions (and answers):

- Presumably, this setting takes precedence over a `codeql.cli.executablePath` setting.
- Should we deprecate `codeql.cli.executablePath` since that would become largely superfluous with the new setting? I realize that this would make using custom builds more challenging, but is that really an issue for our customers or is it only internal people who would need this?

dbartol:
I think `executablePath` is necessary both for internal developers and for external developers who choose to install the CLI in a particular local path. The simplest example is a developer who uses both the extension and runs the CLI directly. They probably want the CLI on their path, and want to use exactly the same CLI from within the extension.

I would have expected the settings to interact as follows:
- If `codeql.cli.executablePath` is set, then use the CLI at that path, else
- If `codeql.cli.cliVersion` is set, download and use that version, else
- If `codeql.cli.includePrerelease` is set, download and use the latest prerelease version, else
- Download and use the latest non-prerelease version

[...]

PR #861, opened 2021-05-12
Use the workspace trust feature

shati-patel:
Will fix https://github.com/github/vscode-codeql/issues/849 and will also (indirectly) fix https://github.com/github/vscode-codeql/issues/509. 

Uses the syntax from https://github.com/microsoft/vscode/issues/120251#issuecomment-825832603 to let the CodeQL extension work with workspace trust. For users who have enabled the workspace trust feature:
- If the user trusts the workspace: all CodeQL features are enabled
- If the user doesn't trust the workspace: Certain settings (that could execute arbitrary code from the workspace) are disabled at the workspace level, and the extension ignores them. This is what the UI looks like:
image

⚠️ ~Note: Currently, you can't use the CodeQL extension at all in untrusted workspaces. This is because it depends on the [Test Explorer](https://marketplace.visualstudio.com/items?itemName=hbenl.vscode-test-explorer), which currently doesn't work in untrusted workspaces. PR: https://github.com/hbenl/vscode-test-explorer/pull/204.~ Fixed!

## Checklist

- [x] Once VS Code is released with workspace trust enabled by default, we need to update the minimum VS Code version required to run this extension. See: 
https://github.com/github/vscode-codeql/blob/72d57eec6ecfb6ff10d59a64d8403acb68ef05b5/extensions/ql-vscode/package.json#L15-L17
  - Will require small update to https://codeql.github.com/docs/codeql-for-visual-studio-code/customizing-settings:
    > To override the default behavior and use a different CLI, you can specify the CodeQL CLI Executable Path. ~Note that this is only available as a user setting, not as a workspace setting.~ 
  - Docs PR opened here: https://github.com/github/codeql/pull/6072

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Display a helpful message and link to documentation when a query produces no results. [#866](https://github.com/github/vscode-codeql/pull/866)
 - Refresh test databases automatically after a test run. [#868](https://github.com/github/vscode-codeql/pull/868)
 - Allow users to specify a custom directory for storing query server logs (`codeQL.runningQueries.customLogDirectory`). The extension will not delete these logs automatically. [#863](https://github.com/github/vscode-codeql/pull/863)
+- Support the "workspace trust" feature. This extension is enabled in untrusted workspaces, but it restricts commands that contain arbitrary paths. [#861](https://github.com/github/vscode-codeql/pull/861)
 
 ## 1.4.8 - 05 May 2021
 
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -6,7 +6,8 @@
 - Display a helpful message and link to documentation when a query produces no results. [#866](https://github.com/github/vscode-codeql/pull/866)
 - Refresh test databases automatically after a test run. [#868](https://github.com/github/vscode-codeql/pull/868)
 - Allow users to specify a custom directory for storing query server logs (`codeQL.runningQueries.customLogDirectory`). The extension will not delete these logs automatically. [#863](https://github.com/github/vscode-codeql/pull/863)
-- Support the "workspace trust" feature. This extension is enabled in untrusted workspaces, but it restricts commands that contain arbitrary paths. [#861](https://github.com/github/vscode-codeql/pull/861)
+- Support the VS Code [Workspace Trust feature](https://code.visualstudio.com/docs/editor/workspace-trust). This extension is now enabled in untrusted workspaces, but it restricts commands that contain arbitrary paths. [#861](https://github.com/github/vscode-codeql/pull/861)
+- Allow the `codeQL.cli.executablePath` configuration setting to be set in workspace-scoped configuration files. This means that each workspace can now specify its own CodeQL CLI compiler, a feature that is unblocked due to implementing Workspace Trust. [#861](https://github.com/github/vscode-codeql/pull/861)
 
 ## 1.4.8 - 05 May 2021
 
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -24,7 +24,7 @@
   "capabilities": {
     "untrustedWorkspaces": {
       "supported": "limited",
-      "description": "Workspace trust is required to execute commands that can contain arbitrary paths. ",
+      "description": "Workspace trust is required to execute commands that can contain arbitrary paths.",
       "restrictedConfigurations": [
         "codeQL.cli.executablePath",
         "codeQL.runningTests.additionalTestArguments"
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -21,6 +21,16 @@
   "extensionDependencies": [
     "hbenl.vscode-test-explorer"
   ],
+  "capabilities": {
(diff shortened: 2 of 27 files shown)


aeisenberg:
This looks great! Two extra things:

1. This PR should also address #509 because users can now specify per-workspace cli versions.
2. We should not merge this until workspace trust is enabled by default in vscode. I'm not sure which version that will be, but I think it's happening soon.
3. We need to update the minimum vscode version required to run this extension when the appropriate vscode is released. See: 
https://github.com/github/vscode-codeql/blob/72d57eec6ecfb6ff10d59a64d8403acb68ef05b5/extensions/ql-vscode/package.json#L15-L17

shati-patel:
> The upstream PR has been merged and the new version of test explorer has been released. Let's ship this!

We should probably still wait until until workspace trust is enabled by default in VS Code, since we've changed the CLI executable setting [back to "window" scope](https://github.com/github/vscode-codeql/pull/861/files#diff-be4c5a31505df7de81fa4900e84e66c04d1a6bca9b3da50f8061d2cb2bfbbde1R131). (i.e. anyone who hasn't explicitly set `"security.workspace.trust.enabled": true` doesn't yet get any of the protection offered by workspace trust.)

Points 2 and 3 in your previous comment (https://github.com/github/vscode-codeql/pull/861#issuecomment-839962213) still hold, I think!

adityasharad:
Nice! Do we know what version of VS Code will ship with workspace trust?

shati-patel:
> Nice! Do we know what version of VS Code will ship with workspace trust?

I've been keeping an eye on https://github.com/microsoft/vscode/issues/120251, but haven't seen anything about an expected release. (Maybe I'm looking in the wrong places though...)

I suppose we could revert the "machine -> window" change and ship the workspace trust implementation itself at any time? That would at least make the extension work for users who _have_ enabled workspace trust and are in an untrusted workspace, without changing the behaviour for anyone else.

aeisenberg:
> We should probably still wait until until workspace trust is enabled by default in VS Code

Yes. Good point. 

There is a block on the package.json and we should update this to the version where workspace trust is going to be released. This will ensure that users on older versions on vscode won't accidentally get the new configuration behaviour.

```json
"engines": {
    "vscode": "^1.43.0"
},
```

alexet:
Running qtests in untrusted workspaces is also unsafe at the moment and hence also needs to be disabled.

shati-patel:
I've added some more commits to update dependencies, since workspace trust is now fully enabled (v1.57.0). I updated to 1.48.0  separately first, using commits from https://github.com/github/vscode-codeql/pull/874 (which will hopefully minimise conflicts with the `qc-development` branch), so that commit (https://github.com/github/vscode-codeql/pull/861/commits/ec4785ea7e6ef11b6baff7c60135097614591784) shouldn't need a re-review.




------
Re:
> Running qtests in untrusted workspaces is also unsafe at the moment and hence also needs to be disabled.

I think that's what line 30 covers: https://github.com/github/vscode-codeql/blob/85a1f6663bd1c5ac382ad64a2c4f73301f0d55df/extensions/ql-vscode/package.json#L28-L31

(Unless we need to disable qltests completely? I'm not sure how to do that.)

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Display a helpful message and link to documentation when a query produces no results. [#866](https://github.com/github/vscode-codeql/pull/866)
 - Refresh test databases automatically after a test run. [#868](https://github.com/github/vscode-codeql/pull/868)
 - Allow users to specify a custom directory for storing query server logs (`codeQL.runningQueries.customLogDirectory`). The extension will not delete these logs automatically. [#863](https://github.com/github/vscode-codeql/pull/863)
+- Support the "workspace trust" feature. This extension is enabled in untrusted workspaces, but it restricts commands that contain arbitrary paths. [#861](https://github.com/github/vscode-codeql/pull/861)
```
aeisenberg, line 9:
```suggestion
- Support the VS Code [Workspace Trust feature](https://code.visualstudio.com/docs/editor/workspace-trust). This extension is now enabled in untrusted workspaces, but it restricts commands that contain arbitrary paths. [#861](https://github.com/github/vscode-codeql/pull/861)
- Allow the `codeQL.cli.executablePath` configuration setting to be set in worskpace-scoped configuration files.  This means that each workspace can now specify its own CLI compiler, a feature that is unblocked due to implementing Workspace Trust. [#861](https://github.com/github/vscode-codeql/pull/861)
```

extensions/ql-vscode/package.json
```diff
@@ -13,14 +13,24 @@
...
   "categories": [
     "Programming Languages"
   ],
   "extensionDependencies": [
     "hbenl.vscode-test-explorer"
   ],
+  "capabilities": {
+    "untrustedWorkspaces": {
+      "supported": "limited",
+      "description": "Workspace trust is required to execute commands that can contain arbitrary paths. ",
```
aeisenberg, line 27:
```suggestion
      "description": "Workspace trust is required to execute commands that can contain arbitrary paths.",
```

aeisenberg (commented):
Looks good. Just some updates for the changelog.

extensions/ql-vscode/CHANGELOG.md
shati-patel: Thanks for adding the extra context!

aeisenberg (approved):

Issue #820, opened 2021-04-06
User-specified directory in which to store evaluation logs


tausbn:
**Is your feature request related to a problem? Please describe.**
In the course of my work, I end up creating a lot of evaluation logs that I subsequently scrutinise in order to do things like fix bugs and improve performance. I also occasionally have to restart VSCode (by choice or by necessity), and the fact that logs do not persist across restarts of VSCode is problematic in this regard.

**Describe the solution you'd like**
I would like the possibility of specifying a place where VSCode should place the evaluation logs. VSCode should _never_ attempt to remove old logs from this directory automatically. The onus will be on the user to do this cleanup themselves (though one _could_ add a way to do this in the VSCode extension as well, for instance as a button associated with the "query history" view).

**Describe alternatives you've considered**
My current alternatives include
- Eagerly backing up all logs before restarting VSCode (but I sometimes forget, leading to much loss of work and subsequent cursing)
- Running the `codeql` CLI command manually, to ensure that I can control where the log ends up (but this adds friction, as it takes me out of the nice VSCode extension).

Also, eagerly backing up logs isn't always possible: on a Codespaces instance, I have no way (that I know of) of accessing the instance _before_ it starts up VSCode, and so my logs will always get purged before I can get to them and safely back them up.

**Additional context**
N/A

adityasharad:
Proposed breakdown:
- add a new user setting to specify where to store query logs (if blank, we use the default of the workspace storage folder)
- if the setting is provided, use it in the output channel loggers as their additional logging location
- if the setting is provided, don't automatically clean up old logs when each logger object is constructed

PR #863, opened 2021-05-14
Specify custom directory for storing evaluation logs

shati-patel:
Will fix https://github.com/github/vscode-codeql/issues/820.

## Checklist

  - Docs-wise, we should probably mention this in https://codeql.github.com/docs/codeql-for-visual-studio-code/troubleshooting-codeql-for-visual-studio-code/.

### DIFF
--- a/extensions/ql-vscode/src/logging.ts
+++ b/extensions/ql-vscode/src/logging.ts
@@ -40,18 +40,24 @@ export class OutputChannelLogger extends DisposableObject implements Logger {
   public readonly outputChannel: OutputChannel;
   private readonly additionalLocations = new Map<string, AdditionalLogLocation>();
   private additionalLogLocationPath: string | undefined;
+  isCustomLogDirectory: boolean;
 
   constructor(private title: string) {
     super();
     this.outputChannel = Window.createOutputChannel(title);
     this.push(this.outputChannel);
+    this.isCustomLogDirectory = false;
   }
 
-  init(storagePath: string): void {
+  setLogStoragePath(storagePath: string, isCustomLogDirectory: boolean): void {
     this.additionalLogLocationPath = path.join(storagePath, this.title);
 
-    // clear out any old state from previous runs
-    fs.remove(this.additionalLogLocationPath);
+    this.isCustomLogDirectory = isCustomLogDirectory;
+
+    if (!this.isCustomLogDirectory) {
+      // clear out any old state from previous runs
+      fs.remove(this.additionalLogLocationPath);
+    }
   }
 
   /**
@@ -80,7 +86,7 @@ export class OutputChannelLogger extends DisposableObject implements Logger {
         this.outputChannel.appendLine(separator);
         this.outputChannel.appendLine(msg);
         this.outputChannel.appendLine(separator);
-        additional = new AdditionalLogLocation(logPath);
+        additional = new AdditionalLogLocation(logPath, !this.isCustomLogDirectory);
         this.additionalLocations.set(logPath, additional);
         this.track(additional);
       }
@@ -112,7 +118,7 @@ export class OutputChannelLogger extends DisposableObject implements Logger {
 }
 
 class AdditionalLogLocation extends Disposable {
-  constructor(private location: string) {
+  constructor(private location: string, private shouldDeleteLogs: boolean) {
     super(() => { /**/ });
   }
 
@@ -128,7 +134,9 @@ class AdditionalLogLocation extends Disposable {
   }
 
   async dispose(): Promise<void> {
-    await fs.remove(this.location);
+    if (this.shouldDeleteLogs) {
+      await fs.remove(this.location);
+    }
   }
 }
 
--- a/extensions/ql-vscode/src/logging.ts
+++ b/extensions/ql-vscode/src/logging.ts
@@ -28,9 +28,16 @@ export interface Logger {
   removeAdditionalLogLocation(location: string | undefined): void;
 
   /**
-   * The base location location where all side log files are stored.
+   * The base location where all side log files are stored.
    */
   getBaseLocation(): string | undefined;
+
+  /**
+   * Sets the location where logs are stored.
+   * @param storagePath The path where logs are stored.
+   * @param isCustomLogDirectory Whether the logs are stored in a custom, user-specified directory.
+   */
+  setLogStoragePath(storagePath: string, isCustomLogDirectory: boolean): Promise<void>;
 }
 
 export type ProgressReporter = Progress<{ message: string }>;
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -84,10 +87,27 @@ export class QueryServerClient extends DisposableObject {
...
+  initLogger() {
+    let storagePath = this.opts.contextStoragePath;
+    let isCustomLogDirectory = false;
+    if (this.config.customLogDirectory) {
+      if (fs.existsSync(this.config.customLogDirectory) && fs.statSync(this.config.customLogDirectory).isDirectory()) {
+        storagePath = this.config.customLogDirectory;
+        isCustomLogDirectory = true;
+      } else if (this.config.customLogDirectory) {
+        helpers.showAndLogErrorMessage(`${this.config.customLogDirectory} is not a valid directory. Logs will be stored in a temporary workspace directory instead.`);
+      }
```
aeisenberg, line 102: If the custom path doesn't exist, maybe you can create it (with an info message). I think it's only an error if it already exists, but isn't a file.

adityasharad:
Looks good so far. Next step is to look at the tests for logging: you'll need to update the tests that use loggers to pass in the new boolean flags you introduced, and check their behaviour in each case.

extensions/ql-vscode/src/queryserver-client.ts
shati-patel: I made an attempt in https://github.com/github/vscode-codeql/pull/863/commits/708baaf1d8ef3f2e82e87dc5e961e34c4596139a, which _seems_ to do the right thing, but any tips on syntax/fixes are very welcome 😄

extensions/ql-vscode/test/pure-tests/logging.test.ts
```diff
@@ -94,11 +109,36 @@ describe('OutputChannelLogger tests', () => {
...
+  it('should delete an existing folder when setting the log storage path', async () => {
     fs.createFileSync(path.join(tempFolders.storagePath.name, 'test-logger', 'xxx'));
-    logger.init(tempFolders.storagePath.name);
+    logger.setLogStoragePath(tempFolders.storagePath.name, false);
     // should be empty dir
 
+    const testLoggerFolder = path.join(tempFolders.storagePath.name, 'test-logger');
+    // TODO: Why does this test pass? I'd expect the length to be 0 if it's correctly deleted the existing folder.
+    expect(fs.readdirSync(testLoggerFolder).length).to.equal(1);
+  });
```
shati-patel, line 135: I couldn't work out why this test passes with "length equal to 1". Can someone explain? I might be misunderstanding what it's doing... 👀

aeisenberg: Is it possible `fs.remove` is not working because the directory is not empty? Try setting a breakpoint on line 129 and step through to see what happens and what is deleted.

aeisenberg: Also, I wonder if that method isn't completely synchronous. You can try addinga a `waitABit` call in between and see what happens.

extensions/ql-vscode/src/logging.ts
```diff
@@ -40,18 +40,24 @@ export class OutputChannelLogger extends DisposableObject implements Logger {
...
+  setLogStoragePath(storagePath: string, isCustomLogDirectory: boolean): void {
     this.additionalLogLocationPath = path.join(storagePath, this.title);
 
-    // clear out any old state from previous runs
-    fs.remove(this.additionalLogLocationPath);
+    this.isCustomLogDirectory = isCustomLogDirectory;
+
+    if (!this.isCustomLogDirectory) {
+      // clear out any old state from previous runs
+      fs.remove(this.additionalLogLocationPath);
```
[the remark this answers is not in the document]
aeisenberg: Yikes...and I realize now that this wasn't async in the past. This change could happen in a future PR.

extensions/ql-vscode/src/logging.ts
```diff
@@ -40,18 +40,24 @@ export class OutputChannelLogger extends DisposableObject implements Logger {
...
 
-  init(storagePath: string): void {
+  setLogStoragePath(storagePath: string, isCustomLogDirectory: boolean): void {
     this.additionalLogLocationPath = path.join(storagePath, this.title);
 
-    // clear out any old state from previous runs
-    fs.remove(this.additionalLogLocationPath);
+    this.isCustomLogDirectory = isCustomLogDirectory;
+
+    if (!this.isCustomLogDirectory) {
```
aeisenberg, line 57: So, what this means is whenever someone changes the query server logs from the default to a custom location, then the old logs are deleted. And if they switch back, there won't be any logs at all. Maybe that's ok.

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -151,6 +172,7 @@ export class QueryServerClient extends DisposableObject {
       args.push('-J=-agentlib:jdwp=transport=dt_socket,address=localhost:9010,server=y,suspend=n,quiet=y');
     }
 
+    this.initLogger();
```
aeisenberg, line 175: I think this needs to go before the call to `resolveRam`.

extensions/ql-vscode/test/pure-tests/logging.test.ts
adityasharad: Oh the sync/async might be it. I was also baffled by why the existing test worked.

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -86,6 +89,26 @@ export class QueryServerClient extends DisposableObject {
...
   }
 
+  async initLogger() {
+    let storagePath = this.opts.contextStoragePath;
+    let isCustomLogDirectory = false;
+    if (this.config.customLogDirectory) {
+      try {
+        if (!fs.pathExists(this.config.customLogDirectory)) {
+          await fs.mkdir(this.config.customLogDirectory);
+        }
```
shati-patel, line 99: Just realised that this logic doesn't quite work 🙃 (When I enter an invalid directory, it doesn't still doesn't `catch` any errors.)

shati-patel: Latest commit (https://github.com/github/vscode-codeql/pull/863/commits/8d711c668e5781ec91dfa96f056c7d91e68a8b17) fixes the logic, but with slightly annoying duplication... I haven't found an easy way to tidy this up yet 😞

aeisenberg: I think `if (!fs.pathExists(this.config.customLogDirectory)) {` should be `if (!(await fs.pathExists(this.config.customLogDirectory))) {` Sorry, I didn't catch that earlier.

aeisenberg: Would that fix the problem here?

shati-patel: That seems to work 🎉  Thank you!

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -127,6 +150,7 @@ export class QueryServerClient extends DisposableObject {
 
   /** Starts a new query server process, sending progress messages to the given reporter. */
   private async startQueryServerImpl(progressReporter: ProgressReporter): Promise<void> {
+    this.initLogger();
```
aeisenberg, line 153:
```suggestion
    async this.initLogger();
```

aeisenberg (commented):
One last change and then :shipit:.

aeisenberg (approved):
Great job!

Issue #768, opened 2021-03-02
Better handling of *.testproj directories as databases


aeisenberg:
`*.testproj` databases are usually ephemeral. They are re-created by the test runner each time tests are run. This causes a problem when importing into vscode. When a database is changed on disk, the query server lock is not released. So, when the database is recreated, queries cannot be run since the original lock is invalid.

There are a few workarounds that users can do:

1. Before importing a `testproj` database, make sure to copy it, so it won't be deleted by the test runner.
2. Before rerunning tests, remove the database from the UI. 

Neither of these solutions are ideal and the extension should handle this automatically. An ideal solution is that immediately before a test is run, check to see if the resulting database is already imported. If so, remove it and then run the tests and re-add when complete.

adityasharad:
I like the idea of removing and deregistering the database when a test is run that would recreate it. Bonus points if we can readd the database again after the tests are done! Worth trying and seeing how our internal users find the experience.

cklin:
I am unable to reproduce the issue as described.  The steps I followed:

1. Run failing test from vscode
2. Import testproj directory as database
3. Change test
4. Rerun still-failing test from vscode
5. Run query from vscode against database

I am able to successfully run the query in step 5.  Interestingly, the result of the query at step 5 reflects the state of the testproj directory at step 2 instead of at step 4.

I discussed with `@aeisenberg`, and we concluded that the behavior I observed is still worth fixing.  Partly to make sure that the database is always up-to-date, and partly to deal with situations where the testproj directory is deleted by a successful test run.

cklin:
I am also exploring an alternative solution that refreshes the database prior to each query.  The benefit of doing so is that it solves the testproj staleness problem not only for tests initiated from the vscode extension, but also for tests initiated elsewhere.  We can limit the automatic refresh to only those situations where the testproj directory has a new modification timestamp, so there should be very little wasted effort if the testproj directory is unchanged.

The same automatic refresh path can also handle situations where the testproj directory is removed by a successful test run.

I am currently in a deep dive trying to understand exactly what needs to be refreshed.  If I unregister a testproj database and then reregister the same directory as database, new queries reflect the new state of the testproj directory.  If I simply click the "Set Current Database" button for the testproj database, new queries would reflect the old (outdated) state of the testproj directory.  Even though clicking "Set Current Database" calls refresh() on the DatabaseItem.  So there is a difference between deregistering-reregistering and DatabaseItem.refresh(), and I want to find out what the difference is.

[...]

PR #864, opened 2021-05-14
Prevent CodeQL tests from interfering with loaded testproj databases

cklin:
This PR is work-in-progress for #768, and I would like to get some early feedback on how any aspect of this PR can be improved.

Here is the general plan:

- [x] Always copy database directories to extension-controlled location before loading
- [ ] Re-copy and re-register the database after each test run

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases.ts
+++ b/extensions/ql-vscode/src/databases.ts
@@ -56,7 +56,10 @@ export interface FullDatabaseOptions extends DatabaseOptions {
 }
 
 interface PersistedDatabaseItem {
+  /** The publicly-visible database URI, which may not be in an extension-controlled location. */
   uri: string;
+  /** The extension-managed database URI, which must be in an extension-controlled location. */
+  managedUri?: string;
   options?: DatabaseOptions;
 }
 
@@ -203,6 +206,8 @@ async function resolveDatabaseContents(uri: vscode.Uri): Promise<DatabaseContent
 export interface DatabaseItem {
   /** The URI of the database */
   readonly databaseUri: vscode.Uri;
+  /** The URI of the database in an extension-controlled location */
+  readonly managedDatabaseUri: vscode.Uri;
   /** The name of the database to be displayed in the UI */
   name: string;
 
@@ -298,6 +303,7 @@ export class DatabaseItemImpl implements DatabaseItem {
 
   public constructor(
     public readonly databaseUri: vscode.Uri,
+    public readonly managedDatabaseUri: vscode.Uri,
     contents: DatabaseContents | undefined,
     private options: FullDatabaseOptions,
     private readonly onChanged: (event: DatabaseChangedEvent) => void
@@ -409,6 +415,7 @@ export class DatabaseItemImpl implements DatabaseItem {
   public getPersistedState(): PersistedDatabaseItem {
     return {
       uri: this.databaseUri.toString(true),
+      managedUri: this.managedDatabaseUri.toString(true),
       options: this.options
     };
   }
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/databases.ts
```diff
@@ -525,7 +532,15 @@ export class DatabaseManager extends DisposableObject {
...
-    const contents = await resolveDatabaseContents(uri);
+    let contents = await resolveDatabaseContents(uri);
+    const managedUri = await this.getDatabaseInManagedLocation(uri);
+    if (managedUri !== uri) {
+      // We always want to run resolveDatabaseContents() before doing any substantial work so that
+      // we can report obvious problems to the user quickly. If we end up copying the database
+      // directory to a new location, then we need to run resolveDatabaseContents() again to get
+      // the correct contents for the copied database directory.
+      contents = await resolveDatabaseContents(managedUri);
+    }
```
aeisenberg, line 543: Is there a reason to resolve database contents twice?

```suggestion
    let contents;
    const managedUri = await this.getDatabaseInManagedLocation(uri);
    if (managedUri !== uri) {
      // We always want to run resolveDatabaseContents() before doing any substantial work so that
      // we can report obvious problems to the user quickly. If we end up copying the database
      // directory to a new location, then we need to run resolveDatabaseContents() again to get
      // the correct contents for the copied database directory.
      contents = await resolveDatabaseContents(managedUri);
    } else {
      contents = await resolveDatabaseContents(uri);
    }
```

extensions/ql-vscode/src/databases.ts
```diff
@@ -545,6 +560,24 @@ export class DatabaseManager extends DisposableObject {
...
+   * @param uri current location of the database directory, which must be a file path pointing to a directory
+   * @returns possibly new location of the database directory, which is guaranteed to be in an extension-controlled location
+   */
+  private async getDatabaseInManagedLocation(uri: vscode.Uri): Promise<vscode.Uri> {
+    if (this.isExtensionControlledLocation(uri)) {
+      return uri;
+    }
+    const baseStoragePath = this.ctx.storagePath || this.ctx.globalStoragePath;
+    const storageFolder = await getStorageFolder(baseStoragePath, uri.toString(true));
+    await fs.copy(uri.fsPath, storageFolder);
```
aeisenberg, line 577: Is `copy` recursive?

extensions/ql-vscode/src/databases.ts
```diff
@@ -545,6 +560,24 @@ export class DatabaseManager extends DisposableObject {
...
 
+  /**
+   * Make a database directory available in an extension-controlled location. If the given URI is already
+   * in a controlled location, this method does nothing and returns the argement unchanged. Otherwise, it
+   * copies the given directory to an extension-controlled location and returns the new location.
+   *
+   * @param uri current location of the database directory, which must be a file path pointing to a directory
+   * @returns possibly new location of the database directory, which is guaranteed to be in an extension-controlled location
+   */
+  private async getDatabaseInManagedLocation(uri: vscode.Uri): Promise<vscode.Uri> {
```
aeisenberg, line 571: I prefer avoiding `get*` named methods for things that do side effects. How about something like this:

```suggestion
  private async ensureDatabaseInManagedLocation(uri: vscode.Uri): Promise<vscode.Uri> {
```

aeisenberg (commented):
This is a big change to always copy databases and it might have negative consequences. For example, some databases are large (> 2GB) and we want to avoid making copies if we can.

As an alternative approach, could you only copy testproj databases?

cklin:
I am abandoning this PR due to performance concerns over database copying.

Issue #551, opened 2020-09-09
Add link to documentation when a query returns no results


aeisenberg:
New users are often confused why a query is returning no results. In practice, though, there are often a small number of well-known reasons why this happens, and they are all documented.

When a query results page opens and it has no results, we should add a link to documentation that possibly explains why this happened and what to do next.

eyskim:
Hi @aeisenberg, I would like to work on this issue if nobody has picked it up yet.

eyskim:
Hi @aeisenberg, upon investigation of this issue, I believe an appropriate approach would be to render a message with the link to the appropriate documentation if the the number of `resultSet.rows` passed to the `<RawTable />` component in `src/view/raw-results-table.tsx` is 0. 

Then, instead of rendering the list of query results as multiple `<RawTableRow />` components, it would render a `<tr>` element with the empty row message.

For example, a change like this:
![image](https://user-images.githubusercontent.com/42628024/118430302-a978b600-b688-11eb-814a-485da5d08392.png)

Such that if there are no results returned from the query, the `emptyTableRow` element would be rendered in `<RawTable />`'s `<tbody>` instead of the list of `<RawTableRow />`.

Before I can make a PR for this issue, I wanted to ask how specific the link to the documentation should be to the user's query. In other words, should it link to specific parts of the documentation based on the query? Or is it sufficient to point to a general link for all empty query results (for example to https://codeql.github.com/docs/codeql-language-guides/)? Perhaps I can do both in separate PRs.

Thanks, and please let me know if there are other considerations I missed!

aeisenberg:
I think that's a fine approach. For now, you can keep the link generic. https://codeql.github.com/docs/codeql-language-guides/ is good. And we can determine a better link later, possibly we can link to the specific language guide.

PR #866, opened 2021-05-17
Add link to language guides for empty query results

eyskim:
Empty table when no query results are returned is replaced by a message with documentation to the [CodeQL Language Guides](https://codeql.github.com/docs/codeql-language-guides/).

Fixes https://github.com/github/vscode-codeql/issues/551.

Please let me know if this warrants a change to the changelogs and if I should cc `@github/docs-content-codeql`.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/raw-results-table.tsx
+++ b/extensions/ql-vscode/src/view/raw-results-table.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { ResultTableProps, className } from './result-table-utils';
+import { ResultTableProps, className, emptyQueryResultsMessage } from './result-table-utils';
 import { RAW_RESULTS_LIMIT, RawResultsSortState } from '../pure/interface-types';
 import { RawTableResultSet } from '../pure/interface-types';
 import RawTableHeader from './RawTableHeader';
@@ -21,6 +21,10 @@ export class RawTable extends React.Component<RawTableProps, {}> {
     const { resultSet, databaseUri } = this.props;
 
     let dataRows = resultSet.rows;
+    if (dataRows.length === 0) {
+      return emptyQueryResultsMessage();
+    }
+
     let numTruncatedResults = 0;
     if (dataRows.length > RAW_RESULTS_LIMIT) {
       numTruncatedResults = dataRows.length - RAW_RESULTS_LIMIT;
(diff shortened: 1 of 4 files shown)


aeisenberg:
Thanks for the PR!

extensions/ql-vscode/src/view/raw-results-table.tsx
```diff
@@ -27,14 +27,19 @@ export class RawTable extends React.Component<RawTableProps, {}> {
...
 
-    const tableRows = dataRows.map((row: ResultRow, rowIndex: number) =>
-      <RawTableRow
-        key={rowIndex}
-        rowIndex={rowIndex + this.props.offset}
-        row={row}
-        databaseUri={databaseUri}
-      />
-    );
+    const emptyTableRow = <tr key={0}><td key={-1}>
```
aeisenberg, line 30: Can you extract this into a function and only call it down below so that we are not generating this virtual DOM unless it's needed?

aeisenberg: Also, this is rendering the message as a row in the table. If there are no results, we shouldn't even display the table. Take a look at `renderNoResults` in `alert-table.tsx`.  This basically does what I'm suggesting, though the message here will also need to be updated similarly.

eyskim: Thanks for the initial review! Just to clarify, should I also change the message returned in `renderNoResults()` in `alert-table.tsx`? For example, replace `No Alerts` in `return <span>No Alerts</span>;` to a similar message as I'm adding in `raw-results-table.tsx`.

aeisenberg: Yes, please.

eyskim: Hi @aeisenberg, as suggested I've extracted the message into a function that is called in `alert-table.tsx` and `raw-results-table.tsx` if there are are no query results.

aeisenberg (commented):
This is looking good. One more thing that needs to happen before merging. Can you add an entry in the `CHANGELOG.md` file? Follow the format of previous entries and add text like this:

> Display a helpful message and link to documentation when a query produces no results.

It would also be nice to style this better, with bigger text and margins. But that's something to work on later.

Once you get the changelog updated, we can merge this.

eyskim:
Thanks! I would love to pick up the backlog that addresses the styling and maybe a dynamic link to the docs depending on the language. Please let me know if it's okay for me to create an issue for this.

aeisenberg (approved):
Great work!

aeisenberg:
> Thanks! I would love to pick up the backlog that addresses the styling and maybe a dynamic link to the docs depending on the language. Please let me know if it's okay for me to create an issue for this.

That would be great. Please create an issue. First, can you look at styling and then we can think of a better link to include. There might be something more useful than going to the language guide.

aeisenberg (approved):

PR #868, opened 2021-05-18
Reregister testproj databases around test runs

cklin:
This PR modifies the extension to reregister testproj databases around test execution.

Closes: #768

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Display CodeQL CLI version being downloaded during an upgrade. [#862](https://github.com/github/vscode-codeql/pull/862)
 - Display a helpful message and link to documentation when a query produces no results. [#866](https://github.com/github/vscode-codeql/pull/866)
+- Refresh test databases automatically after a test run. [#868](https://github.com/github/vscode-codeql/pull/868)
 
 ## 1.4.8 - 05 May 2021
 
--- a/extensions/ql-vscode/src/databases.ts
+++ b/extensions/ql-vscode/src/databases.ts
@@ -266,6 +266,11 @@ export interface DatabaseItem {
   belongsToSourceArchiveExplorerUri(uri: vscode.Uri): boolean;
 
   /**
+   * Whether the database may be affected by test execution for the given path.
+   */
+  isAffectedByTest(testPath: string): Promise<boolean>;
+
+  /**
    * Gets the state of this database, to be persisted in the workspace state.
    */
   getPersistedState(): PersistedDatabaseItem;
@@ -469,6 +474,27 @@ export class DatabaseItemImpl implements DatabaseItem {
       return false;
     return uri.scheme === zipArchiveScheme &&
       decodeSourceArchiveUri(uri).sourceArchiveZipPath === this.sourceArchive.fsPath;
+  }
+
+  public async isAffectedByTest(testPath: string): Promise<boolean> {
+    const databasePath = this.databaseUri.fsPath;
+    if (!databasePath.endsWith('.testproj')) {
+      return false;
+    }
+    try {
+      const stats = await fs.stat(testPath);
+      if (stats.isDirectory()) {
+        return !path.relative(testPath, databasePath).startsWith('..');
+      } else {
+        // database for /one/two/three/test.ql is at /one/two/three/three.testproj
+        const testdir = path.dirname(testPath);
+        const testdirbase = path.basename(testdir);
+        return databasePath == path.join(testdir, testdirbase + '.testproj');
+      }
+    } catch {
+      // No information available for test path - assume database is unaffected.
+      return false;
+    }
   }
 }
 
--- a/extensions/ql-vscode/src/test-adapter.ts
+++ b/extensions/ql-vscode/src/test-adapter.ts
@@ -1,3 +1,4 @@
+import * as fs from 'fs-extra';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import {
@@ -17,8 +18,9 @@ import { QLTestFile, QLTestNode, QLTestDirectory, QLTestDiscovery } from './qlte
 import { Event, EventEmitter, CancellationTokenSource, CancellationToken } from 'vscode';
 import { DisposableObject } from './pure/disposable-object';
 import { CodeQLCliServer } from './cli';
-import { getOnDiskWorkspaceFolders } from './helpers';
+import { getOnDiskWorkspaceFolders, showAndLogErrorMessage, showAndLogWarningMessage } from './helpers';
 import { testLogger } from './logging';
+import { DatabaseItem, DatabaseManager } from './databases';
 
 /**
  * Get the full path of the `.expected` file for the specified QL test.
@@ -57,13 +59,13 @@ function getTestOutputFile(testPath: string, extension: string): string {
  * A factory service that creates `QLTestAdapter` objects for workspace folders on demand.
  */
 export class QLTestAdapterFactory extends DisposableObject {
-  constructor(testHub: TestHub, cliServer: CodeQLCliServer) {
+  constructor(testHub: TestHub, cliServer: CodeQLCliServer, databaseManager: DatabaseManager) {
     super();
 
     // this will register a QLTestAdapter for each WorkspaceFolder
     this.push(new TestAdapterRegistrar(
       testHub,
-      workspaceFolder => new QLTestAdapter(workspaceFolder, cliServer)
+      workspaceFolder => new QLTestAdapter(workspaceFolder, cliServer, databaseManager)
     ));
   }
 }
@@ -91,7 +93,8 @@ export class QLTestAdapter extends DisposableObject implements TestAdapter {
 
   constructor(
     public readonly workspaceFolder: vscode.WorkspaceFolder,
-    private readonly cliServer: CodeQLCliServer
+    private readonly cliServer: CodeQLCliServer,
+    private readonly databaseManager: DatabaseManager
   ) {
     super();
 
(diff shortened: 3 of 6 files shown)


extensions/ql-vscode/src/databases.ts
```diff
@@ -470,6 +476,15 @@ export class DatabaseItemImpl implements DatabaseItem {
     return uri.scheme === zipArchiveScheme &&
       decodeSourceArchiveUri(uri).sourceArchiveZipPath === this.sourceArchive.fsPath;
   }
+
+  /**
+   * Whether the database may be affected by test execution under the given path. It returns true
+   * if the database is backed by a testproj directory at or underneath the given path.
+   */
```
aeisenberg, line 483: It's not useful to duplicate the comment here. Mostly, I'm concerned about one of these copies going out of date.

extensions/ql-vscode/src/test-adapter.ts
```diff
@@ -182,11 +184,33 @@ export class QLTestAdapter extends DisposableObject implements TestAdapter {
...
+        try {
+          const uri = closedDatabase.databaseUri;
+          const reopenedDatabase = await this.databaseManager
+            .openDatabase(_ => { /* no progress reporting */ }, token, uri);
+          await this.databaseManager.renameDatabaseItem(reopenedDatabase, closedDatabase.name);
+          if (currentDatabaseUri === reopenedDatabase.databaseUri) {
+            await this.databaseManager.setCurrentDatabaseItem(reopenedDatabase, true);
+          }
+        } catch (e) {
+          // The database directory may have been deleted by a successful test run.  So failures
```
aeisenberg, line 210: Is there a specific error message we can check here for missing databases? Seems like we only want to catch errors relating to missing databases and rethrow all others.

extensions/ql-vscode/src/test-adapter.ts
```diff
@@ -182,11 +184,33 @@ export class QLTestAdapter extends DisposableObject implements TestAdapter {
     testLogger.outputChannel.show(true);
 
     this.runningTask = this.track(new CancellationTokenSource());
+    const token = this.runningTask.token;
```
aeisenberg, line 187: Maybe instead of creating a `CancellationTokenSource` in the line above, you could wrap the inside of this function with `withProgress` in `commandRunner.ts`. This gives you both a progress monitor and a cancellation token.

You can use them in your calls to `removeDatabaseItem` and `openDatabase` below.

aeisenberg (commented):
I think this generally looks like the right approach. See my comments inline.

extensions/ql-vscode/src/test-adapter.ts
cklin: What do you think we should do if multiple databases fail to open for non-missing errors?  That would lead to multiple exceptions that need to be propagated.  Is there a way to throw multiple exceptions simultaneously in Typescript?  Do we stop at the first exception and throw that?  Or do something else?

aeisenberg: That's a good question. There are two possibilities that I can think of:

1. keep track of all catastrophic errors, concatenate the error messages and throw after trying to re-open all old databases
2. fail fast after the first error, throw that one and don't try to add the remainder.

While 1) may be nicer, I expect that 2) may be sufficient because hopefully this will be a rare occurrence, users are unlikely to have more than one testproj database they are working on at once, and if there is a catastrophic error it's likely that it would affect all of the databases at the same time.

extensions/ql-vscode/src/test-adapter.ts
```diff
@@ -185,12 +188,39 @@ export class QLTestAdapter extends DisposableObject implements TestAdapter {
 
     this._testStates.fire({ type: 'started', tests: tests } as TestRunStartedEvent);
 
+    const currentDatabaseUri = this.databaseManager.currentDatabaseItem?.databaseUri;
+    const databasesUnderTest = this.databaseManager.databaseItems
+      .filter(database => tests.find(testPath => database.isAffectedByTestsUnder(testPath)));
+    await withProgress(
```
aeisenberg, line 194: Darn. I was a little mistaken here and I think your original approach was better.

`this.runningTask` is a `CancellationTokenSource` and `withProgress` gives you a `CancellationToken`. The latter allows you to cancel long running operations and listen for when cancellation happens. The former is for creating and disposing these tokens.

I was hoping we could use a single token and progress monitor for the entire test running operation. This would let users cancel some tests partway through a run.

What you have now creates a first progress monitor, which could be cancelled but does not end the entire operation. And when the tests are running, there is no progress monitor to cancel at all (but the red _stop_ button still works).

There might be a way to get all these operations to work in one operation, but that is a bigger refactor. I think just reverting to what you had before with empty progress monitors is fine.

aeisenberg (commented):
We can chat about my comment tomorrow if you like.

extensions/ql-vscode/src/test-adapter.ts
```diff
@@ -185,12 +188,39 @@ export class QLTestAdapter extends DisposableObject implements TestAdapter {
...
     }
+
+    await withProgress(
+      { location: vscode.ProgressLocation.Notification },
+      async (progress, token) => {
+        for (const closedDatabase of databasesUnderTest) {
+          const uri = closedDatabase.databaseUri;
+          const reopenedDatabase = await this.databaseManager.openDatabase(progress, token, uri);
+          await this.databaseManager.renameDatabaseItem(reopenedDatabase, closedDatabase.name);
+          if (currentDatabaseUri === reopenedDatabase.databaseUri) {
```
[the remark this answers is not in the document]
cklin: Thanks for catching that.  I think in this case using == directly on the URI may be better because currentDatabaseUri could be undefined.

extensions/ql-vscode/src/test-adapter.ts
cklin: Thanks!  That makes sense.  Even with the withProgress() call, I cannot get the progress monitor window to show up at all, so it is not much of a loss anyway.

extensions/ql-vscode/src/databases.ts
```diff
@@ -470,6 +475,25 @@ export class DatabaseItemImpl implements DatabaseItem {
     return uri.scheme === zipArchiveScheme &&
       decodeSourceArchiveUri(uri).sourceArchiveZipPath === this.sourceArchive.fsPath;
   }
+
+  public async isAffectedByTest(testPath: string): Promise<boolean> {
+    const databasePath = this.databaseUri.fsPath;
+    if (!databasePath.endsWith('.testproj')) {
+      return Promise.resolve(false);
```
aeisenberg, line 482: javascript will hoist the result into a promise, so there's no need to explicitly call resolve here.

```suggestion
      return false;
```

Same as below.

extensions/ql-vscode/src/databases.ts
```diff
@@ -470,6 +475,25 @@ export class DatabaseItemImpl implements DatabaseItem {
...
+    }
+    const stats = await fs.stat(testPath);
```
aeisenberg, line 484: Does `stat` return falsy if the path doesn't exist or does it throw?

extensions/ql-vscode/src/test-adapter.ts
```diff
@@ -182,19 +185,79 @@ export class QLTestAdapter extends DisposableObject implements TestAdapter {
...
+    databasesUnderTest: DatabaseItem[], token: vscode.CancellationToken): Promise<void> {
+    for (const database of databasesUnderTest) {
+      try {
+        await this.databaseManager
+          .removeDatabaseItem(_ => { /* no progress reporting */ }, token, database);
+      } catch (e) {
+        // This method is invoked from Test Explorer UI, and testing indicates that Test
+        // Explorer UI swallows any thrown exception without reporting it to the user.
+        // So we need to display the error message ourselves and then rethrow.
+        showAndLogErrorMessage(`Cannot remove database ${database.databaseUri}: ${e}`);
```
aeisenberg, line 221: Database name might be easier for people to understand.
```suggestion
        showAndLogErrorMessage(`Cannot remove database ${database.name}: ${e}`);
```

extensions/ql-vscode/src/vscode-tests/minimal-workspace/databases.test.ts
```diff
@@ -414,6 +414,72 @@ describe('databases', () => {
...
   });
 
+  describe('isAffectedByTest', () => {
+    const directoryStats = new fs.Stats();
+    const fileStats = new fs.Stats();
+
+    before(() => {
+      sinon.stub(directoryStats, 'isDirectory').returns(true);
+      sinon.stub(fileStats, 'isDirectory').returns(false);
+    });
```
aeisenberg, line 424: Nice.

aeisenberg (commented):
This is looking great. A few smaller comments.

extensions/ql-vscode/src/databases.ts
cklin: Good catch!  I ran a test, and it throws if the path does not exist.  I updated the code and the tests accordingly.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Display CodeQL CLI version being downloaded during an upgrade. [#862](https://github.com/github/vscode-codeql/pull/862)
 - Display a helpful message and link to documentation when a query produces no results. [#866](https://github.com/github/vscode-codeql/pull/866)
+- When running tests from the extension, first remove testproj databases affected by the tests and then try to reopen them after the tests complete. [#868](https://github.com/github/vscode-codeql/pull/868)
```
aeisenberg, line 7: Would be better to phrase this as what the feature accomplishes, instead of how it was implemented.

```suggestion
- Test databases are automatically refreshed after a test run. [#868](https://github.com/github/vscode-codeql/pull/868)
```

Or something like that.

aeisenberg (approved):
I'll approve now, assuming you update the changelog message.

PR #874, opened 2021-06-04
Add GitHub authentication

shati-patel:
Adds a command to authenticate to GitHub. This doesn't do anything yet, but will be helpful in future so we can use the GitHub API. When you run the command `CodeQL: Authenticate to GitHub`, you should get a prompt to log in. If you're already logged in, a pop up shows the user you're authenticated as.

The main changes here are `authentication.ts` and `extension.ts` (mostly based on this [VS Code sample](https://github.com/microsoft/vscode-extension-samples/tree/main/github-authentication-sample)). I had to update to VS Code 1.48.0 to use the [authentication](https://code.visualstudio.com/api/references/vscode-api#authentication) API, so there are lots of unrelated changes to make existing code compatible with updated packages.

~Tests are currently broken, presumably because we need to properly implement the `MockExtensionContext` test class.~ EDIT: This was actually due to an outdated version of `eslint`. 

## Checklist

TODO

I'm not sure we want to merge this yet, or wait until we have a use for this authentication 🤔


aeisenberg (commented):
Looks sensible. A few comments.

extensions/ql-vscode/src/authentication.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/** 
+ * Handles authentication to GitHub, using the VS Code [authentication API](https://code.visualstudio.com/api/references/vscode-api#authentication).
+ */
+export class Credentials {
+  private octokit: Octokit.Octokit | undefined;
+
+  async initialize(context: vscode.ExtensionContext): Promise<void> {
+    this.registerListeners(context);
+    this.setOctokit();
```
robertbrignull, line 18: This line needs an `await`.

You may find it useful to add the `"@typescript-eslint/no-floating-promises": "error"` linting rule, though I tried it and there are a lot of cases caused by async logging calls, so probably not something to look at in this PR.

extensions/ql-vscode/src/authentication.ts
```diff
@@ -0,0 +1,60 @@
...
+// https://docs.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps
+const SCOPES = ['repo'];
+
+/** 
+ * Handles authentication to GitHub, using the VS Code [authentication API](https://code.visualstudio.com/api/references/vscode-api#authentication).
+ */
+export class Credentials {
+  private octokit: Octokit.Octokit | undefined;
+
+  async initialize(context: vscode.ExtensionContext): Promise<void> {
```
robertbrignull, line 16: By having a separate initialisation call after construction, it means there two things that the caller has to do, and if a developer doesn't read and understand how to use the `Credentials` class they could make one and forget to initialise it.

Would it be safer to do something like
```typescript
private constructor() {}

static async initialize(context: vscode.ExtensionContext): Promise<Credentials> {
  const c = new Credentials();
  c.registerListeners(context);
  c.setOctokit();
  return c;
}
```

and then to use it just do
```typescript
const credentials = await Credentials.initialize(ctx);
```

Unfortunately you can't do `async` constructors, but I did learn today you can do `static` methods like this so at least it looks a bit better / more understandable.

robertbrignull (commented):
Thought I'd have a read through of this too, so I can learn a bit about how the vscode extension works too.

extensions/ql-vscode/src/extension.ts
```diff
@@ -709,6 +710,19 @@ async function activateWithInstalledDistribution(
...
+   * Currently unused, but will be useful in the future when making API calls.
+   */
+  const credentials = new Credentials();
+  await credentials.initialize(ctx);
+
+  ctx.subscriptions.push(
+    commandRunner('codeQL.authenticateToGitHub', async () => {
+      const octokit = await credentials.getOctokit();
+      const userInfo = await octokit.users.getAuthenticated();
+      helpers.showAndLogInformationMessage(`Authenticated to GitHub as user: ${userInfo.data.login}`);
```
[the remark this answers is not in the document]
shati-patel: VS Code handles login errors quite nicely, so I don't think we need a separate try-catch here:
 
image

extensions/ql-vscode/src/authentication.ts
```diff
@@ -0,0 +1,60 @@
...
+      }
+    }));
+  }
+
+  async getOctokit(): Promise<Octokit.Octokit> {
+    if (this.octokit) {
+      return this.octokit;
+    }
+
+    const session = await vscode.authentication.getSession(GITHUB_AUTH_PROVIDER_ID, SCOPES, { createIfNone: true });
```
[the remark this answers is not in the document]
shati-patel: Since we have `createIfNone: true`, I can't see a way for `session` to be null 🤔

extensions/ql-vscode/src/authentication.ts
```diff
@@ -0,0 +1,65 @@
...
+/** 
+ * Handles authentication to GitHub, using the VS Code [authentication API](https://code.visualstudio.com/api/references/vscode-api#authentication).
+ */
+export class Credentials {
+  private octokit: Octokit.Octokit | undefined;
+
+  // Explicitly make the constructor private, so that we can't accidentally call the constructor from outside the class
+  // without also initializing the class.
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  private constructor() { }
```
shati-patel, line 19: (Not strictly necessary, but it gives a bit of extra safety, as @robertbrignull suggested in https://github.com/github/vscode-codeql/pull/874#discussion_r647269324)

shati-patel:
Still to do: We probably want to feature flag this command, probably using the existing "canary" setting.

extensions/ql-vscode/src/authentication.ts
```diff
@@ -13,33 +13,38 @@ const SCOPES = ['repo'];
...
 
-  private async setOctokit() {
-    // If `createIfNone` were true, a dialog would pop up asking the user to authenticate as soon as the extension starts.
-    // Setting `createIfNone` to false for now, so we can have a more "quiet" prompt, i.e. a numbered label on the "accounts"
-    // icon in the activity bar.
-    const session = await vscode.authentication.getSession(GITHUB_AUTH_PROVIDER_ID, SCOPES, { createIfNone: false });
+  private async initializeOctokit(createIfNone: boolean) {
+    // If `createIfNone` is true, a dialog pops up asking the user to authenticate as soon as the extension starts.
+    // Initializing with `createIfNone: false` for now, so we can have a more quiet prompt, i.e. a numbered label on
+    // the "accounts" icon in the activity bar.
```
aeisenberg, line 31: This comment seems misplaced.

extensions/ql-vscode/src/authentication.ts
```diff
@@ -49,12 +54,12 @@ export class Credentials {
...
-    const session = await vscode.authentication.getSession(GITHUB_AUTH_PROVIDER_ID, SCOPES, { createIfNone: true });
-    this.octokit = new Octokit.Octokit({
-      auth: session.accessToken
-    });
+    await this.initializeOctokit(true);
+    // octokit shouldn't be undefined, since we've set "createIfNone: true".
+    // The following block is mainly here to prevent a compiler error.
+    if (!this.octokit) {
+      throw new Error('Failed to initialize Octokit.');
+    }
```
aeisenberg, line 62: Could `this.octokit` be falsy if someone clicks cancel? In this case, it's not exactly an error if octokit isn't initialized.

Assuming this is the case, I think you could have a better error message.

```suggestion
    // octokit shouldn't be undefined, since we've set "createIfNone: true".
    // The following block is mainly here to prevent a compiler error.
    if (!this.octokit) {
      throw new Error('Did not initialize Octokit.');
    }
```

extensions/ql-vscode/src/authentication.ts
```diff
@@ -13,33 +13,38 @@ const SCOPES = ['repo'];
...
+    await c.initializeOctokit(false);
+    return c;
   }
 
-  private async setOctokit() {
-    // If `createIfNone` were true, a dialog would pop up asking the user to authenticate as soon as the extension starts.
-    // Setting `createIfNone` to false for now, so we can have a more "quiet" prompt, i.e. a numbered label on the "accounts"
-    // icon in the activity bar.
-    const session = await vscode.authentication.getSession(GITHUB_AUTH_PROVIDER_ID, SCOPES, { createIfNone: false });
+  private async initializeOctokit(createIfNone: boolean) {
```
aeisenberg, line 28: Nit: in situations like this, I try to avoid side-effects, so you could have this method return the octokit instance and then set it in all places where you call it. Eg-

```suggestion
  private async createOctokit(createIfNone: boolean):  Octokit.Octokit | undefined {
```

extensions/ql-vscode/src/authentication.ts
```diff
@@ -13,33 +13,38 @@ const SCOPES = ['repo'];
...
 
     if (session) {
       this.octokit = new Octokit.Octokit({
         auth: session.accessToken
       });
-
-      return;
+    } else {
+      this.octokit = undefined;
     }
```
aeisenberg: Here, instead of setting the variable, you return:

```suggestion
    return session
      ? new Octokit.Octokit({
        auth: session.accessToken
      }) : undefined;
```

aeisenberg (commented):
Some stylistic comments that you can take or leave.

extensions/ql-vscode/src/authentication.ts
shati-patel: As far as I can tell, this block is basically unreachable (but I didn't know how to convince the TypeScript compiler that `this.octokit` would never actually be undefined) 🤔 

If a user clicks cancel, then the `await this.createOctokit(true)` promise is rejected and `getOctokit` errors out with `User did not consent to login`.

extensions/ql-vscode/src/authentication.ts
shati-patel: Thanks! I've done this in the latest commit 🙂 (I think. Let me know if I've misunderstood 🙈 )

extensions/ql-vscode/src/authentication.ts
```diff
@@ -21,30 +21,24 @@ export class Credentials {
   static async initialize(context: vscode.ExtensionContext): Promise<Credentials> {
     const c = new Credentials();
     c.registerListeners(context);
-    await c.initializeOctokit(false);
+    await c.createOctokit(false);
```
aeisenberg, line 24: Do you need this:

```suggestion
    c.octokit = await c.createOctokit(false);
```

extensions/ql-vscode/src/authentication.ts
```diff
@@ -21,30 +21,24 @@ export class Credentials {
...
     return c;
   }
 
-  private async initializeOctokit(createIfNone: boolean) {
-    // If `createIfNone` is true, a dialog pops up asking the user to authenticate as soon as the extension starts.
-    // Initializing with `createIfNone: false` for now, so we can have a more quiet prompt, i.e. a numbered label on
-    // the "accounts" icon in the activity bar.
+  private async createOctokit(createIfNone: boolean): Promise<Octokit.Octokit | undefined> {
     const session = await vscode.authentication.getSession(GITHUB_AUTH_PROVIDER_ID, SCOPES, { createIfNone: createIfNone });
```
aeisenberg, line 29: Nit: This can be simplified:

```suggestion
    const session = await vscode.authentication.getSession(GITHUB_AUTH_PROVIDER_ID, SCOPES, { createIfNone });
```

extensions/ql-vscode/src/authentication.ts
```diff
@@ -54,11 +48,11 @@ export class Credentials {
       return this.octokit;
     }
 
-    await this.initializeOctokit(true);
+    this.octokit = await this.createOctokit(true);
     // octokit shouldn't be undefined, since we've set "createIfNone: true".
     // The following block is mainly here to prevent a compiler error.
```
aeisenberg: RIght...I understand now. I think it's best to keep the if-block. We _could_ use a type assertion to force the compiler to ignore the possibly undefined return value: `this.octokit = await this.createOctokit(true)!;` But, I think keeping the if-block and the comment is safer.

aeisenberg (commented):
Comment on line 24 is important, and the one on 32 isn't.

extensions/ql-vscode/src/authentication.ts
shati-patel: Oops, thanks! 🤦🏽

extensions/ql-vscode/src/authentication.ts
aeisenberg: That's a good idea. We can use the `ignoreVoid: true` flag to allow the explicit ignoring of promises. It's a bit of a change to make the switchover, but probably worth it. I'll look at it later today.

aeisenberg:
Trying this out locally. And I realize you need to add the `codeQL.authenticateToGitHub` command to activation `activationEvents`. This will ensure that the extension is loaded if someone tries to run this command.

aeisenberg:
And, is there a way to log out once you log in? There doesn't need to be a command for this, but I would like to be able to test different workflows and make sure that there's no way to break things.

shati-patel:
> And, is there a way to log out once you log in? There doesn't need to be a command for this, but I would like to be able to test different workflows and make sure that there's no way to break things.

I've been using the "Accounts" menu in the sidebar to manage access or sign out: 

image

aeisenberg:
Thanks...

Here's an error path. I'm not sure if there's much you can do about it since the auth flow is a bit error prone in some ways anyway:

1. Run the `codeQL.authenticateToGitHub` command
2. Click _Allow_
3. Navigate to the browser
4. Click _Do not authorize_

Now, you are not authorized and repeatedly trying to run the authorize command does not work.

Again, this might be a bug in the vscode API, but something to explore.

shati-patel:
> Here's an error path. I'm not sure if there's much you can do about it since the auth flow is a bit error prone in some ways anyway:
> 
> 1. Run the `codeQL.authenticateToGitHub` command
> 2. Click _Allow_
> 3. Navigate to the browser
> 4. Click _Do not authorize_
> 
> Now, you are not authorized and repeatedly trying to run the authorize command does not work.

Hmm yes, that is annoying... I've seen the same behaviour with other extensions (for example when signing into Live Share), but it's not ideal in any case 🤔 

I've found a few ways to get out of the non-authorized state (e.g. by clicking "Signing into GitHub..." in the notification bar at the bottom of VS Code, or clicking "Accounts > Sign in to use CodeQL" from the sidebar. So it's probably OK for now, but we might want to add some kind of warning/notification for users when this is eventually ready to go live.

shati-patel:
(PS: Targetting this PR against a new development branch, since we don't want to merge to `main` yet.)

aeisenberg (approved):
Looks great. The remaining auto workflow bugs are outside the scope of this extension.

PR #875, opened 2021-06-09
Update linting rules

aeisenberg:
Add the `@typescript-eslint/no-floating-promises` rule with an allowance
for floating promises if `void` is used.

This increases safety and ensures that we are explicit when we avoid
awaiting a promise. I already caught a few bugish locations.

In general, we don't need to await the results of logging calls.

databases-ui, we were using a deprecated method for removing a
directory. `fs.rmdir` instead of `fs.remove`.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -183,15 +183,15 @@ export class CodeQLCliServer implements Disposable {
   killProcessIfRunning(): void {
     if (this.process) {
       // Tell the Java CLI server process to shut down.
-      this.logger.log('Sending shutdown request');
+      void this.logger.log('Sending shutdown request');
       try {
         this.process.stdin.write(JSON.stringify(['shutdown']), 'utf8');
         this.process.stdin.write(this.nullBuffer);
-        this.logger.log('Sent shutdown request');
+        void this.logger.log('Sent shutdown request');
       } catch (e) {
         // We are probably fine here, the process has already closed stdin.
-        this.logger.log(`Shutdown request failed: process stdin may have already closed. The error was ${e}`);
-        this.logger.log('Stopping the process anyway.');
+        void this.logger.log(`Shutdown request failed: process stdin may have already closed. The error was ${e}`);
+        void this.logger.log('Stopping the process anyway.');
       }
       // Close the stdin and stdout streams.
       // This is important on Windows where the child process may not die cleanly.
@@ -271,7 +271,7 @@ export class CodeQLCliServer implements Disposable {
       // Compute the full args array
       const args = command.concat(LOGGING_FLAGS).concat(commandArgs);
       const argsString = args.join(' ');
-      this.logger.log(`${description} using CodeQL CLI: ${argsString}...`);
+      void this.logger.log(`${description} using CodeQL CLI: ${argsString}...`);
       try {
         await new Promise<void>((resolve, reject) => {
           // Start listening to stdout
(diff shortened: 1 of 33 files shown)


aeisenberg:
Before merging, I need to do some more manual testing.

aeisenberg:
Ran through a bunch of the areas that got touched and everything seems good.

extensions/ql-vscode/src/cli.ts
```diff
@@ -976,7 +976,7 @@ const lineEndings = ['\r\n', '\r', '\n'];
  */
 async function logStream(stream: Readable, logger: Logger): Promise<void> {
   for await (const line of await splitStreamAtSeparators(stream, lineEndings)) {
-    logger.log(line);
+    await logger.log(line);
```
edoardopirovano, line 979: Maybe a quick comment here that we need to `await` this log call unlike most others to avoid mixing up the stream?

extensions/ql-vscode/src/databases.ts
```diff
@@ -858,12 +858,12 @@ export class DatabaseManager extends DisposableObject {
...
 
   private updatePersistedCurrentDatabaseItem(): void {
-    this.ctx.workspaceState.update(CURRENT_DB, this._currentDatabaseItem ?
+    void this.ctx.workspaceState.update(CURRENT_DB, this._currentDatabaseItem ?
       this._currentDatabaseItem.databaseUri.toString(true) : undefined);
   }
 
   private updatePersistedDatabaseList(): void {
-    this.ctx.workspaceState.update(DB_LIST, this._databaseItems.map(item => item.getPersistedState()));
+    void this.ctx.workspaceState.update(DB_LIST, this._databaseItems.map(item => item.getPersistedState()));
```
edoardopirovano, line 866: Perhaps this and `updatePersistedCurrentDatabaseItem` above should be `async` and these should `await` so that callers of them can have the option of waiting for the update to complete if they want to. I guess it doesn't matter if none of the callers _need_ to wait for it, but it kind of seems like this is a case where we _might_ want to wait and should make it clear whether we're waiting or not.

edoardopirovano (commented):
Very nice, just a couple of minor points. Quite scary how many places we had potential bugs that I guess didn't trigger under normal circumstances.

aeisenberg:
Need to look at the integration tests. Put into draft for now.

extensions/ql-vscode/src/databases.ts
aeisenberg: Actually, looking at the calling code, we probably want to await all of these calls. Made that change.

extensions/ql-vscode/src/telemetry.ts
```diff
@@ -209,6 +209,7 @@ export let telemetryListener: TelemetryListener;
 
 export async function initializeTelemetry(extension: Extension<any>, ctx: ExtensionContext): Promise<void> {
   telemetryListener = new TelemetryListener(extension.id, extension.packageJSON.version, key, ctx);
-  telemetryListener.initialize();
+  // do not await initialization, since doing so will cause integration tests to hang.
+  void telemetryListener.initialize();
```
edoardopirovano, line 213: Huh, strange that this is the case. Do you know why the integration tests hang if we wait for the telemetry to initialize? In any case this seems okay to me, since we weren't `await`ing it before so this isn't a behaviour change.

edoardopirovano (approved):

extensions/ql-vscode/src/telemetry.ts
aeisenberg: It's because in some situations, the initialize function can launch a dialog box to ask if a user wants to enable telemetry. The function blocks until the dialog closes. There is no programmatic way (AFAIK) to close a dialog.

edoardopirovano: Aha, makes sense. Could you expand the comment a little bit to explain that to future readers of the code?

PR #701, opened 2020-12-09
Hackathon: Create command to issue multi-queries

aeisenberg:
Need to add your PAT as a setting to codeQL.cli.personalAccessToken.
Make sure to create a new PAT for this and assign `repo` privileges so
that it can dispatch actions.

PR #882, opened 2021-06-16
Workflow dispatch + dev branch tidy-up

shati-patel:
This branch is based on the `qc-development` branch, but (assuming the feature-flagging works) it should now be good to merge to `main` without disrupting our external users.
- First commit is from https://github.com/github/vscode-codeql/pull/874 (already reviewed)
- Second commit adds a basic workflow dispatch command (adapted from hackathon work in https://github.com/github/vscode-codeql/pull/701/files)
- Third commit hides these commands behind the "canary" flag 

More details in internal linked issue and demo!

## Checklist

(n/a—these features aren't useful to external users, so no need to publicize them)

- [N/A] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [N/A] Issues have been created for any UI or other user-facing changes made by this pull request.
- [N/A] `@github/docs-content-codeql` has been cc'd in all issues for UI or other user-facing changes made by this pull request.

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -246,6 +246,10 @@
         "title": "CodeQL: Run Query"
       },
       {
+        "command": "codeQL.runRemoteQuery",
+        "title": "CodeQL: Run Remote Query"
+      },
+      {
         "command": "codeQL.runQueries",
         "title": "CodeQL: Run Queries in Selected Files"
       },
@@ -664,6 +668,10 @@
           "when": "resourceLangId == ql && resourceExtname == .ql"
         },
         {
+          "command": "codeQL.runRemoteQuery",
+          "when": "editorLangId == ql && resourceExtname == .ql"
+        },
+        {
           "command": "codeQL.runQueries",
           "when": "false"
         },
@@ -803,6 +811,10 @@
       "editor/context": [
         {
           "command": "codeQL.runQuery",
+          "when": "editorLangId == ql && resourceExtname == .ql"
+        },
+        {
+          "command": "codeQL.runRemoteQuery",
           "when": "editorLangId == ql && resourceExtname == .ql"
         },
         {
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -38,6 +38,7 @@
     "onView:codeQLAstViewer",
     "onView:test-explorer",
     "onCommand:codeQL.checkForUpdatesToCLI",
+    "onCommand:codeQL.authenticateToGitHub",
     "onCommand:codeQLDatabases.chooseDatabaseFolder",
     "onCommand:codeQLDatabases.chooseDatabaseArchive",
     "onCommand:codeQLDatabases.chooseDatabaseInternet",
@@ -236,6 +237,10 @@
       }
     },
     "commands": [
+      {
+        "command": "codeQL.authenticateToGitHub",
+        "title": "CodeQL: Authenticate to GitHub"
+      },
       {
         "command": "codeQL.runQuery",
         "title": "CodeQL: Run Query"
@@ -868,6 +873,7 @@
     "format-staged": "lint-staged"
   },
   "dependencies": {
+    "@octokit/rest": "^18.5.6",
     "child-process-promise": "^2.2.1",
     "classnames": "~2.2.6",
     "fs-extra": "^9.0.1",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -664,12 +664,16 @@
       ],
       "commandPalette": [
         {
+          "command": "codeQL.authenticateToGitHub",
+          "when": "config.codeQL.canary"
+        },
+        {
           "command": "codeQL.runQuery",
           "when": "resourceLangId == ql && resourceExtname == .ql"
         },
         {
           "command": "codeQL.runRemoteQuery",
-          "when": "editorLangId == ql && resourceExtname == .ql"
+          "when": "config.codeQL.canary && editorLangId == ql && resourceExtname == .ql"
         },
         {
           "command": "codeQL.runQueries",
@@ -815,7 +819,7 @@
         },
         {
           "command": "codeQL.runRemoteQuery",
-          "when": "editorLangId == ql && resourceExtname == .ql"
+          "when": "config.codeQL.canary && editorLangId == ql && resourceExtname == .ql"
         },
         {
           "command": "codeQL.viewAst",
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const octokit = await credentials.getOctokit();
+  const token = await credentials.getToken();
+
+  const queryFile = uri.fsPath;
+  const query = await fs.readFile(queryFile, 'utf8');
+
+  const repositoriesFile = queryFile.substring(0, queryFile.length - '.ql'.length) + '.repositories';
+  if (!(await fs.pathExists(repositoriesFile))) {
+    void showAndLogErrorMessage(`Missing file: '${repositoriesFile}' to specify the repositories to run against.`);
```
aeisenberg, line 26: Maybe add some more context? Something like:

```suggestion
    void showAndLogErrorMessage(`Missing file: '${repositoriesFile}' to specify the repositories to run against. This file must be a sibling of ${queryFile}.`);
```

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -0,0 +1,59 @@
...
+  const config = yaml.safeLoad(await fs.readFile(repositoriesFile, 'utf8')) as Config;
+
+  const ref = config.ref || 'main';
+  const language = config.language;
+  const repositories = JSON.stringify(config.repositories);
+
+  // Test "controller" repository and workflow.
+  const owner = 'dsp-testing';
+  const repo = 'qc-controller';
+  const workflow_id = 'codeql-query.yml';
```
aeisenberg, line 39: Extract these as constants.

extensions/ql-vscode/package.json
```diff
@@ -663,13 +663,17 @@
...
+          "when": "config.codeQL.canary"
+        },
         {
           "command": "codeQL.runQuery",
           "when": "resourceLangId == ql && resourceExtname == .ql"
         },
         {
           "command": "codeQL.runRemoteQuery",
-          "when": "editorLangId == ql && resourceExtname == .ql"
+          "when": "config.codeQL.canary && editorLangId == ql && resourceExtname == .ql"
```
aeisenberg, line 676: Have you verified this is working correctly?

shati-patel: Yep, the commands are hidden in the right situations! Annoyingly, the authentication option itself is still visible from the side bar, even with the canary setting `false`...

image

That comes from the initialization step, which isn't feature-flagged.
https://github.com/github/vscode-codeql/blob/fbec5c920784cdc497b9236135dabfb6b207fb7e/extensions/ql-vscode/src/extension.ts#L729

Will have a look at tidying that up tomorrow! 🧹

aeisenberg: > Yep, the commands are hidden in the right situations! 

Nice!

> Annoyingly, the authentication option itself is still visible from the side bar, even with the canary setting `false`...

Hmmm...is it sufficient to wrap the call to `Credentials.initialize` in a check to see if canary is set?

Also, if someone changes from canary to not canary, it would be nice if they were logged out of github. You can set up a configuration listener and automatically log out if canary goes from true to false (similarly, you can automatically log in when the setting goes in the other direction).

rneatherway:
I've given this a run-through locally and it worked well :heart:. It prompted me to remove the default values for the parameters on the controller workflow as they would be silently used if entries were missing from the `.repositories` file, which was very confusing.

extensions/ql-vscode/package.json
shati-patel: > Also, if someone changes from canary to not canary, it would be nice if they were logged out of github. You can set up a configuration listener and automatically log out if canary goes from true to false (similarly, you can automatically log in when the setting goes in the other direction).

I did some more experimenting, and I don't know if it's possible to log a user out via the public authentication API if we're using a "built-in" [AuthenticationProvider](https://code.visualstudio.com/api/references/vscode-api#AuthenticationProvider) (i.e. GitHub or Microsoft)... I found a few comments where other people were also talking about the [lack of a "logout" option](https://github.com/microsoft/vscode/issues/88309#issuecomment-741569123) and how the [authentication providers are registered](https://github.com/microsoft/vscode/issues/100993#issuecomment-665160371), so I think this might be a wider problem 🤔 

For now, I've at least wrapped the call to `Credentials.initialize` in a canary check, so everything is fully hidden when it needs to be 🎉 Is that okay for now? I don't expect many users toggle the canary flag by accident, so it shouldn't have a noticeable impact!

aeisenberg: Oh well. Thanks for looking into this and apologies for suggesting something that isn't possible.

It's not a big deal since if someone is already logged in and in canary mode, then they are already aware of this feature and will have tried to use it. And it's sufficient to just hide the feature if canary is off. I'll take another look shortly.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -0,0 +1,59 @@
...
+  language: string;
+}
+
+// Test "controller" repository and workflow.
+const OWNER = 'dsp-testing';
+const REPO = 'qc-controller';
+const WORKFLOW_ID = 'codeql-query.yml';
+
+export default async function runRemoteQuery(credentials: Credentials, uri?: Uri) {
+  if (!uri || !uri.fsPath.endsWith('.ql')) {
```
aeisenberg, line 19: nit:

```suggestion
  if (!uri?.fsPath.endsWith('.ql')) {
```

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -0,0 +1,59 @@
...
+      workflow_id: WORKFLOW_ID,
+      ref: ref,
+      inputs: {
+        language,
+        repositories,
+        query,
+        token
+      }
+    });
+    void showAndLogInformationMessage(`Successfully scheduled runs. [Click here to see the progress](https://github.com/${OWNER}/${REPO}/actions).`);
```
aeisenberg, line 54: Neat. This will format markdown in the popup?

Just be aware that the extension log will display this as text.

aeisenberg (approved):
Looks good. One small suggestion, which you can take or leave.

extensions/ql-vscode/src/run-remote-query.ts
shati-patel: > Neat. This will format markdown in the popup?

Yep 😃 

> Just be aware that the extension log will display this as text.

Good point! The link is still clickable in the logs, which is nice 👍🏽

aeisenberg (approved):

PR #883, opened 2021-06-16
[🚧 WIP 🚧] Fix dependabot alerts

shati-patel:
Coming back to this now that https://github.com/github/vscode-codeql/pull/882 is merged 🎉 

See https://github.com/github/vscode-codeql/security/dependabot. I ran `npm audit fix` as a starting point, but this ~might need~ definitely needs some manual work too. 

## Checklist
N/A

### DIFF
--- a/extensions/ql-vscode/package-lock.json
+++ b/extensions/ql-vscode/package-lock.json
@@ -570,14 +570,14 @@
       "dev": true
     },
     "node_modules/@types/gulp": {
-      "version": "4.0.6",
-      "resolved": "https://registry.npmjs.org/@types/gulp/-/gulp-4.0.6.tgz",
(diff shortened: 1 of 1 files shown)


shati-patel:
> Hmmm...taking a look at this now, and I think our best bet is to wait a few more days to fix. The high vulnerability is from `gulp-parent`, which is only used during build, so doesn't affect production.
> 
> Also, there is no fixed version of this available yet, so we need to downgrade to get a safe version. Hopefully, the gulp team will address the problem and put out a new release soon. In that case, it will be easier to fix this dependabot issue.
> 
> If no fix comes in a few weeks, we should downgrade to gulp v3.9.x which is the latest safe version, though might be a bit of a pain.

Thanks for taking a look! ✨  I'll close this PR for now, and hopefully there'll be a patched version of gulp soon 🤞🏽

Issue #890, opened 2021-06-24
Upgrading multiple databases gives duplicate confirmation dialog boxes


shati-patel:
**Describe the bug**

When I select multiple databases for upgrade, the extension pops up an "upgrade confirmation" dialog box for each database. I then get an additional confirmation box for one of the databases that I've already confirmed. 

**Version**

CodeQL extension version: 1.5.1
CodeQL CLI version: 2.5.5 
VS Code version: 1.57.1
Platform: win32 x64

**To reproduce**

1. In the "Databases" UI view, select multiple databases.
2. Right-click and select **Upgrade Database**.
3. Click "Yes" for each upgrade confirmation dialog. (One for each database, as expected.)
4. An additional dialog pops up for one of the databases from step 3.

**Expected behavior**
I only need to confirm upgrades once per database.

**Additional context**

/

shati-patel:
The upgrade command was removed from the database panel in https://github.com/github/vscode-codeql/pull/1501, so this is no longer needed.

PR #891, opened 2021-06-25
New command to add database source folder to workspace

### LINKED ISSUE #798: Support re-adding database source to workspace
Currently when a database is imported, its source folder is automatically added to the workspace.
However, if you for some reason the source folder is not shown anymore (e.g. you manually removed it from workspace), then it appears there is no easy way to add it back.

When right-clicking a database in the list of databases, the extension should check whether the source folder of this database is shown in the current workspace. If not, a context menu "Add database source folder to workspace" (or similar) should be offered.

**Describe alternatives you've considered**
Workaround:
1. Remove the database
2. Import it again

shati-patel:
Will fix #798.

As described in the linked issue, we don't have an easy way of re-adding a database's source folder to the workspace. This PR adds a new command (in the Databases View) that lets you re-add a source folder.

(We also considered re-adding source folders automatically, similar to how the `codeQLDatabases.removeOrphanedDatabases` command is run, but I'm now wondering if that'll be annoying to users. They may have deliberately removed a source folder from their workspace, and don't want it included automatically 🤔 I'm open to suggestions though!)

PS: In the process, I found a small bug in how we upgrade databases. (Opened an issue here #890, but haven't worked out how to fix that yet! 😅)

## Checklist

   - I don't think a docs change is necessary, since the docs don't describe database commands in detail anyway. [We just say](https://codeql.github.com/docs/codeql-for-visual-studio-code/analyzing-your-projects) the following, which is still valid 🙂 
     > To see the menu options for interacting with a database, right-click an entry in the list. You can select multiple databases using Ctrl/Cmd+click.

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Adds the _Add Database Source to Workspace_ command to the right-click context menu in the databases view. This lets users re-add a database's source folder to the workspace and browse the source code. [#891](https://github.com/github/vscode-codeql/pull/891)
+
 ## 1.5.1 - 23 June 2021
 
 No user facing changes.
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -346,6 +346,10 @@
         "title": "Show Database Directory"
       },
       {
+        "command": "codeQLDatabases.addDatabaseSource",
+        "title": "Add Database Source to Workspace"
+      },
+      {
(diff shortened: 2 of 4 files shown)


aeisenberg:
Good point about users explicitly removing the source folder.

aeisenberg (commented):
Nicely done. This should have a changelog entry and then we can merge it.

shati-patel:
> Nicely done. This should have a changelog entry and then we can merge it.

Thanks for the review! I've checked the docs for changes (none needed) and updated the changelog 📝

aeisenberg (approved):
👍🏼

PR #896, opened 2021-07-01
Run tests with nightly CLI

edoardopirovano:
This PR changes the unit tests and integration tests to also download a nightly build of CodeQL and test against this in addition to the stable versions already being tested. This will allow us to catch compatibility issues that arise from changes to the CodeQL CLI _before_ we release them.

## Checklist

N/A: Only affects tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,12 +51,30 @@ jobs:
           name: vscode-codeql-extension
           path: artifacts
 
+  find-nightly:
+    name: Find Nightly Release
+    runs-on: ubuntu-latest
+    outputs:
+      url: ${{ steps.get-url.outputs.nightly-url }}
+    steps:
+      - name: Get Nightly Release URL
+        id: get-url
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+        shell: bash
+        # This workflow step gets an unstable testing version of the CodeQL CLI. It should not be used outside of these tests.
+        run: |
+          LATEST=`gh api repos/dsp-testing/codeql-cli-nightlies/releases --jq '.[].tag_name' --method GET --raw-field 'per_page=1'`
+          echo "::set-output name=nightly-url::https://github.com/dsp-testing/codeql-cli-nightlies/releases/download/$LATEST"
+    
   test:
     name: Test
     runs-on: ${{ matrix.os }}
+    needs: [find-nightly]
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
+        version: [stable, nightly]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -89,7 +107,12 @@ jobs:
       - name: Install CodeQL
         run: |
           mkdir codeql-home
-          curl -L --silent https://github.com/github/codeql-cli-binaries/releases/latest/download/codeql.zip -o codeql-home/codeql.zip
+          if [ ${{ matrix.version }} = "stable" ]
+          then
+            curl -L --silent https://github.com/github/codeql-cli-binaries/releases/latest/download/codeql.zip -o codeql-home/codeql.zip
+          else
+            curl -L --silent ${{ needs.find-nightly.outputs.url }}/codeql.zip -o codeql-home/codeql.zip
+          fi
           unzip -q -o codeql-home/codeql.zip -d codeql-home
           unzip -q -o codeql-home/codeql.zip codeql/codeql.exe -d codeql-home
           rm codeql-home/codeql.zip
@@ -124,12 +147,14 @@ jobs:
   cli-test:
     name: CLI Test
     runs-on: ${{ matrix.os }}
+    needs: [find-nightly]
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        version: ['v2.2.6', 'v2.3.3', 'v2.4.6', 'v2.5.7']
+        version: ['v2.2.6', 'v2.3.3', 'v2.4.6', 'v2.5.7', 'nightly']
     env:
       CLI_VERSION: ${{ matrix.version }}
+      NIGHTLY_URL: ${{ needs.find-nightly.outputs.url }}
       TEST_CODEQL_PATH: '${{ github.workspace }}/codeql'
 
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -123,13 +157,15 @@ jobs:
 
   cli-test:
     name: CLI Test
+    needs: [test]
```
henrymercer, line 160: I'd be in favour of duplicating the "Get Nightly Release URL" step from the `test` job in this job to prevent this job having to wait for the `test` job to complete.

.github/workflows/main.yml
```diff
@@ -95,19 +97,51 @@ jobs:
...
 
-      - name: Run unit tests (Linux)
+      - name: Get Nightly Release URL
+        id: get-url
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+        shell: bash
+        # This workflow step gets an unstable testing version of the CodeQL CLI. It should not be used outside of these tests.
+        run: |
+          export LATEST=`gh release list --repo dsp-testing/codeql-cli-nightlies -L 1 | cut -f 3`
```
henrymercer, line 107: Minor: How confident do we feel about relying on the format of the output here?  If we're not sure that this is stable, we could instead use something like `gh api repos/dsp-testing/codeql-cli-nightlies/releases --jq '.[].tag_name' --method GET --raw-field 'per_page=1'`.

extensions/ql-vscode/src/distribution.ts
```diff
@@ -553,21 +554,28 @@ export class ReleasesApiConsumer {
...
+      if (extractedPath.indexOf(outPath) !== 0 || !(await fs.pathExists(extractedPath))) {
+        return Promise.resolve();
+      }
+      return fs.chmod(extractedPath, file.externalFileAttributes >>> 16);
+    }));
+  } else {
+    await tar.extract({
+      file: archivePath,
+      cwd: outPath,
+    });
```
henrymercer, line 577: Hmm, I'm slightly surprised that we only need to set file permissions when we extract zip archives.

.github/workflows/main.yml
```diff
@@ -95,19 +97,51 @@ jobs:
...
+        shell: bash
+
+      - name: Run unit tests with stable CodeQL (Linux)
         working-directory: extensions/ql-vscode
         if: matrix.os == 'ubuntu-latest'
         run: |
           CODEQL_PATH=$GITHUB_WORKSPACE/codeql-home/codeql/codeql npm run test
 
-      - name: Run unit tests (Windows)
+      - name: Run unit tests with nightly CodeQL (Linux)
```
henrymercer, line 125: We could add the CLI version (stable / nightly) to the job matrix, but given the short duration of these steps, that probably wouldn't be worth the extra compute.

henrymercer (commented):
Have we considered modifying the nightly builds to produce zip archives rather than `.tar.gz` ones?  That would simplify this code and avoid us having to take on an additional dependency.

extensions/ql-vscode/src/distribution.ts
edoardopirovano: My understanding is that `.zip` files don't typically store file permissions so the information is lost after zipping and unzipping (although with the right tooling it is _possible_ to preserve the permissions), whereas `.tar.gz` ones typically do.

.github/workflows/main.yml
edoardopirovano: Good idea, that seems safer.

.github/workflows/main.yml
edoardopirovano: How about just a single job that gets the nightly URL (which should be very quick) and then both the `test` and `cli-test` job can depend on that? That would still let the tests run quickly but avoids duplicating the code and making spurious API calls.

.github/workflows/main.yml
edoardopirovano: Hmm, it's probably still worth doing to make the workflow a bit neater though.

edoardopirovano:
Regarding producing `.zip` files in the nightly releases, it's certainly a possibility. It's worth noting that the Action needs `.tar.gz` files, so we would have to produce _both_ types of files in the release, which seemed a bit clunky and might invite confusion in future. If we really don't want the dependency here I could certainly do that, though.

henrymercer:
I would be slightly in favour of producing `.zip` files in the nightly releases — looking at the code here, the extra clunkiness in the extension seems greater than the additional clunkiness in the nightly release building job.  Interested to hear what others think though (/cc @github/codeql-ci-reviewers @github/codeql-vscode-reviewers).

edoardopirovano:
Note this PR won't go green because the workflows have changed name. Once we've decided on the above and it's approved, it will need force merging following by changing the branch protection rules.

edoardopirovano:
Per discussion above and on Slack, have changed the nightly builds to include a CLI artifact as a `.zip` file and changed the test here to use that 🙂

extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
```diff
@@ -25,11 +25,13 @@ describe('Use cli', function() {
   });
 
   it('should have the correct version of the cli', async () => {
-    expect(
-      (await cli.getVersion()).toString()
-    ).to.eq(
-      new SemVer(process.env.CLI_VERSION || '').toString()
-    );
+    if (process.env.CLI_VERSION !== 'nightly') {
```
henrymercer, line 28: Minor: It might be worth pushing the `if` statement up so we only run this test if we aren't on a nightly CLI version.  Currently if `CLI_VERSION === 'nightly'` this will suggest we've checked that the version of the CLI is correct when we haven't checked anything.

henrymercer (approved):

shati-patel:
👋🏽 @edoardopirovano
Sorry, I've introduced a teeny conflict in https://github.com/github/vscode-codeql/pull/897 🙇🏽‍♀️  (updating the CLI matrix from 2.5.6 -> 2.5.7)

extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
edoardopirovano: Good point, done.

edoardopirovano:
> 👋🏽 @edoardopirovano
> Sorry, I've introduced a teeny conflict in #897 🙇🏽‍♀️ (updating the CLI matrix from 2.5.6 -> 2.5.7)

No worries! I'll rebase 🙂

PR #898, opened 2021-07-02
Add command to run queries on multiple databases

### LINKED ISSUE #569: Add ability to run queries over multiple databases at once
Some enterprise customers have asked for this capability. Even though it's useful, perhaps they should be using code-scanning instead.

shati-patel:
Will fix #569.

First pass at adding a new command "CodeQL: Run Query on Multiple Databases", that lets you select multiple databases from a quick pick menu and run a query over all of them. This currently loops over all selected databases and displays results and query history runs for each database separately. If a database is incompatible, we display an error message and skip to the next database.

This might need some design input if we want to do anything fancy with results display 😅 

<details><summary>Expand for a (somewhat grainy) demo</summary>

  multi-db demo

</details>

**Notes**
- Would this be useful in its current form? (I was mainly experimenting, so suggestions are welcome 🙂)
- The database quick pick currently lists _all_ databases that are in your Databases UI, including those with the wrong language. I couldn't see an obvious way to filter them...

## Checklist

- [x] ~`@github/docs-content-codeql` has been cc'd in all issues for UI or other user-facing changes made by this pull request.~ Opened an internal docs issue!

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -472,7 +472,9 @@ async function activateWithInstalledDistribution(
   ): Promise<void> {
     if (qs !== undefined) {
       const dbItem = databaseQuickPick !== undefined
+        // database selected from multi-database quick pick
         ? databaseQuickPick
+        // database currently selected in Databases UI
         : await databaseUI.getDatabaseItem(progress, token);
       if (dbItem === undefined) {
         throw new Error('Can\'t run query without a selected database');
@@ -587,7 +589,12 @@ async function activateWithInstalledDistribution(
         );
         if (quickpick !== undefined) {
           for (const item of quickpick) {
-            await compileAndRunQuery(false, uri, progress, token, item.databaseItem);
+            try {
+              await compileAndRunQuery(false, uri, progress, token, item.databaseItem);
+            } catch (error) {
+              // Skip databases that are incompatible with the query, e.g. using a different language.
+              void helpers.showAndLogErrorMessage(`Skipped database '${item.label}'. ${error}`);
+            }
           }
         } else {
           void helpers.showAndLogErrorMessage('No databases selected.');
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -10,7 +10,8 @@ import {
   Uri,
   window as Window,
   env,
-  window
+  window,
+  QuickPickItem
 } from 'vscode';
 import { LanguageClient } from 'vscode-languageclient';
 import * as os from 'os';
@@ -559,6 +560,9 @@ async function activateWithInstalledDistribution(
       }
     )
   );
+  interface DatabaseQuickPickItem extends QuickPickItem {
+    databaseItem: DatabaseItem;
+  }
   ctx.subscriptions.push(
     commandRunnerWithProgress(
       'codeQL.runQueryOnMultipleDatabases',
@@ -567,20 +571,23 @@ async function activateWithInstalledDistribution(
         token: CancellationToken,
         uri: Uri | undefined
       ) => {
+        const quickPickItems = dbm.databaseItems.map<DatabaseQuickPickItem>(dbItem => (
+          {
+            databaseItem: dbItem,
+            label: dbItem.name,
+            description: dbItem.language,
+          }
+        ));
         /**
-         * Databases selected from the quick pick menu.
+         * Databases that were selected in the quick pick menu.
          */
-        const quickpick = await window.showQuickPick(
-          // TODO: Return items of type "DatabaseItem" instead of "string"
-          // For this to work, I think we need "DatabaseItem" to extend "QuickPickItem":
-          // https://code.visualstudio.com/api/references/vscode-api#QuickPickItem
-          dbm.databaseItems.map(dbItem => dbItem.name),
-          { canPickMany: true },
+        const quickpick = await window.showQuickPick<DatabaseQuickPickItem>(
+          quickPickItems,
+          { canPickMany: true }
         );
         if (quickpick !== undefined) {
           for (const item of quickpick) {
-            // Instead of "item" (which is a string), we need the associated "DatabaseItem"...
-            await compileAndRunQuery(false, uri, progress, token, item);
+            await compileAndRunQuery(false, uri, progress, token, item.databaseItem);
           }
         } else {
           void helpers.showAndLogErrorMessage('No databases selected.');
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -29,7 +29,7 @@ import {
   QueryServerConfigListener
 } from './config';
 import * as languageSupport from './languageSupport';
-import { DatabaseManager } from './databases';
+import { DatabaseItem, DatabaseManager } from './databases';
 import { DatabaseUI } from './databases-ui';
 import {
   TemplateQueryDefinitionProvider,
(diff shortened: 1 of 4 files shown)


adityasharad:
This looks very neat, and is a nice compact PR! Just had a quick read, will actually test it out later and then do a proper review.

extensions/ql-vscode/src/extension.ts
```diff
@@ -467,9 +468,14 @@ async function activateWithInstalledDistribution(
     selectedQuery: Uri | undefined,
     progress: ProgressCallback,
     token: CancellationToken,
+    databaseQuickPick: DatabaseItem | undefined,
```
adityasharad, line 471: I suggest naming this variable to indicate that it is a database item, not the fact that it comes from a quick pick. The fact that it comes from a quick pick is something that matters only to the calling function. From the perspective of this function: "if you give me a database item I'll use it, if not I'll default to the currently selected database".
```suggestion
    databaseItem: DatabaseItem | undefined,
```

extensions/ql-vscode/src/extension.ts
```diff
@@ -467,9 +468,14 @@ async function activateWithInstalledDistribution(
...
   ): Promise<void> {
     if (qs !== undefined) {
-      const dbItem = await databaseUI.getDatabaseItem(progress, token);
+      const dbItem = databaseQuickPick !== undefined
+        // database selected from multi-database quick pick
+        ? databaseQuickPick
+        // database currently selected in Databases UI
+        : await databaseUI.getDatabaseItem(progress, token);
```
adityasharad, line 478: If you take the above suggestion, this code becomes a little simpler.
```suggestion
     // If no database is specified, use the database currently selected in the Databases UI
     dbItem = dbItem || await databaseUI.getDatabaseItem(progress, token);
```

extensions/ql-vscode/src/extension.ts
```diff
@@ -549,13 +555,57 @@ async function activateWithInstalledDistribution(
...
     )
   );
+  interface DatabaseQuickPickItem extends QuickPickItem {
+    databaseItem: DatabaseItem;
+  }
+  ctx.subscriptions.push(
+    commandRunnerWithProgress(
+      'codeQL.runQueryOnMultipleDatabases',
+      async (
+        progress: ProgressCallback,
```
adityasharad, line 572: Future note: A cool enhancement might be to have the progress bar reflect how many of the selected databases the query has completed on. Dealing with progress monitors is tricky, so let's think about this later, writing it here just to record the idea.

extensions/ql-vscode/src/extension.ts
```diff
@@ -549,13 +555,57 @@ async function activateWithInstalledDistribution(
...
+        token: CancellationToken,
+        uri: Uri | undefined
+      ) => {
+        const quickPickItems = dbm.databaseItems.map<DatabaseQuickPickItem>(dbItem => (
+          {
+            databaseItem: dbItem,
+            label: dbItem.name,
+            description: dbItem.language,
```
adityasharad, line 580: This is a nice choice: it really helps to see the language in the picker.

extensions/ql-vscode/src/extension.ts
```diff
@@ -549,13 +555,57 @@ async function activateWithInstalledDistribution(
...
+          quickPickItems,
+          { canPickMany: true }
+        );
+        if (quickpick !== undefined) {
+          for (const item of quickpick) {
+            try {
+              await compileAndRunQuery(false, uri, progress, token, item.databaseItem);
+            } catch (error) {
+              // Skip databases that are incompatible with the query, e.g. using a different language.
+              void helpers.showAndLogErrorMessage(`Skipped database '${item.label}'. ${error}`);
```
adityasharad, line 596: If multiple databases give an error, this pops up one dialog per database, which can get a bit overwhelming.

I suggest creating a variable outside the loop that holds a list/array of errors (strings). Every time you catch an error, add the error to the list instead of showing the error right away.
Then outside the loop, pop a single dialog that logs all the errors as a single error, with some suitable header message. You can use `.join('\n')` for example to combine the array of strings with newlines in between.

extensions/ql-vscode/src/extension.ts
```diff
@@ -549,13 +555,57 @@ async function activateWithInstalledDistribution(
...
+         * Databases that were selected in the quick pick menu.
+         */
+        const quickpick = await window.showQuickPick<DatabaseQuickPickItem>(
+          quickPickItems,
+          { canPickMany: true }
+        );
+        if (quickpick !== undefined) {
+          for (const item of quickpick) {
+            try {
+              await compileAndRunQuery(false, uri, progress, token, item.databaseItem);
```
adityasharad, line 593: We don't know the language of the query at this point do we? It would be very cool to automatically filter out databases that are of the wrong language. But no problem if we can't do that yet.

adityasharad (commented):
Nice work! Tried it out and this was a good experience to use, and the code is nice and clear. Some suggestions on how we can do error handling - ping me if you'd like more detail.

extensions/ql-vscode/src/extension.ts
adityasharad: As discussed offline, turns out the popup boxes don't support Markdown 😢  So consider logging all the errors, but only creating one popup with the list of database names that failed.

extensions/ql-vscode/src/extension.ts
shati-patel: > We don't know the language of the query at this point do we? It would be very cool to automatically filter out databases that are of the wrong language. But no problem if we can't do that yet.

Indeed 😅 I couldn't work out how to extract that language information _before_ displaying the quick pick, so I stuck with this for now!

extensions/ql-vscode/src/extension.ts
shati-patel: Thank you! I had a go at this in the latest commit. It turns out that `showAndLogErrorMessage` completely ignores newlines (even in the logs), so I've used `showAndLogWarningMessage` instead to make things a bit more readable...

I ended up creating two separate variables (`skippedDatabases` and `errors`) for tracking the errors, which feels a bit messy? Let me know if there's a neater/more compact way of doing this!

extensions/ql-vscode/src/extension.ts
```diff
@@ -467,9 +468,11 @@ async function activateWithInstalledDistribution(
     selectedQuery: Uri | undefined,
     progress: ProgressCallback,
     token: CancellationToken,
+    databaseItem: DatabaseItem | undefined,
   ): Promise<void> {
     if (qs !== undefined) {
-      const dbItem = await databaseUI.getDatabaseItem(progress, token);
+      // If no databaseItem is specified, use the database currently selected in the Databases UI
+      const dbItem = databaseItem || await databaseUI.getDatabaseItem(progress, token);
```
adityasharad, line 475: Minor: you could use only one variable throughout:
```ts
databaseItem = databaseItem || await databaseUI.getDatabaseItem(progress, token);
```
and replace `dbItem` in the two uses below with `databaseItem`.

Why do this? To avoid accidentally introducing code later that uses `databaseItem` when we should be using `dbItem`. (This is a classic mistake I have made often when two variables have similar names but only one variable has the fully initialised value...)

adityasharad (approved):
This looks good! Add a changelog entry and we're good to go.

extensions/ql-vscode/src/extension.ts
shati-patel: Changed! Thanks for the tip—I can definitely see myself getting confused by variables otherwise 🙃

Issue #906, opened 2021-07-05
"Open Query Results" button causes error: "Cannot read property 'result' of undefined"


Marcono1234:
**Describe the bug**
When using the "Open Query Results" button of the Query History view an error occurs.
Button:
![Query History screenshot](https://user-images.githubusercontent.com/11685886/124470442-e9603d80-dd9b-11eb-9da9-0572d9fd6d2d.png)

Log:
```
Cannot read property 'result' of undefined (codeQLQueryHistory.itemClicked)
TypeError: Cannot read property 'result' of undefined
	at InterfaceManager.showResults (#REDACTED#\.vscode\extensions\github.vscode-codeql-1.5.1\out\interface.js:227:21)
	at showResultsForCompletedQuery (#REDACTED#\.vscode\extensions\github.vscode-codeql-1.5.1\out\extension.js:311:20)
	at showResults (#REDACTED#\.vscode\extensions\github.vscode-codeql-1.5.1\out\extension.js:291:41)
	at QueryHistoryManager.invokeCallbackOn (#REDACTED#\.vscode\extensions\github.vscode-codeql-1.5.1\out\query-history.js:207:19)
	at QueryHistoryManager.handleItemClicked (#REDACTED#\.vscode\extensions\github.vscode-codeql-1.5.1\out\query-history.js:318:24)
	at #REDACTED#\.vscode\extensions\github.vscode-codeql-1.5.1\out\query-history.js:189:25
	at #REDACTED#\.vscode\extensions\github.vscode-codeql-1.5.1\out\commandRunner.js:61:26
	...
```

Note that clicking on one of the history entries opens the query results just fine.

Though to be honest, I am not sure if the "Open Query Results" button adds any value in the first place, since you would most likely first have to select the entry from the history (by left clicking?), but by doing so you already opened the query results.

**Version**
CodeQL extension:
```
CodeQL extension version: 1.5.1 
CodeQL CLI version: 2.5.7 
Platform: win32 x64
```

VSCode:
```
Version: 1.57.1 (user setup)
Commit: 507ce72a4466fbb27b715c3722558bb15afa9f48
Date: 2021-06-17T13:28:07.755Z
Electron: 12.0.7
Chrome: 89.0.4389.128
Node.js: 14.16.0
V8: 8.9.255.25-electron.0
OS: Windows_NT x64 10.0.19041
```

**To reproduce**
1. Load any Java database
2. Run the following Quick Query
    ```ql
    import java

    select ""
    ```
[...]

PR #909, opened 2021-07-12
Fix error from "Open Query Results" button

shati-patel:
Fixes https://github.com/github/vscode-codeql/issues/906.

The "Open Query Results" button in the query history view currently doesn't work. Error: `Cannot read property 'result' of undefined (codeQLQueryHistory.itemClicked)`.

I experimented a bit with the `handleItemClicked`, and it looks like changing `singleItem` to `finalSingleItem` works as expected 🙂 

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add the _Add Database Source to Workspace_ command to the right-click context menu in the databases view. This lets users re-add a database's source folder to the workspace and browse the source code. [#891](https://github.com/github/vscode-codeql/pull/891)
 - Fix markdown rendering in the description of the `codeQL.cli.executablePath` setting. [#908](https://github.com/github/vscode-codeql/pull/908)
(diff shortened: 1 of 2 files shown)


adityasharad:
Good catch, this looks sensible! Maybe we should have a query for this kind of pattern.

shati-patel:
Aside: There was another point raised in the [bug report](https://github.com/github/vscode-codeql/issues/906):
> Though to be honest, I am not sure if the "Open Query Results" button adds any value in the first place, since you would most likely first have to select the entry from the history (by left clicking?), but by doing so you already opened the query results.

I agree that the button doesn't add any extra functionality, but I assume it was added for a reason 😄 I guess it makes the commands more obvious/discoverable?

adityasharad:
> I assume it was added for a reason 😄 I guess it makes the command more obvious/discoverable?

Looking at [git blame on that line](https://github.com/github/vscode-codeql/blame/7a58d360fdf3d78a5b7870e2ea61109cf6bc2de7/extensions/ql-vscode/package.json#L398) I think it was added so that we could have UI icons in the query history box for some of the common commands.

adityasharad (approved):

PR #915, opened 2021-07-26
Autodetect what language a query is for

shati-patel:
Adds a `findLanguage` function that uses `codeql resolve queries --format bylanguage` to detect a query's language. This might not work for some queries, so I've added a manual selection as a fallback.

This is used in two places:
- When running a query on multiple databases (we now only show the databases that correspond to the query language)
- When running a remote query (we can now autodetect a language when it isn't specified in the `.repositories` config. This config file is going to be replaced by a different UI soon.)

Question: Does this need tests? I wasn't sure where/how to go about that!

## Checklist

   - Only the "multiple DBs" part is user-facing for now
   - No docs update required (beyond pending update from https://github.com/github/codeql/pull/6322)

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -8,7 +8,7 @@ import { Readable } from 'stream';
 import { StringDecoder } from 'string_decoder';
 import * as tk from 'tree-kill';
 import { promisify } from 'util';
-import { CancellationToken, Disposable } from 'vscode';
+import { CancellationToken, Disposable, Uri } from 'vscode';
 
 import { BQRSInfo, DecodedBqrsChunk } from './pure/bqrs-cli-types';
 import { CliConfig } from './config';
@@ -44,6 +44,16 @@ export interface QuerySetup {
 }
 
 /**
+ * The expected output of `codeql resolve queries --format bylanguage`.
+ */
+export interface QueryInfoByLanguage {
+  // Using `unknown` as a placeholder. For now, the value is only ever an empty object.
+  byLanguage: Record<string, Record<string, unknown>>;
+  noDeclaredLanguage: Record<string, unknown>;
+  multipleDeclaredLanguages: Record<string, unknown>;
+}
+
+/**
  * The expected output of `codeql resolve database`.
  */
 export interface DbInfo {
@@ -70,6 +80,11 @@ export interface UpgradesInfo {
  * The expected output of `codeql resolve qlpacks`.
  */
 export type QlpacksInfo = { [name: string]: string[] };
+
+/**
+ * The expected output of `codeql resolve languages`.
+ */
+export type LanguagesInfo = { [name: string]: string[] };
 
 /**
  * The expected output of `codeql resolve qlref`.
@@ -483,6 +498,20 @@ export class CodeQLCliServer implements Disposable {
   }
 
   /**
+   * Resolves the language for a query.
+   * @param queryUri The URI of the query
+   */
+  async resolveQueryByLanguage(workspaces: string[], queryUri: Uri): Promise<QueryInfoByLanguage> {
+    const subcommandArgs = [
+      '--format', 'bylanguage',
+      queryUri.fsPath,
+      '--additional-packs',
+      workspaces.join(path.delimiter)
+    ];
+    return JSON.parse(await this.runCodeQlCliCommand(['resolve', 'queries'], subcommandArgs, 'Resolving query by language'));
+  }
+
+  /**
    * Finds all available QL tests in a given directory.
    * @param testPath Root of directory tree to search for tests.
    * @returns The list of tests that were found.
@@ -722,6 +751,14 @@ export class CodeQLCliServer implements Disposable {
       args,
       'Resolving qlpack information',
     );
+  }
+
+  /**
+   * Gets information about the available languages.
+   * @returns A dictionary mapping language name to the directory it comes from
+   */
+  async resolveLanguages(): Promise<LanguagesInfo> {
+    return await this.runJsonCodeQlCliCommand<LanguagesInfo>(['resolve', 'languages'], [], 'Resolving languages');
   }
 
   /**
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -73,7 +73,7 @@ import {
 import { CodeQlStatusBarHandler } from './status-bar';
 
 import { Credentials } from './authentication';
-import runRemoteQuery from './run-remote-query';
+import { runRemoteQuery, findLanguage } from './run-remote-query';
 
 /**
  * extension.ts
@@ -570,7 +570,13 @@ async function activateWithInstalledDistribution(
         token: CancellationToken,
         uri: Uri | undefined
       ) => {
-        const quickPickItems = dbm.databaseItems.map<DatabaseQuickPickItem>(dbItem => (
+        const queryLanguage = await findLanguage(cliServer, uri);
+        const filteredDBs = dbm.databaseItems.filter(db => db.language === queryLanguage);
+        if (filteredDBs.length === 0) {
+          void helpers.showAndLogErrorMessage(`No databases found for language ${queryLanguage}`);
+          return;
+        }
+        const quickPickItems = filteredDBs.map<DatabaseQuickPickItem>(dbItem => (
           {
             databaseItem: dbItem,
             label: dbItem.name,
@@ -707,7 +713,7 @@ async function activateWithInstalledDistribution(
     ) => {
       if (isCanary()) {
         const credentials = await Credentials.initialize(ctx);
-        await runRemoteQuery(credentials, uri || window.activeTextEditor?.document.uri);
+        await runRemoteQuery(cliServer, credentials, uri || window.activeTextEditor?.document.uri);
       }
     })
   );
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -570,11 +570,15 @@ async function activateWithInstalledDistribution(
         token: CancellationToken,
         uri: Uri | undefined
       ) => {
+        let filteredDBs = dbm.databaseItems;
+        // If possible, only show databases with the right language (otherwise show all databases).
         const queryLanguage = await findLanguage(cliServer, uri);
-        const filteredDBs = dbm.databaseItems.filter(db => db.language === queryLanguage);
-        if (filteredDBs.length === 0) {
-          void helpers.showAndLogErrorMessage(`No databases found for language ${queryLanguage}`);
-          return;
+        if (queryLanguage) {
+          filteredDBs = dbm.databaseItems.filter(db => db.language === queryLanguage);
+          if (filteredDBs.length === 0) {
+            void helpers.showAndLogErrorMessage(`No databases found for language ${queryLanguage}. Please add a suitable database to your workspace.`);
+            return;
+          }
         }
         const quickPickItems = filteredDBs.map<DatabaseQuickPickItem>(dbItem => (
           {
@@ -588,7 +592,7 @@ async function activateWithInstalledDistribution(
          */
         const quickpick = await window.showQuickPick<DatabaseQuickPickItem>(
           quickPickItems,
-          { canPickMany: true }
+          { canPickMany: true, ignoreFocusOut: true }
         );
         if (quickpick !== undefined) {
           // Collect all skipped databases and display them at the end (instead of popping up individual errors)
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -571,6 +571,10 @@ async function activateWithInstalledDistribution(
         uri: Uri | undefined
       ) => {
         let filteredDBs = dbm.databaseItems;
+        if (filteredDBs.length === 0) {
+          void helpers.showAndLogErrorMessage('No databases found. Please add a suitable database to your workspace.');
+          return;
+        }
         // If possible, only show databases with the right language (otherwise show all databases).
         const queryLanguage = await findLanguage(cliServer, uri);
         if (queryLanguage) {
--- a/extensions/ql-vscode/src/run-remote-query.ts
+++ b/extensions/ql-vscode/src/run-remote-query.ts
@@ -23,12 +23,14 @@ const REPO = 'qc-controller';
 export async function findLanguage(
   cliServer: cli.CodeQLCliServer,
   queryUri: Uri | undefined
-): Promise<string> {
+): Promise<string | undefined> {
   const uri = queryUri || window.activeTextEditor?.document.uri;
   if (uri !== undefined) {
     try {
       const queryInfo = await cliServer.resolveQueryByLanguage(getOnDiskWorkspaceFolders(), uri);
-      return (Object.keys(queryInfo.byLanguage))[0];
+      const language = (Object.keys(queryInfo.byLanguage))[0];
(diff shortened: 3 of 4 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -482,6 +482,30 @@ export class CodeQLCliServer implements Disposable {
     return await this.runJsonCodeQlCliCommand<QuerySetup>(['resolve', 'library-path'], subcommandArgs, 'Resolving library paths');
   }
 
+  /**
+   * Finds the language that a query is analyzing.
+   * @param queryUri The URI of the query
+   */
+
+  // Note: this currently fails for some queries, e.g. the "example" snippets..."
```
shati-patel, line 490: Findings so far:
1. `findLanguage` works fine for all standard queries I've tried in the `<lang>/ql/src` folder
2. It fails for other queries, e.g. 
    - quick queries
    - example snippets in `<lang>/ql/examples/snippets`
    - [custom examples](https://github.com/github/vscode-codeql-starter/tree/main/codeql-custom-queries-cpp) in the starter workspace

For those failures, `codeql` seems unable to resolve the dependency in the qlpack 🤔 

<details><summary>Expand for log</summary>



```
Finding query language using CodeQL CLI: resolve queries -v --log-to-stderr --format bylanguage [redacted]\vscode-codeql-starter\ql\cpp\ql\examples\snippets\addressof.ql...
Sending shutdown request
Sent shutdown request
[2021-07-26 15:11:39] Calling plumbing command: codeql resolve library-path --qlconfig-file=[redacted]\AppData\Local\Programs\Microsoft VS Code\qlconfig.yml --query=[redacted]\vscode-codeql-starter\ql\cpp\ql\examples\snippets\addressof.ql --find-extractors --full-library-path=none --dbscheme=none --format=json
[2021-07-26 15:11:39] [DETAILS] resolve library-path> Resolving query at normalized path [redacted]\vscode-codeql-starter\ql\cpp\ql\examples\snippets\addressof.ql.
[2021-07-26 15:11:39] [DETAILS] resolve library-path> Found enclosing pack 'codeql-cpp-examples' at [redacted]\vscode-codeql-starter\ql\cpp\ql\examples.
[2021-07-26 15:11:39] [DETAILS] resolve library-path> Adding compilation cache at [redacted]\vscode-codeql-starter\ql\cpp\ql\examples\.cache.
[2021-07-26 15:11:39] [DETAILS] resolve library-path> Resolving library dependencies from [redacted]\vscode-codeql-starter\ql\cpp\ql\examples\qlpack.yml.
[2021-07-26 15:11:39] [ERROR] resolve library-path> ERROR: Dependency not found: codeql-cpp ([redacted]\vscode-codeql-starter\ql\cpp\ql\examples\qlpack.yml:1,1-1)
A fatal error occurred: Could not resolve library path for [redacted]\vscode-codeql-starter\ql\cpp\ql\examples
[2021-07-26 15:11:39] Exception caught at top level: Could not resolve library path for [redacted]\vscode-codeql-starter\ql\cpp\ql\examples
                      com.semmle.cli2.resolve.ResolveLibraryPathCommand.executeJSON(ResolveLibraryPathCommand.java:232)
                      com.semmle.cli2.resolve.ResolveLibraryPathCommand.executeJSON(ResolveLibraryPathCommand.java:45)
                      com.semmle.cli2.picocli.SimpleJsonSubcommand.executeInsistingOnJSON(SimpleJsonSubcommand.java:104)
                      com.semmle.cli2.picocli.PlumbingRunner.call(PlumbingRunner.java:54)
                      com.semmle.cli2.picocli.SubcommandCommon.callPlumbingInProcess(SubcommandCommon.java:123)
                      com.semmle.cli2.resolve.ResolveQueriesCommand.splitByLanguage(ResolveQueriesCommand.java:867)
                      com.semmle.cli2.resolve.ResolveQueriesCommand.printNonJsonResult(ResolveQueriesCommand.java:841)
                      com.semmle.cli2.resolve.ResolveQueriesCommand.printNonJsonResult(ResolveQueriesCommand.java:63)
                      com.semmle.cli2.picocli.SimpleJsonSubcommand.executeSubcommand(SimpleJsonSubcommand.java:84)
                      com.semmle.cli2.picocli.SubcommandCommon.executeWithParent(SubcommandCommon.java:432)
                      com.semmle.cli2.execute.CliServerCommand.lambda$executeSubcommand$0(CliServerCommand.java:67)
                      com.semmle.cli2.picocli.SubcommandMaker.runMain(SubcommandMaker.java:205)
                      com.semmle.cli2.execute.CliServerCommand.executeSubcommand(CliServerCommand.java:67)
                      com.semmle.cli2.picocli.SubcommandCommon.call(SubcommandCommon.java:448)
                      com.semmle.cli2.picocli.SubcommandMaker.runMain(SubcommandMaker.java:205)
                      com.semmle.cli2.picocli.SubcommandMaker.runMain(SubcommandMaker.java:214)
                      com.semmle.cli2.CodeQL.main(CodeQL.java:98)
```
</details>

shati-patel: Woops, looks like I just needed to specify some `--additional-packs` 😄

extensions/ql-vscode/src/cli.ts
```diff
@@ -482,6 +482,36 @@ export class CodeQLCliServer implements Disposable {
...
+        '--format', 'bylanguage',
+        queryUri.fsPath,
+        '--additional-packs',
+        workspaces.join(path.delimiter)
+      ];
+      try {
+        const JSONResults = await this.runCodeQlCliCommand(['resolve', 'queries'], subcommandArgs, 'Finding query language');
+
+        // This corresponds to `codeql resolve queries <query> --format bylanguage | jq -r '.byLanguage | keys[]'`
+        language = Object.keys(JSON.parse(JSONResults).byLanguage)[0];
```
edoardopirovano, line 502: I would slightly lean towards adding a new type like we have for other CLI commands (e.g. `TestCompleted`) to capture the output of `codeql resolve queries --format bylanguage`) and then using that result type rather than doing this.

edoardopirovano: To elaborate a bit on why I think that's better: Capturing the whole result of the comment rather than just doing some ad-hoc parsing of it here means that if in future we want to call that command for some other reason that uses another part of the output, it will be much easier to do.

shati-patel: That makes sense, thanks! 🎉 I'll have a go at capturing the result as a type 😃

extensions/ql-vscode/src/cli.ts
```diff
@@ -43,6 +43,16 @@ export interface QuerySetup {
   compilationCache?: string;
 }
 
+/**
+ * The expected output of `codeql resolve queries --format bylanguage`.
+ */
+export interface QueryInfoByLanguage {
+  // What should the value of these objects be? I've only ever seen it as an empty object.
```
edoardopirovano, line 50: Yeah, it's currently just a placeholder and is only ever an empty object. If `unknown` below is enough for the JSON deserializer to be happy then that seems as good a placeholder as any for the moment.

shati-patel: Great, thank you! (And thanks for the review in general ⚡ ) 

I'll convert this back to draft while I tidy up the rest 😀

extensions/ql-vscode/src/cli.ts
```diff
@@ -482,6 +492,20 @@ export class CodeQLCliServer implements Disposable {
     return await this.runJsonCodeQlCliCommand<QuerySetup>(['resolve', 'library-path'], subcommandArgs, 'Resolving library paths');
   }
 
+  /**
+   * Finds the language that a query is analyzing.
```
aeisenberg, line 496: nit: I don't think _analyzing_ is the right word. But I'm not really sure what the best word is. This suggestion is possibly, slightly better.

```suggestion
   * Finds the language that a query targets.
```

aeisenberg (commented):
This looks right to me!

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -1,20 +1,51 @@
...
+    }
+  }
+  const availableLanguages = Object.keys(await cliServer.resolveLanguages());
+  const language = await window.showQuickPick(
+    availableLanguages,
+    { placeHolder: 'Select target language for your query', ignoreFocusOut: true }
+  ) || '';
+  if (language === '') {
+    // This only happens if the user cancels the quick pick.
+    void showAndLogErrorMessage('Language not found. Language must be specified manually.');
```
edoardopirovano, line 43: I notice that in this case we continue to the next line and return an empty string. It might be better to change the type of this function to `string | undefined` and then places where this function are called will be forced to handle the `undefined` case explicitly. In particular, I think you'll want to:
1. When filtering databases, simply skip the filtering and show all databases.
2. When running queries remotely, not make the API call without a language.

edoardopirovano (commented):
Mostly looks good apart from a minor comment inline. I'll hold off on approving it, though, since we paired on this for a bit so it should probably have another set of eyes.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -1,20 +1,51 @@
...
+ */
+export async function findLanguage(
+  cliServer: cli.CodeQLCliServer,
+  queryUri: Uri | undefined
+): Promise<string> {
+  const uri = queryUri || window.activeTextEditor?.document.uri;
+  if (uri !== undefined) {
+    try {
+      const queryInfo = await cliServer.resolveQueryByLanguage(getOnDiskWorkspaceFolders(), uri);
+      return (Object.keys(queryInfo.byLanguage))[0];
```
adityasharad, line 31: Before you return this value, I suggest saving it in a `const` variable and logging a message. That way when looking at the extension log, we can immediately tell whether we followed this code path and successfully detected a language.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -1,20 +1,51 @@
...
+    } catch (e) {
+      void logger.log('Could not autodetect query language. Select language manually.');
+    }
+  }
+  const availableLanguages = Object.keys(await cliServer.resolveLanguages());
+  const language = await window.showQuickPick(
+    availableLanguages,
+    { placeHolder: 'Select target language for your query', ignoreFocusOut: true }
+  ) || '';
+  if (language === '') {
```
adityasharad, line 41: Could this ever be `undefined` rather than the empty string? If yes, `if(!language)` will handle both cases.

extensions/ql-vscode/src/cli.ts
```diff
@@ -482,6 +497,20 @@ export class CodeQLCliServer implements Disposable {
     return await this.runJsonCodeQlCliCommand<QuerySetup>(['resolve', 'library-path'], subcommandArgs, 'Resolving library paths');
   }
 
+  /**
+   * Resolves the language for a query.
+   * @param queryUri The URI of the query
+   */
+  async resolveQueryByLanguage(workspaces: string[], queryUri: Uri): Promise<QueryInfoByLanguage> {
```
adityasharad, line 504: Since you asked about tests:
- `extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts` is a good place to add integration tests for the `resolveQueryByLanguage` and `resolveLanguages` functions.
- You can use the existing tests as an example.
- `resolveLanguages` is probably easier to write a test for because there's no query as input. 
- `resolveQueryByLanguage` will require a test query: there are some examples in the `data` folders, but as far as I can tell none of them depend on an actual language. So you may need to add one, or save this test for later.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -1,20 +1,51 @@
...
 // Test "controller" repository and workflow.
 const OWNER = 'dsp-testing';
 const REPO = 'qc-controller';
 
-export default async function runRemoteQuery(credentials: Credentials, uri?: Uri) {
+/**
+ * Finds the language that a query targets.
+ * If it can't be autodetected, prompt the user to specify the language manually.
+ */
+export async function findLanguage(
```
adityasharad, line 23: Since you asked about tests: I agree that having tests for this function would be handy, and give you confidence in its behaviour. However we don't yet have a test module for `run-remote-query` due to the experimental nature of the entire feature, so adding a test requires some upfront work. (In future it will be easier :) )

I'll record some pointers here about how you could create those tests, but I'm not expecting you to do it in this PR. We can look at test coverage together after this PR.
- For integration tests (which use a real query and actually run the underlying commands), copy `extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts` as a template. I think it would make sense to have:
  - one test function that runs `findLanguage` on a query with a real language dependency and identifies the language
  - one test function that runs it on a dummy query and returns `undefined`
  - any other cases you can think of
- For unit tests (which don't run the actual commands, but simulate the underlying behaviour), use `extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts` as a template. Those tests mock the behaviour of the CLI server.

adityasharad (commented):
Looks good! I think the code is mostly ready to go, with one outstanding suggestion from Edoardo about return types, and one from me about logging.

I've added some suggestions about tests, but I think those are fine to add later, since it will require some upfront work to set up tests for this part of the code.

Have we been adding changelog entries for this feature?

shati-patel:
Thank you for the reviews ✨  I've addressed the main comments in https://github.com/github/vscode-codeql/pull/915/commits/83c68ecbf89415c462c83ac9fd4da8efbe2a2f39 and added a minimal changelog entry that covers the user-facing part of this.

I'll look at adding tests in a separate PR!

extensions/ql-vscode/src/extension.ts
```diff
@@ -570,7 +570,17 @@ async function activateWithInstalledDistribution(
...
+        // If possible, only show databases with the right language (otherwise show all databases).
+        const queryLanguage = await findLanguage(cliServer, uri);
+        if (queryLanguage) {
+          filteredDBs = dbm.databaseItems.filter(db => db.language === queryLanguage);
+          if (filteredDBs.length === 0) {
+            void helpers.showAndLogErrorMessage(`No databases found for language ${queryLanguage}. Please add a suitable database to your workspace.`);
+            return;
+          }
+        }
+        const quickPickItems = filteredDBs.map<DatabaseQuickPickItem>(dbItem => (
```
edoardopirovano, line 583: Note that if the user has no databases and doesn't pick a language when prompted, we try to display this prompt with nothing in it. There should probably be an extra check to display an error in this case and return, similarly to what happens with the filtered list above.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -34,9 +67,13 @@ export default async function runRemoteQuery(credentials: Credentials, uri?: Uri
   const config = yaml.safeLoad(await fs.readFile(repositoriesFile, 'utf8')) as Config;
 
   const ref = config.ref || 'main';
-  const language = config.language;
+  const language = config.language || await findLanguage(cliServer, uri);
   const repositories = config.repositories;
 
+  if (!language) {
+    return;
```
edoardopirovano, line 74: I would add a brief comment stating that we don't need an error message here because `findLanguage` already displays one.

edoardopirovano (commented):
A couple more minor comments, but otherwise looks good!

adityasharad (approved):

extensions/ql-vscode/src/extension.ts
shati-patel: Woops, good point! Thanks 😄

edoardopirovano (approved):
Thanks for addressing all my comments, this LGTM 👍🏼

PR #916, opened 2021-07-26
Update the CLI versions to run integration tests against

aeisenberg:
Small change to ensure this extension tests against the most recent version of the CLI.

### DIFF
--- a/extensions/ql-vscode/src/vscode-tests/ensureCli.ts
+++ b/extensions/ql-vscode/src/vscode-tests/ensureCli.ts
@@ -12,7 +12,7 @@ import { workspace } from 'vscode';
  *       version. Note that for now, we must maintain the default version by hand.
  *       This may be set to `nightly`, in which case the `NIGHTLY_URL` variable must
  *       also be set.
- * 
+ *
  *     - NIGHTLY_URL: The URL for a nightly release of the CodeQL CLI that will be
  *       used if `CLI_VERSION` is set to `nightly`.
  *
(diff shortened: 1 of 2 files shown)


shati-patel (approved):

extensions/ql-vscode/src/vscode-tests/ensureCli.ts
```diff
@@ -44,7 +44,7 @@ const _10MB = _1MB * 10;
 
 // CLI version to test. Hard code the latest as default. And be sure
 // to update the env if it is not otherwise set.
-const CLI_VERSION = process.env.CLI_VERSION || 'v2.5.5';
+const CLI_VERSION = process.env.CLI_VERSION || 'v2.5.8';
```
shati-patel, line 47: (Woops, I didn't know about this one! I haven't updated in the last few release 🤦🏽 )

aeisenberg: I forgot about that too. 🤷🏼 

Maybe we can change this to `latest` and never need to update it again.

edoardopirovano: I think this shouldn't matter too much as it is only relevant when tests are run locally - the Actions workflow all set the `CLI_VERSION` environment variable that overrides this. Note that setting this to `nightly` (I assume that's what you meant since I don't think `latest` is a valid value?) won't work since that requires the `NIGHTLY_URL` environment variable to be set which defeats the point of this line (being able to run the tests locally where the environment variables _aren't_ set).

aeisenberg: Ah...I didn't know that the `NIGHTLY_URL` variable is required, so forget about the idea.  And, yes, this is low priority so if it is missed during an update, not a big deal.

PR #918, opened 2021-08-10
Running remote queries: Save repo lists in VS Code config

shati-patel:
This PR tweaks how you run remote queries 😎 
Previously, each query needed a `.repositories` file that listed which repositories you wanted to analyze.

Now, you can save pre-defined "repository lists" in your VS Code settings, and choose one of those when you run a remote query. 

New flow:
1. If you still have a `.repositories` file, we use the information from there to run your remote query.
2. Otherwise, if you have saved repository lists in your VS Code settings, you can select one of those to run against.
3. If not, you can enter a single repo in a text input box and we run against that.

The `getRepositories` function should give an error if you end up with an empty string/array/object. It doesn't check if your repositories are valid, but the API call itself gives an error in that case (e.g. ❌ `"HttpError: Repository 'xxx' not found"`).

Question:
- Since this isn't a secret, is it okay for this internal setting to be visible in the Settings UI? (@jf205?)
image
  I could hide it completely by omitting the [package.json](https://github.com/github/vscode-codeql/pull/918/files#diff-be4c5a31505df7de81fa4900e84e66c04d1a6bca9b3da50f8061d2cb2bfbbde1) change, but that would remove the validation (that the setting has the correct "shape", i.e. an object where the values are arrays of strings)

## Checklist

N/A: this is part of an internal-only feature (see linked issue for more context/details).

### DIFF
--- a/extensions/ql-vscode/src/config.ts
+++ b/extensions/ql-vscode/src/config.ts
@@ -291,3 +291,17 @@ export function isCanary() {
  * Avoids caching in the AST viewer if the user is also a canary user.
  */
 export const NO_CACHE_AST_VIEWER = new Setting('disableCache', AST_VIEWER_SETTING);
+
+/*
+ * Lists of GitHub repositories that you want to query remotely via the "Run Remote query" command.
+ * Note: This command is only available for internal users.
+ * 
+ * This setting should be a JSON object where each key is a user-specified name (string),
+ * and the value is an array of GitHub repositories (of the form `<owner>/<repo>`).
+ */
+
+const REMOTE_REPO_LISTS = new Setting('remoteRepositoryLists', ROOT_SETTING);
+
+export function getRemoteRepositoryLists(): Record<string, string[]> | undefined {
+  return REMOTE_REPO_LISTS.getValue<Record<string, string[]>>() || undefined;
+}
--- a/extensions/ql-vscode/src/config.ts
+++ b/extensions/ql-vscode/src/config.ts
@@ -292,14 +292,13 @@ export function isCanary() {
  */
 export const NO_CACHE_AST_VIEWER = new Setting('disableCache', AST_VIEWER_SETTING);
 
-/*
+/**
  * Lists of GitHub repositories that you want to query remotely via the "Run Remote query" command.
  * Note: This command is only available for internal users.
  * 
  * This setting should be a JSON object where each key is a user-specified name (string),
  * and the value is an array of GitHub repositories (of the form `<owner>/<repo>`).
  */
-
 const REMOTE_REPO_LISTS = new Setting('remoteRepositoryLists', ROOT_SETTING);
 
 export function getRemoteRepositoryLists(): Record<string, string[]> | undefined {
--- a/extensions/ql-vscode/src/run-remote-query.ts
+++ b/extensions/ql-vscode/src/run-remote-query.ts
@@ -78,7 +78,12 @@ async function getRepositories(): Promise<string[] | undefined> {
     }
   } else {
     void logger.log('No repository lists defined. Displaying text input box.');
-    const repoRegex = /^(?:[a-zA-Z0-9]+-?)*[a-zA-Z0-9]\/[a-zA-Z0-9-_]+$/;
+    /**
+     * This regex matches strings of the form `owner/repo` where:
+     * - `owner` is made up of alphanumeric characters or single hyphens, starting and ending in an alphanumeric character
+     * - `repo` is made up of alphanumeric characters, hyphens, or underscores
+     */
+    const repoRegex = /^(?:[a-zA-Z0-9]+-)*[a-zA-Z0-9]+\/[a-zA-Z0-9-_]+$/;
     const remoteRepo = await window.showInputBox({
       title: 'Enter a GitHub repository in the format <owner>/<repo> (e.g. github/codeql)',
       placeHolder: '<owner>/<repo>',
--- a/extensions/ql-vscode/src/run-remote-query.ts
+++ b/extensions/ql-vscode/src/run-remote-query.ts
@@ -69,7 +69,7 @@ async function getRepositories(): Promise<string[] | undefined> {
         placeHolder: 'Select a repository list. You can define repository lists in the `codeQL.remoteRepositoryLists` setting.',
         ignoreFocusOut: true,
       });
-    if (quickpick && quickpick.repoList.length > 0) {
+    if (quickpick?.repoList.length) {
       void logger.log(`Selected repositories: ${quickpick.repoList}`);
       return quickpick.repoList;
     } else {
@@ -78,6 +78,7 @@ async function getRepositories(): Promise<string[] | undefined> {
     }
   } else {
     void logger.log('No repository lists defined. Displaying text input box.');
+    const repoRegex = /^(?:[a-zA-Z0-9]+-?)*[a-zA-Z0-9]\/[a-zA-Z0-9-_]+$/;
     const remoteRepo = await window.showInputBox({
       title: 'Enter a GitHub repository in the format <owner>/<repo> (e.g. github/codeql)',
       placeHolder: '<owner>/<repo>',
@@ -86,6 +87,9 @@ async function getRepositories(): Promise<string[] | undefined> {
     });
     if (!remoteRepo) {
       void showAndLogErrorMessage('No repositories entered.');
+      return;
+    } else if (!repoRegex.test(remoteRepo)) { // Check if user entered invalid input
+      void showAndLogErrorMessage('Invalid repository format. Must be in the format <owner>/<repo> (e.g. github/codeql)');
       return;
     }
     void logger.log(`Entered repository: ${remoteRepo}`);
--- a/extensions/ql-vscode/src/run-remote-query.ts
+++ b/extensions/ql-vscode/src/run-remote-query.ts
@@ -77,8 +77,19 @@ async function getRepositories(): Promise<string[] | undefined> {
       return;
     }
   } else {
-    void showAndLogErrorMessage('No repository lists defined. You can define repository lists in the `codeQL.remoteRepositoryLists` setting.');
-    return;
+    void logger.log('No repository lists defined. Displaying text input box.');
+    const remoteRepo = await window.showInputBox({
+      title: 'Enter a GitHub repository in the format <owner>/<repo> (e.g. github/codeql)',
+      placeHolder: '<owner>/<repo>',
+      prompt: 'Tip: you can save frequently used repositories in the `codeql.remoteRepositoryLists` setting',
+      ignoreFocusOut: true,
+    });
+    if (!remoteRepo) {
+      void showAndLogErrorMessage('No repositories entered.');
+      return;
+    }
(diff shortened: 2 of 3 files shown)


extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -1,17 +1,20 @@
...
-
+import { getRemoteRepositoryLists } from './config';
 interface Config {
   repositories: string[];
   ref?: string;
   language?: string;
 }
 
+// User-defined repository lists.
+const REPO_LISTS = getRemoteRepositoryLists();
```
aeisenberg, line 16: It looks like this gets the repository lists once and then it is fixed. What happens if someone changes their repository lists? Do they need to restart?

Maybe, instead of a list, call `getRemoteRepositoryLists()` every time immediately before you use it.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -47,6 +50,40 @@ export async function findLanguage(
...
 
+interface RepoListQuickPickItem extends QuickPickItem {
+  repoList: string[];
+}
+
+/**
+ * Gets the repositories to run the query against.
+ */
+async function getRepositories(): Promise<string[] | undefined> {
+  if (REPO_LISTS) {
```
aeisenberg, line 61: Perhaps better to also check that `REPO_LISTS` is not an empty object, eg-

```suggestion
  if (Object.keys(REPO_LISTS || {}).length) {
```

Also, see my comment above. This might be better:

```suggestion
  if (Object.keys(getRemoteRepositoryLists() || {}).length) {
```

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -47,6 +50,40 @@ export async function findLanguage(
...
+    const quickPickItems = Object.keys(REPO_LISTS).map<RepoListQuickPickItem>(key => (
+      {
+        label: key,                // the name of the repository list
+        repoList: REPO_LISTS[key], // the actual array of repositories
+      }
+    ));
```
aeisenberg, line 67: nit: Maybe a little cleaner.

```suggestion
    const quickPickItems = Object.enties(REPO_LISTS).map<RepoListQuickPickItem>(([key, value]) => (
      {
        label: key,                // the name of the repository list
        repoList: value, // the actual array of repositories
      }
    ));
```

aeisenberg: Also, note that repositories lists is a user setting and it may be malformed. Just make sure that if a malformed list (eg- a single string or even a number) is passed, then the error is expressed to the user in an understandable way.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -47,6 +50,40 @@ export async function findLanguage(
...
+    const quickpick = await window.showQuickPick<RepoListQuickPickItem>(
+      quickPickItems,
+      {
+        placeHolder: 'Select a repository list. You can define repository lists in the `codeQL.remoteRepositoryLists` setting.',
+        ignoreFocusOut: true,
+      });
+    if (quickpick && quickpick.repoList.length > 0) {
```
aeisenberg, line 74: nit:

```suggestion
    if (quickpick?.repoList.length > 0) {
```

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -59,19 +96,32 @@ export async function runRemoteQuery(cliServer: cli.CodeQLCliServer, credentials
...
+  if (await fs.pathExists(repositoriesFile)) {
+    void logger.log(`Found '${repositoriesFile}'. Using information from that file to run ${queryFile}.`);
 
-  const config = yaml.safeLoad(await fs.readFile(repositoriesFile, 'utf8')) as Config;
+    const config = yaml.safeLoad(await fs.readFile(repositoriesFile, 'utf8')) as Config;
 
-  const ref = config.ref || 'main';
-  const language = config.language || await findLanguage(cliServer, uri);
-  const repositories = config.repositories;
+    ref = config.ref || 'main';
```
aeisenberg, line 110: Not something to handle now, but this means that users can only choose a ref if they have a .repositories file. If there is no file, then they can only choose `main`. This is fine for now, but we should consider making this more flexible this later.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -59,19 +96,32 @@ export async function runRemoteQuery(cliServer: cli.CodeQLCliServer, credentials
...
+    language = config.language || await findLanguage(cliServer, uri);
+    repositories = config.repositories;
+  } else {
+    ref = 'main';
+    language = await findLanguage(cliServer, uri);
+    repositories = await getRepositories();
```
aeisenberg, line 116: Since `findLanguage` can take a while to run and `getRepositories` is blocked by a user event, here's a trick to get them to run in "parallel":

```suggestion
    [ language, repositories ] = await Promise.all([findLanguage(cliServer, uri), getRepositories()]);;
```

`Promise.all` accepts an array of promises and resolves when all of its arguments resolve. Or, it rejects if any one of them reject. The return value is an array containing all of the resolved values of each promise.

extensions/ql-vscode/src/run-remote-query.ts
shati-patel: I tried this and got an error in the lines below that "Object `quickpick` might be undefined" 😅

extensions/ql-vscode/src/run-remote-query.ts
shati-patel: > Also, note that repositories lists is a user setting and it may be malformed. Just make sure that if a malformed list (eg- a single string or even a number) is passed, then the error is expressed to the user in an understandable way.

👍🏽 Good point. I've re-instated the [package.json](https://github.com/github/vscode-codeql/pull/918/files#diff-be4c5a31505df7de81fa4900e84e66c04d1a6bca9b3da50f8061d2cb2bfbbde1) change, which at least validates that each list is an array of strings. (I removed it from the package.json initially since that stops it from showing up in the settings UI, but maybe we don't need to be too secretive 🤫 )

extensions/ql-vscode/src/run-remote-query.ts
shati-patel: 👍🏽 I expect that we won't need the `ref` at all eventually, though I will double check with the others 😃 
It just specifies from which branch of the controller repository we're running the "remote query" workflow. I can't see why it should ever be different from `main` (except for testing).

extensions/ql-vscode/src/run-remote-query.ts
aeisenberg: Hmmm...ok. Not sure what's happening, but fine to ignore.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -58,11 +55,12 @@ interface RepoListQuickPickItem extends QuickPickItem {
  * Gets the repositories to run the query against.
  */
 async function getRepositories(): Promise<string[] | undefined> {
-  if (REPO_LISTS) {
-    const quickPickItems = Object.keys(REPO_LISTS).map<RepoListQuickPickItem>(key => (
+  const REPO_LISTS = getRemoteRepositoryLists();
```
aeisenberg, line 58: nit: Typically, if something is not a module-level constant, then we avoid using CAPs for it.

```suggestion
  const repoLists = getRemoteRepositoryLists();
```

aeisenberg (commented):
Looking good!

extensions/ql-vscode/src/run-remote-query.ts
shati-patel: (Confirmed with @robertbrignull that we don't need to worry about easily setting the `ref` in VS Code. If required for testing, we can send a manual API request.)

extensions/ql-vscode/src/config.ts
```diff
@@ -291,3 +291,17 @@ export function isCanary() {
...
 export const NO_CACHE_AST_VIEWER = new Setting('disableCache', AST_VIEWER_SETTING);
+
+/*
+ * Lists of GitHub repositories that you want to query remotely via the "Run Remote query" command.
+ * Note: This command is only available for internal users.
+ * 
+ * This setting should be a JSON object where each key is a user-specified name (string),
+ * and the value is an array of GitHub repositories (of the form `<owner>/<repo>`).
+ */
+
```
edoardopirovano, line 302: Extreme nit:
```suggestion

```

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -47,6 +47,52 @@ export async function findLanguage(
...
+    }
+  } else {
+    void logger.log('No repository lists defined. Displaying text input box.');
+    const remoteRepo = await window.showInputBox({
+      title: 'Enter a GitHub repository in the format <owner>/<repo> (e.g. github/codeql)',
+      placeHolder: '<owner>/<repo>',
+      prompt: 'Tip: you can save frequently used repositories in the `codeql.remoteRepositoryLists` setting',
+      ignoreFocusOut: true,
+    });
+    if (!remoteRepo) {
```
edoardopirovano, line 87: Should we also do some basic validation here to ensure the format is indeed `<owner>/<repo>` and fail early (and with a useful error message) if this is indeed the case?

edoardopirovano (commented):
Some very minor comments but otherwise looks good!

extensions/ql-vscode/src/run-remote-query.ts
shati-patel: Thanks for the help on Slack ✨ Cross-posting the regex you shared here for reference: https://regex101.com/r/krDHm5/2

edoardopirovano: No problem! For future reference, I would also add a comment documenting what the regex is matching, i.e. strings of the form `owner/repo` where:
- `owner` is made up of alphanumeric characters or _single_ hyphens, starting and ending in an alphanumeric character
- `repo` is made up of alphanumeric characters, hyphens, or underscores

edoardopirovano:
I'm happy with this once the above comment has been added, will leave it to @aeisenberg to approve if he's happy his comments have been addressed.

shati-patel:
Oops, I've introduced a high-severity code scanning alert 😅 I'll try to figure out what I need to fix first... 👀

edoardopirovano:
> Oops, I've introduced a high-severity code scanning alert sweat_smile I'll try to figure out what I need to fix first... eyes

Ah, yes, Code Scanning is not too happy with our regex... I think changing it to `^(?:[a-zA-Z0-9]+-)*[a-zA-Z0-9]+\/[a-zA-Z0-9-_]+$` should get rid of the alert and matches the same things (see https://regex101.com/r/35eyb8/1).

edoardopirovano (approved):

jf205:
>Since this isn't a secret, is it okay for this internal setting to be visible in the Settings UI? 

👍🏻 (thanks for checking!)

aeisenberg (approved):
Apologies for taking so long to review. This looks good.

PR #924, opened 2021-08-18
Fix type for "remote repo list" setting

shati-patel:
Tiny follow-up to #918. The setting I added was being autofilled in a weird way:

24C31004-BFA8-4E3D-8526-A3A070B372E5

Allowing it to have type `null` seems to fix this 🙂


shati-patel:
Test failures seem unrelated to this change...

edoardopirovano (approved):
LGTM, should be fine to bypass the tests which are currently broken.

PR #926, opened 2021-08-18
PR Checks: Use version of `codeql` repo suitable for CLI version

edoardopirovano:
This PR changes our testing workflow to use a version of the `codeql` repository that is suitable for the CLI being tested against, instead of always checking out `main`. This is needed because some upcoming packaging changes have made the current `main` of `codeql` incompatible with old versions of the CLI that we still need to test against here.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -177,10 +177,26 @@ jobs:
           npm run build
         shell: bash
 
+      - name: Decide on ref of CodeQL repo
+        id: choose-ref
+        shell: bash
+        run: |
+          if [[ "${{ matrix.version }}" == "nightly" ]]
+          then
+            REF="codeql-cli/latest"
+          elif [[ "${{ matrix.version }}" == "v2.2.6" || "${{ matrix.version }}" == "v2.3.3" ]]
+          then
+            REF="codeql-cli/v2.4.5"
+          else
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -177,10 +177,26 @@ jobs:
...
 
+      - name: Decide on ref of CodeQL repo
+        id: choose-ref
+        shell: bash
+        run: |
+          if [[ "${{ matrix.version }}" == "nightly" ]]
+          then
+            REF="main"
+          elif [[ "${{ matrix.version }}" == "v2.2.6" || "${{ matrix.version }}" == "v2.3.3" ]]
+          then
```
aeisenberg, line 188: It's confusing that different versions are being used. Off the top of my head, I don't know if this is correct. But if you've checked the version numbers then this looks fine.

aeisenberg (approved):

.github/workflows/main.yml
edoardopirovano: These are indeed the two versions from [the ones we test against](https://github.com/github/vscode-codeql/blob/3536124fbc597fe282d539d1863e7baf27e443bb/.github/workflows/main.yml#L154) that are too old to have a tag on CodeQL. We could try to do something more clever for these like finding a ref from around the time they were released, but since it works like this and we won't be testing against them forever, I think this is fine.

Issue #832, opened 2021-04-15
Add csv export for non-alert data


aeisenberg:
Currently, csv export is only available for `problem` queries. However, some users are asking for csv export for other kinds of queries. We should be able to support this from within the vscode extension.

edoardopirovano:
Will be closed by https://github.com/github/vscode-codeql/pull/929

PR #929, opened 2021-08-19
Allow exporting of results for non-alert queries

edoardopirovano:
Closes https://github.com/github/vscode-codeql/issues/832

This PR adds a new right click option in the Query History view called _Export Results (CSV)_ which will export the results of a non-alert query to a CSV file (as chosen in a save dialog displayed to the user).

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Add support for filename pattern in history view. [#930](https://github.com/github/vscode-codeql/pull/930)
+- Add an option _Export Results (CSV)_ to export the results of a non-alert query. The existing options for alert queries have been renamed to _View Alerts_ to avoid confusion. [#929](https://github.com/github/vscode-codeql/pull/929)
 
 ## 1.5.3 - 18 August 2021
 
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [UNRELEASED]
 
 - Add support for filename pattern in history view. [#930](https://github.com/github/vscode-codeql/pull/930)
-- Add an option _Export Results (CSV)_ to export the results of a non-alert query. The existing options for alert queries have been renamed to _View Alerts_ to avoid confusion. [#929](https://github.com/github/vscode-codeql/pull/929)
+- Add an option _View Results (CSV)_ to view the results of a non-alert query. The existing options for alert queries have been renamed to _View Alerts_ to avoid confusion. [#929](https://github.com/github/vscode-codeql/pull/929)
 
 ## 1.5.3 - 18 August 2021
 
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -462,15 +462,15 @@
         "title": "Show Query Text"
       },
       {
-        "command": "codeQLQueryHistory.exportCsvResults",
-        "title": "Export Results (CSV)"
+        "command": "codeQLQueryHistory.viewCsvResults",
+        "title": "View Results (CSV)"
       },
       {
-        "command": "codeQLQueryHistory.viewCsvResults",
+        "command": "codeQLQueryHistory.viewCsvAlerts",
         "title": "View Alerts (CSV)"
       },
       {
-        "command": "codeQLQueryHistory.viewSarifResults",
+        "command": "codeQLQueryHistory.viewSarifAlerts",
         "title": "View Alerts (SARIF)"
       },
       {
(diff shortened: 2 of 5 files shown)


edoardopirovano:
Rebased to fix a conflict in the CHANGELOG.

extensions/ql-vscode/package.json
```diff
@@ -461,13 +461,17 @@
         "command": "codeQLQueryHistory.showQueryText",
         "title": "Show Query Text"
       },
+      {
+        "command": "codeQLQueryHistory.exportCsvResults",
+        "title": "Export Results (CSV)"
+      },
       {
         "command": "codeQLQueryHistory.viewCsvResults",
```
shati-patel, line 469: Minor: I see that the command itself is still called `viewCsvResults`. Is it worth changing that to `Alerts` too?
(possibly not, since it isn't user-visible 😄)

extensions/ql-vscode/package.json
```diff
@@ -461,13 +461,17 @@
...
-        "title": "View Results (CSV)"
+        "title": "View Alerts (CSV)"
       },
       {
         "command": "codeQLQueryHistory.viewSarifResults",
-        "title": "View Results (SARIF)"
+        "title": "View Alerts (SARIF)"
```
shati-patel, line 474: > `@github/docs-content-codeql` has been cc'd in all issues for UI or other user-facing changes made by this pull request.

Slightly tangential, but the docs are a bit outdated in this area and still mention a **View SARIF** command, instead of **View Results/Alerts (SARIF)**. Looks like this section (https://github.com/github/codeql/blob/main/docs/codeql/codeql-for-visual-studio-code/analyzing-your-projects.rst#viewing-query-results) should mention the correct name for this command, as well as mentioning CSV. I'll open a quick PR for this (and avoid pinging in the docs team) 🙂

shati-patel (commented):
I'll let someone else review/approve, who's more comfortable with the code changes (especially the changes to `run-queries`), but I tried this locally and it seems to work 😊

extensions/ql-vscode/package.json
```diff
@@ -461,13 +461,17 @@
         "command": "codeQLQueryHistory.showQueryText",
         "title": "Show Query Text"
       },
+      {
+        "command": "codeQLQueryHistory.exportCsvResults",
+        "title": "Export Results (CSV)"
```
aeisenberg, line 466: From a user's perspective (especially a beginner user), it might be confusing that for alert data, you "view" the results, but for non-alert data, you "export" the results. 

Here is one suggestion. Rename this command (or "View Alerts (CSV)") so that they are the same. And for this command instead of asking the user to choose a file name, auto-generate a file name and place it next to the BQRS file. This way, it feels the same to the user (even though under the hood, it is a different action). 

This is just a suggestion and I'm fine with it going in as it is.

extensions/ql-vscode/package.json
edoardopirovano: Good point, probably worth changing it too even it isn't user-facing to avoid confusion.

extensions/ql-vscode/package.json
edoardopirovano: Awesome, thanks for volunteering to do that 🙂

extensions/ql-vscode/package.json
edoardopirovano: Thanks, that does seem more intuitive. I have changed it to do that!

edoardopirovano:
Thanks for the helpful comments @aeisenberg and @shati-patel! I've pushed an additional commit addressing them.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Add support for filename pattern in history view. [#930](https://github.com/github/vscode-codeql/pull/930)
+- Add an option _Export Results (CSV)_ to export the results of a non-alert query. The existing options for alert queries have been renamed to _View Alerts_ to avoid confusion. [#929](https://github.com/github/vscode-codeql/pull/929)
```
shati-patel, line 6: Slight tweak to the change note 📝 
```suggestion
- Add an option _View Results (CSV)_ to view the results of a non-alert query. The existing options for alert queries have been renamed to _View Alerts_ to avoid confusion. [#929](https://github.com/github/vscode-codeql/pull/929)
```

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -216,6 +218,29 @@ export class QueryInfo {
     return this.dilPath;
   }
 
+  async exportCsvResults(qs: qsClient.QueryServerClient, csvPath: string, onFinish: () => void): Promise<void> {
+    let stopDecoding = false;
+    const out = fs.createWriteStream(csvPath);
```
aeisenberg, line 223: If the csv file already exists, then maybe no need to recreate it? Though if the operation is fast, not a big deal either way.

edoardopirovano: This function is already not being called if the file already exists (see [this line](https://github.com/github/vscode-codeql/blob/6564196328680cecee581353b80fa3e8e30b0ec5/extensions/ql-vscode/src/query-history.ts#L587))

shati-patel (approved):
Looks like all comments are addressed :)

Manouchehri:
Super excited to see this in the next release, thanks!

PR #931, opened 2021-08-23
New setting to specify number of paths per alert

### LINKED ISSUE #919: [Feature Request] Customize number of paths per alert
The path view seems to show a maximum of four paths per alert, with no option to customize it. Eclipse had an option to customize this: https://help.semmle.com/ql-for-eclipse/Content/WebHelp/preferences.html. I don't see any similar option in the settings for this extension, could an option be added to customize this?

---

### Implementation notes
- Add a `maxPaths` setting, probably under `runningQueries`
- Pass that value through to `codeql bqrs interpret` and `codeql database interpret-results` using their `--max-paths` argument.

shati-patel:
Will fix #919 .

Added a new setting `codeQL.runningQueries.maxPaths` that lets you configure how many paths to display 🙂 I tested this with a few databases/queries (e.g. [this one](https://lgtm.com/projects/g/codefollower/Tomcat-Research/alerts/?mode=list&id=java%2Fcommand-line-injection%2Cjava%2Fxss))

## Checklist

- [x] ~`@github/docs-content-codeql` has been cc'd in all issues for UI or other user-facing changes made by this pull request.~ No docs updates needed.

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -180,6 +180,11 @@
           "default": false,
           "description": "Enable debug logging and tuple counting when running CodeQL queries. This information is useful for debugging query performance."
         },
+        "codeQL.runningQueries.maxPaths": {
+          "type": "integer",
+          "default": 4,
+          "markdownDescription": "Max number of paths to display for each alert found by a path query (`@kind path-problem`)."
+        },
         "codeQL.runningQueries.autoSave": {
           "type": "boolean",
           "default": false,
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -183,6 +183,8 @@
         "codeQL.runningQueries.maxPaths": {
           "type": "integer",
           "default": 4,
+          "minimum": 1,
+          "maximum": 256,
           "markdownDescription": "Max number of paths to display for each alert found by a path query (`@kind path-problem`)."
         },
         "codeQL.runningQueries.autoSave": {
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -649,6 +649,11 @@ export class CodeQLCliServer implements Disposable {
       this.cliConfig.numberThreads.toString(),
     );
 
+    args.push(
+      '--max-paths',
+      this.cliConfig.maxPaths.toString(),
+    );
+
     args.push(resultsPath);
     await this.runCodeQlCliCommand(['bqrs', 'interpret'], args, 'Interpreting query results');
   }
--- a/extensions/ql-vscode/src/config.ts
+++ b/extensions/ql-vscode/src/config.ts
@@ -79,6 +79,7 @@ const CACHE_SIZE_SETTING = new Setting('cacheSize', RUNNING_QUERIES_SETTING);
 const TIMEOUT_SETTING = new Setting('timeout', RUNNING_QUERIES_SETTING);
 const MEMORY_SETTING = new Setting('memory', RUNNING_QUERIES_SETTING);
 const DEBUG_SETTING = new Setting('debug', RUNNING_QUERIES_SETTING);
+const MAX_PATHS = new Setting('maxPaths', RUNNING_QUERIES_SETTING);
 const RUNNING_TESTS_SETTING = new Setting('runningTests', ROOT_SETTING);
 const RESULTS_DISPLAY_SETTING = new Setting('resultsDisplay', ROOT_SETTING);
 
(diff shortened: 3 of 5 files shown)


aeisenberg (approved):
Nice.

extensions/ql-vscode/src/config.ts
```diff
@@ -112,12 +113,13 @@ export interface QueryHistoryConfig {
...
 }
 
-const CLI_SETTINGS = [ADDITIONAL_TEST_ARGUMENTS_SETTING, NUMBER_OF_TEST_THREADS_SETTING, NUMBER_OF_THREADS_SETTING];
+const CLI_SETTINGS = [ADDITIONAL_TEST_ARGUMENTS_SETTING, NUMBER_OF_TEST_THREADS_SETTING, NUMBER_OF_THREADS_SETTING, MAX_PATHS];
 
 export interface CliConfig {
   additionalTestArguments: string[];
   numberTestThreads: number;
   numberThreads: number;
+  maxPaths: number;
```
adityasharad, line 122: I think you can unit test this - take a look at https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/vscode-tests/minimal-workspace/config.test.ts and add it to the list of values. (I think it will go under the CLIConfigListener rather than the QueryServerConfigListener.)

adityasharad (commented):
Nice! A few suggestions but overall structure looks good :)

extensions/ql-vscode/package.json
```diff
@@ -180,6 +180,11 @@
           "default": false,
           "description": "Enable debug logging and tuple counting when running CodeQL queries. This information is useful for debugging query performance."
         },
+        "codeQL.runningQueries.maxPaths": {
+          "type": "integer",
+          "default": 4,
```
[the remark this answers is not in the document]
shati-patel: Makes sense! Do you have an idea what counts as a sensible upper limit? I've put 256 for now, since that's the highest number I've seen "[in the wild](https://lgtm.com/projects/g/openjdk/jdk/alerts/?mode=list&lang=java)".

extensions/ql-vscode/src/cli.ts
```diff
@@ -649,6 +649,11 @@ export class CodeQLCliServer implements Disposable {
       this.cliConfig.numberThreads.toString(),
     );
 
+    args.push(
```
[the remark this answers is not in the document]
shati-patel: I couldn't find `interpret-results` anywhere, so I guess not?

extensions/ql-vscode/src/config.ts
shati-patel: I've added it to that test file [here](https://github.com/github/vscode-codeql/pull/931/files#diff-f8a6abadae6a8fd37afb3cd61659401d1f7119e14c31ac26f2abb6cd57d37d55). Silly question, but what should `values` be? I couldn't work out if there's a logic to it 🙃

shati-patel:
Thanks for the review and the helpful pointers! 🎉 I've addressed comments in the latest commit (and added some more questions)

extensions/ql-vscode/src/cli.ts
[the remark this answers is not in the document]
adityasharad: Guess not! TIL. Thanks for checking :)

extensions/ql-vscode/package.json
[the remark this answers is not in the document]
adityasharad: I appreciate the data-driven decision. Much better than the random choice I was going to suggest!

extensions/ql-vscode/src/config.ts
adityasharad: Not a silly question, I wondered the same while looking at it. I think the intent is to test a range of values that we expect to be handled: for this kind of test it's a good idea to check the minimum, the maximum, some common value in the middle, and a zero/empty value. If the test is capable of handling errors, then also test some value that shouldn't be allowed.

shati-patel: It looks like the test only lets me test 2 values? https://github.com/github/vscode-codeql/blob/e119218828d5106b0fa2d3547be8208ebeb93938/extensions/ql-vscode/src/vscode-tests/minimal-workspace/config.test.ts#L34-L36

I can't quite work out if the test is doing anything with the values themselves, or just listening for when they change 😕

adityasharad: Ohh you are right. The test is just making sure the listeners handle a change. So `[0, 1]` sounds fine.
We can look separately at creating an integration test that produces multiple paths - right now we have none so there's nothing to update.

adityasharad (approved):

PR #937, opened 2021-08-31
Tidy and add tests for `resolveLanguages` commands

shati-patel:
First stage of tidying up and adding tests. See https://github.com/github/vscode-codeql/pull/915#pullrequestreview-720517460 and the linked (internal) issue.

- I've moved `findLanguage` out of "run-remote-query" and into a more general place ("helpers"), since it's used in a few functions and this seemed like a more logical place
- Added a test for `resolveLanguages` (basically copied from the `resolveQlpacks` test above 😅) 

I haven't yet added a test for the other functions from https://github.com/github/vscode-codeql/pull/915, since they require a bit more setup. Stay tuned for a follow-up PR 😎 

## Checklist

_N/A: no user-facing changes_

### DIFF
--- a/extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
@@ -52,4 +52,15 @@ describe('Use cli', function() {
     expect(qlpacks['codeql-javascript']).not.to.be.undefined;
     expect(qlpacks['codeql-python']).not.to.be.undefined;
   });
+
+  it('should resolve languages', async function() {
+    skipIfNoCodeQL(this);
+    const languages = await cli.resolveLanguages();
+    // should have a bunch of languages. just check that a few known ones exist
+    expect(languages['cpp']).not.to.be.undefined;
+    expect(languages['csharp']).not.to.be.undefined;
+    expect(languages['java']).not.to.be.undefined;
+    expect(languages['javascript']).not.to.be.undefined;
+    expect(languages['python']).not.to.be.undefined;
+  });
 });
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
```diff
@@ -52,4 +52,15 @@ describe('Use cli', function() {
...
+
+  it('should resolve languages', async function() {
+    skipIfNoCodeQL(this);
+    const languages = await cli.resolveLanguages();
+    // should have a bunch of languages. just check that a few known ones exist
+    expect(languages['cpp']).not.to.be.undefined;
+    expect(languages['csharp']).not.to.be.undefined;
+    expect(languages['java']).not.to.be.undefined;
+    expect(languages['javascript']).not.to.be.undefined;
+    expect(languages['python']).not.to.be.undefined;
```
adityasharad, line 64: Reading the [Chai docs](https://www.chaijs.com/api/bdd/#method_property), I think there's an equivalent way of writing what we had before without just crashing if the key is missing. Does this work?
```ts
for (const expectedLanguage of ['cpp', 'csharp', 'go', 'java', 'javascript', 'python']) {
  expect(languages).to.have.property(expectedLanguage).that.is.not.undefined;
}
```
I also added Go since we can expect that to exist, might as well test all supported languages.

adityasharad (approved):
Nice! One minor suggestion that I thought of from reading the existing code and the Chai docs. There are of course many equivalent ways of writing the same test, so pick whatever you think is clear and readable.

extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
shati-patel: Thank you, that looks nicer! (and also works 😄)

I'll merge this and start looking at more interesting tests in a follow-up PR.

PR #940, opened 2021-09-02
Fix dependabot warning

aeisenberg:


shati-patel (approved):
Thanks! Out of interest, how did you update this? 😅 Something like  `npm install path-parse@1.0.7`?

PR #946, opened 2021-09-07
Offer option to rerun remote query on the subset of valid repositories

robertbrignull:
Paired with @shati-patel to write this 🍐 

When a user tries to run a remote CodeQL query (Internal feature only) but it turns out that a subset of the repositories chosen are invalid, offer them the option to rerun their query but only on the subset of valid repositories. The full set of invalid repositories and reasons are included in the logs. It would be nice to also include this information in the popup message but we were struggling to find a convenient or good way to do this.

## Checklist

No changelog needed as this is an internal feature.

### DIFF
--- a/extensions/ql-vscode/src/run-remote-query.ts
+++ b/extensions/ql-vscode/src/run-remote-query.ts
@@ -136,16 +136,16 @@ async function runRemoteQueriesApiRequest(credentials: Credentials, ref: string,
       const invalidRepos = error?.response?.data?.invalid_repos || [];
       const reposWithoutDbUploads = error?.response?.data?.repos_without_db_uploads || [];
       void logger.log('Unable to run query on all repositories');
-      void logger.log('invalidRepos = ' + JSON.stringify(invalidRepos));
-      void logger.log('reposWithoutDbUploads = ' + JSON.stringify(reposWithoutDbUploads));
+      void logger.log('Invalid repos: ' + invalidRepos.join(', '));
+      void logger.log('Repos without DB uploads: ' + reposWithoutDbUploads.join(', '));
 
       if (invalidRepos.length + reposWithoutDbUploads.length === repositories.length) {
         // Every repo is invalid in some way
-        void showAndLogErrorMessage('Unable to run query on the specified repositories.');
+        void showAndLogErrorMessage('Unable to run query on any of the specified repositories.');
         return;
       }
 
-      const popupMessage = 'Unable to run query on the specified repositories. See logs for more details.';
+      const popupMessage = 'Unable to run query on some of the specified repositories. See logs for more details.';
       const rerunQuery = await showInformationMessageWithAction(popupMessage, 'Rerun on the valid repositories only');
       if (rerunQuery) {
         const validRepositories = repositories.filter(r => !invalidRepos.includes(r) && !reposWithoutDbUploads.includes(r));
--- a/extensions/ql-vscode/src/run-remote-query.ts
+++ b/extensions/ql-vscode/src/run-remote-query.ts
@@ -135,9 +135,13 @@ async function runRemoteQueriesApiRequest(credentials: Credentials, ref: string,
     if (typeof error.message === 'string' && error.message.includes('Some repositories were invalid')) {
       const invalidRepos = error?.response?.data?.invalid_repos || [];
       const reposWithoutDbUploads = error?.response?.data?.repos_without_db_uploads || [];
-      void logger.log('Unable to run query on all repositories');
-      void logger.log('Invalid repos: ' + invalidRepos.join(', '));
-      void logger.log('Repos without DB uploads: ' + reposWithoutDbUploads.join(', '));
+      void logger.log('Unable to run query on some of the specified repositories');
+      if (invalidRepos.length > 0) {
+        void logger.log('Invalid repos: ' + invalidRepos.join(', '));
+      }
+      if (reposWithoutDbUploads.length > 0) {
+        void logger.log('Repos without DB uploads: ' + reposWithoutDbUploads.join(', '));
+      }
 
       if (invalidRepos.length + reposWithoutDbUploads.length === repositories.length) {
         // Every repo is invalid in some way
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -128,6 +132,29 @@ export async function runRemoteQuery(cliServer: cli.CodeQLCliServer, credentials
...
+      void logger.log('invalidRepos = ' + JSON.stringify(invalidRepos));
+      void logger.log('reposWithoutDbUploads = ' + JSON.stringify(reposWithoutDbUploads));
+
+      if (invalidRepos.length + reposWithoutDbUploads.length === repositories.length) {
+        // Every repo is invalid in some way
+        void showAndLogErrorMessage('Unable to run query on the specified repositories.');
+        return;
+      }
+
+      const popupMessage = 'Unable to run query on the specified repositories. See logs for more details.';
```
shati-patel, line 148: @aeisenberg (or anyone else!) do you know if there's a way to link to the Extension Log from a message? It would be neat if "See logs" was a clickable link 🖱️

shati-patel (commented):
Thanks for opening the PR! This looks good to me, but I'll let someone else review since I helped write this 😄

(PS: Note to self/reviewer; we don't have any tests yet for the `run-remote-query` module. I have a separate issue for that!)

extensions/ql-vscode/src/run-remote-query.ts
aeisenberg: We do this in `internalShowAndLog`, added as a button to the dialog box. In theory, you can add multiple buttons and handle different actions for each button (though that's not _exactly_ what you are asking).

https://github.com/github/vscode-codeql/blob/7e78a6bc5c852908e5284eb71ed15170a0c28156/extensions/ql-vscode/src/helpers.ts#L72-L72

I think markdown is accepted in these dialog boxes, so you _might_ be able to use a [command uri(https://code.visualstudio.com/api/extension-guides/command#command-uris). But, to do that I thnk you will need to create a command that will open the extension log (I don't think there's a builtin command that does it).

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -128,6 +132,29 @@ export async function runRemoteQuery(cliServer: cli.CodeQLCliServer, credentials
     void showAndLogInformationMessage(`Successfully scheduled runs. [Click here to see the progress](https://github.com/${OWNER}/${REPO}/actions).`);
 
   } catch (error) {
-    void showAndLogErrorMessage(error);
+    if (typeof error.message === 'string' && error.message.includes('Some repositories were invalid')) {
+      const invalidRepos = error?.response?.data?.invalid_repos || [];
+      const reposWithoutDbUploads = error?.response?.data?.repos_without_db_uploads || [];
+      void logger.log('Unable to run query on all repositories');
+      void logger.log('invalidRepos = ' + JSON.stringify(invalidRepos));
```
aeisenberg, line 139: This will look strange if `invalidRepos` is an empty array, so maybe used different text if so. Also, I generally prefer to use template strings instead of concatenation. eg-

```suggestion
      void logger.log(`invalidRepos = ${JSON.stringify(invalidRepos)}`);
```

but maybe something more text friendly would be preferable?

```suggestion
      void logger.log(`InvalidRepos: ${invalidRepos.join(', '}`);
```

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -128,6 +132,29 @@ export async function runRemoteQuery(cliServer: cli.CodeQLCliServer, credentials
...
+      void logger.log('reposWithoutDbUploads = ' + JSON.stringify(reposWithoutDbUploads));
```
aeisenberg, line 140: Same comment as above.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -128,6 +132,29 @@ export async function runRemoteQuery(cliServer: cli.CodeQLCliServer, credentials
...
+
+      if (invalidRepos.length + reposWithoutDbUploads.length === repositories.length) {
+        // Every repo is invalid in some way
+        void showAndLogErrorMessage('Unable to run query on the specified repositories.');
```
aeisenberg, line 144:
```suggestion
        void showAndLogErrorMessage('Unable to run query on any of the specified repositories.');
```

extensions/ql-vscode/src/run-remote-query.ts
aeisenberg:
```suggestion
      const popupMessage = 'Unable to run query on some of the specified repositories. See logs for more details.';
```

robertbrignull: The problem we were having with `internalShowAndLog` and similar methods is that clicking any of the options closes the message. It would be nice if you could click to see more info and then come back to decide whether to re-run your query or not.

The command URIs look pretty interesting and could do what we want. Still haven't quite decided what that is though.

robertbrignull:
I've pushed updates for the simple comments. I'll have a look at command URIs now.

robertbrignull:
Turns out adding a "show logs" command was pretty easy. Here's what it looks like now
Screenshot 2021-09-08 at 12 48 48

I also adjusted the formatting on another log message to make it more consistent.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -128,6 +132,33 @@ export async function runRemoteQuery(cliServer: cli.CodeQLCliServer, credentials
     void showAndLogInformationMessage(`Successfully scheduled runs. [Click here to see the progress](https://github.com/${OWNER}/${REPO}/actions).`);
 
   } catch (error) {
-    void showAndLogErrorMessage(error);
+    if (typeof error.message === 'string' && error.message.includes('Some repositories were invalid')) {
+      const invalidRepos = error?.response?.data?.invalid_repos || [];
+      const reposWithoutDbUploads = error?.response?.data?.repos_without_db_uploads || [];
+      void logger.log('Unable to run query on some of the specified repositories');
+      if (invalidRepos.length > 0) {
+        void logger.log('Invalid repos: ' + invalidRepos.join(', '));
```
shati-patel, line 140: As Andrew mentioned, we mostly use template literals:
```suggestion
        void logger.log(`Invalid repos: ${invalidRepos.join(', ')}`);
```

(also applies to the other strings below 🙂)

shati-patel (approved):
The "open log" link is very cool 😎 I left a few tiny comments, but otherwise this seems good! 

The tests should be fixed now too (🤞🏽), so you might want to rebase/update from `main`.

extensions/ql-vscode/src/run-remote-query.ts
robertbrignull: Thanks. I changed to backticks but forgot to use the templating 🤦

robertbrignull:
Also now merged with latest main. Let's hope the tests are ok now.

aeisenberg:
Was just going to mention that the failing tests are  because the qlpack names have changed.

PR #947, opened 2021-09-07
Wait for document to be saved before running query

hmac:
If `AUTOSAVE_SETTING` is `true` and I trigger the `Run Query` command, I occasionally see race conditions where the query is run before the file is saved, and I see results corresponding to the previous version of the query. I'm not sure how to test this and I haven't been able to reliably reproduce the problem (though it occurs often).

Intuitively, it seems like this change should fix it, since [`editor.document.save`](https://github.com/microsoft/vscode/blob/main/src/vs/vscode.d.ts#L143) returns a `Thenable`, which I think means it returns a promise or the function is async.

My javascript is very rusty, sorry if this is inaccurate!

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Fix bug where a query is sometimes run before the file is saved. [#947](https://github.com/github/vscode-codeql/pull/947)
+
 ## 1.5.4 - 02 September 2021
 
 - Add support for filename pattern in history view. [#930](https://github.com/github/vscode-codeql/pull/930)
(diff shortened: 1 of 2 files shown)


shati-patel (approved):
Thanks for the PR, @hmac! This looks good 🎉 

It's technically a user-facing change, so you could add an entry to the [CHANGELOG](https://github.com/github/vscode-codeql/blob/hmac-wait-for-save/extensions/ql-vscode/CHANGELOG.md) (just under the "unreleased" header). Maybe something like
```
- Fix bug where a query is sometimes run before the file is saved. [#947](https://github.com/github/vscode-codeql/pull/947)
```

hmac:
@shati-patel done - thanks for the quick review!

PR #949, opened 2021-09-08
Fix AST Viewer

aeisenberg:
The previous synthetic query suite was not finding the ast query because
the `qlpack` directive in a query suite only matches queries from the
default suite, which `printAST.ql` is not part of.

This changes to using `from` and `queries` directives.

Also, adds an integration test to ensure we find the queries using
different CLIs. However, this only tests using the latest `main` from
the codeql repository. I wonder if we should start testing using
different versions of the repo.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Fix bug where a query is sometimes run before the file is saved. [#947](https://github.com/github/vscode-codeql/pull/947)
+- Fix broken contextual queries, including _View AST_. [#949](https://github.com/github/vscode-codeql/pull/949)
 
 ## 1.5.4 - 02 September 2021
 
--- a/extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
@@ -6,7 +6,7 @@ import { SemVer } from 'semver';
 import { CodeQLCliServer, QueryInfoByLanguage } from '../../cli';
 import { CodeQLExtensionInterface } from '../../extension';
 import { skipIfNoCodeQL } from '../ensureCli';
-import { getOnDiskWorkspaceFolders } from '../../helpers';
+import { getOnDiskWorkspaceFolders, getQlPackForDbscheme, languageToDbScheme } from '../../helpers';
 import { resolveQueries } from '../../contextual/queryResolver';
 import { KeyType } from '../../contextual/keyType';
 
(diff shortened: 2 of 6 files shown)


aeisenberg:
We will probably need another release for this.

Also, I expect that some of the integration tests will fail, but not sure which ones. Let's see how far back we can still support print asts with the new formulation.

extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
```diff
@@ -68,4 +70,16 @@ describe('Use cli', function() {
     const queryInfo: QueryInfoByLanguage = await cli.resolveQueryByLanguage(getOnDiskWorkspaceFolders(), Uri.file(queryPath));
     expect((Object.keys(queryInfo.byLanguage))[0]).to.eql('javascript');
   });
+
+  it.only('should resolve printAST queries', async function() {
```
adityasharad, line 74: Should we make this loop over all languages? The filenames may differ.

extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
```diff
@@ -68,4 +70,16 @@ describe('Use cli', function() {
...
+    skipIfNoCodeQL(this);
+
+    const result = await resolveQueries(cli, {
+      dbschemePack: 'codeql/javascript-all',
```
adityasharad, line 78: How do these tests get the query repo checkout? If we're splitting over multiple CLI versions, we should use the corresponding `codeql-cli/version` tags from the query repos, as well as latest `main` for the last released CLI.

extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
aeisenberg: Good idea. I will change that.

extensions/ql-vscode/src/vscode-tests/cli-integration/run-cli.test.ts
aeisenberg: As discussed, the integration tests checks out the expected codeql version for each CLI version.

aeisenberg:
Works for all languages in `github/codeql` across all supported versions of the CLI. We only test that the core libraries version released with that version of the CLI are correct. We _could_ add more matrix builds to test different versions of the CLI with different core libraries.

Also, not yet tested are go and ruby since these languages exist in other repos that are not yet being downloaded by the integration tests.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Fix bug where a query is sometimes run before the file is saved. [#947](https://github.com/github/vscode-codeql/pull/947)
+- Fix contextual queries that are broken due to a malformed synthetic query suite. [#949](https://github.com/github/vscode-codeql/pull/949)
```
adityasharad, line 6: Mention the AST viewer here.

adityasharad (approved):
Nice. The missing test combination we discussed is latest stable CLI + `github/codeql:main`, but we can look into adding that separately.

dbartol (approved):
LGTM

PR #956, opened 2021-09-28
Add option to specify a different "controller" repo

shati-patel:
[Internal use only]

Instead of hard-coding the controller repo for viewing remote queries, let users specify a custom one in the settings. This works because we can use a dynamic workflow instead of a workflow file checked in to a controller repo. (Currently behind a feature flag - see linked issue or let me know if you need access for reviewing!) If there's no repo defined in the config, prompt user to enter it in a textbox.

Note: Slightly tangential, but I also renamed the `remoteRepositoryLists` setting and grouped everything related to remote queries under a `codeQL.remoteQueries` category. I'll update the internal instructions with the renamed setting.

## Checklist

N/A - no user-facing changes

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -241,7 +241,7 @@
           "scope": "application",
           "description": "Specifies whether or not to write telemetry events to the extension log."
         },
-        "codeQL.remoteRepositoryLists": {
+        "codeQL.remoteQueries.repositoryLists": {
           "type": [
             "object",
             null
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -256,6 +256,13 @@
           },
           "default": null,
           "markdownDescription": "[For internal use only] Lists of GitHub repositories that you want to query remotely. This should be a JSON object where each key is a user-specified name for this repository list, and the value is an array of GitHub repositories (of the form `<owner>/<repo>`)."
+        },
+        "codeQL.remoteQueries.controllerRepo": {
+          "type": "string",
+          "default": "",
+          "pattern": "^$|^(?:[a-zA-Z0-9]+-)*[a-zA-Z0-9]+/[a-zA-Z0-9-_]+$",
+          "patternErrorMessage": "Please enter a valid GitHub repository",
+          "markdownDescription": "[For internal use only] The name of the GitHub repository where you can view the progress and results of the \"Run Remote query\" command. The repository should be of the form `<owner>/<repo>`)."
         }
       }
     },
--- a/extensions/ql-vscode/src/run-remote-query.ts
+++ b/extensions/ql-vscode/src/run-remote-query.ts
@@ -108,8 +108,8 @@ export async function runRemoteQuery(cliServer: cli.CodeQLCliServer, credentials
   // If it doesn't exist, prompt the user to enter it, and save that value to the config.
   let controllerRepo: string | undefined;
   controllerRepo = getRemoteControllerRepo();
-  if (!controllerRepo) {
-    void logger.log('No controller repository defined.');
+  if (!controllerRepo || !REPO_REGEX.test(controllerRepo)) {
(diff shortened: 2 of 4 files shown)


extensions/ql-vscode/package.json
```diff
@@ -256,6 +256,13 @@
           },
           "default": null,
           "markdownDescription": "[For internal use only] Lists of GitHub repositories that you want to query remotely. This should be a JSON object where each key is a user-specified name for this repository list, and the value is an array of GitHub repositories (of the form `<owner>/<repo>`)."
+        },
+        "codeQL.remoteQueries.controllerRepo": {
+          "type": "string",
+          "default": "",
+          "pattern": "^$|^(?:[a-zA-Z0-9]+-)*[a-zA-Z0-9]+/[a-zA-Z0-9-_]+$",
```
shati-patel, line 263: Copied this from the tried and tested "repo regex" in run-remote-query 😎

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -107,18 +104,44 @@ export async function runRemoteQuery(cliServer: cli.CodeQLCliServer, credentials
...
+    });
+    if (!controllerRepo) {
+      void showAndLogErrorMessage('No controller repository entered.');
+      return;
+    } else if (!REPO_REGEX.test(controllerRepo)) { // Check if user entered invalid input
+      void showAndLogErrorMessage('Invalid repository format. Must be a valid GitHub repository in the format <owner>/<repo>.');
+      return;
+    }
+    void logger.log(`Setting the controller repository as: ${controllerRepo}`);
+    await setRemoteControllerRepo(controllerRepo);
```
shati-patel, line 127: I added this to avoid users having to re-enter the controller repo each time. Not sure if this is the right thing to do, but I don't expect people will need to switch between controller repos regularly...

robertbrignull (commented):
Changes LGTM (though I'll let someone else review too). Thanks for renaming things too as that's much neater.

aeisenberg (commented):
Looks generally good. Just one bit of clarification.

extensions/ql-vscode/src/run-remote-query.ts
shati-patel: What I think this does (esp now that I've pushed a small fix):

- If there's a valid repo in the `codeQL.remoteQueries.controllerRepo` setting, use that
- If not, get the user to enter it in a text box, and set that as `codeQL.remoteQueries.controllerRepo` (so they don't have to re-enter it next time)

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -107,18 +104,44 @@ export async function runRemoteQuery(cliServer: cli.CodeQLCliServer, credentials
     return; // No error message needed, since `getRepositories` already displays one.
   }
 
-  await runRemoteQueriesApiRequest(credentials, ref, language, repositories, query);
+  // Get the controller repo from the config, if it exists.
+  // If it doesn't exist, prompt the user to enter it, and save that value to the config.
+  let controllerRepo: string | undefined;
+  controllerRepo = getRemoteControllerRepo();
+  if (!controllerRepo || !REPO_REGEX.test(controllerRepo)) {
+    void logger.log('No controller repository defined.');
```
aeisenberg, line 112: Should probably have a different error message if `REPO_REGEX.test` fails. Similar message to what you have below.

Maybe something like this:

```suggestion
    void logger.log(controllerRepo ? 'Invalid repository name.' : 'No controller repository defined.');
```

extensions/ql-vscode/src/run-remote-query.ts
aeisenberg: Ah...that makes sense now. Thanks.

extensions/ql-vscode/src/run-remote-query.ts
shati-patel: Good point, thanks! 😎

aeisenberg (approved):
Woohoo!

Issue #894, opened 2021-06-29
LGTM database download progress popup is misleading when not expanded


rdmarsh2:
**Describe the bug**
After starting a database download, when not expanded, the text in the progress popup is `Adding database from LGTM. Choose a language from the d...`, and does not change as progress is made. The text that changes is later in the message, and only visible when the popup is expanded.

**Version**
CodeQL extension version: 1.5.1 
CodeQL CLI version: 2.5.5 
Platform: win32 x64

Version: 1.57.1 (user setup)
Commit: 507ce72a4466fbb27b715c3722558bb15afa9f48
Date: 2021-06-17T13:28:07.755Z
Electron: 12.0.7
Chrome: 89.0.4389.128
Node.js: 14.16.0
V8: 8.9.255.25-electron.0
OS: Windows_NT x64 10.0.19041

**To reproduce**
Start an LGTM database download from the CodeQL sidebar.
Select a language, if needed.
Observe the message in the popup (in my setup, it is unexpanded when it appears)

**Expected behavior**
The unchanging part of the message is short enough that the status (awaiting language choice, downloading, or unzipping) is visible in the unexpanded popup.

**Additional context**
Unexpanded popup:
![image](https://user-images.githubusercontent.com/1377042/123861740-63944c00-d8dc-11eb-93c2-7355fef999e9.png)
Expanded popup:
![image](https://user-images.githubusercontent.com/1377042/123862204-e917fc00-d8dc-11eb-8170-759f6115996d.png)

PR #957, opened 2021-09-30
Remove line about selecting a language from the dropdown.

marcnjaramillo:
Closes [#894](https://github.com/github/vscode-codeql/issues/894)

**Changes made**: removed the line about selecting a language from the dropdown, which now makes the download progress visible even when the popup is collapsed.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases-ui.ts
+++ b/extensions/ql-vscode/src/databases-ui.ts
@@ -295,7 +295,7 @@ export class DatabaseUI extends DisposableObject {
         'codeQLDatabases.chooseDatabaseLgtm',
         this.handleChooseDatabaseLgtm,
         {
-          title: 'Adding database from LGTM. Choose a language from the dropdown, if requested.',
+          title: 'Adding database from LGTM',
         })
     );
     this.push(
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -771,7 +771,7 @@ async function activateWithInstalledDistribution(
     ) =>
       databaseUI.handleChooseDatabaseLgtm(progress, token),
       {
-        title: 'Adding database from LGTM. Choose a language from the dropdown, if requested.',
+        title: 'Adding database from LGTM',
       })
   );
   ctx.subscriptions.push(
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Remove line about selecting a language from the dropdown when downloading database from LGTM. This makes the download progress visible when the popup is not expanded. [#894](https://github.com/github/vscode-codeql/issues/894)
+
 ## 1.5.5 - 08 September 2021
 
 - Fix bug where a query is sometimes run before the file is saved. [#947](https://github.com/github/vscode-codeql/pull/947)


extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -295,7 +295,7 @@ export class DatabaseUI extends DisposableObject {
         'codeQLDatabases.chooseDatabaseLgtm',
         this.handleChooseDatabaseLgtm,
         {
-          title: 'Adding database from LGTM. Choose a language from the dropdown, if requested.',
+          title: 'Adding database from LGTM.',
```
adityasharad, line 298: We can drop the full stop too, since the progress popup message always adds a colon at the end of the title. Same suggestion below.
```suggestion
          title: 'Adding database from LGTM',
```

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -295,7 +295,7 @@ export class DatabaseUI extends DisposableObject {
         'codeQLDatabases.chooseDatabaseLgtm',
         this.handleChooseDatabaseLgtm,
```
adityasharad, line 296: Optional:
Here is a more involved suggestion for you to try out, which I think would work nicely. Consider it an option for further exploration: you can choose whether to include it as part of this PR, or complete this PR and try this suggestion in a follow-up PR.

If we look at the `handleChooseDatabaseLgtm` function, you can see it has a parameter `progress: ProgressCallback`. Progress callbacks work by allowing you to provide a progress message and step number, e.g. the code
```ts
progress({
  message: 'Choose project',
  step: 1,
  maxStep: 2
})
```
will update the progress notification box with the text `title: message` (e.g. "Adding database from LGTM: Choose project") and a 50% complete progress bar.

Instead of having the "Choose language" text in the title, which you've correctly removed in this PR, I think we could add two `progress()` calls -- one that says "Choose project" and one that says "Choose language". Then the popup box exactly describes the action we are asking the user to carry out in the UI at the top of the screen.

You may need to experiment to choose the best place to put those two progress calls. I think the function `promptImportLgtmDatabase` is a good place for the first one, and `promptForLanguage` is a good place for the second one. Both are reached from `handleChooseDatabaseLgtm`. The second one will require passing a `progress` parameter through to a few more functions.

extensions/ql-vscode/src/extension.ts
```diff
@@ -771,7 +771,7 @@ async function activateWithInstalledDistribution(
     ) =>
       databaseUI.handleChooseDatabaseLgtm(progress, token),
       {
-        title: 'Adding database from LGTM. Choose a language from the dropdown, if requested.',
+        title: 'Adding database from LGTM.',
```
adityasharad, line 774:
```suggestion
        title: 'Adding database from LGTM',
```

adityasharad (commented):
Nice, thanks for picking this up so quickly!
Some very small text suggestions, and a larger optional suggestion that would be interesting, but doesn't have to be in this PR.

extensions/ql-vscode/src/extension.ts
marcnjaramillo: Thanks for the feedback. I was wondering about whether to keep this in as well. I think it looks better without it.

extensions/ql-vscode/src/databases-ui.ts
marcnjaramillo: I like this idea! I think for now I'll complete this PR so we can at least have some small improvement, and then I'll spend some time this afternoon and tomorrow working on it.

adityasharad (approved):

aeisenberg (commented):
Nice work! Could you also add a changelog message here? See how the other messages are written in the CHANGELOG.md file.

marcnjaramillo:
@aeisenberg Yes, I will do that. Sorry I missed it the first time.

aeisenberg (approved):

PR #959, opened 2021-10-01
Delete database after removing it from query server control.

alexet:
We need to remove it from the query server before deleting the files it otherwise the query server may access it prior to deletion.


aeisenberg (approved):
Good catch. This makes sense.

Can you also add a changelog entry?

PR #961, opened 2021-10-04
Case insensitive fallback check for github repositories

### LINKED ISSUE #892: Download from LGTM command requires a case-sensitive slug
**Describe the bug**

Using the _Download from LGTM_ command will fail if you enter a URL or slug that does not have the right case. 

https://lgtm.com/projects/g/BOINC/boinc/

**To reproduce**

1. Run _Download from LGTM_
2. Enter the URL https://lgtm.com/projects/g/boinc/boinc/ (Notice that the correct URL is https://lgtm.com/projects/g/BOINC/boinc/)
3. 404 error

**Expected behavior**
Should have downloaded and installed the database.

**Additional context**
It is likely that this is an LGTM issue where it is not able to locate repos case-insensitively. Suggestion from @p0:

> For the `/g/`  provider, you could query github for the canonical case and pass it to LGTM (perhaps only after a 404).

mgsium:
Implements a case-insensitive fallback check for the `/g/` provider where a repository is not found on LGTM using a case-sensitive search. Partially fixes #892 

## Checklist


extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -107,6 +107,15 @@ export async function promptImportLgtmDatabase(
   return;
 }
 
+export async function retrieveCanonicalRepoName(lgtmUrl: string) {
+  const givenRepoName = parseLgtmUrl(lgtmUrl);
+  const repo = await fetch(`https://api.github.com/repos/${givenRepoName}`).then(res => res.json());
```
aeisenberg, line 112: What happens if there is an error here and something besides `200` is returned? Can you take a look at `checkForFailingResponse` in this file? It would be best to to generalize this function to handle this case as well. You could add a new parameter that would replace the error message `Error downloading database.`.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -409,23 +418,31 @@ function convertRawLgtmSlug(maybeSlug: string): string | undefined {
   return;
 }
 
+function parseLgtmUrl(lgtmUrl: string): string | undefined {
+  const re = new RegExp('https://lgtm.com/projects/(g|gl|b|git)/(.*)');
```
aeisenberg, line 422: Only project slugs that begin with a `g` are from github (they are the vast majority of projects on lgtm. It may not be safe to match on other slugs. Eg- if two people put up the different projects with the same project name with different capitalization on github and bitbucket, you would get a mismatch. It's a rare case, but it's probably safer to protect against it.

aeisenberg: Also, please add a comment that this only supports github slugs.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -409,23 +418,31 @@ function convertRawLgtmSlug(maybeSlug: string): string | undefined {
...
-      uri.path.split('/').filter((segment) => segment)
-    ).slice(0, 6);
-    const projectUrl = `https://lgtm.com/${paths.join('/')}`;
-    const projectResponse = await fetch(projectUrl);
-    const projectJson = await projectResponse.json();
+    let projectJson = await pullLgtmProject(lgtmUrl);
 
     if (projectJson.code === 404) {
-      throw new Error();
+      // fallback check for repos with same name but different case
```
aeisenberg, line 439:
```suggestion
      // fallback check for github repos with same name but different case
```

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -409,23 +418,31 @@ function convertRawLgtmSlug(maybeSlug: string): string | undefined {
...
+      let canonicalName = await retrieveCanonicalRepoName(lgtmUrl);
+      if (!canonicalName) {
+        throw new Error();
+      }
+      canonicalName = convertRawLgtmSlug(`g/${canonicalName}`);
+      projectJson = await pullLgtmProject(canonicalName);
```
aeisenberg, line 445: You should check again if `projectJson.code === 404`.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -445,6 +462,16 @@ export async function convertToDatabaseUrl(
   }
 }
 
+async function pullLgtmProject(lgtmUrl: string): Promise<any> {
```
aeisenberg, line 465: `pull` generally has a different meaning when dealing with git operations. This is really about getting the project's metadata. 

```suggestion
async function downloadLgtmProjectMetadata(lgtmUrl: string): Promise<any> {
```

or something like that.

aeisenberg (commented):
Nice work. Some comments inline.

PR #985, opened 2021-10-27
Remote Queries: Create packs for remote queries

aeisenberg:
Remote Queries: Create packs for remote queries

This PR handles two cases:

1. There is a qlpack.yml or codeql-pack.yml file in the same directory
   as the query to run remotely. In this case, run `codeql pack
   packlist` to determine what files to include (and also always include
   the lock file and the query itself. Copy to a temp folder and run
   `pack install`, then `pack bundle`. Finally upload.
2. There is no qlpack in the current directory. Just copy the single
   file to the temp folder and generate a synthetic qlpack before
   installing, bundling and uploading.

Two cases that are not handled:

1. The query file is part of a workspace. Peer dependencies will not be
   found.
2. The query file and its qlpack file are not in the same directory.

These case should be possible to handle later. 

This PR also includes integration tests that build the pack and verify that its components are what we expect.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/run-remote-query.ts
+++ b/extensions/ql-vscode/src/run-remote-query.ts
@@ -119,6 +119,9 @@ async function generateQueryPack(cliServer: cli.CodeQLCliServer, queryFile: stri
         })
     });
 
+    // ensure the qlpack.yml has a valid name
+    await ensureQueryPackName(queryPackDir);
+
     void logger.log(`Copied ${copiedCount} files to ${queryPackDir}`);
 
     language = await findLanguage(cliServer, Uri.file(targetQueryFileName));
@@ -159,6 +162,25 @@ async function generateQueryPack(cliServer: cli.CodeQLCliServer, queryFile: stri
     base64Pack,
     language
   };
+}
+
+/**
+ * Ensure that the qlpack.yml has a valid name. For local purposes,
+ * Anonymous packs and names that are not prefixed by a scope (ie `<foo>/`)
+ * are sufficient. But in order to create a pack, the name must be prefixed.
+ *
+ * @param queryPackDir the directory containing the query pack.
+ */
+async function ensureQueryPackName(queryPackDir: string) {
+  const pack = yaml.safeLoad(await fs.readFile(path.join(queryPackDir, 'qlpack.yml'), 'utf8')) as { name: string; };
+  if (!pack.name || !pack.name.includes('/')) {
+    if (!pack.name) {
+      pack.name = 'codeql-remote/query';
+    } else if (!pack.name.includes('/')) {
+      pack.name = `codeql-remote/${pack.name}`;
+    }
+    await fs.writeFile(path.join(queryPackDir, 'qlpack.yml'), yaml.safeDump(pack));
+  }
 }
 
 async function createRemoteQueriesTempDirectory() {
--- a/extensions/ql-vscode/src/run-remote-query.ts
+++ b/extensions/ql-vscode/src/run-remote-query.ts
@@ -129,14 +129,14 @@ async function generateQueryPack(cliServer: cli.CodeQLCliServer, queryFile: stri
 
     // copy only the query file to the query pack directory
     // and generate a synthetic query pack
-    // TODO this has a limitation that query packs inside of a workspace will not resolve it's peer dependencies.
+    // TODO this has a limitation that query packs inside of a workspace will not resolve its peer dependencies.
     // Something to work on later. For now, we will only support query packs that are not in a workspace.
     void logger.log(`Copying ${queryFile} to ${queryPackDir}`);
     await fs.copy(queryFile, targetQueryFileName);
     void logger.log('Generating synthetic query pack');
     const syntheticQueryPack = {
       name: 'codeql-remote/query',
-      version: '1.0.0',
+      version: '0.0.0',
       dependencies: {
         [`codeql/${language}-all`]: '*',
       }
--- a/extensions/ql-vscode/src/run-remote-query.ts
+++ b/extensions/ql-vscode/src/run-remote-query.ts
@@ -1,13 +1,15 @@
-import { QuickPickItem, Uri, window } from 'vscode';
+import { CancellationToken, QuickPickItem, Uri, window } from 'vscode';
 import * as path from 'path';
 import * as yaml from 'js-yaml';
 import * as fs from 'fs-extra';
 import * as tmp from 'tmp-promise';
-import { findLanguage, showAndLogErrorMessage, showAndLogInformationMessage, showInformationMessageWithAction } from './helpers';
+import { askForLanguage, findLanguage, getOnDiskWorkspaceFolders, showAndLogErrorMessage, showAndLogInformationMessage, showInformationMessageWithAction } from './helpers';
 import { Credentials } from './authentication';
 import * as cli from './cli';
 import { logger } from './logging';
 import { getRemoteControllerRepo, getRemoteRepositoryLists, setRemoteControllerRepo } from './config';
+import { tmpDir } from './run-queries';
+import { ProgressCallback, UserCancellationException } from './commandRunner';
 interface Config {
   repositories: string[];
   ref?: string;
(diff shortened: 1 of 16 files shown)


rneatherway:
> 1. There is a qlpack.yml or codeql-pack.yml file in the same directory
>    as the query to run remotely.

How would we lift this restriction in the end? Search upwards for an enclosing directory containing a `qlpack.yml` file?

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -68,97 +72,291 @@ export async function getRepositories(): Promise<string[] | undefined> {
...
   } else {
-    ref = 'main';
-    [language, repositories] = await Promise.all([findLanguage(cliServer, uri), getRepositories()]);
-  }
+    // open popup to ask for language if not already hardcoded
+    language = fallbackLanguage || await askForLanguage(cliServer);
 
+    // copy only the query file to the query pack directory
+    // and generate a synthetic query pack
+    // TODO this has a limitation that query packs inside of a workspace will not resolve it's peer dependencies.
```
rneatherway, line 132:
```suggestion
    // TODO this has a limitation that query packs inside of a workspace will not resolve its peer dependencies.
```

Trivial nit

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -68,97 +72,291 @@ export async function getRepositories(): Promise<string[] | undefined> {
...
+          const normalizedPath = Uri.file(f).fsPath;
+          const matches = normalizedPath === file || normalizedPath.startsWith(file + path.sep);
+          if (matches) {
+            copiedCount++;
+          }
+          return matches;
+        })
+    });
 
-  // If the user has an explicit `.repositories` file, use that.
```
rneatherway, line 84: Does this PR remove support for `.repositories` entirely? Seems like a good change if so.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -68,97 +72,291 @@ export async function getRepositories(): Promise<string[] | undefined> {
...
-  // Otherwise, prompt user to select repositories from the `codeQL.remoteQueries.repositoryLists` setting.
-  if (await fs.pathExists(repositoriesFile)) {
-    void logger.log(`Found '${repositoriesFile}'. Using information from that file to run ${queryFile}.`);
+    void logger.log(`Copied ${copiedCount} files to ${queryPackDir}`);
 
-    const config = yaml.safeLoad(await fs.readFile(repositoriesFile, 'utf8')) as Config;
+    language = await findLanguage(cliServer, Uri.file(targetQueryFileName));
 
-    ref = config.ref || 'main';
```
rneatherway, line 91: It would be useful to be able to override `ref` at some point for testing purposes. We changed the hard-coded value to test it. Perhaps another non-advertised config setting?

(No change for this PR)

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -68,97 +72,291 @@ export async function getRepositories(): Promise<string[] | undefined> {
...
+    // copy only the query file to the query pack directory
+    // and generate a synthetic query pack
+    // TODO this has a limitation that query packs inside of a workspace will not resolve it's peer dependencies.
+    // Something to work on later. For now, we will only support query packs that are not in a workspace.
+    void logger.log(`Copying ${queryFile} to ${queryPackDir}`);
+    await fs.copy(queryFile, targetQueryFileName);
+    void logger.log('Generating synthetic query pack');
+    const syntheticQueryPack = {
+      name: 'codeql-remote/query',
+      version: '1.0.0',
```
rneatherway, line 139: Minor: would `0.0.0` be a better fake version number? (That's what we've been using up until now: https://github.com/dsp-testing/qc-run2/blob/ff207ffe5d100a9eb55743bc01f032336abdcde7/src/codeql.ts#L45)

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -68,97 +72,291 @@ export async function getRepositories(): Promise<string[] | undefined> {
...
+      dependencies: {
+        [`codeql/${language}-all`]: '*',
```
rneatherway, line 141: This part interests me. I take it this is the equivalent of https://github.com/dsp-testing/qc-run2/blob/ff207ffe5d100a9eb55743bc01f032336abdcde7/src/codeql.ts#L46? How does this behave in the following cases:

1. I've made a qlpack that doesn't depend on the core libraries (perhaps it is language independent). Odd case, I know.
2. I have a qlpack that contains just a query and the qlpack yml file. I've modified one of the core libraries, for example to add a new sink.
3. I have a qlpack that adds some new library files and also uses the core libraries.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -68,97 +72,291 @@ export async function getRepositories(): Promise<string[] | undefined> {
...
+    let fallbackLanguage: string | undefined;
+    let repositories: string[] | undefined;
+
+    progress({
+      maxStep: 5,
+      step: 2,
+      message: 'Determining query target language'
+    });
+
+    // If the user has an explicit `.repositories` file, use that.
```
rneatherway, line 217: Ah, we do still have support for `.repositories`. I think that would be completely fine to remove at this point, but perhaps we should discuss in our next meeting.

extensions/ql-vscode/src/run-remote-query.ts
```diff
@@ -68,97 +72,291 @@ export async function getRepositories(): Promise<string[] | undefined> {
...
+    postfix: 'generated.tgz',
+    prefix: 'qlpack',
+  });
+}
+
+export async function runRemoteQuery(
+  cliServer: cli.CodeQLCliServer,
+  credentials: Credentials,
+  uri: Uri | undefined,
+  dryRun: boolean,
```
rneatherway, line 183: As an alternative to an explicit option like this we have also used stubbing (e.g. https://github.com/github/codeql-action/blob/main/src/database-upload.test.ts), or you could pass in a mocked Octokit.

For the integration test I think it would be quite reasonable to call the API endpoint and run a query in its entirety. If it's against a one or a small number of repositories the cost is extremely small. If the results issue spam is annoying then you could use another repository as the controller.

All up to you.

shati-patel:
I didn't spot this in my earlier review, but I've just tried with a quick query, and it doesn't seem to work 🤔 I'm getting: `A fatal error occurred: Could not resolve library path ...`

Possibly because the qlpack.yml for my quick query uses the old format:
```
# This is an automatically generated file.

name: quick-query
version: 1.0.0
libraryPathDependencies:
  - codeql-go
```

Do quick queries work for you?

aeisenberg:
> > 1. There is a qlpack.yml or codeql-pack.yml file in the same directory
> >    as the query to run remotely.
> 
> How would we lift this restriction in the end? Search upwards for an enclosing directory containing a `qlpack.yml` file?

Yes. I don't think it will be hard to do, but I just want to focus on the simple parts first.

extensions/ql-vscode/src/run-remote-query.ts
aeisenberg: I've been finding it very useful to have it as an explicit parameter. This is mostly because I've been testing locally and I don't want to actually upload anything most of the time.

I was even toying with adding a `codeql.remote.dryRun` undocumented setting, but then I changed my mind.

For now, when I try uploading anything, even though the initial request is successful, the workflow run on the controller repo fails. Once this feature is stable, it may make sense to remove the dry-run flag altogether and actually upload something to run.

extensions/ql-vscode/src/run-remote-query.ts
aeisenberg: I think that's worth talking about. @shati-patel may have some opinions on this.

extensions/ql-vscode/src/run-remote-query.ts
aeisenberg: To be clear, this situation only exists when there is no qlpack associated with this query. So...

1. As long as there is a qlpack.yml that can be resolved and correctly declares itself to be no-lang, it should be fine.
2. I don't think we can (or should?) support this. The `codeql pack create` command compiles the current pack and also grabs all of its dependencies from the package cache, which get copied into the pack. Users are not supposed to manually edit these dependencies (it would be like manually editing your `node_modules`). If this is something that a user feels they need to do, then that's a missing feature we need to work on.
3. This is already handled by `codeql pack create`.

extensions/ql-vscode/src/run-remote-query.ts
aeisenberg: This can be overridden in the .repositories file. Maybe this is a reason to keep it.

aeisenberg:
> Do quick queries work for you?

Hmm...let me check this...

Actually, it's for a trivial reason. The qlpack.yml file is missing a scope and `pack create` won't work with scopeless packages. I can check for this and make sure to add a synthetic scope before trying to upload.

extensions/ql-vscode/src/run-remote-query.ts
rneatherway: > 2\. I don't think we can (or should?) support this. The `codeql pack create` command compiles the current pack and also grabs all of its dependencies from the package cache, which get copied into the pack. Users are not supposed to manually edit these dependencies (it would be like manually editing your `node_modules`). If this is something that a user feels they need to do, then that's a missing feature we need to work on.

That's fine for this PR but I think we are missing an important use-case where a CodeQL developer is working on a new or updated query that also requires changes to the core language libraries. I'll add it to the agenda for discussion.

rneatherway:
My comments are now limited to two discussion points for a future meeting, so I'm happy once @shati-patel is!

extensions/ql-vscode/src/run-remote-query.ts
rneatherway: > Once this feature is stable, it may make sense to remove the dry-run flag altogether and actually upload something to run.

_Hopefully_ it should all work now as we shipped the server-side changes yesterday.

shati-patel:
Thanks 🌞 PR looks good, apart from quick queries 👇🏽 

> > Do quick queries work for you?
> 
> Hmm...let me check this...
> 
> Actually, it's for a trivial reason. The qlpack.yml file is missing a scope and `pack create` won't work with scopeless packages. I can check for this and make sure to add a synthetic scope before trying to upload.

Quick queries still aren't working for me (though possibly for a different reason now 🙃):

```
[2021-11-02 13:24:55] Plumbing command codeql pack resolve-dependencies completed:
                      { }
[2021-11-02 13:24:55] [PROGRESS] pack install> Dependencies resolved. Installing packages...
...
[2021-11-02 13:24:55] [PROGRESS] pack install> Nothing to install.
```
Can't seems to resolve any dependencies or install any packages... Which then causes other stuff to fail further down the line (like` [ERROR] resolve library-path> ERROR: Referenced pack 'codeql-go' not found.`)

aeisenberg:
@shati-patel, it's working for me out of the box, but if I remove the codeql-go submodule from the workspace, then I get the error you mention. Is this what you are seeing?

extensions/ql-vscode/src/run-remote-query.ts
aeisenberg: Hmmm...yes, something to discuss since I don't think we should be encouraging users to change the core libraries. Maybe there's a use case I'm missing.

But, even so, when we run the `pack create` command, all paths in the vscode workspace are added to the `--additional-packs` path, so if the core libraries are in the workspace, they _should_ be used instead of whatever is published in the package registry (I haven't tried this out yet, so not sure).

aeisenberg:
With some help from @shati-patel, I worked out what the issue was. It's a bug in the CLI and I pushed up a private PR to fix. Basically, old style package name references like `codeql-go` (instead of `codeql/go-all`) were not being translated into a correct path. The fix will go out with v2.7.1 of the CLI and I don't think it should block this PR since they are released independently.

shati-patel (approved):
Thanks for fixing that! All good now with the updated CLI 🎉

aeisenberg:
Thanks for the review and trying this out!

PR #988, opened 2021-10-31
Add Generate Qlhelp preview command

mgsium:
Adds a "Generate Query Help" command which generates a .md file from a selected .qhelp file and opens it for preview in vscode. Implements [#481](https://github.com/github/codeql-coreql-team/issues/481)

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fix a bug with importing large databases. Databases over 4GB can now be imported directly from LGTM or from a zip file. This functionality is only available when using CodeQL CLI version 2.6.0 or later. [#971](https://github.com/github/vscode-codeql/pull/971)
 - Replace certain control codes (`U+0000` - `U+001F`) with their corresponding control labels (`U+2400` - `U+241F`)  in the results view. [#963](https://github.com/github/vscode-codeql/pull/963)
 - Allow case-insensitive project slugs for GitHub repositories when adding a CodeQL database from LGTM. [#978](https://github.com/github/vscode-codeql/pull/961)
+- Add a _CodeQL: Preview Query Help_ command to generate Markdown previews of `.qhelp` query help files. This command should only be run in trusted workspaces. See https://codeql.github.com/docs/codeql-cli/testing-query-help-files for more information about query help. [#988](https://github.com/github/vscode-codeql/pull/988)
 - Make "Open Referenced File" command accessible from the active editor menu. [#989](https://github.com/github/vscode-codeql/pull/989)
 
 ## 1.5.6 - 07 October 2021
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -46,6 +46,7 @@
     "onCommand:codeQL.setCurrentDatabase",
     "onCommand:codeQL.viewAst",
     "onCommand:codeQL.openReferencedFile",
+    "onCommand:codeQL.previewQueryHelp",
     "onCommand:codeQL.chooseDatabaseFolder",
     "onCommand:codeQL.chooseDatabaseArchive",
     "onCommand:codeQL.chooseDatabaseInternet",
@@ -294,6 +295,10 @@
       {
         "command": "codeQL.openReferencedFile",
         "title": "CodeQL: Open Referenced File"
+      },
+      {
+        "command": "codeQL.previewQueryHelp",
+        "title": "CodeQL: Preview Query Help"
       },
       {
         "command": "codeQL.quickQuery",
@@ -682,6 +687,11 @@
           "when": "view == codeQLQueryHistory"
         },
         {
+          "command": "codeQL.previewQueryHelp",
+          "group": "9_qlCommands",
+          "when": "view == codeQLQueryHistory && resourceScheme == .qhelp && isWorkspaceTrusted"
+        },
+        {
           "command": "codeQLTests.showOutputDifferences",
           "group": "qltest@1",
           "when": "view == test-explorer && viewItem == testWithSource"
@@ -712,6 +722,11 @@
           "command": "codeQL.openReferencedFile",
           "group": "9_qlCommands",
           "when": "resourceExtname == .qlref"
+        },
+        {
+          "command": "codeQL.previewQueryHelp",
+          "group": "9_qlCommands",
+          "when": "resourceExtname == .qhelp && isWorkspaceTrusted"
         }
       ],
       "commandPalette": [
@@ -742,6 +757,10 @@
         {
           "command": "codeQL.openReferencedFile",
           "when": "resourceExtname == .qlref"
+        },
+        {
+          "command": "codeQL.previewQueryHelp",
+          "when": "resourceExtname == .qhelp && isWorkspaceTrusted"
         },
         {
           "command": "codeQL.setCurrentDatabase",
@@ -900,6 +919,10 @@
         {
           "command": "codeQL.openReferencedFile",
           "when": "resourceExtname == .qlref"
+        },
+        {
+          "command": "codeQL.previewQueryHelp",
+          "when": "resourceExtname == .qhelp && isWorkspaceTrusted"
         }
       ]
     },
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -621,6 +621,20 @@ export class CodeQLCliServer implements Disposable {
 
     return await this.runCodeQlCliCommand(['database', 'unbundle'], subcommandArgs, `Extracting ${archivePath} to directory ${target}`);
   }
+  
+  /**
+   * Uses a .qhelp file to generate Query Help documentation in a specified format.
+   * @param pathToQhelp The path to the .qhelp file
+   * @param format The format in which the query help should be generated {@link https://codeql.github.com/docs/codeql-cli/manual/generate-query-help/#cmdoption-codeql-generate-query-help-format}
+   * @param outputDirectory The output directory for the generated file
+   */
+  async generateQueryHelp(pathToQhelp:string, outputDirectory?: string): Promise<string> {
+    const subcommandArgs = ['--format=markdown'];
+    if(outputDirectory) subcommandArgs.push('--output', outputDirectory);
+    subcommandArgs.push(pathToQhelp);
+
+    return await this.runCodeQlCliCommand(['generate', 'query-help'], subcommandArgs, `Generating qhelp in markdown format at ${outputDirectory}`);
+  }
 
   /**
   * Gets the results from a bqrs.
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -16,6 +16,7 @@ import {
 import { LanguageClient } from 'vscode-languageclient';
 import * as os from 'os';
 import * as path from 'path';
+import * as tmp from 'tmp-promise';
 import { testExplorerExtensionId, TestHub } from 'vscode-test-adapter-api';
 
 import { AstViewer } from './astViewer';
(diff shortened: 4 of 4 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -608,6 +608,21 @@ export class CodeQLCliServer implements Disposable {
  
     return await this.runCodeQlCliCommand(['database', 'unbundle'], subcommandArgs, `Extracting ${archivePath} to directory ${target}`);
   }
+  
+  /**
+   * Uses a .qhelp file to generate Query Help documentation in a specified format.
+   * @param pathToQlhelp The path to the .qhelp file
```
shati-patel, line 614: Here and in other places: `qlhelp` should be `qhelp`. (For consistency, we try to use "query help" as the full name and "qhelp" as the abbreviation!)

shati-patel (commented):
This new command looks exciting! (Not a real review yet, just a small comment about terminology)

extensions/ql-vscode/src/cli.ts
```diff
@@ -613,6 +613,21 @@ export class CodeQLCliServer implements Disposable {
...
     return await this.runCodeQlCliCommand(['database', 'unbundle'], subcommandArgs, `Extracting ${archivePath} to directory ${target}`);
   }
+  
+  /**
+   * Uses a .qhelp file to generate Query Help documentation in a specified format.
+   * @param pathToQhelp The path to the .qhelp file
+   * @param format The format in which the query help should be generated {@link https://codeql.github.com/docs/codeql-cli/manual/generate-query-help/#cmdoption-codeql-generate-query-help-format}
+   * @param outputDirectory The output directory for the generated file
+   */
+  async generateQueryHelp(pathToQhelp:string, format: string, outputDirectory?: string): Promise<string> {
```
adityasharad, line 623: You might as well hardcoded `--format=markdown` in the call below, rather than allowing the format to be a parameter of this function. I don't think we need to support any other format.

extensions/ql-vscode/package.json
```diff
@@ -295,6 +296,10 @@
         "command": "codeQL.openReferencedFile",
         "title": "CodeQL: Open Referenced File"
       },
+      {
+        "command": "codeQL.previewQueryHelp",
+        "title": "CodeQL: Preview Query Help File"
```
adityasharad, line 301: Minor. I think this is clear enough without having to specify it's a file.
```suggestion
        "title": "CodeQL: Preview Query Help"
```

extensions/ql-vscode/src/extension.ts
```diff
@@ -493,6 +494,29 @@ async function activateWithInstalledDistribution(
     }
   }
 
+  async function previewQueryHelp(
+    selectedQuery: Uri
+  ): Promise<void> {
+    const path = selectedQuery ? selectedQuery.path : window.activeTextEditor!.document.uri.fsPath;
```
adityasharad, line 500: If this command is called from the context menu on the editor, I would expect `selectedQuery` to be exactly `window.activeTextEditor.document.uri`. Is that what you observed while testing this?

extensions/ql-vscode/src/extension.ts
```diff
@@ -493,6 +494,29 @@ async function activateWithInstalledDistribution(
...
+    if(path) {
+      // Create temporary directory
+      const tmpDir = tmp.dirSync();
```
adityasharad, line 503: Using a temporary directory is sensible. However we must be extra careful to clean it up properly, especially on Windows.

Have you seen the other uses of temporary directories in `run-queries.ts`? I suggest you follow that pattern as closely as possible. In particular, I recommend passing similar options to `dirSync`, and creating a dispose callback that can be registered with the extension via `ctx.subscriptions.push(callback)`, so that the extension cleans up the temporary directories when it shuts down.

---
More detailed discussion: We could achieve more elegant code by adding an extra field to the `QueryInfo` class in `run-queries.ts`, which currently tracks a `dilPath` and `resultsPath` for each completed query. What do others think?
- Advantages: It handles all the temporary directories already, associates each query with a number of temporary file paths, and doesn't repeat work if the file has already been generated.
- Disadvantages: It only works for queries that have been run, which seems unreasonable for previewing query help. We would also have to track whether the `.qhelp` file had been modified.

extensions/ql-vscode/src/extension.ts
```diff
@@ -493,6 +494,29 @@ async function activateWithInstalledDistribution(
...
+      const absolutePathToMd = `${tmpDir.name}${mdFilePath}`;
+      const uri = Uri.file(absolutePathToMd);
+      try {
+        await cliServer.generateQueryHelp(path , 'markdown', absolutePathToMd);
+        await commands.executeCommand('markdown.showPreviewToSide', uri);
+      } catch (err) {
+        const errorMessage =  err.message.includes('Generating qhelp in markdown') ? (
+          `Could not generate markdown from ${path}: Bad formatting in .qhelp file.`
+        ) : `Could not open a preview of the generated file (${absolutePathToMd}).`;
+        void helpers.showAndLogErrorMessage(errorMessage);
```
adityasharad, line 515: I think this will suppress the error message from the CLI, which is hopefully useful. Could we include `err` in the message we construct here?

extensions/ql-vscode/package.json
```diff
@@ -900,6 +919,10 @@
         {
           "command": "codeQL.openReferencedFile",
           "when": "resourceExtname == .qlref"
+        },
+        {
+          "command": "codeQL.previewQueryHelp",
+          "when": "resourceExtname == .qhelp"
```
adityasharad, line 925:
```suggestion
          "when": "resourceExtname == .qhelp && isWorkspaceTrusted"
```

extensions/ql-vscode/package.json
```diff
@@ -743,6 +758,10 @@
           "command": "codeQL.openReferencedFile",
           "when": "resourceExtname == .qlref"
         },
+        {
+          "command": "codeQL.previewQueryHelp",
+          "when": "resourceExtname == .qhelp"
```
adityasharad, line 763:
```suggestion
          "when": "resourceExtname == .qhelp && isWorkspaceTrusted"
```

extensions/ql-vscode/package.json
```diff
@@ -712,6 +722,11 @@
           "command": "codeQL.openReferencedFile",
           "group": "9_qlCommands",
           "when": "resourceExtname == .qlref"
+        },
+        {
+          "command": "codeQL.previewQueryHelp",
+          "group": "9_qlCommands",
+          "when": "resourceExtname == .qhelp"
```
adityasharad, line 729:
```suggestion
          "when": "resourceExtname == .qhelp && isWorkspaceTrusted"
```

extensions/ql-vscode/package.json
```diff
@@ -681,6 +686,11 @@
           "group": "9_qlCommands",
           "when": "view == codeQLQueryHistory"
         },
+        {
+          "command": "codeQL.previewQueryHelp",
+          "group": "9_qlCommands",
+          "when": "resourceScheme == .qhelp"
```
adityasharad, line 692:
```suggestion
          "when": "view == codeQLQueryHistory && resourceScheme == .qhelp && isWorkspaceTrusted"
```

extensions/ql-vscode/src/extension.ts
```diff
@@ -493,6 +494,29 @@ async function activateWithInstalledDistribution(
...
+  async function previewQueryHelp(
+    selectedQuery: Uri
+  ): Promise<void> {
+    const path = selectedQuery ? selectedQuery.path : window.activeTextEditor!.document.uri.fsPath;
+    if(path) {
+      // Create temporary directory
+      const tmpDir = tmp.dirSync();
+      // Get library name and query file name in the format 'libname/queryname.md
+      const mdFilePath = path.split('/').slice(-2).join('/').replace('.qhelp', '.md');
+      const absolutePathToMd = `${tmpDir.name}${mdFilePath}`;
```
adityasharad, line 506: I would avoid doing path manipulation manually or assuming that the directory separator is always `/`. Let the Node `path` library do this work for you.
```suggestion
      const relativePathToMd = path.basename(path, '.qhelp') + '.md'
      const absolutePathToMd = path.join(tmpDir.name, relativePathToMd)`;
```

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Fix a bug with importing large databases. Databases over 4GB can now be imported directly from LGTM or from a zip file. This functionality is only available when using CodeQL CLI version 2.6.0 or later. [#971](https://github.com/github/vscode-codeql/pull/971)
 - Replace certain control codes (`U+0000` - `U+001F`) with their corresponding control labels (`U+2400` - `U+241F`)  in the results view. [#963](https://github.com/github/vscode-codeql/pull/963)
 - Allow case-insensitive project slugs for GitHub repositories when adding a CodeQL database from LGTM. [#978](https://github.com/github/vscode-codeql/pull/961)
+- Add a _CodeQL: Preview Query Help File_ command to generate and preview query documentation in markdown format from a .qhelp file. [#988](https://github.com/github/vscode-codeql/pull/988)
```
adityasharad, line 10:
```suggestion
- Add a _CodeQL: Preview Query Help_ command to generate Markdown previews of `.qhelp` query help files. See https://codeql.github.com/docs/codeql-cli/testing-query-help-files for more information about query help. [#988](https://github.com/github/vscode-codeql/pull/988)
```

adityasharad (commented):
Thanks! This is a good start to a much-awaited feature.

Most comments are about security (this command should only be run in a trusted workspace) and file/path handling. There is one longer suggestion about whether we should share the temporary file mechanism used for DIL and results files, on which I'd like opinions from the others on the team.

shati-patel (commented):
Just tried this locally and found one more thing! Looks good otherwise 🎇

extensions/ql-vscode/src/extension.ts
```diff
@@ -493,6 +494,32 @@ async function activateWithInstalledDistribution(
...
+        const absolutePathToMd = path.join(o.path, relativePathToMd);
+        const uri = Uri.file(absolutePathToMd);
+        try {
+          await cliServer.generateQueryHelp(pathToQhelp , absolutePathToMd);
+          await commands.executeCommand('markdown.showPreviewToSide', uri);
+        } catch (err) {
+          const errorMessage =  err.message.includes('Generating qhelp in markdown') ? (
+            `Could not generate markdown from ${pathToQhelp}: Bad formatting in .qhelp file.`
+          ) : `Could not open a preview of the generated file (${absolutePathToMd}).`;
+          void helpers.showAndLogErrorMessage(errorMessage);
```
shati-patel, line 514: As @adityasharad mentioned above (in https://github.com/github/vscode-codeql/pull/988#discussion_r740549449), it would be nice to print `err` itself somewhere (perhaps using `logger.log(...)`?) 📝

shati-patel (commented):
Sorry to leave this hanging! I think this is mostly good to merge, though I've highlighted one of the earlier review comments which looks worth doing 😃

extensions/ql-vscode/src/extension.ts
mgsium: Yes, as far as I've observed they are equal. `JSON.stringify(selectedQuery) === JSON.stringify(window.activeTextEditor.document.uri)` also evaluates to `true` in the cases I've tested

adityasharad: In that case I think it should be enough to use `selectedQuery.fsPath`.

adityasharad: Just `path = selectedQuery.fsPath`

mgsium: `path = selectedQuery.fsPath` fails for me when I run the command from the palette

adityasharad (approved):
Looking good, thank you. Final round of minor suggestions :)

extensions/ql-vscode/src/extension.ts
```diff
@@ -493,6 +494,32 @@ async function activateWithInstalledDistribution(
     }
   }
 
+  const tmpDir = tmp.dirSync({ prefix: 'qhelp_', keep: false, unsafeCleanup: true });
```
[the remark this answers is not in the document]
adityasharad: Also I think this pattern does the right thing on Windows (since we use it elsewhere in the extension), but I would not object to a second look from @shati-patel.

extensions/ql-vscode/CHANGELOG.md
adityasharad: I think you have managed to duplicate this changelog entry :)

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -7,7 +7,10 @@
...
 - Allow case-insensitive project slugs for GitHub repositories when adding a CodeQL database from LGTM. [#978](https://github.com/github/vscode-codeql/pull/961)
+- Add a _CodeQL: Preview Query Help_ command to generate Markdown previews of `.qhelp` query help files. This command should only be run in trusted workspaces. See https://codeql.github.com/docs/codeql-cli/testing-query-help-files for more information about query help. [#988](https://github.com/github/vscode-codeql/pull/988)
 - Make "Open Referenced File" command accessible from the active editor menu. [#989](https://github.com/github/vscode-codeql/pull/989)
+- Add a _CodeQL: Preview Query Help File_ command to generate and preview query documentation in markdown format from a .qhelp file. [#988](https://github.com/github/vscode-codeql/pull/988)
+- Add a _CodeQL: Preview Query Help_ command to generate Markdown previews of `.qhelp` query help files. See https://codeql.github.com/docs/codeql-cli/testing-query-help-files for more information about query help. [#988](https://github.com/github/vscode-codeql/pull/988)
```
shati-patel, line 13: There are some outdated copies of the changelog entry here! The top one looks most correct, so you can delete the others :)

shati-patel (approved):
This looks good to me! Once you've tidied up the changelog this should be ready to merge 🥳

extensions/ql-vscode/src/extension.ts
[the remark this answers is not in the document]
shati-patel: > Also I think this pattern does the right thing on Windows (since we use it elsewhere in the extension), but I would not object to a second look from @shati-patel.

Looks good to me too! 😊

shati-patel:
Nice work! Thank you 🎈 🥳

PR #997, opened 2021-11-08
Avoid testing for the correct CLI_VERSION if CLI_PATH is set

aeisenberg:
This should fix the failing CLI integration tests.


hmakholm (approved):

aeisenberg:
Thanks for the quick review!

PR #1007, opened 2021-11-17
Ensure all result set names are loaded

### LINKED ISSUE #1005: Sorting results causes the dropdown to not show other `query predicate` results
When viewing results of queries with multiple `query predicate`s, if you click one of the column headers to sort by that column, you can no longer view the other `query predicate` results in the dropdown unless you go to a previous set of results and back again.

**Version**
CodeQL extension version: 1.5.6 CodeQL CLI version: 2.7.0 Platform: darwin x64

**To reproduce**
Save this query as `resultsBug.ql` and run on any database:
```ql
query predicate foo(string s) {
	s = "foo"
}

query predicate bar(string s) {
	s = "bar"
}
```

Click the dropdown. Notice how you see `foo`, `bar`, and `alerts`. In the dropdown, select `foo`. Click the column header for `s` to sort by that column. Now click the dropdown again. Notice how you now only see `foo` and `alerts`, and that `bar` is missing.

**Expected behavior**
After sorting one `query predicate`'s results by some column, I am still able to switch back to other `query predicate` results with the dropdown.

aeisenberg:
When the extension loads a sorted result set, it takes a shortcut and
avoids loads a file with only the bqrs results for that sorted table.

However, it does not load the results for any other table. This causes
result set names to go away. This change ensures that if we are loading
a sorted table, we also load the result set names for all other tables
in that query.

Fixes #1005.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,12 +2,13 @@
 
 ## [UNRELEASED]
 
-- Fix the _CodeQL: Open Referenced File_ command for Windows systems. [#979](https://github.com/github/vscode-codeql/pull/979) 
+- Fix the _CodeQL: Open Referenced File_ command for Windows systems. [#979](https://github.com/github/vscode-codeql/pull/979)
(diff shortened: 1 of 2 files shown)


shati-patel (approved):
Looks good to me! 🌷

PR #1014, opened 2021-11-22
Add a `codeql` task provider

shati-patel:
Adds a `codeql` task provider. The idea is that we can use this for packaging commands (and others). See internal issue for more context.

Currently, the available tasks are:
- `resolve qlpacks` in the workspace
- `pack download` (only for public packages for now)
- `pack install` (only works for qlpacks in the root of a workspace folder)

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/codeql-task-provider.ts
+++ b/extensions/ql-vscode/src/codeql-task-provider.ts
@@ -21,7 +21,12 @@ interface CodeQLTaskDefinition extends vscode.TaskDefinition {
   /**
    * User-specified properties for the task.
    */
-  prompts: string[];
+  prompts: Prompt[];
+}
+
+interface Prompt {
+  message: string;
+  values?: string[];
 }
 
 export class CodeQLTaskProvider implements vscode.TaskProvider {
@@ -49,30 +54,35 @@ export class CodeQLTaskProvider implements vscode.TaskProvider {
     }
     const cliCommands: CodeQLTaskDefinition[] = [
       {
-        command: 'version',  // TODO: Delete this, just a test command.
-        commandArgs: [],
-        prompts: [],
-        type: CodeQLTaskProvider.CodeQLType,
-      },
-      {
         command: 'resolve qlpacks',
-        commandArgs: ['--additional-packs', getOnDiskWorkspaceFolders().join(path.delimiter)],
+        commandArgs: [
+          '--additional-packs',
+          getOnDiskWorkspaceFolders().join(path.delimiter),
+        ],
         prompts: [],
         type: CodeQLTaskProvider.CodeQLType,
       },
       {
         command: 'pack install',
-        commandArgs: [],       // TODO: Specify which qlpack to install dependencies for. Resolve the commandArgs to all qlpack.ymls in the workspace and add those as --dir args.
-        prompts: [],
+        commandArgs: [],
+        prompts: [
+          {
+            message: 'Select the pack to install dependencies for',
+            values: getWorkspacePacks(),
+          },
+        ],
         type: CodeQLTaskProvider.CodeQLType,
       },
       {
         command: 'pack download',
         commandArgs: [],
-        prompts: ['Enter the <package-scope/name[@version]> of the pack to download:'],
+        prompts: [
+          {
+            message: 'Enter the <package-scope/name[@version]> of the pack to download',
+          },
+        ],
         type: CodeQLTaskProvider.CodeQLType,
-      },  // Needs a PAT for private packs (from octokit?)
-      // 'database create'   // Needs additional args.
+      },
     ];
 
     const tasks: vscode.Task[] | undefined = [];
@@ -99,14 +109,27 @@ class CodeQLTaskTerminal extends DisposableObject implements vscode.Pseudotermin
 
   // TODO: Open the terminal (i.e. run the task) in an appropriate folder (e.g. the enclosing folder of the currently active editor).
   // Currently, it is run from the first workspace folder (codeql-custom-queries-cpp in the starter workspace).
-  constructor(private cliServer: CodeQLCliServer, private command: string, private commandArgs: string[], private prompts: string[]) {
+  constructor(private cliServer: CodeQLCliServer, private command: string, private commandArgs: string[], private prompts: Prompt[]) {
     super();
   }
 
   async open(): Promise<void> {
     const promptedArgs: string[] = [];
     for (const prompt of this.prompts) {
-      const result = await vscode.window.showInputBox({ prompt, ignoreFocusOut: true });
+      let result: string | undefined;
+      // If prompt values are specified, show a quick pick to select the value.
+      // If no values are offered, show a text input box instead.
+      if (prompt.values) {
+        result = await vscode.window.showQuickPick(prompt.values, {
+          placeHolder: prompt.message,
+          ignoreFocusOut: true,
+        });
+      } else {
+        result = await vscode.window.showInputBox({
+          placeHolder: prompt.message,
+          ignoreFocusOut: true,
+        });
+      }
       if (result === undefined) {
         throw new UserCancellationException('No inputs provided.');
       }
@@ -133,13 +156,15 @@ class CodeQLTaskTerminal extends DisposableObject implements vscode.Pseudotermin
 
 /**
  * Lists all workspace folders that contain a qlpack.yml file.
+ * 
+ * Note: This currently only finds packs at the root of a workspace folder.
  */
 function getWorkspacePacks(): string[] {
   const packs: string[] = [];
   const workspaceFolders = getOnDiskWorkspaceFolders();
   for (const folder of workspaceFolders) {
     const qlpackYml = path.join(folder, 'qlpack.yml');
-    if (fs.pathExists(qlpackYml)) {
+    if (fs.pathExistsSync(qlpackYml)) {
       packs.push(folder);
     }
   }
--- a/extensions/ql-vscode/src/codeql-task-provider.ts
+++ b/extensions/ql-vscode/src/codeql-task-provider.ts
@@ -37,7 +37,16 @@ export class CodeQLTaskProvider implements vscode.TaskProvider {
       return this.tasks;
     }
     // Hard-code examples for now. Not sure if we can do this in a better way.
-    const commands: string[] = ['resolve languages', 'version'];
+    const commands: string[] = [
+      'resolve languages',
+      'version',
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  public async provideTasks(): Promise<vscode.Task[]> {
+    return this.getTasks();
+  }
+
+  public resolveTask(_task: vscode.Task): vscode.Task | undefined {
+    const command: string = _task.definition.command;
+    if (command) {
+      const definition: CodeQLTaskDefinition = <any>_task.definition;
+      return this.getTask(definition.command, definition.commandArgs ? definition.commandArgs : [], definition);
```
aeisenberg, line 32: nit:

```suggestion
      return this.getTask(definition.command, definition.commandArgs ?? [], definition);
```

extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,97 @@
...
+  close(): void {
+    // The terminal has been closed. Shutdown the build.
+    if (this.fileWatcher) {
+      this.fileWatcher.dispose();
+    }
+  }
+
+  private async runCli(command: string[], commandArgs: string[]): Promise<void> {
+    this.writeEmitter.fire('Running CodeQL task\r\n');
+    const output = JSON.parse(await this.cliServer.runCodeQlCliCommand(command, commandArgs, ''));
```
aeisenberg, line 94: So, this will use the existing cli server, or restart it if it's not currently running? I think this works.

Definitely need to handle malformed output in case the process abruptly ends partway through. Just add a try-catch with a friendly error message.

extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return new vscode.Task(definition, vscode.TaskScope.Workspace, `${command} ${commandArgs.join(' ')}`,
+      CodeQLTaskProvider.CodeQLType, new vscode.CustomExecution(async (): Promise<vscode.Pseudoterminal> => {
+        // When the task is executed, this callback will run. Here, we setup for running the task.
+        return new CodeQLTaskTerminal(this.workspaceRoot, this.cliServer, command, commandArgs, () => this.sharedState, (state: string) => this.sharedState = state);
+      }));
+  }
+}
+
+class CodeQLTaskTerminal implements vscode.Pseudoterminal {
```
aeisenberg, line 70: I guess there's no way to just use the regular extension logger for output and avoid creating a pseudo terminal? I feel like that would be nicer, but not sure if that's possible.

extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,97 @@
...
+    const commands: string[] = ['resolve languages', 'version'];
+    // Each command can have some commandArgs.
+    const commandArgs: string[][] = [[]];
+
+    this.tasks = [];
+    commands.forEach(command => {
+      commandArgs.forEach(commandArgsGroup => {
+        this.tasks!.push(this.getTask(command, commandArgsGroup));
+      });
+    });
```
aeisenberg, line 50: nit: Trick to avoid the type assertion:

```suggestion
    const tasks = [];
    commands.forEach(command => {
      commandArgs.forEach(commandArgsGroup => {
        tasks.push(this.getTask(command, commandArgsGroup));
      });
    });
    this.tasks = tasks;
```

aeisenberg (commented):
I think this is the right direction. If it's possible to add the output of each command directly to the output logger terminal, then I think that's a better approach. But, don't spend too much time trying to make it work, since it may not be possible.

extensions/ql-vscode/src/codeql-task-provider.ts
shati-patel: I think this should be possible 🤔  Will do some experimenting!

shati-patel: I couldn't work out how to use the regular logger, so I've worked on fixing the pseudo terminal so that it now actually runs the task 😅

extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+    // Hard-code examples for now. Not sure if we can do this in a better way.
+    const commands: string[] = [
+      'resolve languages',
+      'version',
+      'pack install',      // TODO: Specify which qlpack to install dependencies for.
+      'resolve queries',   // TODO: Specify which query, directory, or suite to resolve. Also specify search path/additional packs.
+      'resolve qlpacks',   // TODO: Specify a search path/additional packs.
+      // 'pack publish',     // Needs a PAT (from octokit?)
+      // 'pack download',    // Needs a PAT (from octokit?)
```
aeisenberg, line 47: This one does not need a PAT for public packages.

extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  private fileWatcher: vscode.FileSystemWatcher | undefined;
+
+  // TODO: Open the terminal (i.e. run the task) in an appropriate folder (e.g. the enclosing folder of the currently active editor).
+  // Currently, it is run from the first workspace folder (codeql-custom-queries-cpp in the starter workspace).
+  constructor(private cliServer: CodeQLCliServer, private command: string, private commandArgs: string[]) {
+  }
+
+  async open(): Promise<void> {
+    await this.runCli([this.command], this.commandArgs);
```
aeisenberg, line 91: Something I forgot until this morning is that we are going to need to capture telemetry when users run tasks. It shouldn't be hard to do and might only be 1 line.

See how this is implemented in the command runner: https://github.com/github/vscode-codeql/blob/1d414bac552f5b4ac03527c5666b8545c694cd05/extensions/ql-vscode/src/commandRunner.ts#L196

I'm happy to see this PR go in and then add the telemetry later.

extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  close(): void {
+    // The terminal has been closed. Shutdown the build.
+    if (this.fileWatcher) {
+      this.fileWatcher.dispose();
+    }
+  }
```
aeisenberg, line 99: I think you can make this class a `DisposableObject`. Then when you first create the file watcher, you can just do `this.push(this.fileWatcher)`.

This close method will not be necessary.

aeisenberg: If `close`  is an abstract super method, then you can change the body to `this.dispose();`.

extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (this.tasks !== undefined) {
+      return this.tasks;
+    }
+    // Hard-code examples for now. Not sure if we can do this in a better way.
+    const commands: string[] = [
+      'resolve languages',
+      'version',
+      'pack install',      // TODO: Specify which qlpack to install dependencies for.
+      'resolve queries',   // TODO: Specify which query, directory, or suite to resolve. Also specify search path/additional packs.
+      'resolve qlpacks',   // TODO: Specify a search path/additional packs.
```
aeisenberg, line 45: The additional packs should just be all the folders in the current workspace.

extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,111 @@
...
+  private getTasks(): vscode.Task[] {
+    if (this.tasks !== undefined) {
+      return this.tasks;
+    }
+    // Hard-code examples for now. Not sure if we can do this in a better way.
+    const commands: string[] = [
+      'resolve languages',
+      'version',
+      'pack install',      // TODO: Specify which qlpack to install dependencies for.
+      'resolve queries',   // TODO: Specify which query, directory, or suite to resolve. Also specify search path/additional packs.
```
aeisenberg, line 44: Hmmm...not sure if/how we can specify this. For `pack install`, I think you can look at the active editor (or selected file in the side bar) and assume the qlpack is in that directory, but that is a bit of a pain/

Maybe easier to put up select box where user can choose which pack to run on.

Not sure about `resolve queries`. I think it is less important than the other commands.

aeisenberg:
I'm playing around with this change locally and it is working. But, I'm learning some more things about how task providers work and I think we can make this better (or at least a more canonical task provider implementation). I think best if we have a face to face chat about this. I don't think the changes are very big and may even make things simpler.

extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  /**
+   * Additional command arguments.
+   */
+  commandArgs: string[];
+
+  /**
+   * User-specified properties for the task.
+   */
+  prompts: Prompt[];
```
aeisenberg, line 24: Playing around with this, using the `prompts` makes for some odd interactions if a user saves the task to the tasks.json. I think better would be to create some reasonable defaults and then allow the user to edit manually.

extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,172 @@
...
+      },
+      {
+        command: 'pack install',
+        commandArgs: [],
+        prompts: [
+          {
+            message: 'Select the pack to install dependencies for',
+            values: getWorkspacePacks(),
+          },
+        ],
```
aeisenberg, line 73: Rather than a prompt here, can we just add all of the workspace packs by default, and let the user edit manually?

extensions/ql-vscode/src/codeql-task-provider.ts
```diff
@@ -0,0 +1,172 @@
...
+        type: CodeQLTaskProvider.CodeQLType,
+      },
+      {
+        command: 'pack download',
+        commandArgs: [],
+        prompts: [
+          {
+            message: 'Enter the <package-scope/name[@version]> of the pack to download',
+          },
+        ],
```
aeisenberg, line 83: Rather than a prompt, we can hard code downloading all of the core language query packs and let the user manually override.

shati-patel:
Closing this PR, since we're reconsidering our approach! 😄 (more context in internal issue)

PR #1022, opened 2021-11-26
[Internal only] Support running queries with ML models from CodeQL packs

henrymercer:
This functionality is for internal users only.  Do not use this functionality if you are not a GitHub employee.

This PR allows internal users to run queries that invoke ML models from CodeQL packs.  When a hidden setting is enabled, ML models are resolved using the CLI command `codeql resolve ml-models`.  To help prevent unintentional use of this functionality, we only respect the hidden setting if the workspace is trusted.

Available ML models are passed to the `availableMlModels` property of the `QueryToRun` object as part of the `evaluation/runQueries` queryserver message.  This makes them available for use in CodeQL queries.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -8,7 +8,8 @@ import {
   TextDocument,
   TextEditor,
   Uri,
-  window
+  window,
+  workspace
 } from 'vscode';
 import { ErrorCodes, ResponseError } from 'vscode-languageclient';
 
@@ -617,12 +618,18 @@ export async function compileAndRunQueryAgainstDatabase(
   }
 
   let availableMlModels: cli.MlModelInfo[] = [];
-  if (await cliServer.cliConstraints.supportsResolveMlModels()) {
+  // The `capabilities.untrustedWorkspaces.restrictedConfigurations` entry in package.json doesn't
+  // work with hidden settings, so we manually check that the workspace is trusted before looking at
+  // whether the `shouldInsecurelyLoadMlModelsFromPacks` setting is enabled.
+  if (workspace.isTrusted &&
+    config.shouldInsecurelyLoadMlModelsFromPacks() &&
+    await cliServer.cliConstraints.supportsResolveMlModels()) {
     try {
       availableMlModels = (await cliServer.resolveMlModels(diskWorkspaceFolders)).models;
       void logger.log(`Found available ML models at the following paths: ${availableMlModels.map(x => `'${x.path}'`).join(', ')}.`);
     } catch (e) {
-      const message = `Couldn't resolve available ML models for ${qlProgram.queryPath}: ${e}`;
+      const message = `Couldn't resolve available ML models for ${qlProgram.queryPath}. Running the ` +
+        `query without any ML models: ${e}.`;
       void logger.log(message);
       void showAndLogErrorMessage(message);
     }
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -95,7 +95,7 @@ export class QueryInfo {
 
     const callbackId = qs.registerCallback(res => { result = res; });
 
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/run-queries.ts
```diff
@@ -86,19 +87,23 @@ export class QueryInfo {
...
     upgradeQlo: string | undefined,
+    availableMlModels: cli.MlModelInfo[],
     progress: ProgressCallback,
     token: CancellationToken,
   ): Promise<messages.EvaluationResult> {
     let result: messages.EvaluationResult | null = null;
 
     const callbackId = qs.registerCallback(res => { result = res; });
 
+    const availableMlModelUris: messages.MlModel[] = availableMlModels.map(model => ({ uri: Uri.file(model.path).toString() }));
```
aeisenberg, line 98: I think this is what you want. It skips % encoding of special characters like `=` and `&`.

```suggestion
    const availableMlModelUris: messages.MlModel[] = availableMlModels.map(model => ({ uri: Uri.file(model.path).toString(true) }));
```

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -612,6 +617,24 @@ export async function compileAndRunQueryAgainstDatabase(
     void logger.log(`Couldn't resolve metadata for ${qlProgram.queryPath}: ${e}`);
   }
 
+  let availableMlModels: cli.MlModelInfo[] = [];
+  // The `capabilities.untrustedWorkspaces.restrictedConfigurations` entry in package.json doesn't
+  // work with hidden settings, so we manually check that the workspace is trusted before looking at
+  // whether the `shouldInsecurelyLoadMlModelsFromPacks` setting is enabled.
```
aeisenberg, line 623: Arguably, we should _never_ be insecurely loading anything from an untrusted workspace. So, I think this makes sense regardless.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -612,6 +617,24 @@ export async function compileAndRunQueryAgainstDatabase(
...
+    config.shouldInsecurelyLoadMlModelsFromPacks() &&
+    await cliServer.cliConstraints.supportsResolveMlModels()) {
+    try {
+      availableMlModels = (await cliServer.resolveMlModels(diskWorkspaceFolders)).models;
+      void logger.log(`Found available ML models at the following paths: ${availableMlModels.map(x => `'${x.path}'`).join(', ')}.`);
+    } catch (e) {
+      const message = `Couldn't resolve available ML models for ${qlProgram.queryPath}. Running the ` +
+        `query without any ML models: ${e}.`;
+      void logger.log(message);
+      void showAndLogErrorMessage(message);
```
aeisenberg, line 634: `showAndLogErrorMessage` already logs the error message. No need to do it twice.

```suggestion
      void showAndLogErrorMessage(message);
```

extensions/ql-vscode/src/run-queries.ts
aeisenberg: Though, I am a little confused. It looks like unless `shouldInsecurelyLoadMlModelsFromPacks` is true, there will be _no_ ml models available. 

Although, perhaps what's going on is that this command only loads extra ml models not available in the default distro. And default ml models will always be available.

henrymercer: The `shouldInsecurelyLoadMlModelsFromPacks` setting allows internal users to turn on loading ML models from CodeQL packs.  While it's in beta, we assume that this is an insecure operation, hence the `insecurely`.  However once out of beta we plan for this operation to be a secure one, allowing us to load a limited set of GitHub-authored ML models by default and removing the need for the setting.

adityasharad: If you want to be extra defensive, you could also insist that the canary flag is enabled.

aeisenberg (approved):

PR #1033, opened 2021-12-06
Monitor remote query run and render results

charisk:
This PR introduces logic to monitor the progress of a remote query as well as rendering the remote queries view using real data from the result of the query.

The notification that links to the Actions workflow is still in place:
image

But once the query is complete, you get one more notification (assuming you leave VS Code open):
image

Which leads to a view similar to the following:
image

## Summary of changes
* Added a new `codeQL.monitorRemoteQuery` command that is executed once the remote query has been submitted to run. The new command is not visible to the user.
* Introduced new entities to represent the result of a remote query from a domain point of view and the UI point of view. View entities are inside the /views directory.
* Introduced a `RemoteQueriesManager` that is used as an entry point for running and monitoring remote queries from the extension. This class then uses the `RemoteQuriesMonitor` to monitor the progress of the query and the `RemoteQueriesInterfaceManager` to prepare data for the view and render the view.
* Updated the web view (and message) to retrieve use state and real data.

## Checklist


extensions/ql-vscode/src/extension.ts
```diff
@@ -775,6 +774,14 @@ async function activateWithInstalledDistribution(
       cancellable: true
     })
   );
+
+  ctx.subscriptions.push(
+    commandRunner('codeQL.monitorRemoteQuery', async (
```
aeisenberg, line 779: `commandRunner` does not supply a progress monitor. Use this instead:

```suggestion
    commandRunnerWithProgress('codeQL.monitorRemoteQuery', async (
```

extensions/ql-vscode/src/pure/interface-types.ts
```diff
@@ -370,14 +371,15 @@ export type FromRemoteQueriesMessage =
...
 
 export interface RemoteQueryLoadedMessage {
   t: 'remoteQueryLoaded';
 }
 
-export interface OpenRemoteQueriesViewMessage {
-  t: 'openRemoteQueriesView';
+export interface SetRemoteQueryResultMessage {
+  t: 'setRemoteQueryResult';
+  d: RemoteQueryResult
```
aeisenberg, line 382: We've only been using single char property names for `t`.

```suggestion
  data: RemoteQueryResult
```

or maybe:

```suggestion
  result: RemoteQueryResult
```

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
```diff
@@ -124,11 +144,64 @@ export class RemoteQueriesInterfaceManager extends DisposableObject {
...
+
+  private formatFileSize(bytes: number): string {
+    if (bytes < 1024) {
+      return `${bytes} bytes`;
+    } else if (bytes < 1024 * 1024) {
+      return `${(bytes / 1024).toFixed(2)} kb`;
+    } else if (bytes < 1024 * 1024 * 1024) {
+      return `${(bytes / 1024 / 1024).toFixed(2)} mb`;
+    } else {
+      return `${(bytes / 1024 / 1024 / 1024).toFixed(2)} gb`;
```
aeisenberg, line 182: nit: Can you extract these as constants?

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
```diff
@@ -124,11 +144,64 @@ export class RemoteQueriesInterfaceManager extends DisposableObject {
...
+    const filteredAnalysisResults = analysisResults.filter(r => r.resultCount > 0);
+
+    const sortedAnalysisResults = filteredAnalysisResults.sort((a, b) => {
+      if (a.resultCount > b.resultCount) {
+        return -1;
+      } else if (a.resultCount < b.resultCount) {
+        return 1;
+      } else {
+        return 0;
+      }
```
aeisenberg, line 196: nit: simpler:

```suggestion
      return b.resultCount - a.resultCount;
```

extensions/ql-vscode/src/remote-queries/remote-queries-monitor.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  public async monitorQuery(
+    remoteQuery: RemoteQuery,
+    cancellationToken: vscode.CancellationToken
+  ): Promise<RemoteQueryWorkflowResult> {
+    const credentials = await Credentials.initialize(this.extensionContext);
+
+    if (!credentials) {
+      await showAndLogErrorMessage('Error authenticating with GitHub');
```
aeisenberg, line 24: I think we need to throw instead of log here.

extensions/ql-vscode/src/remote-queries/remote-queries-monitor.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    let attemptCount = 0;
+
+    const octokit = await credentials.getOctokit();
+
+    while (attemptCount <= RemoteQueriesMonitor.twoDaysInSeconds) {
```
aeisenberg, line 31: This expression is comparing the number of attempts with seconds. I see below that you are sleeping for a second between each call, but each call will be more than a second (since it's waiting for a network call to return). So, it's not really going to be two days in the end.

Also, if we decide to wait for more than 1s between calls, the name and value of the constant will need to change.

Suggestions:

1. Choose a max number of calls to the server, instead of a max time
2. Or, instead of sleeping for a fixed amount each time, choose the amount of time to sleep based on the current clock time. Eg- if we want to wait 5 seconds between calls, sleep for this amount:

    ```
    this.sleep(5000 - Date.now() + timeOfLastCall); // need to be careful if timeOfLastCall is more than 5 seconds ago.
    ```

And the while loop expression could look something like this:

```suggestion
    const startTime = Date.now();
    while ((Date.now() - startTime) / 1000 > RemoteQueriesMonitor.twoDaysInSeconds) { 
```

This will give a little more consistency with your checking and make it easier to change intervals.

extensions/ql-vscode/src/remote-queries/remote-queries-monitor.ts
```diff
@@ -0,0 +1,97 @@
...
+      if (this.areStringsSame(workflowRun.data.status, 'completed')) {
+        if (this.areStringsSame(workflowRun.data.conclusion, 'success')) {
+          return { status: 'CompletedSuccessfully' };
+        } else {
+          const error = this.getWorkflowError(workflowRun.data.conclusion);
+          return { status: 'CompletedUnsuccessfully', error };
+        }
+      }
+
+      await this.sleep(1000);
```
aeisenberg, line 51: This feels like it is checking too often. There might be an API rate limit that gets hit here. We need to be careful. Maybe bump it to every 3 or 5 seconds?

extensions/ql-vscode/src/remote-queries/remote-queries-monitor.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    if (this.areStringsSame(conclusion, 'failure')) {
+      // TODO: Get the actual error from the workflow.
+      return 'The remote query execution has failed.';
+    }
+
+    return `Unexpected query execution conclusion: ${conclusion}`;
+  }
+
+  private areStringsSame(a: string | null | undefined, b: string | null | undefined): boolean {
```
aeisenberg, line 84: Is it necessary to perform this kind of comparison? These are strings created by the server and if they use different casing or variants, then presumably, this is a server bug, I think. Is there any realistic situation where we expect differently varianted conclusions?

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
```diff
@@ -16,10 +15,12 @@ import {
 import { Logger } from '../logging';
 import { getHtmlForWebview } from '../interface-utils';
 import { assertNever } from '../pure/helpers-pure';
-import { commandRunner } from '../commandRunner';
+import { AnalysisResult, RemoteQueryResult } from './remote-query-result';
+import { RemoteQuery } from './remote-query';
+import { RemoteQueryResult as RemoteQueryResultViewModel } from './view/remote-query-result';
+import { AnalysisResult as AnalysisResultViewModel } from './view/remote-query-result';
```
aeisenberg, line 21: It would be nice to be more explicit about where the shared code is between the extension and the webview. 

Elsewhere, this shared code goes into the `src/pure` directory. Can you move these files there?

I would like to create a codeql query that asserts that the webviews only import modules that are either `pure` or in a `view` directory.

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -39,74 +36,64 @@ const AnalysisResult = (props: AnalysisResult) => (
...
+        {queryResult.results.map(result =>
+          <li key={result.nwo} className="vscode-codeql__results-list-item">
+            <AnalysisResultItem {...result} />
+          </li>
+        )}
+      </ul>
+    </div>;
+  } catch (err) {
+    console.error(err);
+    return <div>Error!</div>;
```
aeisenberg, line 91: Can you make this error message more descriptive?

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -39,74 +36,64 @@ const AnalysisResult = (props: AnalysisResult) => (
   </span>
 );
 
-export function RemoteQueries(props: Props): JSX.Element {
-  return <div className="vscode-codeql__remote-queries-view">
-    <h1 className="vscode-codeql__query-title">{props.queryTitle}</h1>
+export function RemoteQueries(_: Record<string, never>): JSX.Element {
```
aeisenberg, line 39: nit: I _think_ this will work too:

```suggestion
export function RemoteQueries(): JSX.Element {
```

extensions/ql-vscode/src/extension.ts
charisk: Do we need this even if we're not using the progress API?

extensions/ql-vscode/src/pure/interface-types.ts
charisk: Ah thanks, updated!

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
charisk: 🤦‍♀️ updated

extensions/ql-vscode/src/remote-queries/remote-queries-monitor.ts
charisk: I've gone with the first suggestion which is what I had in mind in the first place but failed to tidy up the code - max number of attempts. I think it keeps the code simpler. And since we don't have a product requirement to check for X hours, it seems more straightforward to do that. Things might change when we start managing query history etc.

extensions/ql-vscode/src/remote-queries/remote-queries-monitor.ts
charisk: Thank you - this is great feedback! I've updated to 5secs.

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
charisk: Would it be possible to apply the rule in all view directories (including subdirs), rather than just the top-level one? Then these files could stay where they are. If not then I'll just move them - but I do like the fact that they're close to all the other remote query stuff.

extensions/ql-vscode/src/extension.ts
```diff
@@ -775,6 +774,14 @@ async function activateWithInstalledDistribution(
...
+      query: RemoteQuery,
+      token: CancellationToken) => {
+      await rqm.monitorRemoteQuery(query, token);
+    }));
```
aeisenberg, line 783: You can remove the progress monitor and the cancellation token.

```suggestion
    commandRunner('codeQL.monitorRemoteQuery', async (
      query: RemoteQuery) => {
      await rqm.monitorRemoteQuery(query, token);
    }));
```

extensions/ql-vscode/src/extension.ts
aeisenberg: No, we don't. You can remove it.

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
aeisenberg: It's probably better to keep these files co-located with the rest of remote-queries. How about making a new directory called `shared` or something and putting them there?

extensions/ql-vscode/src/remote-queries/remote-queries-monitor.ts
aeisenberg: I don't think we need to worry about the server changing behaviour here (remember that we're writing the server, too :) ).

extensions/ql-vscode/src/extension.ts
charisk: Hmm but I'm using the token to allow cancellation. What am I missing?

aeisenberg: Oops...I missed that. Then you need to use `commandRunnerWithProgress` and just ignore the `progress` parameter.

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -0,0 +1,100 @@
...
+      false,
+      progress,
+      token);
+
+    if (querySubmission && querySubmission.query) {
+      void commands.executeCommand('codeQL.monitorRemoteQuery', querySubmission.query);
+    }
+  }
+
+  public async monitorRemoteQuery(
```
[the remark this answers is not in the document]
shati-patel: PS: As discussed offline, we may not need "Monitor Query" in the command palette at all. That might solve the problem 🤔

shati-patel (commented):
Looks good! Just a few minor comments 🌻

extensions/ql-vscode/src/extension.ts
charisk: Are we sure we need `commandRunnerWithProgress` here? If I use that I get a persistent notification (see below), which I don't think it matches the expected UX.

image

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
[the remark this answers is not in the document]
charisk: "Monitor Query" should not be visible any more.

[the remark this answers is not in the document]
shati-patel: Great - this all works for me now! 🎈 🎉

shati-patel (approved):
Thanks for the updates! Looks good from my point of view now

PR #1035, opened 2021-12-06
Add code lens for quick evaluation

aeisenberg:

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Add a CodeLens to make the Quick Evaluation command more accessible. Click the `Quick Evaluation` prompt above a predicate definition in the editor to evaluate that predicate on its own. [#1035](https://github.com/github/vscode-codeql/pull/1035)
   
+- Fix a bug where the _Alerts_ option would show in the results view even if there is no alerts table available. [#1038](https://github.com/github/vscode-codeql/pull/1038)
+
 ## 1.5.8 - 2 December 2021
 
 - Emit a more explicit error message when a user tries to add a database with an unzipped source folder to the workspace. [#1021](https://github.com/github/vscode-codeql/pull/1021)
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,7 +3,6 @@
 ## [UNRELEASED]
 
 - Add a CodeLens to make the Quick Evaluation command more accessible. Click the `Quick Evaluation` prompt above a predicate definition in the editor to evaluate that predicate on its own. [#1035](https://github.com/github/vscode-codeql/pull/1035)
-  
 - Fix a bug where the _Alerts_ option would show in the results view even if there is no alerts table available. [#1038](https://github.com/github/vscode-codeql/pull/1038)
 
 ## 1.5.8 - 2 December 2021
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Add a CodeLens to make the Quick Evaluation command more accessible. Click the `Quick Evaluation` prompt above a predicate definition in the editor to evaluate that predicate on its own. [#1035](https://github.com/github/vscode-codeql/pull/1035)
+  
 ## 1.5.8 - 2 December 2021
 
 - Emit a more explicit error message when a user tries to add a database with an unzipped source folder to the workspace. [#1021](https://github.com/github/vscode-codeql/pull/1021)
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -76,8 +76,8 @@ import {
 import { CodeQlStatusBarHandler } from './status-bar';
 
 import { Credentials } from './authentication';
-import { runRemoteQuery } from './remote-queries/run-remote-query';
-import { RemoteQueriesInterfaceManager } from './remote-queries/remote-queries-interface';
+import { RemoteQueriesManager } from './remote-queries/remote-queries-manager';
+import { RemoteQuery } from './remote-queries/remote-query';
 
 /**
  * extension.ts
@@ -770,12 +770,8 @@ async function activateWithInstalledDistribution(
     )
   );
 
-  void logger.log('Initializing remote queries panel interface.');
-  const rmpm = new RemoteQueriesInterfaceManager(
-    ctx,
-    logger
-  );
-  ctx.subscriptions.push(rmpm);
+  void logger.log('Initializing remote queries interface.');
+  const rqm = new RemoteQueriesManager(ctx, logger, cliServer);
 
   // The "runRemoteQuery" command is internal-only.
   ctx.subscriptions.push(
@@ -790,8 +786,11 @@ async function activateWithInstalledDistribution(
           step: 0,
           message: 'Getting credentials'
         });
-        const credentials = await Credentials.initialize(ctx);
-        await runRemoteQuery(cliServer, credentials, uri || window.activeTextEditor?.document.uri, false, progress, token);
+        await rqm.runRemoteQuery(
+          uri || window.activeTextEditor?.document.uri,
+          progress,
+          token
+        );
       } else {
         throw new Error('Remote queries require the CodeQL Canary version to run.');
       }
@@ -800,6 +799,14 @@ async function activateWithInstalledDistribution(
       cancellable: true
     })
   );
+
+  ctx.subscriptions.push(
+    commandRunner('codeQL.monitorRemoteQuery', async (
+      query: RemoteQuery,
+      token: CancellationToken) => {
+      await rqm.monitorRemoteQuery(query, token);
+    }));
+
   ctx.subscriptions.push(
     commandRunner(
       'codeQL.openReferencedFile',
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -11,7 +11,8 @@ import {
   window as Window,
   env,
   window,
-  QuickPickItem
+  QuickPickItem,
+  Range
 } from 'vscode';
 import { LanguageClient } from 'vscode-languageclient';
 import * as os from 'os';
@@ -21,6 +22,7 @@ import { testExplorerExtensionId, TestHub } from 'vscode-test-adapter-api';
 
 import { AstViewer } from './astViewer';
 import * as archiveFilesystemProvider from './archive-filesystem-provider';
+import QuickEvalCodeLensProvider from './quickEvalCodeLensProvider';
 import { CodeQLCliServer, CliVersionConstraint } from './cli';
 import {
   CliConfigListener,
@@ -156,6 +158,7 @@ export interface CodeQLExtensionInterface {
  * @returns CodeQLExtensionInterface
  */
 export async function activate(ctx: ExtensionContext): Promise<CodeQLExtensionInterface | Record<string, never>> {
+
   void logger.log(`Starting ${extensionId} extension`);
   if (extension === undefined) {
     throw new Error(`Can't find extension ${extensionId}`);
@@ -165,6 +168,9 @@ export async function activate(ctx: ExtensionContext): Promise<CodeQLExtensionIn
   await initializeLogging(ctx);
   await initializeTelemetry(extension, ctx);
   languageSupport.install();
+
+  const codelensProvider = new QuickEvalCodeLensProvider();
+  languages.registerCodeLensProvider({ scheme: 'file', language: 'ql' }, codelensProvider);
 
   ctx.subscriptions.push(distributionConfigListener);
   const codeQlVersionRange = DEFAULT_DISTRIBUTION_VERSION_RANGE;
@@ -471,6 +477,7 @@ async function activateWithInstalledDistribution(
     progress: ProgressCallback,
     token: CancellationToken,
     databaseItem: DatabaseItem | undefined,
+    range?: Range
   ): Promise<void> {
     if (qs !== undefined) {
       // If no databaseItem is specified, use the database currently selected in the Databases UI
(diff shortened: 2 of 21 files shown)


adityasharad (commented):
A good start, and nice to see what we can achieve just from the editor.

extensions/ql-vscode/src/CodeLensProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    for (let index = 0; index < document.lineCount; index++) {
+      const textLine = document.lineAt(index);
+      const regex = new RegExp(/(?=(\w+\(\s*.*(?:,\s*)*\)\s*\{))\1/g);
+      if (textLine.text.match(regex)) {
+        const uri = document.uri;
+        const startIndex = textLine.text.search(regex);
+        const range: Range = new Range(
+          textLine.range.start.line, startIndex,
+          textLine.range.end.line, startIndex - 1
```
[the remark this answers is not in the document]
marcnjaramillo: I was thinking of something else when I made this change, but I didn't need to. Not sure now what that thought was. Changing back now. :)

extensions/ql-vscode/src/CodeLensProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+        );
+
+        const command: Command = {
+          command: 'codeQL.codeLensQuickEval',
+          title: 'CodeQL: Quick Evaluation',
```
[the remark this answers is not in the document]
marcnjaramillo: Title is required for the Command type. If I try deleting it I get an error.

extensions/ql-vscode/src/CodeLensProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+class MyCodeLensProvider implements CodeLensProvider {
+  // Each provider requires a provideCodeLenses function which will give the various documents
+  // the code lenses
+  async provideCodeLenses(document: TextDocument): Promise<CodeLens[]> {
+
+    const codeLenses: CodeLens[] = [];
+
+    for (let index = 0; index < document.lineCount; index++) {
+      const textLine = document.lineAt(index);
+      const regex = new RegExp(/(?=(\w+\(\s*.*(?:,\s*)*\)\s*\{))\1/g);
```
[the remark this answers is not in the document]
marcnjaramillo: Making this change. I've always struggled writing regex (although I'm getting more comfortable with it now), so I always appreciate feedback. The only reason I included the positive lookahead was I was getting a code scanning failure about potential exponential backtracking. But I have also since simplified the regex, so maybe that won't be an issue this time.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ## 1.5.8 - 2 December 2021
 
+- Add CodeLens feature to make the CodeQL quick evaluation command more accessible. [#1035](https://github.com/github/vscode-codeql/pull/1035)
```
aeisenberg, line 7: This should go in the unreleased section of the changelog.

aeisenberg:
Neat. I tried this out and it works. Though, it shows a quick eval suggestion even when the code is commented.

eg- this still shows a quick eval option:

```
// query BlockStmt foo(BlockStmt b) {
//   result = b and b.getNumStmt() >= 1
// }
```

aeisenberg:
And the following code does not show a quick eval option (notice the space before the `(`:

```
query BlockStmt foo (BlockStmt b) {
  result = b and b.getNumStmt() >= 1
}
```

This is an inherent limitation of using regexes. I think it's important to fix the first problem I mentioned, but this one is less important.

extensions/ql-vscode/src/CodeLensProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+  // Each provider requires a provideCodeLenses function which will give the various documents
+  // the code lenses
+  async provideCodeLenses(document: TextDocument): Promise<CodeLens[]> {
+
+    const codeLenses: CodeLens[] = [];
+
+    for (let index = 0; index < document.lineCount; index++) {
+      const textLine = document.lineAt(index);
+      // Match a predicate signature, including predicate name, parameter list, and opening brace.
+      const regex = new RegExp(/\w+\(\s*.*(?:,\s*)*\)\s*\{/);
```
aeisenberg, line 19: It would be _really_ nice to get the name of the predicate in the codelens text. You can do that by capturing the word as below:

```suggestion
      const regex = new RegExp(/(\w+)\(\s*.*(?:,\s*)*\)\s*\{/);
```

Then you can call: `const matches = textLine.text.match(regex);`

`matches[1]` contains the matched word.

extensions/ql-vscode/src/CodeLensProvider.ts
```diff
@@ -0,0 +1,40 @@
...
+        const startIndex = textLine.text.search(regex);
+        if (startIndex !== -1) {
+      const range: Range = new Range(
+        textLine.range.start.line, startIndex,
+        textLine.range.end.line, startIndex + 1
+      );
+
+      const command: Command = {
+        command: 'codeQL.codeLensQuickEval',
+        title: 'CodeQL: Quick Evaluation',
```
aeisenberg, line 29:
```suggestion
        title: `Quick Evaluation: ${matches[1]}`,
```

extensions/ql-vscode/CHANGELOG.md
marcnjaramillo: Whoops. I thought I put it there. I'll fix it.

extensions/ql-vscode/src/CodeLensProvider.ts
marcnjaramillo: Made these changes. I also added an additional whitespace match between the word match and opening '(' to account for any predicate definition that has that space.

extensions/ql-vscode/src/quickEvalCodeLensProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+    const codeLenses: CodeLens[] = [];
+
+    for (let index = 0; index < document.lineCount; index++) {
+      const textLine = document.lineAt(index);
+      // Match a predicate signature, including predicate name, parameter list, and opening brace.
+      const regex = new RegExp(/(\w+)\s*\(\s*.*(?:,\s*)*\)\s*\{/);
+      const matches = textLine.text.match(regex);
+      const startIndex = textLine.text.search(regex);
+
+      if (startIndex !== -1) {
```
adityasharad, line 23: Avoid calling both `match` and `search`, because you're doubling the computation cost at each line by regex matching twice. Now that we're using information from the result of `match`, stick with that.
`matches.index` will give you the start index you need, and `if (matches)` will give you the condition to check whether the match succeeded (because it returns `null` if there is no match.

Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

extensions/ql-vscode/src/extension.ts
```diff
@@ -733,6 +743,22 @@ async function activateWithInstalledDistribution(
...
       })
   );
+
+  ctx.subscriptions.push(
+    commandRunnerWithProgress(
+      'codeQL.codeLensQuickEval',
+      async (
+        progress: ProgressCallback,
+        token: CancellationToken,
+        uri: Uri | undefined,
```
adityasharad, line 753: I don't think this can ever be `undefined`, so might as well make the type strict.

extensions/ql-vscode/src/extension.ts
```diff
@@ -156,6 +158,11 @@ export interface CodeQLExtensionInterface {
  * @returns CodeQLExtensionInterface
  */
 export async function activate(ctx: ExtensionContext): Promise<CodeQLExtensionInterface | Record<string, never>> {
+
+  const codelensProvider = new QuickEvalCodeLensProvider();
+
+  languages.registerCodeLensProvider({ scheme: 'file', language: 'ql' }, codelensProvider);
```
adityasharad, line 164: Please move this further down in the function: I don't think code lens creation should happen before we call `languageSupport.install()`, and certainly not before we've set up the extension and logging.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Add CodeLens feature to make the CodeQL quick evaluation command more accessible. [#1035](https://github.com/github/vscode-codeql/pull/1035)
```
adityasharad, line 5: I've attempted to word this in a way that tells the user how to use the feature. What do you think?
```suggestion
- Add a CodeLens to make the Quick Evaluation command more accessible. Click the `Quick Evaluation` prompt above a predicate definition in the editor to evaluate that predicate on its own. [#1035](https://github.com/github/vscode-codeql/pull/1035)
```

adityasharad (commented):
Good work wiring this together. A few follow up suggestions.

extensions/ql-vscode/src/CodeLensProvider.ts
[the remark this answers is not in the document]
adityasharad: Yup I think your new regex is free of this problem, because of the simpler handling of the `(...)` parameters!

If you'd like to learn more about exponential regexes, the CodeQL docs for the alert you saw are pretty good: https://codeql.github.com/codeql-query-help/javascript/js-redos/. Exponential backtracking usually happens if you have something like `(pattern1|pattern2)*`, because at each step of matching the next character in the `*` sequence, we don't know whether we just saw a `pattern1` or a `pattern2`, and have to backtrack to try both options if the next match fails.

extensions/ql-vscode/CHANGELOG.md
marcnjaramillo: I like it. I tried to make it concise, but I should have included more information.

extensions/ql-vscode/src/extension.ts
marcnjaramillo: Moved it to be after `languageSupport.install()`, but if you think it should go further down in the function I can move it again.

extensions/ql-vscode/src/quickEvalCodeLensProvider.ts
marcnjaramillo: Thanks for pointing this out. I got it fixed now. :)

extensions/ql-vscode/src/quickEvalCodeLensProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+      if (matches) {
+        const range: Range = new Range(
+          textLine.range.start.line, matches.index!,
+          textLine.range.end.line, matches.index! + 1
+        );
+        console.log(range);
+
+        const command: Command = {
+          command: 'codeQL.codeLensQuickEval',
+          title: `Quick Evaluation: ${matches![1]}`,
```
aeisenberg, line 31: Type assertion no longer necessary.

```suggestion
          title: `Quick Evaluation: ${matches[1]}`,
```

aeisenberg (commented):
Looking really good. Just a couple of more small cleanups.

extensions/ql-vscode/src/quickEvalCodeLensProvider.ts
marcnjaramillo: Change has been made.

extensions/ql-vscode/src/quickEvalCodeLensProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+  CodeLensProvider,
+  TextDocument,
+  CodeLens, 
+  Command,
+  Range
+} from 'vscode';
+  
+class QuickEvalCodeLensProvider implements CodeLensProvider {
+  // Each provider requires a provideCodeLenses function which will give the various documents
+  // the code lenses
```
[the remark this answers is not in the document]
marcnjaramillo: Removed this 👍🏻

extensions/ql-vscode/src/quickEvalCodeLensProvider.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    const codeLenses: CodeLens[] = [];
+
+    for (let index = 0; index < document.lineCount; index++) {
+      const textLine = document.lineAt(index);
+      // Match a predicate signature, including predicate name, parameter list, and opening brace.
+      const regex = new RegExp(/(\w+)\s*\(\s*.*(?:,\s*)*\)\s*\{/);
+      const matches = textLine.text.match(regex);
+
+      if (matches) {
```
aeisenberg, line 20: Ensure that commented code removes the code lens. It's too tricky to do in a single regex, so I think doing another check is ok.

```suggestion
      if (matches && textLine.text.search(/^\s*\//)) {
```

aeisenberg (commented):
Looks good, but I can't approve since I was the one who created the PR.

aeisenberg:
@adityasharad Can you approve? I can't since I created this PR.

extensions/ql-vscode/src/quickEvalCodeLensProvider.ts
adityasharad: We had an illuminating discussion offline about why this works (`search` returns 0 when it matches the start of the line and -1 on mismatch, which JS converts into `false` and `true` respectively). I suggested a slightly more verbose but arguably simpler use of `test` instead -- hopefully that's still captured the meaning you suggested here!

aeisenberg: A downside with the suggestion is that the regex is evaluated twice for every line, even if there is no initial match. The original suggestion only ran the second regex _if_ the first one matched. I'm not sure if the extra overhead would be noticeable, but this code is run after every file change (in a background thread).

adityasharad: That's a good point. We can keep using `!test` rather than `search`, but keep it inline instead of in its own variable.
So `if (matches && /.../.test(textLine.text)`

aeisenberg: Yep....that makes sense.

aeisenberg (commented):
:shipit: 

(I can't approve.)

extensions/ql-vscode/src/quickEvalCodeLensProvider.ts
marcnjaramillo: Made this change. Sorry it took a bit. I forgot I had resolved the `CHANGELOG` conflict from here but I hadn't updated my local branch yet. Had to get that squared away. Does it look better?

adityasharad (approved):

PR #1037, opened 2021-12-07
Be nicer about where to open the results webview

aeisenberg:
Currently, the webview _always_ opens next to the currently active
editor. This is a pain if you already have 2 columns open since this
means that the webview will open in a third column, which is rarely
what you want.

This change uses a more sophisticated approach to opening the webview:

1. If there is only one column, open webview to the right of it
2. If there are multiple columns and the active editor is _not_ the
   last column, open to the right of the active editor
3. Otherwise open in the first column.

This will avoid opening a new column unless there is only one column
open right now.

There is no native API that vscode exposed to compare column locations,
so this uses the `ViewColumn` api is a slightly non-standard way.

A limitation is that if the last column is empty and the active editor
is to the left of it, then the webview will not be opened there (which
would be nice). Instead, it will be opened in column 1.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-- Avoid creating a third column when opening the results view. The results view will always open to the right of the active editor, unless the active editor is in the rightmost editor column. In that case open in the leftmost column. [#1037](https://github.com/github/vscode-codeql/pull/1037)]
+- Avoid creating a third column when opening the results view. The results view will always open to the right of the active editor, unless the active editor is in the rightmost editor column. In that case open in the leftmost column. [#1037](https://github.com/github/vscode-codeql/pull/1037)
 - Add a CodeLens to make the Quick Evaluation command more accessible. Click the `Quick Evaluation` prompt above a predicate definition in the editor to evaluate that predicate on its own. [#1035](https://github.com/github/vscode-codeql/pull/1035)
 - Fix a bug where the _Alerts_ option would show in the results view even if there is no alerts table available. [#1038](https://github.com/github/vscode-codeql/pull/1038)
 
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/interface.ts
```diff
@@ -203,6 +204,29 @@ export class InterfaceManager extends DisposableObject {
...
+   * If there is a single view column, then open beside it.
+   * If there are multiple view columns, then open beside the active column,
+   * unless the active editor is the last column. In this case, open in the first column.
+   *
+   * The goal is to avoid opening new columns when there already are two columns open.
+   */
+  private chooseColumnForWebview(): vscode.ViewColumn {
+    // This is not a great way to determine the number of view columns, but I
+    // can't find a vscode API that does it any better.
+    // Here, iterate through all the visible editors and determing the max view column.
```
shati-patel, line 219: minor :) 
```suggestion
    // Here, iterate through all the visible editors and determine the max view column.
```

shati-patel (commented):
I wasn't sure how to review the actual code, but I tried opening webviews in various locations locally, and that worked as expected! 🕸️ 

Could you add a changelog entry for this PR?

aeisenberg:
@shati-patel can I get another look at this one?

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [UNRELEASED]
 
+- Avoid creating a third column when opening the results view. The results view will always open to the right of the active editor, unless the active editor is in the rightmost editor column. In that case open in the leftmost column. [#1037](https://github.com/github/vscode-codeql/pull/1037)]
```
shati-patel, line 5: Extra bracket!

```suggestion
- Avoid creating a third column when opening the results view. The results view will always open to the right of the active editor, unless the active editor is in the rightmost editor column. In that case open in the leftmost column. [#1037](https://github.com/github/vscode-codeql/pull/1037)
```

shati-patel (approved):
Ooops sorry, I completely forgot to approve this 🙈 

Thanks for the ping!!!

extensions/ql-vscode/CHANGELOG.md
aeisenberg: Thanks for your eagle eyes.

PR #1039, opened 2021-12-07
Made the repo list in the remote query view expandable

charisk:
Provides a view all/view less functionality on the repo list in the remote queries view, if there are more than 10 results.

https://user-images.githubusercontent.com/311693/145059360-ed9cf7fa-0a4e-44ef-a8e9-f6c02f81fe2d.mov

## Checklist


extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -58,6 +60,9 @@ export function RemoteQueries(): JSX.Element {
     return <div>Waiting for results to load.</div>;
   }
 
+  const [repoListExpanded, setRepoListExpanded] = useState(false);
```
shati-patel, line 63: I'm still not sure what `useState(...)` means 😵 Did you find any useful explanations about it?

(Don't worry if not, I'm willing to accept that it does the right thing either way 😂)

shati-patel (approved):
Looks good! Thanks for fixing this ✨

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
adityasharad: Quick reference since we were discussing this offline: https://reactjs.org/docs/hooks-state.html.

The idea here is that we declare a new variable `repoListExpanded` that becomes part of the active "state" of the component, together with a setter function `setRepoListExpanded` that we call every time we want to update the value (instead of mutating the variable directly). The argument to `useState` (`false` here) is the initial value of the variable.

I don't work with React that often, so would love to hear how Charis (and others!) think about React state too.

aeisenberg: Yep...that's right. This helps the react engine know exactly what parts of the UI need redrawing and refreshing. If the state of a component changes, then it needs to be refreshed (and all child components). Other components do not need to be redrawn.

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -79,12 +84,18 @@ export function RemoteQueries(): JSX.Element {
       </div>
 
       <ul className="vscode-codeql__results-list">
-        {queryResult.results.map(result =>
+        {queryResult.results.slice(0, numOfReposToShow).map((result, i) =>
```
aeisenberg, line 87: Can we ever get in a situation with 0 results? If so, maybe we want to show a special message here.

Not something that is critical right away, but something to think about.

aeisenberg (approved):
LGTM

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
charisk: That is a great point - I'll raise a UX question!

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
charisk: Thanks everyone!

@shati-patel please resolve the conversation if you're happy or grab me to discuss further if not! :)

PR #1041, opened 2021-12-08
Remote queries: Open query file/text from webview

shati-patel:
Adds some interaction to these buttons in the remote queries webview:

image

- The first one opens the query file that was run, if it exists (similar to what we do in the local results view). 
- The second one opens the exact query text that was run (in a virtual "read-only" file).

### Follow up

- Consolidate the `remote-query` and `codeql` text providers.
- Once #1037 is merged, extract that behaviour so it is used wherever we are opening editors.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
@@ -21,6 +21,8 @@ import { RemoteQuery } from './remote-query';
 import { RemoteQueryResult as RemoteQueryResultViewModel } from './shared/remote-query-result';
 import { AnalysisResult as AnalysisResultViewModel } from './shared/remote-query-result';
 import { showAndLogWarningMessage } from '../helpers';
+import { URLSearchParams } from 'url';
+import { SHOW_QUERY_TEXT_MSG } from '../query-history';
 
 export class RemoteQueriesInterfaceManager {
   private panel: WebviewPanel | undefined;
@@ -65,7 +67,7 @@ export class RemoteQueriesInterfaceManager {
       queryTitle: query.queryName,
       queryFileName: queryFileName,
       queryFilePath: query.queryFilePath,
-      queryTextTmpFilePath: query.queryTextTmpFilePath,
+      queryText: query.queryText,
       totalRepositoryCount: query.repositories.length,
       affectedRepositoryCount: affectedRepositories.length,
       totalResultCount: totalResultCount,
@@ -142,6 +144,22 @@ export class RemoteQueriesInterfaceManager {
     }
   }
 
+  private async openVirtualFile(text: string) {
+    try {
+      const params = new URLSearchParams({
+        queryText: encodeURIComponent(SHOW_QUERY_TEXT_MSG + text)
+      });
+      const uri = Uri.parse(
+        `remote-query:query-text.ql?${params.toString()}`,
+        true
+      );
+      const doc = await workspace.openTextDocument(uri);
+      await Window.showTextDocument(doc, { preview: false });
+    } catch (error) {
+      void showAndLogWarningMessage('Could not open query text');
+    }
+  }
+
   private async handleMsgFromView(
     msg: FromRemoteQueriesMessage
   ): Promise<void> {
@@ -158,6 +176,9 @@ export class RemoteQueriesInterfaceManager {
         break;
       case 'openFile':
         await this.openFile(msg.filePath);
+        break;
+      case 'openVirtualFile':
+        await this.openVirtualFile(msg.queryText);
         break;
       default:
         assertNever(msg);
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
@@ -4,6 +4,7 @@ import {
   window as Window,
   ViewColumn,
   Uri,
+  workspace,
 } from 'vscode';
 import * as path from 'path';
 
@@ -62,6 +63,8 @@ export class RemoteQueriesInterfaceManager {
     return {
       queryTitle: query.queryName,
       queryFile: queryFile,
+      queryPath: query.queryFilePath,
+      queryTextTmpFile: query.queryTextTmpFile,
       totalRepositoryCount: query.repositories.length,
       affectedRepositoryCount: affectedRepositories.length,
       totalResultCount: totalResultCount,
@@ -129,6 +132,11 @@ export class RemoteQueriesInterfaceManager {
     });
   }
 
+  public async openFile(filePath: string) {
+    const textDocument = await workspace.openTextDocument(filePath);
+    await Window.showTextDocument(textDocument, ViewColumn.One);
+  }
+
   private async handleMsgFromView(
     msg: FromRemoteQueriesMessage
   ): Promise<void> {
@@ -142,6 +150,9 @@ export class RemoteQueriesInterfaceManager {
         void this.logger.log(
           `Remote query error: ${msg.error}`
         );
+        break;
+      case 'openFile':
+        await this.openFile(msg.filePath);
         break;
       default:
         assertNever(msg);
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
@@ -20,6 +20,7 @@ import { AnalysisResult, RemoteQueryResult } from './remote-query-result';
 import { RemoteQuery } from './remote-query';
 import { RemoteQueryResult as RemoteQueryResultViewModel } from './shared/remote-query-result';
 import { AnalysisResult as AnalysisResultViewModel } from './shared/remote-query-result';
+import { showAndLogWarningMessage } from '../helpers';
 
 export class RemoteQueriesInterfaceManager {
   private panel: WebviewPanel | undefined;
@@ -54,7 +55,7 @@ export class RemoteQueriesInterfaceManager {
    * @returns A fully created view model.
    */
   private buildViewModel(query: RemoteQuery, queryResult: RemoteQueryResult): RemoteQueryResultViewModel {
-    const queryFile = path.basename(query.queryFilePath);
+    const queryFileName = path.basename(query.queryFilePath);
     const totalResultCount = queryResult.analysisResults.reduce((acc, cur) => acc + cur.resultCount, 0);
     const executionDuration = this.getDuration(queryResult.executionEndTime, query.executionStartTime);
     const analysisResults = this.buildAnalysisResults(queryResult.analysisResults);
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/remote-queries/shared/remote-query-result.ts
```diff
@@ -1,6 +1,8 @@
 export interface RemoteQueryResult {
   queryTitle: string;
   queryFile: string;
```
charisk, line 3: Now that we need the query name as well as the file path, I think `queryFile` isn't a good name any more. 

How about we make the following renames:
* queryFile -> queryName
* queryPath -> queryFilePath
* queryTextTmpFile -> queryTextTmpFilePath

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -59,6 +68,14 @@ export function RemoteQueries(): JSX.Element {
   }
 
   try {
+    const openQueryFile = () => {
+      openFile(queryResult.queryPath);
```
charisk, line 72: What happens if the file is deleted before the results come back? Does VS Code handle it gracefully?

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -583,9 +583,13 @@ function buildRemoteQueryEntity(
     return { owner: owner, name: repo };
   });
 
+  // TODO: Get query text from query file and save it in a temporary .ql file. 
+  const queryTextTmpFile = '';
```
charisk, line 587: Might be worth writing a file here with some dummy text to help with our testing (e.g. we'll see the file opening, it just won't have the full text).

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
```diff
@@ -129,6 +132,11 @@ export class RemoteQueriesInterfaceManager {
     });
   }
 
+  public async openFile(filePath: string) {
```
charisk, line 135: Since this is only used inside the class, it can be made private.

```suggestion
  private async openFile(filePath: string) {
```

extensions/ql-vscode/src/remote-queries/shared/remote-query-result.ts
shati-patel: Done! Though I've used queryFileName instead of queryName (to avoid naming conflicts).

shati-patel:
Latest commit adds a temporary file containing exact query text.

In the local version of showing query text, we create a "text provider" and write to a read-only document: https://github.com/github/vscode-codeql/blob/6195c6552f4f94890b70d791b657ea73b68d9b97/extensions/ql-vscode/src/query-history.ts#L343-L356

I couldn't work out how to re-use that code, or just import the relevant parts (creating a read-only doc), so I've gone with a more lightweight approach of writing to a temporary file for now. Happy to adjust if anyone has suggestions! 🙇🏽‍♀️

extensions/ql-vscode/src/remote-queries/shared/remote-query-result.ts
charisk: That's a good call, I was actually conflating two different things.

charisk:
> In the local version of showing query text, we create a "text provider" and write to a read-only document

Do you want to have a look at that together?

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -567,14 +567,14 @@ export async function getResultIndex(
   }
 }
 
-function buildRemoteQueryEntity(
+async function buildRemoteQueryEntity(
```
charisk, line 570: Can we prepare the file before calling this function and then just passing the file path?  If we do that then the function's responsibility will remain as is (just builds an entity based on some data), rather than also doing stuff like creating files etc.

We can even put the file writing stuff in a function e.g. `writeQueryTextFile()` and then have `runRemoteQuery()` call that, and then `buildRemoteQueryEntity()`.

Does that make sense?

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -72,8 +89,16 @@ export function RemoteQueries(): JSX.Element {
...
-        <span className="vscode-codeql__query-file">{octicons.file} <span>{queryResult.queryFile}</span></span>
-        <span>{octicons.codeSquare} <span>query</span></span>
+        <span className="vscode-codeql__query-file">{octicons.file}
+          <a className="vscode-codeql__query-file-link" href="#" onClick={openQueryFile}>
+            {queryResult.queryFileName}
+          </a>
+        </span>
+        <span>{octicons.codeSquare}
+          <a className="vscode-codeql__query-file-link" href="#" onClick={openQueryTextTmpFile}>
+            query
```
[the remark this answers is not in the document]
shati-patel: I've left it as `query` for now (to match the design doc), but it might be worth changing in future, if it's unclear!

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -583,9 +583,26 @@ function buildRemoteQueryEntity(
     return { owner: owner, name: repo };
   });
 
+  // Get the query text from query file and save it in a temporary .ql file. 
+  const queryTextTmpFilePath = path.join(tmpDir.name, `tmp-${queryName}`);
+  const queryText = await fs.readFile(queryFilePath, 'utf8');
+  await fs.writeFile(
```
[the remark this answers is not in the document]
shati-patel: Moved to a different approach (using a virtual file), so this temporary file no longer exists! ☁️ 

(There's still the general problem that remote query data doesn't persist across restarts, but that's beyond this PR 😅  We have a separate issue for that somewhere)

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
shati-patel: Given that our new approach just passes in the `queryText` now (and no longer writes to a file), I haven't refactored this. Let me know if it's still worth changing though!

shati-patel:
> > In the local version of showing query text, we create a "text provider" and write to a read-only document
> 
> Do you want to have a look at that together?

We worked through this using the existing [`query-history.ts`](https://github.com/github/vscode-codeql/blob/6195c6552f4f94890b70d791b657ea73b68d9b97/extensions/ql-vscode/src/query-history.ts#L343-L356) example and the VS Code [virtual doc sample](https://github.com/microsoft/vscode-extension-samples/blob/main/virtual-document-sample/src/extension.ts) as a guide 🐄 🎉 


I've registered a new "remote-query" text provider to let us write to a readonly file. Ideally, I'd like to slightly refactor the [QueryHistoryManager](https://github.com/github/vscode-codeql/blob/6195c6552f4f94890b70d791b657ea73b68d9b97/extensions/ql-vscode/src/query-history.ts#L343-L356) code, so that we use a single `codeql` text provider everywhere (instead of registering two separate ones). To avoid this PR getting too big, I'm happy to do that in a separate PR, unless there's a more urgent need to do it all together here!

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -64,6 +66,20 @@ export function RemoteQueries(): JSX.Element {
...
 
   try {
+    const openQueryFile = () => {
+      vscode.postMessage({
+        t: 'openFile',
+        filePath: queryResult.queryFilePath
+      });
+    };
+
+    const openQueryTextTmpFile = () => {
```
charisk, line 76: I think this name was a leftover from the previous approach. Shall we call it `openQueryTextVirtualFile`?

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -64,6 +66,20 @@ export function RemoteQueries(): JSX.Element {
   const numOfReposToShow = repoListExpanded ? queryResult.results.length : numOfReposInContractedMode;
 
   try {
+    const openQueryFile = () => {
```
charisk, line 69: These functions don't need to be inside the try block.

charisk (approved):
Looks great! Only super minor things.

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -72,8 +88,16 @@ export function RemoteQueries(): JSX.Element {
         ({queryResult.executionDuration}), {queryResult.executionTimestamp}
       </p>
       <p className="vscode-codeql__paragraph">
-        <span className="vscode-codeql__query-file">{octicons.file} <span>{queryResult.queryFile}</span></span>
-        <span>{octicons.codeSquare} <span>query</span></span>
+        <span className="vscode-codeql__query-file">{octicons.file}
+          <a className="vscode-codeql__query-file-link" href="#" onClick={openQueryFile}>
```
charisk, line 92: Sorry I didn't think about this before pressing "Approve", but I think these may need some styling tidying now that they become links (so that it's like normal text and the visited state doesn't affect the colour).

Or, perhaps more semantically correct if they're buttons rather than links?

charisk (changes requested):
Lets make sure the styling is correct for the new buttons.

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
```diff
@@ -129,6 +135,31 @@ export class RemoteQueriesInterfaceManager {
     });
   }
 
+  private async openFile(filePath: string) {
+    try {
+      const textDocument = await workspace.openTextDocument(filePath);
+      await Window.showTextDocument(textDocument, ViewColumn.One);
```
aeisenberg, line 141: Eventually, it would be nice to do something like https://github.com/github/vscode-codeql/pull/1037, but that can happen later. Once #1037 gets in, we can work on extracting the behaviour so it is used wherever we are opening editors.

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
```diff
@@ -129,6 +135,31 @@ export class RemoteQueriesInterfaceManager {
...
+    try {
+      const params = new URLSearchParams({
+        queryText: encodeURIComponent(SHOW_QUERY_TEXT_MSG + text)
+      });
+      const uri = Uri.parse(
+        `remote-query:query-text.ql?${params.toString()}`,
+        true
+      );
+      const doc = await workspace.openTextDocument(uri);
+      await Window.showTextDocument(doc, { preview: false });
```
aeisenberg, line 157: Is this read-only? It's possible to make happen, but I forget how.

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
shati-patel: Oops, good point, thanks for spotting that! I've added some styling, which keeps the text the same colour in all situations.

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
shati-patel: Indeed, it's read-only! That's what the `remote-query` prefix and the corresponding [provideTextDocumentContent](https://github.com/github/vscode-codeql/pull/1041/files#diff-8ab69a78d0894d504f6993ec074207e3afe8cbcdddf051ba50b98dd2cecc980cR996) function do 😃

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
shati-patel: Oops, good point, thanks for spotting that! I've added some styling, which keeps the text the same colour in all situations.

charisk (approved):

PR #1042, opened 2021-12-09
Bump cli version for integration tests

aeisenberg:


adityasharad (approved):
Make sure to update `ensureCli.ts` too!

aeisenberg:
This test will fail until we have the v2.7.3 tags applied to the different repositories.

PR #1046, opened 2021-12-13
Implement download behaviour in remote queries view

charisk:
Updates the download buttons in the remote queries view to download the results.

Once the file has been downloaded, a notification is shown

image

## Summary of changes
* Introduced the notion of a `DownloadLink` and used it in `RemoteQueryResult` entities and elsewhere.
* Introduced a `ResultIndex` entity that contains all the information we retrieve from the result-index file, as well as any other information needed from the workflow/API.
* Updated the actions API client to build up this new `ResultIndex` entity and also set DownloadLink information where appropriate.
* Updated the web view to send a message to the extension to initiate downloads.

Note that I think error handling in the API client can be improved, and I intend to do that in a separate PR.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/download-link.ts
+++ b/extensions/ql-vscode/src/remote-queries/download-link.ts
@@ -3,13 +3,13 @@
  */
 export interface DownloadLink {
   /**
-   *  A unique id of the file/artifact being downloaded. 
+   *  A unique id of the artifact being downloaded. 
    */
   id: string;
 
   /**
    * The URL path to use against the GitHub API to download the
-   * linked file/artifact. 
+   * linked artifact. 
    */
   urlPath: string;
 
--- a/extensions/ql-vscode/src/remote-queries/download-link.ts
+++ b/extensions/ql-vscode/src/remote-queries/download-link.ts
@@ -1,0 +1,20 @@
+/**
+ * Represents a link to an item to be downloaded. 
+ */
+export interface DownloadLink {
+  /**
+   *  A unique id of the file/artifact being downloaded. 
+   */
+  id: string;
+
+  /**
+   * The URL path to use against the GitHub API to download the
+   * linked file/artifact. 
+   */
+  urlPath: string;
+
+  /**
+   * An optional path to follow inside the downloaded directory.
+   */
+  innerFilePath?: string;
+}
--- a/extensions/ql-vscode/src/remote-queries/download-link.ts
+++ b/extensions/ql-vscode/src/remote-queries/download-link.ts
@@ -1,5 +1,5 @@
 /**
- * Represents a link to an item to be downloaded. 
+ * Represents a link to an artifact to be downloaded. 
  */
 export interface DownloadLink {
   /**
@@ -14,7 +14,7 @@ export interface DownloadLink {
   urlPath: string;
 
   /**
-   * An optional path to follow inside the downloaded directory.
+   * An optional path to follow inside the downloaded archive containing the artifact.
    */
   innerFilePath?: string;
 }
--- a/extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
+++ b/extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
@@ -6,8 +6,11 @@ import { Credentials } from '../authentication';
 import { logger } from '../logging';
 import { tmpDir } from '../run-queries';
 import { RemoteQueryWorkflowResult } from './remote-query-workflow-result';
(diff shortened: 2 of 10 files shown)


extensions/ql-vscode/src/remote-queries/download-link.ts
```diff
@@ -0,0 +1,20 @@
...
+  id: string;
+
+  /**
+   * The URL path to use against the GitHub API to download the
+   * linked file/artifact. 
+   */
+  urlPath: string;
+
+  /**
+   * An optional path to follow inside the downloaded directory.
```
aeisenberg, line 17: Is this more correct?
```suggestion
   * An optional path to follow inside the downloaded archive containing the artifact.
```

extensions/ql-vscode/src/remote-queries/download-link.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Represents a link to an item to be downloaded.
```
aeisenberg, line 2: Are these all GitHub actions _artifacts_? If so, are they always archive files that need to be unzipped before using them?
```suggestion
 * Represents a link to an artifact to be downloaded. 
```

extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
```diff
@@ -142,19 +207,22 @@ async function downloadArtifact(
...
 
+async function saveFile(filePath: string, data: ArrayBuffer): Promise<void> {
+  void logger.log(`Saving file to ${filePath}`);
+  await fs.writeFile(filePath, Buffer.from(data));
+}
+
+async function unzipFile(sourcePath: string, destinationPath: string) {
+  void logger.log(`Unzipping file to ${destinationPath}`);
+  const file = await unzipper.Open.file(sourcePath);
+  await file.extract({ path: destinationPath });
```
aeisenberg, line 223: Careful here...unzipping can be tricky sometimes. There are certain zip files that cannot be extracted this way. See the workaround here: https://github.com/github/vscode-codeql/blob/1014c4bdda3424d3dd8b586e44391b742c1ceaa4/extensions/ql-vscode/src/databaseFetcher.ts#L285-L289

What you have _might_ work, but then again, it may fail unexpectedly on some larger files.

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -52,14 +54,19 @@ export class RemoteQueriesManager {
     const executionEndTime = new Date();
 
     if (queryResult.status === 'CompletedSuccessfully') {
-      const resultIndexItems = await this.downloadResultIndex(credentials, query);
+      const resultIndex = await getRemoteQueryIndex(credentials, query);
+      if (!resultIndex) {
+        await showAndLogErrorMessage(`There was an issue retrieving the result for the query ${query.queryName}`);
```
aeisenberg, line 59: Is there a reason to wait for the popup to close?
```suggestion
        void showAndLogErrorMessage(`There was an issue retrieving the result for the query ${query.queryName}`);
```

extensions/ql-vscode/src/remote-queries/download-link.ts
aeisenberg: Also below, can we standardize on using `artifact` instead of file or item?

It's not a major point. I'm just trying to clarify terms.

charisk: It's a good point, and agree terms should be clear! I didn't want to make it too specific to Actions but I think I just made it confusing. I've updated to use artifact!

shati-patel (approved):
This looks good to me! (Assuming @aeisenberg's comments are addressed 📝)

<sub>PS: thank you for dealing with the conflicts I kindly introduced in #1041 🙃 </sub>

extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
aeisenberg: Arrgh...sorry. I think what you have is fine. I just wasn't reading it carefully. I'm not particularly happy with this library as we've had so many problems with it in the past. Particularly, large zips (> 2GB) can't be handled by the library. But, that isn't a problem here.

aeisenberg (approved):

extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
charisk: That is good to know, thank you!

PR #1048, opened 2021-12-14
Remote queries: No results view

charisk:
Shows a different summary message ("No results found") if there are no results for a query.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
+++ b/extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
@@ -6,6 +6,7 @@ import { AnalysisResult, RemoteQueryResult } from '../shared/remote-query-result
 import * as octicons from '../../view/octicons';
 
 import { vscode } from '../../view/vscode-api';
+import { DownloadLink } from '../download-link';
 
 const numOfReposInContractedMode = 10;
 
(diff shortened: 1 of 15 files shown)


shati-patel (approved):
Thanks, looks good! 🌷 (Tried locally as well, and got the right views 🖼️ )

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -111,13 +127,11 @@ export function RemoteQueries(): JSX.Element {
...
-        <h2 className="vscode-codeql__query-summary-title">Repositories with results ({queryResult.affectedRepositoryCount}):</h2>
-        <a className="vscode-codeql__summary-download-link vscode-codeql__download-link"
-          onClick={() => download(queryResult.downloadLink)}>
-          {octicons.download}Download all
-        </a>
-      </div>
+      {
+        queryResult.affectedRepositoryCount === 0
+          ? <SummaryNoResults />
+          : <SummaryWithResults {...queryResult} />
```
aeisenberg, line 133: Is there a reason to create a copy of `queryResult` here? I guess it's safer than being able to accidentally tamper with the object. Just want to know what you're thoughts are here.

aeisenberg (approved):
Looks good.

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
charisk: I could be wrong, but I think the alternative way of doing this would mean having to write all the property names out. Using the spread operator creates all the props based on the object we pass. So it's about readability rather than immutability.

PR #1049, opened 2021-12-14
Remote queries: Get query name from metadata (if possible)

shati-patel:
In the remote queries webview, display the actual query name (from the metadata) if available. Otherwise, fall back to the file name: 

image

image

## Notes

I started from the existing [`getQueryName()`](https://github.com/github/vscode-codeql/blob/bf350779c9d1a722768ccbdc5d2da8418b5e6862/extensions/ql-vscode/src/query-results.ts#L158-L166) function, which does this exact thing for local queries. However, the types don't match up, so I couldn't work out how to re-use it directly 😕 

Instead, I separated out the code that gets query metadata, and re-used that for remote queries. Other suggestions are also welcome! 😇 

cc @charisk 

### Follow-up

- Add a test for the metadata function
- Restructure/rename the function to make things cleaner (see https://github.com/github/vscode-codeql/pull/1049#discussion_r769411121)

## Checklist

N/A - internal-only, no user-visible impact.

### DIFF
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -334,17 +334,17 @@ async function convertToQlPath(filePath: string): Promise<string> {
 
 /** Gets the selected position within the given editor. */
 async function getSelectedPosition(editor: TextEditor, range?: Range): Promise<messages.Position> {
-    const selectedRange = range || editor.selection;
-    const pos = selectedRange.start;
-    const posEnd = selectedRange.end;
-    // Convert from 0-based to 1-based line and column numbers.
-    return {
-      fileName: await convertToQlPath(editor.document.fileName),
-      line: pos.line + 1,
-      column: pos.character + 1,
-      endLine: posEnd.line + 1,
-      endColumn: posEnd.character + 1
-    };
+  const selectedRange = range || editor.selection;
+  const pos = selectedRange.start;
+  const posEnd = selectedRange.end;
+  // Convert from 0-based to 1-based line and column numbers.
+  return {
+    fileName: await convertToQlPath(editor.document.fileName),
+    line: pos.line + 1,
+    column: pos.character + 1,
+    endLine: posEnd.line + 1,
+    endColumn: posEnd.character + 1
+  };
 }
 
 /**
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/run-queries.ts
```diff
@@ -334,17 +334,17 @@ async function convertToQlPath(filePath: string): Promise<string> {
...
-    const posEnd = selectedRange.end;
-    // Convert from 0-based to 1-based line and column numbers.
-    return {
-      fileName: await convertToQlPath(editor.document.fileName),
-      line: pos.line + 1,
-      column: pos.character + 1,
-      endLine: posEnd.line + 1,
-      endColumn: posEnd.character + 1
-    };
+  const selectedRange = range || editor.selection;
```
shati-patel, line 337: (sure, let's have some random whitespace changes 🤷🏽 )

adityasharad (approved):
Looks very sensible.

aeisenberg (approved):

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -718,6 +712,22 @@ const compilationFailedErrorTail = ' compilation failed. Please make sure there
   ' and the query and database use the same target language. For more details on the error, go to View > Output,' +
   ' and choose CodeQL Query Server from the dropdown.';
 
+/**
+ * Gets metadata for a query, if it exists.
+ * @param cliServer The CLI server.
+ * @param queryPath The path to the query.
+ * @returns A promise that resolves to the query metadata, if available.
+ */
+export async function getQueryMetadata(cliServer: cli.CodeQLCliServer, queryPath: string): Promise<QueryMetadata | undefined> {
```
charisk, line 721: I have two small suggestions, but they shouldn't block this PR and perhaps can trigger wider discussion:
* Move this function to another module (file) e.g. query-metadata-utils or query-metadata-resolver or something like that. I generally have a preference for smaller, more specific/targetted modules and this one is at 700+ lines of code.
* Call this function `tryGetQueryMetadata` to indicate that it will try its best but will not blow up if it can't get the metadata. This can be a general pattern we follow.

charisk (approved):
Awesome work!

shati-patel:
Thanks for the reviews! I'll merge this now and follow-up with tests/tidy-up later 🧹

PR #1057, opened 2021-12-17
Delete output folder before building

### LINKED ISSUE #1056: Build not clearing output directory
When running a build locally, the output directory isn't cleared before rebuilding, which can lead to stale files and confusion.

For example, someone could be doing some testing before releasing a new version of the extension and things appear to work due to old files, only finding out about bugs after releasing.

aeisenberg:
Fixes #1056


extensions/ql-vscode/gulpfile.ts/typescript.ts
```diff
@@ -20,6 +21,10 @@ function goodReporter(): ts.reporter.Reporter {
 
 const tsProject = ts.createProject('tsconfig.json');
 
+export function cleanOuput() {
```
adityasharad, line 24: cleanOutput

adityasharad (approved):
Minor typo in function name, otherwise looks good.
How did we get this far without noticing 😆  Thanks @charisk for pointing it out!

PR #1069, opened 2022-01-06
Add helper command for working on the Remote Queries Results view

charisk:
Adds a helper command that shows the Remote Queries Results view using mocked data instead of having to execute a query and wait for the results.

I found having this sort of thing helpful when working on the view so I thought we might want to check it in. If we don't want a command showing up even in canary, we could check in only the sample data and some commented out code. 

## Checklist
I believe the following isn't applicable:

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -291,7 +291,7 @@
       },
       {
         "command": "codeQL.showFakeRemoteQueryResults",
-        "title": "[Internal] CodeQL: Show fake remote query results"
+        "title": "CodeQL: [Internal] Show fake remote query results"
       },
       {
         "command": "codeQL.runQueries",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -291,7 +291,7 @@
       },
       {
         "command": "codeQL.showFakeRemoteQueryResults",
-        "title": "CodeQL: Show fake remote query results"
+        "title": "[Internal] CodeQL: Show fake remote query results"
       },
       {
         "command": "codeQL.runQueries",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -290,6 +290,10 @@
         "title": "CodeQL: Run Remote Query"
       },
       {
+        "command": "codeQL.showFakeRemoteQueryResults",
+        "title": "CodeQL: Show fake remote query results"
+      },
+      {
         "command": "codeQL.runQueries",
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/remote-queries/sample-data.ts
```diff
@@ -0,0 +1,86 @@
+import { RemoteQuery } from './remote-query';
+import { RemoteQueryResult } from './remote-query-result';
+
+export const sampleRemoteQuery: RemoteQuery = {
+  queryName: 'Inefficient regular expression',
+  queryFilePath: '/Users/charisk/dev/vscode-codeql-starter/ql/javascript/ql/src/Performance/ReDoS.ql',
```
shati-patel, line 6: Probably best not to check in your specific file path... It might be possible to point this to the right location for a general user (assuming they're also in the `vscode-codeql-starter` workspace) 🤔 

Or this could just be a dummy path 😎

shati-patel (commented):
Thanks! I'd been using a similar test command when working on the webview, so I agree it would be helpful to check this in 🎉 

It's probably okay to have this visible to canary users, but otherwise commented code also sounds good! (Assuming there's a way to do that, since you can't add comments in the JSON file?)

extensions/ql-vscode/package.json
```diff
@@ -289,6 +289,10 @@
         "command": "codeQL.runRemoteQuery",
         "title": "CodeQL: Run Remote Query"
       },
+      {
+        "command": "codeQL.showFakeRemoteQueryResults",
+        "title": "CodeQL: Show fake remote query results"
```
aeisenberg, line 294: Would be nice to add some indication that this is for extension dev work only. Maybe something like:

```suggestion
        "title": "[Internal] CodeQL: Show fake remote query results"
```

But, you can come up with your own label. It's nice that it only shows for canary users, but many of our internal users are canary users. It would be nice not to confuse them.

extensions/ql-vscode/src/remote-queries/sample-data.ts
charisk: Good catch - I thought I had dummy-fied everything!

charisk: For info, you're totally right that you generally can't have comments in JSON (there are exceptions), so I meant have the code in extension.ts commented out and leave it to to us to add the dummy command in package.json every time. Not an ideal approach, so glad we're going with checking in the command!

shati-patel (approved):
Thanks! 🦖

PR #1070, opened 2022-01-06
Results view: Don't reopen webview if it's already visible

shati-patel:
A user reported a bug where the results view is opened in the active editor, even if the results view is already visible in a different column:
<details><summary> 🖱️  Expand for example GIF</summary>
<p>

34F3083B-05A6-46EF-8954-6D74615756E0

</p>
</details> 

This PR fixes that behaviour. I.e. when there's already a results view visible (not necessarily active), we display the results in that panel. 

### Notes

A slight weirdness we could still find is that if the results view is open in a tab somewhere, but not visible, it can still move to a different place. I don't think this is a problem, but I can open an issue if it is!

<details><summary>🖱 Expand for details</summary>
<p>

In this case, the results view is open on the right, but it isn't currently visible. Since the right-hand side is the active editor, the new results open on the left. (So the results view moves column.)

8003AFBE-3E96-4AB6-9083-33C1FB4D622A

</p>
</details> 

## Checklist

- [n/a] If this pull request makes user-facing changes that require documentation changes, the `ready-for-doc-review` label has been added to this pull request or the corresponding issue.

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Fix a bug where the results view moved column even when it was already visible. [#1070](https://github.com/github/vscode-codeql/pull/1070)
+
 ## 1.5.9 - 17 December 2021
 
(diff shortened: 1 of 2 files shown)


aeisenberg (approved):

Metabori:
asset

PR #1076, opened 2022-01-13
Add "pack install" and "pack download" commands

shati-patel:
Adds commands to install and download packs 📦 

For the "Download Packs" command, I've added a hard-coded option to download all core packs, and a custom option if users want to specify a custom pack. 

<details><summary>Expand for demo GIF</summary>
<p>

2A0768B5-1B85-4653-9664-992D399DEBB8

</p>
</details> 

For "Install Packs", the user can select packs from their workspace

<details><summary>Expand for demo GIF</summary>
<p>

9E18DA5E-DCBD-433E-AF25-F6563BADC1ED

</p>
</details> 

~PS: I will add a test for these commands 🧪 But I'd be grateful for a quick review first to see if this is a plausible way of doing things!~ 🙈 Done!

## Checklist

- [x] Issues have been created for any UI or other user-facing changes made by this pull request (internal issue)
- [x] ~If this pull request makes user-facing changes that require documentation changes, the `ready-for-doc-review` label has been added to this pull request or the corresponding issue.~ Docs PR: https://github.com/github/codeql/pull/7661

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [UNRELEASED]
 
 - Fix a bug where the results view moved column even when it was already visible. [#1070](https://github.com/github/vscode-codeql/pull/1070)
-- Add packaging-related commands. _CodeQL: Download Packs_ downloads packs from the package registry, and _CodeQL: Install Packs_ installs dependencies for packs in your workspace. [#1076](https://github.com/github/vscode-codeql/pull/1076)
+- Add packaging-related commands. _CodeQL: Download Packs_ downloads query packs from the package registry that can be run locally, and _CodeQL: Install Pack Dependencies_ installs dependencies for packs in your workspace. [#1076](https://github.com/github/vscode-codeql/pull/1076)
 
 ## 1.5.9 - 17 December 2021
 
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Fix a bug where the results view moved column even when it was already visible. [#1070](https://github.com/github/vscode-codeql/pull/1070)
+- Add packaging-related commands. _CodeQL: Download Packs_ downloads packs from the package registry, and _CodeQL: Install Packs_ installs dependencies for packs in your workspace. [#1076](https://github.com/github/vscode-codeql/pull/1076)
 
 ## 1.5.9 - 17 December 2021
 
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -374,8 +374,8 @@
         "title": "CodeQL: Clear Cache"
       },
       {
-        "command": "codeQL.installPacks",
-        "title": "CodeQL: Install Packs"
+        "command": "codeQL.installPackDependencies",
+        "title": "CodeQL: Install Pack Dependencies"
       },
       {
         "command": "codeQL.downloadPacks",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -374,6 +374,14 @@
         "title": "CodeQL: Clear Cache"
       },
       {
+        "command": "codeQL.installPacks",
+        "title": "CodeQL: Install Packs"
+      },
+      {
+        "command": "codeQL.downloadPacks",
+        "title": "CodeQL: Download Packs"
+      },
+      {
         "command": "codeQLDatabases.setCurrentDatabase",
         "title": "Set Current Database"
       },
--- a/extensions/ql-vscode/src/packaging.ts
+++ b/extensions/ql-vscode/src/packaging.ts
@@ -1,4 +1,4 @@
-import { CodeQLCliServer } from './cli';
+import { CliVersionConstraint, CodeQLCliServer } from './cli';
 import {
   getOnDiskWorkspaceFolders,
   showAndLogErrorMessage,
@@ -30,6 +30,10 @@ export async function handleDownloadPacks(
   cliServer: CodeQLCliServer,
   progress: ProgressCallback,
 ): Promise<void> {
+  if (!(await cliServer.cliConstraints.supportsPackaging())) {
+    throw new Error(`Packaging commands are not supported by this version of CodeQL. Please upgrade to v${CliVersionConstraint.CLI_VERSION_WITH_PACKAGING
+      } or later.`);
+  }
   progress({
     message: 'Choose packs to download',
     step: 1,
@@ -87,6 +91,10 @@ export async function handleInstallPacks(
   cliServer: CodeQLCliServer,
   progress: ProgressCallback,
 ): Promise<void> {
+  if (!(await cliServer.cliConstraints.supportsPackaging())) {
+    throw new Error(`Packaging commands are not supported by this version of CodeQL. Please upgrade to v${CliVersionConstraint.CLI_VERSION_WITH_PACKAGING
+      } or later.`);
+  }
   progress({
     message: 'Choose packs to install',
     step: 1,
--- a/extensions/ql-vscode/src/packaging.ts
+++ b/extensions/ql-vscode/src/packaging.ts
@@ -37,7 +37,7 @@ export async function handleDownloadPacks(
     maxStep: 2,
   });
   let packsToDownload: string[] = [];
-  const queryPackOption = 'Download core query packs';
+  const queryPackOption = 'Download all core query packs';
   const customPackOption = 'Download custom specified pack';
   const quickpick = await window.showQuickPick(
     [queryPackOption, customPackOption],
@@ -57,7 +57,7 @@ export async function handleDownloadPacks(
       throw new UserCancellationException('No pack specified.');
     }
   }
-  if (packsToDownload && packsToDownload.length > 0) {
+  if (packsToDownload?.length > 0) {
     progress({
       message: 'Downloading packs. This may take a few minutes.',
       step: 2,
--- a/extensions/ql-vscode/src/packaging.ts
+++ b/extensions/ql-vscode/src/packaging.ts
@@ -3,7 +3,6 @@ import {
   getOnDiskWorkspaceFolders,
   showAndLogErrorMessage,
   showAndLogInformationMessage,
-  showAndLogWarningMessage,
 } from './helpers';
 import { QuickPickItem, window } from 'vscode';
 import { ProgressCallback, UserCancellationException } from './commandRunner';
@@ -123,8 +122,8 @@ export async function handleInstallPacks(
     }
     if (failedPacks.length > 0) {
       void logger.log(`Errors:\n${errors.join('\n')}`);
-      void showAndLogWarningMessage(
-        `Unable to install some packs: ${failedPacks.join(', ')}. See logs for more details.`
+      throw new Error(
+        `Unable to install packs: ${failedPacks.join(', ')}. See logs for more details.`
       );
     } else {
       void showAndLogInformationMessage('Finished installing packs.');
--- a/extensions/ql-vscode/src/packaging.ts
+++ b/extensions/ql-vscode/src/packaging.ts
@@ -123,7 +123,7 @@ export async function handleInstallPacks(
     if (failedPacks.length > 0) {
       void logger.log(`Errors:\n${errors.join('\n')}`);
       throw new Error(
-        `Unable to install packs: ${failedPacks.join(', ')}. See logs for more details.`
+        `Unable to install packs: ${failedPacks.join(', ')}. See log for more details.`
       );
     } else {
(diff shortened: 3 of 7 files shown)


extensions/ql-vscode/src/packaging.ts
```diff
@@ -0,0 +1,144 @@
...
+
+const CORE_PACKS = [
+  'codeql/cpp-all',
+  'codeql/csharp-all',
+  'codeql/go-all',
+  'codeql/java-all',
+  'codeql/javascript-all',
+  'codeql/python-all',
+  'codeql/ruby-all',
+];
```
shati-patel, line 20: Hard-coded these, but maybe there's a way to list the "core" packs programmatically?

extensions/ql-vscode/src/packaging.ts
```diff
@@ -0,0 +1,144 @@
...
+  const packs: string[] = [];
+  const workspaceFolders = getOnDiskWorkspaceFolders();
+  for (const folder of workspaceFolders) {
+    const qlpackYml = path.join(folder, 'qlpack.yml');
+    if (fs.pathExistsSync(qlpackYml)) {
+      packs.push(folder);
+    }
+  }
+  return packs;
+}
```
shati-patel, line 38: This is my (bad) attempt to list all packs in the workspace. Currently only looks for `qlpack.yml` files in the root of the workspace, which excludes anything in the `ql` and `codeql-go` submodules in the starter workspace... Does anyone have a suggestion for how to find packs in subfolders? 

My immediate hacky thought was to hard-code the paths to the main packs in the starter workspace, but surely there's a nicer way 😅

extensions/ql-vscode/src/packaging.ts
```diff
@@ -0,0 +1,144 @@
...
+  progress: ProgressCallback,
+): Promise<void> {
+  progress({
+    message: 'Choose packs to install',
+    step: 1,
+    maxStep: 2,
+  });
+  let packsToInstall: string[] = [];
+  const workspacePackOption = 'Install workspace packs';
+  const customPackOption = 'Install custom specified pack';
```
aeisenberg, line 110: I'm not sure how useful this is. Is the goal to allow users to run install on any source pack on their hard drive? Perhaps to keep things simple, it's good enough to only search for packs in the workspace.

shati-patel: Good idea, thanks!

shati-patel:
Thanks for the review! (And for improving my understanding of how people use these packaging commands 😅)

Pushed some updates based your feedback. Next up: I'll add tests! 👩🏽‍🔬

extensions/ql-vscode/src/packaging.ts
```diff
@@ -0,0 +1,135 @@
...
+  cliServer: CodeQLCliServer,
+  progress: ProgressCallback,
+): Promise<void> {
+  progress({
+    message: 'Choose packs to download',
+    step: 1,
+    maxStep: 2,
+  });
+  let packsToDownload: string[] = [];
+  const queryPackOption = 'Download core query packs';
```
aeisenberg, line 40: Maybe should be explicit about downloading all of the packs.

```suggestion
  const queryPackOption = 'Download all core query packs';
```

extensions/ql-vscode/src/packaging.ts
```diff
@@ -0,0 +1,135 @@
...
+        'Enter the <package-scope/name[@version]> of the pack to download',
+      ignoreFocusOut: true,
+    });
+    if (customPack) {
+      packsToDownload.push(customPack);
+    } else {
+      throw new UserCancellationException('No pack specified.');
+    }
+  }
+  if (packsToDownload && packsToDownload.length > 0) {
```
aeisenberg, line 60: nit:

```suggestion
  if (packsToDownload?.length > 0) {
```

extensions/ql-vscode/src/packaging.ts
```diff
@@ -0,0 +1,135 @@
...
+export async function handleInstallPacks(
+  cliServer: CodeQLCliServer,
+  progress: ProgressCallback,
+): Promise<void> {
+  progress({
+    message: 'Choose packs to install',
+    step: 1,
+    maxStep: 2,
+  });
+  const workspacePacks = await cliServer.resolveQlpacks(getOnDiskWorkspaceFolders());
```
aeisenberg, line 96: Nice. Glad to see this is working.

aeisenberg (commented):
Just tried it out. Works beautifully. I didn't know there was a multi-select option for quick picks.

aeisenberg:
The changes I'm suggesting are not blockers, but I will hold off approval until there are some tests.

extensions/ql-vscode/src/packaging.ts
adityasharad: Could you remind me of the intended audience for the new commands? I think that influences the choice of packs.
- Query writers getting set up to write new queries and query packs without having to clone the `codeql` repo?
- Security teams running the default queries and examining the results? (If so, what does VS Code make easier compared to using the command line or code scanning?)
- Both?

aeisenberg:
To answer a question posed by Aditya in a comment (which I can't see right now, was it deleted?). 

`pack install` is primarily useful for query authors who want to get started developing in vscode. It is one step along the way of making the starter repo no longer necessary.

`pack download`, on its own may not be entirely useful since we also need a way to run one or more queries from downloaded packs.

shati-patel:
I'm struggling with tests (again)... 🙈  The functions I'm testing don't return anything, so I'm trying to just check the logging output.
The tests are failing with "TypeError: Cannot read property 'args' of null" and an "unhandled rejection" error, which I haven't been able to fix... 😅

Pushed my attempt so far. @aeisenberg, would you be able to take a look? 🙏🏽

extensions/ql-vscode/src/packaging.ts
```diff
@@ -0,0 +1,135 @@
...
+      } catch (error) {
+        failedPacks.push(pack.label);
+        errors.push(error);
+      }
+    }
+    if (failedPacks.length > 0) {
+      void logger.log(`Errors:\n${errors.join('\n')}`);
+      void showAndLogWarningMessage(
+        `Unable to install some packs: ${failedPacks.join(', ')}. See logs for more details.`
+      );
```
aeisenberg, line 128: Maybe here you can throw the error instead of just logging it. You can then structure your tests around that.

aeisenberg (commented):
It would also be nice to somehow test that the pack was actually downloaded, but there's no way to do it right now. You would need to run `resolve queries`. So, for now, just testing whether the command throws or not is good enough.

extensions/ql-vscode/src/packaging.ts
```diff
@@ -123,8 +122,8 @@ export async function handleInstallPacks(
     }
     if (failedPacks.length > 0) {
       void logger.log(`Errors:\n${errors.join('\n')}`);
-      void showAndLogWarningMessage(
-        `Unable to install some packs: ${failedPacks.join(', ')}. See logs for more details.`
+      throw new Error(
+        `Unable to install packs: ${failedPacks.join(', ')}. See logs for more details.`
```
aeisenberg, line 126: nit: Maybe make it singular since the button below says "Show log".

```suggestion
        `Unable to install packs: ${failedPacks.join(', ')}. See log for more details.`
```

extensions/ql-vscode/src/vscode-tests/cli-integration/packaging.test.ts
```diff
@@ -88,21 +88,22 @@ describe('Packaging commands', function() {
...
-      'Finished installing packs.'
-    );
+    try {
+      await mod.handleInstallPacks(cli, progress);
+      expect(showAndLogInformationMessageSpy.firstCall.args[0]).to.contain(
+        'Finished installing packs.'
+      );
+    } catch (error) {
+      expect(error.message).to.contain('Unable to install packs:');
+    }
```
aeisenberg, line 107: This test should either test an error condition or a happy path. It looks like there are both kinds of assertions. If you want to test an error condition then you should have a negative assertion after the function that is supposed to throw. Something like this:

```suggestion
      await mod.handleInstallPacks(cli, progress);
      expect(true).to.be.false;
    } catch (error) {
      expect(error.message).to.contain('Unable to install packs:');
    }
```

shati-patel: Thanks! I've split out the test into a "good" and a "bad" version ⚡

shati-patel:
I'd forgotten to add a CLI version constraint for packaging 🤦🏽 🤦🏽 🤦🏽 (fixed in the latest commit)

The mysterious failures have now disappeared, so I'm optimistic that this is now in good shape 🤞🏽 😅

aeisenberg (approved):
Nice job!

I have a _slight_ concern that these tests will now be downloading packs into the global package cache. So, if you run it locally, the tests will hit your cache (and likely not actually download anything). But on CI, the cache will be empty on each run, so packages will be downloaded.

I don't think this is an issue since the tests are not really about the CLI, but only about how he extension interacts with the CLI.

extensions/ql-vscode/src/packaging.ts
```diff
@@ -0,0 +1,142 @@
...
+    step: 1,
+    maxStep: 2,
+  });
+  const workspacePacks = await cliServer.resolveQlpacks(getOnDiskWorkspaceFolders());
+  const quickPickItems = Object.entries(workspacePacks).map<QLPackQuickPickItem>(([key, value]) => ({
+    label: key,
+    packRootDir: value,
+  }));
+  const packsToInstall = await window.showQuickPick(quickPickItems, {
+    placeHolder: 'Select packs to install',
```
adityasharad, line 109: The list of packs that will be shown here are the packs in the workspace, but the packs that we'll actually install will be the dependencies of those packs. Have I got that right?
If yes, I would find it useful to see something like "Select packs to install dependencies for" or "Install dependencies for these packs" in the various prompts and log messages here.

extensions/ql-vscode/src/packaging.ts
```diff
@@ -0,0 +1,142 @@
...
+    canPickMany: true,
+    ignoreFocusOut: true,
+  });
+  if (packsToInstall && packsToInstall.length > 0) {
+    progress({
+      message: 'Installing packs. This may take a few minutes.',
```
adityasharad, line 115:
```suggestion
      message: 'Installing pack dependencies. This may take a few minutes.',
```

extensions/ql-vscode/src/packaging.ts
```diff
@@ -0,0 +1,142 @@
...
+    label: key,
+    packRootDir: value,
+  }));
+  const packsToInstall = await window.showQuickPick(quickPickItems, {
+    placeHolder: 'Select packs to install',
+    canPickMany: true,
+    ignoreFocusOut: true,
+  });
+  if (packsToInstall && packsToInstall.length > 0) {
+    progress({
```
adityasharad, line 114: Minor: You know the total number of root packs, so you could make the progress monitor actually reflect the progress through the root packs. Save that for a future PR though :)

extensions/ql-vscode/package.json
```diff
@@ -373,6 +373,14 @@
         "command": "codeQL.clearCache",
         "title": "CodeQL: Clear Cache"
       },
+      {
+        "command": "codeQL.installPacks",
+        "title": "CodeQL: Install Packs"
```
adityasharad, line 378: Install Pack(age) Dependencies?

adityasharad (commented):
Nice! I have a few non-blocking suggestions about user-visible text, to indicate that `pack install` is really installing pack dependencies. Feel free to incorporate them now or consider them for later.

extensions/ql-vscode/src/packaging.ts
shati-patel: Agreed! Updated wording to be explicit about what we're installing 💪🏽

extensions/ql-vscode/src/packaging.ts
shati-patel: Good idea! I will try that in a follow-up PR 🔄

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Fix a bug where the results view moved column even when it was already visible. [#1070](https://github.com/github/vscode-codeql/pull/1070)
+- Add packaging-related commands. _CodeQL: Download Packs_ downloads packs from the package registry, and _CodeQL: Install Packs_ installs dependencies for packs in your workspace. [#1076](https://github.com/github/vscode-codeql/pull/1076)
```
adityasharad, line 6:
```suggestion
- Add packaging-related commands. _CodeQL: Download Packs_ downloads query packs from the package registry that can be run locally, and _CodeQL: Install Pack Dependencies_ installs dependencies for packs in your workspace. [#1076](https://github.com/github/vscode-codeql/pull/1076)
```

adityasharad (approved):

PR #1078, opened 2022-01-17
Extract base react components

charisk:
While working on some Remote Query view changes I found that the code was getting a bit difficult to manage so I've started extracting out some base components and styles. This could be the start of a component library we can re-use across the codebase.

Note that I looked into possibly using a more proper solution like [styled-components](https://styled-components.com/) but decided it wasn't appropriate/worth it. E.g. for styled-components (and I imagine others) we'd have to make a compromise in our CSP implementation to allow inline styles.

## Checklist
N/A

### DIFF
--- a/extensions/ql-vscode/src/interface-utils.ts
+++ b/extensions/ql-vscode/src/interface-utils.ts
@@ -117,13 +117,19 @@ export function tryResolveLocation(
 export function getHtmlForWebview(
   webview: Webview,
   scriptUriOnDisk: Uri,
-  stylesheetUriOnDisk: Uri
+  stylesheetUrisOnDisk: Uri[],
 ): string {
   // Convert the on-disk URIs into webview URIs.
   const scriptWebviewUri = webview.asWebviewUri(scriptUriOnDisk);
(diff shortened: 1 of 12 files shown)


extensions/ql-vscode/src/interface-utils.ts
```diff
@@ -137,7 +143,7 @@ export function getHtmlForWebview(
   <head>
     <meta http-equiv="Content-Security-Policy"
           content="default-src 'none'; script-src 'nonce-${nonce}'; style-src 'nonce-${nonce}'; connect-src ${webview.cspSource};">
-    <link nonce="${nonce}" rel="stylesheet" href="${stylesheetWebviewUri}">
+        ${stylesheetsHtmlLines.join(' ')}
```
aeisenberg, line 146: nit: Would this make the generated html easier to read?

```suggestion
        ${stylesheetsHtmlLines.join('\n    ')}
```

aeisenberg (approved):
This looks great. Small suggestion that you can take or leave.

extensions/ql-vscode/src/interface-utils.ts
charisk: I've tidied that up but used os.EOL instead of `\n`

PR #1083, opened 2022-01-18
Split download actions in remote queries view

charisk:
We currently have a single action to download a results artifact which means that we lose some context about what we're downloading when we reach the extension code. In order to keep that context and make it easier to track which results have been downloaded, I split the action into two separate ones.

## Checklist
N/A


shati-patel (approved):
Looks good! 🌱

PR #1086, opened 2022-01-19
Wait a few seconds before monitoring remote query run

shati-patel:
We sometimes get an error where the extension successfully schedules a remote query run but then fails with an `HttpError: Not Found` error when it checks the workflow status. Waiting a few seconds before starting to check the status might help 🤞🏽 

See internal linked issue for more info!

## Checklist

n/a, internal feature

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-monitor.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-monitor.ts
@@ -29,10 +29,9 @@ export class RemoteQueriesMonitor {
 
     let attemptCount = 0;
 
-    // Wait a bit before starting to monitor the workflow run.
-    await this.sleep(RemoteQueriesMonitor.sleepTime);
-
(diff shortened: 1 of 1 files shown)


cklin:
Instead of adding a new `sleep` statement, how about moving the existing `sleep` statement in the `while` loop a bit further up to the top of the `while` block?

That should have the same effect of introducing an initial delay, with the added benefit of eliminating the unnecessary final delay if the last `getWorkflowStatus()` call still returns `InProgress` status.

shati-patel:
> Instead of adding a new `sleep` statement, how about moving the existing `sleep` statement in the `while` loop a bit further up to the top of the `while` block?
> 
> That should have the same effect of introducing an initial delay, with the added benefit of eliminating the unnecessary final delay if the last `getWorkflowStatus()` call still returns `InProgress` status.

Good point, that makes a lot of sense 😀

cklin (approved):

PR #1087, opened 2022-01-19
Update progress bar for "install pack dependencies"

shati-patel:
As @adityasharad suggested in https://github.com/github/vscode-codeql/pull/1076#discussion_r787067110, we can have more granular progress tracking by counting how many packs to install 📦 

This sort of works, except we can't set the `maxStep` until after the user has selected the packs... So for the first step ([here](https://github.com/shati-patel/vscode-codeql/blob/d534f88fad43410f07b8ed651726f31d4d55649d/extensions/ql-vscode/src/packaging.ts#L98-L102)) I just stuck with "2" as the max number of steps, which means that the first step takes up a disproportionate amount of space in the progress bar. This looks a bit odd, but I don't think it's a big problem 😅 

## Checklist

n/a - just a follow-up to #1076!

### DIFF
--- a/extensions/ql-vscode/src/packaging.ts
+++ b/extensions/ql-vscode/src/packaging.ts
@@ -110,15 +110,19 @@ export async function handleInstallPackDependencies(
     canPickMany: true,
     ignoreFocusOut: true,
   });
-  if (packsToInstall && packsToInstall.length > 0) {
-    progress({
-      message: 'Installing dependencies. This may take a few minutes.',
-      step: 2,
-      maxStep: 2,
-    });
+  const numberOfPacks = packsToInstall?.length || 0;
+  if (packsToInstall && numberOfPacks > 0) {
     const failedPacks = [];
     const errors = [];
+    // Start at 1 because we already have the first step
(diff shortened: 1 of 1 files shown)


aeisenberg (approved):

aeisenberg:
> So for the first step ([here](https://github.com/shati-patel/vscode-codeql/blob/d534f88fad43410f07b8ed651726f31d4d55649d/extensions/ql-vscode/src/packaging.ts#L98-L102)) I just stuck with "2" as the max number of steps, which means that the first step takes up a disproportionate amount of space in the progress bar.

There are similar issues with other progress bars, where we don't know the complete number of steps until later on. I don't think that's a problem we need to (or can) fix (at least without a lot of work).

adityasharad:
The other alternative is adding some progress calls early to "complete" the progress bar once you work out the total number of steps, and then restart the progress bar for the variable steps (e.g. 1/2, 2/2, <work out the total number of steps>, 1/total, 2/total, ..., total/total). It looks a bit odd in the UI and is...not really worth it 😆

PR #1089, opened 2022-01-20
Download and process analyses results

charisk:
Adds logic to download analyses results, do some minimal processing and make them available to the web view.

Note that this solution doesn't work well with running multiple queries but I'm hoping to address that once we have some notion of query history.

Also it does not support bqrs files as there is some Product/UX work that needs to happen.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
@@ -2,7 +2,6 @@ import { ExtensionContext } from 'vscode';
 import { Credentials } from '../authentication';
 import { Logger } from '../logging';
 import { downloadArtifactFromLink } from './gh-actions-api-client';
-import { DownloadLink } from './download-link';
 import * as path from 'path';
 import * as fs from 'fs-extra';
 import { AnalysisSummary } from './shared/remote-query-result';
@@ -11,27 +10,28 @@ import { AnalysisResults, QueryResult } from './shared/analysis-result';
 
 export class AnalysesResultsManager {
   // Store for the results of various analyses for a single remote query.
-  private readonly analysesResults: { [key: string]: QueryResult[] };
+  private readonly analysesResults: AnalysisResults[];
 
   constructor(
     private readonly ctx: ExtensionContext,
     private readonly logger: Logger,
   ) {
-    this.analysesResults = {};
+    this.analysesResults = [];
   }
 
   public async downloadAnalysisResults(
     analysisSummary: AnalysisSummary,
   ): Promise<void> {
+    if (this.analysesResults.some(x => x.nwo === analysisSummary.nwo)) {
+      // We already have the results for this analysis, don't download again.
+      return;
+    }
+
     const credentials = await Credentials.initialize(this.ctx);
 
     void this.logger.log(`Downloading and processing results for ${analysisSummary.nwo}`);
 
-    const queryResults = await this.downloadSingleAnalysisResults(
-      analysisSummary.downloadLink,
-      credentials);
-
-    this.analysesResults[analysisSummary.nwo] = queryResults;
+    await this.downloadSingleAnalysisResults(analysisSummary, credentials);
   }
 
   public async downloadAllResults(
@@ -42,27 +42,31 @@ export class AnalysesResultsManager {
     void this.logger.log('Downloading and processing all results');
 
     for (const analysis of analysisSummaries) {
-      const queryResults = await this.downloadSingleAnalysisResults(analysis.downloadLink, credentials);
-      this.analysesResults[analysis.nwo] = queryResults;
+      await this.downloadSingleAnalysisResults(analysis, credentials);
     }
   }
 
-  public getFlattenedAnalysesResults(): AnalysisResults[] {
-    return Object.entries(this.analysesResults).map(([nwo, results]) => ({ nwo, results }));
+  public getAnalysesResults(): AnalysisResults[] {
+    return [...this.analysesResults];
   }
 
   private async downloadSingleAnalysisResults(
-    downloadLink: DownloadLink,
+    analysis: AnalysisSummary,
     credentials: Credentials
-  ): Promise<QueryResult[]> {
-    const artifactPath = await downloadArtifactFromLink(credentials, downloadLink);
+  ): Promise<void> {
+    const artifactPath = await downloadArtifactFromLink(credentials, analysis.downloadLink);
+
+    let analysisResults: AnalysisResults;
 
     if (path.extname(artifactPath) === '.sarif') {
-      return await this.readResults(artifactPath);
+      const queryResults = await this.readResults(artifactPath);
+      analysisResults = { nwo: analysis.nwo, results: queryResults };
     } else {
-      // Non-problem or problem-path queries are not currently fully supported.
-      return [];
+      void this.logger.log('Non-problem or problem-path queries are not currently fully supported.');
+      analysisResults = { nwo: analysis.nwo, results: [] };
     }
+
+    this.analysesResults.push(analysisResults);
   }
 
   private async readResults(filePath: string): Promise<QueryResult[]> {
--- a/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
@@ -1,0 +1,90 @@
+import { ExtensionContext } from 'vscode';
+import { Credentials } from '../authentication';
+import { Logger } from '../logging';
+import { downloadArtifactFromLink } from './gh-actions-api-client';
+import { DownloadLink } from './download-link';
+import * as path from 'path';
+import * as fs from 'fs-extra';
+import { AnalysisSummary } from './shared/remote-query-result';
+import * as sarif from 'sarif';
+import { AnalysisResults, QueryResult } from './shared/analysis-result';
+
+export class AnalysesResultsManager {
+  // Store for the results of various analyses for a single remote query.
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -0,0 +1,90 @@
...
+import { AnalysisSummary } from './shared/remote-query-result';
+import * as sarif from 'sarif';
+import { AnalysisResults, QueryResult } from './shared/analysis-result';
+
+export class AnalysesResultsManager {
+  // Store for the results of various analyses for a single remote query.
+  private readonly analysesResults: { [key: string]: QueryResult[] };
+
+  constructor(
+    private readonly ctx: ExtensionContext,
```
aeisenberg, line 17: Rather than storing the `ctx` as a field in this class, does it make more sense to create a credentials instance in the constructor and use that?

I'm not sure if this is a good idea since credentials instances may expire, so take this as a question only.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -0,0 +1,90 @@
...
+  public async downloadAllResults(
+    analysisSummaries: AnalysisSummary[],
+  ): Promise<void> {
+    const credentials = await Credentials.initialize(this.ctx);
+
+    void this.logger.log('Downloading and processing all results');
+
+    for (const analysis of analysisSummaries) {
+      const queryResults = await this.downloadSingleAnalysisResults(analysis.downloadLink, credentials);
+      this.analysesResults[analysis.nwo] = queryResults;
```
aeisenberg, line 46: This downloads each analysis serially, which could take some time if there are lots of them. Should we be parallelizing the requests? We probably don't want to parallelize all requests, but possibly parallelize slices of requests.

This is not something to work on now, only a consideration for the future.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -0,0 +1,90 @@
...
+      // Non-problem or problem-path queries are not currently fully supported.
+      return [];
+    }
+  }
+
+  private async readResults(filePath: string): Promise<QueryResult[]> {
+    const queryResults: QueryResult[] = [];
+
+    const sarifContents = await fs.readFile(filePath, 'utf8');
+    const sarifLog = JSON.parse(sarifContents) as sarif.Log;
```
aeisenberg, line 72: I'm a little concerned here about large sarif files blowing up the process's memory. We do have a streaming sarif parser that avoids polluting memory with bits that aren't used. That parser was developed for local queries, so it may need some tweaking for this case.

See sarif-parser.ts.

And again, not something to worry about right now, but something we will probably need to implement in the future.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -0,0 +1,90 @@
...
+    private readonly ctx: ExtensionContext,
+    private readonly logger: Logger,
+  ) {
+    this.analysesResults = {};
+  }
+
+  public async downloadAnalysisResults(
+    analysisSummary: AnalysisSummary,
+  ): Promise<void> {
+    const credentials = await Credentials.initialize(this.ctx);
```
aeisenberg, line 26: Should we avoid downloading analyses that already have been downloaded? Maybe we would want a flag to either force re-downloading or avoid it.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -0,0 +1,90 @@
...
+    void this.logger.log('Downloading and processing all results');
+
+    for (const analysis of analysisSummaries) {
+      const queryResults = await this.downloadSingleAnalysisResults(analysis.downloadLink, credentials);
+      this.analysesResults[analysis.nwo] = queryResults;
+    }
+  }
+
+  public getFlattenedAnalysesResults(): AnalysisResults[] {
+    return Object.entries(this.analysesResults).map(([nwo, results]) => ({ nwo, results }));
```
aeisenberg, line 51: What's the reason for storing `analysesResults` as an object if the only time it is used is as an array?

Storing as an object would be useful to help prevent duplicate downloads, but that is not implemented yet.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    void this.logger.log('Downloading and processing all results');
+
+    for (const analysis of analysisSummaries) {
+      const queryResults = await this.downloadSingleAnalysisResults(analysis.downloadLink, credentials);
+      this.analysesResults[analysis.nwo] = queryResults;
+    }
+  }
+
+  public getFlattenedAnalysesResults(): AnalysisResults[] {
```
aeisenberg, line 50: nit: Since the only way to interact with `AnalysisResults` is as an array, perhaps you can rename this to:

```suggestion
  public getAnalysesResults(): AnalysisResults[] {
```

aeisenberg (commented):
Looks good. I haven't tried running this yet. I have a handful of suggestions and a few things we will want to think about in the future.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
charisk: Yeah I think storing the `ctx` is safer because the creds might expire.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
charisk: Ah sorry I was planning to leave a comment here but I forgot! I was thinking that some parallelisation would be a good idea too but was unsure about how many concurrent downloads we should be doing. I think the number needs to take into account that people may be running a query against a lot of repos, some with potentially big result files and some with potentially small ones. I was thinking we could start with 3 and see how it goes. What do you think?

Agree we could leave it as is for now - I'll be working on auto-downloads etc. too so might be worth addressing it then.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
charisk: That's a great point! I'll add a TODO to make sure this is covered as part of the overall download behaviour.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
charisk: I thought that this part of the app could be flexible and allow re-downloading (for whatever reason e.g. some sort of partial failure) but the UI will basically not let the user download the same files again and again (the download buttons will be hidden).

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
charisk: Storing as an object removes complexity around removing and items from an array and adding new ones in the case of a re-download. If we want to just drop that behaviour altogether then we could simply have an array of `AnalysisResults` and only add to it.

Looking at everything with fresh eyes I'm thinking that perhaps I should simplify everything for now and add any complexity later if needed.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
charisk: Simplified code - now we won't attempt to download something that has been downloaded already

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
charisk: Updated

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
charisk: Renamed. I think ideally we should be returning a deep copy of the array to make sure that the manager is the only entity manipulating the data but I haven't seen other places in the code doing that so I've not introduced anything here. Perhaps something to discuss though.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
aeisenberg: That all sounds reasonable. 3 is a good place to start.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -42,27 +42,31 @@ export class AnalysesResultsManager {
...
+    let analysisResults: AnalysisResults;
 
     if (path.extname(artifactPath) === '.sarif') {
-      return await this.readResults(artifactPath);
+      const queryResults = await this.readResults(artifactPath);
+      analysisResults = { nwo: analysis.nwo, results: queryResults };
     } else {
-      // Non-problem or problem-path queries are not currently fully supported.
-      return [];
+      void this.logger.log('Non-problem or problem-path queries are not currently fully supported.');
```
aeisenberg, line 65: Can we simplify this message a bit?

```suggestion
      void this.logger.log('Cannot download results. Only alert and path queries are fully supported.');
```

aeisenberg (commented):
Looks good. A small suggestion about updating an error message. I'm not entirely happy with what I have, but we should be using the terms `alert` and `path` queries instead of problem or problem-path.

See https://codeql.github.com/docs/writing-codeql-queries/about-codeql-queries/#overview

After this change, I think it will be good.

aeisenberg (approved):

PR #1108, opened 2022-01-28
Show remote analyses results status

charisk:
Reports status of analyses results:
* Show spinner when downloading results
* Show notification if there was a failure
* Remove download button if results have downloaded successfully
* If all results have been downloaded successfully, hide the "Download all" button

This PR introduces the use of [Primer React](https://primer.style/react) to make use of the Spinner component. See the [Getting started guide](https://primer.style/react/getting-started) to understand some of the changes I've made.

Note that with styled-components being brought in (as a dependency of Primer React) we can now make use of that in our components. In order to keep the size/complexity of the PR reasonable, I didn't take advantage of that yet though.

I will do some more testing around the CSP changes I've made on Monday but I wanted to raise the PR for early feedback.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/interface-utils.ts
+++ b/extensions/ql-vscode/src/interface-utils.ts
@@ -119,6 +119,7 @@ export function getHtmlForWebview(
   webview: Webview,
   scriptUriOnDisk: Uri,
   stylesheetUrisOnDisk: Uri[],
+  allowInlineStyles: boolean
 ): string {
   // Convert the on-disk URIs into webview URIs.
   const scriptWebviewUri = webview.asWebviewUri(scriptUriOnDisk);
@@ -128,8 +129,13 @@ export function getHtmlForWebview(
   // Use a nonce in the content security policy to uniquely identify the above resources.
   const nonce = getNonce();
 
-  const stylesheetsHtmlLines = stylesheetWebviewUris.map(stylesheetWebviewUri =>
-    `<link nonce="${nonce}" rel="stylesheet" href="${stylesheetWebviewUri}">`);
+  const stylesheetsHtmlLines = allowInlineStyles
+    ? stylesheetWebviewUris.map(uri => createStylesLinkWithoutNonce(uri))
+    : stylesheetWebviewUris.map(uri => createStylesLinkWithNonce(nonce, uri));
+
+  const styleSrc = allowInlineStyles
+    ? 'https://*.vscode-webview.net/ vscode-file: \'unsafe-inline\''
+    : `'nonce-${nonce}'`;
 
   /*
    * Content security policy:
@@ -143,7 +149,7 @@ export function getHtmlForWebview(
 <html>
   <head>
     <meta http-equiv="Content-Security-Policy"
-          content="default-src 'none'; script-src 'nonce-${nonce}'; style-src 'nonce-${nonce}'; connect-src ${webview.cspSource};">
+          content="default-src 'none'; script-src 'nonce-${nonce}'; style-src ${styleSrc}; connect-src ${webview.cspSource};">
         ${stylesheetsHtmlLines.join(`    ${os.EOL}`)}
   </head>
   <body>
@@ -242,4 +248,12 @@ export async function jumpToLocation(
       }
     }
   }
+}
+
+function createStylesLinkWithNonce(nonce: string, uri: Uri): string {
+  return `<link nonce="${nonce}" rel="stylesheet" href="${uri}">`;
+}
+
+function createStylesLinkWithoutNonce(uri: Uri): string {
+  return `<link rel="stylesheet" href="${uri}">`;
 }
--- a/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
@@ -8,7 +8,7 @@ import { AnalysisSummary } from './shared/remote-query-result';
 import * as sarif from 'sarif';
 import { AnalysisResults, QueryResult } from './shared/analysis-result';
 import { UserCancellationException } from '../commandRunner';
-import { showAndLogErrorMessage } from '../helpers';
+import * as os from 'os';
 
 export class AnalysesResultsManager {
   // Store for the results of various analyses for a single remote query.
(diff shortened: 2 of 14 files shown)


extensions/ql-vscode/src/interface-utils.ts
```diff
@@ -128,8 +129,13 @@ export function getHtmlForWebview(
...
   const nonce = getNonce();
 
-  const stylesheetsHtmlLines = stylesheetWebviewUris.map(stylesheetWebviewUri =>
-    `<link nonce="${nonce}" rel="stylesheet" href="${stylesheetWebviewUri}">`);
+  const stylesheetsHtmlLines = allowInlineStyles
+    ? stylesheetWebviewUris.map(uri => createStylesLinkWithoutNonce(uri))
+    : stylesheetWebviewUris.map(uri => createStylesLinkWithNonce(nonce, uri));
+
+  const styleSrc = allowInlineStyles
+    ? 'https://*.vscode-webview.net/ vscode-file: \'unsafe-inline\''
```
aeisenberg, line 137: Does this mean that we will be downloading css from `vscode-webview.net`? If so, then we need to make sure that local queries still work even if there's no internet connection. I think that's fine since only the remote queries viewer is using the inline styles.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -72,21 +70,39 @@ export class AnalysesResultsManager {
...
-    let analysisResults: AnalysisResults;
+    let artifactPath;
+    try {
+      artifactPath = await downloadArtifactFromLink(credentials, analysis.downloadLink);
+    }
+    catch (e) {
+      void showAndLogErrorMessage(
+        `There was an issue when downloading the analysis resutls for ${analysis.nwo}: ${e.message}`,
+        { fullMessage: e.stack });
+      return;
```
aeisenberg, line 93: As an alternative to popping up a message box, logging and returning maybe it would be better to log and only pop up a single message box after all promises have completed that lists all analyses that could not be downloaded. I think this would be a better experience, especially if there are lots of analyses that failed.

I can see three ways of doing this:

1. this function can use the results pattern you prototyped in the other branch and the calling function can handle failures appropriately
2. this function can throw an error on failure. The calling function can use `Promise.allSettled` to handle to deal with successes and failures
3. Pass in an error collector callback to this function that keeps track of errors as they happen. After all requests are finished, the calling function can inspect the error collector to handle any errors that were found.

Any of these would be sufficient.

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -80,34 +86,83 @@ const QueryInfo = (queryResult: RemoteQueryResult) => (
...
-);
+const SummaryTitleWithResults = ({
+  queryResult,
+  analysesResults
+}: {
+  queryResult: RemoteQueryResult,
+  analysesResults: AnalysisResults[]
+}) => {
+  const showDownloadButton = queryResult.totalResultCount !== sumAnalysesResults(analysesResults);
+  console.log(`showDownloadButton: ${showDownloadButton}`);
```
aeisenberg, line 97: Accidental commit?

aeisenberg (commented):
Nice.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
charisk: [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) will reject on the first error and since we run the batches serially, we'd only see 1 notification.

However, you're right that this approach isn't ideal as it can possibly hide some errors, so I've restructured a bit - used `Promise.allSettled` and aggregated errors.

extensions/ql-vscode/src/interface-utils.ts
charisk: I've tested running normal queries without internet connection and the view looks + styles are loading.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -85,10 +91,7 @@ export class AnalysesResultsManager {
       artifactPath = await downloadArtifactFromLink(credentials, analysis.downloadLink);
     }
     catch (e) {
-      void showAndLogErrorMessage(
-        `There was an issue when downloading the analysis resutls for ${analysis.nwo}: ${e.message}`,
-        { fullMessage: e.stack });
-      return;
+      throw new Error(`Could not download the analysis resutls for ${analysis.nwo}: ${e.message}`);
```
aeisenberg, line 94:
```suggestion
      throw new Error(`Could not download the analysis results for ${analysis.nwo}: ${e.message}`);
```

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -58,7 +59,12 @@ export class AnalysesResultsManager {
       const nwos = batch.map(a => a.nwo).join(', ');
       void this.logger.log(`Downloading batch ${Math.floor(i / batchSize) + 1} of ${numOfBatches} (${nwos})`);
 
-      await Promise.all(batchTasks);
+      const taskResults = await Promise.allSettled(batchTasks);
+      const failedTasks = taskResults.filter(x => x.status === 'rejected') as Array<PromiseRejectedResult>;
+      if (failedTasks.length > 0) {
+        const failures = failedTasks.map(t => t.reason.message).concat(os.EOL);
+        throw Error(failures.join(os.EOL));
+      }
```
aeisenberg, line 67: So, this leaves us in a state where batches are downloaded until there is a batch with at least one error and then all further downloading stops. I think it should be possible to continue attempting to download even after a single error occurs.

If you accumulate failures in a list within the loop (maybe you can log to the output view so users can see errors as they happen), and then you can throw after _all_ the requests are settled. This way, users can see errors as they occur (if they want), they get a summary of all errors at the end, and a single failure doesn't prevent further downloads from happening.

aeisenberg (commented):
This is looking really good. I have another suggestion, though.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
charisk: That's a good point, sorry I missed it yesterday. Updated, hopefully there now 🤞

aeisenberg (approved):
Thanks for handling this. Looks good now.

PR #1111, opened 2022-02-01
More work on the graph viewer

aeisenberg:
Built on #705.

The viewer is largely implemented now with the following features and
limitations:

1. Any query with `@kind graph` will be opened as a graph
2. Queries that are `@kind graph` and 
   `@tags ide-contextual-queries/print-cfg` will be used in the
   `CodeQL: View CFG` context command. This will be visible
   similar to how the AST viewer works. If there is not exactly
   1 such query for a given language, then the extension will throw
   an error.
3. Unfortunately, the cg viewer assumes that the entire file will
   be added to the graph, so often this will be too big, That leads to
   the following limitation:
4. There is no size checking on the graph. Graphs that are too big will
   crash vscode.
5. Lastly, there is a small bug in how the `@id` is interpreted. Any
   `@id` with a `/` in it will place the `.dot` in a location that
   can't be found by vscode. So, just don't name your queries with any
   `/`.

This feature is only available in canary mode.


aeisenberg:
@hvitved, would you be able to take a look at this? I think it is mostly what you want (except for the limitations mentioned above).

hvitved:
I have tested the extension with [this](https://github.com/github/codeql/blob/main/ruby/ql/test/library-tests/controlflow/graph/Cfg.ql) query, on a Ruby test DB, and that works just fine, except it keeps printing "Rendering graph" even after it has been rendered:

image

I also tested with [this](https://github.com/github/codeql/pull/7820) Print CFG query, running the "View CFG" command, but that results in `Graph is not available.` (an empty graph, I believe).

Thanks again for picking up this task.

aeisenberg:
Yes...the rendering graph message is something I need to fix.

And see my point (5) above about why the print ast query is not working.

hvitved:
> And see my point (5) above about why the print ast query is not working. The most correct fix would be from inside of the CLI, which I don't have capacity to do right now. Another work around could be done in vscode itself. I'll see if I can get to that. First priority is getting this feature released internally.

I agree that this should not be a blocker. I tried changing the ID to `rb-print-cfg`, but that still produces an empty graph.

aeisenberg:
OK... finished and ready for a review. This has been fixed:

> 5\. Lastly, there is a small bug in how the `@id` is interpreted. Any
>         `@id` with a `/` in it will place the `.dot` in a location that
>         can't be found by vscode. So, just don't name your queries with any
>         `/`.

Also, there is a warning on the results view whenever you view a graph. And I have ensured that this feature is disabled if the canary flag is off.

hvitved (approved):

aeisenberg:
Here is a sample query that I have been using for the graph:

```ql
/**
 * @name Print AST
 * @description Outputs a representation of a file's Abstract Syntax Tree. This
 *              query is used by the VS Code extension.
 * @id id/cpp-print-ast
 * @kind graph
 * @tags ide-contextual-queries/print-ast
 */

import cpp
import semmle.code.cpp.PrintAST
import definitions

class Cfg extends PrintASTConfiguration {
  override predicate shouldPrintFunction(Function func) {
    func.getFile().getBaseName() = "cpu.c" and func.getName() = "sandbox_read_fdt_from_file"
  }
}
```

It is adapted from the cpp printAST query. Replace the RHS of the `getBaseName` comparison with a cpp file in a source archive and `getName` with a name of a function.

If you want to try out the `View CFG` command, choose _any_ print AST query and add a new tag `ide-contextual-queries/print-cfg`.  Then go to a _small_ source file in a database of the same language, right click and select `View CFG`.

aeisenberg:
@shati-patel, @cklin this is ready for your review now. I fixed a race condition that sometimes prevented a graph from appearing. See the second to last commit.

aeisenberg:
Forgot to mention that if you want to view a graph, you need to regenerate the web view. To do that, run `npm run  build` in the `vscode-codeql/extensions/ql-vscode` directory.

cc: @cklin since I'm not sure if you're aware of this.

extensions/ql-vscode/src/helpers.ts
```diff
@@ -558,3 +558,28 @@ export async function createTimestampFile(storagePath: string) {
...
+ * @param dir the directory to walk
+ *
+ * @return An iterator of the full path to all files recursively found in the directory.
+ */
+ export async function* walkDirectory(dir: string): AsyncIterableIterator<string> {
+  const seenFiles = new Set<string>();
+  for await (const d of await fs.opendir(dir)) {
+    const entry = path.join(dir, d.name);
+    if (seenFiles.has(entry)) {
+      continue;
```
cklin, line 576: I am not sure how execution would ever reach this line.  Can `fs.opendir` return duplicate directory entries for the same name?

cklin (approved):
The commits I reviewed are in line with our discussion this morning.

extensions/ql-vscode/src/helpers.ts
aeisenberg: Ah...good point. This is a leftover from when I had originally thought it would be handling symlinks. With symlinks, you need to be careful about cycles, but not the case for regular files. I will remove.

shati-patel (approved):
Thanks for the PR, and for your helpful walkthrough 🎉 Looks good, and I've had a brief poke around locally too, which works as expected 🙌🏽

PR #1114, opened 2022-02-02
Update CODEOWNERS

aeisenberg:
code-scanning security experiences team should be helping to review the remote queries part of the extension.


shati-patel (approved):
Makes sense! I didn't even know that team existed 😂

PR #1115, opened 2022-02-03
Simplify the query history objects

aeisenberg:
The goal with this change is to simplify the query history to make it
possible to serialize and de serialize.

I am not a big fan of this approach right now since it relies on mucking
with prototypes when de-serializing.

## Checklist


aeisenberg:
@charisk would you be able to chat about this PR? I'm trying to work on a way to properly (de-)serialize a complex object. I have something that mostly works, but it is messy. Do you have any better ideas?

Can you look at these lines in particular? https://github.com/github/vscode-codeql/pull/1115/files#diff-63143db7fbfcfc5c1d2fc82847fb9e04bf8e51580ae0f5c240d8bf97631e900cR180-R222

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
```diff
@@ -237,19 +248,85 @@ describe('query-results', () => {
...
+        expect(allHistoryActual[0]).to.deep.eq(allHistory[0]);
+        const info = allHistoryActual[i];
+        if (info.completedQuery) {
+          const dbItem = info.completedQuery.query.dbItem;
+          expect(dbItem).to.deep.eq(allHistory[i].completedQuery!.query.dbItem);
+        }
+      }
+      expect(allHistoryActual.length).to.deep.eq(allHistory.length);
+    });
+  });
```
aeisenberg, line 314: This test will ensure that whatever we serialize can be properly deserialized, but as you can see, it's complex.

charisk:
> @charisk would you be able to chat about this PR? I'm trying to work on a way to properly (de-)serialize a complex object. I have something that mostly works, but it is messy. Do you have any better ideas?
> 
> Can you look at these lines in particular? https://github.com/github/vscode-codeql/pull/1115/files#diff-63143db7fbfcfc5c1d2fc82847fb9e04bf8e51580ae0f5c240d8bf97631e900cR180-R222

If I'm reading the code right, I think we could use something like https://github.com/typestack/class-transformer to create the class instances for us. Happy to talk/pair when you're online.

charisk:
> Thanks. Interesting library, though since we're working with nested bespoke objects, it looks like the only way to handle this is using decorators. The bigger complication is that we can't serialize some of the objects (eg- dbItem and config) since they rely on their connections to the global state.
> 
> Maybe I need to read more, but I'm not sure if this library is going to make things much simpler.

I'm not familiar with that part of the code (happy to have a more proper look tomorrow!) so I may be completely off here but it feels like those models are more complex than they should be. I'm not sure what you mean with "connection to the global state" though. Do we need to split data and behaviour to different places?

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -706,7 +738,7 @@ export async function createInitialQueryInfo(
     isQuickEval,
     isQuickQuery: isQuickQueryPath(queryPath),
     databaseInfo,
-    id: queryId++,
+    id: `${path.basename(queryPath)}-${nanoid()}`,
```
charisk, line 741: Does queryPath always exist? Can we run a query without saving it?

extensions/ql-vscode/src/vscode-tests/no-workspace/run-queries.test.ts
```diff
@@ -47,9 +46,13 @@ describe('run-queries', () => {
       const qs = createMockQueryServerClient();
       const mockProgress = 'progress-monitor';
       const mockCancel = 'cancel-token';
+      const mockQlProgram = {
+        mock: 'program'
+      } as unknown as QlProgram;
```
charisk, line 51: I'd personally make a more proper mock to help with debugging.

```ts
const mockQlProgram = {
  dbschemePath: '',
   libraryPath: [],
   queryPath: ''
} as QlProgram;
```

Feel free to ignore though!

aeisenberg: That's fine. I was just being lazy. :)

extensions/ql-vscode/src/query-results.ts
```diff
@@ -174,15 +183,63 @@ export type FullCompletedQueryInfo = FullQueryInfo & {
...
+          // slurped queries do not need to be disposed
+          q.completedQuery.dispose = () => { /**/ };
+        }
+        return q;
+      });
+    } catch (e) {
+      void showAndLogErrorMessage('Error loading query history.', {
+        fullMessage: ['Error loading query history.', e.stack].join('\n'),
+      });
+      return [];
```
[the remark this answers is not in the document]
charisk: I think in general we should leave the caller to decide what to do with the result, but happy to leave it for now as we seem to do this a lot in the current code anyway.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -367,7 +390,7 @@ async function checkDbschemeCompatibility(
 
     // query.program.dbschemePath is the dbscheme of the actual
     // database we're querying.
-    const dbschemeOfDb = await hash(query.program.dbschemePath);
+    const dbschemeOfDb = await hash(dbItem.contents.dbSchemeUri.fsPath);
```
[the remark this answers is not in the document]
charisk: Thanks for the explanation!

PR #1130, opened 2022-02-09
Save query history across restarts

aeisenberg:
Successfully completed queries will be stored on disk and available
across restarts.

- The query results are contained in global storage.
- Metadata and a summary about a query are stored in workspace storage.
- There is a job that runs every 2 hours to determine if any queries are
  old enough to be deleted.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/contextual/templateProvider.ts
+++ b/extensions/ql-vscode/src/contextual/templateProvider.ts
@@ -42,7 +42,7 @@ export class TemplateQueryDefinitionProvider implements DefinitionProvider {
     private cli: CodeQLCliServer,
     private qs: QueryServerClient,
     private dbm: DatabaseManager,
-    private queryStorageLocation: string,
+    private queryStorageDir: string,
   ) {
     this.cache = new CachedOperation<LocationLink[]>(this.getDefinitions.bind(this));
   }
@@ -70,7 +70,7 @@ export class TemplateQueryDefinitionProvider implements DefinitionProvider {
         this.dbm,
         uriString,
         KeyType.DefinitionQuery,
-        this.queryStorageLocation,
+        this.queryStorageDir,
         progress,
         token,
         (src, _dest) => src === uriString
@@ -86,7 +86,7 @@ export class TemplateQueryReferenceProvider implements ReferenceProvider {
     private cli: CodeQLCliServer,
     private qs: QueryServerClient,
     private dbm: DatabaseManager,
-    private queryStorageLocation: string,
+    private queryStorageDir: string,
   ) {
     this.cache = new CachedOperation<FullLocationLink[]>(this.getReferences.bind(this));
   }
@@ -119,7 +119,7 @@ export class TemplateQueryReferenceProvider implements ReferenceProvider {
         this.dbm,
         uriString,
         KeyType.DefinitionQuery,
-        this.queryStorageLocation,
+        this.queryStorageDir,
         progress,
         token,
         (src, _dest) => src === uriString
@@ -140,7 +140,7 @@ export class TemplatePrintAstProvider {
     private cli: CodeQLCliServer,
     private qs: QueryServerClient,
     private dbm: DatabaseManager,
-    private queryStorageLocation: string,
+    private queryStorageDir: string,
   ) {
     this.cache = new CachedOperation<QueryWithDb>(this.getAst.bind(this));
   }
@@ -221,7 +221,7 @@ export class TemplatePrintAstProvider {
         this.qs,
         db,
         initialInfo,
-        this.queryStorageLocation,
+        this.queryStorageDir,
         progress,
         token,
         templates
--- a/extensions/ql-vscode/src/contextual/templateProvider.ts
+++ b/extensions/ql-vscode/src/contextual/templateProvider.ts
@@ -42,6 +42,7 @@ export class TemplateQueryDefinitionProvider implements DefinitionProvider {
     private cli: CodeQLCliServer,
     private qs: QueryServerClient,
     private dbm: DatabaseManager,
+    private queryStorageLocation: string,
   ) {
     this.cache = new CachedOperation<LocationLink[]>(this.getDefinitions.bind(this));
   }
@@ -69,6 +70,7 @@ export class TemplateQueryDefinitionProvider implements DefinitionProvider {
         this.dbm,
         uriString,
         KeyType.DefinitionQuery,
+        this.queryStorageLocation,
         progress,
         token,
         (src, _dest) => src === uriString
@@ -84,6 +86,7 @@ export class TemplateQueryReferenceProvider implements ReferenceProvider {
     private cli: CodeQLCliServer,
     private qs: QueryServerClient,
     private dbm: DatabaseManager,
+    private queryStorageLocation: string,
   ) {
     this.cache = new CachedOperation<FullLocationLink[]>(this.getReferences.bind(this));
   }
@@ -116,6 +119,7 @@ export class TemplateQueryReferenceProvider implements ReferenceProvider {
         this.dbm,
         uriString,
         KeyType.DefinitionQuery,
+        this.queryStorageLocation,
         progress,
         token,
         (src, _dest) => src === uriString
@@ -136,6 +140,7 @@ export class TemplatePrintAstProvider {
     private cli: CodeQLCliServer,
     private qs: QueryServerClient,
     private dbm: DatabaseManager,
+    private queryStorageLocation: string,
   ) {
     this.cache = new CachedOperation<QueryWithDb>(this.getAst.bind(this));
   }
@@ -216,6 +221,7 @@ export class TemplatePrintAstProvider {
         this.qs,
         db,
         initialInfo,
+        this.queryStorageLocation,
         progress,
         token,
         templates
--- a/extensions/ql-vscode/src/query-history-scrubber.ts
+++ b/extensions/ql-vscode/src/query-history-scrubber.ts
@@ -1,0 +1,135 @@
+import * as fs from 'fs-extra';
+import * as os from 'os';
+import * as path from 'path';
+import { Disposable, ExtensionContext } from 'vscode';
+import { logger } from './logging';
+
+const LAST_SCRUB_TIME_KEY = 'lastScrubTime';
+
+type Counter = {
+  increment: () => void;
+};
+
+/**
+ * Registers an interval timer that will periodically check for queries old enought
+ * to be deleted.
+ *
+ * Note that this scrubber will clean all queries from all workspaces. It should not
+ * run too often and it should only run from one workspace at a time.
+ *
+ * Generally, `wakeInterval` should be significantly shorter than `throttleTime`.
+ *
+ * @param wakeInterval How often to check to see if the job should run.
+ * @param throttleTime How often to actually run the job.
+ * @param maxQueryTime The maximum age of a query before is ready for deletion.
+ * @param queryDirectory The directory containing all queries.
+ * @param ctx The extension context.
+ */
+export function registerQueryHistoryScubber(
+  wakeInterval: number,
+  throttleTime: number,
+  maxQueryTime: number,
+  queryDirectory: string,
+  ctx: ExtensionContext,
+
+  // optional counter to keep track of how many times the scrubber has run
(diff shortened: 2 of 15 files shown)


extensions/ql-vscode/src/query-history.ts
```diff
@@ -884,3 +951,106 @@ the file in the file explorer and dragging it into the workspace.`
...
+  queryDirectory: string,
+  ctx: ExtensionContext,
+
+  // optional counter to keep track of how many times the scrubber has run
+  counter?: {
+    increment: () => void;
+  }
+): Disposable {
+  const deregister = setInterval(async () => {
+    const lastScrubTime = ctx.globalState.get<number>(LAST_SCRUB_TIME_KEY);
```
charisk, line 985: Do we also want to store the fact that we're currently scrubbing, in global state, and check that before kicking off the process? That would help avoid multiple workspaces running the scrubber at the same time. We'd need to be careful though about reseting that though.

Or are we happy that if two scrubbing processes did happen to run at the same time, we have enough defensiveness that it wouldn't cause trouble?

extensions/ql-vscode/src/query-results.ts
```diff
@@ -253,13 +270,15 @@ export class FullQueryInfo {
...
-    private readonly source?: CancellationTokenSource
+    private source?: CancellationTokenSource // used to cancel in progress queries
   ) {
     this.setConfig(config);
   }
 
   cancel() {
     this.source?.cancel();
+    // query is no longer in progress, can delete the cancellation token source
+    delete this.source;
```
charisk, line 281: Why do we need to do this?

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -651,19 +664,21 @@ export async function compileAndRunQueryAgainstDatabase(
...
     hasMetadataFile,
     packConfig.dbscheme,
     initialInfo.quickEvalPosition,
     metadata,
     templates
   );
+  await query.createTimestampFile();
 
-  const upgradeDir = await tmp.dir({ dir: upgradesTmpDir.name, unsafeCleanup: true });
+  let upgradeDir: tmp.DirectoryResult | undefined;
```
charisk, line 677: Is this change relevant to query history? 🤔

extensions/ql-vscode/src/query-history.ts
```diff
@@ -488,19 +540,21 @@ export class QueryHistoryManager extends DisposableObject {
     singleItem: FullQueryInfo,
     multiSelect: FullQueryInfo[]
   ): Promise<void> {
-    if (!this.assertSingleQuery(multiSelect)) {
+    const { finalSingleItem, finalMultiSelect } = this.determineSelection(singleItem, multiSelect);
+
+    if (!this.assertSingleQuery(finalMultiSelect)) {
```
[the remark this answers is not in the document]
charisk: Thanks for pointing it out, and for making the fix! 

In the future, it'd personally help me a lot of we had separate PRs for such changes. I can live with it if it's too much trouble, I just find that I miss stuff more in bigger PRs that do multiple things.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -884,3 +951,106 @@ the file in the file explorer and dragging it into the workspace.`
...
+                void logger.log(`  ${dir} is older than ${maxQueryTime / 1000} seconds. Deleting.`);
+                await fs.remove(dir);
+                scrubCount++;
+              } else {
+                void logger.log(`  ${dir} is not older than ${maxQueryTime / 1000} seconds. Keeping.`);
+              }
+            }
+          } catch (err) {
+            errors.push(`  Could not delete '${dir}': ${err}`);
+          }
```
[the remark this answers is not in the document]
charisk: That seems like the right place to me!

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -98,6 +94,16 @@ export class QueryEvaluationInfo {
     return path.join(this.querySaveDir, `sortedResults-${resultSetName}.bqrs`);
   }
 
+  /**
+   * Creates a file in the query directory that indicates when this query was created.
+   * This is important for keeping track of when queries should be removed.
+   */
+  async createTimestampFile() {
```
charisk, line 101: Shouldn't this be part of the information we store for the query in the all-history.json?

Happy to be told that that would be too big of a change right now.

extensions/ql-vscode/src/query-history.ts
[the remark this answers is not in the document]
aeisenberg: In general, I agree and prefer that for reviewing. But for larger changes like these, where nothing is working (until everything is), I find it quite hard to break out these smaller bits, especially since these bits are often somewhat dependent on the larger change. Sometimes, I try separating these out into individual commits in the same PR.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -381,6 +400,16 @@ export class QueryHistoryManager extends DisposableObject {
     this.push(
       queryHistoryConfigListener.onDidChangeConfiguration(() => {
         this.treeDataProvider.refresh();
+        // recreate the history scrubber
+        this.queryHistoryScrubber.dispose();
+        this.queryHistoryScrubber = this.push(
+          registerQueryHistoryScubber(
```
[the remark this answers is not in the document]
aeisenberg: Actually, yes. I should. There are two instances of calling this and they should be identical (currently, they are not).

extensions/ql-vscode/src/query-history.ts
```diff
@@ -398,6 +427,27 @@ export class QueryHistoryManager extends DisposableObject {
...
+    );
+  }
+
+  async readQueryHistory(): Promise<void> {
+    const history = await FullQueryInfo.slurp(this.queryMetadataStorageLocation, this.queryHistoryConfigListener);
+    this.treeDataProvider.allHistory = history;
+  }
+
+  async writeQueryHistory(): Promise<void> {
+    const toSave = this.treeDataProvider.allHistory.filter(q => q.isCompleted());
```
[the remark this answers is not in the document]
aeisenberg: No. Not cancelled queries. It does include queries that are unsuccessful. This allows users to view the logs and find out why the query failed.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -445,14 +495,16 @@ export class QueryHistoryManager extends DisposableObject {
...
-
-    (finalMultiSelect || [finalSingleItem]).forEach((item) => {
-      // Removing in progress queries is not supported yet
+    const toDelete = (finalMultiSelect || [finalSingleItem]);
+    await Promise.all(toDelete.map(async (item) => {
+      // Removing in progress queries is not supported. They must be cancelled first.
       if (item.status !== QueryStatus.InProgress) {
         this.treeDataProvider.remove(item);
         item.completedQuery?.dispose();
+        await item.completedQuery?.query.cleanUp();
```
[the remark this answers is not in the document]
aeisenberg: Two reasons, we can't do this:

1. Dispose is an inherited method and cannot be made async
2. `cleanUp` will delete the entire directory of the query. We don't want to do this on a normal dispose, which gets called when the application exits. We only want to do this when the user explicitly wants to remove the query.

I can add a comment here. And I think `cleanUp` is not the right name. I think `deleteQuery` is better.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -884,3 +951,106 @@ the file in the file explorer and dragging it into the workspace.`
...
+ *
+ * Generally, `wakeInterval` should be significantly shorter than `throttleTime`.
+ *
+ * @param wakeInterval How often to check to see if the job should run.
+ * @param throttleTime How often to actually run the job.
+ * @param maxQueryTime The maximum age of a query before is ready for deletion.
+ * @param queryDirectory The directory containing all queries.
+ * @param ctx The extension context.
+ */
+export function registerQueryHistoryScubber(
```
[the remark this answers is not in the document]
aeisenberg: Done.

extensions/ql-vscode/src/query-history.ts
aeisenberg: I think a safe and easy thing to do is to move the `ctx.globalState.update` to happen as soon as possible after the `get`. This minimizes the window when two workspaces would be able to run the scrubber at the same time. Fundamentally though, I don't think this would be a problem. There might be some extra error messages in one of the workspaces if it gets a list of query directories, but by the time it checks for them they are gone.

extensions/ql-vscode/src/query-results.ts
aeisenberg: The `source` object is complex and contains workspace state. I'm doing it to make sure workspace resources aren't leaked. I realized that this object is a `Disposable` as well, so I'll add a call to `dispose`.

Slightly unrelated, this object can't be serialized to JSON because it's recursive, so it also needs to be deleted before serialization.

extensions/ql-vscode/src/run-queries.ts
aeisenberg: It can't be stored there. We need to store this value in the query directory so that all workspaces know how old a given query folder is.

Remember, that the goal is to be able to clean up queries for workspaces that you never intend to open again.

extensions/ql-vscode/src/contextual/templateProvider.ts
```diff
@@ -42,6 +42,7 @@ export class TemplateQueryDefinitionProvider implements DefinitionProvider {
     private cli: CodeQLCliServer,
     private qs: QueryServerClient,
     private dbm: DatabaseManager,
+    private queryStorageLocation: string,
```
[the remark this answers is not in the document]
charisk: Yep that works

extensions/ql-vscode/src/query-history-scrubber.ts
```diff
@@ -0,0 +1,135 @@
...
+  const deregister = setInterval(scrubber, wakeInterval, throttleTime, maxQueryTime, queryDirectory, ctx, counter);
+
+  return {
+    dispose: () => {
+      clearInterval(deregister);
+    }
+  };
+}
+
+async function scrubber(
```
charisk, line 47: How about calling this `scrub` instead (or `scrubHistory`)? I generally try to use verbs for functions.

extensions/ql-vscode/src/query-history.ts
charisk: Ok cool - we can see how it goes in practise too and adjust if necessary.

extensions/ql-vscode/src/query-results.ts
charisk: That sounds fun 😱

extensions/ql-vscode/src/run-queries.ts
charisk: > It can't be stored there. We need to store this value in the query directory so that all workspaces know how old a given query folder is.
Is all-history.json not meant to be stored in global storage?

extensions/ql-vscode/src/query-history.ts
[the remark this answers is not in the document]
charisk: I think it's still ok to separate things out in separate PRs even when you work on a larger bit of work. 

If I'm working on something fairly big I start off the work and if I find something that could be split into its own PR then I do that. It's a bit more hassle juggling things but I find that 
a) it's easier for the reviewer 
b) things get done more quickly overall and 
c) there's less of a chance that things get missed

I'm probably "preaching to the converted" and apologies for that, just wanted to make it clear what my personal preferences (and weaknesses!) are as a reviewer.

extensions/ql-vscode/src/query-history-scrubber.ts
charisk, line 47: How about calling this scrub instead (or scrubHistory)? I generally try to use verbs for functions.

extensions/ql-vscode/src/run-queries.ts
aeisenberg: That's correct. `all-history.json` is in workspace storage. This way each workspace only sees the queries relevant to itself. Other workspaces' queries will not show.

I can change the name and be more clear.

charisk (approved):

PR #1107, opened 2022-01-28
Run CodeQL analysis with latest CLI to opt into ML-powered queries beta

henrymercer:
We need to run the CodeQL analysis with at least CodeQL CLI v2.7.6 to opt into the ML-powered queries beta. This CLI hasn't yet made it into the Actions VM images, so we add `tools: latest` to download the CodeQL Bundle v2.7.6 from `github/codeql-action`. Once CodeQL Bundle v2.7.6 ships in the Actions VM images, we will then be able to remove `tools: latest`.

## Checklist

- [x] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [x] Issues have been created for any UI or other user-facing changes made by this pull request.
- [x] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #1133, opened 2022-02-10
Use default bundle for Code Scanning analysis

henrymercer:
The Actions VM image containing v2.7.6 of the CodeQL CLI has now fully rolled out, so we no longer need to download the latest CodeQL bundle to use this CLI.

Reverts github/vscode-codeql#1107


aeisenberg:
I wonder if it would be best to continue using the latest CLI (or maybe even the nightly version). If there is an incompatibility, I'd rather find it on this repo first before it hits external customers.

PR #1134, opened 2022-02-10
Remove feature flag for loading ML models from packs

henrymercer:
This functionality is now ready for use by all users, and should be enabled for everyone. I don't think now is the right time to advertise it though, so haven't added a changelog note.

**Removal of `workspace.isTrusted` guard**

The restriction to trusted workspaces only was only necessary to confirm secure usage of the `shouldInsecurelyLoadMlModelsFromPacks` setting (see the comment). This setting has now been removed, and the security mechanisms built into `codeql resolve ml-models` mean that we don't need or want to restrict loading ML models to trusted workspaces.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -10,8 +10,7 @@ import {
   TextDocument,
   TextEditor,
   Uri,
-  window,
-  workspace
+  window
 } from 'vscode';
 import { ErrorCodes, ResponseError } from 'vscode-languageclient';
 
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/run-queries.ts
```diff
@@ -632,13 +631,7 @@ export async function compileAndRunQueryAgainstDatabase(
   const metadata = await tryGetQueryMetadata(cliServer, qlProgram.queryPath);
 
   let availableMlModels: cli.MlModelInfo[] = [];
-  // The `capabilities.untrustedWorkspaces.restrictedConfigurations` entry in package.json doesn't
-  // work with hidden settings, so we manually check that the workspace is trusted before looking at
-  // whether the `shouldInsecurelyLoadMlModelsFromPacks` setting is enabled.
-  if (workspace.isTrusted &&
```
aeisenberg, line 638: Should we still prevent this setting in untrusted workspaces?

henrymercer: I'm not sure I understand your question — this setting is no longer used.

aeisenberg: Sorry...that was unclear. I just meant is it safe to use ml models in an untrusted workspace? The workspace may contain an untrusted model, which as I understand it, could run untrusted code.

I know there are SHAs that ship with the CLI to restrict which models are allowed to run. Does that mean there is no danger of running an untrusted model?

PR #1138, opened 2022-02-11
Fix upgrades path

aeisenberg:
Ensure that upgrades can be resolved even when the upgrades pack is not
in the workspace. This is the situation when the core libraries are
resolved from the package cache.

This change works because `qlProgram.libraryPath` is the resolved
search path for compiling the query. We are guaranteed that the
appropriate core libraries are included in this query.

Note that this change avoids using extra source folders from the
workspace. Previously without using packages, we assume that all
relevant query paths are already inside the workspace. With
packaging, this is no longer the case.

It is theoretically possible that there will be extra upgrade scripts
that are not on the resolved search path, but are included in the
workspace. This situation would have worked in the past.This is not a
situation that we expect to happen in practice. And if this does happen,
I believe this is an error and all upgrades should be added explicitly
to the search path.

An open question is if this will work with downgrade scripts. If it does
not, then I don't think this change makes things any worse than before.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -437,12 +437,23 @@ async function compileNonDestructiveUpgrade(
   progress: ProgressCallback,
   token: CancellationToken,
 ): Promise<string> {
-  const searchPath = getOnDiskWorkspaceFolders();
 
   if (!dbItem?.contents?.dbSchemeUri) {
     throw new Error('Database is invalid, and cannot be upgraded.');
   }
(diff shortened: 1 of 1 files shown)


cklin:
It looks like the test database might need to be rebuilt for the new javascript dbscheme?

aeisenberg:
The failure is when running against an old version of the CLI. It's passing on newer versions. I'm going to see if I can reproduce locally.

aeisenberg:
Ah ha....the problem is that older codeql versions (pre-packaging) did not have the upgrades packs on the library path of the src packs. There was no dependency chain. 

What this means is that for older CLIs, we need to use the old mechanism and put all workspace folders onto the search path.

It _may_ necessary to combine the workspace folders with the resolved path, but I'd rather not go there unless I have to.

cklin (approved):

Johnstone456 (commented):
MPESA

PR #1142, opened 2022-02-14
Store remote query artifacts in global storage

aeisenberg:
This moves all artifacts downloaded for a remote query into the global
storage directory. Each remote query gets its own directory. The
parent directory is the shared query storage directory.

Each remote query directory also gets a timestamp file.

With these changes, remote queries will be persisted across restarts
and deleted automatically on the same schedule as local queries.

Note: This does _not_ add remote queries to the query history view yet.
This part of the feature is coming next.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
@@ -8,6 +8,8 @@ import { AnalysisResults, QueryResult } from './shared/analysis-result';
 import { UserCancellationException } from '../commandRunner';
 import * as os from 'os';
 import { sarifParser } from '../sarif-parser';
+import { createTimestampFile } from '../helpers';
+import { nanoid } from 'nanoid';
 
 export class AnalysesResultsManager {
   // Store for the results of various analyses for a single remote query.
@@ -15,6 +17,7 @@ export class AnalysesResultsManager {
 
   constructor(
     private readonly ctx: ExtensionContext,
+    private readonly storagePath: string,
     private readonly logger: Logger,
   ) {
     this.analysesResults = [];
@@ -78,6 +81,20 @@ export class AnalysesResultsManager {
     return [...this.analysesResults];
   }
 
+  /**
+   * Prepares a directory for storing analysis results for a single query run.
+   * This directory initially contains only a timestamp file, which will be
+   * used by the query history manager to determine when the directory
+   * should be deleted.
+   *
+   * @param queryName The name of the query that was run.
+   */
+  public async prepareDownloadDirectory(queryName: string): Promise<void> {
+    // Prepare the storage directory.
+    const artifactStorageDir = path.join(this.storagePath, `${queryName}-${nanoid()}`);
+    await createTimestampFile(artifactStorageDir);
+  }
+
   private async downloadSingleAnalysisResults(
     analysis: AnalysisSummary,
     credentials: Credentials,
@@ -94,7 +111,7 @@ export class AnalysesResultsManager {
 
     let artifactPath;
     try {
-      artifactPath = await downloadArtifactFromLink(credentials, analysis.downloadLink);
+      artifactPath = await downloadArtifactFromLink(credentials, analysis.downloadLink, this.storagePath);
     }
     catch (e) {
       throw new Error(`Could not download the analysis results for ${analysis.nwo}: ${e.message}`);
@@ -119,7 +136,7 @@ export class AnalysesResultsManager {
 
     // Read the sarif file and extract information that we want to display
     // in the UI. For now we're only getting the message texts but we'll gradually
-    // extract more information based on the UX we want to build. 
+    // extract more information based on the UX we want to build.
 
     sarifLog.runs?.forEach(run => {
       run?.results?.forEach(result => {
--- a/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
@@ -8,8 +8,6 @@ import { AnalysisResults, QueryResult } from './shared/analysis-result';
 import { UserCancellationException } from '../commandRunner';
 import * as os from 'os';
 import { sarifParser } from '../sarif-parser';
-import { createTimestampFile } from '../helpers';
-import { nanoid } from 'nanoid';
 
 export class AnalysesResultsManager {
   // Store for the results of various analyses for a single remote query.
@@ -17,7 +15,7 @@ export class AnalysesResultsManager {
 
   constructor(
     private readonly ctx: ExtensionContext,
-    private readonly storagePath: string,
+    readonly storagePath: string,
     private readonly logger: Logger,
   ) {
     this.analysesResults = [];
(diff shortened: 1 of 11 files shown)


extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -78,6 +81,20 @@ export class AnalysesResultsManager {
...
 
+  /**
+   * Prepares a directory for storing analysis results for a single query run.
+   * This directory initially contains only a timestamp file, which will be
+   * used by the query history manager to determine when the directory
+   * should be deleted.
+   *
+   * @param queryName The name of the query that was run.
+   */
+  public async prepareDownloadDirectory(queryName: string): Promise<void> {
```
charisk, line 92: Do we only want analyses results to be stored or do we also want to store other artifacts and query information in this directory? 

If it's more general, then I think it belongs in the `RemoteQueriesManager` class.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -78,6 +81,20 @@ export class AnalysesResultsManager {
...
+    // Prepare the storage directory.
+    const artifactStorageDir = path.join(this.storagePath, `${queryName}-${nanoid()}`);
```
charisk, line 94: Not sure this is a useful comment but just something to think about.. We seem to be going down an implementation path where various parts of the code control how query history data is stored. For example, in this case, the `AnalysesResultsManager` decides to follow the <queryName>-<id> pattern. But it could decide to follow a different pattern and (I imagine) that would make other bits of the code unhappy. 

I've not thought about query history nearly as much as you have so I could be wrong, but it feels like the responsibility and knowledge around query history should be in a single place.

Perhaps what we're thinking here is overall data of a query, not "historical" data only. Maybe instead of a `storagePath` we could to be giving these objects a class/function that will tell them where to store their data.

charisk (commented):
Thanks for looking into this! I'll have another look once you've updated with latest main as there's a bit of noise.

aeisenberg:
It's hard to manage this chain of PRs. An extra commit snuck its way in there. Should be clearer now.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
aeisenberg: What else will need to be stored? Currently, it's only the analyses results. You probably have a better understanding here than I do.

extensions/ql-vscode/src/remote-queries/download-link.ts
```diff
@@ -1,20 +1,25 @@
...
 
   /**
    * An optional path to follow inside the downloaded archive containing the artifact.
    */
   innerFilePath?: string;
+
+  /**
+   * The full path to the directory where the artifacts for this link is stored.
+   */
+  artifactStorageDir: string;
```
charisk, line 24: This doesn't feel like it belongs here to me. The download link is meant to hold information about where to find a download, not how to store it. 

How about we store it on the `RemoteQuery` entity itself? 

Or, as I mentioned in [this](https://github.com/github/vscode-codeql/pull/1142#discussion_r806610367) comment, perhaps we need something more central that manages where query data is stored rather than spreading the knowledge/responsibility around.

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -79,13 +90,19 @@ export class RemoteQueriesManager {
...
-      await showAndLogErrorMessage(`Remote query execution failed. Error: ${queryResult.error}`);
-      return;
-    } else if (queryResult.status === 'Cancelled') {
+    } else if (queryWorkflowResult.status === 'CompletedUnsuccessfully') {
+      await showAndLogErrorMessage(`Remote query execution failed. Error: ${queryWorkflowResult.error}`);
+
+    } else if (queryWorkflowResult.status === 'Cancelled') {
       await showAndLogErrorMessage('Remote query monitoring was cancelled');
+
+    } else if (queryWorkflowResult.status === 'InProgress') {
```
charisk, line 100: Since this should never happen, why is it not in the `else` clause (L105)?

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -15,6 +15,7 @@ export class AnalysesResultsManager {
 
   constructor(
     private readonly ctx: ExtensionContext,
+    readonly storagePath: string,
```
charisk, line 18: I don't think this is being used anywhere anymore so can be reverted.

extensions/ql-vscode/src/remote-queries/remote-query-result.ts
```diff
@@ -3,6 +3,7 @@ import { DownloadLink } from './download-link';
 export interface RemoteQueryResult {
   executionEndTime: Date;
   analysisSummaries: AnalysisSummary[];
+  artifactStorageDir: string;
```
charisk, line 6: Nit, but I think this should be called `artifactsStorageDir` (plural for artifacts)

Or, if it's made part of the `RemoteQuery` (rather than `RemoteQueryResult`) as a I mention in another comment perhaps it could just be called `storageDir` since it's meant to contain general query data.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
charisk: I reckon we'll need to be storing other information like actions workflow run id etc. We'll probably want to store the whole `RemoteQuery` object.

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
aeisenberg: In order for `assertNever` to work, all cases need to be explicitly handled. Otherwise, `assertNever` will have a compile error.

extensions/ql-vscode/src/remote-queries/download-link.ts
aeisenberg: I'll see what I can do. The difficulty is that there's no single place that is persistently accessible to make this information available in all places it's needed. It will need to be a bigger change. But, now that i have this piece in and it's working, I think I'm at a stage where I can add this abstraction.

aeisenberg:
See the message on the most recent commit.

charisk (approved):
Approving this since it works but I'm hoping we'll be able to tidy up some bits later on (happy to have a look when I've got some time).

extensions/ql-vscode/src/remote-queries/remote-query-result.ts
aeisenberg: Removed.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
aeisenberg: It is being used.

aeisenberg:
Yes, there will need to be cleanup later. Just trying to get this in incrementally. Thanks for being patient with this.

PR #1151, opened 2022-02-16
Integrate evaluator structured logging per query server instance

angelapwen:
This pull request is intended to ensure the query server runs correctly with structured evaluator logging enabled. It does not alert the user to the generated structured logfile, but a user could manually find it amongst other query artifacts. All structured logs related to all queries in a single query server instance will be output to a single file within `structured-evaluator-log.json` in the storage path that includes other query artifacts. The logs will not be minified for now, and will be set to the highest verbosity, as these are the use cases we expect from most of our users. 

Because we do not expose any configurations or indications to the user that we are integrating structured logging with this change, I have not updated the CHANGELOG. That said, we are now writing an extra file to user's disk so I am open to adding the change too. 

### Testing
This change has been tested locally against the current version of the CLI that will be included into 2.8.2, and does produce the structured evaluator logfile as intended. It continues to pass all existing CLI integration tests. 

### Expected future work 
Note that this change is intended to work with the current version of the query server (ie. will be released in 2.8.2), which will print all structured logs related to all queries in a single query server instance to a single file. When the changes to the query server allow for individual queries to have structured logs in separate files (expected in 2.8.3) the changes to the extension should include:

-  creating a directory for all the structured evaluator logfiles (perhaps making this configurable similar to the existing log directory?)
-  passing unique logfile names for each query to the query server
- changing the feature to be flagged under 2.8.3 

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -209,6 +209,24 @@
           "default": null,
           "description": "Path to a directory where the CodeQL extension should store query server logs. If empty, the extension stores logs in a temporary workspace folder and deletes the contents after each run."
         },
+        "codeQL.runningQueries.structuredEvaluatorLog": {
+          "type": [
+            "string",
+            null
+          ],
+          "default": "structured-evaluator-log.json",
+          "description": "Path to a file where the evaluator should store structured evaluator logs. If empty, the query server stores the logs in a default location and overwrites them after each query server run."
+        },
+        "codeQL.runningQueries.structuredEvaluatorLog.minify": {
+          "type": "boolean",
+          "default": false,
+          "description": "Minifies the JSON output of structured evaluator logs, removing all whitespace except a single newline between event objects. If not passed, logs are not minified."
+        },
+        "codeQL.runningQueries.structuredEvaluatorLog.verbosity": {
+          "type": "integer",
+          "default": 1,
+          "description": "Specifies the verbosity of the structured evaluator logs, from 1 (least verbose) to 5 (most verbose). If not passed or empty, verbosity is set to 1. "
+        },
         "codeQL.runningQueries.quickEvalCodelens": {
           "type": "boolean",
           "default": true,
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -209,24 +209,6 @@
           "default": null,
           "description": "Path to a directory where the CodeQL extension should store query server logs. If empty, the extension stores logs in a temporary workspace folder and deletes the contents after each run."
         },
-        "codeQL.runningQueries.structuredEvaluatorLog": {
-          "type": [
-            "string",
-            null
-          ],
-          "default": "structured-evaluator-log.json",
-          "description": "Path to a file where the evaluator should store structured evaluator logs. If empty, the query server stores the logs in a default location and overwrites them after each query server run."
-        },
-        "codeQL.runningQueries.structuredEvaluatorLog.minify": {
-          "type": "boolean",
-          "default": false,
-          "description": "Minifies the JSON output of structured evaluator logs, removing all whitespace except a single newline between event objects. If not passed, logs are not minified."
-        },
-        "codeQL.runningQueries.structuredEvaluatorLog.verbosity": {
-          "type": "integer",
-          "default": 1,
-          "description": "Specifies the verbosity of the structured evaluator logs, from 1 (least verbose) to 5 (most verbose)."
-        },
         "codeQL.runningQueries.quickEvalCodelens": {
           "type": "boolean",
           "default": true,
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -225,6 +225,8 @@
         "codeQL.runningQueries.structuredEvaluatorLog.verbosity": {
           "type": "integer",
           "default": 1,
+          "minimum": 1,
+          "maximum": 5,
           "description": "Specifies the verbosity of the structured evaluator logs, from 1 (least verbose) to 5 (most verbose). If not passed or empty, verbosity is set to 1. "
         },
         "codeQL.runningQueries.quickEvalCodelens": {
--- a/extensions/ql-vscode/src/queryserver-client.ts
+++ b/extensions/ql-vscode/src/queryserver-client.ts
@@ -169,6 +169,21 @@ export class QueryServerClient extends DisposableObject {
 
     if (await this.cliServer.cliConstraints.supportsOldEvalStats()) {
       args.push('--old-eval-stats');
+    } 
+
+    if (this.config.structuredEvalLogFile) {
+      args.push('--evaluator-log');
+      args.push(this.config.structuredEvalLogFile);
+
+      if (this.config.structuredEvalLogMinify) {
+        args.push('--evaluator-log-minify');
+      }
+
+      // 1 is default behavior in query server if argument is not passed.
+      if (this.config.structuredEvalLogVerbosity > 1) {
+        args.push('--evaluator-log-level');
+        args.push(this.config.structuredEvalLogVerbosity.toString());
+      }
     }
 
     if (this.config.debug) {
--- a/extensions/ql-vscode/src/queryserver-client.ts
+++ b/extensions/ql-vscode/src/queryserver-client.ts
@@ -172,19 +172,13 @@ export class QueryServerClient extends DisposableObject {
     }
 
     if (await this.cliServer.cliConstraints.supportsStructuredEvalLog()) {
-      if (this.config.structuredEvalLogFile) {
-        args.push('--evaluator-log');
-        args.push(this.config.structuredEvalLogFile);
+      args.push('--evaluator-log');
+      args.push(`${this.opts.contextStoragePath}/structured-evaluator-log.json`);
   
-        if (this.config.structuredEvalLogMinify) {
-          args.push('--evaluator-log-minify');
-        }
-        // 1 is default behavior in query server if argument is not passed.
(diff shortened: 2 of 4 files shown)


extensions/ql-vscode/package.json
```diff
@@ -209,6 +209,24 @@
...
           "description": "Path to a directory where the CodeQL extension should store query server logs. If empty, the extension stores logs in a temporary workspace folder and deletes the contents after each run."
         },
+        "codeQL.runningQueries.structuredEvaluatorLog": {
+          "type": [
+            "string",
+            null
+          ],
+          "default": "structured-evaluator-log.json",
+          "description": "Path to a file where the evaluator should store structured evaluator logs. If empty, the query server stores the logs in a default location and overwrites them after each query server run."
+        },
```
aeisenberg, line 219: What is the path relative to? If it must be an absolute path, then we should ensure this is only settable in local workspace settings.

extensions/ql-vscode/package.json
```diff
@@ -209,6 +209,24 @@
...
+        "codeQL.runningQueries.structuredEvaluatorLog.minify": {
+          "type": "boolean",
+          "default": false,
+          "description": "Minifies the JSON output of structured evaluator logs, removing all whitespace except a single newline between event objects. If not passed, logs are not minified."
```
aeisenberg, line 223: Is there ever a reason for this to be minified? Ie- do we need this option at all?

extensions/ql-vscode/package.json
```diff
@@ -209,6 +209,24 @@
...
+        },
+        "codeQL.runningQueries.structuredEvaluatorLog.verbosity": {
+          "type": "integer",
+          "default": 1,
+          "description": "Specifies the verbosity of the structured evaluator logs, from 1 (least verbose) to 5 (most verbose). If not passed or empty, verbosity is set to 1. "
```
aeisenberg, line 228: Second sentence is not necessary since the default is 1 and that is apparent to the user already.


```suggestion
          "description": "Specifies the verbosity of the structured evaluator logs, from 1 (least verbose) to 5 (most verbose)."
```

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -171,6 +171,22 @@ export class QueryServerClient extends DisposableObject {
...
+    if (await this.cliServer.cliConstraints.supportsStructuredEvalLog()) {
+      if (this.config.structuredEvalLogFile) {
+        args.push('--evaluator-log');
+        args.push(this.config.structuredEvalLogFile);
+  
+        if (this.config.structuredEvalLogMinify) {
+          args.push('--evaluator-log-minify');
+        }
+        // 1 is default behavior in query server if argument is not passed.
+        if (this.config.structuredEvalLogVerbosity > 1) {
```
aeisenberg, line 183: Should you throw an error if the value is > 5?

Is there a way to specify the 1 < x < 5 constraint in the setting itself?

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -171,6 +171,22 @@ export class QueryServerClient extends DisposableObject {
       args.push('--old-eval-stats');
     }
 
+    if (await this.cliServer.cliConstraints.supportsStructuredEvalLog()) {
+      if (this.config.structuredEvalLogFile) {
+        args.push('--evaluator-log');
+        args.push(this.config.structuredEvalLogFile);
```
aeisenberg, line 177: We should make sure the structured logs go in the same directory as all of the other query artifacts. I don't see much of a reason to allow users to set the name. But, we _do_ want to create a new command that allows users to open and view the logs (like we do for the regular log). See `codeQLQueryHistory.showQueryLog` command.

aeisenberg: It's a little more complicated since I see that there is a `codeQL.runningQueries.customLogDirectory` setting. I'm not even sure if that's working any more.

extensions/ql-vscode/package.json
angelapwen: Hm, like you mentioned in another comment, there doesn't seem to be a need for us to allow the user to pass this in (as the final state should be that the user views the log via our extension and not as a file itself). 

I can change this to a boolean — just so that users have the ability to have structured logs at all. Or should we enable them by default so that they would only have to toggle the "show" configuration?

aeisenberg: Is it an either/or? So, users can't produce both normal and structured logs? My thinking is that unless you're trying to process the logs programmatically, non-structured logs would be preferable?

So, we may want to have a boolean for a user to choose.

extensions/ql-vscode/package.json
angelapwen: I'm not entirely sure why a user would flip this option on — perhaps only if the file is expected to be massive? I don't know if that's a use case we really expect here. cc @edoardopirovano

extensions/ql-vscode/package.json
angelapwen: It's not an either-or on the query server side: the `--log-file` argument to the query server is still in effect and both can be produced (although I guess there should be some sort of check in place to make sure the file names passed in aren't the same!) 

I imagined having the "traditional" logs going to the existing logging directory and the structured evaluator logs also going to the same directory, but with a different name, if both are enabled.

extensions/ql-vscode/src/queryserver-client.ts
angelapwen: Just did both: specified the constraint in the setting (as part of `package.json`) and ignored the value if it wasn't within bounds!

extensions/ql-vscode/package.json
aeisenberg: Presumably, users want the structured logs so they can do things with it. Is this an important use case that we need to support right now?

Also, in the future, the extension itself will be using the logs to provide interesting data directly inline.

I'm fine with getting something out there that is in the general direction we want to go, and then iterating until we have all the bits of the feature in place. I think it's worth having a bit of a think of the use cases so we avoid implementing things we don't really need.

extensions/ql-vscode/package.json
edoardopirovano: Hmm, I guess this depends entirely on how the logs are being consumed. In my mind, the plan was that the extension would consume and manage these logs itself and not directly expose them to the user in any way. If this is the case, then in fact there's no reason for the logs to _not_ be minimised, since no human will ever be looking at them so we may as well save a bit of disk space. The disk-space savings are non-negligible, e.g. for a standard suite of queries on a large-ish database:

image

If, however, there's any plans for the logs to be viewed directly by a user, then it makes sense to have them not be minified. Also, I guess having them unminified is helpful for _us_ when debugging things.

extensions/ql-vscode/package.json
```diff
@@ -209,6 +209,24 @@
...
+          "description": "Path to a file where the evaluator should store structured evaluator logs. If empty, the query server stores the logs in a default location and overwrites them after each query server run."
+        },
+        "codeQL.runningQueries.structuredEvaluatorLog.minify": {
+          "type": "boolean",
+          "default": false,
+          "description": "Minifies the JSON output of structured evaluator logs, removing all whitespace except a single newline between event objects. If not passed, logs are not minified."
+        },
+        "codeQL.runningQueries.structuredEvaluatorLog.verbosity": {
+          "type": "integer",
+          "default": 1,
```
edoardopirovano, line 227: I think we might just want to make the extension always pass in the highest verbosity. As we develop features of the extension to display different information from the structured logs, it will be annoying if the information may or may not be present depending on some user-supplied value. It seems easier to just make sure all the information is always there.

extensions/ql-vscode/package.json
angelapwen: Thanks for sharing the graph. Hm 🤔 I think it's possible that a user might want to pull the logfile themselves — perhaps they would want to `grep` the log themselves, or something else that our features do not yet support. 

Though, can a user view other existing files in the standard VSCode query run artifacts?)

extensions/ql-vscode/src/queryserver-client.ts
angelapwen: I have updated the PR to create the file in the same directory as the other query artifacts. When the change to the CLI is made to create logfiles per query (see PR description), we can create a directory for the structured log files (potentially a custom one as well). 

After we are able to consume the individual logfiles per query, we could add the open/view log command.

extensions/ql-vscode/src/queryserver-client.ts
angelapwen: We decided to hard-code this to highest verbosity for now (and anticipate this to be the same for the per-query logs).

extensions/ql-vscode/package.json
angelapwen: The updated commits no longer allow the user to configure a custom location for the logfile. In the future we may want to allow a custom additional location the way the other logfiles do, though.

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -171,6 +171,16 @@ export class QueryServerClient extends DisposableObject {
       args.push('--old-eval-stats');
     }
 
+    if (await this.cliServer.cliConstraints.supportsStructuredEvalLog()) {
+      args.push('--evaluator-log');
+      args.push(`${this.opts.contextStoragePath}/structured-evaluator-log.json`);
```
aeisenberg, line 176: This is fine for now. And we will make sure this writes to the query storage location later.

aeisenberg (approved):
Looks good. Let's get this in and get people using it.

aeisenberg:
Should we remove the `DO NOT MERGE YET` label?

PR #1157, opened 2022-02-18
Ensure `--addtional-packs` arg not used for empty workspace

### LINKED ISSUE #1066: Resolving upgrades takes a long time when the current workspace has no workspace roots
**Describe the bug**

Resolving upgrades will scan the user's home directory for upgrades if the current workspace has no open folders. Normally, all workspace folders are concatenated and passed to the `--additional-packs` argument. However, when there are no folders, the value is an empty string, which corresponds to the CWD of the vscode process (typically, the user's home directory).

**To reproduce**

1. Open an empty workspace
2. Add a database
3. Remove that database's source folder
4. Open a quick query editor tab
5. run a simple query like `select 1`
6. BUG: `resolve upgrades` stage takes a long time

The proper solution is to avoid using an `--additional-packs` argument if there are no workspace folders. Also, we should verify that only `file:` based folders are included. Remote folders and archive folders should be excluded as well. This may already be happening, but we should verify this.

aeisenberg:
Fixes #1066

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -514,8 +514,7 @@ export class CodeQLCliServer implements Disposable {
   async resolveLibraryPath(workspaces: string[], queryPath: string): Promise<QuerySetup> {
     const subcommandArgs = [
       '--query', queryPath,
-      '--additional-packs',
-      workspaces.join(path.delimiter)
+      ...this.getAdditionalPacksArg(workspaces)
     ];
     return await this.runJsonCodeQlCliCommand<QuerySetup>(['resolve', 'library-path'], subcommandArgs, 'Resolving library paths');
   }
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -929,6 +930,12 @@ export class CodeQLCliServer implements Disposable {
         throw new Error('No distribution found');
     }
   }
+
+  private getAdditionalPacksArg(paths: string[]): string[] {
```
henrymercer, line 934: Minor: Is `Args` more appropriate here since the return type is an array?

henrymercer (approved):

extensions/ql-vscode/src/cli.ts
aeisenberg: I used `Arg` (singular) since this method returns the `--additional-packs` argument.

PR #1163, opened 2022-02-22
Allow multiple analyses for same repo to be downloaded

aeisenberg:
Removes the limitation specified in #1089 where analyses for the same
repo and different queries will overwrite each other.

This PR does not save analysis results across restarts. After restarting, results will need to be re-downloaded. That is what I will work on next.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
@@ -10,15 +10,16 @@ import * as os from 'os';
 import { sarifParser } from '../sarif-parser';
 
 export class AnalysesResultsManager {
-  // Store for the results of various analyses for a single remote query.
-  private readonly analysesResults: AnalysisResults[];
+  // Store for the results of various analyses for each remote query.
+  // The key is the queryId and is also the name of the directory where results are stored.
+  private readonly analysesResults: Map<string, AnalysisResults[]>;
 
   constructor(
     private readonly ctx: ExtensionContext,
     readonly storagePath: string,
     private readonly logger: Logger,
   ) {
-    this.analysesResults = [];
+    this.analysesResults = new Map();
   }
 
   public async downloadAnalysisResults(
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -139,6 +146,6 @@ export class AnalysesResultsManager {
   }
 
   private isAnalysisDownloaded(analysis: AnalysisSummary): boolean {
-    return this.analysesResults.some(x => x.nwo === analysis.nwo);
+    return (this.analysesResults.get(analysis.downloadLink.queryId) || []).some(x => x.nwo === analysis.nwo);
```
charisk, line 149: Here and in a few other places we do `this.analysesResults.get(queryId)` .. I think it'd be easier to read if we had a `getQueryResults` and did

```suggestion
  return this.getQueryResults(queryId).some(x => x.nwo === analysis.nwo);
```

where `getQueryResults` is:
```ts
private getQueryResults(queryId: string) { 
  return this.analysesResults.get(analysis.downloadLink.queryId) || []
}
```

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -302,7 +302,7 @@ export async function runRemoteQuery(
       message: 'Sending request'
     });
 
-    const workflowRunId = await runRemoteQueriesApiRequest(credentials, 'main', language, repositories, owner, repo, base64Pack, dryRun);
+    const workflowRunId = await runRemoteQueriesApiRequest(credentials, 'better-errors', language, repositories, owner, repo, base64Pack, dryRun);
```
charisk, line 305: Just checking we're intentionally checking this in. Fine if we decided to do that!

charisk (approved):

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
aeisenberg: There is already the `getAnalysesResults`, which is public. It makes a copy of the array before returning it. I can't use this method internally, since there is one place that relies on using the original array. Instead, I will make a `internalGetAnalysesResults` method that returns the results without copying them.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
aeisenberg: This PR is broken unless we point to the `better-errors` branch. When that branch is merged on qc-run2, then we can change this back to `main`. I'll add a comment to that effect.

PR #1164, opened 2022-02-23
Avoid downloading a result if it already exists

aeisenberg:
This commit adds a check if a results artifact already exists before
trying to download it again.

This is not a complete solution since the page icon will still have a
download button even if the artifact already exists. In this case,
clicking on it will avoid downloading it a second time.

The next step is to read in the downloaded artifacts and display them
appropriately.

With this PR, saving remote queries across restarts is working in its most basic state. Once this PR is ready, I will take a step back, write some tests, ensure there are no problems, and do a release.

## Checklist


extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -145,7 +148,11 @@ export class AnalysesResultsManager {
     return queryResults;
   }
 
-  private isAnalysisDownloaded(analysis: AnalysisSummary): boolean {
+  private isAnalysisInMemory(analysis: AnalysisSummary): boolean {
     return (this.analysesResults.get(analysis.downloadLink.queryId) || []).some(x => x.nwo === analysis.nwo);
   }
+
+  async isAnalysisDownloaded(analysis: AnalysisSummary): Promise<boolean> {
```
charisk, line 155: Where is this used? Should it be async?

aeisenberg: This is not used (yet). Once we start trying to proactively determine what has already been downloaded, we're going to need this function. I had started trying to do this, but gave up when I realized it is harder than I thought. I'll work on this part later. And will remove this function for now.

charisk (approved):

PR #1166, opened 2022-02-24
Add unit tests for query history and remote queries

aeisenberg:
Adds some tests for reading in the history and manipulating.
There are some more tests to come later. Maybe in another PR, maybe in
this one.

Note that this PR uses a new node 16 API String.prototype.replaceAll.
I think this is ok since vscode ships with node 16. If this causes
problems, I can separate to a different PR and we can discuss there.

## Checklist


extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
```diff
@@ -0,0 +1,252 @@
...
+
+  const EXTENSION_PATH = path.join(__dirname, '../../../');
+  const STORAGE_DIR = path.join(tmpDir.name, 'remote-queries');
+  const asyncNoop = async () => { /** noop */ };
+
+  let sandbox: sinon.SinonSandbox;
+  let qhm: QueryHistoryManager;
+  let rawQueryHistory: any;
+  let remoteQueryResult0: RemoteQueryResult;
+  // let remoteQueryResult1: RemoteQueryResult;
```
charisk, line 37: Should this be removed?

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
```diff
@@ -0,0 +1,252 @@
...
+  afterEach(() => {
+    deleteHistoryState();
+  });
+
+  beforeEach(() => {
+    sandbox = sinon.createSandbox();
+    disposables = new DisposableBucket();
+    rawQueryHistory = fs.readJSONSync(path.join(STORAGE_DIR, 'workspace-query-history.json'));
+    remoteQueryResult0 = fs.readJSONSync(path.join(STORAGE_DIR, 'queries', rawQueryHistory[0].queryId, 'query-result.json'));
+    // remoteQueryResult1 = fs.readJSONSync(path.join(STORAGE_DIR, 'queries', rawQueryHistory[1].queryId, 'query-result.json'));
```
charisk, line 55: Should this be removed?

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    for (const p of walk(STORAGE_DIR)) {
+      replacePlaceholder(path.join(p));
+    }
+  }
+
+  function deleteHistoryState() {
+    fs.removeSync(STORAGE_DIR);
+  }
+
+  function* walk(dir: string): IterableIterator<string> {
```
charisk, line 233: Nice! At first I was thinking it might be an overkill to have an a generator for tests, but there's no harm in having it and puts us in a good position if we want to test a lot/big files.

charisk (approved):
Thanks for adding some much needed tests!

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -106,7 +106,8 @@ export class AnalysesResultsManager {
     const resultsForQuery = this.internalGetAnalysesResults(queryId);
     resultsForQuery.push(analysisResults);
     this.analysesResults.set(queryId, resultsForQuery);
-    void publishResults(resultsForQuery);
+    void publishResults([...resultsForQuery]);
+    const pos = resultsForQuery.length - 1;
```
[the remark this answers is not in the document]
aeisenberg: I've always like immutable.js for things like this, but I don't like introducing it into an ongoing project. Maybe it's not such a bad idea in this case.

We do need to be careful. Now that we are saving data across restarts, any change in format will cause error when there is an upgrade. It's not a _terrible_ situation since the worst case is that we need to delete some old queries on upgrade.

I should add version number to each saved artifact and the extension can read this and upgrade it appropriately when we change versions.

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
aeisenberg: I'll be writing more tests and will likely need this variable.

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
aeisenberg: I should probably move this to a shared location since it might be useful in other places.

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
charisk: Shall we add it then? Happy either way.

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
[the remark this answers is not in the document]
charisk: I'd be up for using immutable.js or any other way to enforce immutability.

Agree about versions - we'll need to be careful. Perhaps the overall query history store has a version? We can always deal with it the first time we need a new version I think.

aeisenberg:
Putting this into draft mode so I can work on it more.

extensions/ql-vscode/package.json
```diff
@@ -1064,7 +1064,7 @@
     "@types/jszip": "~3.1.6",
     "@types/mocha": "^9.0.0",
     "@types/nanoid": "^3.0.0",
-    "@types/node": "^12.14.1",
+    "@types/node": "^16.11.25",
```
aeisenberg, line 1067: Need to update to v16 to get the `replaceAll` method.

extensions/ql-vscode/src/cli.ts
```diff
@@ -403,7 +403,7 @@ export class CodeQLCliServer implements Disposable {
     try {
       if (cancellationToken !== undefined) {
         cancellationRegistration = cancellationToken.onCancellationRequested(_e => {
-          tk(child.pid);
+          tk(child.pid || 0);
```
aeisenberg, line 406: In v16, `pid` is potentially undefined.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -676,8 +676,10 @@ export class QueryHistoryManager extends DisposableObject {
       // show original query file on double click
       await this.handleOpenQuery(finalSingleItem, [finalSingleItem]);
     } else {
-      // show results on single click
-      await this._onWillOpenQueryItem.fire(finalSingleItem);
+      // show results on single click only if query is completed successfully.
+      if (finalSingleItem.status === QueryStatus.Completed) {
+        await this._onWillOpenQueryItem.fire(finalSingleItem);
+      }
```
aeisenberg, line 682: Fix small bug to avoid an error when someone clicks on an in progress item.

extensions/ql-vscode/src/blob.d.ts
```diff
@@ -1,11 +0,0 @@
-/**
```
aeisenberg, line 1: With node 16, this is no longer needed.

extensions/ql-vscode/src/pure/disposable-object.ts
```diff
@@ -83,3 +83,17 @@ export abstract class DisposableObject implements Disposable {
     }
   }
 }
+
+/**
+ * A simple disposable object that does nothing other than contain a list of disposable objects.
+ * This is useful for implementing a `Disposable` that owns other disposable objects.
+ */
+export class DisposableBucket extends DisposableObject {
```
aeisenberg, line 91: Used for tests.

aeisenberg:
@charisk can you take another look? I've made a lot of changes.

extensions/ql-vscode/src/pure/disposable-object.ts
charisk: Shall we move this to the tests dir?

extensions/ql-vscode/src/vscode-tests/no-workspace/data/remote-queries/queries/MRVA Integration test 2-UL-vbKAjP8ffObxjsp7hN/query-result.json
```diff
@@ -0,0 +1,18 @@
+{
+  "executionEndTime": 1645645150738,
+  "analysisSummaries": [
+    {
+      "nwo": "github/vscode-codeql",
+      "resultCount": 5,
+      "fileSizeInBytes": 81237,
+      "downloadLink": {
+        "id": "171544171",
+        "urlPath": "/repos/dsp-testing/qc-run2/actions/artifacts/171544171",
```
charisk, line 10: Perhaps shouldn't use `dsp-testing` testing?

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
```diff
@@ -0,0 +1,342 @@
...
+    expect(spy.getCall(0).args[0]).to.deep.eq(rawQueryHistory[0]);
+    expect(spy.getCall(1).args[0]).to.deep.eq(rawQueryHistory[1]);
+    expect(spy.callCount).to.eq(2);
+
+    expect(qhm.treeDataProvider.allHistory[0]).to.deep.eq(rawQueryHistory[0]);
+    expect(qhm.treeDataProvider.allHistory[1]).to.deep.eq(rawQueryHistory[1]);
+    expect(qhm.treeDataProvider.allHistory.length).to.eq(2);
+  });
+
+  it('should remove and then a query from history', async () => {
```
charisk, line 99:
```suggestion
  it('should remove and then add query from history', async () => {
```

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
```diff
@@ -0,0 +1,342 @@
...
+      sandbox.stub(Credentials, 'initialize').resolves(mockCredentials);
+
+      arm = new AnalysesResultsManager(
+        {} as ExtensionContext,
+        path.join(STORAGE_DIR, 'queries'),
+        mockLogger
+      );
+    });
+
+    it('should avoid downloading an analysis result', async () => {
```
charisk, line 192:
```suggestion
    it('should avoid re-downloading an analysis result', async () => {
```

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
```diff
@@ -0,0 +1,342 @@
...
+
+  describe('AnalysisResultsManager', () => {
+
+    let mockCredentials: any;
+    let mockOctokit: any;
+    let mockLogger: any;
+    let arm: AnalysesResultsManager;
+
+    beforeEach(() => {
+      mockOctokit = {
```
charisk, line 174: I'm a little surprised we have to mock at this level and not at `gh-api-client` level which is what the `AnalysesResultsManager` is using.. Is this because we want to test as closer to the "edge" as possible?

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
```diff
@@ -0,0 +1,342 @@
...
+      const sarif = fs.readJSONSync(path.join(STORAGE_DIR, 'queries', rawQueryHistory[0].queryId, '171543249', 'results.sarif'));
+      const queryResults = sarif.runs
+        .flatMap((run: any) => run.results)
+        .map((result: any) => ({ message: result.message.text }));
+
+      expect(publisher.getCall(1).args[0][0].results).to.deep.eq(queryResults);
+    });
+  });
+
+  // Since this test changes the state of the query history manager, we need to copy the original
```
charisk, line 320: This comment doesn't look like it belongs here exactly - should it be near the test?

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
```diff
@@ -0,0 +1,342 @@
...
+      const analysisSummaries1 = [...remoteQueryResult1.analysisSummaries];
+
+      await arm.downloadAnalysesResults(analysisSummaries0, undefined, publisher);
+      await arm.downloadAnalysesResults(analysisSummaries1, undefined, publisher);
+
+      const result0 = arm.getAnalysesResults(rawQueryHistory[0].queryId);
+      const result1 = arm.getAnalysesResults(rawQueryHistory[1].queryId);
+
+      // Shoule be equal, but not equivalent
+      expect(result0).to.deep.eq((arm as any).analysesResults.get(rawQueryHistory[0].queryId));
```
charisk, line 298: This looks like it's interrogating the private state of the AnalysisResultsManager which doesn't feel right.

Should we not instead check that what we get back from `getAnalysesResults` is what we "downloaded"?

extensions/ql-vscode/src/vscode-tests/test-helpers.ts
```diff
@@ -0,0 +1,21 @@
+import * as path from 'path';
```
charisk, line 1: Can I make a suggestion that we start using more specific util modules? e.g. `file-utils.ts` rather than `test-helpers` which could be a bucket for pretty much anything.

extensions/ql-vscode/src/vscode-tests/no-workspace/data/remote-queries/queries/MRVA Integration test 2-UL-vbKAjP8ffObxjsp7hN/query-result.json
aeisenberg: I can change the repo.

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
aeisenberg: I'm specifically testing here that the results are an equivalent copy of what is stored by the `AnalysesResultsManager`. Instead of inspecting private state, I could call `arm.getAnalysesResults` twice and ensure that the objects returned are equivalent but differemnt.

charisk: Ah I see, I think that would make more sense.

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
charisk: Oh sorry I didn't realise that was the case. That's a bit yucky, but appreciate it would be more work at this point!

charisk (approved):
Approving since all the comments are actually small suggestions!

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-query-history.test.ts
aeisenberg: I added these tests because keeping them the same instance created weird race conditions, so just wanted to make sure we don't do this again.

Another reason to use immutable.

aeisenberg:
All comments addressed.

PR #1167, opened 2022-02-24
Add new endpoints of the query server protocol

edoardopirovano:
Adds in the interfaces and methods for calling two new (still unreleased) end-points for the query server in the VS code extension. I haven't called these anywhere yet nor added tests because (a) we haven't released a version of CodeQL with these in yet (b) we probably need to discuss a bit quite how/where we want to rotate the logs.

I have, however, tested this manually (with a query server built from a branch adding the endpoints in) by doing the most naive thing of adding in the calls at a suitable point in the `run` method of `run-queries.ts` and can confirm that they work.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/pure/messages.ts
+++ b/extensions/ql-vscode/src/pure/messages.ts
@@ -646,6 +646,31 @@ export interface ClearCacheParams {
    */
   dryRun: boolean;
 }
+
+/**
+ * Parameters to start a new structured log
+ */
+ export interface StartLogParams {
+  /**
+   * The dataset for which we want to start a new structured log
+   */
+  db: Dataset;
+  /**
+   * The path where we want to place the new structured log
+   */
+  logPath: string;
+}
+
+/**
+ * Parameters to terminate a structured log
+ */
+ export interface EndLogParams {
+  /**
+   * The dataset for which we want to terminated the log
+   */
+  db: Dataset;
+}
+
 /**
  * Parameters for trimming the cache of a dataset
  */
@@ -680,6 +705,26 @@ export interface ClearCacheResult {
    * deleted.
    */
   deletionMessage: string;
+}
+
+/**
+ * The result of starting a new structured log.
+ */
+export interface StartLogResult {
+  /**
+   * A user friendly message saying what happened.
+   */
+  successMessage: string;
+}
+
+/**
+ * The result of terminating a structured.
+ */
+export interface EndLogResult {
+  /**
+   * A user friendly message saying what happened.
+   */
+  successMessage: string;
 }
 
 /**
@@ -1017,6 +1062,16 @@ export const compileUpgrade = new rpc.RequestType<WithProgressId<CompileUpgradeP
  * Compile an upgrade script to upgrade a dataset.
  */
 export const compileUpgradeSequence = new rpc.RequestType<WithProgressId<CompileUpgradeSequenceParams>, CompileUpgradeSequenceResult, void, void>('compilation/compileUpgradeSequence');
+
+/**
+ * Start a new structured log in the evaluator, terminating the previous one if it exists
+ */
+ export const startLog = new rpc.RequestType<WithProgressId<StartLogParams>, StartLogResult, void, void>('evaluation/startLog');
+
+/**
+ * Terminate a structured log in the evaluator
+ */
+ export const endLog = new rpc.RequestType<WithProgressId<EndLogParams>, StartLogResult, void, void>('evaluation/endLog');
 
 /**
  * Clear the cache of a dataset


extensions/ql-vscode/src/pure/messages.ts
```diff
@@ -682,6 +707,26 @@ export interface ClearCacheResult {
...
 }
 
+/**
+ * The result of starting a new structured log.
+ */
+export interface StartLogResult {
+  /**
+   * A user friendly message saying what happened.
+   */
+  successMessage: string;
```
aeisenberg, line 717: What message is written on failure? It might be too hard to change the CLI here, but should this be `completionMessage` instead to allow for the possibility of failure?

extensions/ql-vscode/src/pure/messages.ts
```diff
@@ -682,6 +707,26 @@ export interface ClearCacheResult {
...
+}
+
+/**
+ * The result of terminating a structured.
+ */
+export interface EndLogResult {
+  /**
+   * A user friendly message saying what happened.
+   */
+  successMessage: string;
```
aeisenberg, line 727: Same as above.

extensions/ql-vscode/src/pure/messages.ts
```diff
@@ -1018,6 +1063,16 @@ export const compileUpgrade = new rpc.RequestType<WithProgressId<CompileUpgradeP
  */
 export const compileUpgradeSequence = new rpc.RequestType<WithProgressId<CompileUpgradeSequenceParams>, CompileUpgradeSequenceResult, void, void>('compilation/compileUpgradeSequence');
 
+/**
+ * Start a new structured log in the evaluator, terminating the previous one if it exists
```
aeisenberg, line 1067: Does this mean we can only have one structured log at the same time? What if we have multiple queries running? I guess there's no way to separate logs for each of them.

angelapwen: Edoardo and I discussed this this morning — it sounds like if we have multiple queries running at the same time, the interleaved messages will write to the log file for the most recent query. This is also what happens to existing logs so we thought this would be acceptable for the time being.

extensions/ql-vscode/src/pure/messages.ts
edoardopirovano: Currently this cannot fail. I agree that calling it something like `completionMessage` to allow for the possibility of a failure in the future seems reasonable. This doesn't even _really_ need to be an object, but it seemed better to make it one, also for future extensibility.

extensions/ql-vscode/src/pure/messages.ts
edoardopirovano: As above.

extensions/ql-vscode/src/pure/messages.ts
aeisenberg: OK...At the time we wrote the interleaved  log messages, there wasn't an easy way to distinguish messages for different queries.

 With structured logging, in theory, it should be easier to add a unique identifier that distinguishes where the query came from. 

When we invoke a query run, the extension passes in an id. See: https://github.com/github/vscode-codeql/blob/39f9c082b988610cf291df1d5e1eecbab845e7d4/extensions/ql-vscode/src/run-queries.ts#L123-L132

This id is used for callbacks, cancellation, and progress. Is there any possibility of adding the id to the structure log message? This is not something for now, but some future work.

edoardopirovano: Indeed, as I discussed with Angela I don't think we can do any better than this (and already don't for the unstructured logs). The problem is that from the evaluator's point of view there isn't really a concept of what evaluations are being done for which query. When we start a query we just enqueue the whole dependency graph of predicates that it needs to evaluate that query, and that can partially overlap with other already running queries.

aeisenberg:
Is this something you want to get into the extension now? Or is this preparation for something that doesn't need to be merged until later?

extensions/ql-vscode/src/pure/messages.ts
aeisenberg: Interesting, so in theory, if predicates are shared across queries, the same work can be done for both of them?

edoardopirovano:
> Is this something you want to get into the extension now? Or is this preparation for something that doesn't need to be merged until later?

More just a convenient way to share some code with Angela, and to give you (and others on the team) a chance to have some input on the protocol. This doesn't necessarily need merging (and indeed I definitely wouldn't merge until the corresponding internal PR is merged).

extensions/ql-vscode/src/pure/messages.ts
angelapwen: Yes, I believe within a single run of the query server, the more recent queries can pull from some cache, but please correct me if incorrect @edoardopirovano !

edoardopirovano: Concretely in the evaluator what happens is that the shared predicate will become a single node in the evaluation graph that is needed by both queries. Then, it's evaluated once and the result is used by both queries wherever it is needed (directly in the query, or in one of it's transitive dependencies).

We do track demand in the evaluation graph, so for many predicates we can associate a query that most recently created demand for it in the `queryCaysingWork` field of the predicate start event. However, this is a little misleading because often this predicate would've been needed by other queries, so it isn't particularly meaningful to associate the time spend evaluating it to one query.

Rather than trying to achieve something similar to what we do with the cancellation tokens for the logs, I think I would much prefer to have a warning that running queries in parallel on the same DB will produce unreliable logs and timing information. Perhaps the extension could even have an option to queue queries for sequential running rather than running them in parallel, for users who care about getting accurate benchmarking information.

PR #1168, opened 2022-02-24
Preemptively add a version number to the query history json file

aeisenberg:
Since we are now storing query history on disk, we will need to handle
situations where versions change. For now, there is only version 1. In
the future, we may need to make breaking changes to this format and we
need the flexibility to detect and possibly handle different versions.

In this case, users don't often downgrade their vscode versions, so
most likely, we only need to be forward compatible. Ie- we need to
handle moving from v1 to v2, but not the other way around.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
@@ -18,7 +18,7 @@ import { slurpQueryHistory, splatQueryHistory } from '../../query-serialization'
 chai.use(chaiAsPromised);
 const expect = chai.expect;
 
-describe.only('query-results', () => {
+describe('query-results', () => {
   let disposeSpy: sinon.SinonSpy;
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
```diff
@@ -18,7 +18,7 @@ import { slurpQueryHistory, splatQueryHistory } from '../../query-serialization'
 chai.use(chaiAsPromised);
 const expect = chai.expect;
 
-describe('query-results', () => {
+describe.only('query-results', () => {
```
charisk, line 21:
```suggestion
describe('query-results', () => {
```

charisk (approved):
Good stuff! Just need to revert a small change that has crept in.

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
aeisenberg: Thanks for catching that and re-reviewing.

PR #1171, opened 2022-02-28
Add SARIF processing and basic alert rendering

charisk:
Updated logic that reads the SARIF file in order to extract information about analysis results and then used that data to start rendering a list of results. 

This is quite basic at the minute and will be fleshed out in future PRs. More information in the internal issue.

Please let me know if I'm using incorrect terminology.

To view the changes, enable rending or analysis results in RemoteQueries.tsx L333

## Checklist
N/A:


extensions/ql-vscode/src/remote-queries/sarif-processor.ts
```diff
@@ -0,0 +1,137 @@
+import * as sarif from 'sarif';
+
+import { AnalysisAlert, ResultSeverity } from './shared/analysis-result';
+
+export class ProcessedSarif {
```
aeisenberg, line 5: Is there a reason to make this a class? Is it because you are planning on adding more behaviour to it over time? However, if these bits are going to need to be sent to the webview, or serialized to disk for query history, we should avoid classes with behaviour since that makes deserialization hard.

extensions/ql-vscode/src/remote-queries/sarif-processor.ts
```diff
@@ -0,0 +1,137 @@
...
+  public constructor(
+    public readonly alerts: AnalysisAlert[],
+    public readonly errors: string[]) { }
+}
+
+export function processSarif(sarifLog: sarif.Log): ProcessedSarif {
```
aeisenberg, line 11: Could you add some unit tests for this function. It is nicely contained and should be straight forward.

extensions/ql-vscode/src/remote-queries/sarif-processor.ts
```diff
@@ -0,0 +1,137 @@
...
+          }
+        });
+      }
+    }
+  }
+
+  return new ProcessedSarif(alerts, errors);
+}
+
+function getSeverity(sarifRun: sarif.Run, result: sarif.Result): ResultSeverity {
```
aeisenberg, line 109: And unit tests as well for this one?

extensions/ql-vscode/src/remote-queries/sample-data.ts
```diff
@@ -99,7 +99,23 @@ export const sampleRemoteQueryResult: RemoteQueryResult = {
...
+  {
+    message: 'This shell command depends on an uncontrolled [absolute path](1).',
+    severity: 'Error',
+    filePath: 'npm-packages/meteor-installer/config.js',
+    contextRegion: {
+      startLine: 253,
+      endLine: 257,
+      text: '  if (isWindows()) {\n    //set for the current session and beyond\n    child_process.execSync(`setx path "${meteorPath}/;%path%`);\n    return;\n  }\n',
+    },
+    codeRegion: {
```
shati-patel, line 112: (Ran out of time for a proper review today, but here's a small comment before I forget 💭 )

Minor: I'm still getting a bit confused about what "codeRegion" means 😅 Perhaps something like "highlightedRegion" is clearer?

extensions/ql-vscode/src/remote-queries/sarif-processor.ts
```diff
@@ -0,0 +1,137 @@
...
+  const defaultSeverity = 'Warning';
+
+  const ruleId = result.ruleId;
```
adityasharad, line 112: I think you can expect to find `result.rule.id`, which will be the same as `result.ruleId`, but is the more stable/meaninguful field.

extensions/ql-vscode/src/remote-queries/sarif-processor.ts
```diff
@@ -0,0 +1,137 @@
...
+  if (!ruleId) {
+    return defaultSeverity;
+  }
+
+  const rule = sarifRun.tool.driver.rules?.find(r => r.id === ruleId);
```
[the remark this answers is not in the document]
adityasharad: Generally speaking, this highlights the fact that we're doing the same kinds of SARIF processing task in various places, such as the CodeQL CLI and now the VS Code extension. Unfortunately there isn't yet a SARIF SDK for JavaScript or Java, which would make this easier. The best thing we can do for now is share knowledge internally so that the various processors behave consistently on SARIF produced by CodeQL.

adityasharad (commented):
Nice. Chiming in with a few suggestions on SARIF processing.

extensions/ql-vscode/src/remote-queries/sample-data.ts
charisk: I also find it a bit confusing, but was trying to stay close to the SARIF terms. Since you're having the same issue as me, I think it's a signal that it's worth changing the names. How about `codeSnippet` and `highlightedRegion`?

shati-patel: > I also find it a bit confusing, but was trying to stay close to the SARIF terms. Since you're having the same issue as me, I think it's a signal that it's worth changing the names. How about `codeSnippet` and `highlightedRegion`?

Yeah, I think we should either match the SARIF terms exactly (i.e. `contextRegion` and just `region`), or pick something more self-explanatory! I'd vote for your suggestion of `codeSnippet` and `highlightedRegion` 😀

extensions/ql-vscode/src/remote-queries/sample-data.ts
```diff
@@ -99,7 +99,23 @@ export const sampleRemoteQueryResult: RemoteQueryResult = {
 };
 
 
-const createAnalysisResults = (n: number) => Array(n).fill({ 'message': 'Sample text' });
+const createAnalysisResults = (n: number) => Array(n).fill(
```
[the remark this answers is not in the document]
charisk: Agree we can do something different here now that we have a query history (perhaps have a util that puts a dummy query in your history)

extensions/ql-vscode/src/remote-queries/sarif-processor.ts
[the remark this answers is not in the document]
charisk: Thanks so much for this - super helpful. I've updated the logic, hopefully I've translated sentences to code correctly 😬

extensions/ql-vscode/src/remote-queries/sample-data.ts
charisk: Updated!

extensions/ql-vscode/src/remote-queries/sarif-processor.ts
charisk: Thanks for the nudge, I was being lazy. I've added a few unit tests for the various functions. They don't cover everything as there are way too many possibilities, but I think they cover enough.

extensions/ql-vscode/src/remote-queries/shared/analysis-result.ts
```diff
@@ -3,9 +3,28 @@ export type AnalysisResultStatus = 'InProgress' | 'Completed' | 'Failed';
...
+  highlightedRegion: HighligtedRegion
 }
+
+export interface CodeSnippet {
+  startLine: number;
+  endLine: number;
+  text: string;
+}
+
+export interface HighligtedRegion {
```
shati-patel, line 23: 🤏🏽 tiny typo 😄 (here and above)

```suggestion
export interface HighlightedRegion {
```

charisk: 🤦‍♀️

extensions/ql-vscode/src/remote-queries/sarif-processor.ts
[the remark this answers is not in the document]
adityasharad: Looks good!

extensions/ql-vscode/src/remote-queries/sarif-processor.ts
charisk: It started as something but ended up being something else!

I've deleted the model completely.

charisk:
@aeisenberg @shati-patel I've just pushed some code changes to make the code snippets a bit less ugly.

The original plan was to use monaco editor but I'm having too much trouble with it so for now I'll just try and make it look decent with plain React/CSS. 

Apologies for the churn!

extensions/ql-vscode/src/remote-queries/view/AnalysisAlertResult.tsx
```diff
@@ -0,0 +1,171 @@
...
+  lineNumber: number,
+  highlightedRegion: HighlightedRegion
+}) => {
+  if (!shouldHighlightLine(lineNumber, highlightedRegion)) {
+    return <PlainLine text={line} />;
+  }
+
+  const section1 = line.substring(0, highlightedRegion.startColumn - 1);
+  const section2 = line.substring(highlightedRegion.startColumn - 1, highlightedRegion.endColumn - 1);
+  const section3 = line.substring(highlightedRegion.endColumn - 1, line.length);
```
aeisenberg, line 113: Make sure that you can gracefully handle invalid regions. These regions can be created by user queries which could return invalid source locations. Eg- `endLine < startLine` or `endLine == startLine && endColumn <= startColumn`. In general (I believe) that end column must be strictly greater than start column on the same line.

Also, note that CodeQL/Sarif source locations have a slightly different semantics than vscode locations. See here for how it is handled for local queries: https://github.com/github/vscode-codeql/blob/735f177283ed7e02674ed6ff1af889d61f9a878c/extensions/ql-vscode/src/interface-utils.ts#L64-L78

I don't know if this will affect remote queries since everything is self contained, but just something to be aware of.

It would be fine to just keep the handling simpler for now and make it more robust in a future PR.

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -0,0 +1,179 @@
+import * as sarif from 'sarif';
```
aeisenberg, line 1: Note there is already `extensions/ql-vscode/src/pure/sarif-utils.ts`. Does it make sense to move some of these functions over to that file? The other file can be renamed to `sarif-processing.ts` since that is a better name.

charisk: I'd like to do that in a separate PR if that's okay - I noticed the file after I made my changes and I think there are potentially some conflicting function names etc.

extensions/ql-vscode/src/remote-queries/view/AnalysisAlertResult.tsx
charisk: Thanks! I've added some basic validation but will probably revisit to tidy things up in later PR.

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
aeisenberg: That's fine. There is some cleanup work to be done and it doesn't need to happen now.

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -177,3 +185,19 @@ export function tryGetRule(
...
+
+  if (alert.codeSnippet.startLine > alert.codeSnippet.endLine) {
+    errors.push('The code snippet start line is greater than the end line');
+  }
+
+  const highlightedRegion = alert.highlightedRegion;
+  if (highlightedRegion.endLine === highlightedRegion.startLine &&
+    highlightedRegion.endColumn < highlightedRegion.startColumn) {
+    errors.push('The highlighted region end column is greater than the start column');
+  }
```
aeisenberg, line 200: There are other kinds of errors to handle, too. Eg- negative values, or values larger than the code itself. I don't think you need to be adding error messages for all of these. You just need to make sure that the UI can do something reasonable when these errors occur. 

It looks like here you are adding error messages, and still adding the region to the list of alerts. I think that's fine as long as there's somewhere else that can clean up invalid regions, or the web view is robust enough to do something reasonable for them. Does the web view already handle this?

charisk: TBH I wanted to focus on happy path first so that we can start seeing the view populated. I just added the errors because the compiler was shouting at me 😂 the web view doesn't handle errors gracefully atm, but remember that all this is behind a "feature flag".

WRT validation, here are some thoughts I've had:
- How far do we want to go, considering we own the whole toolchain?
- Adding custom validation code like I've started doing here seems a pain. I'm wondering if we can replace that with something automated e.g. using a JSON schema validator like [ajv](https://ajv.js.org/guide/typescript.html).

charisk:
@aeisenberg are you happy enough with this change to be merged?

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
aeisenberg: We don't really own the whole toolchain since we are running queries created by others and it is the query writers who inject source locations. Most of the time (but not always), the source locations come from codeql core libraries and most of the time (but not always) these libraries produce valid source locations. I bring this up because we've hit errors before for local queries. Bad source locations usually means a bug in the query or one of the query libraries. However, the user running the query may not have any control over that, so we should be handling this gracefully.

Also, ajv won't be much help here since invalid source locations are not checkable in the json schema for sarif. The way we've handled this for local queries is that we do some reasonable things to normalize source locations (eg- negative numbers are changed to 0, values after the end of the line are changed to the end of the line, and if an end location comes before the start location we swap them). I don't think we are even logging these invalid source locations (but maybe we should).

I agree that we don't need to handle it now, but it is something we should put in for a general release. So, let's get this in and handle invalid cases later. I just wanted to make sure you are aware of the difficulties so we don't repeat the same problems we had with local queries. :)

aeisenberg (approved):
As mentioned, let's get this in and we can iterate on making this more robust later.

Issue #1174, opened 2022-02-28
Error when displaying log from a previous VS Code run after restarting


edoardopirovano:
**Describe the bug**
After a restart, items in the Query History view preserved from previous runs still have a _Show Query Log_ button when right-clicked. However, the query log is gone as it is stored in temporary storage that is cleared on restart, so clicking the button gives an error.

**Version**
Latest `main` of both CodeQL and VS Code.

**To reproduce**
1. Run a query.
2. Restart VS Code.
3. Right click the Query History item for the run and select _Show Query Log_.

**Expected behavior**

Either we don't show _Show Query Log_ any more, or we place the log somewhere where it is preserved across runs.

aeisenberg:
Thanks for reporting. I think the best thing to do would be to ensure the log file is saved to the query history location in the global storage.  The problem manifests here: https://github.com/github/vscode-codeql/blob/cdb95065839d370012e26c6196dc59579f795cb5/extensions/ql-vscode/src/queryserver-client.ts#L211-L216

Unfortunately, the logic is a bit twisted and we don't have the query history location available at the moment when we start writing the logs. It will require a little bit of reworking to get there.

edoardopirovano:
Indeed. I noticed this because I replicated the existing logic for some new options to display per-query structured logs and encountered this there too.

I started looking into fixing this but I was actually going the opposite way and trying to figure out when a query comes from a previous run to hide the option. That isn't too easy to do either, so your solution is probably better.

aeisenberg:
Let me take this over since this is related to work I've just done. I'll focus on the existing log and I'll make sure that whatever I do will be easily extendable to the structured log.

aeisenberg:
Going through this now. The good news is that a lot of this logic can be significantly simplified since log files no longer need to be deleted when the workspace closes because they are deleted when the rest of the results are deleted.

The bad (or complicating) news is that there is a preference `codeQL.runningQueries.customLogDirectory` that no longer makes sense to support since its reason for being around was to allow users to retain logs across restarts. We _could_ retain this behaviour for backwards compatibility. Doing so will add much of the complexity back. _Or_ we could remove the handling of the setting and if the extension detects the setting, we can issue a pop-up warning describing the new behaviour.

I won't make any changes until we can discuss this some more.

edoardopirovano:
>_Or_ we could remove the handling of the setting and if the extension detects the setting, we can issue a pop-up warning describing the new behaviour.

[...]

PR #1179, opened 2022-03-01
Add new command to open the query history directory

aeisenberg:
Allows users to explore the contents and all artifacts of the query
that was just run.

Fixes https://github.com/github/vscode-codeql/issues/1174#issuecomment-1055885346

cc: @tausbn 

## Checklist


charisk (approved):

tausbn:
Does this work with Codespaces? As far as I can tell, it instructs the host OS to open the folder, and I'm not sure that works properly when the machine that has the directory is hosted in the cloud somewhere.

aeisenberg:
Good question. I have not tried this on codespaces. I'm not even sure what the expected behaviour should be. Probably the best result would be to disable the command if the file system is remote.

aeisenberg:
Added this line and verified on codespaces that the option does not show:

https://github.com/github/vscode-codeql/pull/1179/files#diff-be4c5a31505df7de81fa4900e84e66c04d1a6bca9b3da50f8061d2cb2bfbbde1R707

tausbn:
> Added this line and verified on codespaces that the option does not show: [...]

In that case, I do not see how this PR "Fixes https://github.com/github/vscode-codeql/issues/1174#issuecomment-1055885346", as that comment pertains directly to the use of Codespaces.

aeisenberg:
Hmmm...right. At least now, a broken command won't be visible in codespaces. Also, this logs _will_ stick around after restarts. All the data is stored in the workspace storage folder. You need to ensure that the workspace remains the same across restarts. 

This command is quite useful in non-remote situations and is something I wanted to implement anyway.

I will take a look at a way to copy the path to the clipboard later.

aeisenberg:
Query logs will be saved in subdirectories of `/home/codespace/.vscode-remote/data/User/globalStorage/github.vscode-codeql/queries`. One directory for each query in a file called `query.log`.

PR #1186, opened 2022-03-04
Expose per-query structured evaluator logs

edoardopirovano:
Do not merge until the corresponding internal PR to the query server has been merged and the estimated version in this PR has been updated to match when we actually ship that.

This PR adds two new options to the Query History view:

1. _Show Evaluator Log (Raw)_ displays the raw JSON from the structured logs of a given query run.
2. _Show Evaluator Log (Summary)_ creates a human-readable summary of the structured logs and displays that.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/pure/messages.ts
+++ b/extensions/ql-vscode/src/pure/messages.ts
@@ -646,6 +646,35 @@ export interface ClearCacheParams {
    */
   dryRun: boolean;
 }
+
+/**
+ * Parameters to start a new structured log
+ */
+ export interface StartLogParams {
+  /**
+   * The dataset for which we want to start a new structured log
+   */
+  db: Dataset;
+  /**
+   * The path where we want to place the new structured log
+   */
+  logPath: string;
+}
+
+/**
+ * Parameters to terminate a structured log
+ */
+ export interface EndLogParams {
+  /**
+   * The dataset for which we want to terminated the log
+   */
+  db: Dataset;
+  /**
+   * The path of the log to terminate, will be a no-op if we aren't logging here
+   */
+  logPath: string;
+}
+
 /**
  * Parameters for trimming the cache of a dataset
  */
@@ -680,6 +709,26 @@ export interface ClearCacheResult {
    * deleted.
    */
   deletionMessage: string;
+}
+
+/**
+ * The result of starting a new structured log.
+ */
+export interface StartLogResult {
+  /**
+   * A user friendly message saying what happened.
+   */
+  outcomeMessage: string;
+}
+
+/**
+ * The result of terminating a structured.
+ */
+export interface EndLogResult {
+  /**
+   * A user friendly message saying what happened.
+   */
+  outcomeMessage: string;
 }
 
 /**
@@ -1017,6 +1066,16 @@ export const compileUpgrade = new rpc.RequestType<WithProgressId<CompileUpgradeP
  * Compile an upgrade script to upgrade a dataset.
  */
 export const compileUpgradeSequence = new rpc.RequestType<WithProgressId<CompileUpgradeSequenceParams>, CompileUpgradeSequenceResult, void, void>('compilation/compileUpgradeSequence');
+
+/**
+ * Start a new structured log in the evaluator, terminating the previous one if it exists
+ */
+ export const startLog = new rpc.RequestType<WithProgressId<StartLogParams>, StartLogResult, void, void>('evaluation/startLog');
+
+/**
+ * Terminate a structured log in the evaluator. Is a no-op if we aren't logging to the given location
+ */
+ export const endLog = new rpc.RequestType<WithProgressId<EndLogParams>, EndLogResult, void, void>('evaluation/endLog');
 
 /**
  * Clear the cache of a dataset
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -767,13 +767,15 @@ export class QueryHistoryManager extends DisposableObject {
     singleItem: QueryHistoryInfo,
     multiSelect: QueryHistoryInfo[]
   ) {
-    // Local queries only
-    if (!this.assertSingleQuery(multiSelect) || singleItem?.t !== 'local') {
+    const { finalSingleItem, finalMultiSelect } = this.determineSelection(singleItem, multiSelect);
+
+    // Only applicable to an individual local query
+    if (!this.assertSingleQuery(finalMultiSelect) || !finalSingleItem || finalSingleItem.t !== 'local') {
       return;
     }
 
-    if (singleItem.evalLogLocation) {
-      await this.tryOpenExternalFile(singleItem.evalLogLocation);
+    if (finalSingleItem.evalLogLocation) {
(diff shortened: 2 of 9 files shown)


extensions/ql-vscode/src/pure/messages.ts
```diff
@@ -646,6 +646,35 @@ export interface ClearCacheParams {
...
 }
+
+/**
+ * Parameters to start a new structured log
+ */
+ export interface StartLogParams {
+  /**
+   * The dataset for which we want to start a new structured log
+   */
+  db: Dataset;
```
aeisenberg, line 657: So, these logs are specific to a dataset, not a query?

extensions/ql-vscode/src/query-history.ts
```diff
@@ -736,6 +750,46 @@ export class QueryHistoryManager extends DisposableObject {
...
+  private warnNoEvalLog() {
+    void showAndLogWarningMessage('No evaluator log is available for this run. Perhaps it failed before evaluation, or you are running with a version of CodeQL before ' + CliVersionConstraint.CLI_VERSION_WITH_PER_QUERY_EVAL_LOG + '?');
+  }
+
+  async handleShowEvalLog(
+    singleItem: QueryHistoryInfo,
+    multiSelect: QueryHistoryInfo[]
+  ) {
+    // Local queries only
+    if (!this.assertSingleQuery(multiSelect) || singleItem?.t !== 'local') {
```
aeisenberg, line 763: Best if you call `const { finalSingleItem, finalMultiSelect } = this.determineSelection(singleItem, multiSelect);` before this so that you handle some edge cases where you right-click on an item that is not selected.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -736,6 +750,46 @@ export class QueryHistoryManager extends DisposableObject {
...
+      this.warnNoEvalLog();
+    }
+  }
+
+  async handleShowEvalLogSummary(
+    singleItem: QueryHistoryInfo,
+    multiSelect: QueryHistoryInfo[]
+  ) {
+    // Local queries only
+    if (!this.assertSingleQuery(multiSelect) || singleItem?.t !== 'local') {
```
aeisenberg, line 779: See previous comment.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -736,6 +750,46 @@ export class QueryHistoryManager extends DisposableObject {
...
+      return;
+    }
+
+    if (singleItem.evalLogLocation) {
+      const summaryLocation = singleItem.evalLogLocation + '.summary';
```
aeisenberg, line 784: I'd prefer to have these file names encapsulated in the `LocalQueryInfo` object. Can you create a getter called `evalLogSummaryLocation`? and return the proper path?

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -258,3 +258,7 @@ export class QueryServerClient extends DisposableObject {
 export function findQueryLogFile(resultPath: string): string {
   return path.join(resultPath, 'query.log');
 }
+
+export function findQueryStructLogFile(resultPath: string): string {
+  return path.join(resultPath, 'evaluator-log.jsonl');
```
aeisenberg, line 263: Just making sure this isn't a typo and it should be `evaluator-log.json`.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -657,6 +675,7 @@ export async function compileAndRunQueryAgainstDatabase(
   progress: ProgressCallback,
   token: CancellationToken,
   templates?: messages.TemplateDefinitions,
+  queryInfo?: LocalQueryInfo,
```
aeisenberg, line 678: So, this value will only be available for queries directly initiated by the user? Contextual queries won't have the `queryInfo`? That's fine, but I just want to be sure I understand. Could you add a comment here to explain that?

extensions/ql-vscode/src/pure/messages.ts
edoardopirovano: In a sense, yes. We rotate the logs when new queries are started, but if there are multiple concurrent queries running then events from the evaluator will end up in the latest query to be started.

extensions/ql-vscode/src/query-history.ts
edoardopirovano: Thanks, done.

extensions/ql-vscode/src/query-history.ts
edoardopirovano: Also done.

extensions/ql-vscode/src/run-queries.ts
edoardopirovano: Yes, that's exactly right. There's a few places where queries aren't initiated by the user (as you say, contextual queries, and also test code). In these cases we don't have a sensible place to put the structured logs, nor any reason to store them, so we don't create a log for these cases.

edoardopirovano:
Thanks for the review @aeisenberg! I've believe I have addressed all your comments either with responses above or in the new commit 🙂

aeisenberg (approved):
Looks good.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -736,6 +750,49 @@ export class QueryHistoryManager extends DisposableObject {
...
+
+  async handleShowEvalLog(
+    singleItem: QueryHistoryInfo,
+    multiSelect: QueryHistoryInfo[]
+  ) {
+    const { finalSingleItem, finalMultiSelect } = this.determineSelection(singleItem, multiSelect);
+
+    // Only applicable to an individual local query
+    if (!this.assertSingleQuery(finalMultiSelect) || !finalSingleItem || finalSingleItem.t !== 'local') {
+      return;
```
angelapwen, line 766: Should this situation also result in some kind of error message?

extensions/ql-vscode/src/query-history.ts
```diff
@@ -736,6 +750,49 @@ export class QueryHistoryManager extends DisposableObject {
...
+
+  async handleShowEvalLogSummary(
+    singleItem: QueryHistoryInfo,
+    multiSelect: QueryHistoryInfo[]
+  ) {
+    const { finalSingleItem, finalMultiSelect } = this.determineSelection(singleItem, multiSelect);
+
+    // Only applicable to an individual local query
+    if (!this.assertSingleQuery(finalMultiSelect) || !finalSingleItem || finalSingleItem.t !== 'local') {
+      return;
```
angelapwen, line 784: Same comment as above.

angelapwen (commented):
✔️ just a few comments since we are waiting for the per-query structured logging PR to merge and be released anyway.

extensions/ql-vscode/src/query-history.ts
edoardopirovano: Hmm, good question. I followed what @aeisenberg did above for other similar drop downs. Any particular reason you went with just a `return` rather than a more explicit error?

edoardopirovano:
Rebased onto the latest `main` to check everything is still working since this PR has been open for a while and we're now at a point where we can merge it.

PR #1187, opened 2022-03-04
Add support for showing code flows

charisk:
Adds the ability to read code flow information from SARIF files and render those in the Web View.

A few notes:
- I've used an RC version of the Primer React library to take advantage of some of the new menu/overlay stuff. I don't intend to merge the PR until the final version is released (see https://github.com/primer/react/discussions/1918).
- As discussed previously, we need a better solution for the validation aspect of SARIF data.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/sarif-processing.ts
+++ b/extensions/ql-vscode/src/remote-queries/sarif-processing.ts
@@ -1,6 +1,6 @@
 import * as sarif from 'sarif';
 
-import { AnalysisAlert, ResultSeverity } from './shared/analysis-result';
+import { AnalysisAlert, CodeFlow, CodeSnippet, HighlightedRegion, ResultSeverity, ThreadFlow } from './shared/analysis-result';
 
 const defaultSeverity = 'Warning';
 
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -186,18 +147,149 @@ export function tryGetRule(
...
+  const regionErrors = validateRegion(region);
+  errors.push(...regionErrors);
+
+  const filePath = location.physicalLocation?.artifactLocation?.uri;
+  if (!filePath) {
+    errors.push('No file path found in the SARIF result location');
+  }
+
+  if (errors.length > 0) {
+    return { processedLocation: undefined, errors };
```
aeisenberg, line 238: In this case, I think it would still be nice to return a processed location, but with some fixups. It would be unfortunate if query results couldn't be displayed because the query muddled up some source locations. I think the important thing is to avoid failing when trying to display bad locations and make some sort of best effort to fix up the locations (even if that means reducing the entire source location to a single character. It would still be useful to warn that the source locs are bad, but just do our best to show something.

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -186,18 +147,149 @@ export function tryGetRule(
...
+          if (locationErrors.length > 0) {
+            errors.push(...locationErrors);
+            continue;
+          }
+
+          threadFlows.push({
+            filePath: processedLocation!.filePath,
+            codeSnippet: processedLocation!.codeSnippet,
+            highlightedRegion: processedLocation!.highlightedRegion,
+            message: processedLocation!.message
```
aeisenberg, line 285: If my comment above it taken to account, `processedLocation` will never be undefined and the `!` can go away.

extensions/ql-vscode/src/remote-queries/view/CodePaths.tsx
```diff
@@ -0,0 +1,180 @@
...
+  codeFlows: CodeFlow[],
+  ruleDescription: string,
+  message: string,
+  severity: ResultSeverity
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const [selectedCodeFlow, setSelectedCodeFlow] = useState(codeFlows[0]);
+
+  const anchorRef = useRef<HTMLDivElement>(null);
+  const linkRef = useRef<HTMLAnchorElement>(null);
```
aeisenberg, line 132: Isn't `useRef` no arg?

```suggestion
  const anchorRef = useRef<HTMLDivElement>();
  const linkRef = useRef<HTMLAnchorElement>();
```

extensions/ql-vscode/src/remote-queries/view/CodePaths.tsx
```diff
@@ -0,0 +1,180 @@
...
+        onClick={() => setIsOpen(true)}
+        ref={linkRef}
+        sx={{ cursor: 'pointer' }}>
+        Show paths
+      </Link>
+      {isOpen && (
+        <Overlay
+          returnFocusRef={linkRef}
+          onEscape={closeOverlay}
+          onClickOutside={closeOverlay}
```
aeisenberg, line 148: Nice.

aeisenberg (approved):
Looks good. My main comment about trying your best to display malformed source locs can be delayed until later.

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
charisk: Thank you, I really appreciate these comments - I'm gradually understanding the domain more and more! I'll leave as is for now and we can loop back round with all the other validation bits.

extensions/ql-vscode/src/remote-queries/view/CodePaths.tsx
charisk: We use `null` to set an initial value because otherwise it can be `undefined` that's not acceptable for `Box` or `Link`.

charisk:
There is a bit of a delay with releasing a new version of Primer React so I'll go ahead and merge this PR since none of the RC issues seem to be affecting us. I'll come back and upgrade the package versions once a new version is released.

PR #1189, opened 2022-03-07
Add pre-push hook check to block leftover .only()s

charisk:
## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/scripts/forbid-mocha-only
+++ b/extensions/ql-vscode/scripts/forbid-mocha-only
@@ -1,4 +1,4 @@
-if grep -rq --include '*.test.ts' 'it.only\|describe.only' './test'; then
+if grep -rq --include '*.test.ts' 'it.only\|describe.only' './test' './src'; then
     echo 'There is a .only() in the tests. Please remove it.'
     exit 1;
 else
--- a/extensions/ql-vscode/scripts/forbid-mocha-only
+++ b/extensions/ql-vscode/scripts/forbid-mocha-only
@@ -1,0 +1,7 @@
+if grep -rq --include '*.test.ts' 'it.only\|describe.only' ..; then
+    echo 'There is a .only() in the tests. Please remove it.'
+    exit 1;
+else
+    echo 'All good'
+    exit 0;
+fi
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/scripts/forbid-mocha-only
```diff
@@ -0,0 +1,7 @@
+if grep -rq --include '*.test.ts' 'it.only\|describe.only' ..; then
```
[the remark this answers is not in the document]
charisk: I just realised that all the tests are actually in ./test, so I've updated the command to just look in there.

extensions/ql-vscode/scripts/forbid-mocha-only
```diff
@@ -0,0 +1,6 @@
+if grep -rq --include '*.test.ts' 'it.only\|describe.only' './test'; then
```
shati-patel, line 1: > I just realised that all the tests are actually in ./test, so I've updated the command to just look in there.

⚠️ 
The `./test` folder doesn't contain all tests! (I think only the "pure" tests that don't rely on any VS Code APIs.)
The other tests are in `./src/vscode-tests` 🧪

charisk: 🤦‍♀️ thanks, I missed those. Do you happen to know if there is a reason those tests are there?

I've updated to look in ./test and ./src.

shati-patel: > Do you happen to know if there is a reason those tests are there and not in ./test?

I asked the same question when I first started looking at the tests, but I can't quite remember the answer 😳 
The tests in `./src` are integration tests and need access to import APIs and load up the extension, so maybe they need to be next to the actual source code? 🙈

charisk: Ok thanks, was just being nosey :D

shati-patel (approved):

extensions/ql-vscode/scripts/forbid-mocha-only
aeisenberg: @shati-patel is exactly right. There's no _good_ reason for the separation other than that they are run differently. I don't think it's really worth the effort to combine them, but it is confusing.

PR #1195, opened 2022-03-09
Deal with analysis messages that have links to locations

charisk:
Analysis messages can have tokens in them that need to be replaced to link to locations. This PR adds support for rendering these messages properly. Note that the link to the location isn't populated yet - once we have support for that we'll be able to take advantage of it here.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/pure/sarif-utils.ts
+++ b/extensions/ql-vscode/src/pure/sarif-utils.ts
@@ -148,7 +148,7 @@ export function parseSarifRegion(
   // We assume that the SARIF we're given always has startLine
   // This is not mandated by the SARIF spec, but should be true of
   // SARIF output by our own tools.
-  const startLine = region.startLine!;
+  const startLine = region.startLine ?? 1;
 
   // These defaults are from SARIF 2.1.0 spec, section 3.30.2, "Text Regions"
   // https://docs.oasis-open.org/sarif/sarif/v2.1.0/cs01/sarif-v2.1.0-cs01.html#_Ref493492556
@@ -160,7 +160,7 @@ export function parseSarifRegion(
   // length we don't know at this point in the code.
   //
   // It is off by one with respect to the way vscode counts columns in selections.
-  const endColumn = region.endColumn! - 1;
+  const endColumn = (region.endColumn ?? 1) - 1;
 
   return {
     startLine,
--- a/extensions/ql-vscode/src/pure/sarif-utils.ts
+++ b/extensions/ql-vscode/src/pure/sarif-utils.ts
@@ -145,9 +145,8 @@ export function parseSarifRegion(
   startColumn: number,
   endColumn: number
 } {
-  // We assume that the SARIF we're given always has startLine
-  // This is not mandated by the SARIF spec, but should be true of
-  // SARIF output by our own tools.
+  // The SARIF we're given should have a startLine, but we
+  // fall back to 1, just in case something has gone wrong.
   const startLine = region.startLine ?? 1;
 
   // These defaults are from SARIF 2.1.0 spec, section 3.30.2, "Text Regions"
@@ -155,9 +154,10 @@ export function parseSarifRegion(
   const endLine = region.endLine === undefined ? startLine : region.endLine;
   const startColumn = region.startColumn === undefined ? 1 : region.startColumn;
 
-  // We also assume that our tools will always supply `endColumn` field, which is
-  // fortunate, since the SARIF spec says that it defaults to the end of the line, whose
-  // length we don't know at this point in the code.
+  // Our tools should always supply `endColumn` field, which is fortunate, since 
+  // the SARIF spec says that it defaults to the end of the line, whose
+  // length we don't know at this point in the code. We fall back to 1,
+  // just in case something has gone wrong.
   //
   // It is off by one with respect to the way vscode counts columns in selections.
   const endColumn = (region.endColumn ?? 1) - 1;
--- a/extensions/ql-vscode/src/pure/sarif-utils.ts
+++ b/extensions/ql-vscode/src/pure/sarif-utils.ts
@@ -127,33 +127,47 @@ export function parseSarifLocation(
       userVisibleFile
     } as ParsedSarifLocation;
   } else {
-    const region = physicalLocation.region;
-    // We assume that the SARIF we're given always has startLine
-    // This is not mandated by the SARIF spec, but should be true of
-    // SARIF output by our own tools.
-    const startLine = region.startLine!;
-
-    // These defaults are from SARIF 2.1.0 spec, section 3.30.2, "Text Regions"
-    // https://docs.oasis-open.org/sarif/sarif/v2.1.0/cs01/sarif-v2.1.0-cs01.html#_Ref493492556
-    const endLine = region.endLine === undefined ? startLine : region.endLine;
-    const startColumn = region.startColumn === undefined ? 1 : region.startColumn;
-
-    // We also assume that our tools will always supply `endColumn` field, which is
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/remote-queries/shared/analysis-result.ts
```diff
@@ -7,12 +7,12 @@ export interface AnalysisResults {
...
 
 export interface AnalysisAlert {
-  message: string;
+  message: AnalysisMessage;
   shortDescription: string;
   severity: ResultSeverity;
   filePath: string;
   codeSnippet: CodeSnippet;
-  highlightedRegion: HighlightedRegion;
+  highlightedRegion?: HighlightedRegion;
```
charisk, line 15: I've made this optional because it made my code simpler but also because I think it makes sense for this to be optional. We won't always want to highlight a specific region.

aeisenberg (commented):
I think this is generally fine. Based on our conversation this morning, I think this will be changing somewhat anyway. It would be nice to get some more unit tests in for the parsing.

extensions/ql-vscode/src/remote-queries/shared/analysis-result.ts
```diff
@@ -37,7 +37,33 @@ export interface ThreadFlow {
   filePath: string;
   codeSnippet: CodeSnippet;
   highlightedRegion: HighlightedRegion;
+  message?: AnalysisMessage;
+}
+
+export interface AnalysisMessage {
+  tokens: AnalysisMessageToken[]
```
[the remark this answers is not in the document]
charisk: Correct!

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -272,7 +286,10 @@ function extractCodeFlows(
 
       for (const threadFlow of codeFlow.threadFlows) {
         for (const location of threadFlow.locations) {
-          const { processedLocation, errors: locationErrors } = extractLocation(location);
+          if (!location?.location) {
+            errors.push('No location found in a SARIF thread flow');
+          }
+          const { processedLocation, errors: locationErrors } = extractLocation(location!.location!);
```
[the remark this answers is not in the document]
charisk: Removed

charisk:
> I think this is generally fine. Based on our conversation this morning, I think this will be changing somewhat anyway. It would be nice to get some more unit tests in for the parsing.

Yep, I'll make the changes we discussed in a separate PR and will make sure functionality is covered by unit tests.

extensions/ql-vscode/src/pure/sarif-utils.ts
```diff
@@ -127,35 +127,49 @@ export function parseSarifLocation(
...
+): {
+  startLine: number,
+  endLine: number,
+  startColumn: number,
+  endColumn: number
+} {
+  // We assume that the SARIF we're given always has startLine
+  // This is not mandated by the SARIF spec, but should be true of
+  // SARIF output by our own tools.
+  const startLine = region.startLine!;
```
aeisenberg, line 151: I don't think we can assume this if someone writes a bad query.

```suggestion
  // This is not mandated by the SARIF spec, but should be true of
  // SARIF output by our own tools.
  const startLine = region.startLine ?? 1;
```

extensions/ql-vscode/src/pure/sarif-utils.ts
```diff
@@ -127,35 +127,49 @@ export function parseSarifLocation(
...
+  // https://docs.oasis-open.org/sarif/sarif/v2.1.0/cs01/sarif-v2.1.0-cs01.html#_Ref493492556
+  const endLine = region.endLine === undefined ? startLine : region.endLine;
+  const startColumn = region.startColumn === undefined ? 1 : region.startColumn;
+
+  // We also assume that our tools will always supply `endColumn` field, which is
+  // fortunate, since the SARIF spec says that it defaults to the end of the line, whose
+  // length we don't know at this point in the code.
+  //
+  // It is off by one with respect to the way vscode counts columns in selections.
+  const endColumn = region.endColumn! - 1;
```
aeisenberg, line 163: Same as above.

```suggestion
  const endColumn = (region.endColumn ?? 1) - 1;
```

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -19,77 +20,63 @@ export function extractAnalysisAlerts(
...
-      const shortDescription = getShortDescription(rule, message!);
-
-      for (const location of result.locations) {
-        const { processedLocation, errors: locationErrors } = extractLocation(location);
-
-        if (locationErrors.length > 0) {
-          errors.push(...locationErrors);
+  try {
+    for (const run of sarifLog.runs!) {
+      for (const result of run.results!) {
```
aeisenberg, line 28: To be safe, maybe something like this:

```suggestion
    for (const run of sarifLog.runs ?? []) {
      for (const result of run.results ?? []) {
```

It's _probably_ fine, but not sure what would happen if a run legitimately has no results.

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -19,77 +20,63 @@ export function extractAnalysisAlerts(
...
+): AnalysisAlert[] {
+  const alerts: AnalysisAlert[] = [];
+
+  const message = getMessage(result);
+  const rule = tryGetRule(run, result);
+  const severity = tryGetSeverity(run, result, rule) || defaultSeverity;
+  const codeFlows = getCodeFlows(result);
+  const shortDescription = getShortDescription(rule, message!);
+
+  for (const location of result.locations!) {
```
aeisenberg, line 56: If a result has no location associated with it (this can happen with a query that doesn't include locations), does this array exist? I'm not sure. Maybe safer to just do:

```suggestion
  for (const location of result.locations ?? []) {
```

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -19,77 +20,63 @@ export function extractAnalysisAlerts(
...
+    const physicalLocation = location.physicalLocation!;
+    const filePath = physicalLocation.artifactLocation!.uri!;
+    const codeSnippet = getCodeSnippet(physicalLocation.contextRegion!);
+    const highlightedRegion = physicalLocation.region
+      ? getHighlightedRegion(physicalLocation.region!)
+      : undefined;
```
aeisenberg, line 62: It's probably safe to leave all of this as is. If a result has a location, then it should be constructed properly in the sarif.

aeisenberg (commented):
It's really hard to know right now which properties will never be undefined, and which ones might. I'm injecting some of my knowledge of how the CLI and sarif interpretation works, but I'm not 100% certain.

If any of these properties are undefined, then it would be for poorly defined queries (ie- queries that don't use source locations at all, or construct invalid locations).

I would lean towards trying not to throw an error if a property is null or undefined. I think just replacing most type assertions with nullish coalescing would be sufficient (especially on the arrays).

extensions/ql-vscode/src/pure/sarif-utils.ts
charisk: I was a bit reluctant to change this as it's touching local queries code, but I've gone ahead and made the change since you gave it your blessing :P I've also updated the comments.

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
charisk: Based on what I've seen in the code base etc. I don't think it's possible, but I guess it doesn't hurt adding the fallbacks.

aeisenberg (approved):

PR #1205, opened 2022-03-14
Get database sha from result index

shati-patel:
We want to be able to link from remote query results to the exact file that's being analysed (i.e. the correct GitHub URL for the repository + the exact commit SHA that the database was built from).

https://github.com/dsp-testing/qc-run2/pull/639 adds this to the result index, and this PR reads the SHA and adds it to the `AnalysisSummary`. (If no SHA is found, we use `HEAD` as a fall back) 

We don't use the SHA anywhere yet, but we'll need it in the webview soon. (I "tested" the change by setting some breakpoints and checking the variables, but there's no visible change 🤷🏽 ) 

## Checklist

N/A - internal only 🤫 

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #1209, opened 2022-03-15
Create remote file links to GitHub URL

shati-patel:
Follow-up to #1205.

Creates a clickable link in the results view that links to the file location on GitHub. e.g.
image
links to 
https://github.com/meteor/meteor/blob/HEAD/npm-packages/meteor-installer/install.js#L253-L257 (using the exact SHA instead of HEAD, when available).

## Checklist

N/A - still internal only 🗳️ 

### DIFF
--- a/extensions/ql-vscode/src/pure/bqrs-utils.ts
+++ b/extensions/ql-vscode/src/pure/bqrs-utils.ts
@@ -101,7 +101,12 @@ export function tryGetRemoteLocation(
   if (loc === undefined) {
     return undefined;
   } else if (isWholeFileLoc(loc) || isLineColumnLoc(loc)) {
-    return fileLinkPrefix + loc.uri.replace(FILE_LOCATION_REGEX, '');
+    // Trim the location
+    // file:/home/runner/work/turboscan/turboscan/cmd/enumgenerator/main.go
+    const parts = loc.uri.split('/');
+    const trimmedLocation = parts.slice(6, parts.length).join('/');
+    return `${fileLinkPrefix}\\${trimmedLocation}`;
+
   } else if (isStringLoc(loc)) {
     const location = tryGetLocationFromString(loc);
     return location?.uri;
--- a/extensions/ql-vscode/src/pure/bqrs-utils.ts
+++ b/extensions/ql-vscode/src/pure/bqrs-utils.ts
@@ -93,3 +93,19 @@ export function isWholeFileLoc(loc: UrlValue): loc is WholeFileLocation {
 export function isStringLoc(loc: UrlValue): loc is string {
   return typeof loc === 'string';
 }
+
+export function tryGetRemoteLocation(
+  loc: UrlValue | undefined,
+  fileLinkPrefix: string
+): string | undefined {
+  if (loc === undefined) {
+    return undefined;
+  } else if (isWholeFileLoc(loc) || isLineColumnLoc(loc)) {
+    return fileLinkPrefix + loc.uri.replace(FILE_LOCATION_REGEX, '');
+  } else if (isStringLoc(loc)) {
+    const location = tryGetLocationFromString(loc);
+    return location?.uri;
+  } else {
+    return undefined;
+  }
+}
--- a/extensions/ql-vscode/src/pure/bqrs-utils.ts
+++ b/extensions/ql-vscode/src/pure/bqrs-utils.ts
@@ -4,6 +4,7 @@ import {
   LineColumnLocation,
   WholeFileLocation
 } from './bqrs-cli-types';
+import { createRemoteFileRef } from './location-link-utils';
 
 /**
  * The CodeQL filesystem libraries use this pattern in `getURL()` predicates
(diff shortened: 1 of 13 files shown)


shati-patel:
Old query history items (i.e. whose result index didn't contain `sha`) still returned a link like https://github.com/meteor/meteor/blob/undefined/npm-packages/meteor-installer/install.js#L253-L257. Latest commit (https://github.com/github/vscode-codeql/pull/1209/commits/dd6385bbedf543ec36cb755d9c08b6fe5b45c504) just adds a fallback to `HEAD` here as well.

extensions/ql-vscode/src/pure/bqrs-utils.ts
```diff
@@ -93,3 +94,25 @@ export function isWholeFileLoc(loc: UrlValue): loc is WholeFileLocation {
...
+  const resolvableLocation = tryGetResolvableLocation(loc);
+  if (!resolvableLocation) {
+    return undefined;
+  }
+
+  // Remote locations have the following format:
+  // file:/home/runner/work/<repo>/<repo/relative/path/to/file
+  // So we need to drop the first 6 parts of the path.
+  const locationParts = resolvableLocation.uri.split('/');
+  const trimmedLocation = locationParts.slice(6, locationParts.length).join('/');
```
adityasharad, line 111: This is a clever trick, and is good enough as a starting point for most runs right now. Unfortunately in the long run, I suspect there is a risk of this breaking quietly because it depends on the exact filesystem layout during the workflow run that originally created the database. Here are some examples I could think of where the assumptions will not hold:
- the run might be on Windows, so you have something like `C:\runner\work\<nwo>` as the prefix (likely)
- the run might have checked out the repo to a different folder instead of the root of the workspace, in which case the prefix is `/home/runner/work/<nwo>/differentFolder` (somewhat likely)
- the run might be on a self-hosted runner that doesn't follow the `/home/runner/work/<nwo>` convention (unlikely but possible)

Assuming we have to handle these cases, what should we do? Fortunately I think there is some logic from the local queries part of the extension that we can learn from, and a useful field populated by the CLI. CodeQL databases have a `sourceLocationPrefix` field in `codeql-database.yml` that describes the path prefix to the source code (`/home/runner/work/<nwo>/` in your example). Have you got that data already from the remote run?

Once you have that, you can remove the `sourceLocationPrefix` from the beginning of the `resolvableLocation.uri`, and what remains will be the relative path you're looking for.

(The key difference between local and remote queries here is that local queries have to resolve the location against the source archive on the local filesystem, while remote queries have to resolve the location against GitHub repo URLs. There is the possibility of sharing code between them and making this distinction clear, but it's overkill for this PR.)

extensions/ql-vscode/src/remote-queries/analyses-results-manager.ts
```diff
@@ -166,4 +168,8 @@ export class AnalysesResultsManager {
   private isAnalysisInMemory(analysis: AnalysisSummary): boolean {
     return this.internalGetAnalysesResults(analysis.downloadLink.queryId).some(x => x.nwo === analysis.nwo);
   }
+
+  private createFileLinkPrefix(nwo: string, sha: string): string {
```
adityasharad, line 172: Minor: Consider giving this a name that indicates it's a GitHub.com link prefix. In future if we add GHES support that will help us find this code and make it more flexible.

adityasharad (commented):
This looks neat, and the wiring of the new links generally looks good. I have one suggestion on how to resolve non-SARIF locations, which can be fiddly -- let me know if you want to talk through the details separately.

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -369,7 +370,7 @@ export function RemoteQueries(): JSX.Element {
     return <div>Waiting for results to load.</div>;
   }
 
-  const showAnalysesResults = false;
+  const showAnalysesResults = true;
```
charisk, line 373: Optional: Change this back to false so I don't get merge conflicts 😂

charisk (approved):
Woohoo!

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -273,7 +273,7 @@ const AnalysesResultsDescription = ({
   const showMaxResultsMessage = analysesResults.some(a => a.rawResults?.capped);
   const maxRawResultsMessage = <>
     <VerticalSpace size={1} />
-    Some repositories have more than {MAX_RAW_RESULTS} results. We will only show you up to 
+    Some repositories have more than {MAX_RAW_RESULTS} results. We will only show you up to
```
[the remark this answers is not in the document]
shati-patel: Thanks! Will this happen every time we edit this file? 😬 Do we have a more permanent way of fixing it?

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
shati-patel: very sensible 😁

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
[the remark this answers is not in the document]
charisk: Yes and yes! Was about to ping you to discuss :)

extensions/ql-vscode/src/pure/bqrs-utils.ts
shati-patel: Thank you for the detailed breakdown! 🙇🏽‍♀️ 

You're definitely right that this won't be reliable in the long run 🙈 Since we have enough control over the actions runs for the upcoming private beta, this should work for now.

I'll open an issue to fix this properly though 😄 Your suggestion of using the `sourceLocationPrefix` from the action sounds good! We don't currently have access to that in the extension, but we can download it as part of the metadata (along with things like `nwo`, `sha` etc)

PR #1217, opened 2022-03-17
Minor webview fixes

shati-patel:
Puts the annotation for an alert below the highlighted line instead of above. (PS: I'm not sure if this is the right fix, but it _looks_ good at least 🤷🏽)
image

(Also converts a space to `&nbsp;` to stop an overenthusiastic linter from stripping it off the end of the line.) 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/view/FileCodeSnippet.tsx
+++ b/extensions/ql-vscode/src/remote-queries/view/FileCodeSnippet.tsx
@@ -206,13 +206,6 @@ const FileCodeSnippet = ({
       <CodeContainer>
         {code.map((line, index) => (
           <div key={index}>
-            {message && severity && <Message
-              message={message}
-              currentLineNumber={startingLine + index - 1}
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/remote-queries/view/FileCodeSnippet.tsx
```diff
@@ -208,7 +208,7 @@ const FileCodeSnippet = ({
           <div key={index}>
             {message && severity && <Message
               message={message}
-              currentLineNumber={startingLine + index}
+              currentLineNumber={startingLine + index - 1}
```
charisk, line 211: As you say, this works, however I think it'll be more intuitive if we left this line of code as is and just moved the `Message` component below the `Box` component instead.

shati-patel: That's far too sensible 😂 

(done)

PR #1225, opened 2022-03-17
Log most expensive predicates and timings to query server window

angelapwen:
I have made a few additions ✨ to a work in progress PR to add per-query structured evaluator logs that cannot be merged yet. Changes include:
- generating evaluator summary logs as soon as the raw logs are written rather than only when the user clicks "Show Evaluator Log (Summary)" so that
- the most expensive predicates and timing summary from the end of the summary file can be printed to the query server output window where our users expect it
- renames all instances where we call the logs "structured" to "eval" or "evaluator" log to match naming across the repository
- deprecates structured logging for a single query server instance as this is no longer necessary (rolls back #1151)

query-server-view

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -675,7 +675,7 @@ export async function compileAndRunQueryAgainstDatabase(
   progress: ProgressCallback,
   token: CancellationToken,
   templates?: messages.TemplateDefinitions,
-  queryInfo?: LocalQueryInfo,
+  queryInfo?: LocalQueryInfo, // May be omitted for queries not initiated by the user. If omitted we won't create a structured log for the query.
 ): Promise<QueryWithResults> {
   if (!dbItem.contents || !dbItem.contents.dbSchemeUri) {
     throw new Error(`Database ${dbItem.databaseUri} does not have a CodeQL database scheme.`);
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -94,12 +94,12 @@ export class QueryEvaluationInfo {
     return qsClient.findQueryLogFile(this.querySaveDir);
   }
 
-  get structLogPath() {
-    return qsClient.findQueryStructLogFile(this.querySaveDir);
+  get evalLogPath() {
+    return qsClient.findQueryEvalLogFile(this.querySaveDir);
   }
 
-  get structLogSummaryPath() {
-    return qsClient.findQueryStructLogSummaryFile(this.querySaveDir);
+  get evalLogSummaryPath() {
+    return qsClient.findQueryEvalLogSummaryFile(this.querySaveDir);
   }
 
   get resultsPaths() {
@@ -167,7 +167,7 @@ export class QueryEvaluationInfo {
     if (queryInfo && await qs.cliServer.cliConstraints.supportsPerQueryEvalLog()) {
       await qs.sendRequest(messages.startLog, {
         db: dataset,
-        logPath: this.structLogPath,
+        logPath: this.evalLogPath,
       });
     }
     const params: messages.EvaluateQueriesParams = {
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/run-queries.ts
```diff
@@ -199,6 +209,30 @@ export class QueryEvaluationInfo {
...
+        if (err) {
+          throw new Error(`Could not read structured evaluator log summary file at ${this.evalLogSummaryPath}.`);
+        }
+        const summaryText = buffer.toString();
+        const mostExpensivePredicatesIndex = summaryText.indexOf('Most expensive predicates');
+
+        if (mostExpensivePredicatesIndex != -1) {
+          void qs.logger.log(' --- Evaluator Log Summary --- ', { additionalLogLocation: this.logPath });
+          void qs.logger.log(summaryText.substring(mostExpensivePredicatesIndex), { additionalLogLocation: this.logPath });
+        }
```
angelapwen, line 229: Because there could be a variable number of "most expensive predicates" per stage, I couldn't think of a better way to parse out only this part of the logs for the output. cc @adityasharad because we discussed trying to avoid text parsing previously.

edoardopirovano: Hmm, I am indeed a little uneasy about parsing the logs to do this. Perhaps a better option to achieve this without relying on parsing the logs could be to extend `codeql generate log-summary` with a flag that tells it to produce just this end-of-query summary (i.e. most expensive predicates and total evaluation times) without all the other information. Then, we could invoke that here to get the information without having to rely on log parsing. That gives us some robustness to changes to the log, and provides us with an isolated place we can potentially add more end-of-query information in future. It also avoids having to deal with the issue I mentioned above about consuming the file in a streaming manner.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -199,6 +209,30 @@ export class QueryEvaluationInfo {
...
   }
 
+  /**
+   * Prints the end of the structured log summary file (starting from most expensive predicates)
+   * to the query server Output Window.
+   * @param qs The current QueryServerClient instance for this query.
+   */
+  printEvalLogSummary(qs: qsClient.QueryServerClient) {
+    if (this.hasEvalLogSummary()) {
+      fs.readFile(this.evalLogSummaryPath, (err, buffer) => {
```
edoardopirovano, line 219: Note that `fs.readFile` will read the whole file into memory which for very long evaluation runs could potentially cause problems, especially for users with not much RAM on their machine. Anecdotally, I've seen these logs get to a few hundred megabytes before, and we're storing them twice here at a point because they'll be in both `buffer` and `summaryText` after we've done `const summaryText = buffer.toString()`. Reading using [streams](https://nodejs.dev/learn/nodejs-streams) would be a better option to keep our memory consumption down, especially since we're throwing most of the file out anyway.

extensions/ql-vscode/src/run-queries.ts
angelapwen: That makes sense to me. I am thinking of just extending the `--format` flag with a new option called `final` (or something along those lines) that only outputs the information we want in this window. We can workshop the option name 😄

adityasharad: What does `codeql generate log-summary` currently print to stdout/stderr? Could we include the timing summary in that output, and just have the extension display that output in the Output window?

angelapwen: I believe `generate log-summary` command does not print to stdout/stderr unless the output file for the summary is unspecified (in which case the entire summary text file is printed to stdout). It's possible that other values are printing to stdout/stderr that I haven't noticed in the codebase though 🤔 will defer to @edoardopirovano 

We were thinking of using a hidden option for `codeql generate log-summary` that only works with the `--format=text` option. Something like `codeql generate log-summary --format=text --end-of-query=${FILE}` It would write the end of query summaries to another file (which I guess is similar to writing it to stdout/stderr) and the Extension would consume that other file for the output window.

adityasharad: Sounds reasonable; either of those will achieve what you want without requiring two command invocations.

angelapwen: Ah yes, we came to that conclusion after I had implemented the `--format=final` option 🐢

edoardopirovano: I did actually think of suggesting stdout for the end summary but I strayed away from it because it's already possible for the whole log to go to stdout so it seemed a bit strange to have that part of it always go there. Maybe stderr could be an option, though - I don't think we log anything there.

I think a separate file works too, though, and it has the advantage of leaving us stderr for something else in the future (maybe progress updates?).

PR #952, opened 2021-09-09
Remove `token` for running a remote query

shati-patel:
[internal feature]

The API for remote querying has changed (see linked issue), so we no longer need to provide a token in the Octokit request 🙂 This is a partial revert of [this commit](https://github.com/github/vscode-codeql/pull/882/commits/3c678ca39f90a429999e4a197c2ccc64f2718c00#diff-39def7429bdcd0a61f1dab2ef3277a28e1f8ff7186cced600a49b10c8b9d8276).

## Checklist

N/A, no user-facing changes

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] `@github/docs-content-codeql` has been cc'd in all issues for UI or other user-facing changes made by this pull request.

PR #1229, opened 2022-03-22
Download databases from GitHub

### LINKED ISSUE #1123: Support importing databases directly from github
For github repositories that save a copy of their databases, the extension should have a command that downloads and installs the database, similar to how we do it for LGTM databases.

There is already an API endpoint available:

```
gh api /repos/{owner}/{repo}/code-scanning/codeql/databases/<lang>
```

Is there a way to determine which database languages are available for any given repo before downloading?

shati-patel:
🍐 _with @robertbrignull_

Will fix #1123.

Adds a new command to download a database directly from GitHub. This borrows heavily from existing functionality like the "Download database from LGTM" command 😅 

download-db

The main difference is that we need to pass in credentials to authenticate to the GitHub API. We could use octokit directly for downloading the DB, but then we wouldn't be able to re-use all the nice functionality from [`databaseArchiveFetcher`](https://github.com/github/vscode-codeql/blob/3d21b203be4a3c863ffd8fdfd2f82c6d1f314781/extensions/ql-vscode/src/databaseFetcher.ts#L164-L225). So instead we get the octokit token (as previously done in https://github.com/github/vscode-codeql/pull/952) and pass in some custom API headers.

Still TODO:
- [x] Get a GitHub logo for the database UI (any ideas where I can find this in light/dark versions?)
   - See https://github.com/github/vscode-codeql/pull/1229#issuecomment-1075492445
- [x] Add some tests

(probably in a follow-up PR)
- [ ] Give the imported database a more helpful name
- [ ] Allow GitHub repo URLs as well as NWOs in the input box (like we do for LGTM)

## Checklist

N/A, since this is now also internal-only 🔐 


extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -61,6 +64,61 @@ export async function promptImportInternetDatabase(
...
+  const githubRepo = await window.showInputBox({
+    title: 'Enter a GitHub repository in the format <owner>/<repo> (e.g. github/codeql)',
+    placeHolder: '<owner>/<repo>',
+    ignoreFocusOut: true,
+  });
+  if (!githubRepo) {
+    return;
+  }
+
+  if (REPO_REGEX.test(githubRepo)) {
```
adityasharad, line 97: Style: You can reduce the number of nested if-else blocks to keep track of by flipping this test and returning early if the test returns false. In fact you used this trick just above!
```ts
if (!REPO_REGEX.test(githubRepo)) {
  return;
}
// rest of logic here
```

extensions/ql-vscode/src/authentication.ts
```diff
@@ -41,22 +49,33 @@ export class Credentials {
...
+    if (this.octokitAndToken) {
+      return this.octokitAndToken.token;
+    }
+    this.octokitAndToken = await this.createOctokit(true);
+    if (!this.octokitAndToken) {
       throw new Error('Did not initialize Octokit.');
     }
-    return this.octokit;
+    return this.octokitAndToken.token;
   }
```
adityasharad, line 80: Minor: You have a neat "lazy initialisation" pattern here, and you can make it even simpler to avoid the repetition between the `getOctokit` and `getToken` functions.

Couple of ideas:
- Is it possible to get the token string back out of an `Octokit` object? Then you can avoid the `OctokitAndToken` type with its separate `token` field. `await octokit.auth()` might do it, but I haven't tried! Otherwise stick with what you have.
- Consider having a private helper function here `getOctokitAndToken()` which does most of the work, including the creation and error handling, and returns `this.octokitAndToken`. Then `getOctokit()` becomes `return (await this.getOctokitAndToken()).octokit` (and `getToken` is similar).

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -166,13 +226,15 @@ export async function importArchiveDatabase(
...
  * @param databaseUrl URL from which to grab the database
+ * @param requestHeaders Headers to send with the request
  * @param databaseManager the DatabaseManager
  * @param storagePath where to store the unzipped database.
  * @param progress callback to send progress messages to
  * @param token cancellation token
  */
 async function databaseArchiveFetcher(
   databaseUrl: string,
+  requestHeaders: { [key: string]: string },
```
adityasharad, line 237: Clever, I see this lets you share the logic of fetching and unzipping archives between GitHub and LGTM.

adityasharad (commented):
Code changes look good! Minor suggestions only. Would it be possible to add unit tests for the new command, based on any we have for the existing LGTM command?

adityasharad:
For logos, we currently use both https://github.com/primer/octicons and https://github.com/microsoft/vscode-icons in this repo, and both have a GitHub icon. Give those a try :)

extensions/ql-vscode/src/authentication.ts
robertbrignull: Thank you for pointing out `octokit.auth()`. I didn't think there was a way to get the token back out of the octokit object. But I've tried this out locally and it works perfectly, so I agree we can remove all these changes to `authentication.ts` and instead use `.auth()` when we need the token.

extensions/ql-vscode/src/databaseFetcher.ts
shati-patel: Yes! Props to @robertbrignull ⚡

shati-patel:
Thanks for the helpful review! 😍 Addressed the comments, and made another small change to hide this command behind the canary flag.

Currently working on unit tests! 🧪 😅 

> For logos, we currently use icons from both https://github.com/primer/octicons and https://github.com/microsoft/vscode-icons in this repo, and both have a GitHub icon. Give those a try :)

Thank you! Found in https://github.com/microsoft/vscode-icons. Only thing is that some of the other logos have a small green "plus" ➕ 
image
I don't think it matters too much that we don't have that for the GitHub (and general "cloud") logos, especially since it's hidden behind canary.

shati-patel:
Added some tests in https://github.com/github/vscode-codeql/pull/1229/commits/d7a1e426834a9004cb84570a11a1f85aa50227e0!
It's not ideal, since we can't easily make any real octokit requests, but I've tested with a few mock API responses instead. Re-reviews/suggestions welcome 🌅 🙏🏽

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -506,7 +618,7 @@ async function promptForLanguage(
     maxStep: 2
   });
   if (!languages.length) {
-    return;
+    throw new Error('No databases found');
```
aeisenberg, line 621: This is also a change to downloading LGTM databases. I can't remember if there was a reason why we didn't throw when there were no languages found for an LGTM project. Can you just make sure that nothing awkward happens when downloading LGTM dbs?

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -43,7 +43,7 @@ interface QueriesResponse {
  * - `owner` is made up of alphanumeric characters or single hyphens, starting and ending in an alphanumeric character
  * - `repo` is made up of alphanumeric characters, hyphens, or underscores
  */
-const REPO_REGEX = /^(?:[a-zA-Z0-9]+-)*[a-zA-Z0-9]+\/[a-zA-Z0-9-_]+$/;
+export const REPO_REGEX = /^(?:[a-zA-Z0-9]+-)*[a-zA-Z0-9]+\/[a-zA-Z0-9-_]+$/;
```
aeisenberg, line 46: Probably good to move this to a new file. I don't think `databaseFetcher` should be depending on `run-remote-query`. Maybe `helpers` or `helpers-pure`.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -381,6 +462,37 @@ export async function findDirWithFile(
...
+    // TODO: In future, we could accept GitHub URLs in addition to NWOs.
+    // Similar to "looksLikeLgtmUrl".
+    if (!REPO_REGEX.test(githubRepo)) {
+      throw new Error('Invalid repository format. Must be in the format <owner>/<repo> (e.g. github/codeql)');
+    }
+
+    const [owner, repo] = githubRepo.split('/');
+
+    const octokit = await credentials.getOctokit();
+    const response = await octokit.request('GET /repos/:owner/:repo/code-scanning/codeql/databases', { owner, repo });
```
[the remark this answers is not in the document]
robertbrignull: The API currently does not allow anonymous requests because it is only staff shipped. Even in the private beta it will still require logging in. Once we get to public beta it will likely start accepting anonymous requests for public repos.

However for the extension I'd argue we still want to always make the user log in. It'll simplify our lives because we only have to handle one case. And I would expect the requirement to log in is not a blocked to users here.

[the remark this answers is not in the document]
aeisenberg: Thanks for the explanation. I think that's fine for now. It does seem like a small barrier for users who are not using any other github features. This is something we can discuss later when the API endpoint starts accepting anonymous requests.

extensions/ql-vscode/src/databaseFetcher.ts
shati-patel: I wasn't able to find an LGTM project that doesn't have any languages, so I suspect that's a sufficiently rare case that we've never needed to handle it previously 😅

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
shati-patel: Done! ✔️

extensions/ql-vscode/media/dark/github.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
adityasharad, line 1: Consider an HTML comment (if SVG allows them) or a README file stating where you got these icons from.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -61,6 +64,76 @@ export async function promptImportInternetDatabase(
...
+  if (!githubRepo) {
+    return;
+  }
+
+  if (!REPO_REGEX.test(githubRepo)) {
+    throw new Error(`Invalid GitHub repository: ${githubRepo}`);
+  }
+
+  const databaseUrl = await convertGithubNwoToDatabaseUrl(githubRepo, credentials, progress);
+  if (databaseUrl) {
```
adityasharad, line 102: Minor: You can do the same early-exit trick here:
```ts
if (!databaseUrl) {
  return
}
// continue
```

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -61,6 +64,76 @@ export async function promptImportInternetDatabase(
...
+    /** 
+     * A token object returned by `octokit.auth()`. It's undocumented, but looks something like this:
+     * {
+     *   token: 'xxxx',
+     *   tokenType: 'oauth',
+     *   type: 'token',
+     * }
+     * We only need the actual token string.
+     */
+    const octokitToken = await octokit.auth() as { token: string };
```
adityasharad, line 113: Minor: Could this be `(await octokit.auth())?.token`, and then you don't need to use `.token` everywhere below?

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -61,6 +64,76 @@ export async function promptImportInternetDatabase(
...
+      storagePath,
+      progress,
+      token,
+      cli
+    );
+    if (item) {
+      await commands.executeCommand('codeQLDatabases.focus');
+      void showAndLogInformationMessage('Database downloaded and imported successfully.');
+    }
+    return item;
```
adityasharad, line 132: Minor: Move this up into the `if(item)` case, so it's more obvious what's going on. The `return` at the end of the function will handle all the fall-through cases where we didn't return early.

adityasharad (approved):
Good stuff. I think you've addressed my comments, and I'll leave @aeisenberg to approve when he's happy.

extensions/ql-vscode/src/databaseFetcher.ts
shati-patel: Makes sense! I still had to keep `as { token: string }`, since `await octokit.auth()` is of type "unknown" otherwise 🤷🏽

aeisenberg (approved):
Thanks for looking into all of my comments. I think you've addressed Aditya's as well. So... 🚢.

PR #1243, opened 2022-03-28
Gate show eval log and summary commands behind CLI v2.8.4

angelapwen:
This PR will make it so that the "Show Evaluation Log (Raw)" and "Show Evaluation Log (Summary)" commands in the query history view only if the CLI version is >2.9.0. Previously, the commands always showed regardless of CLI version, and the user received an error pop-up if their CLI checkout was an earlier version. Once this is merged to the appropriate base branch, this change will allow us to merge #1186 before CLI v2.8.4 is released.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -8,7 +8,7 @@ import { Readable } from 'stream';
 import { StringDecoder } from 'string_decoder';
 import * as tk from 'tree-kill';
 import { promisify } from 'util';
-import { CancellationToken, Disposable, Uri } from 'vscode';
+import { CancellationToken, commands, Disposable, Uri } from 'vscode';
 
 import { BQRSInfo, DecodedBqrsChunk } from './pure/bqrs-cli-types';
 import { CliConfig } from './config';
@@ -957,6 +957,9 @@ export class CodeQLCliServer implements Disposable {
   public async getVersion() {
     if (!this._version) {
       this._version = await this.refreshVersion();
+      await commands.executeCommand(
+        'setContext', 'codeql.supportsEvalLog', await this.cliConstraints.supportsPerQueryEvalLog()
+      );
     }
     return this._version;
   }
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -957,6 +957,9 @@ export class CodeQLCliServer implements Disposable {
   public async getVersion() {
     if (!this._version) {
       this._version = await this.refreshVersion();
+      await commands.executeCommand(
+        'setContext', 'codeql.supportsEvalLog', await this.cliConstraints.supportsPerQueryEvalLog()
+      );
```
adityasharad, line 962: I like the simplicity of this change. Is this the best place to execute this command, though? It will be invoked every time the extension asks for the CodeQL CLI version, which may be frequent. I expect this command to be executed each time there is a potential change to the configured CLI -- see what the `CliConfigListener` and `QueryServerConfigListener` classes in `config.ts` do, and whether that can be adapted to suit your purpose.

aeisenberg: This is actually what is happening. The only time `this._version` is undefined is after a config change, where the version may have potentially changed. The version and context are cached until the next time the config changes. By putting it in this if statement, we are guaranteed to call the command the minimal number of times.

adityasharad: Fancy. Looks good then! Perhaps add a comment to explain that, so I don't pester you about it again :)

extensions/ql-vscode/src/cli.ts
```diff
@@ -1276,7 +1279,7 @@ export class CliVersionConstraint {
    /**
     * CLI version that supports rotating structured logs to produce one per query.
     */
-    public static CLI_VERSION_WITH_PER_QUERY_EVAL_LOG = new SemVer('2.8.4');
+    public static CLI_VERSION_WITH_PER_QUERY_EVAL_LOG = new SemVer('2.9.0');
```
edoardopirovano, line 1282: Why the change of version here? I believe everything we need for this is in 2.8.4.

angelapwen: Ah, I got a bit mixed up with all the different PRs — the incoming versions to log to the query server output (that I had been planning to merge into your PR) will be in 2.9.0, but it does seem unnecessary to wait until 2.9.0 for these changes. 

So if a user is using 2.8.4, they would see the commands (and log files) but not any output in the query server console; and if they are using 2.9.0+ they would see the query server console output as well. I think that is reasonable (without adding another gate for 2.9.0).

edoardopirovano (approved):

PR #1246, opened 2022-03-28
Add repositories search box

aeisenberg:
A simple, webview-only search box for filtering repositories from the remote queries results view.

- [x] ~TODO: Need to handle dark mode.~ 

Do we want any animations for showing/hiding repos?

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
+++ b/extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
@@ -17,7 +17,6 @@ const RepositoriesSearch = ({ filterValue, setFilterValue }: RepositoriesSearchP
         color: 'var(--vscode-editor-foreground);',
         width: 'calc(100% - 14px)',
       }}
-      className="vscode-codeql__repositories-search"
       leadingVisual={SearchIcon}
       aria-label="Repository search"
       name="repository-search"
--- a/extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
+++ b/extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
@@ -14,7 +14,8 @@ const RepositoriesSearch = ({ filterValue, setFilterValue }: RepositoriesSearchP
       block
       sx={{
         backgroundColor: 'var(--vscode-editor-background);',
-        color: 'var(--vscode-editor-foreground);'
+        color: 'var(--vscode-editor-foreground);',
+        width: 'calc(100% - 14px)',
       }}
       className="vscode-codeql__repositories-search"
       leadingVisual={SearchIcon}
--- a/extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
+++ b/extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
@@ -1,0 +1,30 @@
+import * as React from 'react';
+import { ChangeEvent } from 'react';
+import { TextInput } from '@primer/react';
+import { SearchIcon } from '@primer/octicons-react';
+
+interface RepositoriesSearchProps {
+  filterValue: string;
+  setFilterValue: (value: string) => void;
+}
+
+const RepositoriesSearch = ({ filterValue, setFilterValue }: RepositoriesSearchProps) => {
+  return <>
+    <TextInput
+      block
+      sx={{
+        backgroundColor: 'var(--vscode-editor-background);',
+        color: 'var(--vscode-editor-foreground);'
+      }}
+      className="vscode-codeql__repositories-search"
+      leadingVisual={SearchIcon}
+      aria-label="Repository search"
+      name="repository-search"
+      placeholder="Filter by repository owner/name"
+      value={filterValue}
+      onChange={(e: ChangeEvent) => setFilterValue((e.target as HTMLInputElement).value)}
+    />
+  </>;
+};
+
+export default RepositoriesSearch;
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
```diff
@@ -0,0 +1,30 @@
...
+interface RepositoriesSearchProps {
+  filterValue: string;
+  setFilterValue: (value: string) => void;
+}
+
+const RepositoriesSearch = ({ filterValue, setFilterValue }: RepositoriesSearchProps) => {
+  return <>
+    <TextInput
+      block
+      sx={{
```
charisk, line 15: This doesn't look like it's needed. We what it looks like without it and it looks fine in both modes to us.

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
```diff
@@ -0,0 +1,30 @@
...
+        backgroundColor: 'var(--vscode-editor-background);',
+        color: 'var(--vscode-editor-foreground);'
+      }}
+      className="vscode-codeql__repositories-search"
```
charisk, line 19: This class is added to add some space to the top. I've generally used `<VerticalSpace />` for that e.g. `<VerticalSpace size={2} />`

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { SearchIcon } from '@primer/octicons-react';
+
+interface RepositoriesSearchProps {
+  filterValue: string;
+  setFilterValue: (value: string) => void;
+}
+
+const RepositoriesSearch = ({ filterValue, setFilterValue }: RepositoriesSearchProps) => {
+  return <>
+    <TextInput
```
charisk, line 13: In a large screen this looks a bit silly. Shall we set max-width? e.g. `max-width: 30em`.

image

charisk (commented):
Looking good, just some minor comments.

Reviewed with @robertbrignull @shati-patel

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
aeisenberg: Should we have a max-width for the entire results view and just let the search box follow that?

The wireframes have the search box exactly as wide as the NWO results.

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
aeisenberg: Sure, I can change that. Just wondering what the benefit of using extra dom for adding space instead of using css.

aeisenberg:
@charisk what do you think of the latest commit? It adds a max-width to the entire page and all sub-elements respect that value. Now the search bar is always _exactly_* as wide as the results below it.

* Actually, I need to change from % to fixed difference since on narrow screens, the values are not correct. Let me fix that.

aeisenberg:
done

extensions/ql-vscode/src/remote-queries/view/remoteQueries.css
```diff
@@ -41,7 +45,6 @@
 }
 
 .vscode-codeql__repositories-search {
```
charisk, line 47: This can be deleted now.

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
charisk: IMO it's much easier to follow/reason with rather than poking with margins and padding (and dealing with conflicting margins, different context/layout etc.)

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
charisk: Hmm yeah I have no idea what it should look like in red 😂 we just checked light and dark mode. Ideally we'd be setting colors through the Primer theme but that's something to look into separately!

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
```diff
@@ -14,7 +14,8 @@ const RepositoriesSearch = ({ filterValue, setFilterValue }: RepositoriesSearchP
       block
       sx={{
         backgroundColor: 'var(--vscode-editor-background);',
-        color: 'var(--vscode-editor-foreground);'
+        color: 'var(--vscode-editor-foreground);',
+        width: 'calc(100% - 14px)',
```
charisk, line 18: Why is this needed? If I turn it off things look the same in my view. Is it trying to add a padding of some sort?

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
charisk: Good call!

charisk (approved):

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
aeisenberg: Yes. The width of the search bar needs to be exactly the same as the width of the analysis results section. I was trying to find a way to get this to work using padding or margin, but because we are using primer, we can't attach css to the elements that would be affected by padding or margin elements. This is the only way I could figure out how to get it working.

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
aeisenberg: Handling all kinds of themes will be very, very tricky. But, I think we should at least make a best effort for the common themes (though, I don't know if red is very common).

PR #1264, opened 2022-03-30
Print end-of-query summary logs to Query Server Console

angelapwen:
This change passes the `--end-summary=$eoq-evaluator-log.jsonl.summary` option to the CLI so that a separate file in the given location is written, with only the end-of-query summary information (most expensive predicate and timings). It then directly logs the contents of this file to the query server console.

It also changes the CLI version we gate all evaluator logs behind to 2.9.0 for simplicity. Otherwise, we would need to pass in one set of arguments to the CLI for 2.8.4 and another for 2.9.0. If in the future a user must be on a 2.8.x release and would like evaluator logs, we will be able to modify this for them easily.

This has been tested locally on VSCode (by modifying the CLI version expected to 2.8.3 and using the main checkout of the CLI) and works as we expect:
Screen Shot 2022-03-30 at 4 18 36 PM

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -781,6 +781,11 @@ export class QueryHistoryManager extends DisposableObject {
     void showAndLogWarningMessage('No evaluator log is available for this run. Perhaps it failed before evaluation, or you are running with a version of CodeQL before ' + CliVersionConstraint.CLI_VERSION_WITH_PER_QUERY_EVAL_LOG + '?');
   }
 
+  private warnNoEvalLogSummary() {
+    void showAndLogWarningMessage('No evaluator log summary is available for this run. Perhaps it failed before evaluation, or you are running with a version of CodeQL before ' + CliVersionConstraint.CLI_VERSION_WITH_PER_QUERY_EVAL_LOG + '?');
+  }
+
+
   async handleShowEvalLog(
     singleItem: QueryHistoryInfo,
     multiSelect: QueryHistoryInfo[]
@@ -810,13 +815,10 @@ export class QueryHistoryManager extends DisposableObject {
       return;
     }
 
-    if (finalSingleItem.evalLogLocation) {
-      if (!fs.existsSync(finalSingleItem.evalLogSummaryLocation)) {
-        await this.qs.cliServer.generateLogSummary(finalSingleItem.evalLogLocation, finalSingleItem.evalLogSummaryLocation);
-      }
-      await this.tryOpenExternalFile(finalSingleItem.evalLogSummaryLocation);
+    if (finalSingleItem.evalLogSummaryLocation) {
+        await this.tryOpenExternalFile(finalSingleItem.evalLogSummaryLocation);
     } else {
-      this.warnNoEvalLog();
+      this.warnNoEvalLogSummary();
     }
   }
 
--- a/extensions/ql-vscode/src/queryserver-client.ts
+++ b/extensions/ql-vscode/src/queryserver-client.ts
@@ -259,14 +259,14 @@ export function findQueryLogFile(resultPath: string): string {
   return path.join(resultPath, 'query.log');
 }
 
-export function findQueryStructLogFile(resultPath: string): string {
+export function findQueryEvalLogFile(resultPath: string): string {
   return path.join(resultPath, 'evaluator-log.jsonl');
 }
 
-export function findQueryStructLogSummaryFile(resultPath: string): string {
+export function findQueryEvalLogSummaryFile(resultPath: string): string {
   return path.join(resultPath, 'evaluator-log.jsonl.summary');
 }
 
-export function findQueryStructLogEndSummaryFile(resultPath: string): string {
+export function findQueryEvalLogEndSummaryFile(resultPath: string): string {
   return path.join(resultPath, 'eoq-evaluator-log.jsonl.summary');
 }
--- a/extensions/ql-vscode/src/queryserver-client.ts
+++ b/extensions/ql-vscode/src/queryserver-client.ts
@@ -262,3 +262,11 @@ export function findQueryLogFile(resultPath: string): string {
 export function findQueryStructLogFile(resultPath: string): string {
   return path.join(resultPath, 'evaluator-log.jsonl');
 }
+
+export function findQueryStructLogSummaryFile(resultPath: string): string {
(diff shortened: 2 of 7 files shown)


extensions/ql-vscode/src/query-history.ts
```diff
@@ -781,6 +781,11 @@ export class QueryHistoryManager extends DisposableObject {
     void showAndLogWarningMessage('No evaluator log is available for this run. Perhaps it failed before evaluation, or you are running with a version of CodeQL before ' + CliVersionConstraint.CLI_VERSION_WITH_PER_QUERY_EVAL_LOG + '?');
   }
 
+  private warnNoEvalLogSummary() {
+    void showAndLogWarningMessage('No evaluator log summary is available for this run. Perhaps it failed before evaluation, or you are running with a version of CodeQL before ' + CliVersionConstraint.CLI_VERSION_WITH_PER_QUERY_EVAL_LOG + '?');
```
aeisenberg, line 785: Minor:

```suggestion
    void showAndLogWarningMessage(`No evaluator log summary is available for this run. Perhaps it failed before evaluation, or you are running with a version of CodeQL before ${CliVersionConstraint.CLI_VERSION_WITH_PER_QUERY_EVAL_LOG}?`);
```

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -259,6 +259,14 @@ export function findQueryLogFile(resultPath: string): string {
   return path.join(resultPath, 'query.log');
 }
 
-export function findQueryStructLogFile(resultPath: string): string {
+export function findQueryEvalLogFile(resultPath: string): string {
   return path.join(resultPath, 'evaluator-log.jsonl');
 }
+
+export function findQueryEvalLogSummaryFile(resultPath: string): string {
+  return path.join(resultPath, 'evaluator-log.jsonl.summary');
```
aeisenberg, line 267: Is this a valid json or jsonl file? If so, it should probably use the proper file extension. Maybe:

```suggestion
  return path.join(resultPath, 'evaluator-log.summary.jsonl');
```

extensions/ql-vscode/src/queryserver-client.ts
```diff
@@ -259,6 +259,14 @@ export function findQueryLogFile(resultPath: string): string {
...
+}
+
+export function findQueryEvalLogEndSummaryFile(resultPath: string): string {
+  return path.join(resultPath, 'eoq-evaluator-log.jsonl.summary');
```
aeisenberg, line 271: Also here:

```suggestion
  return path.join(resultPath, 'eoq-evaluator-log.summary.jsonl');
```

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -186,9 +195,21 @@ export class QueryEvaluationInfo {
...
         });
-        queryInfo.evalLogLocation = this.structLogPath;
+        if (await this.hasEvalLog()) {
+          queryInfo.evalLogLocation = this.evalLogPath;
+          await qs.cliServer.generateLogSummary(this.evalLogPath, this.evalLogSummaryPath, this.evalLogEndSummaryPath);
+          fs.readFile(this.evalLogEndSummaryPath, (err, buffer) => {
+            if (err) {
+              throw new Error(`Could not read structured evaluator log end of summary file at ${this.evalLogEndSummaryPath}.`);
+            }
+            void qs.logger.log(' --- Evaluator Log Summary --- ', { additionalLogLocation: this.logPath });
```
aeisenberg, line 207: I _think_ this will add the `' --- Evaluator Log Summary --- '` to the file. Is this what we want? Is it better/possible to ensure the output is proper json or jsonl?

extensions/ql-vscode/src/queryserver-client.ts
angelapwen: It's not — it's a generated text file summarizing the contents of the jsonl file.

extensions/ql-vscode/src/run-queries.ts
angelapwen: This adds the ' --- Evaluator Log Summary --- ' line to the Query Server Console, which I believe is what we want. The file itself (the end summary file) should never be read by a user; it's generated on the CLI side specifically for this use case.

extensions/ql-vscode/src/queryserver-client.ts
aeisenberg: OK...then maybe avoid the `jsonl` component entirely?

extensions/ql-vscode/src/queryserver-client.ts
angelapwen: Same as above, it's not a JSONL file but rather a condensed version of the summary text file. I didn't know how to name it so appended `eoq` to the beginning of the file rather than at the end, but

`evaluator-log.jsonl.summary.eoq` might also work?

extensions/ql-vscode/src/run-queries.ts
aeisenberg: The additionalLogLocation will add the contents to `this.logPath`. If it's not meant to be read by a user and it is not jsonl syntax then adding this line to it as well is fine.

angelapwen: Hm, the user is meant to read this line (and the following lines) only from the Query Server Console. I can remove the `additionalLogLocation` part of the logging so that the user will not see traces of the evaluator structured logs in the additional log where they may not expect it. If they want the full summary of the structured log, they can view the actual summary file.

aeisenberg: You are much more familiar with this feature than I am. I am fine with it if you decide to leave this in.

extensions/ql-vscode/src/queryserver-client.ts
edoardopirovano: Just calling it `.summary` seems perfectly reasonable to me, yep!

extensions/ql-vscode/src/queryserver-client.ts
edoardopirovano: Maybe `evaluator-log-end.summary`? As I understand it, this is only used internally so I don't suppose it matters too much.

edoardopirovano (approved):
Looking good, thanks! I'm happy with this at it is, modulo you may want some simpler names for the files as discussed above 🙂

PR #1268, opened 2022-04-04
Use the repos defined in the query result instead of the query

charisk:
In order to support system defined repo lists (see linked internal issue) we need to stop relying on the repository names being available to us on the query object, but instead get that information from the query result.

This PR updates a couple of uses of the repo lists to be using the query result instead of the query.

## Checklist
N/A:


shati-patel (approved):
Looks good! 🌼

------

Just a comment about a minor behaviour change: Do we now only count repositories with results? E.g. 

Before this PR: 
image

vs 
After this PR:
image


I don't think this is necessarily a problem, especially for security researchers running on loads of repos at once! But (in future) we might want to change the design slightly, so we're not showing the number of repos with results in two places? 🤔

shati-patel (approved):
Ignore my previous comment! That turned out to be a different issue 😅

PR #1291, opened 2022-04-12
Handle cancelling of remote queries

aeisenberg:
This change issues a cancel request when the user clicks on "cancel" for
a remote query.

The cancel can take quite a while to complete, so a message is popped up
to let the user know.

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -36,6 +36,8 @@ import { QueryStatus } from './query-status';
 import { slurpQueryHistory, splatQueryHistory } from './query-serialization';
 import * as fs from 'fs-extra';
 import { CliVersionConstraint } from './cli';
+import { Credentials } from './authentication';
+import { cancelRemoteQuery } from './remote-queries/gh-actions-api-client';
 
 /**
  * query-history.ts
@@ -318,6 +320,7 @@ export class QueryHistoryManager extends DisposableObject {
     private queryStorageDir: string,
     ctx: ExtensionContext,
     private queryHistoryConfigListener: QueryHistoryConfig,
+    private readonly getCredentials: () => Promise<Credentials>,
     private doCompareCallback: (
       from: CompletedLocalQueryInfo,
       to: CompletedLocalQueryInfo
@@ -816,7 +819,7 @@ export class QueryHistoryManager extends DisposableObject {
     }
 
     if (finalSingleItem.evalLogSummaryLocation) {
-        await this.tryOpenExternalFile(finalSingleItem.evalLogSummaryLocation);
+      await this.tryOpenExternalFile(finalSingleItem.evalLogSummaryLocation);
     } else {
       this.warnNoEvalLogSummary();
     }
@@ -830,11 +833,20 @@ export class QueryHistoryManager extends DisposableObject {
     // In the future, we may support cancelling remote queries, but this is not a short term plan.
     const { finalSingleItem, finalMultiSelect } = this.determineSelection(singleItem, multiSelect);
 
-    (finalMultiSelect || [finalSingleItem]).forEach((item) => {
-      if (item.status === QueryStatus.InProgress && item.t === 'local') {
-        item.cancel();
+    const selected = finalMultiSelect || [finalSingleItem];
+    const results = selected.map(async item => {
+      if (item.status === QueryStatus.InProgress) {
+        if (item.t === 'local') {
+          item.cancel();
+        } else if (item.t === 'remote') {
+          void showAndLogInformationMessage('Cancelling remote query. This may take a while.');
+          const credentials = await this.getCredentials();
+          await cancelRemoteQuery(credentials, item.remoteQuery);
+        }
       }
     });
+
+    await Promise.all(results);
   }
 
   async handleShowQueryText(
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -318,9 +318,8 @@ export class QueryHistoryManager extends DisposableObject {
     private qs: QueryServerClient,
     private dbm: DatabaseManager,
     private queryStorageDir: string,
-    ctx: ExtensionContext,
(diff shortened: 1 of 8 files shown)


extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -155,9 +155,16 @@ export class RemoteQueriesManager extends DisposableObject {
...
       }
     } else if (queryWorkflowResult.status === 'CompletedUnsuccessfully') {
-      queryItem.failureReason = queryWorkflowResult.error;
-      queryItem.status = QueryStatus.Failed;
-      void showAndLogErrorMessage(`Variant analysis execution failed. Error: ${queryWorkflowResult.error}`);
+      if (queryWorkflowResult.error?.includes('cancelled')) {
+        // workflow was cancelled on the server
+        queryItem.failureReason = 'Cancelled';
+        queryItem.status = QueryStatus.Failed;
+        void showAndLogErrorMessage('Variant analysis monitoring was cancelled');
```
charisk, line 162: It feels like this should be an informational message rather than an error message. But I'm not 100%.

extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
```diff
@@ -74,6 +74,18 @@ export async function getRemoteQueryIndex(
   };
 }
 
+export async function cancelRemoteQuery(
+  credentials: Credentials,
+  remoteQuery: RemoteQuery
+): Promise<void> {
+  const octokit = await credentials.getOctokit();
+  const { actionsWorkflowRunId, controllerRepository: { owner, name } } = remoteQuery;
+  const response = await octokit.request(`POST /repos/${owner}/${name}/actions/runs/${actionsWorkflowRunId}/cancel`);
```
shati-patel, line 83: Optional suggestion 😺 

You could use the octokit "wrapper" for cancelling a workflow run from https://octokit.github.io/rest.js/v18#actions-cancel-workflow-run

extensions/ql-vscode/src/query-history.ts
```diff
@@ -830,11 +833,20 @@ export class QueryHistoryManager extends DisposableObject {
...
-    (finalMultiSelect || [finalSingleItem]).forEach((item) => {
-      if (item.status === QueryStatus.InProgress && item.t === 'local') {
-        item.cancel();
+    const selected = finalMultiSelect || [finalSingleItem];
+    const results = selected.map(async item => {
+      if (item.status === QueryStatus.InProgress) {
+        if (item.t === 'local') {
+          item.cancel();
+        } else if (item.t === 'remote') {
+          void showAndLogInformationMessage('Cancelling remote query. This may take a while.');
```
shati-patel, line 842: Probably?
```suggestion
          void showAndLogInformationMessage('Cancelling variant analysis. This may take a while.');
```

extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
```diff
@@ -74,6 +74,18 @@ export async function getRemoteQueryIndex(
...
+  if (response.status >= 300) {
+    throw new Error(`Error cancelling remote query: ${response.status}`);
```
shati-patel, line 85:
```suggestion
    throw new Error(`Error cancelling variant analysis: ${response.status}`);
```

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
shati-patel, line 162: Minor: Is "monitoring" the right word here (and below on line 171)? It looks like the variant analysis (i.e. the workflow run) itself is cancelled 🐱
```suggestion
        void showAndLogErrorMessage('Variant analysis was cancelled');
```

shati-patel (commented):
Nice! Left a few tiny comments too 🔍

extensions/ql-vscode/src/query-history.ts
aeisenberg: Yes. There are some other places where we use "remote query" or "remote queries" in log messages that we should fix. I'll make the change here. And we should create a new PR for other instances of the old terminology.

aeisenberg: https://github.com/github/code-scanning/issues/5782

extensions/ql-vscode/src/query-history.ts
```diff
@@ -318,6 +320,7 @@ export class QueryHistoryManager extends DisposableObject {
     private queryStorageDir: string,
     ctx: ExtensionContext,
     private queryHistoryConfigListener: QueryHistoryConfig,
+    private readonly getCredentials: () => Promise<Credentials>,
```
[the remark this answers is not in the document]
aeisenberg: This is the same pattern being used in the constructor to [`DatabasesUI`](https://github.com/github/vscode-codeql/blob/61d4305593a2c001ee3cc5e846ffaee189d1d5ca/extensions/ql-vscode/src/extension.ts#L438).

My original intention was to do this to make testing easier, but since I'm not directly testing this feature maybe I can change it to what you suggest.

extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
aeisenberg: Hmmm...didn't know about that. But, I have a preference for keeping it closer to the actual rest API since it will be easier for us to look up the docs if we ever need to.

extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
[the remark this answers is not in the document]
aeisenberg: Yes. I think you're right.

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
aeisenberg: Yes

charisk (approved):

PR #1292, opened 2022-04-12
Rename remote queries -> variant analysis

aeisenberg:
In some user facing text.

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -667,11 +667,11 @@ export class CodeQLCliServer implements Disposable {
 
   /**
   * Generate a summary of an evaluation log.
-  * @param endSummaryPath The path to write only the end of query part of the human-readable summary to. 
+  * @param endSummaryPath The path to write only the end of query part of the human-readable summary to.
   * @param inputPath The path of an evaluation event log.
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -1255,7 +1255,7 @@ export class CliVersionConstraint {
   public static CLI_VERSION_WITH_NO_PRECOMPILE = new SemVer('2.7.1');
 
   /**
-   * CLI version where remote queries are supported.
+   * CLI version where remote queries (variant analysis) are supported.
```
charisk, line 1258: I think it's ok to assume that people reading the code know that remote queries = variant analysis (otherwise where we could end up adding `(variant analysis)` everywhere.

extensions/ql-vscode/src/config.ts
```diff
@@ -322,7 +322,7 @@ export function isCanary() {
  */
 export const NO_CACHE_AST_VIEWER = new Setting('disableCache', AST_VIEWER_SETTING);
 
-// Settings for remote queries
+// Settings for variant analysis
```
charisk, line 325: Same comment - this is not user facing.

charisk (approved):
Thanks for tidying that up! Just a couple of suggestions but LGTM.

PR #1298, opened 2022-04-14
Avoid loading wrong results into an open window

aeisenberg:
This fixes a bug where an open results view will accumulate results from
other queries who have their results downloaded while this view is open.

The fix is to ensure that the results view for the query is open when
some results are downloaded.


shati-patel (approved):
Looks good to me! 🚢

PR #1299, opened 2022-04-14
Remove jsonc dependency

aeisenberg:
This dependency was only used to parse package.json and
this can be just as easily parsed by regular JSON object.

jsonc can also parse JSON with comments, but there are no
comments in package.json.

See https://github.com/github/vscode-codeql/pull/1262


angelapwen (approved):
🚢

PR #1304, opened 2022-04-20
Bug fix for evaluator log commands in query history view

angelapwen:
I realized I had not opened a PR for this change. Previously, the options for showing raw evaluator logs and summaries only populated for full query runs but not quick evaluator runs. We discovered this was due to the [VSCode `when` clause](https://code.visualstudio.com/api/references/when-clause-contexts#add-a-custom-when-clause-context) not supporting parentheses. This change is a quick fix to display the raw and summary logs for quick evaluation runs as well.

Note that I have not updated the CHANGELOG because these commands are gated behind CLI v.2.9.0, which is not yet released, so this should not affect any existing users. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -735,12 +735,12 @@
         {
           "command": "codeQLQueryHistory.showEvalLog",
           "group": "9_qlCommands",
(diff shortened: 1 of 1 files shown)


aeisenberg (approved):
Looks good. In the future, we should change this to using regex to match the viewItems, but that is a larger change we should apply throughout.

PR #1305, opened 2022-04-20
Update the warning message after running variant analysis

aeisenberg:
Adds more information about onboarding new repos.

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
+++ b/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
@@ -353,6 +353,7 @@ export function parseResponse(owner: string, repo: string, response: QueriesResp
     }
     if (response.errors.repositories_without_database?.length) {
       logMessage += `\n\nRepositories without databases:\n${response.errors.repositories_without_database.join(', ')}`;
+      logMessage += '\nThese repositories have been added to the database storage service and we will attempt to create a database for them next time the store is updated.';
     }
   }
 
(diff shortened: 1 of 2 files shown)


shati-patel (approved):
🚢

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -353,6 +353,7 @@ export function parseResponse(owner: string, repo: string, response: QueriesResp
     }
     if (response.errors.repositories_without_database?.length) {
       logMessage += `\n\nRepositories without databases:\n${response.errors.repositories_without_database.join(', ')}`;
+      logMessage += '\nThese repositories have been added to the database storage service and we will attempt to create a database for them next time the store is updated.';
```
charisk, line 356: Should we be using `os.EOL` instead of `\n`?

aeisenberg: Sure. I'll make that change.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -338,22 +339,25 @@ async function runRemoteQueriesApiRequest(
 }
 
 // exported for testng only
```
charisk, line 341: Might want to move the comment to be on top of the function.

charisk (approved):

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -338,22 +339,25 @@ async function runRemoteQueriesApiRequest(
...
+const eol = os.EOL;
+const eol2 = os.EOL + os.EOL;
+
 export function parseResponse(owner: string, repo: string, response: QueriesResponse) {
```
aeisenberg, line 345:
```suggestion
const eol = os.EOL;
const eol2 = os.EOL + os.EOL;

// exported for testng only
export function parseResponse(owner: string, repo: string, response: QueriesResponse) {
```

PR #1310, opened 2022-04-21
MRVA: Display available results, even if some jobs are cancelled

shati-patel:
Instead of fully failing when a workflow run is cancelled, we check whether there is a result index available. If so, we display the available results! See internal issue for more info 😎 

Note: There's no indication of a "partially cancelled" query in the query history, but we do still display a popup in both cases saying that the variant analysis was cancelled. See screenshots below. I'm very open to alternative UX suggestions 😄 

- Cancelled with no artifacts to download: 
   image

- Cancelled, but there are artifacts to download:
   image

## Checklist

N/A - still internal 🌱

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
@@ -145,10 +145,10 @@ export class RemoteQueriesManager extends DisposableObject {
         void showAndLogErrorMessage(`Variant analysis execution failed. Error: ${queryWorkflowResult.error}`);
       }
     } else if (queryWorkflowResult.status === 'Cancelled') {
-      // workflow was cancelled from within VS Code
       queryItem.failureReason = 'Cancelled';
       queryItem.status = QueryStatus.Failed;
-      void showAndLogErrorMessage('Variant analysis was cancelled');
+      await this.downloadAvailableResults(queryItem, credentials, executionEndTime);
+      void showAndLogInformationMessage('Variant analysis was cancelled');
     } else if (queryWorkflowResult.status === 'InProgress') {
       // Should not get here. Only including this to ensure `assertNever` uses proper type checking.
       void showAndLogErrorMessage(`Unexpected status: ${queryWorkflowResult.status}`);
@@ -268,7 +268,11 @@ export class RemoteQueriesManager extends DisposableObject {
    * Checks whether there's a result index artifact available for the given query.
    * If so, set the query status to `Completed` and auto-download the results.
    */
-  private async downloadAvailableResults(queryItem: RemoteQueryHistoryItem, credentials: Credentials, executionEndTime: number): Promise<void> {
+  private async downloadAvailableResults(
+    queryItem: RemoteQueryHistoryItem,
+    credentials: Credentials,
+    executionEndTime: number
+  ): Promise<void> {
     const resultIndex = await getRemoteQueryIndex(credentials, queryItem.remoteQuery);
     if (resultIndex) {
       queryItem.completed = true;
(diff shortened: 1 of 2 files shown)


charisk (approved):

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -131,40 +131,21 @@ export class RemoteQueriesManager extends DisposableObject {
...
-        queryItem.status = QueryStatus.Failed;
-      }
+      await this.downloadAvailableResults(queryItem, credentials, executionEndTime);
     } else if (queryWorkflowResult.status === 'CompletedUnsuccessfully') {
       if (queryWorkflowResult.error?.includes('cancelled')) {
         // workflow was cancelled on the server
         queryItem.failureReason = 'Cancelled';
         queryItem.status = QueryStatus.Failed;
+        await this.downloadAvailableResults(queryItem, credentials, executionEndTime);
         void showAndLogInformationMessage('Variant analysis was cancelled');
```
aeisenberg, line 141: I don't really like my suggestion, but we should get the point across that we are downloading the results that exist.

```suggestion
        void showAndLogInformationMessage('Variant analysis was cancelled. Downloading partial results.');
```

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -131,40 +131,21 @@ export class RemoteQueriesManager extends DisposableObject {
...
       } else {
         queryItem.failureReason = queryWorkflowResult.error;
         queryItem.status = QueryStatus.Failed;
         void showAndLogErrorMessage(`Variant analysis execution failed. Error: ${queryWorkflowResult.error}`);
       }
     } else if (queryWorkflowResult.status === 'Cancelled') {
+      // workflow was cancelled from within VS Code
```
aeisenberg, line 148: Are there any partial results to download in this case?

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -282,4 +263,34 @@ export class RemoteQueriesManager extends DisposableObject {
...
+
+      // Ask if the user wants to open the results in the background.
+      void this.askToOpenResults(queryItem.remoteQuery, queryResult).then(
+        noop,
+        err => {
+          void showAndLogErrorMessage(err);
+        }
+      );
+    } else {
+      void showAndLogErrorMessage(`There was an issue retrieving the result for the query ${queryItem.remoteQuery.queryName}`);
```
aeisenberg, line 292: Since we know what the URL of the workflow is, maybe we should add it here. (Not exactly part of this PR, but still a nice to have.)

aeisenberg (commented):
Looks good. Some suggestions inline.

aeisenberg:
One more suggestion I just thought of. I think it would be nice to have a different icon for cancelled runs that have partial results. It's not really a scary red X, nor is it a happy globe. Maybe some sort of `!` would be better? I don't remember how complex the logic is getting for the icons, so if it's not easy, we can delay this until later.

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
charisk: FWIW I don't think we necessarily need this - the usual view results notification is shown.

charisk:
> One more suggestion I just thought of. I think it would be nice to have a different icon for cancelled runs that have partial results. It's not really a scary red X, nor is it a happy globe. Maybe some sort of `!` would be better? I don't remember how complex the logic is getting for the icons, so if it's not easy, we can delay this until later.

We think that would need introduction of a "partially failed" query status for query history items. This is something we'll probably want to do fairly soon but we wanted to hear some feedback from Product/UX first. It can certainly be done in a separate PR. Thoughts?

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
shati-patel: Agreed that we could do with some better UI 🙈 I'll leave as-is for now, and we can follow-up with a nicer experience later. (Along with query history icons, as mentioned separately!)

shati-patel:
> > One more suggestion I just thought of. I think it would be nice to have a different icon for cancelled runs that have partial results. It's not really a scary red X, nor is it a happy globe. Maybe some sort of `!` would be better? I don't remember how complex the logic is getting for the icons, so if it's not easy, we can delay this until later.
> 
> We think that would need introduction of a "partially failed" query status for query history items. This is something we'll probably want to do fairly soon but we wanted to hear some feedback from Product/UX first (see internal issue). It can certainly be done in a separate PR. Thoughts?

Agreed that we should do this, but let's wait for some feedback on the overall UX first 🙂  I'll merge this (since it's an improvement over the current situation), and we can follow-up with some polish separately! 💅🏽

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
aeisenberg: It's possible that this branch is never hit any more now that we are handling cancels from the server. Not sure...just a possibility.

PR #1312, opened 2022-04-22
MRVA: Make markdown code snippets look nicer

shati-patel:
A quick PR to remove some extraneous newlines in our markdown rendering! 🔪 (Because otherwise the code blocks looked huge) 

See [before](https://github.com/github/vscode-codeql/blob/1a03c0e4acc3a449b6e20ddc0e90c59d630273ba/extensions/ql-vscode/test/pure-tests/remote-queries/markdown-generation/data/interpreted-results/path-problem/results-repo1.md) 😢 and [after](https://github.com/github/vscode-codeql/blob/b8fef48757f0a4090acbac731d317c757f3aff3a/extensions/ql-vscode/test/pure-tests/remote-queries/markdown-generation/data/interpreted-results/path-problem/results-repo1.md) 😄 

## Checklist

N/A - internal only 🤭

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #1313, opened 2022-04-22
Markdown rendering: Display paths

shati-patel:
_(based on #1312)_

Render path results in an expandable section (as shown in the example in the linked internal issue!) 

Click [here](https://github.com/github/vscode-codeql/blob/shati-patel/sharing-path-results/extensions/ql-vscode/test/pure-tests/remote-queries/markdown-generation/data/interpreted-results/path-problem/results-repo1.md) for a cute rendered preview  🖼️ 

## Checklist

N/A - internal only 😄 

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
@@ -89,12 +89,18 @@ function generateMarkdownForInterpretedResult(interpretedResult: AnalysisAlert, 
       ...generateMarkdownForCodeSnippet(codeSnippet, language, highlightedRegion),
     );
   }
-  const alertMessageLines = buildMarkdownAlertMessage(interpretedResult, language);
-  lines.push(...alertMessageLines);
+  const alertMessage = generateMarkdownForAlertMessage(interpretedResult);
+  lines.push(alertMessage, '');
+
+  // If available, show paths
+  const hasPathResults = interpretedResult.codeFlows.length > 0;
+  if (hasPathResults) {
+    const pathLines = generateMarkdownForPathResults(interpretedResult, language);
+    lines.push(...pathLines);
+  }
 
   // Padding between results
   lines.push(
-    '',
     '----------------------------------------',
     '',
   );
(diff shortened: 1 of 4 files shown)


aeisenberg (approved):

extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
```diff
@@ -144,22 +144,67 @@ function highlightCodeLines(
...
 
-function buildMarkdownAlertMessage(interpretedResult: AnalysisAlert): string {
+function buildMarkdownAlertMessage(
+  interpretedResult: AnalysisAlert,
+  language: string
+): string[] {
+  const hasPathResults = interpretedResult.codeFlows.length > 0;
+  if (hasPathResults) {
+    // For path-problem queries, the "alert message" is an expandable section containing the path results.
+    return buildMarkdownPathResults(interpretedResult, language);
```
charisk, line 154: 🤔 I'm a little surprised by this - do we not want to expand message tokens for path-problem queries?

extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
```diff
@@ -144,22 +144,67 @@ function highlightCodeLines(
   return `${partiallyHighlightedLine.plainSection1}<strong>${partiallyHighlightedLine.highlightedSection}</strong>${partiallyHighlightedLine.plainSection2}`;
 }
 
-function buildMarkdownAlertMessage(interpretedResult: AnalysisAlert): string {
+function buildMarkdownAlertMessage(
```
charisk, line 147: What do you think about appending a `For` to these new `buildMarkdown` functions e.g. `buildMarkdownForAlertMessage`? Or, to match other function names in this module, you could call them `generateMarkdownForXXX`

extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
```diff
@@ -144,22 +144,67 @@ function highlightCodeLines(
...
-        token.text,
+    alertMessage += token.text;
+  }
+  const pathLines: MarkdownFile = [];
+  pathLines.push('#### Paths', '');
+  for (const codeFlow of interpretedResult.codeFlows) {
+    const stepCount = codeFlow.threadFlows.length;
+    pathLines.push(`Path with ${stepCount} steps`);
+    let index = 1;
+    for (const threadFlow of codeFlow.threadFlows) {
```
charisk, line 190: You could use a "traditional" for loop here instead to avoid defining and index and incrementing it. E.g.

```ts
for (let i = 0; i < codeFlow.threadFlows.length; i++) {
  const threadFlow = codeFlow.threadFlows[i];
  ...
}
```

extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
```diff
@@ -89,8 +89,8 @@ function generateMarkdownForInterpretedResult(interpretedResult: AnalysisAlert,
       ...generateMarkdownForCodeSnippet(codeSnippet, language, highlightedRegion),
     );
   }
-  const alertMessage = buildMarkdownAlertMessage(interpretedResult);
-  lines.push(alertMessage);
+  const alertMessageLines = buildMarkdownAlertMessage(interpretedResult, language);
```
charisk, line 92: At first I found it a little confusing that the code paths stuff is shown as part of the "message" - they are 2 different things, it just happens that the UX bundles them together - you get an expandable message if there are code paths to show.

This is not necessarily a good suggestion or one that should be implemented, but I wonder if the code would be simpler/more readable if we first build some markdown for code paths (if there are any) and then hand it to the `buildMarkdownAlertMessage` that knows whether to build a normal message or an expandable section based on that.

It might be worth trying that out. Or not! See what you think 😸

charisk (commented):
Generally looking good - I have some questions/comments but if you need to get the PR in quickly, don't let me stop you.

extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
shati-patel: This is another limitation of markdown 😢 As far as I'm aware, you can't have a clickable link (e.g. to an alert location) in the title of an expandable section, so it has to be plain-text for the path-problem case 🤷🏽 

(Maybe we can play around with the design a bit more though - not sure what the best "look" is 😅 )

charisk: Oh interesting, thanks for the explanation! I think the current approach could lead to messages looking a bit weird (containing `[1]` etc.) so perhaps we should come up with some other "UX". How about having a "Show paths" section that is expandable then instead of using the message? A bit like we do with UI rendering of results.

extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
shati-patel: Nice, thank you!

extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
shati-patel: I've (hopefully!) tidied this up a bit: The alert message is always rendered the same, but if there are paths we show an expandable "Show paths" section ([preview](https://github.com/github/vscode-codeql/blob/shati-patel/sharing-path-results/extensions/ql-vscode/test/pure-tests/remote-queries/markdown-generation/data/interpreted-results/path-problem/results-repo1.md))

extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
shati-patel: Done!

shati-patel:
> Generally looking good - I have some questions/comments but if you need to get the PR in quickly, don't let me stop you.

Thanks for the review! Tidied up the path rendering a bit and added an example that actually has multiple paths 😄 

(I see there's another suggestion in the internal issue to collapse each individual path—that might be an idea for a follow-up PR, but I think this is probably okay for now!)

extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
```diff
@@ -152,14 +160,42 @@ function buildMarkdownAlertMessage(interpretedResult: AnalysisAlert): string {
...
 
+function generateMarkdownForPathResults(
+  interpretedResult: AnalysisAlert,
+  language: string
+): MarkdownFile {
+  const pathLines: MarkdownFile = [];
+  for (const codeFlow of interpretedResult.codeFlows) {
+    const stepCount = codeFlow.threadFlows.length;
+    pathLines.push(`#### Path with ${stepCount} steps`);
+    for (let i = 0; i < codeFlow.threadFlows.length; i++) {
```
charisk, line 179: Feel free to ignore this but you could use the `stepCount` here
```suggestion
    for (let i = 0; i < stepCount; i++) {
```

charisk (approved):
This looks great, thank you for making the changes!

Agree that we could make further adjustments later on depending on Product/UX feedback.

PR #1326, opened 2022-04-29
Upgrade node version minimum requirement

charisk:
I believe we're expecting use of v15 and above so I've updated the engine definition.

## Checklist
N/A:


shati-patel (approved):
Thanks, makes sense! 🚢

edoardopirovano:
Should we not also be checking in an updated `package-lock.json`?

aeisenberg (commented):
It's a little strange that we have `"@types/node": "^16.11.25"` in dev dependencies, but the minimum node version is 15. Can we synchronize the two? And also, the package lock should be updated as @edoardopirovano mentions.

aeisenberg:
We also need to make sure that we run the internal integration tests using the same version of node. This is something I can take care of.

PR #1330, opened 2022-05-04
MRVA: Include filename when generating markdown files

shati-patel:
This PR updates the `MarkdownFile` type to include the filename as well as the file contents. This is so we can keep track of the filenames more consistently and link to them correctly, e.g. when uploading as a gist. _(Originally suggested in https://github.com/github/vscode-codeql/pull/1288#discussion_r847672799.)_

Note: there's no behaviour change here yet, I've just updated the existing functions/tests to work with the updated type. I will tweak the tests (to check that file name is expected) in a follow-up PR.

## Checklist

N/A - internal only 🦈

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #1335, opened 2022-05-05
MRVA: Also test names of generated markdown files

shati-patel:
Quick follow-up to #1330. I've updated the markdown test data to use the expected file names and I've updated the tests to actually check that the generated names are correct 🦒 

## Checklist

N/A - internal only 🍉


extensions/ql-vscode/test/pure-tests/remote-queries/markdown-generation/markdown-generation.test.ts
```diff
@@ -3,6 +3,8 @@ import * as path from 'path';
 import * as fs from 'fs-extra';
 import { generateMarkdown } from '../../../../src/remote-queries/remote-queries-markdown-generation';
 
+const expectedFileNames = ['summary', 'github-codeql', 'meteor-meteor'];
```
charisk, line 6: This introduces knowledge about the test data outside of the test data dir and into the source code. It works, but it's not ideal IMO.

Instead of keeping the list of expected file names and then looping over them etc. we could move all the files we expect into a subdir e.g. `expected` and then look at all the contents of the dir and iterating over those.

charisk (approved):
Thanks for making this part of the tests!

I have a suggestion that I'd like us to implement but I don't want to block your progress on delivering the feature so I'll approve. Perhaps can be done in a future PR?

extensions/ql-vscode/test/pure-tests/remote-queries/markdown-generation/markdown-generation.test.ts
shati-patel: Done in latest commits! Thanks for helping with this 🍐 🎉

charisk (approved):
🚢

PR #1340, opened 2022-05-09
Fix cli-integration tests

aeisenberg:
The main problem this commit fixes is with vscode 1.67.0, an error is
thrown when inside of integration tests and a dialog box is opened. We
were opening the telemetry dialog box. Now, an env variable is set
during cli-integration tests that prevents the dialog from being
opened.

There are also other cleanups and improvements with cli-integration
tests that assist with running locally:

- `vscode-test` dependency has been renamed to `@vscode/test-electron`,
  so use that instead and make the small API changes to support it.
- Commit the codeql-pack.lock.yml file so it isn't recreated on each
  test run.
- Ensure all databases are removed before _and after_ each test run
  that manipulates the set of installed databases
- Similarly, for quick query files, delete them before and after each
  test.
- Change some async `forEach` blocks to for loops in order to support
  sequential operations more easily.

## Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,7 +12,6 @@
         // Add a reference to a workspace to open. Eg-
         // "${workspaceRoot}/../vscode-codeql-starter/vscode-codeql-starter.code-workspace"
       ],
-      "stopOnEntry": false,
       "sourceMaps": true,
       "outFiles": [
         "${workspaceRoot}/extensions/ql-vscode/out/**/*.js",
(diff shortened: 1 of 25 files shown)


aeisenberg:
I had to revert back to `vscode-test` from `@vscode/test-electron`. Even though the former is deprecated, the latter was preventing the integration tests from passing and I couldn't figure out what the error is.

The failing _Dependency Review_ job is mostly out of our control. It is failing due to a dependency on `glob-parent`, which is vulnerable, but only used in building and tests. There is no way to update that dependency without downgrading(!) `gulp`. And downgrading gulp causes other (minor) security errors to fail. All of these potential vulnerabilities are in test/dev code and is not shipped to users.

cklin (approved):

PR #1341, opened 2022-05-11
MRVA: Add command to export markdown results to gist

shati-patel:
Adds a canary command "Export Variant Analysis Results" that lets you create a gist from the results in the currently-open MRVA webview:

demo of exporting variant analysis to Gist

We will also have an option to create a local markdown file with results. I'll leave that for a separate PR, to avoid this getting too big!

(See internal issue for more details.)

## Checklist

N/A - internal only 🐣

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
@@ -5,14 +5,14 @@ import * as fs from 'fs-extra';
 
 import { Credentials } from '../authentication';
 import { CodeQLCliServer } from '../cli';
-import { ProgressCallback } from '../commandRunner';
+import { ProgressCallback, UserCancellationException } from '../commandRunner';
 import { createTimestampFile, showAndLogErrorMessage, showAndLogInformationMessage, showInformationMessageWithAction } from '../helpers';
 import { Logger } from '../logging';
 import { runRemoteQuery } from './run-remote-query';
 import { RemoteQueriesInterfaceManager } from './remote-queries-interface';
 import { RemoteQuery } from './remote-query';
 import { RemoteQueriesMonitor } from './remote-queries-monitor';
-import { getRemoteQueryIndex } from './gh-actions-api-client';
+import { createGist, getRemoteQueryIndex } from './gh-actions-api-client';
 import { RemoteQueryResultIndex } from './remote-query-result-index';
 import { RemoteQueryResult } from './remote-query-result';
 import { DownloadLink } from './download-link';
@@ -23,6 +23,7 @@ import { QueryHistoryManager } from '../query-history';
 import { QueryStatus } from '../query-status';
 import { DisposableObject } from '../pure/disposable-object';
 import { QueryHistoryInfo } from '../query-results';
+import { generateMarkdown } from './remote-queries-markdown-generation';
 
 const autoDownloadMaxSize = 300 * 1024;
 const autoDownloadMaxCount = 100;
@@ -38,7 +39,7 @@ export class RemoteQueriesManager extends DisposableObject {
     private readonly cliServer: CodeQLCliServer,
     private readonly qhm: QueryHistoryManager,
     private readonly storagePath: string,
-    logger: Logger,
+    private readonly logger: Logger,
   ) {
     super();
     this.analysesResultsManager = new AnalysesResultsManager(ctx, cliServer, storagePath, logger);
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -301,4 +302,59 @@ export class RemoteQueriesManager extends DisposableObject {
...
+
+    if (!queryId || !queryHistoryItem || !queryHistoryItem.completed || queryHistoryItem.t !== 'remote') {
+      throw new Error('No variant analysis results currently open. To open results, click an item in the query history view.');
+    }
+
+    void this.logger.log(`Exporting variant analysis results for query: ${queryId}`);
+    const query = queryHistoryItem.remoteQuery;
+    const analysesResults = this.analysesResultsManager.getAnalysesResults(queryId);
+
+    const credentials = await Credentials.initialize(this.ctx);
```
charisk, line 318: Shall we move this closer to its usage? We don't need to get credentials if we're going to just create local files.

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -301,4 +302,59 @@ export class RemoteQueriesManager extends DisposableObject {
...
+      const markdownFiles = generateMarkdown(query, analysesResults, 'gist');
+
+      // Convert markdownFiles to the appropriate format for uploading to gist
+      const gistFiles = markdownFiles.reduce((acc, cur) => {
+        acc[`${cur.fileName}.md`] = { content: cur.content.join('\n') };
+        return acc;
+      }, {} as { [key: string]: { content: string } });
+
+      const gistUrl = await createGist(credentials, description, gistFiles);
+      void showAndLogInformationMessage(`Variant analysis results exported to [gist](${gistUrl}).`);
```
charisk, line 353: This will show a "Show log" button. I don't think we want that. Shall we have an "Open gist" action instead that opens the URL? (assuming that's not too hard)

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -301,4 +302,59 @@ export class RemoteQueriesManager extends DisposableObject {
       queryItem.status = QueryStatus.Failed;
     }
   }
+
+  public async exportVariantAnalysisResults(): Promise<void> {
```
charisk, line 306: A suggestion (that is definitely not necessary) around how this bit of code could be made a bit cleaner (and btw I think what you have is fine):
- Move the `exportVariantAnalysisResults` function to another file/module and pass in any dependencies (e.g. `logger`, `qhm`). This can be done as a class that receives the dependencies in the constructor (like we do in the `RemoteQueriesManager`), or it could be simply be a function within a `results-sharing.ts` module. 
- Once we're in a new file, it'd be easier to break down into smaller ones a bit. E.g. The part around selecting which mode to use (gist or local) can be in one function, and the bits around generating gists or local files could be in separate functions.
 
Totally up to you if you want to do any of that. I just thought it might be useful to write down.


Also while writing this I realise that we've generally used the term remote query rather than variant analysis in non user-facing stuff. I'd personally try and stick to that to have some sort of consistency (although I appreciate it's by definition inconsistent 😬 )

charisk (commented):
Looking really good! Some comments + I'll come back to it later.

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
shati-patel: Done! Pop-up window now looks like this 
image

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -301,4 +302,65 @@ export class RemoteQueriesManager extends DisposableObject {
...
+    const queryHistoryItem = this.qhm.getCurrentQueryHistoryItem();
+
+    if (!queryHistoryItem || queryHistoryItem.t !== 'remote') {
+      throw new Error('No variant analysis results currently open. To open results, click an item in the query history view.');
+    } else if (!queryHistoryItem.completed) {
+      throw new Error('Variant analysis results are not yet available. Wait for results to be downloaded.');
```
charisk, line 312: I'd skip the last sentence (it's not quite true - the problem is that the actions workflow hasn't finished yet).

```suggestion
      throw new Error('Variant analysis results are not yet available.');
```

charisk (approved):
Let's definitely add some tests in a separate PR!

PR #1344, opened 2022-05-12
MRVA: Export results to local markdown files

shati-patel:
Expands on the exporting functionality from #1341 to allow exporting to local markdown instead of gist. A pop-up gives you the location of the saved markdown, and lets you open the summary file. (You can also find the markdown files later, via the "Open query directory" command in the query history.)

Here's a little GIF:

demo of exporting variant analysis to local markdown

(See internal issue for more info 🔍 )

## Checklist

N/A - internal only 🦔

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/export-results.ts
+++ b/extensions/ql-vscode/src/remote-queries/export-results.ts
@@ -102,7 +102,7 @@ async function exportResultsToGist(
 
 /**
  * Converts the results of a remote query to markdown and saves the files locally
- * in the query directory (i.e. the workspace directory where query results and logs are saved).
+ * in the query directory (where query results and metadata are also saved).
  */
 async function exportResultsToLocalMarkdown(
   queryDirectoryPath: string,
--- a/extensions/ql-vscode/src/remote-queries/export-results.ts
+++ b/extensions/ql-vscode/src/remote-queries/export-results.ts
@@ -1,0 +1,97 @@
+import { window, commands, Uri, ExtensionContext, QuickPickItem } from 'vscode';
+import { Credentials } from '../authentication';
+import { UserCancellationException } from '../commandRunner';
+import { showInformationMessageWithAction, showAndLogInformationMessage } from '../helpers';
+import { logger } from '../logging';
+import { QueryHistoryManager } from '../query-history';
+import { createGist } from './gh-actions-api-client';
+import { RemoteQueriesManager } from './remote-queries-manager';
+import { generateMarkdown } from './remote-queries-markdown-generation';
+import { RemoteQuery } from './remote-query';
+import { AnalysisResults } from './shared/analysis-result';
+
+/**
+ * Exports the results of the currently-selected remote query.
+ * The user is prompted to select the export format.
+ */
+export async function exportRemoteQueryResults(
+  queryHistoryManager: QueryHistoryManager,
+  remoteQueriesManager: RemoteQueriesManager,
+  ctx: ExtensionContext,
+): Promise<void> {
+  const queryHistoryItem = queryHistoryManager.getCurrentQueryHistoryItem();
+  if (!queryHistoryItem || queryHistoryItem.t !== 'remote') {
+    throw new Error('No variant analysis results currently open. To open results, click an item in the query history view.');
+  } else if (!queryHistoryItem.completed) {
+    throw new Error('Variant analysis results are not yet available.');
+  }
+  const queryId = queryHistoryItem.queryId;
+  void logger.log(`Exporting variant analysis results for query: ${queryId}`);
+  const query = queryHistoryItem.remoteQuery;
+  const analysesResults = remoteQueriesManager.getAnalysesResults(queryId);
+
+  const gistOption = {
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/remote-queries/export-results.ts
```diff
@@ -95,3 +97,32 @@ async function exportResultsToGist(
...
   }
 }
+
+/**
+ * Converts the results of a remote query to markdown and saves the files locally
+ * in the query directory (i.e. the workspace directory where query results and logs are saved).
+ */
+async function exportResultsToLocalMarkdown(
+  queryHistoryManager: QueryHistoryManager,
+  queryHistoryItem: QueryHistoryInfo,
```
aeisenberg, line 107: minor: I see you are passing in two parameters here, `queryHistoryManager` and `queryHistoryItem`, only being used to get the `basePath`. It might be simpler to get the `basePath` outside of this function and pass it in as a parameter.

It cuts down on the # of parameters to this function by one, avoids an import

extensions/ql-vscode/src/remote-queries/export-results.ts
```diff
@@ -95,3 +97,32 @@ async function exportResultsToGist(
...
+  query: RemoteQuery,
+  analysesResults: AnalysisResults[]
+) {
+  const markdownFiles = generateMarkdown(query, analysesResults, 'local');
+  const basePath = await queryHistoryManager.getQueryHistoryItemDirectory(
+    queryHistoryItem
+  );
+  for (const markdownFile of markdownFiles) {
+    const filePath = path.join(basePath, `${markdownFile.fileName}.md`);
+    await fs.writeFile(filePath, markdownFile.content.join('\n'), 'utf8');
```
charisk, line 117: Have you tested whether this works if you export multiple times? I think it should, but can't be bothered checking the docs 😬

extensions/ql-vscode/src/remote-queries/export-results.ts
```diff
@@ -95,3 +97,32 @@ async function exportResultsToGist(
...
+  }
+  const shouldOpenSummary = await showInformationMessageWithAction(
+    `Variant analysis results exported to \"${basePath}\".`,
+    'Open exported results'
+  );
+  if (shouldOpenSummary) {
```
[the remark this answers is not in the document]
shati-patel: Hmm, that's a good point. You can currently access the results files in the "query directory" from here:
image

That isn't at all obvious though 😅 Perhaps we could put all the results in a "exported results" subdirectory of the query directory, and then open that directory instead of just opening the summary file?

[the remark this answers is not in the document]
aeisenberg: I think that would work. Could you do both? ie- open the markdown file as an editor as well as open the exported results directory? I think it's nice to see the prettified markdown version of the results.

extensions/ql-vscode/src/remote-queries/export-results.ts
aeisenberg: Oh, right...one thing we're going to need to think about is exporting multiple times with different selections. Presumably, we are going to want to delete the old results before creating new ones. Is this something we can implement later?

charisk: Yeah I think we can worry about that a bit later on!

I specifically meant running the export multiple times e.g. maybe I forgot I already exported stuff and I tried it again.

shati-patel: Exporting the same thing multiple times works fine!

shati-patel: I guess we technically overwrite the markdown files every time though 🤔 Would that be a problem? Perhaps we should check if the files exist first?

aeisenberg: > Exporting the same thing multiple times works fine!

My concern is about exporting a slightly different selection each time. Since selection isn't implemented yet, we don't need to worry about it until later, but this is something we should keep in mind.

extensions/ql-vscode/src/remote-queries/export-results.ts
```diff
@@ -100,29 +102,28 @@ async function exportResultsToGist(
...
 ) {
   const markdownFiles = generateMarkdown(query, analysesResults, 'local');
-  const basePath = await queryHistoryManager.getQueryHistoryItemDirectory(
-    queryHistoryItem
-  );
+  const exportedResultsPath = path.join(queryDirectoryPath, 'exported-results');
+  await fs.ensureDir(exportedResultsPath);
   for (const markdownFile of markdownFiles) {
-    const filePath = path.join(basePath, `${markdownFile.fileName}.md`);
+    const filePath = path.join(exportedResultsPath, `${markdownFile.fileName}.md`);
```
aeisenberg, line 116: Nice...so now all markdown files will live in a single directory. This will make it easier to manage and delete across exports.

aeisenberg (approved):

charisk (approved):

PR #1345, opened 2022-05-13
MRVA: Add webview button to export results

shati-patel:
Adds a button to the webview that calls the "Export Results" command (from #1341). This makes the export functionality more discoverable:
image

I couldn't find an exact match for a Primer component (https://primer.style/react/Button wasn't quite right), so I copied the VS Code style. The button doesn't look _exactly_ like our design (padding/font look bit off), but it's close? 🤞🏽 😅 

## Checklist

N/A - internal only 🤫 

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
@@ -4,7 +4,8 @@ import {
   window as Window,
   ViewColumn,
   Uri,
-  workspace
+  workspace,
+  commands
 } from 'vscode';
 import * as path from 'path';
 
@@ -209,6 +210,9 @@ export class RemoteQueriesInterfaceManager {
         break;
       case 'remoteQueryDownloadAllAnalysesResults':
         await this.downloadAllAnalysesResults(msg);
+        break;
+      case 'remoteQueryExportResults':
+        await await commands.executeCommand('codeQL.exportVariantAnalysisResults');
         break;
       default:
         assertNever(msg);
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
@@ -212,7 +212,7 @@ export class RemoteQueriesInterfaceManager {
         await this.downloadAllAnalysesResults(msg);
         break;
       case 'remoteQueryExportResults':
-        await await commands.executeCommand('codeQL.exportVariantAnalysisResults');
+        await commands.executeCommand('codeQL.exportVariantAnalysisResults');
         break;
       default:
         assertNever(msg);
--- a/extensions/ql-vscode/src/remote-queries/view/ExportButton.tsx
+++ b/extensions/ql-vscode/src/remote-queries/view/ExportButton.tsx
@@ -1,0 +1,22 @@
+import * as React from 'react';
+import styled from 'styled-components';
+
+const Button = styled.a`
+  color: var(--vscode-button-foreground);
+  background-color: var(--vscode-button-background);
+  &:hover {
+    color: var(--vscode-button-foreground);
+    text-decoration: none;
+    background-color: var(--vscode-button-hoverBackground);
+  }
+  cursor: pointer;
+  padding: 5px 10px;
+`;
+
+const ExportButton = ({ text, onClick }: { text: string, onClick: () => void }) => (
+  <Button className="monaco-button monaco-text-button" onClick={onClick}>
+    {text}
+  </Button>
+);
+
+export default ExportButton;
--- a/extensions/ql-vscode/src/remote-queries/view/ExportButton.tsx
+++ b/extensions/ql-vscode/src/remote-queries/view/ExportButton.tsx
@@ -1,20 +1,20 @@
 import * as React from 'react';
 import styled from 'styled-components';
 
-const Button = styled.a`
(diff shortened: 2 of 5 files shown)


extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
```diff
@@ -210,6 +211,9 @@ export class RemoteQueriesInterfaceManager {
       case 'remoteQueryDownloadAllAnalysesResults':
         await this.downloadAllAnalysesResults(msg);
         break;
+      case 'remoteQueryExportResults':
+        await await commands.executeCommand('codeQL.exportVariantAnalysisResults');
```
aeisenberg, line 215: Remove extra extra word.

```suggestion
        await commands.executeCommand('codeQL.exportVariantAnalysisResults');
```

extensions/ql-vscode/src/remote-queries/view/ExportButton.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import styled from 'styled-components';
+
+const Button = styled.a`
+  color: var(--vscode-button-foreground);
+  background-color: var(--vscode-button-background);
+  &:hover {
+    color: var(--vscode-button-foreground);
```
aeisenberg, line 8: Is this right? It's the same class as non-hover mode. Did you mean this?

```suggestion
    color: var(--vscode-button-hoverForeground);
```

extensions/ql-vscode/src/remote-queries/view/ExportButton.tsx
```diff
@@ -0,0 +1,22 @@
...
+    text-decoration: none;
+    background-color: var(--vscode-button-hoverBackground);
+  }
+  cursor: pointer;
+  padding: 5px 10px;
+`;
+
+const ExportButton = ({ text, onClick }: { text: string, onClick: () => void }) => (
```
aeisenberg, line 16: Minor: this isn't really an _export_ button. What you've made here is something more general. Maybe rename to something like `ActionButton` or `StyledButton` or ...

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -313,9 +320,16 @@ const AnalysesResults = ({
...
-        totalAnalysesResults={totalAnalysesResults}
-        totalResults={totalResults} />
+      <Box display="flex">
+        <Box flexGrow={1}>
+          <AnalysesResultsTitle
+            totalAnalysesResults={totalAnalysesResults}
+            totalResults={totalResults} />
+        </Box>
+        <Box>
+          <ExportButton text="Export all" onClick={() => exportResults()}></ExportButton>
```
aeisenberg, line 330: Embarrasingly trivial: This is more succinct, though either way is fine.

```suggestion
          <ExportButton text="Export all" onClick={exportResults}></ExportButton>
```

aeisenberg:
Also, I think it's fine based on the css you have,  but please verify that this looks reasonable given different styles.

extensions/ql-vscode/src/remote-queries/view/ExportButton.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import styled from 'styled-components';
+
+const Button = styled.a`
```
charisk, line 4: If you make the following changes I think you get the same result but it hopefully makes a bit more sense:
- Use a `<button>` instead of `<a>` (`styled.button...`)
- Set `border: 0;`
- Remove `className="monaco-button monaco-text-button"` from L17

charisk (commented):
Just adding a suggestion to Andrew's comments but wanted to also said great job making a new re-usable component!

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
shati-patel: Not too trivial at all 🙈 I'm always happy to see neater ways of writing things!

extensions/ql-vscode/src/remote-queries/view/ExportButton.tsx
shati-patel: Thank you! That worked 🎉

extensions/ql-vscode/src/remote-queries/view/ExportButton.tsx
shati-patel: Good point! The button's only used for exporting at the moment, so I'll keep the name for now (for the sake of easily remembering where it's used 🔍 ). But we could rename it in future and use it more generally! 💭

shati-patel:
Thank you for the reviews! Can confirm the button looks sensible in other themes too 🎨

extensions/ql-vscode/src/remote-queries/view/ExportButton.tsx
charisk: Just my 2p.. there is value in renaming now. Think of a new user coming to our codebase. They're making some UI changes that include adding a button that executes an action. They look the current components and they see "ExportButton". At that point they might think that it's not what they want and create their own component, as what they want has nothing to do with exporting. `ActionButton` is more likely to be considered in such case.

shati-patel: Makes sense, I will be less lazy and do it now! 😂

PR #1350, opened 2022-05-16
Show query results before structured evaluator log summary completes

angelapwen:
Previously, we were `await`ing the structured evaluator log summary to be fully generated before the query result and raw structured log could be viewed, even though they were already available.

This change allows the structured evaluator log summary to execute asynchronously. If the user attempts to click the "Show Structured Evaluator Log (Summary)" command on the query history item before the summary is complete, the user receives a popup saying "The evaluator log summary is still being generated. Please try again later. The summary generation process is tracked in the "CodeQL Extension Log" view." — 

Screen Shot 2022-05-16 at 4 41 16 PM

The CodeQL Extension Log view will indicate the summary generation is in-progress and when it is complete. It will also log the same warning line as above every time the user attempts to view the summary file before it is completed (in the following screenshot the command was clicked twice):
Screen Shot 2022-05-16 at 4 45 41 PM

and at this point the user will be able to view the log summary by clicking the query history command. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -212,7 +212,7 @@ export class QueryEvaluationInfo {
             })
 
             .catch(err => {
-              void logger.log(`Failed to generate structured evaluator log summary. Reason: ${err.message}`);
+              void showAndLogWarningMessage(`Failed to generate structured evaluator log summary. Reason: ${err.message}`);
             });
         } else {
           void showAndLogWarningMessage(`Failed to write structured evaluator log to ${this.evalLogPath}.`);
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -199,15 +199,21 @@ export class QueryEvaluationInfo {
         });
         if (await this.hasEvalLog()) {
           queryInfo.evalLogLocation = this.evalLogPath;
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/run-queries.ts
```diff
@@ -199,15 +199,21 @@ export class QueryEvaluationInfo {
...
+                if (err) {
+                 throw new Error(`Could not read structured evaluator log end of summary file at ${this.evalLogEndSummaryPath}.`);
+                }
+                void qs.logger.log(' --- Evaluator Log Summary --- ');
+                void qs.logger.log(buffer.toString());
+              });
+            })
+
+            .catch(err => {
+              void logger.log(`Failed to generate structured evaluator log summary. Reason: ${err.message}`);
```
edoardopirovano, line 215: Should we use `showAndLogWarningMessage` here so that the error is displayed in a pop-up as well as being logged?

angelapwen: Hm, I think because all the other related popups in this realm are warning messages I'll keep it as a warning message.

edoardopirovano: I think I'm inclined to agree that everything in this code should be a warning since failures here aren't fatal in the sense they don't result in a failure to produce results for the query which is what most users care about.

edoardopirovano (approved):
Seems reasonable to me, and a surprisingly simple fix! Thanks for picking this up quickly 🙂

PR #1353, opened 2022-05-18
Add sorting to variant analysis results

aeisenberg:
Sort by stars, number of results, and name.

This also includes a graphql query that retrieves all the stars
for relevant repositories.

Here is an example of what it looks like:

sorting

This PR also includes an integration test that ensures the graphql call to get the stargazers counts for each repo works.  I had to make a change to the credentials class to allow a way of instantiating based on an environment variable.

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
@@ -12,7 +12,7 @@ import { runRemoteQuery } from './run-remote-query';
 import { RemoteQueriesInterfaceManager } from './remote-queries-interface';
 import { RemoteQuery } from './remote-query';
 import { RemoteQueriesMonitor } from './remote-queries-monitor';
-import { getRemoteQueryIndex } from './gh-actions-api-client';
+import { getRemoteQueryIndex, getStargazers } from './gh-actions-api-client';
 import { RemoteQueryResultIndex } from './remote-query-result-index';
 import { RemoteQueryResult } from './remote-query-result';
 import { DownloadLink } from './download-link';
@@ -182,7 +182,6 @@ export class RemoteQueriesManager extends DisposableObject {
   }
 
   private mapQueryResult(executionEndTime: number, resultIndex: RemoteQueryResultIndex, queryId: string): RemoteQueryResult {
-
     const analysisSummaries = resultIndex.successes.map(item => ({
       nwo: item.nwo,
       databaseSha: item.sha || 'HEAD',
@@ -281,6 +280,8 @@ export class RemoteQueriesManager extends DisposableObject {
       queryItem.failureReason = undefined;
       const queryResult = this.mapQueryResult(executionEndTime, resultIndex, queryItem.queryId);
 
+      await this.addStargazers(resultIndex, credentials, queryResult);
+
       await this.storeJsonFile(queryItem, 'query-result.json', queryResult);
 
       // Kick off auto-download of results in the background.
@@ -303,7 +304,13 @@ export class RemoteQueriesManager extends DisposableObject {
     }
   }
 
-  // Pulled from the analysis results manager, so that we can get access to 
+  private async addStargazers(resultIndex: RemoteQueryResultIndex, credentials: Credentials, queryResult: RemoteQueryResult) {
+    const nwos = resultIndex.successes.map(s => s.nwo);
+    const stargazers = await getStargazers(credentials, nwos);
+    queryResult.analysisSummaries.forEach(analysis => analysis.starCount = stargazers[analysis.nwo]);
+  }
+
+  // Pulled from the analysis results manager, so that we can get access to
   // analyses results from the "export results" command.
   public getAnalysesResults(queryId: string): AnalysisResults[] {
     return [...this.analysesResultsManager.getAnalysesResults(queryId)];
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
@@ -181,7 +181,12 @@ export class RemoteQueriesManager extends DisposableObject {
       results => this.interfaceManager.setAnalysisResults(results, queryResult.queryId));
   }
 
-  private mapQueryResult(executionEndTime: number, resultIndex: RemoteQueryResultIndex, queryId: string): RemoteQueryResult {
+  private mapQueryResult(
+    executionEndTime: number,
+    resultIndex: RemoteQueryResultIndex,
+    queryId: string,
+    stargazers: Record<string, number>
+  ): RemoteQueryResult {
     const analysisSummaries = resultIndex.successes.map(item => ({
       nwo: item.nwo,
       databaseSha: item.sha || 'HEAD',
@@ -192,6 +197,7 @@ export class RemoteQueriesManager extends DisposableObject {
         urlPath: `${resultIndex.artifactsUrlPath}/${item.artifactId}`,
         innerFilePath: item.sarifFileSize ? 'results.sarif' : 'results.bqrs',
         queryId,
+        starCount: stargazers[item.nwo]
       } as DownloadLink
     }));
     const analysisFailures = resultIndex.failures.map(item => ({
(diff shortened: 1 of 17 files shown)


extensions/ql-vscode/src/remote-queries/view/StarCount.tsx
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react';
+import { StarIcon } from '@primer/octicons-react';
+
+type Props = { starCount?: number };
+
+const StarCount = ({ starCount }: Props) => (
+  Number.isFinite(starCount) ? (
+    <>
+      <span className="vscode-codeql__analysis-star">
```
charisk, line 9: How come you're using CSS classes here? My aim so far with these components is that they would be re-usable and self-contained. We can use StyledComponents and inline styles.

 E.g. 

```ts
const Star = styled.span`
  ...
`;

const Count = styled.span`
 ...
`;
```

extensions/ql-vscode/src/remote-queries/view/StarCount.tsx
```diff
@@ -0,0 +1,38 @@
...
+        <StarIcon size={16} />
+      </span>
+      <span className='vscode-codeql__analysis-count'>
+        {displayStars(starCount!)}
+      </span>
+    </>
+  ) : (
+    <>
```
charisk, line 17: It feels a bit wrong to have empty spans. Can we use a `Box` component to get the layout we want?

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import { FilterIcon } from '@primer/octicons-react';
+import { ActionList, ActionMenu } from '@primer/react';
+
+export type Sort = 'name' | 'stars' | 'results';
+type SortBy = { name: string, sort: Sort }[];
```
charisk, line 6: Shall we drop the array part here? I think it hurts readability.

```suggestion
type SortBy = { name: string, sort: Sort };
```

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
```diff
@@ -0,0 +1,68 @@
...
+type Props = {
```
charisk, line 7: Personal preference (feel free to ignore) but I've generally avoided having a `Props` type because it makes you jump around in the code - I just define it inline where the component that has the props is defined.

This helps if you have multiple components in the same file too - avoids the reader having to wonder for which component the props are for.

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
```diff
@@ -0,0 +1,68 @@
...
+  setSort: (sort: Sort) => void;
+};
+
+type Sortable = {
+  nwo: string;
+  starCount?: number;
+  resultCount?: number;
+};
+
+const sortBy: SortBy = [
```
charisk, line 18: We have a type and a value both called "sort by", which I found a bit difficult to follow. Perhaps `sorts` is a better name for this one? Or `sortByOptions`?

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
```diff
@@ -0,0 +1,68 @@
...
+  { name: 'Sort by Name', sort: 'name' },
+  { name: 'Sort by Results', sort: 'results' },
+  { name: 'Sort by Stars', sort: 'stars' },
+];
+
+export function sorter(sort: Sort) {
```
charisk, line 24: Let's add a return type - make it clearer that this returns a function.

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
```diff
@@ -0,0 +1,68 @@
...
+      }
+    }
+
+    // Fall back on name compare if results or stars are equal
+    return left.nwo.localeCompare(right.nwo, undefined, { sensitivity: 'base' });
+  };
+}
+
+const SortRepoFilter = ({ sort, setSort }: Props) => {
+  return <span className="vscode-codeql__analysis-sorter">
```
charisk, line 47: Similar comment with `StarCount` - why are we using CSS classes?

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <ActionMenu>
```
charisk, line 48: This doesn't look right locally for me for some reason 🤔 

image
image
image

charisk (commented):
Some comments for now, I'll come back to it tomorrow!

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
charisk: Actually this looked better after running `npm run build`. I'm not sure why, I thought `npm run watch` would refresh everything when it first runs.

There are some styling issues still though. 
- The colours are a bit off in light mode
- The menu opens up to the right - it should open to the left according to the designs
image

aeisenberg:
Thanks for the review!

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -281,6 +280,8 @@ export class RemoteQueriesManager extends DisposableObject {
       queryItem.failureReason = undefined;
       const queryResult = this.mapQueryResult(executionEndTime, resultIndex, queryItem.queryId);
 
+      await this.addStargazers(resultIndex, credentials, queryResult);
```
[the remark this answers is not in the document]
aeisenberg: Yep...that's cleaner.

extensions/ql-vscode/src/remote-queries/view/StarCount.tsx
aeisenberg: It was the easiest way to ensure the layouts in columns. I'll see if I can use `Box` to make it cleaner.

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
aeisenberg: I agree, that's confusing. I'll change that.

aeisenberg: This type is actually not needed. It's gone.

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
aeisenberg: It's the capitalization that should help distinguish a type from a value. But, anyway, the type is gone now. I think it's more understandable now.

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
aeisenberg: I will change.

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
aeisenberg: Don't know if i have control over where the drop-down comes. I'll try to fix.

I'll also work on the background color.

aeisenberg: The menu open location is dependent on the width of the viewport. With a smaller screen, the menu will open on the other side. It's the default behaviour of this component. Do you think this is something to spend time on?

aeisenberg: And another thing that's odd is that on my screen, I'm getting the reverse colours. I wonder if it has something to do with what theme the IDE is using when you initially run compile??

_Extension_Development_Host__-_CodeQL_Query_Results_—_vscode-codeql-starter__Workspace_

Odd. I'll have to look tomorrow.

aeisenberg:
I addressed all of your comments except for the theme styling. Something odd is happening and I'll need to look tomorrow.

extensions/ql-vscode/src/remote-queries/view/SortRepoFilter.tsx
charisk: Discussed offline - we'll use the approach we have in the CodePaths component and work on our Primer theming solution.

charisk (approved):
As discussed, there are some comments to be addressed but most will be done before merging and a separate PR will address light theme issues.

aeisenberg:
Will work on theming and styling next.

PR #1365, opened 2022-05-26
Extract time functions

aeisenberg:
Create the `time.ts` module as a place to put fime functions.
Move two time functions there and create tests for them.

The `humanizeUnit` function now uses ECMAscript apis. This ensures
that pluralization happens appropriately.

Also, fix a small bug in the results view to enure `repository`
is correctly pluralized.
## Checklist

### DIFF
--- a/extensions/ql-vscode/src/pure/time.ts
+++ b/extensions/ql-vscode/src/pure/time.ts
@@ -7,29 +7,71 @@ const durationFormatter = new Intl.RelativeTimeFormat('en', {
   numeric: 'auto',
 });
 
-// All these are approximate, specifically months and years
+// Months and years are approximate
 const MINUTE_IN_MILLIS = 1000 * 60;
 const HOUR_IN_MILLIS = 60 * MINUTE_IN_MILLIS;
 const DAY_IN_MILLIS = 24 * HOUR_IN_MILLIS;
 const MONTH_IN_MILLIS = 30 * DAY_IN_MILLIS;
 const YEAR_IN_MILLIS = 365 * DAY_IN_MILLIS;
 
-export function humanizeDuration(diffInMs?: number) {
-  if (diffInMs === undefined) {
+/**
+ * Converts a number of milliseconds into a human-readable string with units, indicating a relative time in the past or future.
+ *
+ * @param relativeTimeMillis The duration in milliseconds. A negative number indicates a duration in the past. And a positive number is
+ *  the future.
+ * @returns A humanized duration. For example, "in 2 minutes", "2 minutes ago", "yesterday", or "tomorrow".
+ */
+export function humanizeRelativeTime(relativeTimeMillis?: number) {
+  if (relativeTimeMillis === undefined) {
     return '';
   }
 
-  if (Math.abs(diffInMs) < HOUR_IN_MILLIS) {
-    return durationFormatter.format(Math.floor(diffInMs / MINUTE_IN_MILLIS), 'minute');
-  } else if (Math.abs(diffInMs) < DAY_IN_MILLIS) {
-    return durationFormatter.format(Math.floor(diffInMs / HOUR_IN_MILLIS), 'hour');
-  } else if (Math.abs(diffInMs) < MONTH_IN_MILLIS) {
-    return durationFormatter.format(Math.floor(diffInMs / DAY_IN_MILLIS), 'day');
-  } else if (Math.abs(diffInMs) < YEAR_IN_MILLIS) {
-    return durationFormatter.format(Math.floor(diffInMs / MONTH_IN_MILLIS), 'month');
+  if (Math.abs(relativeTimeMillis) < HOUR_IN_MILLIS) {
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/remote-queries/view/LastUpdated.tsx
```diff
@@ -23,7 +25,7 @@ const LastUpdated = ({ lastUpdated }: Props) => (
         <RepoPushIcon size={16} />
       </IconContainer>
       <Duration>
-        {humanizeDuration(lastUpdated)}
+        {humanizeDuration(lastUpdated && -lastUpdated)}
```
angelapwen, line 28: I'm not sure I understand this line of code — could you explain it?

angelapwen (approved):
✅ assuming the line I commented on makes sense 😸

extensions/ql-vscode/src/remote-queries/view/LastUpdated.tsx
shati-patel: I was about to ask the same question! What does it mean to `&&` these two numbers? 😅

aeisenberg: lastUpdated may be undefined. If so, do not want to negate it (this would convert it to NaN, which is BAD).

If lastUpdated is 0, then negating is a noop. So we don't need to negate.

Would it be easier to understand if I did this?

```

lastUpdated === undefined ? undefined : -lastUpdated
```

aeisenberg:
```suggestion
        {humanizeDuration(lastUpdated === undefined ? undefined : -lastUpdated)}
```

angelapwen: Hm, yes, I find that easier to understand. What happens if `lastUpdated` is an integer other than 0 though? Do we expect it to be a positive or negative integer (and why do we need to negate it at all if the absolute value is taken in `humanizeDuration()`?)

extensions/ql-vscode/src/pure/time.ts
```diff
@@ -0,0 +1,69 @@
...
+
+function createFormatter(unit: string) {
+  return Intl.NumberFormat('en-US', {
+    style: 'unit',
+    unit,
+    unitDisplay: 'long'
+  });
+}
+
+export function humanizeUnit(diffInMs?: number): string {
```
charisk, line 43: I have a very tired brain, but I can't easily understand the difference between `humanizeDuration` and `humanizeUnit` - can we think of different names?

extensions/ql-vscode/src/remote-queries/view/RemoteQueries.tsx
```diff
@@ -71,14 +71,19 @@ const openQueryTextVirtualFile = (queryResult: RemoteQueryResult) => {
   });
 };
 
+function createResultsDescription(queryResult: RemoteQueryResult) {
+  return `${queryResult.totalResultCount} results from running against ${queryResult.totalRepositoryCount
```
charisk, line 75: I'd personally break this down a bit to make it easier to follow.

extensions/ql-vscode/src/pure/time.ts
```diff
@@ -0,0 +1,69 @@
...
+});
+
+// All these are approximate, specifically months and years
+const MINUTE_IN_MILLIS = 1000 * 60;
+const HOUR_IN_MILLIS = 60 * MINUTE_IN_MILLIS;
+const DAY_IN_MILLIS = 24 * HOUR_IN_MILLIS;
+const MONTH_IN_MILLIS = 30 * DAY_IN_MILLIS;
+const YEAR_IN_MILLIS = 365 * DAY_IN_MILLIS;
+
+export function humanizeDuration(diffInMs?: number) {
```
charisk, line 17: I think it's worth having comments on these functions to help users of them know what they do.

extensions/ql-vscode/src/remote-queries/view/LastUpdated.tsx
aeisenberg: Good questions. If `lastUpdated` is something other than 0, then it is negated (meaning, a time in the past).

Absolute value is used to determine which unit to use. A positive integer indicates "XXX units _from now_". A negative integer indicates "XXX units ago".

Maybe it would be clearer if we ensured that `lastUpdated` was always negative and I can put a comment there.

extensions/ql-vscode/src/pure/time.ts
aeisenberg: I'll add comments. I can't think of better names.

extensions/ql-vscode/src/remote-queries/view/LastUpdated.tsx
angelapwen: oh yes! I see how it works with the new change now. I think that would make it clearer, yes!

aeisenberg:
All good now?

PR #1369, opened 2022-05-31
Be consistent about casing in Query History menu

elenatanasoiu:
Reported here: https://github.com/github/code-scanning/issues/6008

We originally started out by capitalizing each word [1], but made some
small changes [2] which resulted in our Query History options
being inconsistent.

Let's fix that.

[1]: https://github.com/github/vscode-codeql/blame/a5da5564969a0f37f95f1fe8b856f60810fc5caa/extensions/ql-vscode/package.json
[1]: https://github.com/github/vscode-codeql/commit/b470e4143126d0ceb65aee00fdb25e135deba328

## Questions

Building the extension: I've followed the [contribution docs](https://github.com/github/vscode-codeql/blob/main/CONTRIBUTING.md#building-the-extension) to do this using `npm build`. However, I could only complete the build successfully once I downgraded to node v16.14.2. For node v18 the build fails. Would we be okay with adding a `.nvmrc` file in this repo? 

To provide a bit more context: I only found the correct version by spinning up a codespace and snooping around. "Why don't you build the extension in a codespace?" you ask. While it does build successfully, it doesn't seem like you're able to then install the resulting .vsix file into VSCode. Happy to be told otherwise.

TL;DR: are we comfortable with pinning node to v16.14.2 in an `.nvmrc` file?

## Checklist


extensions/ql-vscode/package.json
```diff
@@ -471,7 +471,7 @@
       },
       {
         "command": "codeQLQueryHistory.openQuery",
-        "title": "Open the query that produced these results",
+        "title": "Open The Query That Produced These Results",
```
aeisenberg, line 474: minor: Generally, small words shouldn't be capitalized unless they are the first in the sentence.

```suggestion
        "title": "Open the Query that Produced these Results",
```

aeisenberg (commented):
Thanks for the PR!

aeisenberg:
Hmmm...I wasn't aware that the extension wasn't building in node v18. It's definitely worth fixing eventually, but in general, we need to stay in sync with the node version shipped with vscode (currently v16.13.0).

I think it is fine to add a `.nvmrc` file. Also, we should update the `engine` field in `package.json` to include a max version as well as min. Using `^16.11.25` should be fine since it allows for anything less than `17.0.0`.

Would you be able to raise an issue for the broken build on v18? It's something we should fix before vscode moves to v18.

aeisenberg (approved):

charisk:
Seems a bit silly but do we need to update the CHANGELOG file? The PR description check-list suggests yes. cc @shati-patel

shati-patel:
> Seems a bit silly but do we need to update the CHANGELOG file? The PR description check-list suggests yes. cc @shati-patel

No harm in updating the changelog, but it's definitely not necessary in this case since it's such a minor change!

elenatanasoiu:
Thanks @charisk & @shati-patel. Have added a line to the changelog.

shati-patel (approved):
:shipit: ✅

PR #1375, opened 2022-06-01
Use base styling for MRVA results font

elenatanasoiu:
It's been pointed out that MRVA results are hard to read as the font is
small and narrowly spaced. It also doesn't match the font type normally
used in source files.

We can instead switch to using the `font-family` values we've defined in 
our [base styles](https://github.com/github/vscode-codeql/commit/e8e50c43813ab6b6607850a044a2417209c099bb#diff-138b48c6b2640e1e37df0a473cdd159b69aeed8c4ea1a336d6c46771dac47197R2) and increase the font size from `x-small` to `small`.

## Before

Screenshot 2022-06-01 at 18 24 00

## After

Screenshot 2022-06-01 at 18 13 43

## Checklist


extensions/ql-vscode/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Prints end-of-query evaluator log summaries to the Query Log. [#1349](https://github.com/github/vscode-codeql/pull/1349)
 - Be consistent about casing in Query History menu. [#1369](https://github.com/github/vscode-codeql/pull/1369)
+- Use base styling for MRVA results font [#1375](https://github.com/github/vscode-codeql/pull/1375)
```
aeisenberg, line 7: I don't think we're putting MRVA changes into the changelog since this feature is not yet publicly released. So, this line can be removed.

extensions/ql-vscode/src/remote-queries/view/FileCodeSnippet.tsx
```diff
@@ -24,8 +24,8 @@ const getSeverityColor = (severity: ResultSeverity) => {
 const replaceSpaceAndTabChar = (text: string) => text.replaceAll(' ', '\u00a0').replaceAll('\t', '\u00a0\u00a0\u00a0\u00a0');
 
 const Container = styled.div`
-  font-family: ui-monospace, SFMono-Regular, SF Mono, Menlo, Consolas, Liberation Mono, monospace;
-  font-size: x-small;
+  font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Helvetica, Arial, sans-serif, Apple Color Emoji, Segoe UI Emoji;
```
aeisenberg, line 27: It looks like the font is no longer monospace. Is that intentional?

shati-patel: The original issue wasn't very clear, but I think the user (rightly) pointed out that our monospace font in the results view was much harder to read than the monospace font in the VS Code editor. So we probably want to match that editor font.

I expect we can do this by using the VS Code variable: `font-family: var(--vscode-editor-font-family);` (warning, I haven't properly tested this 🙈 )

shati-patel: > Yes, the original issue was:
> "The font used in the code snippet is quite small and narrowly spaced compared to the source files."
> The base styles don't use monospace:

Optional extra, if you want to fiddle around with the individual containers: While the code snippet should stay monospace, I could see the file link + message text looking nice in the base styles, e.g. something like this. What do you think? 🎨 

image

(Feel free to ignore though. Updating the overall font and font size is already an improvement! 💅🏽 ) 

---------- 
Side note: you probably already know this, but it's helpful to poke around in the VS Code webview using `Developer: Open Webview Developer Tools` (I didn't know about this command for ages 😆 )

extensions/ql-vscode/src/remote-queries/view/FileCodeSnippet.tsx
```diff
@@ -24,8 +24,8 @@ const getSeverityColor = (severity: ResultSeverity) => {
...
+  font-size: small;
```
shati-patel, line 28: I see that we still have
```js
const MessageText = styled.div`
  font-size: x-small;
  padding-left: 0.5em;
`;
```
on line 49 below, which means that the message text looks comparatively tiny 🔍 

Could we make the message text the same size as the rest of the container, i.e. also `small`? 😃

shati-patel (commented):
Thank you for the PR ⚡ 😸 

I've left a couple of suggestions, mainly to clarify the original issue!

extensions/ql-vscode/src/remote-queries/view/FileCodeSnippet.tsx
elenatanasoiu: Thanks, @shati-patel, I've updated the font family to use `var(--vscode-editor-font-family)` as you suggested and adjusted the size of `MessageText`. Both changes have taken effect. 

## Before

Screenshot 2022-06-13 at 12 58 13

## After

Screenshot 2022-06-13 at 13 16 26

I didn't know about `Developer: Open Webview Developer Tools`. Thanks for teaching me that! It makes changing props a lot easier.

In terms of the title and file link, I believe that because they're inside the `Container` component, they're already inheriting the font size:

Screenshot 2022-06-13 at 13 21 06

aeisenberg (approved):

Issue #1373, opened 2022-06-01
Broken build on Node v18


elenatanasoiu:
**Description**

Building the extension via `npm run build` breaks on Node v18, but works fine on previous versions. 

Full error output:

```
$ npm run build

> vscode-codeql@1.6.7 build
> gulp

[09:05:19] Requiring external module ts-node/register
[09:05:20] Using gulpfile ~/projects/vscode-codeql/extensions/ql-vscode/gulpfile.ts
[09:05:20] Starting 'default'...
[09:05:20] Starting 'cleanOutput'...
[09:05:20] Finished 'cleanOutput' after 14 ms
[09:05:20] Starting 'compileTypeScript'...
[09:05:20] Starting 'compileTextMateGrammar'...
[09:05:20] Starting 'compileView'...
[09:05:20] Starting 'copyTestData'...
[09:05:20] Starting 'copyViewCss'...
[09:05:20] Finished 'copyTestData' after 205 ms
[09:05:20] Finished 'compileTextMateGrammar' after 318 ms
[09:05:20] Finished 'copyViewCss' after 330 ms
webpack compiled successfully
[09:05:31] Finished 'compileView' after 11 s
[09:05:35] 'compileTypeScript' errored after 15 s
[09:05:35] Error: You must provide the URL of lib/mappings.wasm by calling SourceMapConsumer.initialize({ 'lib/mappings.wasm': ... }) before using SourceMapConsumer
    at readWasm (/Users/elenatanasoiu/projects/vscode-codeql/extensions/ql-vscode/node_modules/gulp-typescript/node_modules/source-map/lib/read-wasm.js:8:13)
    at wasm (/Users/elenatanasoiu/projects/vscode-codeql/extensions/ql-vscode/node_modules/gulp-typescript/node_modules/source-map/lib/wasm.js:25:16)
    at /Users/elenatanasoiu/projects/vscode-codeql/extensions/ql-vscode/node_modules/gulp-typescript/node_modules/source-map/lib/source-map-consumer.js:264:14
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
[09:05:35] 'default' errored after 15 s
```

**To reproduce**

```
cd extensions/ql-vscode
nvm install 18
nvm use 18
npm install
npm run build
```

**Expected behavior**
The build command should complete without errors.

**Additional context**
This passes in codespaces as a codespace currently defaults to v16. Discovered while working on https://github.com/github/vscode-codeql/pull/1369

[...]

PR #1378, opened 2022-06-03
Fix running build on node 18

aeisenberg:
Two problems:

1. source-map@0.7.3 does not work on node 18
2. need to set --openssl-legacy-provider because webpack is using
   openssl in a legacy way that was changed in node 17.

Fixes https://github.com/github/vscode-codeql/issues/1373

I'm reticent to merge this change even if it does work since we are pointing directly to a git repo and an unreleased sha for `source-map`. If a new release is not made before we need to upgrade to node 18, we may need to fork the repo and release a version ourselves.

## Checklist


elenatanasoiu:
> I'm reticent to merge this change even if it does work since we are pointing directly to a git repo and an unreleased sha for source-map.

I think the danger with this approach is that we might forget to replace the unreleased SHA with an official version of `source-map` as there's nothing prompting us to check for a newer version when we need to switch to Node 18.

Since the ultimate goal of this is to unblock us to use Node 18, I'd be inclined to suggest we wait until we'll need to run on 18 instead of half-heartedly upgrading `source-map` now. We'll then be prompted to upgrade source-map as part of that process and by that time it's likely an official version will be released.

What do you think?

aeisenberg:
There's no rush to upgrade right now, but also, the request to upgrade source-map has been open for a while and there still hasn't been a release.

My concern is that when vscode moves to node 18, this extension will be broken until we get a new version of source-map (and fix the other problems). At this point, we're going to need to scramble to fix it fast.

It's hard to tell how often vscode updates its node versions. Maybe every 8 months? And the last time was in March. I don't want to get blindsided by this in the future, but we don't need to fix this _now_.

elenatanasoiu:
@aeisenberg Good news! Looks like a fix was merged into source-map to fix the failing build (in Oct 2020): https://github.com/mozilla/source-map/issues/423. 

And a new release ([0.7.4](https://github.com/mozilla/source-map/tags)) containing the fix was pushed 10 days ago. Which is quite Nice of them. ;)

I've just tested this ([branch](https://github.com/github/vscode-codeql/compare/elenatanasoiu/node-18?expand=1))with the steps described [in the original issue](https://github.com/github/vscode-codeql/issues/1373) and the build now works correctly.

aeisenberg:
Thanks for looking into this. That's good news. Would you mind pushing up your change?

aeisenberg:
Closing in favour of #1378.

Issue #1377, opened 2022-06-03
Dumping CSV does not escape ',' character from results


JLLeitschuh:
**Describe the bug**

Execute the following CodeQL query:

```
import java
import semmle.code.java.dataflow.ExternalFlow

from string namespace, string type, boolean subtypes, string name, string signature, string ext,
  string input, string output, string kind, boolean generated, string row
where summaryModel(namespace, type, subtypes, name, signature, ext, input, output, kind, generated, row)
select namespace, type, subtypes, name, signature, ext, input, output, kind, generated, row
```

Then "Views Results CVS" on the result.

The output CVS will not have the `,` in entries escaped properly.

**Version**

```
CodeQL extension version: 1.6.5 
CodeQL CLI version: 2.9.1 
Platform: darwin arm64
```

**To reproduce**

See above

**Expected behavior**

Output CSV should escape the `,` character in entries

**Additional context**

aeisenberg:
Thanks for reporting. I can reproduce the error.

aeisenberg:
Problem is here: https://github.com/github/vscode-codeql/blob/2f9aca785e5e6c5616b3b3c20e360b1f68096db4/extensions/ql-vscode/src/run-queries.ts#L362

The `const chunk: DecodedBqrsChunk` that gets returned contains `tuples` and `columns`. The columns contain type information. Each string column should be quoted.

Issue #1376, opened 2022-06-02
Error when trying to view the CSV results of a quick evaluation


aeisenberg:
1. Perform a quick evaluation anywhere
2. After the results view opens, Click on View CSV results for this query in the history view
3. BUG
```
Exception caught at top level: /Users/andrew.eisenberg/Library/Application Support/Code/User/globalStorage/github.vscode-codeql/queries/ExternalFlow.qll-utKlrrgKpMXWbhpVDQOh2/results.bqrs has no result set called #select.
                      com.semmle.cli2.bqrs.DecodeCommand.generateOutput(DecodeCommand.java:237)
                      com.semmle.cli2.bqrs.DecodeCommand.executeSubcommand(DecodeCommand.java:213)
                      com.semmle.cli2.picocli.SubcommandCommon.executeWithParent(SubcommandCommon.java:484)
                      com.semmle.cli2.execute.CliServerCommand.lambda$executeSubcommand$0(CliServerCommand.java:67)
                      com.semmle.cli2.picocli.SubcommandMaker.runMain(SubcommandMaker.java:205)
                      com.semmle.cli2.execute.CliServerCommand.executeSubcommand(CliServerCommand.java:67)
                      com.semmle.cli2.picocli.SubcommandCommon.call(SubcommandCommon.java:500)
                      com.semmle.cli2.picocli.SubcommandMaker.runMain(SubcommandMaker.java:205)
                      com.semmle.cli2.picocli.SubcommandMaker.runMain(SubcommandMaker.java:214)
                      com.semmle.cli2.CodeQL.main(CodeQL.java:98)
```

What's happening is that the view CSV results is expecting a `#select` results set, but in this case, the results set does not exist. The only results set available is some renaming of the selected quick eval text.

I can think of two possible solutions:

1. Turn off csv results for quick eval (not great)
2. Arbitrarily choose a single results set to view if the `#select` results set is not available. 99% of the time (maybe 100% of the time) this will do the right thing.

adityasharad:
Can we get the correct result set name using `bqrs info`? I expect there to be only one result set.

aeisenberg:
If there is more than one result set, what does it mean to be "correct"? (Maybe it's the result set prefixed by `#`?)

If there is exactly one result set, do we even need the name of the result set? We just need to get the single set.

In the case of quick eval, I think there's only one result set.

PR #1379, opened 2022-06-03
Fix quoting of string columns in csv

aeisenberg:
Fixes two issues with CSV exporting and creates a unit test along with a generic framework for creating a mock CLI server. Commit-by-commit review is suggested.

Fixes https://github.com/github/vscode-codeql/issues/1377
Fixes https://github.com/github/vscode-codeql/issues/1376

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -956,11 +956,11 @@ export class QueryHistoryManager extends DisposableObject {
       void this.tryOpenExternalFile(query.csvPath);
       return;
     }
-    await query.exportCsvResults(this.qs, query.csvPath, () => {
+    if (await query.exportCsvResults(this.qs, query.csvPath)) {
       void this.tryOpenExternalFile(
         query.csvPath
       );
-    });
+    }
   }
 
   async handleViewCsvAlerts(
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -341,24 +341,33 @@ export class QueryEvaluationInfo {
   /**
    * Creates the CSV file containing the results of this query. This will only be called if the query
    * does not have interpreted results and the CSV file does not already exist.
+   *
+   * @return Promise<true> if the operation creates the file. Promise<false> if the operation does
+   * not create the file.
+   *
+   * @throws Error if the operation fails.
    */
-  async exportCsvResults(qs: qsClient.QueryServerClient, csvPath: string, onFinish: () => void): Promise<void> {
-    let stopDecoding = false;
-    const out = fs.createWriteStream(csvPath);
-    out.on('finish', onFinish);
-    out.on('error', () => {
-      if (!stopDecoding) {
-        stopDecoding = true;
-        void showAndLogErrorMessage(`Failed to write CSV results to ${csvPath}`);
-      }
-    });
+  async exportCsvResults(qs: qsClient.QueryServerClient, csvPath: string): Promise<boolean> {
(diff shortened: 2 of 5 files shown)


aeisenberg:
Adding @edoardopirovano as a reviewer since he is already familiar with this area.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -956,11 +956,11 @@ export class QueryHistoryManager extends DisposableObject {
       void this.tryOpenExternalFile(query.csvPath);
       return;
     }
-    await query.exportCsvResults(this.qs, query.csvPath, () => {
+    if (await query.exportCsvResults(this.qs, query.csvPath)) {
       void this.tryOpenExternalFile(
         query.csvPath
       );
-    });
+    }
```
edoardopirovano, line 963: Should we be displaying an error in the `else` case here? I see we `reject()` where we were previously using `showAndLogErrorMessage()` but as far as I can tell we are then swallowing that error.

extensions/ql-vscode/src/vscode-tests/no-workspace/run-queries.test.ts
```diff
@@ -53,6 +58,51 @@ describe('run-queries', () => {
...
 
+  [SELECT_QUERY_NAME, 'other'].forEach(resultSetName => {
+    it(`should export csv results for result set ${resultSetName}`, async () => {
+      const csvLocation = path.join(tmpDir.name, 'test.csv');
+      const qs = createMockQueryServerClient(
+        createMockCliServer({
+          bqrsInfo: [{ 'result-sets': [{ name: resultSetName }, { name: 'hucairz' }] }],
+          bqrsDecode: [{
+            columns: [{ kind: 'NotString' }, { kind: 'String' }],
+            tuples: [['a', 'b'], ['c', 'd']],
```
edoardopirovano, line 69: Let's expand the tuples here to include some quote characters and make sure those get escaped correctly.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -358,9 +358,12 @@ export class QueryEvaluationInfo {
         pageSize: 100,
         offset: nextOffset,
       });
-      for (const tuple of chunk.tuples) {
-        out.write(tuple.join(',') + '\n');
-      }
+      const quotes = chunk.columns.map(col => col.kind === 'String' ? '"' : '');
+      chunk.tuples.forEach((tuple) => {
+        out.write(tuple.map((v, i) => {
+          return `${quotes[i]}${v}${quotes[i]}`;
```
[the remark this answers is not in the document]
aeisenberg: Good point. I'll do that.

extensions/ql-vscode/src/query-history.ts
edoardopirovano: @aeisenberg I don't think you've addressed this comment. It's less blocking the other others, so I'm happy to approve without any changes here if you don't think any changes are needed, but just checking it wasn't missed accidentally.

aeisenberg: Oh...sorry. I forgot to comment here. The error is not being swallowed. Take a look at `commandRunnerWithProgress`, you can see that there is an error barrier there that catches and logs all thrown and rejected errors. So, you don't ever need to explicitly catch and log an error when you're inside a command as long as that command is wrapped in `commandRunnerWithProgress`.

edoardopirovano: Gotcha, thanks for the explanation 🙂

edoardopirovano (approved):
LGTM, thanks for addressing my comments!

PR #1386, opened 2022-06-15
Follow guidance for git tagging in contribution docs

elenatanasoiu:
Adding two things:
- A bit more detail on how to add a tag and how to delete a badly named one
- Switch to the official way of sharing tags according to the git docs [^1]

[^1]: https://git-scm.com/book/en/v2/Git-Basics-Tagging

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,15 +115,34 @@ From inside of VSCode, open the `launch.json` file and in the _Launch Integratio
     * Create a new commit with a message the same as the branch name.
     * Create a PR for this branch.
     * Wait for the PR to be merged into `main`
-1. Trigger a release build on Actions by adding a new tag on branch `main` named after the release, as above. Note that when you push to upstream, you will need to fully qualify the ref. A command like this will work:
+1. Switch to `main` and add a new tag on the `main` branch with your new version (named after the release), e.g.
+    ```bash
+    git checkout main
+    git tag v1.3.6
+    ```
+
+   If you've accidentally created a badly named tag, you can delete it via 
+    ```bash
+    git tag -d badly-named-tag
+    ```
+1. Push the new tag up:
+
+   a. If you're using a fork of the repo:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -115,15 +115,26 @@ From inside of VSCode, open the `launch.json` file and in the _Launch Integratio
...
 
+   If you've accidentally created a badly named tag, you can delete it via 
     ```bash
-    git push upstream refs/tags/v1.3.6
+    git tag -d badly-named-tag
     ```
+1. Push the new tag upstream:
+
+    ```bash
+    git push origin v1.3.6
```
shati-patel, line 131: Should we still use `refs/tags/v1.3.6` here? I think we originally used this fully-qualified ref in case there was a branch or something that was also called `v1.3.6`? 🤔 :thinking-about-git-intensifies:

Also minor: The example previously used `upstream` instead of `origin`. I personally use `upstream` for this repo and `origin` for my fork, so `upstream` was correct, but this may be different for others! 🍴  🆙

shati-patel (approved):
Thanks for updating this and being explicit about how tagging works 🤗 🏷️ 

Just a minor comment from me 💭

sampart:
Folk can also do this via https://github.com/github/vscode-codeql/releases/new. I've not tested on this repo, but on others I've found that creating a tag this way still triggers the necessary build steps. It may not be worth including, but I thought I'd mention it.

CONTRIBUTING.md
elenatanasoiu: Ah yes, using a fork complicates things. Thanks!

I've added both options as I don't think everyone works with a fork (I don't use one specifically to avoid the overcomplication).

elenatanasoiu:
Thanks @sampart! It's not 100% clear to me if you can create a "draft" release via that link so I've chosen to stick with the flow we already have.

PR #1387, opened 2022-06-15
Prepare for Node 18 upgrade

elenatanasoiu:
We'd like to pave the way to upgrade from Node 16.13.0 (current version) to Node 18 (when VSCode decides to switch to it).

As we discovered in https://github.com/github/vscode-codeql/pull/1378, our build is currently breaking on Node 18 for two reasons:
- `source-map@0.7.3` does not work on Node 18
- our current version of `webpack` relies on OpenSSL for its hashing algorithm but OpenSSL is not supported by default from Node 17 onwards. While we can provide a legacy flag to revert to using OpenSSL in Node 18, this flag doesn't work with our current version of Node (16.13.0) and is actually re-activating insecure behaviour.

To address these issues we're:
- upgrading to `source-map@0.7.4` which has very recently been released and works with Node 18
- switch to a newer version of `webpack` which doesn't rely on OpenSSL for its hashing algorithm and therefore doesn't require legacy support for OpenSSL in Node 18.

Happily, both these packages run with both Node 16.13.0 and Node 18.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package-lock.json
+++ b/extensions/ql-vscode/package-lock.json
@@ -26,6 +26,7 @@
         "react": "^17.0.2",
         "react-dom": "^17.0.2",
         "semver": "~7.3.2",
+        "source-map": "^0.7.4",
         "source-map-support": "^0.5.21",
         "stream": "^0.0.2",
         "stream-chain": "~2.2.4",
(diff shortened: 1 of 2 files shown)


aeisenberg:
If this works, it is much better than my approach. Also need to make sure the build still runs using node 16 (or at least the local build works).

PR #1388, opened 2022-06-15
Arcane workaround to fix a flaky test

aeisenberg:
For an inexplicable reason, the first time the selection
occurs, the value is incorrect. We often miss this error
in our tests if the expectation is reached before the
selection changed event fires.

It seems that the _second_ time the selection changed
event fires, the value is correct.

This change ensures we wait for the second selection change.
And we avoid running expectations until then.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/vscode-tests/no-workspace/query-history.test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/no-workspace/query-history.test.ts
@@ -402,14 +402,29 @@ describe('query-history', () => {
       });
     });
 
-    it('should get the selection from the treeView when both selections are empty', async () => {
+    it('should get the selection from the treeView when both selections are empty', async function() {
+      this.timeout(999999999);
       queryHistoryManager = await createMockQueryHistory(allHistory);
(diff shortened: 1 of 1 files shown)


shati-patel (approved):
I can't claim to understand what's going on either 😂 But it seems to work, so here's an approval! (anyone with more knowledge, feel free to chime in) 

Hopefully this'll stop those annoying random failures 🤞🏽

extensions/ql-vscode/src/vscode-tests/no-workspace/query-history.test.ts
```diff
@@ -402,14 +402,29 @@ describe('query-history', () => {
       });
     });
 
-    it('should get the selection from the treeView when both selections are empty', async () => {
+    it('should get the selection from the treeView when both selections are empty', async function() {
+      this.timeout(999999999);
```
aeisenberg, line 406: Eeek...this should go away.

```suggestion
    it('should get the selection from the treeView when both selections are empty', async () => {
```

aeisenberg:
@shati-patel do you want me to walk you through what I think is happening here?

PR #1389, opened 2022-06-16
Remove use of Primer's box component

charisk:
Since we're moving away from Primer, it makes sense to stop using `Box` for layout, so I've replaced usages of `Box` with simple `div`. 

BTW I noticed that we're using a mixture of ems and rems but I've decided to not touch that to allow minimal and straight-forward changes.

## Checklist
N/A

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/view/FileCodeSnippet.tsx
+++ b/extensions/ql-vscode/src/remote-queries/view/FileCodeSnippet.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import styled from 'styled-components';
 import { CodeSnippet, FileLink, HighlightedRegion, AnalysisMessage, ResultSeverity } from '../shared/analysis-result';
-import { Box } from '@primer/react';
 import VerticalSpace from './VerticalSpace';
 import { createRemoteFileRef } from '../../pure/location-link-utils';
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/remote-queries/view/FileCodeSnippet.tsx
```diff
@@ -188,36 +187,36 @@ const FileCodeSnippet = ({
...
-              </Box>
-            </Box>
+              </div>
+            </div>
             {message && severity && <Message
               message={message}
               currentLineNumber={startingLine + index}
               highlightedRegion={highlightedRegion}
-              borderColor={getSeverityColor(severity)}>
+              borderLeftColor={getSeverityColor(severity)}>
```
aeisenberg, line 219: This looks like a difference from the old implementation. Is this intentional?

aeisenberg (commented):
Looks good. Just one question.

extensions/ql-vscode/src/remote-queries/view/FileCodeSnippet.tsx
charisk: Yes it is. I've updated the prop for the `Message` component to be `borderLeftColor` instead of `borderColor` to avoid ambiguity with the variable called `bordedColor`.

aeisenberg (approved):

PR #1391, opened 2022-06-16
Allow repo names with periods

shati-patel:
Turns out that repo names can include periods (e.g. https://github.com/mozilla/pdf.js)! This PR updates the regex for permitted repo names, so that we no longer throw an error: 
image

Note: There are possibly some restrictions around whether the period (or hyphen/underscore) can appear anywhere in the repo name or only in the middle. I don't think it's necessary to be that specific (especially because I don't know what the exact rules are 😄) 

## Checklist

N/A: This is currently only used in MRVA and for downloading databases from GitHub, both of which are still internal-only.

### DIFF
--- a/extensions/ql-vscode/src/pure/helpers-pure.ts
+++ b/extensions/ql-vscode/src/pure/helpers-pure.ts
@@ -34,9 +34,9 @@ export const asyncFilter = async function <T>(arr: T[], predicate: (arg0: T) => 
 /**
  * This regex matches strings of the form `owner/repo` where:
  * - `owner` is made up of alphanumeric characters or single hyphens, starting and ending in an alphanumeric character
- * - `repo` is made up of alphanumeric characters, hyphens, or underscores
+ * - `repo` is made up of alphanumeric characters, hyphens, periods, or underscores
  */
-export const REPO_REGEX = /^(?:[a-zA-Z0-9]+-)*[a-zA-Z0-9]+\/[a-zA-Z0-9-_]+$/;
+export const REPO_REGEX = /^(?:[a-zA-Z0-9]+-)*[a-zA-Z0-9]+\/[a-zA-Z0-9-_\.]+$/;
 
 export function getErrorMessage(e: any) {
   return e instanceof Error ? e.message : String(e);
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/pure/helpers-pure.ts
```diff
@@ -34,9 +34,9 @@ export const asyncFilter = async function <T>(arr: T[], predicate: (arg0: T) =>
 /**
  * This regex matches strings of the form `owner/repo` where:
  * - `owner` is made up of alphanumeric characters or single hyphens, starting and ending in an alphanumeric character
- * - `repo` is made up of alphanumeric characters, hyphens, or underscores
+ * - `repo` is made up of alphanumeric characters, hyphens, periods, or underscores
  */
-export const REPO_REGEX = /^(?:[a-zA-Z0-9]+-)*[a-zA-Z0-9]+\/[a-zA-Z0-9-_]+$/;
+export const REPO_REGEX = /^(?:[a-zA-Z0-9]+-)*[a-zA-Z0-9]+\/[a-zA-Z0-9-_\.]+$/;
```
aeisenberg, line 39: Can owners have `.` in them? Is there any danger in being slightly more expansive in what we accept?

shati-patel: Good point. I'm not 100% sure about owner names, but it might be worth accepting ones with `.` (We'll still throw an error later if the owner/repo doesn't exist) 🤔

edoardopirovano: Also: the owner regex above requires the owner to end in an alphanumeric character. This is the current requirement for usernames but it hasn't always been the case, so there's users out there who's name ends in a symbol, e.g. https://github.com/mattr- which the regex wouldn't accept.

Maybe relaxing it to something like `/^[a-zA-Z0-9-_\.]+\/[a-zA-Z0-9-_\.]+/` and possibly failing later on is the simplest way forward.

shati-patel: Ah thanks! Good point about "legacy" names which might follow a different format... I'll update to a less strict regex 👍🏽 😸

edoardopirovano (approved):

PR #1396, opened 2022-06-21
Refactor: Invert dependency between query history and remote quries managers

charisk:
Currently, the `RemoteQueriesManager` depends on the `QueryHistoryManager` and related models e.g. `QueryHistoryInfo`. This is a bit problematic because the `RemoteQueriesManager` ends up having knowledge about query history stuff and can receive events that are potentially for local queries. 

This PR inverts that dependency and makes `RemoteQueriesManager` a dependency of QueryHistoryManager. The `RemoteQueriesManager` is completely independent of query history and simply publishes events that the `QueryHistoryManager` can consume. I think this way makes more sense because the query history area **has** to be aware of local/remote queries, but remote queries don't have to necessarily be aware of the query history.

## Checklist
N/A

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -544,26 +544,24 @@ export class QueryHistoryManager extends DisposableObject {
   }
 
   private registerToRemoteQueriesEvents() {
-    this.remoteQueriesManager.onRemoteQueryAdded(event => {
-      const historyItem: RemoteQueryHistoryItem = {
+    const queryAddedSubscription = this.remoteQueriesManager.onRemoteQueryAdded(event => {
+      this.addQuery({
         t: 'remote',
         status: QueryStatus.InProgress,
         completed: false,
         queryId: event.queryId,
         remoteQuery: event.query,
-      };
-
-      this.addQuery(historyItem);
+      });
     });
 
-    this.remoteQueriesManager.onRemoteQueryRemoved(async (event) => {
+    const queryRemovedSubscription = this.remoteQueriesManager.onRemoteQueryRemoved(async (event) => {
       const item = this.treeDataProvider.allHistory.find(i => i.t === 'remote' && i.queryId === event.queryId);
       if (item) {
         await this.removeRemoteQuery(item as RemoteQueryHistoryItem);
       }
     });
 
-    this.remoteQueriesManager.onRemoteQueryStatusUpdate(async (event) => {
+    const queryStatusUpdateSubscription = this.remoteQueriesManager.onRemoteQueryStatusUpdate(async (event) => {
       const item = this.treeDataProvider.allHistory.find(i => i.t === 'remote' && i.queryId === event.queryId);
       if (item) {
         const remoteQueryHistoryItem = item as RemoteQueryHistoryItem;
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/query-history.ts
```diff
@@ -307,21 +311,11 @@ export class QueryHistoryManager extends DisposableObject {
...
-    ._onDidRemoveQueryItem.event;
-
-  private readonly _onWillOpenQueryItem = super.push(new EventEmitter<QueryHistoryInfo>());
-  readonly onWillOpenQueryItem: Event<QueryHistoryInfo> = this
-    ._onWillOpenQueryItem.event;
-
   constructor(
     private readonly qs: QueryServerClient,
     private readonly dbm: DatabaseManager,
+    private readonly localQueriesInterfaceManager: InterfaceManager,
```
aeisenberg, line 317: We _could_ rename `InterfaceManager` to something like `LocalQueriesInterfaceManager`.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -548,12 +543,47 @@ export class QueryHistoryManager extends DisposableObject {
...
+    this.remoteQueriesManager.onRemoteQueryAdded(event => {
+      const historyItem: RemoteQueryHistoryItem = {
+        t: 'remote',
+        status: QueryStatus.InProgress,
+        completed: false,
+        queryId: event.queryId,
+        remoteQuery: event.query,
+      };
+
+      this.addQuery(historyItem);
```
aeisenberg, line 556: Minor:

```suggestion
      this.addQuery({
        t: 'remote',
        status: QueryStatus.InProgress,
        completed: false,
        queryId: event.queryId,
        remoteQuery: event.query,
      });
```

extensions/ql-vscode/src/query-history.ts
```diff
@@ -548,12 +543,47 @@ export class QueryHistoryManager extends DisposableObject {
     );
   }
 
+  private registerToRemoteQueriesEvents() {
+    this.remoteQueriesManager.onRemoteQueryAdded(event => {
```
aeisenberg, line 547: Pretty sure these need to be added to the subscription for disposal. All of the `onRemoteQuery*` event handlers.

aeisenberg (commented):
Looks sensible.

extensions/ql-vscode/src/query-history.ts
charisk: Good catch! Sorted.

charisk:
> Thanks for tidying this up! 🧹
> 
> I tried locally, and I wonder if something has changed with reporting progress of the `Run Variant Analysis` command? 🔄
> 
> Specifically, the final stage of the command ("Sending request") now stays on-screen the whole time until the results are back. I'd expect it to disappear as soon as we've successfully scheduled the run:
> 
> image

Great catch! Not sure how I missed this.

I'm not sure what could be causing this. I'll take a look tomorrow. @aeisenberg please let me know if you have any ideas. I don't think I've changed the "run remote query" code path so I'm a bit puzzled as to why this would be happening.

charisk:
Bug fixed now - there was an `await` on the execution of the monitor command and things just _happened_ to work before my changes.

elenatanasoiu (approved):
This makes sense 👍

PR #1397, opened 2022-06-21
MRVA: Improve experience when controller repo does not exist

elenatanasoiu:
The controller repo is set via the `codeQL.variantAnalysis.controllerRepo`
setting in VSCode.

While we have validation to check that the repo is not null and the
format of the controller repo is correct: `<owner>/<repo>`, we still
allow you to provide a non-existent repo (e.g. a mispelled one).

When the MRVA request is sent over to the API, it will verify that the
repo exists and return a very generic "Not Found" response.

This will then be logged out in the "Output" tab for VSCode.

We'd like to give users a better indication of what has gone wrong in
this case so we're making the error message more verbose.

Screenshot 2022-06-21 at 21 47 26

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
+++ b/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
@@ -333,8 +333,12 @@ async function runRemoteQueriesApiRequest(
     const { popupMessage, logMessage } = parseResponse(owner, repo, response.data);
     void showAndLogInformationMessage(popupMessage, { fullMessage: logMessage });
     return response.data.workflow_run_id;
-  } catch (error) {
-    void showAndLogErrorMessage(getErrorMessage(error));
+  } catch (error: any) {
+    if (error.status === 404) {
+      void showAndLogErrorMessage(`Controller repository was not found. Please make sure it's a valid repo name.${eol}`);
+    } else {
+      void showAndLogErrorMessage(getErrorMessage(error));
+    }
   }
 }
 


extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -334,7 +335,11 @@ async function runRemoteQueriesApiRequest(
     void showAndLogInformationMessage(popupMessage, { fullMessage: logMessage });
     return response.data.workflow_run_id;
   } catch (error) {
-    void showAndLogErrorMessage(getErrorMessage(error));
+    if ((error as Response).status === 404) {
```
charisk, line 338: Hmm what made you use `Response` from `node-fetch`? I think that response type is probably `RequestError` (https://octokit.github.io/types.ts/modules/RequestError.html) but I notice we don't have `@octokit/types` in our package.json. They both have `status` so TypeScript is happy.

I suggest we do the following:
- Bring in `@octokit/types` as a dev dependency and try to use `RequestError` from there.
- If for some reason that causes you headache, just go with `any` instead of `Response`:

```typescript
catch (error: any) {
  if (error.status === 404) {
   ...
}
```

elenatanasoiu: Thanks! Had a look at this with @shati-patel and we realized we're already using `OctokitResponse<QueriesResponse, number>)` when we make the API call in the same `try ... catch()` block. Using that type builds successfully so we didn't need to install `@octokit/types` as a dev dependency in the end. 

We did try to add `@octokit/types` initially but at the moment this is causing a Typescript error in another plugin:

```
/Users/elenatanasoiu/projects/vscode-codeql/extensions/ql-vscode/node_modules/@octokit/rest/node_modules/@octokit/plugin-rest-endpoint-methods/dist-types/generated/parameters-and-response-types.d.ts(425,60): error TS2339: Property 'POST 
```

There's an issue opened 20 hours ago on this, but it's yet unresolved: https://github.com/octokit/plugin-rest-endpoint-methods.js/issues/488

charisk: 🤔 we do use `OctokitResponse<QueriesResponse, number>)` but as the return type of `octokit.request()`. Here, we're trying to access the type of the error.

It's a shame that the typings are broken - thanks for checking though!

This is pretty minor but I think it could cause confusion. I'd rather we use `any` than the wrong type. What do you think?

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -334,7 +334,11 @@ async function runRemoteQueriesApiRequest(
     void showAndLogInformationMessage(popupMessage, { fullMessage: logMessage });
     return response.data.workflow_run_id;
   } catch (error) {
-    void showAndLogErrorMessage(getErrorMessage(error));
+    if ((error as OctokitResponse<QueriesResponse, number>).status === 404) {
+      void showAndLogErrorMessage(`Controller repository was not found. Please make sure it\'s a valid repo name.${eol}`);
```
charisk, line 338: How come we're adding a new line in the end?

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
elenatanasoiu: Fine by me! 😉 Have changed.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
elenatanasoiu: To make the error message easier to spot, like so:

Screenshot 2022-06-22 at 14 26 13

charisk (approved):
Thanks for the patience with this!

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -333,8 +333,12 @@ async function runRemoteQueriesApiRequest(
     const { popupMessage, logMessage } = parseResponse(owner, repo, response.data);
     void showAndLogInformationMessage(popupMessage, { fullMessage: logMessage });
     return response.data.workflow_run_id;
-  } catch (error) {
-    void showAndLogErrorMessage(getErrorMessage(error));
+  } catch (error: any) {
+    if (error.status === 404) {
+      void showAndLogErrorMessage(`Controller repository was not found. Please make sure it\'s a valid repo name.${eol}`);
```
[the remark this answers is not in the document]
elenatanasoiu: Ah I missed the escaping! Thanks!

For the eol, there's already one being added above this line so I'll choose to leave it at the end.

elenatanasoiu:
Thanks @charisk !

PR #1400, opened 2022-06-22
Fail fast if codeql repo is missing from the workspace

elenatanasoiu:
In order to run our cli-integration tests, we're required to have a
local copy of the `github/codeql` repo. We can then run the tests by 
running the `Launch Integration Tests - With CLI` task from inside 
VS Code.

(See CONTRIBUTING.md for details.)

If we don't have the repo cloned locally or we're not pointing to it
in `launch.json`, we don't get a clear indication of what the problem is.

The tests will still attempt to run.

Let's fail fast instead and add an actionable error message to the output.

Screenshot 2022-06-22 at 19 24 13

We're also updating `CONTRIBUTING.md` to make it a bit friendlier for 
newcomers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,9 +95,11 @@ Running from a terminal, you _must_ set the `TEST_CODEQL_PATH` variable to point
 
 ### Running the integration tests
 
-The _Launch Integration Tests - With CLI_ tests require a CLI instance in order to run. There are several environment variables you can use to configure this.
+You will need to run CLI tests using a task from inside of VS Code called _Launch Integration Tests - With CLI_.
 
-From inside of VSCode, open the `launch.json` file and in the _Launch Integration Tests - With CLI_ uncomment and change the environment variables appropriate for your purpose.
+The CLI integration tests require the CodeQL standard libraries in order to run so you will need to clone a local copy of the `github/codeql` repository.
+
+From inside of VSCode, open the `launch.json` file and in the _Launch Integration Tests - With CLI_ task, uncomment the `"${workspaceRoot}/../codeql"` line. If necessary, replace value with a path to your checkout, and then run the task.
 
 ## Releasing (write access required)
 
--- a/extensions/ql-vscode/src/vscode-tests/cli-integration/global.helper.ts
+++ b/extensions/ql-vscode/src/vscode-tests/cli-integration/global.helper.ts
@@ -75,7 +75,7 @@ export default function(mocha: Mocha) {
     }
   );
 
-  // ensure etension is cleaned up.
+  // ensure extension is cleaned up.
   (mocha.options as any).globalTeardown.push(
     async () => {
       const extension = await extensions.getExtension<CodeQLExtensionInterface | Record<string, never>>('GitHub.vscode-codeql')!.activate();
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -95,9 +95,13 @@ Running from a terminal, you _must_ set the `TEST_CODEQL_PATH` variable to point
...
 
-The _Launch Integration Tests - With CLI_ tests require a CLI instance in order to run. There are several environment variables you can use to configure this.
+You will need to run CLI tests using a task from inside of VS Code called _Launch Integration Tests - With CLI_.
 
-From inside of VSCode, open the `launch.json` file and in the _Launch Integration Tests - With CLI_ uncomment and change the environment variables appropriate for your purpose.
+The CLI integration tests require the CodeQL standard libraries in order to run so you will need to clone a local copy of the `github/codeql` repository.
+
+You will then need to go to `launch.json` and uncomment the `"${workspaceRoot}/../codeql"` line and then run the task.
+
+From inside of VSCode, open the `launch.json` file and in the _Launch Integration Tests - With CLI_ uncomment the `"${workspaceRoot}/../codeql"` line and then run the task.
```
aeisenberg, line 104:
```suggestion
From inside of VSCode, open the `launch.json` file and in the _Launch Integration Tests - With CLI_ task, uncomment the `"${workspaceRoot}/../codeql"` line. If necessary, replace value with a path to your checkout, and then run the task.
```

extensions/ql-vscode/src/vscode-tests/cli-integration/global.helper.ts
```diff
@@ -93,6 +93,22 @@ export default function(mocha: Mocha) {
...
+
+  // check that the codeql folder is found in the workspace
+  (mocha.options as any).globalSetup.push(
+    async () => {
+      const folders = workspace.workspaceFolders;
+      if (!folders) {
+        fail('\n\n\nNo workspace folders found.\nYou will need a local copy of the codeql repo.\nMake sure you specify the path to it in launch.json.\nIt should be something along the lines of "${workspaceRoot}/../codeql" depending on where you have your local copy of the codeql repo.\n\n\n');
+      } else {
+        // look for the codeql basename in the workspace folders
+        const codeqlFolder = folders.find(folder => folder.name === 'codeql');
```
aeisenberg, line 105: I was thinking that you do a quick check for a specific file inside of the repository. Something like this instead:

```suggestion
        const codeqlFolder = folders.find(f => fs.pathExistsSync(path.join(f.uri.fsPath, 'codeql-workspace.yml'));
```

I haven't tried this out myself, so the syntax may be off.

Also, usually, we don't want to run synchronous functions in `fs`, but since this is just test setup I think we're fine.

extensions/ql-vscode/src/vscode-tests/cli-integration/global.helper.ts
```diff
@@ -93,6 +93,21 @@ export default function(mocha: Mocha) {
...
   );
+
+  // check that the codeql folder is found in the workspace
+  (mocha.options as any).globalSetup.push(
+    async () => {
+      const folders = workspace.workspaceFolders;
+      if (!folders) {
+        fail('\n\n\nNo workspace folders found.\nYou will need a local copy of the codeql repo.\nMake sure you specify the path to it in launch.json.\nIt should be something along the lines of "${workspaceRoot}/../codeql" depending on where you have your local copy of the codeql repo.\n\n\n');
+      } else {
+        const codeqlFolder = folders.find(f => f.uri.fsPath.endsWith('codeql'));
```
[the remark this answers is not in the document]
elenatanasoiu: I couldn't find `fsPath.basename()`, but doing `folder.name` works detects when the folder exists. 

I've tested with & without pointing to the repo in `launch.json`.

elenatanasoiu:
Thanks for reviewing @adityasharad 💐

extensions/ql-vscode/src/vscode-tests/cli-integration/global.helper.ts
elenatanasoiu: I'm afraid this doesn't seem to be working ** old-man-screams-at-file-paths **

I also tried: 

```
const codeqlFolder = folders.find(folder => folder.name === 'codeql');
const codeqlWorkspaceFile = fs.readdirSync(codeqlFolder.uri.fsPath).find(file => file.match(/^codeql-workspace\.yaml$/));
```

but no joy.

CONTRIBUTING.md
elenatanasoiu: I've merged this 👍

extensions/ql-vscode/src/vscode-tests/cli-integration/global.helper.ts
elenatanasoiu: How bad would it be if we just checked for the folder name?

aeisenberg: That's fine. Better than what we had before.

aeisenberg (approved):

PR #1405, opened 2022-06-23
Parse summary of evaluator logs into data model

angelapwen:
This PR is the first step towards adding a visualizer for evaluator log data. It includes the following logic:
- running the command to generate log summaries, after a query is complete, in JSON format for ease of parsing
- parsing each interesting JSON object from the log summary into a data model (where interesting means the object contains all fields that our data model require) 
- adding a data model encapsulating fields that we expect to surface in our visualizer

After this PR is merged, we will merge a PR displaying the data in the data model in a [VSCode TreeView](https://code.visualstudio.com/api/extension-guides/tree-view).

The PR is marked as draft due to the number of comments labeled `REVIEW`. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/pure/log-summary-parser.ts
+++ b/extensions/ql-vscode/src/pure/log-summary-parser.ts
@@ -1,0 +1,31 @@
+import { EvaluatorLogData } from '../run-queries';
+
+/**
+ * A pure method that parses a string of evaluator log summaries into
+ * an array of EvaluatorLogData objects. 
+ * 
+ */
+ export function parseVisualizerData(logSummary: string): EvaluatorLogData[] {
+    // Remove newline delimiters because summary is in .jsonl format.
+    const jsonSummaryObjects: string[] = logSummary.split('\n\n');
+    const visualizerData: EvaluatorLogData[] = [];
+  
+    for (const obj of jsonSummaryObjects) {
+      const jsonObj = JSON.parse(obj);
+  
+      // Only convert log items that have an RA and millis field
+      // REVIEW: Not sure this is exactly what we want. Are we now excluding too many items?
+      // REVIEW: Is there a way to make this less brittle? 
+      if (jsonObj.ra != undefined && jsonObj.millis != undefined) {
+        const newLogData: EvaluatorLogData = {
+          queryName: jsonObj.queryCausingWork,
+          predicateName: jsonObj.predicateName,
+          timeInMillis: jsonObj.millis,
+          resultSize: jsonObj.resultSize
+          // TODO: need to also parse RA, pipeline arrays into the object. 
+        };
+        visualizerData.push(newLogData);
+      }
+    } 
+    return visualizerData;
+}
--- a/extensions/ql-vscode/src/run-queries.ts
+++ b/extensions/ql-vscode/src/run-queries.ts
@@ -37,6 +37,7 @@ import { ensureMetadataIsComplete } from './query-results';
 import { SELECT_QUERY_NAME } from './contextual/locationFinder';
 import { DecodedBqrsChunk } from './pure/bqrs-cli-types';
 import { getErrorMessage } from './pure/helpers-pure';
+import { parseVisualizerData } from './pure/log-summary-parser';
 
 /**
  * run-queries.ts
@@ -44,6 +45,28 @@ import { getErrorMessage } from './pure/helpers-pure';
  *
  * Compiling and running QL queries.
  */
+
+// REVIEW: I think these should be moved to another class (EvaluatorLogVisualizer
+// or EvaluatorLogData on its own once we wire it up with the tree visualizer)? 
+
+// REVIEW: Suggestions on other fields that are useful for a performance 
+// debugger would be welcome!
+export interface EvaluatorLogData {
+  queryName: string;
+  predicateName: string;
+  timeInMillis: number;
+  resultSize: number;
+  ra?: Pipeline[];
+}
+
+interface Pipeline {
+  pipelineName: string;
+  steps: PipelineStep[];
+}
+
+interface PipelineStep {
+  body: string;
+}
 
 /**
  * Information about which query will be to be run. `quickEvalPosition` and `quickEvalText`
@@ -103,6 +126,10 @@ export class QueryEvaluationInfo {
     return qsClient.findQueryEvalLogSummaryFile(this.querySaveDir);
   }
 
+  get jsonEvalLogSummaryPath() {
+    return qsClient.findJsonQueryEvalLogSummaryFile(this.querySaveDir);
+  }  
+  
   get evalLogEndSummaryPath() {
     return qsClient.findQueryEvalLogEndSummaryFile(this.querySaveDir);
   }
@@ -212,7 +239,28 @@ export class QueryEvaluationInfo {
             })
 
             .catch(err => {
-              void showAndLogWarningMessage(`Failed to generate structured evaluator log summary. Reason: ${err.message}`);
+              void showAndLogWarningMessage(`Failed to generate human-readable structured evaluator log summary. Reason: ${err.message}`);
+            });
+
+          // REVIEW: it seems clunky to have to run this command again just to get the same summary file
+          // in JSON format, but I'm not sure it's worth changing the way it works in the CLI for just
+          // this use case? 
+          void qs.cliServer.generateJsonLogSummary(this.evalLogPath, this.jsonEvalLogSummaryPath)
+            .then(() => {
+              // Convert summary into appropriate JSON object here. Perhaps move the parsing logic
+              // into another class later on?
+              
+              // REVIEW: We may want to stream this later on; we are generating the human-readable
+              // file in memory above but the JSON formatted one should be larger?
+              fs.readFile(this.jsonEvalLogSummaryPath, (err, buffer) => {
+                if (err) {
+                  throw new Error(`Could not read structured evaluator log end of summary JSON file at ${this.jsonEvalLogSummaryPath}.`);
+                }
+                parseVisualizerData(buffer.toString()); // Eventually this return value will feed into the tree visualizer.
+              });
+            })
+            .catch(err => {
+              void showAndLogWarningMessage(`Failed to generate JSON structured evaluator log summary. Reason: ${err.message}`);
             });
         } else {
           void showAndLogWarningMessage(`Failed to write structured evaluator log to ${this.evalLogPath}.`);
--- a/extensions/ql-vscode/test/pure-tests/log-summary-parser.test.ts
+++ b/extensions/ql-vscode/test/pure-tests/log-summary-parser.test.ts
@@ -1,0 +1,129 @@
+import { expect } from 'chai';
+import 'mocha';
+
+import { parseVisualizerData } from '../../src/pure/log-summary-parser';
+
+describe('Parsing Valid Summary Text', () => {
+    let validSummaryText = `
+    {
+      "completionTime" : "2022-06-23T14:02:42.007Z",
+      "raHash" : "e77dfaa9ciimqv7gb3imoesdb11",
+      "predicateName" : "query::ClassPointerType::getBaseType#dispred#f0820431#ff",
+      "appearsAs" : {
+        "query::ClassPointerType::getBaseType#dispred#f0820431#ff" : {
+          "uboot-taint.ql" : [ 3 ]
+        }
+      },
+      "queryCausingWork" : "uboot-taint.ql",
+      "evaluationStrategy" : "COMPUTE_SIMPLE",
+      "millis" : 11,
+      "resultSize" : 1413,
+      "dependencies" : {
+        "derivedtypes_2013#join_rhs" : "0da8f1fqbiin9i0mcitjedvlbc7",
+        "query::Class#class#f0820431#f" : "2e5d24rnudi1l99mvtse9u567b7",
+        "derivedtypes" : "070e120iu2i3dhj6pt13oqnbi66"
+      },
+      "ra" : {
+        "pipeline" : [
+          "    {1} r1 = CONSTANT(unique int)[1]",
+          "    {3} r2 = JOIN r1 WITH derivedtypes_2013#join_rhs ON FIRST 1 OUTPUT Rhs.3, Rhs.1, Rhs.2",
+          "    {4} r3 = JOIN r2 WITH query::Class#class#f0820431#f ON FIRST 1 OUTPUT Lhs.1, Lhs.2, 1, Lhs.0",
+          "    {2} r4 = JOIN r3 WITH derivedtypes ON FIRST 4 OUTPUT Lhs.0, Lhs.3",
+          "    return r4"
+        ]
+      },
+      "pipelineRuns" : [ {
+        "raReference" : "pipeline"
+      } ]
+    }
+
+    {
+      "completionTime" : "2022-06-23T14:02:42.072Z",
+      "raHash" : "0a0e3cicgtsru1m0cun896qhi61",
+      "predicateName" : "query::DefinedMemberFunction#class#f0820431#f",
+      "appearsAs" : {
+        "query::DefinedMemberFunction#class#f0820431#f" : {
+          "uboot-taint.ql" : [ 3 ]
+        }
+      },
+      "queryCausingWork" : "uboot-taint.ql",
+      "evaluationStrategy" : "COMPUTE_SIMPLE",
+      "millis" : 20,
+      "resultSize" : 8740,
+      "dependencies" : {
+        "fun_decls" : "e6e2e6vn02fcrrtp1ks1f75cd01",
+        "fun_def" : "6bec314lcq8sm3skg9mpecscp02",
+        "function_instantiation" : "8caddafufc3it9svjph9m8d43me",
+        "fun_decls_10#join_rhs" : "16474d6lehg7ssk83gnv2q7r8t8"
+      },
+      "ra" : {
+        "pipeline" : [
+          "    {2} r1 = SCAN fun_decls OUTPUT In.0, In.1",
+          "    {2} r2 = STREAM DEDUP r1",
+          "    {1} r3 = JOIN r2 WITH fun_def ON FIRST 1 OUTPUT Lhs.1",
+          "",
+          "    {2} r4 = SCAN function_instantiation OUTPUT In.1, In.0",
+          "    {2} r5 = JOIN r4 WITH fun_decls_10#join_rhs ON FIRST 1 OUTPUT Rhs.1, Lhs.1",
+          "    {1} r6 = JOIN r5 WITH fun_def ON FIRST 1 OUTPUT Lhs.1",
+          "",
+          "    {1} r7 = r3 UNION r6",
+          "    return r7"
+        ]
+      },
+      "pipelineRuns" : [ {
+        "raReference" : "pipeline"
+      } ]
+    }    
+  `
+  it('should return only valid EvaluatorLogData objects', () => {
+    const evaluatorLogData = parseVisualizerData(validSummaryText);
+    expect (evaluatorLogData.length).to.eq(2);
+    for (let item of evaluatorLogData) {
+        expect(item.queryName).to.not.be.empty;
+        expect(item.predicateName).to.not.be.empty;
+        expect(item.timeInMillis).to.be.a('number');
+        expect(item.resultSize).to.be.a('number');
+    }
+  });
+
+  let invalidHeaderText = `
+  {
+    "summaryLogVersion" : "0.3.0",
+    "codeqlVersion" : "2.9.0+202204201304plus",
+    "startTime" : "2022-06-23T14:02:41.607Z"
+  }
+  `
+
+  it('should not parse a summary header object', () => {
+    const evaluatorLogData = parseVisualizerData(invalidHeaderText);
+    expect (evaluatorLogData.length).to.eq(0);
+    for (let item of evaluatorLogData) {
+      expect(item).to.be.empty;
+    }
+  });
+
+  let invalidSummaryText = `
+  {
+    "completionTime" : "2022-06-23T14:02:42.019Z",
+    "raHash" : "6bec314lcq8sm3skg9mpecscp02",
+    "predicateName" : "fun_def",
+    "appearsAs" : {
+      "fun_def" : {
+        "uboot-taint.ql" : [ 3, 21, 22, 24 ]
+      }
+    },
+    "queryCausingWork" : "uboot-taint.ql",
+    "evaluationStrategy" : "EXTENSIONAL",
+    "millis" : 3,
+    "resultSize" : 8857
+  }  
+  `
+
+  it('should not parse a log event missing RA or millis fields', () => {
+    const evaluatorLogData = parseVisualizerData(invalidSummaryText);
+    expect (evaluatorLogData.length).to.eq(0);
+    for (let item of evaluatorLogData) {
+      expect(item).to.be.empty;
+    }
+  });
+});
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -689,6 +689,23 @@ export class CodeQLCliServer implements Disposable {
   }
 
   /**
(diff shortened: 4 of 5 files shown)


extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -0,0 +1,31 @@
...
+          predicateName: jsonObj.predicateName,
+          timeInMillis: jsonObj.millis,
+          resultSize: jsonObj.resultSize
+          // TODO: need to also parse RA, pipeline arrays into the object. 
+        };
+        visualizerData.push(newLogData);
+      }
+    } 
+    return visualizerData;
+}
```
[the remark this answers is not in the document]
angelapwen: Ah... I can move this back into `run-queries.ts`, but I think this will cause the test setup to be more complex.

extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -0,0 +1,31 @@
...
+ export function parseVisualizerData(logSummary: string): EvaluatorLogData[] {
+    // Remove newline delimiters because summary is in .jsonl format.
+    const jsonSummaryObjects: string[] = logSummary.split('\n\n');
+    const visualizerData: EvaluatorLogData[] = [];
+  
+    for (const obj of jsonSummaryObjects) {
+      const jsonObj = JSON.parse(obj);
+  
+      // Only convert log items that have an RA and millis field
+      // REVIEW: Not sure this is exactly what we want. Are we now excluding too many items?
```
edoardopirovano, line 17: I think that's reasonable. The only predicates missing a `millis` field are ones where we don't have a meaningful evaluation time for them (e.g. because we skipped evaluating due to a sentinel being empty or a cache hit), which aren't likely to be ones a user debugging performance cares about. The only ones missing `ra` are extensionals (that is: predicates that are part of the actual database). Those do have a meaningful evaluation time (because we have to load them from disk and into the cache), but it's usually pretty small so again not likely to be something a user debugging performance cares about.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -212,7 +239,28 @@ export class QueryEvaluationInfo {
             })
 
             .catch(err => {
-              void showAndLogWarningMessage(`Failed to generate structured evaluator log summary. Reason: ${err.message}`);
+              void showAndLogWarningMessage(`Failed to generate human-readable structured evaluator log summary. Reason: ${err.message}`);
+            });
+
+          // REVIEW: it seems clunky to have to run this command again just to get the same summary file
+          // in JSON format, but I'm not sure it's worth changing the way it works in the CLI for just
+          // this use case?
```
edoardopirovano, line 247: I think this is fine for now. I do think it's looking like we want a way to specify which types of output to send where in `codeql generate log-summary`. Then we could write a single invocation like `codeql generate log-summary --output "end-table:some-file" --output "summary:some-other-file" --output "human-readable:yet-another-file"`. But having something like that would be a pretty big rewrite of the already quite complex `codeql generate log-summary`.

extensions/ql-vscode/src/pure/log-summary-parser.ts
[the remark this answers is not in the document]
aeisenberg: `EvaluatorLogData` is just an interface. If you move the interface into this file, then you will avoid importing from a non-pure location.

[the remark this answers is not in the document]
angelapwen: Done! Thanks.

extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -0,0 +1,51 @@
+// REVIEW: Perhaps this file should be renamed now that it also includes the 
+// data model on top of the parsing logic?
+
+// REVIEW: Suggestions on other fields that are useful for a performance 
+// debugger would be welcome!
+export interface EvaluatorLogData {
```
adityasharad, line 6: One suggestion here: it might make your life easier to have the field names exactly match the fields from the evaluator log. Then it's easy to update this file (we could even derive it in future from the JSON schema, but that's a future idea).

extensions/ql-vscode/src/run-queries.ts
adityasharad: Agreed. Another option I just suggested verbally was to have an `--output ide` or similar, which produces all the different output formats the extension expects at this point, saved in separately named output files. But this can all wait for the future.

extensions/ql-vscode/test/pure-tests/log-summary-parser.test.ts
```diff
@@ -0,0 +1,129 @@
+import { expect } from 'chai';
+import 'mocha';
+
+import { parseVisualizerData } from '../../src/pure/log-summary-parser';
+
+describe('Parsing Valid Summary Text', () => {
+    const validSummaryText = `
```
adityasharad, line 7: At the risk of complicating your test code by introducing file reading, I think it would be useful to keep the JSON test data in standalone files checked into the repo. Then each test case only needs to know the input file name and assertions on the expected output.

extensions/ql-vscode/src/pure/log-summary-parser.ts
angelapwen: Thanks for sharing your thoughts on this! Glad to hear this is on the right track. I have a feeling there may be other edge cases that I'm not seeing (and therefore might fail while parsing) that I'll catch when I make the tests more robust.

extensions/ql-vscode/src/run-queries.ts
angelapwen: Sounds good. Just filed https://github.com/github/codeql-core/issues/2480 to track for later and move discussion there!

extensions/ql-vscode/src/pure/log-summary-parser.ts
angelapwen: That makes sense — I think I'd named them whatever was intuitive to me, but it'd be better to keep it consistent with the actual log for future maintenance. Deriving it from a JSON schema certainly seems like it'd make the parsing logic less brittle. I'll update it now to match the evaluator log fields.

extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -0,0 +1,51 @@
...
+    const jsonSummaryObjects: string[] = logSummary.split('\n\n');
+    const visualizerData: EvaluatorLogData[] = [];
+  
+    for (const obj of jsonSummaryObjects) {
+      const jsonObj = JSON.parse(obj);
+  
+      // Only convert log items that have an RA and millis field
+      // REVIEW: Not sure this is exactly what we want. Are we now excluding too many items?
+      // REVIEW: Is there a way to make this less brittle? 
+      if (jsonObj.ra != undefined && jsonObj.millis != undefined) {
```
adityasharad, line 39: Generally safer and easier to reason about!
```suggestion
      if (jsonObj.ra !== undefined && jsonObj.millis !== undefined) {
```

extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+  
+  interface Pipeline {
+    pipelineName: string;
+    steps: PipelineStep[];
+  }
+  
+  interface PipelineStep {
+    body: string;
+  }
```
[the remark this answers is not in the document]
angelapwen: Thank you! Resolved.

extensions/ql-vscode/src/pure/log-summary-parser.ts
angelapwen: I'd forgotten about `!==` 😄

extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -0,0 +1,42 @@
...
+    predicateName: string;
+    millis: number;
+    resultSize: number;
+    ra?: Pipeline[];
+}
+  
+interface Pipeline {
+    // Key: pipeline identifier; Value: array of pipeline steps 
+    pipeline: Map<string, string[]>;
+}
```
angelapwen, line 14: I have simplified the model behind `Pipeline` to match what the log actually emits more closely (a map where the key is the name of the pipeline and the value is an array of pipeline steps).

Previously I had `PipelineStep` as a separate interface that only held a `body` of strings, but that additional layer of indirection seemed unnecessary. Open to other thoughts here though!

extensions/ql-vscode/test/pure-tests/log-summary-parser.test.ts
angelapwen: I've made this change. I think it would still be good to make the tests more robust (right now I am testing for the edge cases that I have paid attention to and caught, but it may be good to test an entire full summary file to make sure nothing in the JSON file crashes). This may be a pretty large file to check in though, and I'm not entirely sure what I'd test for other than that it doesn't crash — given that it would be kind of laborious to manually check how many objects I'd expect to be parsed out of it. 

Maybe it would be good to just continuously add individual test cases that test different configurations of the JSON objects? 🤔

extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -0,0 +1,42 @@
...
+  
+/**
+ * A pure method that parses a string of evaluator log summaries into
+ * an array of EvaluatorLogData objects. 
+ * 
+ */
+ export function parseVisualizerData(logSummary: string): EvaluatorLogData[] {
+    // Remove newline delimiters because summary is in .jsonl format.
+    const jsonSummaryObjects: string[] = logSummary.split('\n\n');
```
aeisenberg, line 23: Try using:

```suggestion
    const jsonSummaryObjects: string[] = logSummary.split(os.EOL + os.EOL);
```

need to import `os` at the start of the file as well.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -212,7 +239,28 @@ export class QueryEvaluationInfo {
...
+          // REVIEW: it seems clunky to have to run this command again just to get the same summary file
+          // in JSON format, but I'm not sure it's worth changing the way it works in the CLI for just
+          // this use case? 
+          void qs.cliServer.generateJsonLogSummary(this.evalLogPath, this.jsonEvalLogSummaryPath)
+            .then(() => {
+              // Convert summary into appropriate JSON object here. Perhaps move the parsing logic
+              // into another class later on?
+              
+              // REVIEW: We may want to stream this later on; we are generating the human-readable
+              // file in memory above but the JSON formatted one should be larger?
```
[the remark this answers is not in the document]
angelapwen: After discussion with @adityasharad, decided to stream in a follow-up PR given this will go in under canary flag and make this PR ready for review. I have documented this as a TODO in the code comments.

extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -0,0 +1,31 @@
...
+      // Only convert log items that have an RA and millis field
+      // REVIEW: Not sure this is exactly what we want. Are we now excluding too many items?
+      // REVIEW: Is there a way to make this less brittle? 
+      if (jsonObj.ra != undefined && jsonObj.millis != undefined) {
+        const newLogData: EvaluatorLogData = {
+          queryName: jsonObj.queryCausingWork,
+          predicateName: jsonObj.predicateName,
+          timeInMillis: jsonObj.millis,
+          resultSize: jsonObj.resultSize
+          // TODO: need to also parse RA, pipeline arrays into the object.
```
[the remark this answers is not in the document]
angelapwen: Thanks for this comment, I hadn't considered saving something like the position in file in bytes. During 🍐 with @aeisenberg we discussed that this would be a good idea but will follow-on to the PR (and the front end visualizer PR) with this change, so that we can view simple/small logs in the visualizer first. The change is still under a canary flag 🐦 so we don't expect the memory pressure to be too pressing at the moment.

extensions/ql-vscode/test/pure-tests/log-summary-parser.test.ts
```diff
@@ -0,0 +1,46 @@
...
+      const validSummaryText = await fs.readFile(path.join(__dirname, 'evaluator-log-summaries/valid-summary.jsonl'), 'utf8');
+      const evaluatorLogData = parseVisualizerData(validSummaryText.toString());
+      expect (evaluatorLogData.length).to.eq(2);
+      for (const item of evaluatorLogData) {
+        expect(item.queryCausingWork).to.not.be.empty;
+        expect(item.predicateName).to.not.be.empty;
+        expect(item.millis).to.be.a('number');
+        expect(item.resultSize).to.be.a('number');
+        expect(item.ra).to.not.be.empty;
+        if (item.ra !== undefined) {
```
adityasharad, line 20: Turn this into an expectation: `expect(item.ra.to.not.be.undefined)`, before you check it is not empty.

extensions/ql-vscode/src/pure/log-summary-parser.ts
adityasharad: This does not appear to match your test data, where `ra` is an object with a single field called `pipeline`, whose value is an array of strings.

extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+  
+/**
+ * A pure method that parses a string of evaluator log summaries into
+ * an array of EvaluatorLogData objects. 
+ * 
+ */
+ export function parseVisualizerData(logSummary: string): EvaluatorLogData[] {
+    // Remove newline delimiters because summary is in .jsonl format.
+    const jsonSummaryObjects: string[] = logSummary.split(os.EOL + os.EOL);
```
adityasharad, line 25: Can you explain why we split on two EOLs? I see your test data is formatted to make each item multiline and has 2 EOLs between items, but is that identical to what the CLI produces?. The JSONL format requires only one EOL between lines.

adityasharad (commented):
I think this is shaping up well and you've correctly limited the scope for this PR, but I've confused myself about the type/structure of the expected data. Several questions and naming suggestions.

extensions/ql-vscode/src/pure/log-summary-parser.ts
angelapwen: Yes, I have taken the test data from summary files that the CLI produces. A single EOL delimits each line in the file, so splitting on one splits every single line in the file. Splitting on 2 EOLs splits by every object in the file. Let me know if that makes sense!

extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -0,0 +1,44 @@
...
+      const jsonObj = JSON.parse(obj);
+  
+      // Only convert log items that have an RA and millis field
+      if (jsonObj.ra !== undefined && jsonObj.millis !== undefined) {
+        const newLogData: EvaluatorLogData = {
+          queryCausingWork: jsonObj.queryCausingWork,
+          predicateName: jsonObj.predicateName,
+          millis: jsonObj.millis,
+          resultSize: jsonObj.resultSize,
+          ra: jsonObj.ra? jsonObj.ra: null
```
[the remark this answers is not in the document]
angelapwen: Hm, good point. I was originally trying to guard against this but I believe that the `undefined` line should take care of that condition. I have removed it.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -338,6 +331,48 @@ export class QueryEvaluationInfo {
...
+   * Calls the appropriate CLI command to generate a JSON log summary and parse it 
+   * into the appropriate data model for the log visualizer. 
+   */
+  parseJsonLogSummary(cliServer: cli.CodeQLCliServer): void {
+    void cliServer.generateJsonLogSummary(this.evalLogPath, this.jsonEvalLogSummaryPath)
+      .then(() => {         
+        // TODO(angelapwen): Stream the file in. 
+        fs.readFile(this.jsonEvalLogSummaryPath, (err, buffer) => {
+          if (err) {
+            throw new Error(`Could not read structured evaluator log end of summary JSON file at ${this.jsonEvalLogSummaryPath}.`);
```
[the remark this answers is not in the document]
angelapwen: Yes, thanks for catching this!

extensions/ql-vscode/test/pure-tests/log-summary-parser.test.ts
angelapwen: Ah, that makes sense. Done. I also made the `ra` field not-optional given we are not parsing in objects missing the field.

extensions/ql-vscode/test/pure-tests/log-summary-parser.test.ts
```diff
@@ -0,0 +1,46 @@
...
+          for (const pipeline of Object.entries(item.ra)) {
+            expect (pipeline).to.not.be.empty;
```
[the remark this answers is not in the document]
angelapwen: Ah, you're right. I think I actually meant something like

```
for (const pipeline of item.ra) {
  expect (pipeline).to.not.be.empty;
}
```

because I wanted to check that each pipeline in the array (`item.ra`) was non-empty. But when I use the above code I get `TypeError: item.ra is not iterable` 🤔 which I find odd, because it is of type `Pipeline[]`.

extensions/ql-vscode/test/pure-tests/log-summary-parser.test.ts
```diff
@@ -0,0 +1,46 @@
...
+      const evaluatorLogData = parseVisualizerData(invalidHeaderText);
+      expect (evaluatorLogData.length).to.eq(0);
+      for (const item of evaluatorLogData) {
+        expect(item).to.be.empty;
+      }
+    });
+  
+    it('should not parse a log event missing RA or millis fields', async function () {
+      const invalidSummaryText = await fs.readFile(path.join(__dirname, 'evaluator-log-summaries/invalid-summary.jsonl'), 'utf8');
+      const evaluatorLogData = parseVisualizerData(invalidSummaryText);
```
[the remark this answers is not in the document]
angelapwen: Makes sense! Done.

extensions/ql-vscode/test/pure-tests/log-summary-parser.test.ts
```diff
@@ -0,0 +1,46 @@
...
+      }
+    });
+  
+    it('should not parse a summary header object', async function () {
+      const invalidHeaderText = await fs.readFile(path.join(__dirname, 'evaluator-log-summaries/invalid-header.jsonl'), 'utf8');
+      const evaluatorLogData = parseVisualizerData(invalidHeaderText);
+      expect (evaluatorLogData.length).to.eq(0);
+      for (const item of evaluatorLogData) {
+        expect(item).to.be.empty;
+      }
```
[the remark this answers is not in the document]
angelapwen: Done.

extensions/ql-vscode/test/pure-tests/log-summary-parser.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    });
+  
+    it('should not parse a log event missing RA or millis fields', async function () {
+      const invalidSummaryText = await fs.readFile(path.join(__dirname, 'evaluator-log-summaries/invalid-summary.jsonl'), 'utf8');
+      const evaluatorLogData = parseVisualizerData(invalidSummaryText);
+      expect (evaluatorLogData.length).to.eq(0);
+      for (const item of evaluatorLogData) {
```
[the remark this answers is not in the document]
angelapwen: Done!

extensions/ql-vscode/src/pure/log-summary-parser.ts
angelapwen: Ah, there are actually cases where `ra` contains many `pipelines` with different names, which is why I used this schema. I've added an example of this case to the `valid-summary.jsonl` file but for convenience here is the related part:

```
"ra" : {
    "base" : [
      "    {2} r1 = namespaces AND NOT namespacembrs_1#antijoin_rhs(Lhs.0)",
      "    return r1"
    ],
    "standard" : [
      "    {2} r1 = JOIN QualifiedName::Namespace::getAQualifierForMembers#f0820431#ff#prev_delta WITH QualifiedName::Namespace::getAQualifierForMembers#f0820431#ff#join_rhs#1 ON FIRST 1 OUTPUT Rhs.1, Lhs.1",
      "",
      "    {2} r2 = JOIN QualifiedName::Namespace::getAQualifierForMembers#f0820431#ff#prev_delta WITH QualifiedName::Namespace::getAQualifierForMembers#f0820431#ff#join_rhs ON FIRST 1 OUTPUT Rhs.1, (Lhs.1 ++ \"::\" ++ Rhs.2)",
      "",
      "    {2} r3 = r1 UNION r2",
      "    {2} r4 = r3 AND NOT QualifiedName::Namespace::getAQualifierForMembers#f0820431#ff#prev(Lhs.0, Lhs.1)",
      "    return r4"
    ],
    "order_500000" : [
      "    {2} r1 = JOIN QualifiedName::Namespace::getAQualifierForMembers#f0820431#ff#prev_delta WITH QualifiedName::Namespace::getAQualifierForMembers#f0820431#ff#join_rhs#1 ON FIRST 1 OUTPUT Rhs.1, Lhs.1",
      "",
      "    {2} r2 = JOIN QualifiedName::Namespace::getAQualifierForMembers#f0820431#ff#prev_delta WITH namespacembrs ON FIRST 1 OUTPUT Rhs.1, Lhs.1",
      "    {2} r3 = JOIN r2 WITH namespaces ON FIRST 1 OUTPUT Lhs.0, (Lhs.1 ++ \"::\" ++ Rhs.1)",
      "",
      "    {2} r4 = r1 UNION r3",
      "    {2} r5 = r4 AND NOT QualifiedName::Namespace::getAQualifierForMembers#f0820431#ff#prev(Lhs.0, Lhs.1)",
      "    return r5"
    ]
  },
```

From this I thought it would be most accurate for `ra` to be an array of `pipeline` objects, that have a key (the name of the pipeline such as `base`) and a value of the array of strings.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -358,14 +359,11 @@ export class QueryEvaluationInfo {
...
-        // Convert summary into appropriate JSON object here. Perhaps move the parsing logic
-        // into another class later on?            
-        // REVIEW: We may want to stream this later on; we are generating the human-readable
-        // file in memory above but the JSON formatted one should be larger?
+      .then(() => {         
+        // TODO(angelapwen): Stream the file in. 
         fs.readFile(this.jsonEvalLogSummaryPath, (err, buffer) => {
           if (err) {
-            throw new Error(`Could not read structured evaluator log end of summary JSON file at ${this.jsonEvalLogSummaryPath}.`);
+            throw new Error(`Could not read structured evaluator log summary JSON file at ${this.jsonEvalLogSummaryPath}`);
```
aeisenberg, line 366: minor: Error messages should generally be complete sentences.

```suggestion
            throw new Error(`Could not read structured evaluator log summary JSON file at ${this.jsonEvalLogSummaryPath}.`);
```

aeisenberg (approved):
This is looking good to me. I think you've already addressed Aditya's comments.

extensions/ql-vscode/src/pure/log-summary-parser.ts
adityasharad: I think a closer representation would be for `ra` to be a map/object itself, rather than an array. The keys of this map would be the (string) pipeline names, and the values of this map would be arrays of RA log strings for each pipeline.

This may also explain what you are seeing when trying to parse the JSON logs into `Pipeline[]`.

angelapwen: I see; for some reason I added an extra layer of arrays in my mental model. I have renamed the interface `Pipelines` to indicate that the object contains all pipelines rather than an array of `Pipeline`s. This did indeed fix the `Pipeline[]` error. Thank you!

angelapwen:
> This is looking good to me. I think you've already addressed Aditya's comments.

Thank you all for the reviews! Merging for now ✨

PR #1408, opened 2022-06-28
Add cutoff repos and counts to error message

robertbrignull:
The queries API is changing so that it won't fail when you try to run a query on more than 1000 repos. Instead it'll  cap the repos off at 1000 and tell you in the response which repos weren't queried (up to a limit of 100 repos).

This PR is for handling that change in the API and display the information in the log message. While there I also updated the log message in general to include the counts of the repositories in the various states. I think this makes it more consistent if we're including the count for the cutoff repositories case.

Any input on the wording of the log message is very welcome. Writing these messages is very hard 😢 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
+++ b/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
@@ -39,6 +39,8 @@ interface QueriesResponse {
     invalid_repositories?: string[],
     repositories_without_database?: string[],
     private_repositories?: string[],
+    cutoff_repositories?: string[],
+    cutoff_repositories_count?: number,
   },
   repositories_queried: string[],
 }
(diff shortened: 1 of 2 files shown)


charisk (approved):
Messages look good to me!

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -358,18 +360,31 @@ export function parseResponse(owner: string, repo: string, response: QueriesResp
...
-    if (response.errors.repositories_without_database?.length) {
-      logMessage += `${eol2}Repositories without databases:${eol}${response.errors.repositories_without_database.join(', ')}`;
+    if (repositories_without_database?.length) {
+      logMessage += `${eol2}${repositories_without_database.length} repositories did not have a CodeQL database available:${eol}${repositories_without_database.join(', ')}`;
       logMessage += `${eol}For each public repository that has not yet been added to the database service, we will try to create a database next time the store is updated.`;
     }
-    if (response.errors.private_repositories?.length) {
-      logMessage += `${eol2}Non-public repositories:${eol}${response.errors.private_repositories.join(', ')}`;
+    if (private_repositories?.length) {
+      logMessage += `${eol2}${private_repositories.length} repositories are not public:${eol}${private_repositories.join(', ')}`;
```
elenatanasoiu, line 373: Since we're building a log message from:
- two end of lines
- a number
- a message
- another eol
- a list
in three separate situations, extracting those lines into a method might improve readability. 

e.g. 
```
logMessage += buildCutOffMessage(private_repositories, "repositories are not public");

func buildCutOffMessage(list, message) => `${eol2}${list.length} ${message}:${eol}${list.join(', ')}`; 
```

elenatanasoiu (approved):
This looks great 👍 

I've left an optional suggestion, but feel free to ignore it as I think the way we do message building is fine.

PR #1412, opened 2022-06-29
Fix error messages in extension log when quick evalling library files with CodeQL CLI v2.10.0

henrymercer:
Currently `resolve ml-models` only supports queryspecs, i.e. .ql, .qls, directory, and query pack specifications. However, a quick evaluation within a library will have a path ending in .qll, and currently the extension tries to call `resolve ml-models` on this path, logging an error message like this to the extension log:

```
[2022-06-29 10:15:25] Calling plumbing command: codeql resolve queries --additional-packs=/home/edoardo/semmle-code/ql --qlconfig-file=/home/edoardo/semmle-code/ql/qlconfig.yml --format startingpacks -- /home/edoardo/semmle-code/ql/cpp/ql/lib/semmle/code/cpp/internal/ResolveClass.qll
A fatal error occurred: /home/edoardo/semmle-code/ql/cpp/ql/lib/semmle/code/cpp/internal/ResolveClass.qll is not a .ql file, .qls file, a directory, or a query pack specification.
[2022-06-29 10:15:25] Exception caught at top level: /home/edoardo/semmle-code/ql/cpp/ql/lib/semmle/code/cpp/internal/ResolveClass.qll is not a .ql file, .qls file, a directory, or a query pack specification.
                      com.semmle.cli2.resolve.ResolveQueriesCommand.resolveQuerySpecifier(ResolveQueriesCommand.java:237)
                      com.semmle.cli2.resolve.ResolveQueriesCommand.executeJSON(ResolveQueriesCommand.java:174)
                      com.semmle.cli2.resolve.ResolveQueriesCommand.executeJSON(ResolveQueriesCommand.java:71)
                      com.semmle.cli2.picocli.SimpleJsonSubcommand.executeInsistingOnJSON(SimpleJsonSubcommand.java:117)
                      com.semmle.cli2.picocli.PlumbingRunner.call(PlumbingRunner.java:54)
                      com.semmle.cli2.picocli.SubcommandCommon.callPlumbingInProcess(SubcommandCommon.java:126)
                      com.semmle.cli2.resolve.ResolveMlModelsCommand.resolvePackDependencies(ResolveMlModelsCommand.java:233)
                      com.semmle.cli2.resolve.ResolveMlModelsCommand.executeJSON(ResolveMlModelsCommand.java:92)
                      com.semmle.cli2.resolve.ResolveMlModelsCommand.executeJSON(ResolveMlModelsCommand.java:49)
[...]

PR #1418, opened 2022-06-30
Resolve ml-queries from directory

aeisenberg:
Previously, there was a bug where quick eval queries would crash when
the eval snippet is in a library file.

The problem was that the `codeql resolve queries` command fails when
passed a library file. The fix is to avoid passing the library file at
all. Instead, pass the directory. This is safe because the resolve
queries command only needs to know which query pack the file is
contained in. Passing in the parent directory is the same as passing in
a file in this particular case.

This PR removes a previous workaround implemented here: https://github.com/github/vscode-codeql/pull/1412. This previous workaround simply disabled ml packs for quick eval.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -606,7 +606,8 @@ export class CodeQLCliServer implements Disposable {
   /** Resolves the ML models that should be available when evaluating a query. */
   async resolveMlModels(additionalPacks: string[], queryPath: string): Promise<MlModelsInfo> {
     const args = await this.cliConstraints.supportsPreciseResolveMlModels()
-      ? [...this.getAdditionalPacksArg(additionalPacks), queryPath]
+      // use the dirname of the path so that we can handle query libraries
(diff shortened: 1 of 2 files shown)


aeisenberg:
I don't think this needs a changelog entry since we didn't previously mention that ml models won't be used for quick evals in libraries (which is all this PR enables).

edoardopirovano (approved):
LGTM

henrymercer (approved):
This should always work because the command ultimately calls `codeql resolve queries --format startingpacks`, and the parent directory of a query will always live in (or be) the same pack as that query lives in.  The fact we need it probably indicates that something isn't quite right with the design here — I think we really want a command that takes an extension of queryspecs that includes qll files.

aeisenberg:
Yes. I think there's something we could do in the resolve queries command itself, and I started working on that, but it just made the command more complicated and more confusing. I didn't want to make this change to the codeql command for this one edge case. If we see more instances of this happening, then we may want to look into it.

PR #1420, opened 2022-07-01
Add API retries for octokit requests

robertbrignull:
The aim is to add automatic retrying for HTTP requests made with octokit. We've seen a few cases of spurious failure causing problems and this will go a long way to avoiding them. We may also want to add the ability to fully retry some operations like downloading MRVA results, but that's a separate change to this.

Is this all that's needed to use the `@octokit/plugin-retry` module and get automatic retrying? Reading https://github.com/octokit/plugin-retry.js this appears to be it. Just not sure how to tell if it's working or not. 🤔 
Maybe someone could work with me to write a test.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package-lock.json
+++ b/extensions/ql-vscode/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.6.9",
       "license": "MIT",
       "dependencies": {
+        "@octokit/plugin-retry": "^3.0.9",
         "@octokit/rest": "^18.5.6",
         "@primer/octicons-react": "^16.3.0",
         "@primer/react": "^35.0.0",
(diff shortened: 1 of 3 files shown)


charisk:
Thanks for taking a look at this! The default behaviour is that `400, 401, 403, 404, 422` codes won't be retried (according to [this](https://github.com/octokit/plugin-retry.js/blob/master/src/index.ts#L10)). Do you think we might want to retry 404s? I'm wondering if we ever want to protect against something not being there _yet_.

robertbrignull:
> The default behaviour is that `400, 401, 403, 404, 422` codes won't be retried (according to [this](https://github.com/octokit/plugin-retry.js/blob/master/src/index.ts#L10)). 

That default list of statuses to not retry looks sensible to me. Those should all be user errors where retrying won't make a difference, as opposed to 5xx errors indicating something went wrong out of the user's control.

> Do you think we might want to retry 404s? I'm wondering if we ever want to protect against something not being there _yet_.

In terms of a 404, that would only mean the controller repository doesn't exist _yet_ and I don't think we want to retry and wait for that. For repos not having databases those are filtered out and the request either succeeds with those repos filtered out or fails with a 422 if no repos have databases. Again here I don't think we want to attempt to wait for them. The time frame of how long the user probably wants to wait vs how long the thing will take to be available isn't really compatible.

Does that make sense and is that what you meant?

charisk (approved):

aeisenberg (approved):
This sounds reasonable.

PR #1424, opened 2022-07-12
Replace "repository search" filter box with VS Code UI Toolkit

shati-patel:
Another part of transitioning the MRVA webview from Primer to the VS Code UI Toolkit! See internal linked issue for details.
The repository filter box now uses a VS Code "codicon" and the `VSCodeTextField` component. 

-----

1. First commit (https://github.com/github/vscode-codeql/commit/84fafa445871bbec7460876fd9c3532d198c8367) just adds the [codicons](https://github.com/microsoft/vscode-codicons) package. This also involved updating the CSP and webview sources, for which I followed the Microsoft example from [here](https://github.com/microsoft/vscode-extension-samples/tree/main/webview-codicons-sample) with some additional tweaks until it worked 🙈 Please double-check and let me know if I can do that better!
2. Second commit (https://github.com/github/vscode-codeql/commit/6b1b103694b459963a340fe2d3dec2ec1b5321e2) actually replaces the filter box: 

| Before | After |
|--------|--------|
|  C4756BE5-3F6E-4285-9D0F-B04E4BB4B541|  D6A27C12-151A-4EB0-9A6B-7A724D00EB00  |

## Checklist

N/A - internal only 🐘 


extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
```diff
@@ -10,20 +9,16 @@ interface RepositoriesSearchProps {
...
+    <VSCodeTextField
+      placeholder='Filter by repository owner/name'
+      size='150'
+      ariaLabel="Repository search"
       name="repository-search"
-      placeholder="Filter by repository owner/name"
       value={filterValue}
-      onChange={(e: ChangeEvent) => setFilterValue((e.target as HTMLInputElement).value)}
-    />
+      onInput={(e: ChangeEvent) => setFilterValue((e.target as HTMLInputElement).value)}
```
shati-patel, line 18: For some reason, `onChange` didn't quite do the right thing for this new component, so I tried `onInput`, which worked fine!

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
```diff
@@ -10,20 +9,16 @@ interface RepositoriesSearchProps {
...
-      sx={{
-        backgroundColor: 'var(--vscode-editor-background);',
-        color: 'var(--vscode-editor-foreground);',
-        width: 'calc(100% - 14px)',
-      }}
-      leadingVisual={SearchIcon}
-      aria-label="Repository search"
+    <VSCodeTextField
+      placeholder='Filter by repository owner/name'
+      size='150'
```
shati-patel, line 14: Another slight change from the old component: I wasn't able to set an equivalent to `width: 'calc(100% - 14px)'`. 

`size='100'` made the box slightly less than full-width, so I picked `150` as something slightly higher. The box still resizes appropriately when you resize the screen 🤷🏽

extensions/ql-vscode/src/interface-utils.ts
```diff
@@ -149,7 +150,7 @@ export function getHtmlForWebview(
 <html>
   <head>
     <meta http-equiv="Content-Security-Policy"
-          content="default-src 'none'; script-src 'nonce-${nonce}'; style-src ${styleSrc}; connect-src ${webview.cspSource};">
+          content="default-src 'none'; script-src 'nonce-${nonce}'; font-src ${styleSrc}; style-src ${styleSrc}; connect-src ${webview.cspSource};">
```
charisk, line 153: 🤔 I'm not sure this makes sense. Does setting `font-src '${webview.cspSource}'` work?

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
```diff
@@ -111,6 +111,7 @@ export class RemoteQueriesInterfaceManager {
           localResourceRoots: [
             Uri.file(this.analysesResultsManager.storagePath),
             Uri.file(path.join(this.ctx.extensionPath, 'out')),
+            Uri.file(path.join(this.ctx.extensionPath, 'node_modules/@vscode/codicons/dist')),
```
charisk, line 114: I feel like this part should be dealt by webpack somehow. @aeisenberg what do you think?

The reason I mention webpack is because webpack is the tool that is building up our "executable" and sticks it in the `out` directory. It'd be cool if it included any codicons resources in it too.

extensions/ql-vscode/src/remote-queries/view/RepositoriesSearch.tsx
charisk: I believe you can set any CSS property you want by setting `style` e.g.

```jsx
<VSCodeTextField style={{ width: 'calc(100% - 14px)' }} 
```

but if it looks ok then I don't mind which way we do it!

charisk (commented):
Thank you for getting this working! I have some comments/questions that may be hard to answer. I'm happy to pair if it helps.

extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts
aeisenberg: It might be that without extra webpack instructions, this file won't exist when the extension is built.

shati-patel: I couldn't find any examples of other icons/things that we require from node_modules 😿 The other ones ([like the primer/octicons](https://github.com/github/vscode-codeql/blob/9c29c5c9c6b7c9f073034f79118f647138552e0b/extensions/ql-vscode/src/remote-queries/view/RepoListCopyButton.tsx#L4)) are generally just components that you can import 🤔 


Also, good call about building/installing the extension properly! I downloaded an actions build from https://github.com/github/vscode-codeql/actions/runs/2657604754 and the codicons aren't visible 🤦🏽 

I'll take a quick look at webpack - though I may need a hand to make changes 😅

shati-patel: Now that I've [moved `codicons` to the regular `dependencies`](https://github.com/github/vscode-codeql/pull/1424/commits/9669cd850a5e84516e318f0107a7dc0464010e0c) (instead of `devDependencies`), the current approach seems to work fine 😌 

Let me know if we still need/want to change webpack! Otherwise, this should be good to go 🤞🏽

charisk: Lets get it merged and we can play around with webpack some other time.

PR #1427, opened 2022-07-14
MRVA: Include more info in query history label

shati-patel:
Updates the default query history label for remote items to also include the language and number of repos queried. See internal issue for more details.

- Before: 
 
  image

- After: (Note: old query items still say `xxx on <controller-repo>`, but new ones give the number of repos)

  image

---------

### Comments/questions

- I'm not particularly happy about the different behaviour for older runs (i.e. using the controller repo name vs # repositories), but I couldn't see an obvious way to "backfill" the number of repos for older items 🤔 Happy for any suggestions!
- I've currently added the `(language)` to the end of the "query name" (`%q`) for remote queries. Do we want to make the `(language)` label a separate interpolated "thing" (e.g `%l`) or is this okay? 🔍 If we add it as a new interpolation option, we'd also need to handle it for local queries, which is probably beyond the scope of this PR 😅 

## Checklist

N/A - internal only 🍉 

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
+++ b/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
@@ -258,18 +258,19 @@ export async function runRemoteQuery(
     });
 
     const actionBranch = getActionBranch();
-    const apiResponses = await runRemoteQueriesApiRequest(credentials, actionBranch, language, repoSelection, owner, repo, base64Pack, dryRun);
+    const apiResponse = await runRemoteQueriesApiRequest(credentials, actionBranch, language, repoSelection, owner, repo, base64Pack, dryRun);
     const queryStartTime = Date.now();
     const queryMetadata = await tryGetQueryMetadata(cliServer, queryFile);
 
     if (dryRun) {
       return { queryDirPath: remoteQueryDir.path };
     } else {
-      if (!apiResponses) {
+      if (!apiResponse) {
         return;
       }
 
-      const { workflowRunId, numRepositoriesQueried } = apiResponses;
+      const workflowRunId = apiResponse.workflow_run_id;
+      const numRepositoriesQueried = apiResponse.repositories_queried.length;
       const remoteQuery = await buildRemoteQueryEntity(
         queryFile,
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -258,25 +258,27 @@ export async function runRemoteQuery(
     });
 
     const actionBranch = getActionBranch();
-    const workflowRunId = await runRemoteQueriesApiRequest(credentials, actionBranch, language, repoSelection, owner, repo, base64Pack, dryRun);
+    const apiResponses = await runRemoteQueriesApiRequest(credentials, actionBranch, language, repoSelection, owner, repo, base64Pack, dryRun);
```
charisk, line 261: Minor: there's still 1 API response, it just happens that we return 2 pieces of data from that response. So `apiResponse` would be the correct name IMO.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -336,7 +338,10 @@ async function runRemoteQueriesApiRequest(
     );
     const { popupMessage, logMessage } = parseResponse(owner, repo, response.data);
     void showAndLogInformationMessage(popupMessage, { fullMessage: logMessage });
-    return response.data.workflow_run_id;
+    return {
```
charisk, line 341: Feel free to ignore this, but just a thought: Here we could also return the whole response (`response.data`) and let the caller select the fields it needs. Then your return signature (`Promise<void | { workflowRunId: number, numRepositoriesQueried: number }>`) would be a bit more tidy too. E.g. `Promise<void | RunQueryResponse>` or something like that.

charisk (approved):

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
shati-patel: Ah yes, good point - we already have the type `QueriesResponse`, so I've changed the function to return the whole thing 👍🏽

PR #1430, opened 2022-07-14
Initial implementation of sourcemap-based jump-to-QL command

dbartol:
This PR implements a "Go to QL Code" command that jumps from RA code in a human-readable log summary to the corresponding QL code that generated it.

The actual mapping is done based on the sourcemap generated alongside the log summary. Debug info generation and sourcemap generation are now enabled by default for all queries run from the extension.

In order to appropriately enable or disable the command based on whether or not there is QL code corresponding to the current selection, we track changes to the active editor and current selection, setting a context variable that is used in the `enablement` clause for the command.

I've added an integration test that opens a summary log and runs the new command to jump to the QL code. There's a little bit of funny business to fix up the paths in the sourcemap because they normally include absolute paths.


extensions/ql-vscode/package.json
```diff
@@ -1084,6 +1096,10 @@
         {
           "command": "codeQL.previewQueryHelp",
           "when": "resourceExtname == .qhelp && isWorkspaceTrusted"
+        },
+        {
+          "command": "codeQL.gotoQL",
+          "when": "editorLangId == ql-summary"
```
aeisenberg, line 1102: Do you need to add `"enablement": "codeql.hasQLSource"` here so that the command is only enabled when the context variable is true? Not sure if it works the way you have it. Might be fine...just not sure.

extensions/ql-vscode/src/log-insights/summary-language-support.ts
```diff
@@ -0,0 +1,139 @@
...
+      }
+
+      const mapPath = document.uri.fsPath + '.map';
+
+      try {
+        const sourceMapText = await fs.promises.readFile(mapPath, 'utf-8');
+        const rawMap: RawSourceMap = JSON.parse(sourceMapText);
+        this.sourceMap = await new SourceMapConsumer(rawMap);
+      } catch {
+        // Error reading sourcemap. Pretend there was no sourcemap
```
aeisenberg, line 83: Maybe emit a log message to the the output view, so we can track this.

extensions/ql-vscode/src/log-insights/summary-language-support.ts
```diff
@@ -0,0 +1,139 @@
...
+    const document = editor.document;
+    if (document.languageId !== 'ql-summary') {
+      return undefined;
+    }
+
+    if (document.uri.scheme !== 'file') {
+      return undefined;
+    }
+
+    if (this.lastDocument !== document) {
```
aeisenberg, line 70: Are we guaranteed that these will always be instance equals? What if there are two editors open for the same file? Can we get away with sharing the same sourcemap?

Would it be safer to store the uri to the document and compare that instead? This would avoid saving a reference to the document instance in this object. It just feels unsafe to keep a handle on the document since we don't know what the object lifecycle is from inside of vscode. If it works, then better to save the uri.

extensions/ql-vscode/package.json
adityasharad: Shall we also gate this behind the canary flag?

extensions/ql-vscode/src/log-insights/summary-language-support.ts
```diff
@@ -0,0 +1,139 @@
+import * as fs from 'fs';
+import { RawSourceMap, SourceMapConsumer } from 'source-map';
+import { commands, Position, Selection, TextDocument, TextEditor, TextEditorRevealType, TextEditorSelectionChangeEvent, ViewColumn, window, workspace } from 'vscode';
+import { DisposableObject } from '../pure/disposable-object';
+import { commandRunner } from '../commandRunner';
+
+/** A `Position` within a specified file on disk. */
+interface PositionInFile {
+  filePath: string;
+  position: Position;
```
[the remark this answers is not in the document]
dbartol: Indeed:
- In the CLI, `Position` uses one-based lines and columns.
- The sourcemap standard uses zero-based lines and columns.
- VS Code's `Position` class uses one-based lines and columns.
- The `source-map` library used by the VS Code extension uses one-based lines and zero-based columns 🙄

At this point, I've either handled them all correctly, or I've made an even number of mistakes :grinning:

extensions/ql-vscode/src/log-insights/summary-language-support.ts
```diff
@@ -76,11 +77,16 @@ export class SummaryLanguageSupport extends DisposableObject {
...
         this.sourceMap = await new SourceMapConsumer(rawMap);
-      } catch {
-        // Error reading sourcemap. Pretend there was no sourcemap
+      } catch (e: unknown) {
+        // Error reading sourcemap. Pretend there was no sourcemap.
+        if (e instanceof Error) {
+          void logger.log(`Error reading sourcemap file '${mapPath}': ${e.message}`);
+        } else {
+          void logger.log(`Error reading sourcemap file '${mapPath}'`);
+        }
```
aeisenberg, line 89: Minor: There is a helper function that will make this simpler. 99.9% of the time (maybe 100% if the time here) this will be an instance of `Error`. See the `getErrorMessage` in `helpers.ts`.

dbartol: Thanks. Done.

extensions/ql-vscode/src/log-insights/summary-language-support.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Updates the `codeql.hasQLSource` context variable based on the current selection. This variable
+   * controls whether or not the `codeQL.gotoQL` command is enabled.
+   */
+  private async updateContext(): Promise<void> {
+    const position = await this.getQLSourceLocation();
+
+    const result = await commands.executeCommand('setContext', 'codeql.hasQLSource', position !== undefined);
+    void result;
```
aeisenberg, line 123: This is semantically equivalent.

```suggestion
    await commands.executeCommand('setContext', 'codeql.hasQLSource', position !== undefined);
```

extensions/ql-vscode/src/log-insights/summary-language-support.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  handleDidChangeActiveTextEditor = async (editor: TextEditor | undefined): Promise<void> => {
+    void editor;
```
aeisenberg, line 127: You can remove this line.
```suggestion
    void editor;
```

This might make the linter unhappy. If so, just delete the parameter.

extensions/ql-vscode/src/log-insights/summary-language-support.ts
aeisenberg: As long as we are guaranteed to release the text document reference when the editor closes. I _think_ that's already happening because the event handler below will be fired when the active editor changes.

extensions/ql-vscode/src/log-insights/summary-language-support.ts
```diff
@@ -0,0 +1,142 @@
...
+    await this.updateContext();
+  }
+
+  handleDidChangeTextEditorSelection = async (e: TextEditorSelectionChangeEvent): Promise<void> => {
+    void e;
```
aeisenberg, line 132: Same here.

aeisenberg (commented):
Some minor suggestions.

extensions/ql-vscode/src/log-insights/summary-language-support.ts
dbartol: I was actually caching the sourcemap of the most recently active _summary_ file, so merely switching to another active editor would not invalidate the cache or release the document unless that new editor was a different summary file. I've added the event handler to discard the cache when the cached document is closed.

extensions/ql-vscode/src/log-insights/summary-language-support.ts
dbartol: I forgot that the usual way of shutting up the linter was just to put an underscore in front of the parameter name. I've done that.

extensions/ql-vscode/src/log-insights/summary-language-support.ts
dbartol: Fixed with underscore

aeisenberg (approved):
Not sure what the status of this PR is, but it looks good to me. Is this still a draft?

extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -1,16 +1,16 @@
...
   resultSize: number;
-  // Key: pipeline identifier; Value: array of pipeline steps 
+  // Key: pipeline identifier; Value: array of pipeline steps
   ra: Record<string, string[]>;
 }
 
 /**
  * A pure method that parses a string of evaluator log summaries into
- * an array of EvalLogData objects.
+ * an array of EvaluatorLogData objects.
```
dbartol, line 13:
```suggestion
 * an array of EvalLogData objects.
```

extensions/ql-vscode/src/vscode-tests/minimal-workspace/sourcemap.test.ts
```diff
@@ -0,0 +1,71 @@
+import { fail } from 'assert';
+import { commands, Selection, window, workspace } from 'vscode';
+import * as path from 'path';
+import * as assert from 'assert';
+import { expect } from 'chai';
+import { tmpDir } from '../../helpers';
+import * as fs from 'fs-extra';
+
+export function run() {
```
aeisenberg, line 9: I don't think it is necessary to wrap and export the tests.

extensions/ql-vscode/src/vscode-tests/minimal-workspace/sourcemap.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Integration tests for queries
+   */
+  describe('SourceMap', function() {
+    this.timeout(200000);
+
+    it('should jump to QL code', async () => {
+      try {
+        this.timeout(60000);
```
aeisenberg, line 18: This is a little bit strange. `this` is the enclosing describe object. I am not sure if the `timeout()` is actually affecting this test.  I think you want to do this.

```suggestion
    it('should jump to QL code', async function() {
      try {
        this.timeout(60000);
```

extensions/ql-vscode/src/vscode-tests/minimal-workspace/sourcemap.test.ts
```diff
@@ -0,0 +1,71 @@
...
+import { expect } from 'chai';
+import { tmpDir } from '../../helpers';
+import * as fs from 'fs-extra';
+
+export function run() {
+  /**
+   * Integration tests for queries
+   */
+  describe('SourceMap', function() {
+    this.timeout(200000);
```
aeisenberg, line 14: Quite a large timeout. Is this necessary?

extensions/ql-vscode/src/vscode-tests/minimal-workspace/sourcemap.test.ts
```diff
@@ -0,0 +1,71 @@
...
+        fail(e as Error);
+      }
+    });
+
+    async function copyFilesToTempDirectory<T extends Record<string, string>>(files: T): Promise<T> {
+      const tempDir = path.join(tmpDir.name, 'log-summary');
+      await fs.ensureDir(tempDir);
+      const result: Record<string, string> = {};
+      for (const key in files) {
+        const srcPath = files[key];
```
aeisenberg, line 62: minor:

```suggestion
      for (const srcPath in Object.values(files)) {
```

aeisenberg (commented):
I reviewed this before I realized it was still in draft, so apologies if I commented on a bunch of stuff you were going to remove anyway.

extensions/ql-vscode/src/vscode-tests/minimal-workspace/sourcemap.test.ts
dbartol: Isn't that how we ensure that the activation test runs before this new test?

extensions/ql-vscode/src/vscode-tests/minimal-workspace/sourcemap.test.ts
dbartol: `key` is reused at the bottom of the loop to set the corresponding result property.

aeisenberg: Ahh...right:

```suggestion
      for (const [key, srcPath] in Object.entries(files)) {
```

extensions/ql-vscode/src/vscode-tests/minimal-workspace/sourcemap.test.ts
aeisenberg: Oh...not sure. If this is working for you, can you add a comment here?

dbartol: This is irrelevant now that I moved the test to the `cli-integration` suite

extensions/ql-vscode/src/vscode-tests/minimal-workspace/sourcemap.test.ts
dbartol: I removed that timeout anyway.

extensions/ql-vscode/src/vscode-tests/minimal-workspace/sourcemap.test.ts
dbartol: Reduced. That was just to avoid a timeout when debugging.

extensions/ql-vscode/src/vscode-tests/minimal-workspace/sourcemap.test.ts
dbartol: Ah, that works. Fixed.

extensions/ql-vscode/test/data/lib/semmle/code/cpp/Namespace.qll
```diff
@@ -0,0 +1,237 @@
+/**
```
adityasharad, line 1: Minor: I think a comment saying this is test data would also be a helpful reminder. The contents don't matter so much as the fact that it exercises various QL features.

extensions/ql-vscode/src/vscode-tests/cli-integration/query.test.ts
```diff
@@ -151,7 +151,8 @@ describe('using the query server', function() {
             localChecking: false,
             noComputeGetUrl: false,
             noComputeToString: false,
-            computeDefaultStrings: true
+            computeDefaultStrings: true,
+            emitDebugInfo: true
```
adityasharad, line 155: Future note: we may want to parameterise/matrix this test so that we check both `true/false` for `emitDebugInfo`.

adityasharad (approved):
Tests also look good.

extensions/ql-vscode/test/data/lib/semmle/code/cpp/Namespace.qll
dbartol: Fixed.

PR #1432, opened 2022-07-18
Add result count to remote queries in Query History 

elenatanasoiu:
When you run a remote query, we'd like to display more information about
it in the Query History panel.

At the moment [we've improved this][1] by adding the language and number of repositories.

Here we're also adding the number of results for a remote query.

So the final format of the query history item will change from:

`<query_name> - <query_status>`

to

`<query_name> (<language>) on x repositories - <query_status>  (y results)`

While we're here we're also:
- changing local queries to display number of results in parentheses to match remote queries
- make remote queries sortable by number of results

## Before

Screenshot 2022-07-18 at 17 16 07

## After

Screenshot 2022-07-18 at 17 12 04

[1]: https://github.com/github/vscode-codeql/pull/1427

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/history-item-label-provider.ts
+++ b/extensions/ql-vscode/src/history-item-label-provider.ts
@@ -67,10 +67,10 @@ export class HistoryItemLabelProvider {
 
   // Return the number of repositories queried if available. Otherwise, use the controller repository name.
   private buildRepoLabel(item: RemoteQueryHistoryItem): string {
-    const numRepositoriesQueried = item.remoteQuery.numRepositoriesQueried;
+    const repositoryCount = item.remoteQuery.repositoryCount;
 
-    if (numRepositoriesQueried) {
-      return pluralize(numRepositoriesQueried, 'repository', 'repositories');
+    if (repositoryCount) {
+      return pluralize(repositoryCount, 'repository', 'repositories');
     }
 
     return `${item.remoteQuery.controllerRepository.owner}/${item.remoteQuery.controllerRepository.name}`;
--- a/extensions/ql-vscode/src/history-item-label-provider.ts
+++ b/extensions/ql-vscode/src/history-item-label-provider.ts
@@ -74,8 +74,7 @@ export class HistoryItemLabelProvider {
       // Return the number of repositories queried if available. Otherwise, use the controller repository name.
       d: numRepositoriesQueried ? numRepositoriesLabel : `${item.remoteQuery.controllerRepository.owner}/${item.remoteQuery.controllerRepository.name}`,
 
-      // There is no synchronous way to get the results count.
-      r: '',
+      r: item.resultCount === undefined ? '' : `(${item.resultCount} results)`,
       s: item.status,
       f: path.basename(item.remoteQuery.queryFilePath),
       '%': '%'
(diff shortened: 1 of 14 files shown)


charisk:
Perhaps a Product/UX question, but do we really need the status of the query in the label? I think the icons cover the status. 

It's just getting longer and longer 😬

extensions/ql-vscode/src/history-item-label-provider.ts
```diff
@@ -74,8 +74,7 @@ export class HistoryItemLabelProvider {
       // Return the number of repositories queried if available. Otherwise, use the controller repository name.
       d: numRepositoriesQueried ? numRepositoriesLabel : `${item.remoteQuery.controllerRepository.owner}/${item.remoteQuery.controllerRepository.name}`,
 
-      // There is no synchronous way to get the results count.
-      r: '',
+      r: item.resultCount === undefined ? '' : `(${item.resultCount} results)`,
```
charisk, line 77: Elsewhere we've been careful with pluralization (`result` if 1 result, `results` if more than one), shall we do the same here?

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -42,6 +42,7 @@ export interface UpdatedQueryStatusEvent {
   status: QueryStatus;
   failureReason?: string;
   numRepositoriesQueried?: number;
```
charisk, line 44: We use numXXX in some cases (`numRepositoriesQueried`) and XXXcount in others (`resultCount`). It'd be nice to use a convention. I prefer `count` so the comment is not really related to this PR I guess 😬 but we could change it later on.

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -317,10 +318,12 @@ export class RemoteQueriesManager extends DisposableObject {
     if (resultIndex) {
       const metadata = await this.getRepositoriesMetadata(resultIndex, credentials);
       const queryResult = this.mapQueryResult(executionEndTime, resultIndex, queryId, metadata);
+      const resultCount = queryResult.analysisSummaries.reduce((acc, cur) => acc + cur.resultCount, 0);
```
charisk, line 321: Do we have this available in one of our helper functions somewhere now (or is it in a PR)?

charisk (approved):
LGTM, just some minor comments/thoughts.

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
charisk: We don't! Ignore me 😄

extensions/ql-vscode/src/history-item-label-provider.ts
elenatanasoiu: Makes sense. I've extracted a `pluralize` method here: https://github.com/github/vscode-codeql/pull/1432/commits/0577f762df76b83630097e973c81c8c470c3fc15

Let me know what you think.

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
elenatanasoiu: It was a good idea so I've extracted a method for that: https://github.com/github/vscode-codeql/pull/1432/commits/9078f411b9d112ad9887ed37c5daac69a72b5b15

PR #1433, opened 2022-07-18
Add a tree viewer UI for the evaluator logs

angelapwen:
This PR is the first iteration for a tree viewer for evaluator logs. All below feature are gated behind canary 🐦  

Upon extension startup, a container for the tree visualizer can be seen (similarly to the AST Viewer):
startup-container

There is a new command for each query run in the Query History view called "Show Evaluator Log (UI)":
viewer-command

When the command is clicked, the visualizer view for a single query run is created based off of the JSON parsed in #1405:
viewer

If the query run did not have any applicable data (ie. we hit the cache) the tree will simply say "No predicates evaluated in this query run.": 
no-predicates-viewer

The PR includes integration tests for the tree builder and viewer code, and relies on unit tests from #1405 to determine correctness. 

**Not in scope, for follow-up PRs:**

- suggested by @adityasharad: Cache the parsed data, without re-parsing each time we call `getRoots` similar to `astBuilder.getRoots()`
- Allow multiple query runs' logs to be visualized at once — some more design discussions needed on when trees are added to the visualizer (upon clicking the command?). At that point we would also want to distinguish two runs of the same query from one another — maybe with the evaluation start time. 
- suggested by @edoardopirovano: sort the tree view by time spent evaluating the predicate. This would make it easier to find predicates that are causing performance problems, which is often the goal of someone digging through this data.
- "Copy text" command — in case a user wants to search logs themselves and would like to get the predicate name easily.
- "Jump to query history item" command — to better materialize the relationship between query history items and visualizer items 
- "Jump to log" command — because this viewer is created with the JSON summary log, there are some design considerations around which log the user would like to jump to. Currently the viewer is only expected to view the JSON raw log and the human-readable summary log.

## Checklist


adityasharad:
>  🐞 / UI ❓: currently there is a button for "Show Evaluator Log Visualizer" in the view (see first screenshot above) that doesn't work when clicked. Does it make sense to just remove it? Should the button show the visualizer for the last query in the query history view?

What does the equivalent button for the AST viewer do when there is no source file selected? And can we conditionally hide/show suggestions in welcome views?

> UI ❓: the current design only allows for 1 query run's log to be visualized at a time. Ideally we would be able to view multiple query runs' visualizations at once (with each run being a root-level element of the tree). Is this worth doing in the current PR, or a follow-up?

Definitely a follow-up. We can always assume that there is a single run at a time. In practice that usually means only one query, but in future we can attempt to handle the case where the run involved an entire folder of queries.

> We do not have any tests 🧪 for this visualizer. Is it necessary to add tests for correctness given that the correctness of data parsed was already tested in https://github.com/github/vscode-codeql/pull/1405? What about UI tests?

Yes, I agree the correctness tests should go in the set of unit tests you already wrote. We don't have true UI tests for the extension. A middle ground worth adding in this PR is an integration test that creates a query run and attempts to call your new command -- see what we do for the AST viewer.

angelapwen:
> What does the equivalent button for the AST viewer do when there is no source file selected? And can we conditionally hide/show suggestions in welcome views?
> 
> >

It shows the following warning.
Screen Shot 2022-07-18 at 6 33 55 PM
The warning that the current change in this PR defaults to is the one that happens when the log is still being generated (something like the log is not ready yet), which is not ideal, but makes sense for the case where the log generation is still being run. That's why I thought it might make sense to just remove the button altogether here. I can't imagine too many cases where someone would want to click the button to visualize a log without selecting a specific run.


> Definitely a follow-up. We can always assume that there is a single run at a time. In practice that usually means only one query, but in future we can attempt to handle the case where the run involved an entire folder of queries.
> 
> >
Ok, that makes sense. In this situation we have the query showing up in each root element of the visualizer — the element has the form `$PREDICATE - $SIZE tuples in $TIME ms for query $QUERY`, but `$QUERY` is always the same. Perhaps it would be better to show the query name along with the log name in the top tree message?


> Yes, I agree the correctness tests should go in the set of unit tests you already wrote. We don't have true UI tests for the extension. A middle ground worth adding in this PR is an integration test that creates a query run and attempts to call your new command -- see what we do for the AST viewer.

I can look into adding these tomorrow 😸

angelapwen:
>  🐞: there is an error when attempting to load a new view of a query that wasn't the first query to be visualized after startup. The error is specifically around registering the command to clear the visualizer when it is already registered:

This bug is resolved — I worked out that the command is registered every time the visualizer's constructor was called, which was every time the command was run. I'm now initializing the visualizer upon extension startup. 

>  🐞: when loading the view, I get this error popup but not further information in the logs.

I am now no longer getting this error popup so marking this resolved for now.

angelapwen:
>  🐞 / UI ❓: currently there is a button for "Show Evaluator Log Visualizer" in the view (see first screenshot above) that doesn't work when clicked. Does it make sense to just remove it? Should the button show the visualizer for the last query in the query history view?

I couldn't think of a good reason for the button to be there other than that it looks slightly more aesthetically pleasing and would match the AST Viewer container more. I removed the button so that the container looks like this upon startup: 
container-startup

extensions/ql-vscode/src/vscode-tests/no-workspace/eval-log-visualizer.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    
+        expect((visualizer as any).treeDataProvider.roots).to.eq(roots);
+        expect((visualizer as any).treeView.message).to.eq(`Visualizer for ${evalLogPath}`);
+    });
+
+    it('should clear the visualizer\'s roots', () => {
+        // REVIEW: I can't call visualizer.clear() here unless I make the function public, but
+        // I'm not sure it's worth making it public just for the sake of testing. Or is there a 
+        // way to call the "Clear" command in these tests?
+    });
```
angelapwen, line 80: See comment

extensions/ql-vscode/src/vscode-tests/no-workspace/eval-log-tree-builder.test.ts
```diff
@@ -0,0 +1,56 @@
...
+            predicateName: 'quick_eval#query#ffffffff',
+            millis: 1,
+            resultSize: 596,
+            ra: 
+                    { pipeline: [
+                        '{1} r1',
+                        '{2} r2',
+                        'return r2'
+                    ] }
+        }
```
angelapwen, line 31: I'm having an issue with the formatting of the data in this integration test, again related to the ra field — the current format is what is actually passed into the tree builder and parsed (correctly); I have copied it from debugging. But this doesn't compile as `ra` expects a `Map` type rather than an object. 

When I use the following format which compiles:
```
ra: 
   new Map (Object.entries(
      { pipeline: [
           '...', '...', '...'
      ] }
    ))
```

then the code in the builder that actually parses the items is unable to parse the `Map` because it expects an `Object` that it then uses `Object.entries` on rather than a `Map`.

I'm not really sure how to resolve this issue to get the test to both run correctly and compile, but feel there should be a simple solution 🤔

adityasharad: As suggested elsewhere, I think changing `Map` to `Record` in the expected interface will resolve this. `Map` is a concrete class from JS (in particular it has a hash implementation), whereas `Record` is a more permissive interface from TS that can be satisfied by any object.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -904,7 +913,37 @@ export class QueryHistoryManager extends DisposableObject {
...
+  ) {
+    const { finalSingleItem, finalMultiSelect } = this.determineSelection(singleItem, multiSelect);
+    // Only applicable to an individual local query
+    if (!this.assertSingleQuery(finalMultiSelect) || !finalSingleItem || finalSingleItem.t !== 'local') {
+      return;
+    }
+
+    // If visualizer data in memory does exist, then build tree and display
+    if (finalSingleItem.evalLogVisualizerData) {
+      const evalLogTreeBuilder = new EvalLogTreeBuilder(finalSingleItem.evalLogVisualizerData);
```
adityasharad, line 933: My previous comment about caching is not relevant yet, because I see we are constructing a fresh builder here each time we run the command. Let's not complicate it further, but there might be room to make this faster in future.

extensions/ql-vscode/src/vscode-tests/no-workspace/eval-log-visualizer.test.ts
adityasharad: Try disposing the visualiser and then inspecting its roots? Otherwise just make the clear function public, with a comment that it's for testing.

adityasharad (commented):
Nice work. A lot of comments but I think you're on the right track.

The most complex part is the machinery for creating the tree builder and passing around the log items in memory -- I think that could be simplified (or cached), but it is better to revisit that separately.

extensions/ql-vscode/src/vscode-tests/no-workspace/eval-log-tree-builder.test.ts
angelapwen: Interesting, I didn't know about `Record` — this worked! Thank you.

extensions/ql-vscode/src/pure/log-summary-parser.ts
```diff
@@ -2,35 +2,31 @@ import * as os from 'os';
...
     predicateName: string;
     millis: number;
     resultSize: number;
-    ra: Pipelines;
-}
-  
-interface Pipelines {
     // Key: pipeline identifier; Value: array of pipeline steps 
-    pipelineNamesToSteps: Map<string, string[]>;
+    ra: Map<string, string[]>;
```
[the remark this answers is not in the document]
angelapwen: Made this change in a separate commit~

extensions/ql-vscode/package.json
```diff
@@ -533,6 +534,10 @@
         "command": "codeQLQueryHistory.showEvalLogSummary",
         "title": "Show Evaluator Log (Summary)"
       },
+      {
+        "command": "codeQLQueryHistory.showEvalLogVisualizer",
+        "title": "Show Evaluator Log (Visualizer)"
```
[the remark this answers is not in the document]
angelapwen: Just as a note, the summary file we are exposing to users (at least up until now) isn't the summary JSON file — it's the "human-readable" text version of the summary file. Our assumption was that users would prefer to have a easily consumable version of the summary. 

I _could_ change it to display the summary JSON file (which is used to build the UI) — this would align better with a future "jump to log" command in the UI. Right now, the log it would "jump to" would be neither one of the logs shown with other commands, which would be really confusing.

On the other hand, we could defer that change until the "jump to log" command is actually implemented. In that case maybe the commands could be `Raw JSON`, `Summary Text`, and `UI`?

extensions/ql-vscode/src/eval-log-tree-builder.ts
```diff
@@ -0,0 +1,52 @@
...
+        const roots: EvalLogTreeItem[] = [];
+
+        // For each item, create a TreeItem object with appropriate parents/children 
+        this.evalLogDataItems.forEach(logDataItem => {
+
+            // TODO(angelapwen): Once the viewer can show logs for multiple queries, then queries should be the root level.
+            // Every object is a root except the RA for now
+            const label = `${logDataItem.predicateName} - ${logDataItem.resultSize} tuples in ${logDataItem.millis} ms for query ${logDataItem.queryCausingWork}`;
+            const rootItem: EvalLogTreeItem = {
+                label: label,
```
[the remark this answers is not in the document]
angelapwen: 🧠 thank you!

extensions/ql-vscode/src/eval-log-tree-builder.ts
```diff
@@ -0,0 +1,52 @@
...
+                children: [] // Will assign pipeline items as children shortly.
+            };
+
+            for (const [pipelineName, steps] of Object.entries(logDataItem.ra)) {
+                const pipelineLabel = `Pipeline ID: ${pipelineName}`;
```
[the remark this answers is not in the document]
angelapwen: Many pipelines are simply named `pipeline` and I think `Pipeline pipeline` is a bit confusing. I'll try `Pipeline:` then?

extensions/ql-vscode/src/eval-log-visualizer.ts
```diff
@@ -0,0 +1,96 @@
...
+            : TreeItemCollapsibleState.None;
+        const treeItem = new TreeItem(element.label || '', state);
+        treeItem.tooltip = `${treeItem.label} || ''}`;
+        return treeItem;
+    }
+
+    getChildren(element?: EvalLogTreeItem): ProviderResult<EvalLogTreeItem[]> {
+        // If no item is passed, return the root.
+        if (!element) {
+            return this.roots ? this.roots : [];
```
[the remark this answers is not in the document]
angelapwen: Added in separate commit! Thanks

extensions/ql-vscode/src/eval-log-tree-builder.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    private async parseRoots(): Promise<EvalLogTreeItem[]> {
+        const roots: EvalLogTreeItem[] = [];
+
+        // For each item, create a TreeItem object with appropriate parents/children 
+        this.evalLogDataItems.forEach(logDataItem => {
+
+            // TODO(angelapwen): Once the viewer can show logs for multiple queries, then queries should be the root level.
+            // Every object is a root except the RA for now
```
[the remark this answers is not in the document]
angelapwen: Ah, I meant that the RA field (which creates a `pipeline` and `steps` children) is the only field that doesn't produce a root item for the tree. I see your other comment about making the root object the name of the query run, though, so this comment will be removed regardless.

extensions/ql-vscode/src/eval-log-tree-builder.ts
```diff
@@ -0,0 +1,52 @@
...
+        
+                steps.forEach( (step: string) => {
+                    const stepLabel = step;
+                    const stepItem: ChildEvalLogTreeItem =  {
+                        label: stepLabel,
+                        parent: pipelineItem,
+                        children: []
+                    };
+                    pipelineItem.children.push(stepItem);
+                });
```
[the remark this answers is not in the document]
angelapwen: Oops, I committed this but reverted because of a type issue. `steps.map(...)` returns a `void[]` but it needs to be a `ChildEvalLogTreeItem[]`. I tried a few different type coercions/ways to get each item in the array to be of type `ChildEvalLogTreeItem` but I'm not sure if I'm just not finding the right syntax or if this isn't possible 😄

edoardopirovano:
Firstly, very nice work! I played around with this a bit and it's super cool, very much looking forward to having this in a release. I haven't reviewed the code for this since Aditya left many useful comments that you're still working through. I do have a few pointers on polishing the UX after using this a bit, though. Please don't consider any of these blocking in any way, they're merely suggestions.

- The right click in Query History is gated behind the canary flag, but the panel is not. This is potentially quite confusing because the panel tells you to right click to populate it. We should hide either both or neither.
- If a run evaluated nothing (e.g. we hit the cache) we get a strangely empty panel with just the log name, maybe we could show something nicer here? I'm thinking of a message like `No predicates evaluated in this run.`
- I agree with your point above that showing the query name rather than the log file name at the top of the panel would be more helpful since the file is always just called `evaluator-log.jsonl`.
- There's lots of follow-up work I would consider. Along with the things you have already mentioned in the PR description, I would suggest that a hopefully fairly simple change that could be a big UX improvement would be an option to sort the tree view by time spent evaluating the predicate. This would make it easier to find predicates that are causing performance problems, which is often the goal of someone digging through this data.

extensions/ql-vscode/src/eval-log-tree-builder.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    private async parseRoots(): Promise<EvalLogTreeItem[]> {
+        const roots: EvalLogTreeItem[] = [];
+
+        // For each item, create a TreeItem object with appropriate parents/children 
+        this.evalLogDataItems.forEach(logDataItem => {
+
+            // TODO(angelapwen): Once the viewer can show logs for multiple queries, then queries should be the root level.
+            // Every object is a root except the RA for now
+            const label = `${logDataItem.predicateName} - ${logDataItem.resultSize} tuples in ${logDataItem.millis} ms for query ${logDataItem.queryCausingWork}`;
```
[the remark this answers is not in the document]
angelapwen: That makes sense. I have updated the code to make the root level the query being evaluated for now. It's not quite specific to an individual run, though — in the future if someone is hoping to compare/contrast two runs, the root labels would be the same. 

I don't think we have a unique identifier for the run, but perhaps something like `$QUERY at $START+TIME` would suffice? cc @edoardopirovano for additional UX input here 😸

angelapwen:
Thanks both for the reviews! I've marked 'resolved' the comments I've already addressed and I haven't forgotten about the remainder of them — working on them/will get to them soon.

I'll take a look at the UX suggestions from @edoardopirovano as well. I've addressed one: 
> I agree with your point above that showing the query name rather than the log file name at the top of the panel would be more helpful since the file is always just called evaluator-log.jsonl.

Since we decided to make the root-level of the tree the query name for now (in the future maybe query name + start time of run if a user decides to view multiple runs of the same query file at the same time), I made the 'message' just `Visualizer for query:` (in the future `queries`).

Eg: 
new-container-message

extensions/ql-vscode/package.json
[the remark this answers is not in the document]
adityasharad: Ah! I hadn't realised the summary was text. You are right of course. I like the distinction of `Raw JSON`, `Summary Text`, and `UI`.

angelapwen:
> The right click in Query History is gated behind the canary flag, but the panel is not. This is potentially quite confusing because the panel tells you to right click to populate it. We should hide either both or neither.
I've gated the panel/view for the visualizer behind the canary flag as well. It ended up being easy to do with the `when` context clause! Thanks for catching 😄

extensions/ql-vscode/src/eval-log-tree-builder.ts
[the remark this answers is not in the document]
adityasharad: Ugh my fault for not trying it out myself. Put an extra pair of parentheses around the `{}` object, so you get
```js
steps.map((step: string) => ({
  label: step,
  parent: pipelineItem,
  children: []
}));
```

Without the extra parentheses, I think the opening `{` is treated as starting the body of a function (which has no return statement, hence the `void` type), instead of starting an object literal like we intended.

(The `: string` type annotation is optional btw; TS will infer it.)

angelapwen:
> If a run evaluated nothing (e.g. we hit the cache) we get a strangely empty panel with just the log name, maybe we could show something nicer here? I'm thinking of a message like `No predicates evaluated in this run`.

Thanks for catching this ✨ I tried out a few different experiences, including simply having that message at the top of the panel, but decided to make it a child tree item of the root query item so that in the future it'll be easier to display it for multiple queries at a time. 

With predicates — 
Screen Shot 2022-07-21 at 10 59 44 AM

Without predicates — 
Screen Shot 2022-07-21 at 11 00 09 AM

While I did that, I realized that I could take the query name from the Extension rather than the log itself, and stopped parsing the query causing work from the log. Now the query names in the visualizer match those in the query history view which I think is nice (eg. it preserves whether it is a quick eval or not in the name, and which line), but without the information about the database it was run on and total query time. 

Another item to add to the future improvements list could be a jump-(back)-to-query-history-item command to make the relationship even more clear.

extensions/ql-vscode/package.json
```diff
@@ -608,6 +613,14 @@
           "light": "media/light/clear-all.svg",
           "dark": "media/dark/clear-all.svg"
         }
+      },
+      {
+        "command": "codeQLEvalLogVisualizer.clear",
```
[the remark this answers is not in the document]
angelapwen: 👍 I've done all the renaming internally now — I stuck with "viewer" for the container to match the AST viewer. The user-visible text always says "evaluator log" but internally I used "eval log" so the variable names weren't quite so long. 

The only user-visible text that doesn't say "Viewer" is the command we discussed below that now says "Show Evaluator Log (UI)".

Screen Shot 2022-07-21 at 11 57 01 AM

extensions/ql-vscode/package.json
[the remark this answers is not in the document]
angelapwen: Done 😄

extensions/ql-vscode/src/eval-log-tree-builder.ts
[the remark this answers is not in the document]
angelapwen: Ah, I see! I had thought that it was an object of type `void` because TypeScript couldn't figure out how to make it of type `ChildEvalLogTreeItem[]` but it was actually inferring that it was a function returning `void` 🤯 

have made the change now.

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -357,15 +357,15 @@ export class QueryEvaluationInfo {
...
     void cliServer.generateJsonLogSummary(this.evalLogPath, this.jsonEvalLogSummaryPath)
-      .then(() => {         
+      .then(() => {
         // TODO(angelapwen): Stream the file in. 
         fs.readFile(this.jsonEvalLogSummaryPath, (err, buffer) => {
           if (err) {
             throw new Error(`Could not read structured evaluator log summary JSON file at ${this.jsonEvalLogSummaryPath}.`);
           }
-          parseVisualizerData(buffer.toString()); // Eventually this return value will feed into the tree visualizer.
+          queryInfo.evalLogVisualizerData = parseVisualizerData(buffer.toString());
```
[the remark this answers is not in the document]
angelapwen: I ended up moving this method to `query-history.ts`, triggered only when the viewer command is clicked, so that I could avoid passing around all of the data in the `queryInfo` as you mentioned earlier. Now, the JSON summary log for the viewer is generated here (in canary mode) but it isn't read until the command is clicked. 

I think the parsing logic makes more sense in the Query History view anyway as well 😄

angelapwen:
I think I've addressed all comments above. The most recent screenshots in the PR comments should be the most up-to-date. (I will make sure to update the PR body with the final screenshots before merging to prevent future confusion about what changes went in in this PR). I also added all the follow-up issues/PRs that came up during this PR's review in the PR body to track for the future.

Marking for re-review now ✨ thank you!

extensions/ql-vscode/src/eval-log-tree-builder.ts
```diff
@@ -0,0 +1,66 @@
+import { ChildEvalLogTreeItem, EvalLogTreeItem } from './eval-log-viewer';
+import { EvalLogData as EvalLogData } from './pure/log-summary-parser';
+
+export default class EvalLogTreeBuilder {
```
adityasharad, line 4: Or something to this effect.
```suggestion
/** Builds the tree data for the evaluator log viewer for a single query run. */
export default class EvalLogTreeBuilder {
```

extensions/ql-vscode/src/eval-log-tree-builder.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    // Once the viewer can show logs for multiple queries, there will be more than 1 item at the root
+    // level. For now, there will always be one root (the one query being shown).
+    const queryItem: EvalLogTreeItem = {
+      label: this.queryName,
+      children: [] // Will assign predicate items as children shortly.
+    };
+
+    // Display descriptive message when no data exists 
+    if (this.evalLogDataItems.length == 0) {
```
adityasharad, line 28: Minor: I find it useful to avoid `==`, since there's a risk of surprise when an argument is `null` or `undefined`. Either `!this.evalLogDataItems.length` or `this.evalLogDataItems.length === 0` will do.

extensions/ql-vscode/src/eval-log-viewer.ts
```diff
@@ -0,0 +1,92 @@
...
+// Manages a tree viewer of structured evaluator logs.
+export class EvalLogViewer extends DisposableObject {
+  private treeView: TreeView<EvalLogTreeItem>;
+  private treeDataProvider: EvalLogDataProvider;
+
+  constructor() {
+    super();
+
+    this.treeDataProvider = new EvalLogDataProvider();
+    this.treeView = window.createTreeView('codeQLEvalLogViewer', {
```
adityasharad, line 58: Do we need to dispose the tree view when this object is disposed?

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -204,8 +203,9 @@ export class QueryEvaluationInfo {
         });
         if (await this.hasEvalLog()) {
           this.displayHumanReadableLogSummary(queryInfo, qs);
-          if (config.isCanary()) {
-            this.parseJsonLogSummary(qs.cliServer);
+          if (config.isCanary()) { // Generate JSON summary for viewer.
+            void qs.cliServer.generateJsonLogSummary(this.evalLogPath, this.jsonEvalLogSummaryPath);
```
adityasharad, line 207: How come we don't need to `await` this?

extensions/ql-vscode/src/run-queries.ts
```diff
@@ -353,26 +353,6 @@ export class QueryEvaluationInfo {
...
-        // TODO(angelapwen): Stream the file in. 
-        fs.readFile(this.jsonEvalLogSummaryPath, (err, buffer) => {
-          if (err) {
-            throw new Error(`Could not read structured evaluator log summary JSON file at ${this.jsonEvalLogSummaryPath}.`);
-          }
-          parseVisualizerData(buffer.toString()); // Eventually this return value will feed into the tree visualizer.
-        });
-      })
-      .catch(err => {
-        void showAndLogWarningMessage(`Failed to generate JSON structured evaluator log summary. Reason: ${err.message}`);
```
adityasharad, line 372: You lose this error handler, but I think that's covered by the CLI server's default error handling.

adityasharad (approved):
Great work! Some outstanding minor comments.

extensions/ql-vscode/src/eval-log-tree-builder.ts
angelapwen: Thanks! I was just thinking that I didn't add many class- or function-level comments this time around. I'll do a final check for other places where comments might be helpful before merging.

extensions/ql-vscode/src/run-queries.ts
angelapwen: Hm — I was thinking that because the viewer doesn't need to populate its data until the command is clicked, it might be fine to not `await` it for now (while the query is run) to help the query finish execution more quickly.

On the other hand, I can see that it may take a really long time to generate the log summary and then the user would repeatedly hit the error message that the data is not ready. I'll change it to `await` as this is under canary mode anyway. We can see if we get reports that the double summary generation is taking too long.

extensions/ql-vscode/src/run-queries.ts
angelapwen: I thought this would be ok given the user doesn't really know that there is a JSON structured log summary that the viewer is going to consume (it may unnecessarily consume them given that they can still view the raw JSON and summary text logs).

extensions/ql-vscode/src/eval-log-viewer.ts
adityasharad: Hah so it is!

PR #1444, opened 2022-07-21
Don't show parentheses when results are not yet fetched in Query History

elenatanasoiu:
We missed a place where we needed to check if results are present before attempting to show them.

## Before

Screenshot 2022-07-21 at 16 52 21

## After

Screenshot 2022-07-21 at 17 08 19

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
@@ -2,7 +2,7 @@ import { env } from 'vscode';
 import { expect } from 'chai';
 import { QueryHistoryConfig } from '../../config';
 import { HistoryItemLabelProvider } from '../../history-item-label-provider';
-import { CompletedLocalQueryInfo, CompletedQueryInfo, InitialQueryInfo } from '../../query-results';
+import { CompletedLocalQueryInfo, CompletedQueryInfo, InitialQueryInfo, QueryHistoryInfo } from '../../query-results';
 import { RemoteQueryHistoryItem } from '../../remote-queries/remote-query-history-item';
 
 
@@ -84,7 +84,7 @@ describe('HistoryItemLabelProvider', () => {
 
   describe('remote queries', () => {
     it('should interpolate query when user specified', () => {
-      const fqi = createMockRemoteQueryInfo('xxx');
+      const fqi = createMockRemoteQueryInfo({ userSpecifiedLabel: 'xxx' });
 
       expect(labelProvider.getLabel(fqi)).to.eq('xxx');
 
@@ -96,7 +96,7 @@ describe('HistoryItemLabelProvider', () => {
     });
 
     it('should interpolate query when not user-specified', () => {
-      const fqi = createMockRemoteQueryInfo();
+      const fqi = createMockRemoteQueryInfo({});
 
       expect(labelProvider.getLabel(fqi)).to.eq('xxx query-name (javascript) xxx');
 
@@ -109,14 +109,14 @@ describe('HistoryItemLabelProvider', () => {
     });
 
     it('should use number of repositories instead of controller repo if available', () => {
-      const fqi = createMockRemoteQueryInfo(undefined, 2);
+      const fqi = createMockRemoteQueryInfo({ repositoryCount: 2 });
 
       config.format = '%t %q %d %s %f %r %%';
       expect(labelProvider.getLabel(fqi)).to.eq(`${dateStr} query-name (javascript) 2 repositories in progress query-file.ql (16 results) %`);
     });
 
     it('should get query short label', () => {
-      const fqi = createMockRemoteQueryInfo('xxx');
+      const fqi = createMockRemoteQueryInfo({ userSpecifiedLabel: 'xxx' });
 
       // fall back on user specified if one exists.
       expect(labelProvider.getShortLabel(fqi)).to.eq('xxx');
@@ -126,7 +126,15 @@ describe('HistoryItemLabelProvider', () => {
       expect(labelProvider.getShortLabel(fqi)).to.eq('query-name');
     });
 
-    function createMockRemoteQueryInfo(userSpecifiedLabel?: string, repositoryCount?: number) {
+    function createMockRemoteQueryInfo({
+      resultCount = 16,
+      userSpecifiedLabel = undefined,
+      repositoryCount = 0
+    }: {
+      resultCount?: number;
+      userSpecifiedLabel?: string;
+      repositoryCount?: number;
+    }): QueryHistoryInfo {
       return {
         t: 'remote',
         userSpecifiedLabel,
@@ -142,7 +150,7 @@ describe('HistoryItemLabelProvider', () => {
           repositoryCount,
         },
         status: 'in progress',
-        resultCount: 16,
+        resultCount,
       } as unknown as RemoteQueryHistoryItem;
     }
   });
--- a/extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
@@ -138,7 +138,31 @@ describe('HistoryItemLabelProvider', () => {
       it('should skip displaying them', () => {
         const fqi = createMockRemoteQueryInfo({ resultCount: 0, repositoryCount: 2 });
         config.format = '%t %q %d %s %f %r %%';
-        expect(labelProvider.getLabel(fqi)).to.eq(`${dateStr} query-name (javascript) 2 repositories in progress query-file.ql  %`);
+        expect(labelProvider.getLabel(fqi)).to.eq(`${dateStr} query-name (javascript) 2 repositories in progress query-file.ql %`);
+      });
+    });
+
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
```diff
@@ -126,7 +126,35 @@ describe('HistoryItemLabelProvider', () => {
...
+        expect(labelProvider.getLabel(fqi)).to.eq(`${dateStr} query-name (javascript) 2 repositories in progress query-file.ql (16 results) %`);
+      });
+    });
+
+    describe('when results are not present', () => {
+      it('should skip displaying them', () => {
+        const fqi = createMockRemoteQueryInfo({ resultCount: 0, repositoryCount: 2 });
+
+        config.format = '%t %q %d %s %f %r %%';
+        expect(labelProvider.getLabel(fqi)).to.eq(`${dateStr} query-name (javascript) 2 repositories in progress query-file.ql  %`);
```
robertbrignull, line 143: Hmm, there's a double space at the end here. It shows up in the screenshots too, though it isn't too noticeable unless you're looking for it. I'm wondering how much we care about that. It's better than having the `()` but it's not perfect.

We could try moving the space after the paren into the replacement string and out of the format string. What do you think?

extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
```diff
@@ -126,7 +126,35 @@ describe('HistoryItemLabelProvider', () => {
...
+      });
+    });
+
+    interface mockRemoteQueryInterface {
```
charisk, line 147: Interface names should start with a capital letter.

I'm not sure why we're defining an interface here though, we could just say that the return type of `createMockRemoteQueryInfo` is `RemoteQueryHistoryItem`.

BTW (and this is nothing to do with your changes, I'm just sharing some thoughts) I'm not very keen on the approach we've been taking around "forcing" a type on an object when it's not entirely true. I think it can lead to confusing code. What I mean with that is that we do `as unknown as RemoteQueryHistoryItem` (and not just `as RemoteQueryHistoryItem`), presumably because the object doesn't define all the fields it needs in order to be a proper `RemoteQueryHistoryItem`. Then the user of this object will assume it's a proper `RemoteQueryHistoryItem` and could fail if trying to access one of its values. I think that's confusing from an engineer's perspective. We should set all the fields instead. If that's a pain then we need to build helpers around creating these objects.

extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
```diff
@@ -95,8 +95,8 @@ describe('HistoryItemLabelProvider', () => {
       expect(labelProvider.getLabel(fqi)).to.eq(`${dateStr} query-name (javascript) github/vscode-codeql-integration-tests in progress %::${dateStr} query-name (javascript) github/vscode-codeql-integration-tests in progress %`);
     });
 
-    it('should interpolate query when not user specified', () => {
-      const fqi = createMockRemoteQueryInfo();
+    it('should interpolate query when user not specified', () => {
```
shati-patel, line 98: Maybe I'm reading this incorrectly, but this change seems wrong 🤔 We're testing the case where the label is not user specified (not whether the user is specified).

The original is correct, though could probably be clearer! 🙈

shati-patel (commented):
Good spot! I hadn't noticed the empty parentheses either 😄 

+1 to the existing reviews, and just a tiny extra comment from me 🤏🏽

extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
elenatanasoiu: > Interface names should start with a capital letter.

Thanks 👍 

> we could just say that the return type of createMockRemoteQueryInfo is RemoteQueryHistoryItem.

In this case we were setting `userSpecifiedLabel` and `repositoryCount` which aren't present on `RemoteQueryHistoryItem`. 

But I agree this could be a lot cleaner.

extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
elenatanasoiu: Thanks @shati-patel! I have gone with `should interpolate query when not user-specified` in the end.

extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
elenatanasoiu: I had a look at moving the space into the replacement string. Unfortunately, the spaces are set extension-wide [via `this.config.format`](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/history-item-label-provider.ts#L26-L28). 

The config is built from whatever list of elements are passed to it, like we do [here](https://github.com/github/vscode-codeql/blob/2cf5b39cfeaf5306bd8ade30607c62112deb1823/extensions/ql-vscode/src/history-item-label-provider.ts#L81-L87):

```
private getRemoteInterpolateReplacements(item: RemoteQueryHistoryItem): InterpolateReplacements {
    return {
      t: new Date(item.remoteQuery.executionStartTime).toLocaleString(env.language),
      q: `${item.remoteQuery.queryName} (${item.remoteQuery.language})`,
      d: this.buildRepoLabel(item),
      r: `(${pluralize(item.resultCount, 'result', 'results')})`,
      s: item.status,
      f: path.basename(item.remoteQuery.queryFilePath),
      '%': '%'
    };
  }
```

and then builds something like what we define in the tests: 

```
config.format = '%t %q %d %r %s %f  %%';
```

elenatanasoiu:
Thanks for reviewing everyone!

extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
aeisenberg: We _could_ add some extra handling of spaces during interpolation. After the interpolated string is generated, we can then squash all whitespace into a single space. I don't see a use case for fine-grained control over whitespace. Any repeated whitespace is most likely due to odd interactions with interpolation, like this. So, I think squashing would give a nicer experience.

This is just like HTML does, which makes me think that since vscode is an electron app, do you know if duplicate spaces actually render in the UI?

extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
```diff
@@ -140,7 +140,7 @@ describe('HistoryItemLabelProvider', () => {
         const fqi = createMockRemoteQueryInfo({ resultCount: 0, repositoryCount: 2 });
 
         config.format = '%t %q %d %s %f %r %%';
-        expect(labelProvider.getLabel(fqi)).to.eq(`${dateStr} query-name (javascript) 2 repositories in progress query-file.ql  %`);
+        expect(labelProvider.getLabel(fqi)).to.eq(`${dateStr} query-name (javascript) 2 repositories in progress query-file.ql %`);
```
aeisenberg, line 143: Would be nice to have a new test that exercises all sorts of whitespace squashing. After that, looks good.

elenatanasoiu: Have added some extra tests.

aeisenberg (approved):

extensions/ql-vscode/src/vscode-tests/no-workspace/history-item-label-provider.test.ts
charisk: (feel free to merge the PR, just some thoughts from me - we could tidy up in a future PR)

 I'm really not following what this object is supposed to be. It seems to have a bunch of properties - some from `RemoteQueryHistoryItem` and some not. The function under test is `getLabel` and that takes a `QueryHistoryInfo` so I think that's what this object is supposed to be, but it has a few issues. I think the following would remove type errors, but I'm unsure why `in progress` is expected and not `InProgress` so that will likely fail:

```
function createMockRemoteQueryInfo({
  resultCount = 16,
  userSpecifiedLabel = undefined,
  repositoryCount = 0
}: {
  resultCount?: number;
  userSpecifiedLabel?: string;
  repositoryCount?: number;
}): QueryHistoryInfo {
  return {
    t: 'remote',
    userSpecifiedLabel,
    remoteQuery: {
      executionStartTime: date.getTime(),
      queryName: 'query-name',
      queryFilePath: 'query-file.ql',
      controllerRepository: {
        owner: 'github',
        name: 'vscode-codeql-integration-tests'
      },
      language: 'javascript',
      actionsWorkflowRunId: 123,
      queryText: 'query-text',
      repositoryCount,
    },
    resultCount,
    completed: false,
    queryId: 'query-id'
  };
}
```

elenatanasoiu: Thank for this @charisk! 

I've added `QueryHistoryInfo` as a return type without the extra fields and the tests run fine. 

Oddly it didn't complain about missing fields. 🤔

PR #1454, opened 2022-08-04
Report suspicious join orders

dbartol:
This PR adds support for scanning the evaluator log to detect suspicious join orders that may be causing performance problems in a query. The actual detection algorithm is cut-and-pasted, and slightly refactored, from DCA. I've attempted to keep the portions of the code that scan the log clean of any dependencies on VS Code, so that we can eventually share the relevant code between the extension and DCA.

The algorithm computes a numeric metric for each relation in the RA, attempting to estimate how expensive the actual join order is compared to an ideal join order. Based on usage in DCA, we've set the threshold metric at 50. Anything higher than that is flagged as suspicious. In practice, most good join orders tend to have metrics in the single digits, and most bad join orders have metrics a couple orders of magnitude higher than that.

Relevant classes/interfaces include:

### Non-VSCode types/functions
These types don't depend on VS Code, so could be reused in DCA to share much of the implementation.

- `EvaluationLogScanner` - Handles events from scanning a single JSON log summary
- `EvaluationLogScannerProvider` - Factory for `EvaluationLogScanner`s.
- `EvaluationLogScannerSet` - Registry of scanner providers. When a log is scanned, this creates an instance of each registered scanner and passes the individual log records to each scanner.
- `EvaluationLogProblemReporter` - Callback interface used by scanners to report errors/warnings.
- `JoinOrderScannerProvider` - Implementation of the scanner interfaces that detects suspicious join orders.
- `generateSummarySymbolsFile()` - Generates a JSON file that maps the name of each relation from a human-readable summary to the location of its RA. This currently depends on the exact format of the human-readable summary, so I'd like to move this into the CLI itself once it's working well.
- `readJsonlFile()` - Reads a file in our particular flavor of "human-readable JSONL", invoking a callback for each top-level object. Note that true JSONL separates objects with single newline, requiring each object to be on a single line. Our flavor separates objects with a double newline, so that single newlines can be used within an object. This function is currently just a minor refactoring of what we already used to parse the logs for the log viewer, so it still reads the entire file into memory at once. The interface will allow us to change the implementation to stream objects, however.

### VS Code types
- `LogScannerService` - Connects an `EvaluationLogScannerSet` to all the right VS Code events and UI. It scans the log whenever the current query history item changes, or when the current query completes. All reported problems are routed to the `Problems` view in VS Code.

The best I was able to come up with for testing was a unit test that scans a log fixture and expects the proper warning.


dbartol:
Note to reviewers: The actual implementation of the join order metric probably doesn't need a very thorough review, since it's just a slightly reorganized version of the existing DCA code. @MathiasVP you might want to take a look at `join-order.ts` just to see what I did, though.

extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -0,0 +1,459 @@
...
+            }
+          })
+          .map((prev): [string, number] => {
+            const size = this.prevDeltaSizes(event, prev, iteration);
+            return [prev, size];
+          })
+      );
+    }
+
+    const resultSizes = inLayerEvent.resultSize;
```
MathiasVP, line 399: This turns out to be incorrect. We just merged a [fix](https://github.com/github/codeql-dca/pull/984) on this line a couple of days ago. This fix should most likely be ported to this file as well.

extensions/ql-vscode/src/log-insights/jsonl-reader.ts
```diff
@@ -0,0 +1,23 @@
...
+ * it will stream the document to improve the performance with large documents.
+ *
+ * @param path The path to the file.
+ * @param handler Callback to be invoked for each top-level JSON object in order.
+ */
+export async function readJsonlFile(path: string, handler: (value: any) => Promise<void>): Promise<void> {
+  const logSummary = await fs.readFile(path, 'utf-8');
+
+  // Remove newline delimiters because summary is in .jsonl format.
+  const jsonSummaryObjects: string[] = logSummary.split(/\r?\n\r?\n/g);
```
aeisenberg, line 17: I _think_ this will help with memory, but maybe the runtime is smart enough to know that `logSummary` isn't used any more and can be garbage collected.

```suggestion
  const jsonSummaryObjects: string[] = logSummary.split(/\r?\n\r?\n/g);
  delete logSummary;
```

extensions/ql-vscode/src/log-insights/log-scanner-service.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+function predicateSymbolKey(name: string, raHash: string): string {
+  return `${name}@${raHash.substring(0, 8)}`;
+}
+
+/**
+ * Implementation of `EvaluationLogProblemReporter` that generates `Diagnostic` objects to display
+ * in the VS Code "Problems" view.
+ */
+class ProblemReporter implements EvaluationLogProblemReporter {
```
aeisenberg, line 24: Not sure if you've already implemented this, but there will need to be a command to clear out these problems, or else they will be stuck in the problems view.

extensions/ql-vscode/src/log-insights/log-scanner-service.ts
```diff
@@ -0,0 +1,102 @@
...
+export class LogScannerService extends DisposableObject {
+  public readonly scanners = new EvaluationLogScannerSet();
+  private readonly diagnosticCollection = this.push(languages.createDiagnosticCollection('ql-eval-log'));
+  private currentItem: QueryHistoryInfo | undefined = undefined;
+
+  constructor(qhm: QueryHistoryManager) {
+    super();
+
+    this.push(qhm.onDidChangeCurrentQueryItem(async (item) => {
+      if (item !== this.currentItem) {
```
aeisenberg, line 53: This will only work for local queries, right? So, need to make sure this is local and not remote. The type is `LocalQueryInfo`. So make sure to check for the `t` property being correct.

extensions/ql-vscode/src/log-insights/log-scanner-service.ts
```diff
@@ -0,0 +1,102 @@
...
+      const range = new Range(predicateInfo.raStartLine, 0, predicateInfo.raEndLine + 1, 0);
+      this.diagnostics.push(new Diagnostic(range, message, DiagnosticSeverity.Error));
+    }
+  }
+}
+
+export class LogScannerService extends DisposableObject {
+  public readonly scanners = new EvaluationLogScannerSet();
+  private readonly diagnosticCollection = this.push(languages.createDiagnosticCollection('ql-eval-log'));
+  private currentItem: QueryHistoryInfo | undefined = undefined;
```
aeisenberg, line 47: I think you want this:

```suggestion
  private currentItem: LocalQueryInfo | undefined = undefined;
```

extensions/ql-vscode/src/log-insights/log-scanner-service.ts
```diff
@@ -0,0 +1,102 @@
...
+   * Scan the evaluation log for a query, and report any diagnostics.
+   *
+   * @param query The query whose log is to be scanned.
+   */
+  public async scanEvalLog(
+    query: QueryHistoryInfo | undefined
+  ): Promise<void> {
+    this.diagnosticCollection.clear();
+
+    if ((query === undefined) || (query.t !== 'local') || (query.jsonEvalLogSummaryLocation === undefined)) {
```
aeisenberg, line 76: Ah..I see you're taking care of local queries here. Either way is fine.

```suggestion
    if (query?.t !== 'local' || query.jsonEvalLogSummaryLocation === undefined) {
```

extensions/ql-vscode/src/log-insights/log-scanner-service.ts
```diff
@@ -0,0 +1,102 @@
...
+      return;
+    }
+
+    const diagnostics = await this.scanLog(query.jsonEvalLogSummaryLocation, query.evalLogSummarySymbolsLocation);
+    const uri = Uri.file(query.evalLogSummaryLocation!);
```
aeisenberg, line 81: Can `query.evalLogSummaryLocation` be undefined?

extensions/ql-vscode/src/log-insights/summary-parser.ts
```diff
@@ -0,0 +1,114 @@
...
+    lineNumber++;
+    const startLine = lines[startLineNumber];
+    const nonRecursiveMatch = startLine.match(NON_RECURSIVE_TUPLE_COUNT_REGEXP);
+    let predicateName: string | undefined = undefined;
+    let iteration = 0;
+    if (nonRecursiveMatch) {
+      predicateName = nonRecursiveMatch.groups!.predicateName;
+    } else {
+      const recursiveMatch = startLine.match(RECURSIVE_TUPLE_COUNT_REGEXP);
+      if (recursiveMatch) {
```
aeisenberg, line 78: Avoid the `!`?

```suggestion
      if (recursiveMatch.?groups) {
```

extensions/ql-vscode/src/log-insights/summary-parser.ts
```diff
@@ -0,0 +1,114 @@
...
+      while ((lineNumber < lines.length) && (raEndLine === undefined)) {
+        const raLine = lines[lineNumber];
+        const returnMatch = raLine.match(RETURN_REGEXP);
+        if (returnMatch) {
+          raEndLine = lineNumber;
+        }
+        lineNumber++;
+      }
+      if (raEndLine === undefined) {
+        raEndLine = lineNumber - 1;
```
aeisenberg, line 96: This means that `raEndLine` < `raStartLine`. Is that ok?

extensions/ql-vscode/src/query-history.ts
```diff
@@ -260,7 +266,10 @@ export class HistoryTreeDataProvider extends DisposableObject {
   }
 
   setCurrentItem(item?: QueryHistoryInfo) {
-    this.current = item;
+    if (item !== this.current) {
+      this.current = item;
+      this._onDidChangeCurrentQueryItem.fire(item);
```
aeisenberg, line 271: Looks like this is a bug that wasn't noticed before.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -926,7 +951,7 @@ export class QueryHistoryManager extends DisposableObject {
     }
 
     // Summary log file doesn't exist.
-    if (finalSingleItem.evalLogLocation && fs.pathExists(finalSingleItem.evalLogLocation)) {
+    if (finalSingleItem.evalLogLocation && await fs.pathExists(finalSingleItem.evalLogLocation)) {
```
aeisenberg, line 954: We should have had a linter error because of this. Good catch.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -950,15 +975,14 @@ export class QueryHistoryManager extends DisposableObject {
...
 
-    // TODO(angelapwen): Stream the file in. 
-    void fs.readFile(finalSingleItem.jsonEvalLogSummaryLocation, async (err, buffer) => {
-      if (err) {
-        throw new Error(`Could not read evaluator log summary JSON file to generate viewer data at ${finalSingleItem.jsonEvalLogSummaryLocation}.`);
-      }
-      const evalLogData: EvalLogData[] = parseViewerData(buffer.toString());
+    // TODO(angelapwen): Stream the file in.
+    try {
+      const evalLogData: EvalLogData[] = await parseViewerData(finalSingleItem.jsonEvalLogSummaryLocation);
```
aeisenberg, line 980: This changes from a callback to awaiting processing of the summary log. Which means that `handleShowEvalLogViewer` won't complete until the log is fully processed. That seems ok to me, but @angelapwen should look at this since she originally implemented this.

aeisenberg (commented):
Lots of little comments. Nothing concerning.

extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -0,0 +1,459 @@
...
+          const metric = computeJoinOrderBadness(maxTupleCount, maxDependentPredicateSize, resultSize!);
+          this.joinOrderMetricMap.set(key, metric);
+          this.reportProblemIfNecessary(event, 0, metric);
+        }
+        break;
+      }
+
+      case 'COMPUTE_RECURSIVE': {
+        // Compute the badness metric for a recursive predicate for each ordering.
+        // See https://github.com/github/codeql-coreql-team/issues/1289#issuecomment-1007237055 for
```
MathiasVP, line 238: Since this is now code in a public repo we should probably not link to an internal GitHub issue. We could write out the definition here if you think it's helpful for future readers

extensions/ql-vscode/src/query-history.ts
angelapwen: Thanks!! Didn't notice this either.

extensions/ql-vscode/src/query-history.ts
angelapwen: I _think_ this seems fine; it makes sense that the user would expect the command to complete once the log is processed.

extensions/ql-vscode/src/log-insights/jsonl-reader.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Read a file consisting of multiple JSON objects. Each object is separated from the previous one
+ * by a double newline sequence. This is basically a more human-readable form of JSONL.
+ *
+ * The current implementation reads the entire text of the document into memory, but in the future
+ * it will stream the document to improve the performance with large documents.
+ *
+ * @param path The path to the file.
+ * @param handler Callback to be invoked for each top-level JSON object in order.
+ */
```
angelapwen, line 12: This is great for future work! Thanks.

extensions/ql-vscode/src/log-insights/summary-parser.ts
```diff
@@ -0,0 +1,114 @@
...
+
+// Tuple counts for Expr::Expr::getParent#dispred#f0820431#ff@76d6745o:
+const NON_RECURSIVE_TUPLE_COUNT_REGEXP = /^Evaluated relational algebra for predicate (?<predicateName>\S+) with tuple counts:$/;
+// Tuple counts for Expr::Expr::getEnclosingStmt#f0820431#bf@923ddwj9 on iteration 0 running pipeline base:
+const RECURSIVE_TUPLE_COUNT_REGEXP = /^Evaluated relational algebra for predicate (?<predicateName>\S+) on iteration (?<iteration>\d+) running pipeline (?<pipeline>\S+) with tuple counts:$/;
+const RETURN_REGEXP = /^\s*return /;
+
+/**
+ * Parse a human-readable evaluation log summary to find the location of the RA for each pipeline
+ * run.
```
angelapwen, line 38: Interesting that we now have 2 parser implementations of the evaluation log summary (one for the JSON version that I used for the tree viewer and this one for human-readable). I wonder if it's worth combining them later on.

extensions/ql-vscode/src/query-history.ts
dbartol: It's a new event, and the previous code just had a potentially redundant field assignment. I don't think there was a real problem.

extensions/ql-vscode/src/log-insights/summary-parser.ts
dbartol: This parser for the human-readable log is really just to figure out where _in the human-readable log_ each relation is. We couldn't really tell that by parsing the JSON version. My plan (and I mention this in a comment somewhere) is to move the code that generates the "symbols" file into the CLI as part of generating the human-readable summary, where we'll already know the location of each relation as we emit it.

edoardopirovano: I internally screamed a little when I read that we were parsing the human-readable log. Having looked at it, it's not actually too bad though, since the scope of what we're parsing is pretty limited and there's a clear path to get rid of that parsing too. 👍🏼

extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -0,0 +1,459 @@
...
+    // id AND NOT id
+    String.raw`([0-9a-zA-Z:#_]+)\s+AND\s+NOT\s+([0-9a-zA-Z:#_]+)`,
+    // INVOKE HIGHER-ORDER RELATION rel ON <id, ..., id>
+    String.raw`INVOKE\s+HIGHER-ORDER\s+RELATION\s[^\s]+\sON\s+<([0-9a-zA-Z:#_<>]+)((?:,[0-9a-zA-Z:#_<>]+)*)>`,
+    // SELECT id
+    String.raw`SELECT\s+([0-9a-zA-Z:#_]+)`
+  ];
+  const r = new RegExp(
+    `${String.raw`\{[0-9]+\}\s+[0-9a-zA-Z]+\s=\s(?:` + regexps.join('|')})`
+  );
```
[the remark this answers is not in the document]
dbartol: Fixed.

extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -0,0 +1,459 @@
...
+  return I.List(operations).flatMap(operation => {
+    const matches = r.exec(operation.trim());
+    return I.List(matches!)
```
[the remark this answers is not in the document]
dbartol: Fixed.

extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -0,0 +1,459 @@
...
+      .rest() // Skip the first group as it's just the entire string
+      .filter(x => !!x && !x.match('r[0-9]+|PRIMITIVE')) // Only keep the references to predicates.
```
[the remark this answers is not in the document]
dbartol: This turned out to be subtly different logic in the case where `x` is `null`, so I reverted back to the original code.

extensions/ql-vscode/src/log-insights/summary-parser.ts
dbartol: I will note that between the time I originally wrote this code a month or so ago, and the time I opened the PR, the human-readable format had changed.

extensions/ql-vscode/src/log-insights/summary-parser.ts
dbartol: Fixed

extensions/ql-vscode/src/log-insights/summary-parser.ts
dbartol: I've changed the code to just skip creating the symbol if we hit EOF before seeing the `return`. Should only happen for invalid logs, and won't be an issue once the symbol generation moves into the CLI.

extensions/ql-vscode/src/log-insights/log-scanner-service.ts
dbartol: The current item can be a remote query. We just won't generate any diagnostics for a remote query. This is the simplest way to ensure that the problems being displayed always reflect the current history item (local or remote).

extensions/ql-vscode/src/log-insights/jsonl-reader.ts
dbartol: I'll just make a follow-up PR that does proper streaming. It's not that hard, but I didn't want to add more to this PR.

extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    const resultSizes = inLayerEvent.resultSize;
+    return { dependentPredicateSizes, resultSizes };
+  }
+
+  /**
+   * Compute the metric input for all the events in a SCC that starts with main node `event`
+   */
+  private badnessInputsForRecursiveDelta(event: ComputeRecursive) {
```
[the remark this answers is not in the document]
dbartol: Added

extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -0,0 +1,459 @@
...
+
+  /**
+   * Compute the maximum tuple count and maximum dependent predicate size for a non-recursive pipeline
+   */
+  private badnessInputsForNonRecursiveDelta(
+    pipelineRun: PipelineRun,
+    event: ComputeSimple
+  ): { maxTupleCount: number; maxDependentPredicateSize: number } {
+    const dependentPredicateSizes = Object.values(event.dependencies).map(hash =>
+      this.predicateSizes.get(hash)!  // REVIEW: '!'
```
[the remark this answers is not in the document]
dbartol: Fixed. We'll just treat it as zero.

extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -0,0 +1,459 @@
...
+    // If an iteration isn't present in the map it means it was skipped because the optimizer
+    // inferred that it was empty. So its size is 0.
+    return this.curDeltaSizes(event, predicate, i - 1);
+  }
+
+  private curDeltaSizes(event: ComputeRecursive, predicate: string, i: number) {
+    // If an iteration isn't present in the map it means it was skipped because the optimizer
+    // inferred that it was empty. So its size is 0.
+    return (
+      this.layerEvents.get(event.raHash)!.find(x => x.predicateName === predicate)?.deltaSizes[i] || 0
```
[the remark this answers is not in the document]
dbartol: Fixed to treat it as zero, which was already the fallback for failed lookups in the returned array.

extensions/ql-vscode/src/log-insights/log-scanner-service.ts
dbartol: Whenever the current history item changes (include to `undefined`), we clear the problems unconditionally before we conditionally scan for any problems with the new history item.

extensions/ql-vscode/src/log-insights/log-scanner-service.ts
dbartol: See below.

extensions/ql-vscode/src/log-insights/log-scanner-service.ts
dbartol: Probably not, but I've removed the `!` and added a check for `undefined` to the existing guard condition above.

extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -0,0 +1,459 @@
...
+  const result = new Int32Array(length);
+  for (let i = 0; i < length; i++) {
+    const ai = a[i] || 0;
+    const bi = b[i] || 0;
+    // -1 is used to represent the absence of a tuple count for a line in the pretty-printed RA (e.g. an empty line), so we ignore those.
+    if (i < a.length && i < b.length && (ai === -1 || bi === -1)) {
+      result[i] = -1;
+      reportIfInconsistent(ai, bi);
+    } else {
+      result[i] = ai + bi;
```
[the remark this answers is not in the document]
dbartol: I think so, but I don't think it causes any problems.

extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -0,0 +1,459 @@
...
+  }
+}
+
+/**
+ * Sum arrays a and b element-wise, and pad with 0s if the arrays are not the same length.
+ */
+function pointwiseSum(a: Int32Array, b: Int32Array): Int32Array {
+  function reportIfInconsistent(ai: number, bi: number) {
+    if (ai === -1 && bi !== -1) {
+      console.warn(
```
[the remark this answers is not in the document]
dbartol: I've rerouted them through a new function on `EvaluationLogProblemReporter`. For VS Code, they'll go through our regular logger.

extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -0,0 +1,459 @@
...
+        `Operation was not evaluated in the first pipeline, but it was evaluated in the accumulated pipeline (with tuple count ${bi}).`
```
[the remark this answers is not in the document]
dbartol: I think it's more trouble than it's worth, since this isn't a message a user is going to take action on.

extensions/ql-vscode/src/log-insights/join-order.ts
dbartol: I've removed the reference. It might be helpful to write out the definition, but I'm not sure that I can write it accurately based on the contents of the internal issue.

extensions/ql-vscode/src/log-insights/join-order.ts
dbartol: Applied the fix.

dbartol:
I believe that I've addressed or responded to all feedback. Should be ready for re-review.

extensions/ql-vscode/src/log-insights/join-order.ts
MathiasVP: No worries. I'll be happy to open a PR with the precise definition in a comment block once we've merged this one.

extensions/ql-vscode/src/log-insights/join-order.ts
MathiasVP: Great! The fix LGTM 👍.

MathiasVP:
The bad-join-order-detection code in join-order.ts LGTM now! In fact, I hope we can backport some of these lovely improvements to DCA at some point (or even share some of the code in a glorious future)

aeisenberg (approved):
Nice work! LGTM.

PR #1004, opened 2021-11-15
Fix large sarif handling

marcnjaramillo:
Replace this with a description of the changes your pull request makes.

## Checklist

- [x] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [x] Issues have been created for any UI or other user-facing changes made by this pull request.
- [x] `@github/docs-content-codeql` has been cc'd in all issues for UI or other user-facing changes made by this pull request.

PR #1457, opened 2022-08-09
Use sarif parser for reopened results

angelapwen:
When we attempt to re-open query results, instead of using the streaming SARIF parser written in #1004, we used `JSON.parse()`. This couldn't handle large SARIF files and we ran into an error.

This change uses the streaming SARIF parser instead. Note that this is a draft pull request as it is untested locally. cc @aschackmull (sorry, wrong Anders tagged at first!)

## Checklist


extensions/ql-vscode/src/query-results.ts
```diff
@@ -160,10 +161,12 @@ export async function interpretResultsSarif(
   sourceInfo?: cli.SourceInfo
 ): Promise<SarifInterpretationData> {
   const { resultsPath, interpretedResultsPath } = resultsPaths;
+  let res;
   if (await fs.pathExists(interpretedResultsPath)) {
-    return { ...JSON.parse(await fs.readFile(interpretedResultsPath, 'utf8')), t: 'SarifInterpretationData' };
+    res = await sarifParser(interpretedResultsPath);
```
adityasharad, line 166: I think you still need the extra `t: 'SarifInterpretationData'` property, so the results view knows what to do with this object. See what we do just below.

angelapwen: I am appending it a few lines down in the `return` statement — does it need to be assigned to `res` here?

angelapwen:
Looks like there are some genuine unit test errors on the method I changed so this will need more 🕵️ work

extensions/ql-vscode/src/query-results.ts
adityasharad: Gah didn't read the diff closely enough. You have done exactly that!

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
```diff
@@ -216,7 +219,81 @@ describe('query-results', () => {
...
+
+    await expect(
+      interpretResultsSarif(
+        mockServer,
+        metadata,
+        {
+          resultsPath, interpretedResultsPath
+        },
+        sourceInfo as SourceInfo)
+    ).to.be.rejectedWith('Parsing output of interpretation failed: Invalid SARIF file: expecting at least one run with result.');
```
angelapwen, line 293: @aeisenberg: this test is failing locally for me and I was wondering if you could help me understand why. I'm writing to the filestream with the same loop as above. 

Instead of getting the error we expect about how we expect at least one run with result, we get the error

```
Parser cannot parse input: unexpected characters
```
which indicates that we hit this error prior to hitting EOF, where the error I've added in this PR would be thrown.

Also of note, when I decrease the number of iterations through the loop (even, say, to a single iteration), the error changes to:

```
Parser cannot parse input: expected a value
```

rather than the error we expect, even though the file we've written should be functionally the same as the file in the 4th test. This makes me think it's something to do with the write stream, but I'm surprised that it works fine for the last test (invalid large SARIF file). 

Let me know if you have any pointers on this ☝️ 😸

angelapwen: I think I'm having success with using `fs.unlink()` to delete the previous large SARIF file. (Not sure why overwriting it with an option of `w` passed to the `WriteStream` wasn't doing the job, but I'll take it!)

adityasharad: I think we also have `del` for this purpose, which may be more robust at deletion.

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
```diff
@@ -216,7 +219,88 @@ describe('query-results', () => {
...
```
angelapwen, line 300: This invalid large SARIF file test is not rejecting as expected from the parser (which is what we'd expect, similar to the small invalid file), but rather returning empty/undefined and `interpretResultsSarif` is appending and returning `{ t: 'SarifInterpretationData' }` alone. Will investigate further.

angelapwen: Hm.. it may still be related to some issue with the `WriteStream`. When I only write a single object to the file, which should make it identical to test 4, I still get this result rather than the error thrown.

```
    for (let i = 0; i < 1; i++) {
      invalidSarifStream.write(JSON.stringify({
        a: '6'
      }), 'utf8');
    }
```

(using this loop)

extensions/ql-vscode/src/query-results.ts
aeisenberg, line 166: Just a thought....if this function throws an error because the file is not parseable, should we delete the file and re-run `cli.interpretBqrsSarif`?

Possibly, we can do this as a followup.

extensions/ql-vscode/src/sarif-parser.ts
```diff
@@ -1,33 +1,34 @@
 import * as Sarif from 'sarif';
 import * as fs from 'fs-extra';
-import { parser } from 'stream-json';
-import { pick } from 'stream-json/filters/Pick';
 import Assembler = require('stream-json/Assembler');
-import { chain } from 'stream-chain';
 import { getErrorMessage } from './pure/helpers-pure';
+import Pick = require('stream-json/filters/Pick');
```
aeisenberg, line 5: Is there a reason why you are using `require` instead of standard imports?

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
```diff
@@ -216,7 +219,88 @@ describe('query-results', () => {
...
+      {
+        resultsPath, interpretedResultsPath
+      },
+      sourceInfo as SourceInfo
+    );
+    // We do not re-interpret if we are reading from a SARIF file. 
+    expect(spy).to.not.have.been.called;
+
+    expect(results5).to.have.property('t', 'SarifInterpretationData');
+    expect(results5).to.have.nested.property('runs[0].results');
```
aeisenberg, line 274: Nice....I didn't know about this assertion.

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
aeisenberg: 😕

extensions/ql-vscode/src/sarif-parser.ts
angelapwen: Hm, I had thought that I needed to import it this way as a namespace to use the `Pick.withParser()` method (similar to the way `Assembler` was imported to use `Assembler.connectTo`

extensions/ql-vscode/src/query-results.ts
angelapwen: Hm, that's a good point 👍 I'll leave this comment up for now, while we work out the last invalid SARIF test issue..

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
angelapwen: I had to find it out of necessity because it would've been much too long to write out the full object for `deepEqual` in this case 😸

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
angelapwen: Replaced with `del` 👍 (but it didn't fix the other large SARIF test problem unfortunately)

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
angelapwen: It's really odd, I had thought deleting the file resolved this issue, but it just creates a different error (not throwing any parsing error at all rather than throwing unexpected parsing errors). But it's definitely something related to file system writing 🤔

extensions/ql-vscode/src/sarif-parser.ts
aeisenberg: I think this will work:

```suggestion
import * as pick from 'stream-json/filters/Pick';
```

Or possibly better:

```suggestion
import { withParser } from 'stream-json/filters/Pick';
```

angelapwen: Ah yes.. i'll do the same for `connectTo` from `Assembler` as well.

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
```diff
@@ -155,68 +156,193 @@ describe('query-results', () => {
...
-    );
-    expect(results3).to.deep.eq({ a: 6, t: 'SarifInterpretationData' });
+    afterEach(async () => {
+      sandbox.restore();
+      await safeDel(interpretedResultsPath);
+    });
+
+    it('should interpretResultsSarif', async function() {
+      // up to 2 minutes per test
+      this.timeout(2 * 60 * 1000);
```
aeisenberg, line 187: Can you move this (and the other) timeout to the describe block above? You only need to set the timeout once up there.

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
```diff
@@ -364,4 +490,12 @@ describe('query-results', () => {
...
     return fqi;
   }
+
+  function safeDel(file: string) {
+    try {
+      fs.unlinkSync(file);
+    } catch (e) {
+      // ignore
+    }
+  }
```
aeisenberg, line 500: I have a PR up that removes this from `queries.test.ts`. As I explain above, can you use `del`?

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
```diff
@@ -155,68 +156,193 @@ describe('query-results', () => {
...
-      metadata,
-      {
-        resultsPath, interpretedResultsPath
-      },
-      sourceInfo as SourceInfo
-    );
-    expect(results3).to.deep.eq({ a: 6, t: 'SarifInterpretationData' });
+    afterEach(async () => {
+      sandbox.restore();
+      await safeDel(interpretedResultsPath);
```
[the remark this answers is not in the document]
angelapwen: Actually the test was failing in CI (not locally) when I used `del` — it said that the file wasn't found when it tried to delete it. See https://github.com/github/vscode-codeql/actions/runs/3285438706/jobs/5412531729

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
angelapwen: I get an `Object is possibly undefined` error when I move it to the describe block or the `beforeEach` block — is there a way to get around that error?

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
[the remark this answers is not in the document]
angelapwen:
```
1) query-results
       interpretResultsSarif
         "after each" hook for "should use sarifParser on a valid small SARIF file":
     Error: Cannot delete files/directories outside the current working directory. Can be overridden with the `force` option.
      at safeCheck (node_modules/del/index.js:37:9)
      at mapper (node_modules/del/index.js:73:4)
      at /home/runner/work/vscode-codeql/vscode-codeql/extensions/ql-vscode/node_modules/p-map/index.js:57:28
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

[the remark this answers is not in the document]
angelapwen: I'm going to remove the `del` call entirely and see if it works, given that I am over-writing the existing file when the `WriteStream` is created.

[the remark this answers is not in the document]
angelapwen: Removing `del` altogether worked on ubuntu. There is still a failure in this same test in Windows

```
[main 2022-10-20T17:11:53.979Z] Extension host with pid 6888 exited with code: 134, signal: null.

Test runner caught exception (Failed)
Exit code:   134
Done
Tried running suite 3 time(s), still failed, giving up.

Error: Process completed with exit code 1.
```

that isn't very descriptive. Exit code 134 seems to be a `SIGABRT`. I seem to remember there being some intricacies around Windows and file I/O — maybe I'm hitting one of them here?

aeisenberg:
I can't tell why the windows tests are failing. It seems to be failing consistently. If you have access to a windows box, maybe try running the tests there.

angelapwen:
> I can't tell why the windows tests are failing. It seems to be failing consistently. If you have access to a windows box, maybe try running the tests there.

I don't have one, but will 🍐 with Dave tomorrow on this!

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
aeisenberg: Hmmm...I had to remove `del` from my PR sine it was hanging. Apologies for the back and forth, but using `safeDel` here is fine.

angelapwen: Ugh... still getting that Windows failure

```
[main 2022-10-20T20:56:10.030Z] Extension host with pid 756 exited with code: 134, signal: null.

Test runner caught exception (Failed)
Tried running suite 3 time(s), still failed, giving up.
Exit code:   134
Done
```
now with `safeDel`.

aeisenberg: That's annoying. Just for fun, can you try rebasing this PR on top of https://github.com/github/vscode-codeql/pull/1611?

angelapwen: I just did — getting some errors locally (`Error: Cannot find module '/Users/angelapwen/repos/vscode-codeql/extensions/ql-vscode/out/vscode-tests/no-workspace/index'`) after rebasing when I try to debug now but pushed up to see if it's a local dev environment thing 🤷

angelapwen: Same failure, except now I can't run it locally 😆 going to revert the rebase

aeisenberg:
Now, looks like the tests are failing because `interpreted.json` is trying to be deleted, but a lock on it is being held by another process or something. It has to do with the `safeDel` method. For some reason, it is not able to be deleted and it looks like the tests can't even read it.

Not a pleasing solution, but perhaps, you can use a different name for interpreted results file for each test. This way, you won't need to delete it in `afterEach` and you can be sure you can read it in each test.

angelapwen:
> Now, looks like the tests are failing because `interpreted.json` is trying to be deleted, but a lock on it is being held by another process or something. It has to do with the `safeDel` method. For some reason, it is not able to be deleted and it looks like the tests can't even read it.
> 
> Not a pleasing solution, but perhaps, you can use a different name for interpreted results file for each test. This way, you won't need to delete it in `afterEach` and you can be sure you can read it in each test.

Dave and I just came up with something like this in pairing too. It looked like there was a separate issue with Windows Defender that needed a sleep after the file was written (because it couldn't be opened while Windows Defender was scanning it), so we added 10s sleep as well. After that we discovered the problem in the final test where it wasn't able to be deleted/read. 

I've renamed the file for just the last test to see if it passes for now (can't reproduce locally without a Windows machine). I can also rename the others for consistency.

extensions/ql-vscode/src/vscode-tests/no-workspace/query-results.test.ts
angelapwen: Figured out how move this to the `beforeEach` hook and pushed up.

angelapwen: Nevermind, that didn't work... had to revert

aeisenberg:
Not sure if you've tried this already, but you can launch the test version of vscode with `--verbose and `--log` `trace` to get some more info while you are running. Augment the CLI options in [getLaunchArgs](https://github.com/alexet/vscode-codeql/blob/main/extensions/ql-vscode/src/vscode-tests/run-integration-tests.ts#L123).

angelapwen:
Phew, renaming the path worked 😸 so this PR is finally ready for re-review

aeisenberg (approved):
Great job sticking with this and working through all these frustrating details. It seemed like a simple task, but there was a lot of subtlety with it.

PR #1462, opened 2022-08-16
Escape HTML characters when rendering MRVA results as markdown

shati-patel:
Our MRVA markdown export uses HTML `<pre>` and `<code>` blocks to render code snippets. When the code snippet itself is HTML (or XML or similar), the snippet is treated as HTML instead of raw code. To fix this, we can escape certain HTML characters (such as `<`).

See internal linked issue for some examples and a bit more context! 🖼️ 

## Checklist

N/A - this is part of the internal variant analysis feature 🦺 

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
@@ -138,7 +138,7 @@ function generateMarkdownForCodeSnippet(
   const codeLines = codeSnippet.text
     .split('\n')
     .map((line, index) =>
-      highlightCodeLines(line, index + snippetStartLine, highlightedRegion)
+      highlightAndEscapeCodeLines(line, index + snippetStartLine, highlightedRegion)
     );
 
   // Make sure there are no extra newlines before or after the <code> block:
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/test/pure-tests/remote-queries/markdown-generation/data/interpreted-results/problem/expected/meteor-meteor.md
```diff
@@ -46,3 +46,26 @@ pp.strictDirective = function(start) {
 *This part of the regular expression may cause exponential backtracking on strings starting with '"' and containing many repetitions of '\!'.*
 
 ----------------------------------------
+
+[app/src/main/AndroidManifest.xml](https://github.com/AlexRogalskiy/android-nrf-toolbox/blob/034cf3aa7d2a3a4145177de32546ca518a462a66/app/src/main/AndroidManifest.xml#L239-L249)
```
shati-patel, line 50: Note: I'm aware that this example doesn't make any sense in the context of the given repo and language (`meteor/meteor` and `javascript`) 😅 

But since it's just test data, I thought it would be easier to tweak an existing test than add a completely new example!

extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
```diff
@@ -330,3 +335,10 @@ function createFileName(nwo: string) {
   const [owner, repo] = nwo.split('/');
   return `${owner}-${repo}`;
 }
+
+/**
+ * Escape characters that could be interpreted as HTML instead of raw code.
+ */
+function escapeHtmlCharacters(text: string): string {
+  return text.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;');
```
robertbrignull, line 343: tl;dr what's here is fine 👍 

Out of curiosity I wondered about the performance of doing multiple regex `replace` calls. Turns out it is pretty good. Possibly the characteristics will change for a longer string, but I think this test case is roughly representative of real data and also I'm confident it'll be fine anyway.

I tried an online benchmarking site to test against using `replaceAll` to avoid the regex argument, or a single `replace` that matches all cases and uses a function to determine the replacement. The results are as follows:
Screenshot 2022-08-16 at 14 29 49

robertbrignull (approved):
Changes and tests LGTM

aeisenberg (commented):
Code looks fine, but just want to make sure there are no other characters that need to be escaped. Eg- multiple spaces, single and double quotes, tabs, and other whitespace.

You've probably already tested this, but just want to make sure.

shati-patel:
> Code looks fine, but just want to make sure there are no other characters that need to be escaped. Eg- multiple spaces, single and double quotes, tabs, and other whitespace.
> 
> You've probably already tested this, but just want to make sure.

Thanks for checking! 😊 Whitespace/tabs etc haven't been an issue for the markdown export (though we do escape them [in the webview](https://github.com/github/vscode-codeql/blob/6808d7dcaf09521b5d2156f3633fd807ed24648c/extensions/ql-vscode/src/remote-queries/view/FileCodeSnippet.tsx#L24))

Issue #1065, opened 2022-01-04
Startup time is slow when there are many installed databases


aeisenberg:
When there are many of installed databases in the workspace, extension startup time is slow. This issue is about determining the root cause of this slowness and then fixing it.

aeisenberg:
Adding the "Good first issue" label since determining what the problem is will be helpful even if the solution is quite complicated.

nickrolfe:
I'm not sure it's related to the *number* of databases, necessarily. I just observed the extension taking ~50 seconds to finish loading with only a single (large) database installed.

Even with zero databases installed, it takes ~30 seconds for the extension to finish loading my query history. That may or may not be a separate problem.

aeisenberg:
It's possible that the slowness is related to loading the source archive file system folders. Is it possible that your workspace with 0 databases still had some source archive folders?

aeisenberg:
We had a discussion internally about the recently. Transcribing the interesting parts of the discussion here.

- _Person A_: I noticed that CodeQL extension spent a long time loading databases on startup just not. At least 30+ seconds with my 14 databases :thinking_face: is there some known slowdown for this?

- _Person B_: I had a look before. The problem is we add a virtual filesystem for each database and vscode decides to immediately scan it. This involves unzipping the source archive and takes time and as JS is single threaded this blocks everything else. The solution would be to either delay responding until things are loaded (not great but may be better UI, still blocks all extensions at some point) or move the work to a worker.

- _Person C_: That makes sense. So, it's not really about the number of databases, but the size of the archives.
We could also delay scanning the archive until the user actually tries to open it. I wonder if we could  be even lazier and only scan individual directories on request. I'd have to take a deeper look at the implementation to see if this is possible.

aeisenberg:
@stonish I think this is something more appropriate for your team now. It may already be getting looked at with the improvements to single-rooted workspaces. I'm going to take this off of our planning board. I'm not sure how you want to track this now.

trganda:
> Please try moving the codeql instance one level down. Eg from /Users/trganda/.codeql to something like /Users/trganda/.codeql/v2.15.2.

It's work and the start-up being so fast. Tks :)

PR #1465, opened 2022-08-22
Improve startup time

alexet:
Greatly improves #1065

In both these cases we have a cache initialised by a promise. However if we call the method again before the promise finishes we start another copy. This makes it so we cache the promise instead and only do the work once. I also applied the transformation it to `getVersion` as I noticed multiple calls to codeql to get the version as well (much less important as the codeql work is parallel to npm).

## Checklist


aeisenberg (approved):
Thanks for looking into this. I tried this locally. This is definitely an improvement. But, things are still slow. 

Seems like loading the archive contents is still a bottleneck (even though we are doing it only once now). It doesn't seem like there is a way to do any caching.

PR #1466, opened 2022-08-23
Remove canary requirement for GitHub database download

koesie10:
This adds the ability for users to download CodeQL databases from GitHub without the canary flag being set. Since this feature requires GitHub authentication, the "Authenticate to GitHub" command is now also available.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -664,7 +664,7 @@
         },
         {
           "command": "codeQLDatabases.chooseDatabaseGithub",
-          "when": "config.codeQL.canary && view == codeQLDatabases",
+          "when": "view == codeQLDatabases",
           "group": "navigation"
         },
         {
@@ -879,10 +879,6 @@
       ],
       "commandPalette": [
         {
-          "command": "codeQL.authenticateToGitHub",
-          "when": "config.codeQL.canary"
-        },
-        {
           "command": "codeQL.runQuery",
           "when": "resourceLangId == ql && resourceExtname == .ql"
         },
@@ -925,10 +921,6 @@
         {
           "command": "codeQL.viewCfg",
           "when": "resourceScheme == codeql-zip-archive && config.codeQL.canary"
-        },
-        {
-          "command": "codeQL.chooseDatabaseGithub",
-          "when": "config.codeQL.canary"
         },
         {
           "command": "codeQLDatabases.setCurrentDatabase",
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/package.json
```diff
@@ -878,10 +878,6 @@
         }
       ],
       "commandPalette": [
-        {
```
charisk, line 881: How come we're removing this command? Should we just remove the `when` clause?

extensions/ql-vscode/package.json
```diff
@@ -926,10 +922,6 @@
           "command": "codeQL.viewCfg",
           "when": "resourceScheme == codeql-zip-archive && config.codeQL.canary"
         },
-        {
```
charisk, line 929: How come we're removing this command? Should we just remove the when clause?

extensions/ql-vscode/package.json
```diff
@@ -1175,7 +1167,7 @@
       },
       {
         "view": "codeQLDatabases",
-        "contents": "Add a CodeQL database:\n[From a folder](command:codeQLDatabases.chooseDatabaseFolder)\n[From an archive](command:codeQLDatabases.chooseDatabaseArchive)\n[From a URL (as a zip file)](command:codeQLDatabases.chooseDatabaseInternet)\n[From LGTM](command:codeQLDatabases.chooseDatabaseLgtm)"
+        "contents": "Add a CodeQL database:\n[From a folder](command:codeQLDatabases.chooseDatabaseFolder)\n[From an archive](command:codeQLDatabases.chooseDatabaseArchive)\n[From a URL (as a zip file)](command:codeQLDatabases.chooseDatabaseInternet)\n[From GitHub](command:codeQLDatabases.chooseDatabaseGithub)\n[From LGTM](command:codeQLDatabases.chooseDatabaseLgtm)"
```
charisk, line 1170: Nice!

charisk (commented):
Thanks for looking into this!

extensions/ql-vscode/package.json
koesie10: The default is for all commands to be visible to the user in the command palette. So, when a command is not in the command palette, it will always be visible to the user. For example, the `codeQL.chooseDatabaseLgtm` command is also not in the `commandPalette` section, only in the `commands` section.

extensions/ql-vscode/package.json
koesie10: See my comment above, when there's no condition the command doesn't need to be in the `commandPalette` section.

extensions/ql-vscode/package.json
charisk: Ah awesome! Thanks for explaining - wasn't aware of that.

charisk (approved):
Great work!

PR #1468, opened 2022-08-23
Add `implements` and `signature` to syntax highlighting

dbartol:
This PR adds syntax highlighting for the new, still-experimental "parameterized modules" feature in the QL language.

`signature` is added as an "annotation keyword", highlighted only if it appears before a module, class, or predicate declaration. `implements` is added as a context-sensitive keyword, available only within a module defintion (but outside the module body).


extensions/ql-vscode/syntaxes/ql.tmLanguage.yml
```diff
@@ -668,6 +679,19 @@ repository:
     - match: '(?#simple-id)'
       name: entity.name.type.namespace.ql
 
+  # The `extends` clause of a class definition.
```
aeisenberg, line 682:
```suggestion
  # The `implements` clause of a class definition.
```

aeisenberg (commented):
Hmmm...`signature` is working as expected, but `implements` is not. Is something wrong with this syntax?


_Extension_Development_Host__definitions_ql_—_vscode-codeql-starter__Workspace_●

extensions/ql-vscode/syntaxes/ql.tmLanguage.yml
dbartol: Doesn't `implements` only apply to modules?

dbartol: (I admit that would have been more obvious had I remembered to change the comment...)

aeisenberg (approved):
Ah...yes. That works now. Thanks.

PR #1472, opened 2022-08-24
Fix data not being loaded in MRVA results panel

koesie10:
When the MRVA results panel is closed (so the panel gets disposed) and opened again, it would not load the MRVA data (such as whether a query has already been downloaded). This fixes it by also resetting the internal state of whether the panel is loaded when the panel is disposed.

## Checklist


shati-patel (approved):
Nice, this seems to fix the bug 😻 Thanks! ⚡

PR #1477, opened 2022-08-26
Unify the Webpack bundle

koesie10:
This will move all webviews into a single Webpack bundle. This will make it easier to add new webviews since we don't need to add a new bundle, but just need to add a new directory with an `index.tsx` file.

It also moves the CSS processing to Webpack so that we don't need to specify the CSS files to use separately, but can simply do so in the TypeScript files.

The next step will be abstracting/simplifying the interface classes which create the panel.

## Checklist


extensions/ql-vscode/src/interface-utils.ts
```diff
@@ -116,11 +117,38 @@ export function tryResolveLocation(
...
-  allowInlineStyles: boolean
+  view: 'results' | 'compare' | 'remote-queries',
+  {
+    allowInlineStyles,
+    includeCodicons
+  }: {
+    allowInlineStyles?: boolean;
+    includeCodicons?: boolean;
+  } = {
+      allowInlineStyles: false,
```
aeisenberg, line 130: I believe this is only `true` for the mrva page so that it can use inline styles while loading react components from the primer UI toolkit library. Since we are moving away from primer and using the vscode UI toolkit, is this still necessary?

Again, not related to this PR, but perhaps something to check later.

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
```
aeisenberg, line 1: I think this file can be deleted.

aeisenberg (commented):
Thanks for taking this on! Looks really good. I have a few questions, but nothing really blocking. I haven't downloaded this and tried it out, which I plan on doing.

extensions/ql-vscode/src/interface-utils.ts
```diff
@@ -116,11 +117,38 @@ export function tryResolveLocation(
...
 ): string {
+  const scriptUriOnDisk = Uri.file(
+    ctx.asAbsolutePath('out/webview.js')
+  );
+
+  const stylesheetUrisOnDisk = [
+    Uri.file(ctx.asAbsolutePath('out/webview.css'))
+  ];
+
+  if (includeCodicons) {
```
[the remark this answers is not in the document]
koesie10: As you have noticed in #1478, I think we should always include the codicons and will do it in that PR. However, to include the codicons the path should also be included in the `localResourceRoots`, so that would require duplicating [this line](https://github.com/github/vscode-codeql/blob/45da1e0f1fa845c6f7b48d5a5ca52b69a4680646/extensions/ql-vscode/src/remote-queries/remote-queries-interface.ts#L120) to the other two interface managers. I didn't really think that would be worth it, especially since that part is going to be replaced anyway in #1478.

aeisenberg (approved):

PR #1482, opened 2022-08-29
Bundle Codicons using Webpack

koesie10:
This will include the Codicons inside the webview bundle, reducing the number of files that need to be loaded and the resource roots that need to be included.

The only place where a codicon is currently used is in the repository search bar for MRVA results.

## Checklist

### DIFF
--- a/extensions/ql-vscode/gulpfile.ts/webpack.config.ts
+++ b/extensions/ql-vscode/gulpfile.ts/webpack.config.ts
@@ -57,13 +57,15 @@ export const config: webpack.Configuration = {
         ]
       },
       {
-        test: /\.(woff(2)?|ttf|eot|svg)$/,
+        test: /\.(woff(2)?|ttf|eot)$/,
         use: [
           {
             loader: 'file-loader',
             options: {
               name: '[name].[ext]',
               outputPath: 'fonts/',
+              // We need this to make Webpack use the correct path for the fonts.
+              // Without this, the CSS file will use `url([object Module])`
               esModule: false
             }
           },
--- a/extensions/ql-vscode/gulpfile.ts/webpack.config.ts
+++ b/extensions/ql-vscode/gulpfile.ts/webpack.config.ts
@@ -55,6 +55,19 @@ export const config: webpack.Configuration = {
             loader: 'css-loader'
           }
         ]
+      },
+      {
+        test: /\.(woff(2)?|ttf|eot|svg)$/,
+        use: [
+          {
+            loader: 'file-loader',
+            options: {
+              name: '[name].[ext]',
+              outputPath: 'fonts/',
+              esModule: false
+            }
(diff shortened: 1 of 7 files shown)


aeisenberg (commented):
With this change, does this mean that we can remove the references to codicons when generating the webview? LIke here https://github.com/github/vscode-codeql/blob/d2af550bccba566e7de263daa71cc7afee074bd7/extensions/ql-vscode/src/abstract-interface-manager.ts#L59 and here https://github.com/github/vscode-codeql/blob/d2af550bccba566e7de263daa71cc7afee074bd7/extensions/ql-vscode/src/interface-utils.ts#L144?

koesie10:
Thanks for noticing, those changes probably got lost when I merged in `main`. Those can indeed be removed.

extensions/ql-vscode/gulpfile.ts/webpack.config.ts
```diff
@@ -55,6 +55,19 @@ export const config: webpack.Configuration = {
...
         ]
+      },
+      {
+        test: /\.(woff(2)?|ttf|eot|svg)$/,
+        use: [
+          {
+            loader: 'file-loader',
+            options: {
+              name: '[name].[ext]',
+              outputPath: 'fonts/',
```
charisk, line 66: It doesn't feel right to be putting svg files in `fonts/` to me 🤔 do we need svgs for codicons or can we leave that out for now?

extensions/ql-vscode/gulpfile.ts/webpack.config.ts
```diff
@@ -55,6 +55,19 @@ export const config: webpack.Configuration = {
...
+              esModule: false
```
charisk, line 67: Are we setting this to help with tree-shaking? 

It might be worth adding a code comment to explain - webpack config is a bit too magical.

charisk (approved):
Thanks for tidying this up! Just a couple of questions/comments.

extensions/ql-vscode/gulpfile.ts/webpack.config.ts
koesie10: We can leave that out for now, this just includes all possible font extension types. However, Codicons doesn't use SVG so we can leave it out.

extensions/ql-vscode/gulpfile.ts/webpack.config.ts
koesie10: I've added a comment. If this is not set, Webpack doesn't know this is just a file path and will use `src: url([object Module]) format("truetype");` as the CSS property for the font.

charisk: Cool, thanks for the explanation!

charisk (approved):

PR #1483, opened 2022-08-29
Bundle extension files using esbuild

koesie10:
This bundles the extension files using esbuild, as recommended by VSCode. This reduces the size of the extension from 17MB to 6MB.

Gulp will still run TypeScript to check types, but will not use the TypeScript compiler output in the bundle.

See: https://code.visualstudio.com/api/working-with-extensions/bundling-extension

## TODO

- [ ] Integration tests are broken, probably because they are all using a different version of `mocha`.
- [ ] This is quite a high-risk change since it changes how all files and modules are loaded.

## Checklist


dbartol:
If the TypeScript output is no longer being bundled, I assume that the bundled extension winds up using `ts-loader` or something like that to handle the TypeScript at runtime? How will our debug targets be affected? Will they also use `ts-loader`, or will they use precompiled TypeScript? I'd prefer that the debug targets be debugging something as close to the end-user environment as possible.

extensions/ql-vscode/package.json
```diff
@@ -1325,5 +1326,11 @@
   },
   "resolutions": {
     "glob-parent": "6.0.0"
+  },
+  "__metadata": {
+    "id": "56d1e1f8-892a-4abc-b402-c473b832e324",
+    "publisherDisplayName": "GitHub",
+    "publisherId": "7c1c19cd-78eb-4dfb-8999-99caf7679002",
+    "isPreReleaseVersion": false
```
aeisenberg, line 1334: This looks generated. Is this required?

koesie10:
I'm sorry about the confusion, the PR body is not clear about this. The TypeScript files are still being compiled to JavaScript, but it's not using the TypeScript compiler anymore. ESBuild will take in the TypeScript files, strip the types, bundle it and create a .js and .js.map file.

Previously, the TypeScript compiler would create a `.js` file for every `.ts` file: `extension.js`, `ide-server.ts`, `packaging.ts`, etc. In combination with the `node_modules` directory, this would allow VSCode to run the `extension.js` file, which imports the other files and files in `node_modules`. However, this creates thousands of files, which can be quite slow to load.

In comparison, ESBuild will take in all the files in the extension and all `node_modules` and create 1 file: `extension.js` (and `extension.js.map`). This way, we don't need to copy `node_modules` anymore and are left with just this 1 file which VSCode needs to load, which can be much faster.

In terms of debugging, there will be no real difference since ESBuild will still generate a source map, so VSCode will automatically map the compiled bundle file to the original source paths (it might be using `source-map`/`source-map-support`, I'm not sure about that part). There is no difference in how the files are loaded in development or production, both will load just a single `extension.js` file.

Overall, this reduces the number of files from thousands (I believe 7,000+) down to 1 file which needs to be loaded by VSCode (or could potentially be loaded). In the current state of this PR, we are only including the `node_modules/@vscode/codicons` since that's used by the webview. However, this can potentially also be removed if we do #1482.

Unfortunately, we can't completely remove the `node_modules` folder. While it won't be included in the final bundled VSCode extension, we still require it for tests due to the setup of the tests. We might be able to get rid of it by restructuring the tests, but I don't think that's worth the time investment.

dbartol:
>There is no difference in how the files are loaded in development or production, both will load just a single extension.js file.

Great, this is what I was hoping for.

aeisenberg:
Nice. Thanks for the explanation. This makes sense.

Looks like there's a failing build step for injecting the Application Insights key (for telemetry). The file that the script is looking for no longer exists since all files are concatenated, that step will need to change. The failure is here: https://github.com/github/vscode-codeql/blob/de6c523bad8b3383b05c722a4682f0066b3b7097/extensions/ql-vscode/gulpfile.ts/appInsights.ts#L14

I think it would be sufficient to just change that to `extension.js`. The key it is replacing is unique enough that I don't think it will find any false positives.

extensions/ql-vscode/gulpfile.ts/typescript.ts
```diff
@@ -41,35 +43,33 @@ export function compileEsbuild() {
...
+  gulp.watch('src/**/*.ts', compileTypeScriptTests);
 }
 
 export function checkTypeScript() {
   // This doesn't actually output the TypeScript files, it just
   // runs the TypeScript compiler and reports any errors.
-  return tsProject.src()
-    .pipe(tsProject(goodReporter()));
+  return tsProject.src();
+  // .pipe(tsProject(goodReporter()));
```
aeisenberg, line 72: delete?

extensions/ql-vscode/src/vscode-tests/no-workspace/contextual/queryResolver.test.ts
```diff
@@ -93,9 +94,9 @@ describe('queryResolver', () => {
     getPrimaryDbschemeSpy = sinon.stub();
     return proxyquire('../../../contextual/queryResolver', {
       'fs-extra': {
+        ensureDirSync,
```
aeisenberg, line 97: On line 9, there is:

```
const proxyquire = pq.noPreserveCache().noCallThru();
```

I don't think `noCallThru` is necessary. If you remove that, do you still need to have this line?

extensions/ql-vscode/src/vscode-tests/no-workspace/sarifParser.test.ts
```diff
@@ -1,6 +1,6 @@
 import * as path from 'path';
 import * as chai from 'chai';
-import chaiAsPromised from 'chai-as-promised';
+import * as chaiAsPromised from 'chai-as-promised';
```
aeisenberg, line 3: I'm pretty sure this line and line 7 are not actually needed since `chaiAsPromised` is configured in `extensions/ql-vscode/src/vscode-tests/no-workspace/index.ts`.

aeisenberg (commented):
I'm a little concerned about using a vastly difference structure for running tests and for what we distribute. Doing so may mean that we don't catch extension packaging issues unless we test manually. 

I mean...we _should_ be testing manually anyway before a release because we _have_ had other packaging issues. But these changes seem like it would be more likely for packaging bugs to slip through.

I think this change is very valuable since it helps make our published extension smaller and have fewer files. But I also think it is very valuable to (as much as possible) test what we ship. Can you be more explicit about what needs to change in order to run the tests from the esbuild extension?

My understanding is that the following is broken:

- proxyquire
- loading external test data

Anything else?

koesie10:
> My understanding is that the following is broken:
>
> - proxyquire
> - loading external test data
> Anything else?

In general the problem is that we cannot access anything from the extension itself when running bundled tests. That also means that when a test exists to test it in the environment of VSCode, a different function will be tested than is actually loaded in the extension. For example, if we call `sarifParser`, we won't call the `sarifParser` function in the loaded extension, but a separate `sarifParser` function that is loaded as part of the test suite.

Loading external test data is actually fine since that doesn't involve us accessing the extension. Proxyquire is indeed unusable because when running a bundled test suite we're not loading any external modules.

I think it will be quite a challenge to solve this problem using our current test suite. The only safe way to test the extension is to call commands and retrieve the output from somewhere in VSCode. It is not safe to call a function inside of the extension. Some of our existing tests can be moved to the unit test suite which should not impact the actual functionality testing (I think the `sarifParser` test is safe to be run outside of VSCode). Others will be a lot harder, such as the cli-integration tests.

aeisenberg:
Thanks for the explanation. That is indeed unfortunate. My concern is that we wind up testing something that is not being shipped and therefore miss something. I am not sure how likely this really is and I don't want to block what seems like a legitimate improvement in the extension.

Most likely, any problem that the automated tests miss would be around some sort of packaging and distribution change. If we were to commit to more manual (smoke) testing, or some level of end to end testing before releasing, I think this change would be fine.

koesie10:
Closing this in favour of #1799

PR #1485, opened 2022-08-30
Remove canary requirement for GitHub database download

koesie10:
This adds the ability for users to download CodeQL databases from GitHub without the canary flag being set.

If the canary flag is set, we will always require authentication. If it is not set, we do not require authentication, but will use an authentication token if it is available. The "Authenticate to GitHub" command is also available, so users which are not using the canary flag can still download private CodeQL databases using that command, although it's only available using the command palette.

See #1466 for the initial implementation of this.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/authentication.ts
+++ b/extensions/ql-vscode/src/authentication.ts
@@ -76,16 +76,27 @@ export class Credentials {
     }));
   }
 
-  async getOctokit(): Promise<Octokit.Octokit> {
+  /**
+   * Creates or returns an instance of Octokit.
+   *
+   * @param requireAuthentication Whether the Octokit instance needs to be authentication as user.
+   * @returns An instance of Octokit.
+   */
+  async getOctokit(requireAuthentication = true): Promise<Octokit.Octokit> {
     if (this.octokit) {
       return this.octokit;
     }
 
-    this.octokit = await this.createOctokit(true);
-    // octokit shouldn't be undefined, since we've set "createIfNone: true".
-    // The following block is mainly here to prevent a compiler error.
+    this.octokit = await this.createOctokit(requireAuthentication);
+
     if (!this.octokit) {
-      throw new Error('Did not initialize Octokit.');
+      if (requireAuthentication) {
+        throw new Error('Did not initialize Octokit.');
+      }
+
+      // We don't want to set this in this.octokit because that would prevent
+      // authenticating when requireCredentials is true.
+      return new Octokit.Octokit({ retry });
     }
     return this.octokit;
   }
--- a/extensions/ql-vscode/src/authentication.ts
+++ b/extensions/ql-vscode/src/authentication.ts
@@ -79,7 +79,7 @@ export class Credentials {
   /**
    * Creates or returns an instance of Octokit.
    *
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/authentication.ts
```diff
@@ -76,16 +76,27 @@ export class Credentials {
...
+
     if (!this.octokit) {
-      throw new Error('Did not initialize Octokit.');
+      if (requireAuthentication) {
+        throw new Error('Did not initialize Octokit.');
+      }
+
+      // We don't want to set this in this.octokit because that would prevent
+      // authenticating when requireCredentials is true.
+      return new Octokit.Octokit({ retry });
```
aeisenberg, line 99: It _might_ be safe to store this octokit instance in `this.octokit`. Above, a listener is being registered that will recreate the octokit instance when credentials change.

aeisenberg: Though, what you have now is also fine.

aeisenberg (approved):
A couple of minor suggestions, but this can go in without them.

extensions/ql-vscode/src/authentication.ts
koesie10: I think it's not safe to store it in the instance variable. That would fail in the following scenario:
1. You first download a database from GitHub (`this.octokit` is set to an unauthenticated instance)
2. You now try to run a variant analysis, which requires authentication. However, `this.octokit` is set so an unauthenticated instance, so an unauthenticated instance is returned while `requireAuthentication` is `true`.

shati-patel (approved):
Thanks for the updates! This looks good to me 😺 ✨

robertbrignull (commented):
I have a couple of questions about points in the code. I have tried it out though and it seems to be working as intended.

extensions/ql-vscode/src/authentication.ts
```diff
@@ -76,16 +76,27 @@ export class Credentials {
...
   }
 
-  async getOctokit(): Promise<Octokit.Octokit> {
+  /**
+   * Creates or returns an instance of Octokit.
+   *
+   * @param requireAuthentication Whether the Octokit instance needs to be authenticated as user.
+   * @returns An instance of Octokit.
+   */
+  async getOctokit(requireAuthentication = true): Promise<Octokit.Octokit> {
```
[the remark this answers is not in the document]
koesie10: No, not currently, but see https://github.com/github/vscode-codeql/pull/1485/commits/23a0e03cef87611ac35737ee62a1549e36f10c55 where I give a reasoning for why I left it like this

extensions/ql-vscode/src/extension.ts
```diff
@@ -1018,19 +1018,16 @@ async function activateWithInstalledDistribution(
     }
   };
 
-  // The "authenticateToGitHub" command is internal-only.
   ctx.subscriptions.push(
     commandRunner('codeQL.authenticateToGitHub', async () => {
-      if (isCanary()) {
```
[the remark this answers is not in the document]
koesie10: This command should never be called in non-canary mode, so this shouldn't really matter

PR #1490, opened 2022-08-31
Check CLI version and fix race condition in join-order scanning

dbartol:
Two fixes, both serious regressions introduced by the "bad join order detection" feature.

See individual commit messages for details.


adityasharad (approved):
Looks reasonable. Do we have a test that exercising the reading of the log summary (granted that may not have failed consistently with this bug)?

dbartol:
Any of our tests that run a query would hit the racy code path. I never hit it in my initial development, but I hit it pretty regularly while debugging the version check issue. I think it had something to do with where my breakpoints were set.

aeisenberg (approved):
This makes sense. Thanks for the fix!

PR #1492, opened 2022-09-01
Add retry for finding result-index artifact

koesie10:
It seems like the result-index artifact may not be available immediately after the workflow run has finished. This adds a retry mechanism to wait for the result-index to be available. It will retry at most 10 times with a wait of 1 second between each retry.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
+++ b/extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
@@ -43,8 +43,7 @@ export async function getRemoteQueryIndex(
   const workflowUri = `https://github.com/${owner}/${repoName}/actions/runs/${workflowRunId}`;
   const artifactsUrlPath = `/repos/${owner}/${repoName}/actions/artifacts`;
 
-  const artifactList = await listWorkflowRunArtifacts(credentials, owner, repoName, workflowRunId);
-  const resultIndexArtifactId = tryGetArtifactIDfromName('result-index', artifactList);
+  const [artifactList, resultIndexArtifactId] = await waitForArtifact(credentials, owner, repoName, workflowRunId, 'result-indx');
   if (!resultIndexArtifactId) {
     return undefined;
   }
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
```diff
@@ -255,6 +254,44 @@ function tryGetArtifactIDfromName(
...
+ * Wait for an artifact to be available in a workflow run.
+ * @param credentials Credentials for authenticating to the GitHub API.
+ * @param owner
+ * @param repo
+ * @param workflowRunId The ID of the workflow run to get the artifact for.
+ * @param artifactName The artifact name, as a string.
+ * @param maxAttempts The maximum number of attempts to download the artifact.
+ * @returns An array containing the full list of artifacts and the ID of the artifact with the given name.
+ */
+async function waitForArtifact(
```
robertbrignull, line 267: This approach seems like it'll help, but I don't know how long it's reasonable to wait for the artefacts to show up and whether 10s will be enough.

As an alternate suggestion, would it work to add a check of the `result-index` artifact to [`monitorQuery`](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/remote-queries/remote-queries-monitor.ts#L20) so that method only returns `"CompletedSuccessfully"` when the artifact is successful?

If that method finds that work the workflow claims to be completed but the required artifact doesn't exist yet, then it'll go round the loop again and sleep for a while. This means it'll be able to handle any period where the artifacts aren't populated, though we could specify a duration of maybe an hour where it'll consider the run failed.

If we add the check there we theoretically shouldn't need any extra code here. This part will work as before but we'll be sure that the artifacts exist by this point.

extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
```diff
@@ -43,8 +43,7 @@ export async function getRemoteQueryIndex(
   const workflowUri = `https://github.com/${owner}/${repoName}/actions/runs/${workflowRunId}`;
   const artifactsUrlPath = `/repos/${owner}/${repoName}/actions/artifacts`;
 
-  const artifactList = await listWorkflowRunArtifacts(credentials, owner, repoName, workflowRunId);
-  const resultIndexArtifactId = tryGetArtifactIDfromName('result-index', artifactList);
+  const [artifactList, resultIndexArtifactId] = await waitForArtifact(credentials, owner, repoName, workflowRunId, 'result-indx');
```
robertbrignull, line 46:
```suggestion
  const [artifactList, resultIndexArtifactId] = await waitForArtifact(credentials, owner, repoName, workflowRunId, 'result-index');
```
Just a typo here

extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
```diff
@@ -255,6 +254,44 @@ function tryGetArtifactIDfromName(
...
+ * @returns An array containing the full list of artifacts and the ID of the artifact with the given name.
+ */
+async function waitForArtifact(
+  credentials: Credentials,
+  owner: string,
+  repo: string,
+  workflowRunId: number,
+  artifactName: string,
+  maxAttempts = 10,
+  intervalBetweenAttempts = 1000
```
aeisenberg, line 274: Better to use exponential backoff here? Eg- interval of 1s, 2s, 4s, 8s, ....?

koesie10: I would expect Actions to make an artifact available reasonably soon, in the order of seconds rather than minutes. I don't think exponential back-off would have been necessary for this.

extensions/ql-vscode/src/remote-queries/gh-actions-api-client.ts
koesie10: Thanks, that does sound like a better idea. I've changed it to have the check in `monitorQuery` now. I didn't implement the "specify a duration of maybe an hour where it'll consider the run failed" because I couldn't think of a reliable way to detect this. We can't really store it in the `RemoteQueriesMonitor` because that will not persist between restarts of VSCode/extension. The only real alternative would be check when the workflow run completed and comparing it to the current time, but that is dependent on the local timestamp being correct.

aeisenberg (approved):

robertbrignull (approved):
LGTM too. Thanks for updating!

In general yes the actions artifacts only take seconds or less to appear, but there can sometimes be problems so it's good to plan for it.

PR #1497, opened 2022-09-06
Setup Storybook for testing UI components

koesie10:
This sets up Storybook for testing of React components. It adds stories for some of the MRVA components. It does not add stories for the main MRVA views since those are not independent of VSCode and need to be run from within VSCode.

https://user-images.githubusercontent.com/1112623/188593710-edf734b0-796a-43bb-81ce-7c1514f9f3ee.mov

## Checklist


charisk:
> It does not add stories for the main MRVA views since those are not independent of VSCode and need to be run from within VSCode

Do you think it's possible to fake out being inside VS Code? From a very simplistic point of view, the webviews are "just" web pages rendered inside VS Code. I appreciate that reality is much more nuanced, but I was hoping we'd look into potentially setting up the whatever we need within Storybook to pretend we're inside the VS Code extension. I don't think we'll have many components that are completely independent of VS Code unfortunately.

charisk:
> We might be able to simulate some of VSCode's environment, or at least mock the function calls so that they are not undefined. However, I think this should not be necessary.

I agree. We could design our components to be not directly use VS Code function calls. I was thinking more in terms of styling etc rather than interactivity e.g. so that we're able to use VS Code colour variables etc.

koesie10:
> > We might be able to simulate some of VSCode's environment, or at least mock the function calls so that they are not undefined. However, I think this should not be necessary.
> 
> I agree. We could design our components to be not directly use VS Code function calls. I was thinking more in terms of styling etc rather than interactivity e.g. so that we're able to use VS Code colour variables etc.

VSCode colour variables do completely work (see https://github.com/github/vscode-codeql/pull/1497/files#diff-9b6fe650c4162362634b90f6d2fec716787e077e6db1febcd621986c5633155e), it's just the JavaScript `window`/`vscode` objects which do not match/do not exist which are normally inserted by VSCode.

koesie10:
I've now added a mock of the VSCode API available in the webview, so we can now also render views which depend on the VSCode API:

Screenshot 2022-09-07 at 14 40 12

extensions/ql-vscode/.npmrc
```diff
@@ -0,0 +1 @@
+legacy-peer-deps=true
```
charisk, line 1: This makes me a bit uncomfortable as it could potentially introduce breaking changes (from my understanding). Is it possible to avoid it and add the dependency that is causing the trouble explicitly?

extensions/ql-vscode/.storybook/preview.js
```diff
@@ -0,0 +1,38 @@
+import { themes } from '@storybook/theming';
+import { action } from '@storybook/addon-actions';
+
+// Allow all stories/components to use Codicons
+import '@vscode/codicons/dist/codicon.css';
+
+import '../src/stories/vscode-theme.css';
```
charisk, line 7: Instead of having our own file that we have to manage/keep up-to date etc., is there a way to get everything from VSCode automatically?

extensions/ql-vscode/.storybook/preview.js
```diff
@@ -0,0 +1,38 @@
+import { themes } from '@storybook/theming';
+import { action } from '@storybook/addon-actions';
+
+// Allow all stories/components to use Codicons
+import '@vscode/codicons/dist/codicon.css';
```
charisk, line 5: Can you remind me why this isn't just part of our webpack bundle?

extensions/ql-vscode/.storybook/preview.js
```diff
@@ -0,0 +1,38 @@
...
+
+import '../src/stories/vscode-theme.css';
+
+export const parameters = {
```
charisk, line 9: Were these just generated? Would be good to add some context about what they are, maybe a link to the docs?

extensions/ql-vscode/.storybook/preview.js
```diff
@@ -0,0 +1,38 @@
...
+    values: [
+      {
+        name: 'dark',
+        value: '#1e1e1e',
+      },
+    ],
+  },
+  options: {
+    storySort: {
+      order: ['WebView UI Toolkit', 'MRVA'],
```
charisk, line 31: How come we have stories for the toolkit? It's not something we're building - it's an external dependency.

charisk, line 31: Can we just have one set of stories (storybook?) and call it something generic? The component library we've started building with remote queries/variant analysis is meant to be used for local queries too eventually so it would be cool to avoid calling anything MRVA etc. Appreciate the current repo structure has nothing to imply that though.

extensions/ql-vscode/src/stories/vscode-webview-ui-toolkit/Overview.stories.mdx
```diff
@@ -0,0 +1,9 @@
+<!-- Checkbox.stories.mdx -->
+
```
charisk, line 2: Would be cool to have an Overview page like this one for the "MRVA" stories too.

charisk (commented):
Thank you for looking into this and getting all the bits working! It looks great and it will really improve our DX while building our new components and views!

extensions/ql-vscode/.npmrc
koesie10: Unfortunately, this is the recommended fix by Storybook (see https://github.com/storybookjs/storybook/issues/18298). I'm not sure we can avoid adding the dependency since it's one of the Storybook packages which causes this to be necessary.

extensions/ql-vscode/.storybook/preview.js
koesie10: This is part of our normal Webpack bundle, but it's only imported by the `webview.ts` file, which we're not explicitly importing in Storybook. Storybook uses a different Webpack config, so we need to explicitly import it.

extensions/ql-vscode/.storybook/preview.js
koesie10: This was how I started testing Storybook, but I agree with you that it doesn't really make sense to add stories for the WebView UI Toolkit.

koesie10: 👍, I'll remove the MRVA prefix from the existing stories such that all components are just listed as top-level stories.

extensions/ql-vscode/.storybook/preview.js
koesie10: Most of these are from the template, do you think it would be enough to link to [this documentation page](https://storybook.js.org/docs/react/configure/overview#configure-story-rendering)?

extensions/ql-vscode/.npmrc
charisk: Thanks for the link! It mentions that this will be fixed with storybook v7. Is it worth using an alpha version of that rather than 6? The reason I'm asking is because the current change potentially affects the actual extension ("prod") but using an alpha version of the storybook only potentially makes things a bit unstable for dev. 

I'm saying that without any idea about storybook versions and how mature the alpha is though!

extensions/ql-vscode/.storybook/preview.js
charisk: Ok thanks. I think there's a way to tell webpack to always include it without having to import but that's a problem for another day :D

extensions/ql-vscode/.storybook/preview.js
koesie10: We might be able to generate it from the VSCode source code, but I don't think we can import it from VSCode itself. 

The colors are defined [here](https://github.com/microsoft/vscode/blob/main/src/vs/platform/theme/common/colorRegistry.ts) and the styles are created [here](https://github.com/microsoft/vscode/blob/a6272c941610ec1da55f48ea425cdb4bf4868543/src/vs/workbench/contrib/webview/browser/themeing.ts#L51-L82). They are then injected into the WebView using [this](https://github.com/microsoft/vscode/blob/a6272c941610ec1da55f48ea425cdb4bf4868543/src/vs/workbench/contrib/webview/browser/webviewElement.ts#L740-L750) on the host side and [this](https://github.com/microsoft/vscode/blob/a6272c941610ec1da55f48ea425cdb4bf4868543/src/vs/workbench/contrib/webview/browser/pre/index.html#L497-L515) on the WebView side.

I think it will be a lot more work to generate the CSS variables from this than to copy the variables from VSCode once in a while.

extensions/ql-vscode/.storybook/preview.js
koesie10: There's a way to specify the entrypoint for Webpack so that it always imports this file, but that won't really solve our problem since we're using two different Webpack configs (one for the extension and one for Storybook). So, we would still need to import the Codicons twice.

extensions/ql-vscode/.storybook/preview.js
charisk: Ah that's a shame. I suspected it would be a can of warms, but I thought I'd ask anyway 😂 

In that case I think we should have some docs to tell people about this (maybe in the storybook part in the README?) and how to update them.

extensions/ql-vscode/.npmrc
koesie10: The PR that is referenced in the issue has also been backported to Storybook 6, but it does not actually fix the problem. The workaround which was implemented is to automatically set this parameter in the `.npmrc` file (which is how it got added here). It doesn't seem like they have a real fix yet.

extensions/ql-vscode/.storybook/preview.js
koesie10: Perhaps I can also add it to an Overview page in Storybook itself and refer to that from the CONTRIBUTING documentation?

charisk: SGTM!

extensions/ql-vscode/.npmrc
charisk: Ah 💩 okay. Thanks for the links and explanations!

Shall we add a comment to this file to explain why we're setting this flag?

extensions/ql-vscode/.storybook/preview.js
charisk: I just had a brainwave! Feel free to ignore this if you want (totally up to you), but the actual toolkit has a storybook, so perhaps this is something they've solved already? See https://github.com/microsoft/vscode-webview-ui-toolkit/tree/main/.storybook

charisk (approved):
I'm approving and leaving the last 2 unresolved comments up to you to look at or ignore!

Again, thanks for sorting this!

extensions/ql-vscode/.storybook/preview.js
koesie10: I did see whether it was possible to replicate the way they are using the VSCode variables, but they are using a completely custom system. They have a system of "design tokens" (https://github.com/microsoft/vscode-webview-ui-toolkit/blob/6fc339ce26e7da15b9ba97c063f56d8a4eae85ad/src/design-tokens.ts) which specify a variable and a default. These are then used in [the styles](https://github.com/microsoft/vscode-webview-ui-toolkit/blob/6fc339ce26e7da15b9ba97c063f56d8a4eae85ad/src/button/button.styles.ts) and will automatically use either the variable or the default. I don't think that is going to work for us since we'd need to modify the way we write the CSS files and set defaults for all variables we use. It could also result in having different defaults than the toolkit, so it might also get out of sync at some point.

koesie10:
The failing *Dependency Review* is due to dependencies of Storybook and is only used in dev environments. It would be quite hard to fix this since the dependencies are transitive dependencies of Storybook. Since they won't be used in production code, I think it should be fine to merge this.

charisk:
> The failing _Dependency Review_ is due to dependencies of Storybook and is only used in dev environments. It would be quite hard to fix this since the dependencies are transitive dependencies of Storybook. Since they won't be used in production code, I think it should be fine to merge this.

Does this mean that all future PRs will have a failing check?

koesie10:
> > The failing _Dependency Review_ is due to dependencies of Storybook and is only used in dev environments. It would be quite hard to fix this since the dependencies are transitive dependencies of Storybook. Since they won't be used in production code, I think it should be fine to merge this.
> 
> Does this mean that all future PRs will have a failing check?

No, the check only checks new dependencies, so future PRs will not fail this check. We already have some open Dependabot alerts for other development dependencies and those do not fail any PRs either.

PR #1498, opened 2022-09-06
QueryServer: Abstract over the query running parts.

alexet:
This creates an abstraction `QueryRunner` over the actual running of queries and splits about half of the query running code out (the parts about handling results and 

This should have no functional changes.

PR #1499, opened 2022-09-06
QueryServer: Add support for new query-server

alexet:
Add support for the new query server.

For local testing purposes it hardcodes cli support to true. We need to change to the correct version before merging.


extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -390,9 +388,9 @@ export class DatabaseUI extends DisposableObject {
...
     progress: ProgressCallback,
     token: CancellationToken
-  ): Promise<DatabaseItem | undefined> => {
+  ): Promise<void> => {
     try {
-      return await this.chooseAndSetDatabase(true, progress, token);
+      await this.chooseAndSetDatabase(true, progress, token);
     } catch (e) {
       void showAndLogErrorMessage(getErrorMessage(e));
       return undefined;
```
aeisenberg, line 396: This line is not necessary.

extensions/ql-vscode/src/extension.ts
```diff
@@ -70,12 +70,12 @@ import { asError, assertNever, getErrorMessage } from './pure/helpers-pure';
 import { spawnIdeServer } from './ide-server';
 import { InterfaceManager } from './interface';
 import { WebviewReveal } from './interface-utils';
-import { ideServerLogger, logger, queryServerLogger } from './logging';
+import { ideServerLogger, logger, ProgressReporter, queryServerLogger } from './logging';
 import { QueryHistoryManager } from './query-history';
 import { CompletedLocalQueryInfo, LocalQueryInfo } from './query-results';
-import * as qsClient from './queryserver-client';
+import * as qsClient from './legacy-query-server/queryserver-client';
+import * as qsClient2 from './query-server/queryserver-client';
```
aeisenberg, line 77: Maybe this is more understandable?

```suggestion
import * as legacyQsClient from './legacy-query-server/queryserver-client';
import * as newQsClient from './query-server/queryserver-client';
```

extensions/ql-vscode/src/extension.ts
```diff
@@ -782,12 +770,13 @@ async function activateWithInstalledDistribution(
           });
         }
 
-        if (queryUris.length > 1) {
+        if (queryUris.length > 1 && !await cliServer.cliConstraints.supportsNonDestructiveUpgrades()) {
```
aeisenberg, line 773: Interesting...we never had this before. Probably a bug that we always tried to upgrade the database. Most likely in these cases, this was a no-op that just took some time.

extensions/ql-vscode/src/legacy-query-server/upgrades.ts
```diff
@@ -1,13 +1,12 @@
...
-import { getOnDiskWorkspaceFolders, showAndLogErrorMessage, tmpDir } from './helpers';
-import { ProgressCallback, UserCancellationException } from './commandRunner';
-import { logger } from './logging';
-import * as messages from './pure/messages';
-import * as qsClient from './queryserver-client';
+import { getOnDiskWorkspaceFolders, showAndLogErrorMessage, tmpDir } from '../helpers';
+import { ProgressCallback, UserCancellationException } from '../commandRunner';
+import { logger } from '../logging';
+import * as messages from '../pure/legacy-messages';
+import * as qsClient from '../legacy-query-server/queryserver-client';
```
aeisenberg, line 6: Unnecessary change?

extensions/ql-vscode/src/pure/new-messages.ts
```diff
@@ -0,0 +1,208 @@
...
+}
+
+
+
+export interface UpgradeParams {
+  db: string,
+  additionalPacks: string[],
+}
+
+export type UpgradeResult = Record<string, unknown>;
```
aeisenberg, line 150: Surprised we still need upgrades here.

extensions/ql-vscode/src/query-server/query-runner.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+  async registerDatabase(progress: ProgressCallback, token: CancellationToken, dbItem: DatabaseItem): Promise<void> {
+    if (dbItem.contents && (await this.qs.cliServer.cliConstraints.supportsDatabaseRegistration())) {
+      const databases: string[] = [dbItem.databaseUri.fsPath];
+      await this.qs.sendRequest(registerDatabases, { databases }, token, progress);
+    }
+  }
+
+  async upgradeDatabaseExplicit(dbItem: DatabaseItem, progress: ProgressCallback, token: CancellationToken): Promise<void> {
```
aeisenberg, line 59: Ah... I see. Is this only for very old databases? Maybe include a note saying something like:

> Databases created by CLI versions later than 2.x.y do not need to be explicitly upgraded.

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -0,0 +1,596 @@
...
+   */
+  async ensureDilPath(cliServer: CodeQLCliServer): Promise<string> {
+    if (await this.hasDil()) {
+      return this.dilPath;
+    }
+    const compiledQuery = path.join(this.querySaveDir, 'compiledQuery.qlo');
+    if (!(await fs.pathExists(compiledQuery))) {
+      if (await cliServer.cliConstraints.supportsNewQueryServer()) {
+        // This could be from the new query server
+        // in whcih case we expect the qlo to be missing so we should ignore it
```
aeisenberg, line 199:
```suggestion
        // in which case we expect the qlo to be missing so we should ignore it
```

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -0,0 +1,596 @@
...
+        throw new Error(
```
aeisenberg, line 200: When would this happen? Is there anything the user can do to fix this situation? The error message as it is is not very helpful.

aeisenberg (commented):
Looks generally reasonable. I haven't tried this out myself.

aeisenberg:
I think you want to base this PR on top of https://github.com/github/vscode-codeql/pull/1498, which might be difficult since that PR is from your local fork.

extensions/ql-vscode/src/query-server/query-runner.ts
alexet: This is because there are a couple of minor reasons why we may want an explicit upgrade. Old databases don't actually work but there are still some small reasons why you may want a destructive upgrade (usually performance testing).

extensions/ql-vscode/src/run-queries-shared.ts
alexet: Originally I though that this should not happen unless the user deleted the dil or a bug happened in the query server. But I have also realised that a compilation error would also cause this. It is a bit tricky to work out which case but hopefully it is understandable that if your query doesn't compile then you won't be able to see the DIL.

aeisenberg:
Superseded by https://github.com/github/vscode-codeql/pull/1508

aeisenberg:
Closing in favour of #1508.

PR #1500, opened 2022-09-07
Avoid prompting for upgrades when they wouldn't be prompted.

alexet:
We prompt for destructive upgrades once at the start of running multiple queries so the user doesn't get repeated prompts but with non-destructive they wouldn't any way and it would be worse. This also moves the version check to live with all the other version checks.

I pulled this out of #1498 because it isn't directly related.

## Checklist


aeisenberg (approved):
Thanks! A nice little cleanup.

PR #1501, opened 2022-09-07
Don't display destructive upgrades any more.

alexet:
Most users don't want to destructively upgrade their database any more. This hides the option from the right click menu. It is still visible in the command picker for those users who actually want to do that.

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #1507, opened 2022-09-13
Don't display destructive upgrades any more

aeisenberg:
This is actually coming from @alexet's PR #1498. This is a separate bit and I wanted to get this in first.


shati-patel (commented):
I believe these changes (+ typo fixes in changelog) were already merged to `main` in https://github.com/github/vscode-codeql/pull/1501. Does that look right? 🤔

aeisenberg:
Yes. You're right! Thanks.

PR #1511, opened 2022-09-14
Consolidate view loading messages into one

charisk:
We currently have different message types for when a view is loaded. However, they're all handled in the same way. I've consolidated them all into one.

### DIFF
--- a/extensions/ql-vscode/src/compare/compare-view.ts
+++ b/extensions/ql-vscode/src/compare/compare-view.ts
@@ -111,7 +111,7 @@ export class CompareView extends AbstractWebview<ToCompareViewMessage, FromCompa
 
   protected async onMessage(msg: FromCompareViewMessage): Promise<void> {
     switch (msg.t) {
-      case 'compareViewLoaded':
+      case 'viewLoaded':
         this.onWebViewLoaded();
         break;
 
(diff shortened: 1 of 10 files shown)


koesie10 (approved):
Thanks, this is definitely an improvement! Would it make sense to handle this message in the `AbstractWebview` so the subclasses don't have to call `onWebViewLoaded` manually? I am thinking we can create a new method in the abstract class that will be called from here:

https://github.com/github/vscode-codeql/blob/3f3b934736d31492451a5837c64b14b75895adf4/extensions/ql-vscode/src/abstract-webview.ts#L84

Then, if the message is `viewLoaded` it will call `onWebViewLoaded`, otherwise it will call `this.onMessage`.

charisk:
> Thanks, this is definitely an improvement! Would it make sense to handle this message in the `AbstractWebview` so the subclasses don't have to call `onWebViewLoaded` manually? I am thinking we can create a new method in the abstract class that will be called from here:
> 
> https://github.com/github/vscode-codeql/blob/3f3b934736d31492451a5837c64b14b75895adf4/extensions/ql-vscode/src/abstract-webview.ts#L84
> 
> Then, if the message is `viewLoaded` it will call `onWebViewLoaded`, otherwise it will call `this.onMessage`.

I considered that (and I think it's a good idea), but I didn't go ahead with it because I think we should do that if we have base types for our messages. We currently have message types explicitly called in the corresponding type e.g. `FromResultsViewMsg`. If we had a base type (or one that these ones union with) I think it would make sense to make the change you're describing too.

PR #1512, opened 2022-09-14
Add variant analysis top header

koesie10:
This adds the variant analysis top header. It also adds React testing library to make automated testing of this possible and adds stories for the newly added component.

It might be easier to review this commit-by-commit.

Since this is only part of header, we need a better name for this component. However, I couldn't really think of one. The best one I could come up with is `VariantAnalysisTitleRow` but that doesn't really cover everything.

## Checklist


extensions/ql-vscode/src/view/variant-analysis/VariantAnalysis.tsx
```diff
@@ -1,5 +1,21 @@
...
 
 export function VariantAnalysis(): JSX.Element {
-  return <span>Hello!</span>;
+  return (
+    <VariantAnalysisContainer>
+      <VariantAnalysisHeader
+        queryName="Example query"
+        queryFileName="example.ql"
+        status={VariantAnalysisStatus.InProgress}
+        onOpenQueryClick={() => void 0}
```
charisk, line 13: Shall we put some `console.log` for now?

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
```diff
@@ -0,0 +1,91 @@
+import * as React from 'react';
+import { VariantAnalysisStatus } from '../../remote-queries/shared/variant-analysis';
+import ViewTitle from '../remote-queries/ViewTitle';
+import { CodeSquareIcon, FileCodeIcon } from '@primer/octicons-react';
```
charisk, line 4: We should try and get our icons from codicons rather than Primer.

extensions/ql-vscode/package.json
```diff
@@ -1187,7 +1187,9 @@
     "watch": "npm-run-all -p watch:*",
     "watch:extension": "tsc --watch",
     "watch:webpack": "gulp watchView",
-    "test": "mocha --exit -r ts-node/register test/pure-tests/**/*.ts",
+    "test": "npm-run-all -p test:*",
+    "test:mocha": "mocha --exit -r ts-node/register test/pure-tests/**/*.ts",
```
charisk, line 1191: Rather than splitting based on the technology used, shall we split on the kinds of tests? the jest tests are currently React component tests (e.g. `test:ui` or `test:view`), and mocha tests are other tests (e.g. `unit`?)

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { LinkIconButton } from './LinkIconButton';
+import styled from 'styled-components';
+import { VSCodeButton } from '@vscode/webview-ui-toolkit/react';
+
+export type VariantAnalysisHeaderProps = {
+  queryName: string;
+  queryFileName: string;
+  status: VariantAnalysisStatus;
+
+  onOpenQueryClick: () => void;
```
charisk, line 14: These sound quite similar. How about

```ts
onOpenQueryFileClick: () => void;
onViewQueryTextClick: () => void;
```

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
```diff
@@ -0,0 +1,91 @@
...
+  margin-left: auto;
+  display: flex;
+  gap: 1em;
+`;
+
+const Button = styled(VSCodeButton)`
+  white-space: nowrap;
+`;
+
+export const VariantAnalysisHeader = ({
```
charisk, line 47: So are you imaging this component to have the whole header area (as described in Figma) or just what we have now? I think the size of what you have now is big enough and ideally shouldn't get too much bigger.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
```diff
@@ -0,0 +1,91 @@
...
+import ViewTitle from '../remote-queries/ViewTitle';
+import { CodeSquareIcon, FileCodeIcon } from '@primer/octicons-react';
+import { LinkIconButton } from './LinkIconButton';
+import styled from 'styled-components';
+import { VSCodeButton } from '@vscode/webview-ui-toolkit/react';
+
+export type VariantAnalysisHeaderProps = {
+  queryName: string;
+  queryFileName: string;
+  status: VariantAnalysisStatus;
```
charisk, line 12: I think i'd be more explicit here and call it `variantAnalysisStatus`

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
```diff
@@ -0,0 +1,91 @@
...
+const Container = styled.div`
+  display: flex;
+  align-items: center;
+`;
+
+const TopRow = styled.div`
+  max-width: 100%;
+`;
+
+const TopLinkButtons = styled.div`
```
charisk, line 32: `QueryActions` might be a better name for this.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
```diff
@@ -0,0 +1,91 @@
...
+  display: flex;
+  gap: 1em;
+`;
+
+const BottomButtons = styled.div`
```
charisk, line 37: `VariantAnalysisActions` might be a better name for this.

charisk (commented):
This is looking great! A partial review from me atm.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
```diff
@@ -0,0 +1,91 @@
...
+            {queryFileName}
+          </LinkIconButton>
+          <LinkIconButton onClick={onViewQueryClick}>
+            <CodeSquareIcon size={16} />
+            View query
+          </LinkIconButton>
+        </TopLinkButtons>
+      </TopRow>
+      <BottomButtons>
+        {status === VariantAnalysisStatus.InProgress && (
```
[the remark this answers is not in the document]
koesie10: 👍, although I did move it to its own file. I think having each component in a separate file makes it a lot easier to navigate the code. I only put the styled components in the same file as where they're used since those don't contain any logic.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
charisk: Perhaps the fact that we can't come up with a good name is an indication that the component is doing a bit too much. I think that's also shown by the fact that we need to pass quite a few inputs.

It's currently doing two things:
1. Allows users to see query information (title and links to query file/text).
2. Allows users to execute certain actions against a variant analysis.

Perhaps we should split into two components - `QueryDetails` and `VariantAnalysisActions` or something like that?

extensions/ql-vscode/src/view/variant-analysis/__tests__/VariantAnalysisHeader.spec.tsx
```diff
@@ -0,0 +1,83 @@
+import * as React from 'react';
```
charisk, line 1: Is the `__` in dir names a convention when using jest? e.g. `__tests__`, `__mocks__` etc.

extensions/ql-vscode/src/view/remote-queries/ViewTitle.tsx
```diff
@@ -1,9 +1,13 @@
 import styled from 'styled-components';
```
charisk, line 1: Is it worth moving this file into a shared location (`view/common` rather than having some stuff in remote-queries and some in `variant-analysis`)? I'm happy for this to be in a subsequent PR if we agree it's the right thing to do.

extensions/ql-vscode/src/view/variant-analysis/__tests__/VariantAnalysisHeader.spec.tsx
koesie10: Sort of; `__tests__` is [the default value](https://jestjs.io/docs/configuration#testmatch-arraystring) for `testMatch`, but I can't really find a specific page documenting this convention. The `__mocks__` naming is also from the Jest documentation, specifically about [setting up Jest with Webpack](https://jestjs.io/docs/webpack#handling-static-assets).

I think it makes sense to follow these defaults and I personally prefer the tests to be close to the components. This convention also makes it easy to distinguish between Mocha and Jest tests.

extensions/ql-vscode/src/view/remote-queries/ViewTitle.tsx
koesie10: 👍, I think that would be very useful. We can also consider making these importable by just specifying the `../common` when we place an `index.ts` file inside the `common` directory.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
koesie10: 👍, I think that makes sense. The actions were already split into a separate component, so now I've also split out the query header into a separate component. I've kept the `VariantAnalysisHeader` component, but it now just contains a container and the two components. We can also hopefully add the upcoming stats component to this without the file becoming too large.

I haven't created stories for the two separate components since that might result in stories which are too granular. I'm also not sure whether we want to create separate unit tests for these two components or whether we want to keep the current unit tests on the parent component. I have now added some tests for the actions since those actually contain logic, but have not added any for the query details.

charisk: If you were starting fresh, would you create unit tests and stories against the individual components or the top level one? I think we'd do the smaller components. 

We potentially don't even need the `VariantAnalysisHeader` component - at the moment it feels like it just adds to prop drilling. But I don't have a strong sense about that yet.

extensions/ql-vscode/src/view/remote-queries/ViewTitle.tsx
charisk: Great! Are you happy to do that as part of the issue you're working on, as a separate PR?

koesie10: It might be better to do that in a separate PR. There are some other components which we might also be able to move to the `common` directory such as `VerticalSpace`, so the changeset might get quite large (especially since all imports also need to be changed).

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
koesie10: Good point, thanks! I've added stories and tests for the smaller components and removed the tests for the `VariantAnalysisHeader` component. I still see value in having the `VariantAnalysisHeader` and the stories around for now while we're still working on these components to see how everything fits together.

I definitely think we should refactor the `VariantAnalysisHeader` component (probably just remove it or change its props to take a complete variant analysis interface once we have that). However, I think that will be easier once we actually have an interface from which we can get the data, so I've left it like this for now.

extensions/ql-vscode/src/view/remote-queries/ViewTitle.tsx
charisk: Yeah I'd definitely prefer to see it in a separate PR!

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
charisk: Awesome thanks!

charisk (approved):
Really great work! Thanks @koesie10!

PR #1521, opened 2022-09-20
Define variant analysis domain models

charisk:
Define some initial models that represent the variant analysis domain. These will be used in the various extension flows and new view/components.

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
+++ b/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
@@ -1,24 +1,5 @@
 import { Repository } from './repository';
 
-export interface VariantAnalysisSubmission {
-  startTime: number,
-  controllerRepoId: number,
-  actionRepoRef: string,
-  query: {
-    name: string,
-    filePath: string,
-    language: VariantAnalysisQueryLanguage,
-
-    // Base64 encoded query pack.
-    pack: string,
-  },
-  databases: {
-    repositories?: string[],
-    repositoryLists?: string[],
-    repositoryOwners?: string[],
-  }
-}
-
 export interface VariantAnalysis {
   id: number,
   controllerRepoId: number,
@@ -42,8 +23,9 @@ export interface VariantAnalysis {
 export enum VariantAnalysisQueryLanguage {
   CSharp = 'csharp',
   Cpp = 'cpp',
-  Javascript = 'javascript',
   Go = 'go',
+  Java = 'java',
+  Javascript = 'javascript',
   Python = 'python',
   Ruby = 'ruby'
 }
@@ -79,14 +61,14 @@ export interface VariantAnalysisScannedRepository {
 export interface VariantAnalysisSkippedRepositories {
   accessMismatchRepos?: VariantAnalysisSkippedRepositoryGroup,
   notFoundRepos?: VariantAnalysisSkippedRepositoryGroup,
-  noCodeql_dbRepos?: VariantAnalysisSkippedRepositoryGroup,
+  noCodeqlDbRepos?: VariantAnalysisSkippedRepositoryGroup,
   overLimitRepos?: VariantAnalysisSkippedRepositoryGroup
 }
 
 export interface VariantAnalysisSkippedRepositoryGroup {
   repoCount: number,
   repositories: Array<{
-    id: number,
+    id?: number,
     fullName: string
   }>
 }
--- a/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
+++ b/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
@@ -66,7 +66,7 @@ export interface VariantAnalysisSkippedRepositories {
 }
 
 export interface VariantAnalysisSkippedRepositoryGroup {
-  repoCount: number,
+  repositoryCount: number,
   repositories: Array<{
     id?: number,
     fullName: string
--- a/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
+++ b/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
@@ -1,5 +1,92 @@
+import { Repository } from './repository';
+
+export interface VariantAnalysisSubmission {
+  startTime: number,
+  controllerRepoId: number,
+  actionRepoRef: string,
+  query: {
+    name: string,
+    filePath: string,
+    language: VariantAnalysisQueryLanguage,
+
+    // Base64 encoded query pack.
+    pack: string,
+  },
+  databases: {
+    repositories?: string[],
+    repositoryLists?: string[],
+    repositoryOwners?: string[],
+  }
+}
+
+export interface VariantAnalysis {
+  id: number,
+  controllerRepoId: number,
+  query: {
+    name: string,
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -1,5 +1,92 @@
...
+  startTime: number,
+  controllerRepoId: number,
+  actionRepoRef: string,
+  query: {
+    name: string,
+    filePath: string,
+    language: VariantAnalysisQueryLanguage,
+
+    // Base64 encoded query pack.
+    pack: string,
```
koesie10, line 13: Do we need this to be stored like this? Would it perhaps make more sense to store it as a regular file?

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -1,5 +1,92 @@
+import { Repository } from './repository';
+
+export interface VariantAnalysisSubmission {
```
koesie10, line 3: What will be the difference between the `VariantAnalysisSubmission` and the `VariantAnalysis`? It seems like they are very similar and might not need separate models?

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -1,5 +1,92 @@
...
+  accessMismatchRepos?: VariantAnalysisSkippedRepositoryGroup,
+  notFoundRepos?: VariantAnalysisSkippedRepositoryGroup,
+  noCodeql_dbRepos?: VariantAnalysisSkippedRepositoryGroup,
+  overLimitRepos?: VariantAnalysisSkippedRepositoryGroup
+}
+
+export interface VariantAnalysisSkippedRepositoryGroup {
+  repoCount: number,
+  repositories: Array<{
+    id: number,
```
koesie10, line 89: We won't have an ID for repos which are not found, so this might need to be optional (or there needs to be a separate type for the `notFoundRepos`)

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -1,5 +1,92 @@
...
+  analysisStatus: VariantAnalysisRepoStatus,
+  resultCount?: number,
+  artifactSizeInBytes?: number,
+  failureMessage?: string
+}
+
+export interface VariantAnalysisSkippedRepositories {
+  accessMismatchRepos?: VariantAnalysisSkippedRepositoryGroup,
+  notFoundRepos?: VariantAnalysisSkippedRepositoryGroup,
+  noCodeql_dbRepos?: VariantAnalysisSkippedRepositoryGroup,
```
koesie10, line 82:
```suggestion
  noCodeqlDbRepos?: VariantAnalysisSkippedRepositoryGroup,
```

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -1,5 +1,92 @@
...
+  Canceled = 'canceled',
+  TimedOut = 'timedOut',
+}
+
+export interface VariantAnalysisScannedRepository {
+  repository: Repository,
+  analysisStatus: VariantAnalysisRepoStatus,
+  resultCount?: number,
+  artifactSizeInBytes?: number,
+  failureMessage?: string
```
koesie10, line 76: Do we need to store the `database_commit_sha` and `source_location_prefix` as well, or will those be stored separately as part of the query result?

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -1,5 +1,92 @@
...
+
+export interface VariantAnalysisSkippedRepositories {
+  accessMismatchRepos?: VariantAnalysisSkippedRepositoryGroup,
+  notFoundRepos?: VariantAnalysisSkippedRepositoryGroup,
+  noCodeql_dbRepos?: VariantAnalysisSkippedRepositoryGroup,
+  overLimitRepos?: VariantAnalysisSkippedRepositoryGroup
+}
+
+export interface VariantAnalysisSkippedRepositoryGroup {
+  repoCount: number,
```
koesie10, line 87: I think it might be clearer for this to be called `repositoryCount` to match with the name in the API.

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
robertbrignull: I'm also unclear about the difference between these two. I think the rest of the types introduced in this PR are clear enough, but these two could use comments explaining their purpose, or the names need to be much more descriptive.

About their actual purpose, is one the data we send to the API to submit the variant analysis and the other one is the response? They don't match up exactly since some fields like `databases` aren't in the response, so it's not an exact copy of an API response but rather we've added extra data into it (which is fine and good). So in that case I that reduces the benefit from having two separate types.

I suppose one benefit of having separate types is you can make more fields non-nullable, so that way just all the rest of the code can be simpler with fewer type checks.

robertbrignull (commented):
In general looks good to me, though I agree with all of @koesie10's review comments and it would be good to discuss those.

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
charisk: `VariantAnalysisSubmission` is a model that will be used within the extension "core" and I don't expect it to be used in the web views at all. The idea is that it contains all the information needed to request a variant analysis. It might be worth removing `VariantAnalysisSubmission` for now and introducing it along with the relevant functionality though. I think I'll do that!

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
charisk: I think the short answer is that I don't know yet so I thought we'd leave it until we need them.

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -1,5 +1,92 @@
...
+
+export enum VariantAnalysisQueryLanguage {
+  CSharp = 'csharp',
+  Cpp = 'cpp',
+  Javascript = 'javascript',
+  Go = 'go',
+  Python = 'python',
+  Ruby = 'ruby'
+}
+
```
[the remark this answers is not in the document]
charisk: Oops!

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
charisk: This is gone now :)

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
charisk: I've gone with making it optional for now - I think the view will only use the nwo anyway.

robertbrignull (approved):
LGTM

Thanks for addressing the comments ✨

PR #1524, opened 2022-09-20
Move GitHub Actions API client to gh-api directory

charisk:
Moved the `gh-actions-api-client` into a `gh-api` subdir. We will soon have a few more files around calling the GitHub API so I think it's worth starting a subdir for them.

I've also noticed that the `gh-actions-api-client` also has logic around non-action stuff, but I'll leave that as a separate change.

## Checklist
N/A - refactoring.

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -40,7 +40,7 @@ import * as fs from 'fs-extra';
 import { CliVersionConstraint } from './cli';
 import { HistoryItemLabelProvider } from './history-item-label-provider';
 import { Credentials } from './authentication';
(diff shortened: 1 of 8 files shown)


koesie10 (approved):
LGTM. Perhaps we can add an `index.ts` file in the future to shorten the imports paths and make the code independent of where exactly a function lives within the `gh-api` directory.

charisk:
> LGTM. Perhaps we can add an `index.ts` file in the future to shorten the imports paths and make the code independent of where exactly a function lives within the `gh-api` directory.

I'm thinking it'd be good to do that in general in the codebase but I haven't poked anyone around this yet. I also don't _love_ the index.ts solution because you end up with a lot of index.ts files, but I don't think there's a better solution (at least there wasn't last time I looked!)

PR #1526, opened 2022-09-20
Release PR for v1.7.0

robertbrignull:
PR for v1.7.0 release.

Bumping to 1.7.0 because of two potentially breaking changes, though each is mitigated:
- Remove ability to download databases from LGTM.
  - This is still available behind the canary flag, to be removed at a later date.
- Removed the ability to manually upgrade databases from the context menu on databases.
  - This is still available in the command palate.


henrymercer (approved):
A minor version bump seems consistent with how we've handled other similar changes we've made in the past.

PR #1527, opened 2022-09-20
Add new GitHub API client with functions for new MRVA flow

charisk:
Add a new `gh-api-client` module that implements functions to help with the new flow. 

There is an existing gh-actions-api-client` module that contains functions that could be moved into the new module, but I've left that for a future PR.

No error handling is added at this point - the callers of this module will be responsible for catching and handling exceptions.

## Checklist
N/A - code not used yet:


extensions/ql-vscode/src/remote-queries/gh-api/gh-api-client.ts
```diff
@@ -0,0 +1,86 @@
...
+  credentials: Credentials,
+  submissionDetails: VariantAnalysisSubmission
+): Promise<VariantAnalysis> {
+  const octokit = await credentials.getOctokit();
+
+  const { actionRepoRef, query, databases, controllerRepoId } = submissionDetails;
+
+  const data: VariantAnalysisSubmissionRequest = {
+    action_repo_ref: actionRepoRef,
+    language: query.language as VariantAnalysisQueryLanguage,
```
koesie10, line 21: I don't think the casting is necessary if the enum and the string union match in the values they accept. By skipping the casting here, we will catch errors when they are out-of-sync.

```suggestion
    language: query.language,
```

extensions/ql-vscode/src/remote-queries/gh-api/gh-api-client.ts
```diff
@@ -0,0 +1,86 @@
...
+    query_pack: query.pack,
+    repositories: databases.repositories,
+    repository_lists: databases.repositoryLists,
+    repository_owners: databases.repositoryOwners,
+  };
+
+  const response: OctokitResponse<VariantAnalysis, number> = await octokit.request(
```
koesie10, line 28: `number` is already the default of the second parameter to `OctokitResponse`, so this should be sufficient:

```suggestion
  const response: OctokitResponse<VariantAnalysis> = await octokit.request(
```

extensions/ql-vscode/src/remote-queries/gh-api/gh-api-client.ts
```diff
@@ -0,0 +1,86 @@
...
+  VariantAnalysisQueryLanguage,
+  VariantAnalysisRepoTask,
+  VariantAnalysisSubmissionRequest
+} from './variant-analysis';
+
+export async function submitVariantAnalysis(
+  credentials: Credentials,
+  submissionDetails: VariantAnalysisSubmission
+): Promise<VariantAnalysis> {
+  const octokit = await credentials.getOctokit();
```
koesie10, line 15: Would it make sense to make `octokit` an argument to the function rather than the `credentials`? I realize that we're currently using the `credentials` everywhere, but we don't really care about the actual credentials; we only need the Octokit client.

This might also make testing a bit easier since we don't need to mock this promise but can directly mock Octokit.

koesie10 (approved):

PR #1528, opened 2022-09-21
Fix "Export All" not always exporting the correct query

koesie10:
The "Export All" button was always exporting the selected query, while a different query could be open in a VSCode panel. This will ensure that the query ID is passed to the export function, so that the correct query is exported.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -680,7 +680,7 @@ export class QueryHistoryManager extends DisposableObject {
     return this.treeDataProvider.getCurrent();
   }
 
-  getQueryById(queryId: string): RemoteQueryHistoryItem | undefined {
+  getRemoteQueryById(queryId: string): RemoteQueryHistoryItem | undefined {
     return this.treeDataProvider.allHistory.find(i => i.t === 'remote' && i.queryId === queryId) as RemoteQueryHistoryItem;
   }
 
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/query-history.ts
```diff
@@ -680,6 +680,10 @@ export class QueryHistoryManager extends DisposableObject {
     return this.treeDataProvider.getCurrent();
   }
 
+  getQueryById(queryId: string): RemoteQueryHistoryItem | undefined {
+    return this.treeDataProvider.allHistory.find(i => i.t === 'remote' && i.queryId === queryId) as RemoteQueryHistoryItem;
```
charisk, line 684: Why are we filtering only for remote queries? Is it to return `RemoteQueryHistoryItem`? In that case perhaps the function should be called `getRemoteQueryById`?

It'd be nice if we just had a generic `getQueryById` that returns QueryHistoryItem's though rather than anything more specific.

charisk (approved):
Generally looks good, just some minor non-blocking comments!

extensions/ql-vscode/src/query-history.ts
koesie10: I'll rename it to `getRemoteQueryById`. The reason for not returning any local queries is that local queries [don't have a query ID](https://github.com/github/vscode-codeql/blob/1b6685ef6ff208812af7350be013c842dcefd484/extensions/ql-vscode/src/query-results.ts#L213), so the `i.t === 'remote'` is necessary to even be able to access the `queryId` field. Without it, the union type doesn't necessarily have a `queryId` field.

PR #1538, opened 2022-09-22
Implement submitting a live-results variant analysis

robertbrignull:
Implements submitting a variant analysis to the new live results endpoint when the live results config is set. Doesn't yet implement the full notification when the analysis is submitted and doesn't implement any monitoring, so currently the variant analysis ID is printed to the logs but I expect that to change in the future.

Also adds a dedicated `getControllerRepoId` method before we get into anything else. This is not strictly necessary but it lets us give better error messages and means that later on we can sure that the controller repo exists.

Finally, also adds some tests that have the live results config set. I've copied the existing tests by adding all four cases of different query packs. Not sure it's strictly needed to have all four in both cases but I've gone with consistency.

## Checklist

  - Not necessary. Not yet a visible change.
  - No user-facing changes

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
+++ b/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
@@ -243,8 +243,8 @@ export async function runRemoteQuery(
 
     if (isVariantAnalysisLiveResultsEnabled()) {
       const queryName = getQueryName(queryMetadata, queryFile);
-      const variantAnalysisLangauge = parseVariantAnalysisQueryLanguage(language);
-      if (variantAnalysisLangauge === undefined) {
+      const variantAnalysisLanguage = parseVariantAnalysisQueryLanguage(language);
+      if (variantAnalysisLanguage === undefined) {
         throw new UserCancellationException(`Found unsupported language: ${language}`);
       }
 
@@ -256,7 +256,7 @@ export async function runRemoteQuery(
           name: queryName,
           filePath: queryFile,
           pack: base64Pack,
-          language: variantAnalysisLangauge,
+          language: variantAnalysisLanguage,
         },
         databases: {
           repositories: repoSelection.repositories,
--- a/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
+++ b/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
@@ -269,7 +269,7 @@ export async function runRemoteQuery(
         credentials,
         variantAnalysisSubmission
       );
-      await showAndLogInformationMessage('Variant analysis submitted for processing');
+      void showAndLogInformationMessage('Variant analysis submitted for processing');
       void logger.log(`Variant analysis result:\n${JSON.stringify(variantAnalysisResponse)}`);
       return;
 
--- a/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
+++ b/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
@@ -19,11 +19,12 @@ import * as cli from '../cli';
 import { logger } from '../logging';
 import { getActionBranch, getRemoteControllerRepo, setRemoteControllerRepo } from '../config';
 import { ProgressCallback, UserCancellationException } from '../commandRunner';
-import { OctokitResponse } from '@octokit/types/dist-types';
+import { OctokitResponse, RequestError } from '@octokit/types/dist-types';
 import { RemoteQuery } from './remote-query';
 import { RemoteQuerySubmissionResult } from './remote-query-submission-result';
 import { QueryMetadata } from '../pure/interface-types';
 import { getErrorMessage, REPO_REGEX } from '../pure/helpers-pure';
+import * as ghApiClient from './gh-api/gh-api-client';
 import { getRepositorySelection, isValidSelection, RepositorySelection } from './repository-selection';
 
 export interface QlPack {
@@ -212,29 +213,31 @@ export async function runRemoteQuery(
 
     // Get the controller repo from the config, if it exists.
     // If it doesn't exist, prompt the user to enter it, and save that value to the config.
-    let controllerRepo: string | undefined;
-    controllerRepo = getRemoteControllerRepo();
-    if (!controllerRepo || !REPO_REGEX.test(controllerRepo)) {
-      void logger.log(controllerRepo ? 'Invalid controller repository name.' : 'No controller repository defined.');
-      controllerRepo = await window.showInputBox({
+    let controllerRepoNwo: string | undefined;
+    controllerRepoNwo = getRemoteControllerRepo();
+    if (!controllerRepoNwo || !REPO_REGEX.test(controllerRepoNwo)) {
+      void logger.log(controllerRepoNwo ? 'Invalid controller repository name.' : 'No controller repository defined.');
+      controllerRepoNwo = await window.showInputBox({
         title: 'Controller repository in which to run the GitHub Actions workflow for this variant analysis',
         placeHolder: '<owner>/<repo>',
         prompt: 'Enter the name of a GitHub repository in the format <owner>/<repo>',
         ignoreFocusOut: true,
       });
-      if (!controllerRepo) {
+      if (!controllerRepoNwo) {
         void showAndLogErrorMessage('No controller repository entered.');
         return;
-      } else if (!REPO_REGEX.test(controllerRepo)) { // Check if user entered invalid input
+      } else if (!REPO_REGEX.test(controllerRepoNwo)) { // Check if user entered invalid input
         void showAndLogErrorMessage('Invalid repository format. Must be a valid GitHub repository in the format <owner>/<repo>.');
         return;
       }
-      void logger.log(`Setting the controller repository as: ${controllerRepo}`);
-      await setRemoteControllerRepo(controllerRepo);
+      void logger.log(`Setting the controller repository as: ${controllerRepoNwo}`);
+      await setRemoteControllerRepo(controllerRepoNwo);
     }
 
-    void logger.log(`Using controller repository: ${controllerRepo}`);
-    const [owner, repo] = controllerRepo.split('/');
+    void logger.log(`Using controller repository: ${controllerRepoNwo}`);
+    const [owner, repo] = controllerRepoNwo.split('/');
+    const controllerRepoId = await getControllerRepoId(credentials, owner, repo);
+    void logger.log(`Controller repository ID: ${controllerRepoId}`);
 
     progress({
       maxStep: 4,
@@ -450,4 +453,16 @@ async function buildRemoteQueryEntity(
     actionsWorkflowRunId: workflowRunId,
     repositoryCount,
   };
+}
+
+async function getControllerRepoId(credentials: Credentials, owner: string, repo: string): Promise<number> {
+  try {
+    return await ghApiClient.getRepositoryIdFromNwo(credentials, owner, repo);
+  } catch (e: any) {
+    if ((e as RequestError).status === 404) {
+      throw new Error(`Controller repository "${owner}/${repo}" not found`);
+    } else {
+      throw new Error(`Error getting controller repository "${owner}/${repo}": ${e.message}`);
+    }
+  }
 }
--- a/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
+++ b/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
@@ -17,7 +17,7 @@ import {
 import { Credentials } from '../authentication';
 import * as cli from '../cli';
 import { logger } from '../logging';
-import { getActionBranch, getRemoteControllerRepo, setRemoteControllerRepo } from '../config';
+import { getActionBranch, getRemoteControllerRepo, isVariantAnalysisLiveResultsEnabled, setRemoteControllerRepo } from '../config';
 import { ProgressCallback, UserCancellationException } from '../commandRunner';
 import { OctokitResponse, RequestError } from '@octokit/types/dist-types';
 import { RemoteQuery } from './remote-query';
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts
```diff
@@ -58,6 +60,9 @@ describe('Remote queries', function() {
     // always run in the vscode-codeql repo
     await setRemoteControllerRepo('github/vscode-codeql');
     await setRemoteRepositoryLists({ 'vscode-codeql': ['github/vscode-codeql'] });
+
+    // Consider live results disabled unless specifically enabled in a test
+    await setVariantAnalysisLiveResultsEnabled(false);
```
koesie10, line 65: We are now setting this value before each test, but that means that the value after this test file has run is that the value might be set to true depending on the order of tests. Since other test files will execute in the same VSCode instance, that might result in some different behavior for those tests as well.

I think it would be better to move this to the `afterEach` hook to reset the value after every test; that will ensure that the value will be `false` for certain after the test file has completed.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -451,3 +483,20 @@ async function buildRemoteQueryEntity(
...
 }
+
+function getQueryName(queryMetadata: QueryMetadata | undefined, queryFilePath: string): string {
+  // The query name is either the name as specified in the query metadata, or the file name.
+  return queryMetadata?.name ?? path.basename(queryFilePath);
+}
+
+async function getControllerRepoId(credentials: Credentials, owner: string, repo: string): Promise<number> {
+  try {
+    return await ghApiClient.getRepositoryIdFromNwo(credentials, owner, repo);
```
koesie10, line 494: Would it make sense to drop the `owner` and `repo` parameters from this method and instead call `getRemoteControllerRepo` in this method? The only thing that now differentiates this from the `getRepositoryIdFromNwo` method is the way error messages are handled; it doesn't actually just get the controller repo ID, you can pass any arbitrary owner/repo combination into it.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -259,31 +263,61 @@ export async function runRemoteQuery(
...
+      const queryName = getQueryName(queryMetadata, queryFile);
+      const variantAnalysisSubmission: VariantAnalysisSubmission = {
+        startTime: queryStartTime,
+        actionRepoRef: actionBranch,
+        controllerRepoId: controllerRepoId,
+        query: {
+          name: queryName,
+          filePath: queryFile,
+          pack: base64Pack,
+          language: language as VariantAnalysisQueryLanguage,
```
koesie10, line 279: Since this is being cast to the `VariantAnalysisQueryLanguage`, should we validate that the `language` is a supported MRVA language?

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts
```diff
@@ -259,6 +264,101 @@ describe('Remote queries', function() {
     }
   });
 
+  it('should run a variant analysis that is part of a qlpack', async () => {
+    await setVariantAnalysisLiveResultsEnabled(true);
```
koesie10, line 268: Perhaps we can move this to a nested `describe` block which has this as a `beforeEach` hook? Something like:

```ts
describe('when live results are enabled', () => {
  beforeEach(async () => {
    await setVariantAnalysisLiveResultsEnabled(true);
  });

  it('should run a variant analysis that is part of a qlpack', async () => {
    // ...
  });

  // ...
});
```

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts
robertbrignull: Thanks for bringing this up. So values like this aren't reset between tests? I wasn't sure so I added this before each test, but was hoping to clarify before merging the PR.

Unless it's too expensive (which I doubt) should we do it before and after, so we don't rely on other tests leaving a consistent state? I note we don't go to those lengths for the other config settings, but that's a different issue and doesn't mean we can't do something else.

What I'd like even more is a `resetConfig` method that would set all of the config to a consistent state. Then every test could call that in its `beforeEach` method and not have to worry about any leftover config changes. If you think that's a good idea I can open an issue to look at it.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
robertbrignull: I think that would be a good way of cleaning this code up by putting more implementation into a method. Unfortunately it's not quite as clean as it could be, mostly because we use the `owner` and `repo` fields later on in `runRemoteQuery` so we have to pass those back as well. But a nice way of doing that could be changing to passing around a `Repository` object instead of just the ID or owner or repo name separately. I've had a go at this and I think it's worked out quite nicely. It's slightly more code changes than strictly necessary but I think it's all going in the right direction.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
robertbrignull: I'm not sure. On the one hand it's nice to have accurate types, but then it's seems weird to do the type checking here. We are storing the language as a string elsewhere in the extension, so shouldn't we be applying this type earlier on (I realise we've only just introduced the type which is why it's not used elsewhere).

On the other hand the API itself does validation that the language is from the known set of languages, and returns an error message. So just passing the string to the API and letting it decide if the language is ok or not also seems like a good option to me.

Another question is are we going to ever do anything in the VSCode extension that depends on the exact language? I can't think of many cases. We'll generally either just display the language or pass it to other places like the CodeQL CLI or the github API. We might want to display a better formatted name (i.e. "JavaScript" instead of "javascript") but that's about all.

So my preference would actually be to delete the `VariantAnalysisQueryLanguage` type and just treat the language as a string. We could also keep the type but make it an alias of string. I just don't know if that's too radical of a decision to make.

What do you think?

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts
robertbrignull: Ooh, that's a good idea. I wasn't aware you could do nested describes with their own before/after hooks.

robertbrignull: This has increased the total number of lines changed quite a lot, but if you enable the mode to ignore whitespace then it'll look a lot more manageable.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
robertbrignull: I'm changing my mind. For this PR I'll just add a method to check if the language is valid or not, and we'll throw an exception if it's not. I'll leave the discussion of whether we want this type of not for a later time.

extensions/ql-vscode/test/pure-tests/variant-analysis.test.ts
```diff
@@ -0,0 +1,12 @@
+import { expect } from 'chai';
+import { parseVariantAnalysisQueryLanguage, VariantAnalysisQueryLanguage } from '../../src/remote-queries/shared/variant-analysis';
+
+describe('parseVariantAnalysisQueryLanguage', () => {
```
robertbrignull, line 4: Maybe this function is too simple to require / benefit from a test. I'm happy to remove it if that would be simpler.

I also wasn't sure where to put it. The test is completely pure in that it doesn't need a workspace, and I have found other examples of tests in the `test/pure-tests` directory that test things from `src/remove-queries`. But I could move it back to `src/vscode-tests/no-workspace/remote-queries/shared/` if that would be a better location for it.

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts
koesie10: I don't think the config is reset between each test, the only integration between VSCode and the test runner is the script in [`extensions/ql-vscode/src/vscode-tests/index-template.ts`](https://github.com/github/vscode-codeql/blob/8c7c197b2200a7547feafb052644fd7373239f04/extensions/ql-vscode/src/vscode-tests/index-template.ts) which doesn't seem to reset the config.

I think it would indeed be a good idea to have a `resetConfig` function which will reset all config values. We would then be able to run this before every test and have a consistent state without needing to reset the state before/after tests.

extensions/ql-vscode/test/pure-tests/variant-analysis.test.ts
koesie10: I'm also not quite sure how the distinction between those tests was originally determined. However, I do think we should always prefer the `test` directory over `vscode-tests`; the former is much quicker to run and less flaky. We could also consider placing it in a directory like `test/remote-queries` and adding that to the `test:unit` `package.json` script.

robertbrignull:
I've hit barriers which have necessitated a few more changes. The main one is that I've removed the `setVariantAnalysisLiveResultsEnabled` method and instead am using `sandbox.stub(config, 'isVariantAnalysisLiveResultsEnabled')` to control whether live results is enabled.

The reason here is that to set a value in this way, VSCode appears to require that the config field is registered in a defined list of [properties](https://github.com/github/vscode-codeql/blob/8c7c197b2200a7547feafb052644fd7373239f04/extensions/ql-vscode/package.json#L144) and doing this would make the `liveResults` configuration option appear as an autocomplete option and in the generated extension settings in the UI. It's my understanding that we don't want to do this and we want to hide this setting from anybody who isn't reading our code or pull requests. Using a stub is also the way that we [control `isCanary` in the tests](https://github.com/github/vscode-codeql/blob/8c7c197b2200a7547feafb052644fd7373239f04/extensions/ql-vscode/src/vscode-tests/no-workspace/run-queries.test.ts#L57).

extensions/ql-vscode/test/pure-tests/variant-analysis.test.ts
robertbrignull: I think I'll leave it here for now unless you have strong thoughts on this. Moving it in the future will be very easy.

robertbrignull:
Hmm, the new tests are timing out but I can't see why it's happening. I can't see anything obviously wrong or different. I can use the new code paths manually and they all seem to function, so perhaps the issue is in the tests. I just can't see where 😦 

```
  1) Remote queries
       when live results are enabled
         should run a variant analysis that is part of a qlpack:
     Error: Timeout of 180000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/vscode-codeql/vscode-codeql/extensions/ql-vscode/out/vscode-tests/cli-integration/remote-queries/run-remote-query.test.js)
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)
```

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -259,31 +238,65 @@ export async function runRemoteQuery(
...
-      return { query: remoteQuery };
+      const variantAnalysisSubmission: VariantAnalysisSubmission = {
+        startTime: queryStartTime,
+        actionRepoRef: actionBranch,
+        controllerRepoId: controllerRepo.id,
+        query: {
+          name: queryName,
+          filePath: queryFile,
+          pack: base64Pack,
+          language: variantAnalysisLangauge,
```
elenatanasoiu, line 259: Small thing: `variantAnalysisLangauge` looks like a typo.

robertbrignull: Yep. Fixed now.

robertbrignull:
All tests successful 🥳

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts
```diff
@@ -334,7 +334,7 @@ describe('Remote queries', function() {
       const fileUri = getFile('data-remote-qlpack-nested/subfolder/in-pack.ql');
 
       const querySubmissionResult = await runRemoteQuery(cli, credentials, fileUri, true, progress, token);
-      expect(querySubmissionResult).to.be.ok;
+      expect(querySubmissionResult).to.be.undefined;
```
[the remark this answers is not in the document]
elenatanasoiu: Alternatively I can see in other places we're calling this method without expecting a result to be returned. If that's the new behaviour, I think we should change the tests to stop expecting a result in the first place. So instead of:

```
const querySubmissionResult = await runRemoteQuery(cli, credentials, fileUri, true, progress, token);
```

we'd just call the method: 
```
await runRemoteQuery(cli, credentials, fileUri, true, progress, token);
```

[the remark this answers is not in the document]
robertbrignull: I was kinda putting this off for this PR and hoping to address it in the future when we know better what data we'll want returned from this method. But I think we can actually just return a `VariantAnalysis` object and that likely be exactly what we want, or at least it's a good starting point.

So I've adjusted `RemoteQuerySubmissionResult` to looks like the following so it can handle the live and non-live results cases.
```typescript
export interface RemoteQuerySubmissionResult {
  queryDirPath?: string;
  query?: RemoteQuery;
  variantAnalysis?: VariantAnalysis;
}
```

In the tests we can now assert that the object is non-undefined. I've gone for only checking a couple of fields on it. We could assert the values of more field but I think this is good enough for now?

Here's what the logging looks like. I expect we'll change this in the future now but this is a placeholder until we do more.
Screenshot 2022-09-27 at 10 14 31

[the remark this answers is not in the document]
elenatanasoiu: Nice! Thanks for changing it 👍

elenatanasoiu (approved):

PR #1539, opened 2022-09-23
Add Alert component for showing warnings and errors

koesie10:
This adds an `Alert` component which is able to show warnings and errors in the correct styles.

### Dark theme

Screenshot 2022-09-23 at 15 24 57

Screenshot 2022-09-23 at 15 25 13

Screenshot 2022-09-23 at 15 25 26

Screenshot 2022-09-23 at 15 25 40

Screenshot 2022-09-23 at 15 27 14

### Light theme

Screenshot 2022-09-23 at 15 26 13

Screenshot 2022-09-23 at 15 26 24

Screenshot 2022-09-23 at 15 26 37

Screenshot 2022-09-23 at 15 26 47

Screenshot 2022-09-23 at 15 28 03

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/common/Alert.tsx
+++ b/extensions/ql-vscode/src/view/common/Alert.tsx
@@ -42,6 +42,15 @@ const getBorderColor = ({ type }: ContainerProps): string => {
   }
 };
 
+const getTypeText = (type: ContainerProps['type']): string => {
+  switch (type) {
+    case 'warning':
+      return 'Warning';
+    case 'error':
+      return 'Error';
+  }
+};
+
 const Container = styled.div<ContainerProps>`
   display: flex;
   flex-direction: column;
@@ -81,7 +90,7 @@ type Props = {
 export const Alert = ({ type, title, message, actions, inverse }: Props) => {
   return (
     <Container type={type} inverse={inverse}>
-      <Title>{type}: {title}</Title>
+      <Title>{getTypeText(type)}: {title}</Title>
       <span>{message}</span>
       {actions && <ActionsContainer>{actions}</ActionsContainer>}
     </Container>
--- a/extensions/ql-vscode/src/view/common/Alert.tsx
+++ b/extensions/ql-vscode/src/view/common/Alert.tsx
@@ -1,0 +1,98 @@
+import * as React from 'react';
+import { ReactNode } from 'react';
+import styled from 'styled-components';
+
+type ContainerProps = {
+  type: 'warning' | 'error';
+  inverse?: boolean;
+};
+
+const getBackgroundColor = ({ type, inverse }: ContainerProps): string => {
+  if (!inverse) {
+    return 'var(--vscode-notifications-background)';
+  }
+
+  switch (type) {
+    case 'warning':
+      return 'var(--vscode-editorWarning-foreground)';
+    case 'error':
+      return 'var(--vscode-debugExceptionWidget-border)';
+  }
+};
+
+const getTextColor = ({ type, inverse }: ContainerProps): string => {
+  if (!inverse) {
+    return 'var(--vscode-tab-activeForeground)';
+  }
+
+  switch (type) {
+    case 'warning':
+      return 'var(--vscode-panel-background)';
+    case 'error':
+      return 'var(--vscode-list-activeSelectionForeground)';
+  }
+};
+
+const getBorderColor = ({ type, inverse }: ContainerProps): string => {
+  if (inverse) {
+    switch (type) {
+      case 'warning':
+        return 'var(--vscode-inputValidation-warningBorder)';
+      case 'error':
+        return 'var(--vscode-inputValidation-errorBorder)';
+    }
+  } else {
+    switch (type) {
+      case 'warning':
+        return 'var(--vscode-notificationsWarningIcon-foreground)';
+      case 'error':
+        return 'var(--vscode-notificationsErrorIcon-foreground)';
+    }
+  }
+};
+
+const Container = styled.div<ContainerProps>`
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  gap: 1em;
+  padding: 1em;
+
+  color: ${props => getTextColor(props)};
+  background-color: ${props => getBackgroundColor(props)};
+  border: 1px solid ${props => getBorderColor(props)};
+`;
+
+const Title = styled.div`
+  font-size: 0.85em;
+  font-weight: 800;
+  text-transform: uppercase;
+`;
+
+const ActionsContainer = styled.div`
+  display: flex;
+  flex-direction: row;
+  gap: 0.75em;
+  margin-left: auto;
+`;
+
+type Props = {
+  type: 'warning' | 'error';
+  title: string;
+  message: ReactNode;
+
+  actions?: ReactNode;
+
+  // Inverse the color scheme
+  inverse?: boolean;
+};
+
+export const Alert = ({ type, title, message, actions, inverse }: Props) => {
+  return (
+    <Container type={type} inverse={inverse}>
+      <Title>{type}: {title}</Title>
+      <span>{message}</span>
+      {actions && <ActionsContainer>{actions}</ActionsContainer>}
+    </Container>
+  );
+};
--- a/extensions/ql-vscode/src/view/common/Alert.tsx
+++ b/extensions/ql-vscode/src/view/common/Alert.tsx
@@ -22,32 +22,23 @@ const getBackgroundColor = ({ type, inverse }: ContainerProps): string => {
 
 const getTextColor = ({ type, inverse }: ContainerProps): string => {
   if (!inverse) {
-    return 'var(--vscode-tab-activeForeground)';
+    return 'var(--vscode-editor-foreground)';
   }
 
   switch (type) {
     case 'warning':
-      return 'var(--vscode-panel-background)';
+      return 'var(--vscode-editor-background)';
     case 'error':
       return 'var(--vscode-list-activeSelectionForeground)';
   }
 };
 
-const getBorderColor = ({ type, inverse }: ContainerProps): string => {
-  if (inverse) {
-    switch (type) {
-      case 'warning':
-        return 'var(--vscode-inputValidation-warningBorder)';
-      case 'error':
-        return 'var(--vscode-inputValidation-errorBorder)';
-    }
-  } else {
-    switch (type) {
-      case 'warning':
-        return 'var(--vscode-notificationsWarningIcon-foreground)';
-      case 'error':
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/view/common/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+import styled from 'styled-components';
+
+type ContainerProps = {
+  type: 'warning' | 'error';
+  inverse?: boolean;
+};
+
+const getBackgroundColor = ({ type, inverse }: ContainerProps): string => {
+  if (!inverse) {
+    return 'var(--vscode-notifications-background)';
```
robertbrignull, line 12: For the colours chosen throughout this file (not just this particular one I've put the comment on) are these the ones that most-closely match the designs? It's unfortunate they don't really form a consistent set, as in we've got ones from `notifications`, and `inputValidation`, and `debugExceptionWidget`.

Does there exist a set we could pick that are meant to go together, and that way it would fit the VSCode colour scheme even though it's slightly different from the designs? What approach have we taken for other components?

I'm going to have a look through for ones I think might fit, but I'll submit those as a separate review so they're all grouped together.

robertbrignull (commented):
Component looks great. I'm also going to take some bits as inspiration and copy them over to my PR at https://github.com/github/vscode-codeql/pull/1541

I have one question about colours, and as I say in the comment I'm going to submit a second review focussed on that.

extensions/ql-vscode/src/view/common/Alert.tsx
robertbrignull, line 12:
```suggestion
    return 'var(--vscode-editor-background)';
```
Changes from #252526 to #1e1e1e

extensions/ql-vscode/src/view/common/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  switch (type) {
+    case 'warning':
+      return 'var(--vscode-editorWarning-foreground)';
+    case 'error':
+      return 'var(--vscode-debugExceptionWidget-border)';
```
robertbrignull, line 19:
```suggestion
      return 'var(--vscode-editorError-foreground)';
```
Changes from #a31515 to #f14c4c

extensions/ql-vscode/src/view/common/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+  }
+};
+
+const getTextColor = ({ type, inverse }: ContainerProps): string => {
+  if (!inverse) {
+    return 'var(--vscode-tab-activeForeground)';
```
robertbrignull, line 25:
```suggestion
    return 'var(--vscode-editor-foreground)';
```
Changes from #ffffff to #d4d4d4

Probably the biggest change in these suggestions. Does VSCode use pure white for text in dark mode? From what I can tell it does seem to be slightly darker generally, which is nicer on the eyes.

extensions/ql-vscode/src/view/common/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  switch (type) {
+    case 'warning':
+      return 'var(--vscode-panel-background)';
+    case 'error':
+      return 'var(--vscode-list-activeSelectionForeground)';
```
robertbrignull, line 32:
```suggestion
      return 'var(--vscode-editor-foreground)';
```
Changes from #ffffff to #d4d4d4

extensions/ql-vscode/src/view/common/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+};
+
+const getTextColor = ({ type, inverse }: ContainerProps): string => {
+  if (!inverse) {
+    return 'var(--vscode-tab-activeForeground)';
+  }
+
+  switch (type) {
+    case 'warning':
+      return 'var(--vscode-panel-background)';
```
robertbrignull, line 30:
```suggestion
      return 'var(--vscode-editor-background)';
```
Doesn't actually change the colour but would fit in since I'm suggesting other `editor` colours elsewhere 🤷

extensions/ql-vscode/src/view/common/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+    case 'error':
+      return 'var(--vscode-list-activeSelectionForeground)';
+  }
+};
+
+const getBorderColor = ({ type, inverse }: ContainerProps): string => {
+  if (inverse) {
+    switch (type) {
+      case 'warning':
+        return 'var(--vscode-inputValidation-warningBorder)';
```
robertbrignull, line 40:
```suggestion
        return 'var(--vscode-editorWarning-foreground)';
```
Changes from #b89500 to #cca700

extensions/ql-vscode/src/view/common/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+      case 'error':
+        return 'var(--vscode-inputValidation-errorBorder)';
```
robertbrignull, line 42:
```suggestion
        return 'var(--vscode-editorError-foreground)';
```
Changes from #be1100 to #f14c4c

extensions/ql-vscode/src/view/common/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+    }
+  } else {
+    switch (type) {
+      case 'warning':
+        return 'var(--vscode-notificationsWarningIcon-foreground)';
```
robertbrignull, line 47:
```suggestion
        return 'var(--vscode-editorWarning-foreground)';
```
Doesn't change the colour, but standardises on using the same name everywhere.

extensions/ql-vscode/src/view/common/Alert.tsx
```diff
@@ -0,0 +1,98 @@
...
+      case 'error':
+        return 'var(--vscode-notificationsErrorIcon-foreground)';
```
robertbrignull, line 49:
```suggestion
        return 'var(--vscode-editorError-foreground)';
```
Doesn't change the colour, but standardises on using the same name everywhere.

robertbrignull (commented):
Potential suggestions for colour changes. My selection criteria here has been:
- Use the minimal number of colours. So if we refer to a red in multiple places, then use the same red everywhere.
- Try to pick a general colour and not one that's meant for a specialised purpose, e.g. preferring `editor` colours over `list-activeSelection`.

I'm not trying to say the colours I've picked here are perfect, but I'm interested to hear what you think. I definitely think we should standardise so we aren't using slightly different colours in different modes, or multiples names for the same colour.

extensions/ql-vscode/src/view/common/__tests__/Alert.spec.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { render, screen } from '@testing-library/react';
+import { Alert } from '../Alert';
+
+describe(Alert.name, () => {
+  it('renders a warning correctly', () => {
+    render(<Alert type="warning" title="Warning title" message="Warning content" />);
+
+    expect(screen.getByText('warning: Warning title')).toBeInTheDocument();
```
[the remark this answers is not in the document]
koesie10: It is indeed uppercased by the styling, but you're right that when you copy the text it would show as `warning`. Perhaps it would be even better to use `Warning`? That would make the most sense when copying this text, such that is shows as e.g. `Warning: Access mismatch`.

extensions/ql-vscode/src/view/common/Alert.tsx
koesie10: I agree that it would be better to use standardized color names, but I chose the colors that fit the designs. If we can deviate from the design, I would also prefer to use less colors and standardize on them.

koesie10: I think this change deviates quite a lot from the design. The design specifies that the alerts have a different background color from the editor background and this would completely remove that. I think `notifications-background` does make sense for this, since we are essentially showing a notification inline.

extensions/ql-vscode/src/view/common/Alert.tsx
koesie10: I did think about using the normal `editorError` color, but the contrast ratio between the foreground and background color is quite small with `editorError-foreground`. This makes the text somewhat hard to read. These are the results from the [WebAIM contrast checker](https://webaim.org/resources/contrastchecker/):

Screenshot 2022-09-26 at 13 51 14

Screenshot 2022-09-26 at 13 51 24

Using black text instead of white text improves this somewhat, but since the text in the alerts is quite small, I still don't think it will be enough.

Screenshot 2022-09-26 at 13 51 35

extensions/ql-vscode/src/view/common/Alert.tsx
koesie10: VSCode does use light gray instead of white and it makes sense to be consistent there, so I'll change this.

extensions/ql-vscode/src/view/common/Alert.tsx
koesie10: Unfortunately, this makes the text quite unreadable when combined with the different error background color:

Screenshot 2022-09-26 at 13 56 11

Even with the current background color, this makes it harder to read the text.

extensions/ql-vscode/src/view/common/__tests__/Alert.spec.tsx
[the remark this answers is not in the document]
robertbrignull: I agree that capitalisation would be the ideal 👍

extensions/ql-vscode/src/view/common/Alert.tsx
robertbrignull: Thanks for checking. I didn't know about that contrast checker, but that's a very handy tool to bear in mind.

extensions/ql-vscode/src/view/common/Alert.tsx
robertbrignull: Good point. I had not considered that it needs to sit on top of an editor background. And you're right it is a notification.

robertbrignull: I don't know whether the exact colours need to be adhered to, so long as we have the basic colours and readable text, and colours are either the same or different where they're meant to be. Something that's in my mind is we're optimizing for the standard dark mode but then there are other colour schemes that people could be using.

I'm also aware that changing these colours at a later date will be very easy, so I don't want to hold up our progress here unnecessarily. I'm thinking that for this PR we stick closer to the designs, and I'm happy to go with what in this PR currently, and then we check with `@asiermartinez` internally on how to approach this question in the future.

robertbrignull (approved):

PR #1540, opened 2022-09-23
Ensure `safeMax` is safe for undefined values

aeisenberg:
I came across this when I had a query that threw an error while running for unrelated reasons. At this point, the query results were in a bad state, but this caused `safeMax` to be called with `undefined` and it prevented the extension from starting. This changed fixed the error.


angelapwen (approved):
I looked into what happens when `Math.max()` is called on an empty value and it looks like it [returns infinity](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max). So, that makes sense because then the next line will return 0. Today I learned!

PR #1541, opened 2022-09-23
Add loading component

robertbrignull:
Implements a component to show when the variant analysis submission is being processed. This component doesn't have any props as there's only state it needs to show. The decision on whether to show it or not based on the variant analysis status will be decided by the parent component.

Storybook screenshot:
Screenshot 2022-09-23 at 17 21 48

See the internal issue for how it's meant to look. I've tried to copy the style as closely as possible but please do tell me if this could fit in better.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisLoading.stories.tsx
+++ b/extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisLoading.stories.tsx
@@ -1,0 +1,25 @@
+import React from 'react';
+
+import { ComponentMeta, ComponentStory } from '@storybook/react';
+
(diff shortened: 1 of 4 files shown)


robertbrignull:
Ok, just spotted that I should hook this to the main view as well. I'll look at doing that next week.

koesie10 (approved):

robertbrignull:
@koesie10, I've pushed a couple more commits, if you could re-review.

I've adjusted the styling to fit better with how we do it in other components, and also to add some padding because it looked bad without it.

I've also hooked it into the main variant analysis component, and it would be shown if you don't set the action workflow ID. I haven't made it so you can view both states in the story book because currently the variant analysis component doesn't take any props, but that could be something to do in the future.

koesie10 (approved):

PR #1545, opened 2022-09-28
Implement monitoring for variant analysis live results 

elenatanasoiu:

Continues from a [previous PR](https://github.com/github/vscode-codeql/pull/1538) which introduced a way to submit new
variant analyses to the API.

Here we're creating a way to keep track of variant analysis results as
they become ready. To do this we're introducing a new 
`VariantAnalysisMonitor` class. This will poll the API every 5 seconds 
for changes to the variant analysis. By default it will continue to run for 
a maximum of 2 days, or when the user closes VSCode.

The monitor will receive a `variantAnalysis` summary from the API that
will contain an up-to-date list of scanned repos and will return a list of
scanned repo ids.

In a future PR, we'll add the functionality to:
- update the UI for in-progress/completed states
- raise an error on timeout
- download the results

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
@@ -49,7 +49,7 @@ export class VariantAnalysisMonitor {
         variantAnalysis.id
       );
 
-      if (variantAnalysisSummary.status == 'in_progress' && variantAnalysisSummary.failure_reason) {
+      if (variantAnalysisSummary.failure_reason) {
         variantAnalysis.status = VariantAnalysisStatus.Failed;
         variantAnalysis.failureReason = processFailureReason(variantAnalysisSummary.failure_reason);
         return {
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
@@ -69,6 +69,10 @@ export class VariantAnalysisMonitor {
         });
       }
 
+      if (variantAnalysisSummary.status === 'completed') {
+        break;
+      }
+
       attemptCount++;
     }
 
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
@@ -1,0 +1,81 @@
+import * as vscode from 'vscode';
+import { Credentials } from '../authentication';
+import { Logger } from '../logging';
+import * as ghApiClient from './gh-api/gh-api-client';
+
+import { VariantAnalysis, VariantAnalysisStatus } from './shared/variant-analysis';
+import {
+  VariantAnalysis as VariantAnalysisApiResponse
+} from './gh-api/variant-analysis';
+import { VariantAnalysisMonitorResult } from './shared/variant-analysis-monitor-result';
+import { processFailureReason } from './variant-analysis-processor';
+
+export class VariantAnalysisMonitor {
+  // With a sleep of 5 seconds, the maximum number of attempts takes
+  // us to just over 2 days worth of monitoring.
+  public static maxAttemptCount = 17280;
+  public static sleepTime = 5000;
+
+  constructor(
+    private readonly extensionContext: vscode.ExtensionContext,
+    private readonly logger: Logger
+  ) {
+  }
+
+  public async monitorVariantAnalysis(
+    variantAnalysis: VariantAnalysis,
+    cancellationToken: vscode.CancellationToken
+  ): Promise<VariantAnalysisMonitorResult> {
+
+    const credentials = await Credentials.initialize(this.extensionContext);
+    if (!credentials) {
+      throw Error('Error authenticating with GitHub');
+    }
+
+    let variantAnalysisSummary: VariantAnalysisApiResponse;
+    let attemptCount = 0;
+    const scannedReposDownloaded: number[] = [];
+
+    while (attemptCount <= VariantAnalysisMonitor.maxAttemptCount) {
+      await this.sleep(VariantAnalysisMonitor.sleepTime);
+
+      if (cancellationToken && cancellationToken.isCancellationRequested) {
+        return { status: 'Cancelled', error: 'Variant Analysis was canceled.' };
+      }
+
+      variantAnalysisSummary = await ghApiClient.getVariantAnalysis(
+        credentials,
+        variantAnalysis.controllerRepoId,
+        variantAnalysis.id
+      );
+
+      if (variantAnalysisSummary.status == 'in_progress' && variantAnalysisSummary.failure_reason) {
+        variantAnalysis.status = VariantAnalysisStatus.Failed;
+        variantAnalysis.failureReason = processFailureReason(variantAnalysisSummary.failure_reason);
+        return {
+          status: 'Failed',
+          error: `Variant Analysis has failed: ${variantAnalysisSummary.failure_reason}`,
+          variantAnalysis: variantAnalysis
+        };
+      }
+
+      void this.logger.log('****** Retrieved variant analysis' + JSON.stringify(variantAnalysisSummary));
+
+      if (variantAnalysisSummary.scanned_repositories) {
+        variantAnalysisSummary.scanned_repositories.forEach(scannedRepo => {
+          if (!scannedReposDownloaded.includes(scannedRepo.repository.id) && scannedRepo.analysis_status === 'succeeded') {
+            scannedReposDownloaded.push(scannedRepo.repository.id);
+          }
+        });
+      }
+
+      attemptCount++;
+    }
+
+    return { status: 'CompletedSuccessfully', scannedReposDownloaded: scannedReposDownloaded };
+  }
+
+  private async sleep(ms: number) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
+}
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-processor.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-processor.ts
@@ -102,7 +102,7 @@ function processRepoGroup(repoGroup: ApiVariantAnalysisSkippedRepositoryGroup): 
 }
 
 function processNotFoundRepoGroup(repoGroup: ApiVariantAnalysisNotFoundRepositoryGroup): VariantAnalysisSkippedRepositoryGroup {
-  const repo_nwos = repoGroup.repository_nwos.map(nwo => {
+  const repo_full_names = repoGroup.repository_full_names.map(nwo => {
     return {
       fullName: nwo
     };
@@ -110,7 +110,7 @@ function processNotFoundRepoGroup(repoGroup: ApiVariantAnalysisNotFoundRepositor
 
   return {
     repositoryCount: repoGroup.repository_count,
-    repositories: repo_nwos
+    repositories: repo_full_names
   };
 }
 
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-processor.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-processor.ts
@@ -60,11 +60,8 @@ export function processVariantAnalysis(
 function processScannedRepositories(
   scannedRepos: ApiVariantAnalysisScannedRepository[]
 ): VariantAnalysisScannedRepository[] {
-
-  const result: VariantAnalysisScannedRepository[] = [];
-
-  scannedRepos.forEach(function(scannedRepo) {
-    const parsedRepo: VariantAnalysisScannedRepository = {
+  return scannedRepos.map(scannedRepo => {
+    return {
       repository: {
         id: scannedRepo.repository.id,
         fullName: scannedRepo.repository.full_name,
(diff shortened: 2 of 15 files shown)


extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/variant-analysis-submission.ts
```diff
@@ -0,0 +1,20 @@
...
+    query: {
+      name: 'query-name',
+      filePath: 'query-file-path',
+      language: VariantAnalysisQueryLanguage.Javascript,
+      pack: 'base64-encoded-string',
+    },
+    databases: {
+      repositories: ['1', '2', '3'],
+      repositoryLists: ['top10', 'top100'],
+      repositoryOwners: ['mona', 'lisa'],
```
koesie10, line 17: Would we usually have all three populated? I would expect this to normally only have one of these fields populated.

extensions/ql-vscode/src/vscode-tests/factories/remote-queries/gh-api/variant-analysis-api-response.ts
```diff
@@ -2,14 +2,18 @@ import {
...
 import {
   VariantAnalysisQueryLanguage
 } from '../../../../remote-queries/shared/variant-analysis';
+import { createMockScannedRepos } from './scanned-repositories';
+import { createMockSkippedRepos } from './skipped-repositories';
 
 export function createMockApiResponse(
-  scannedRepos: VariantAnalysisScannedRepository[],
-  skippedRepos: VariantAnalysisSkippedRepositories
+  status = 'in_progress',
```
koesie10, line 14: Should this have the `VariantAnalysisStatus` type instead of casting the string to `VariantAnalysisStatus` on line 29?

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
```diff
@@ -0,0 +1,81 @@
...
+    const scannedReposDownloaded: number[] = [];
+
+    while (attemptCount <= VariantAnalysisMonitor.maxAttemptCount) {
+      await this.sleep(VariantAnalysisMonitor.sleepTime);
+
+      if (cancellationToken && cancellationToken.isCancellationRequested) {
+        return { status: 'Cancelled', error: 'Variant Analysis was canceled.' };
+      }
+
+      variantAnalysisSummary = await ghApiClient.getVariantAnalysis(
```
koesie10, line 46: Does it make sense to move the variable declaration to inside the `while` loop? It doesn't seem to be used outside of it, so making it a `const` would make this somewhat more robust.

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
```diff
@@ -0,0 +1,81 @@
...
+          if (!scannedReposDownloaded.includes(scannedRepo.repository.id) && scannedRepo.analysis_status === 'succeeded') {
+            scannedReposDownloaded.push(scannedRepo.repository.id);
+          }
+        });
+      }
+
+      attemptCount++;
+    }
+
+    return { status: 'CompletedSuccessfully', scannedReposDownloaded: scannedReposDownloaded };
```
koesie10, line 75: Is it just the IDs of the repos that we want to return? Do we not want to return the complete summary which includes the result count and artifact URL of the scanned repos so that we can show/download those?

extensions/ql-vscode/src/remote-queries/variant-analysis-processor.ts
```diff
@@ -0,0 +1,157 @@
...
+        fullName: scannedRepo.repository.full_name,
+        private: scannedRepo.repository.private,
+      },
+      analysisStatus: processApiRepoStatus(scannedRepo.analysis_status),
+      resultCount: scannedRepo.result_count,
+      artifactSizeInBytes: scannedRepo.artifact_size_in_bytes,
+      failureMessage: scannedRepo.failure_message
+    };
+
+    result.push(parsedRepo);
```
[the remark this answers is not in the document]
elenatanasoiu: Yup! Have changed.

extensions/ql-vscode/src/remote-queries/variant-analysis-processor.ts
```diff
@@ -0,0 +1,157 @@
...
+  });
+
+  return {
+    repositoryCount: repoGroup.repository_count,
+    repositories: repos
+  };
+}
+
+function processNotFoundRepoGroup(repoGroup: ApiVariantAnalysisNotFoundRepositoryGroup): VariantAnalysisSkippedRepositoryGroup {
+  const repo_nwos = repoGroup.repository_nwos.map(nwo => {
```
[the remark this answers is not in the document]
elenatanasoiu: Thanks, have changed now. 

There's still the `VariantAnalysisSkippedRepositories` which has a key called `not_found_repo_nwos`. I'm going to leave it as is since it doesn't look like we've changed the API for this key.

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-processor.test.ts
```diff
@@ -0,0 +1,231 @@
...
+    const repoName1 = 'github' + faker.random.word();
+    const repoName2 = 'github' + faker.random.word();
+
+    return {
+      repository_count: 2,
+      repository_nwos: [repoName1, repoName2]
+    };
+  }
+
+  function transformSkippedRepos(
```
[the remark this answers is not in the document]
elenatanasoiu: That's fair, I got a little bit carried away. Have made it more verbose for skipped repos.

extensions/ql-vscode/src/vscode-tests/factories/remote-queries/gh-api/scanned-repositories.ts
```diff
@@ -0,0 +1,31 @@
...
+    repository: {
+      id: faker.datatype.number(),
+      name: name,
+      full_name: 'github/' + name,
+      private: isPrivate,
+    },
+    analysis_status: analysisStatus,
+    result_count: faker.datatype.number(),
+    artifact_size_in_bytes: faker.datatype.number(),
+    failure_message: ''
```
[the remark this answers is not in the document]
elenatanasoiu: I've left it out 👍

extensions/ql-vscode/src/vscode-tests/factories/remote-queries/gh-api/skipped-repositories.ts
```diff
@@ -0,0 +1,45 @@
...
+        full_name: 'github/' + faker.random.word(),
+        private: false
+      }
+    ]
+  };
+}
+
+export function createMockNotFoundSkippedRepoGroup(): VariantAnalysisNotFoundRepositoryGroup {
+  const repoName1 = 'github' + faker.random.word();
+  const repoName2 = 'github' + faker.random.word();
```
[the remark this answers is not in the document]
elenatanasoiu: Have added a missing `/` to them.

extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/variant-analysis-submission.ts
elenatanasoiu: You're right, at the moment we only ever send one and even validate to check that only one of these fields is present in the API. 

However in the future we do intend to make it possible to send combinations of these. 

🤔  I think for now it does makes sense to keep just one of them to be consistent with what the API is doing.

elenatanasoiu: I've left `repositories`.

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
elenatanasoiu: Your observation is spot on. At the moment there's not much point to declaring it outside the while loop. 

However, I've left it like this since I intend to use it in a future PR in order to check on the summary's finished stated (e.g. did it time out or fail?). You can see a sketch of this in one of my other branches: https://github.com/github/vscode-codeql/compare/elenatanasoiu/monitor-variant-analysis-continued#diff-65d29f358913c786586c8fdf81a804c51bc1853333b8b02ccc6c950a4f0ee757R88-R98.

extensions/ql-vscode/src/remote-queries/variant-analysis-processor.ts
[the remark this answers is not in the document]
koesie10: Thanks!

I don't think the API is still using `not_found_repo_nwos` anywhere. From the REST API documentation it seems like that field should be called `not_found_repos`.

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
```diff
@@ -0,0 +1,81 @@
...
+        return { status: 'Cancelled', error: 'Variant Analysis was canceled.' };
+      }
+
+      variantAnalysisSummary = await ghApiClient.getVariantAnalysis(
+        credentials,
+        variantAnalysis.controllerRepoId,
+        variantAnalysis.id
+      );
+
+      if (variantAnalysisSummary.status == 'in_progress' && variantAnalysisSummary.failure_reason) {
```
robertbrignull, line 52: Is it possible to have `status: in_progress` but also have a `failure_reason`? Could you explain this bit here a little more?

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
elenatanasoiu: Just for the sake of clarity, I had a chat with Robert about what this monitor is meant to return and based on what he described in [the comment](https://github.com/github/vscode-codeql/pull/1545#discussion_r982591809) below, we decided not to return anything from the monitor and instead rely on emitting events to the UI for updates on the state (we'll refactor this in a future PR). 

We'd also like to trigger separate commands for actually downloading the results rather than performing it in the same process as the monitor. The reason is because it would delay the polling and isn't really connected to what the monitor is supposed to do.

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
elenatanasoiu: We discussed this and decided:

1. We'd like to trigger separate commands for actually downloading the results rather than performing it in the same process as the monitor. The reason is because it would delay the polling and isn't really connected to what the monitor is supposed to do.

2. In a future PR we'll refactor the monitor to stop returning anything and instead rely on emitting events to the UI. We have a separate ticket for looking into defining the interaction between them.

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
elenatanasoiu: This was a leftover from an initial sketch where I was trying to map the states of the monitor to the states of the variant analysis ('In progress', 'CompletedSuccessfully', 'CompletedUnsuccessfully' etc). Discussed this with Robert and came to the conclusion that we don't actually want to have a one-to-one direct mapping.

This is a long way to say I'm removing the check for `status` and relying on the existence of a `failure_reason` to mark the variant analysis as failed. 

Later we'll flesh this out further by defining when to emit events to the UI based on the state of the variant analysis summary.

elenatanasoiu: Have made the change here: https://github.com/github/vscode-codeql/pull/1545/commits/e41b0ff779438d05e6cca0f179eefdf32e698215.

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      if (variantAnalysisSummary.scanned_repositories) {
+        variantAnalysisSummary.scanned_repositories.forEach(scannedRepo => {
+          if (!scannedReposDownloaded.includes(scannedRepo.repository.id) && scannedRepo.analysis_status === 'succeeded') {
+            scannedReposDownloaded.push(scannedRepo.repository.id);
+          }
+        });
+      }
+
+      attemptCount++;
```
robertbrignull, line 72: Can you add some kind of exit clause for the success case? Currently I think it'll keep polling forever even if the analyses are all successful.

Maybe something like:
```typescript
if (variantAnalysisSummary.status === 'completed') {
  break;
}
```

Overall I think we might need more complex checking of states to handle everything robustly. But we can work on that iteratively and I don't want to hold this PR up unnecessarily. For this PR it's fine so long as the monitor process stops when the analysis is in the "completed" status and doesn't just poll forever (or up to 2 days).

robertbrignull (approved):
I have one more suggestion to make the monitor behave like a good citizen, but otherwise I think we should merge this and more forward. It's a good building base to work from and the code doesn't have to be 100% perfect on every edge case while it's still feature flagged.

robertbrignull:
Btw, I really like the test cases that were added and all the mocking and factories! I didn't comment on any of them in the code as everything looked really good. But those test additions are also really impactful and useful!

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
elenatanasoiu: Yeah I left it quite barebones since there was some setup to get it to this point. Have added the extra exit clause.

robertbrignull (approved):
Thanks for updating!

PR #1548, opened 2022-09-28
Hide repositories with raw results without rows

koesie10:
Currently, when running a query which produces raw results, we will show all repositories, even if they do not have any results. This change will ensure that we are only showing repositories which have results. This matches the behavior for queries which produce interpreted results.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/remote-queries/RemoteQueries.tsx
+++ b/extensions/ql-vscode/src/view/remote-queries/RemoteQueries.tsx
@@ -377,7 +377,7 @@ const AnalysesResults = ({
 
       <ul className="vscode-codeql__flat-list">
         {analysesResults
-          .filter(a => a.interpretedResults.length > 0 || (a.rawResults?.resultSet.rows && a.rawResults.resultSet.rows.length > 0))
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/view/remote-queries/RemoteQueries.tsx
```diff
@@ -377,7 +377,7 @@ const AnalysesResults = ({
 
       <ul className="vscode-codeql__flat-list">
         {analysesResults
-          .filter(a => a.interpretedResults.length > 0 || a.rawResults)
+          .filter(a => a.interpretedResults.length > 0 || (a.rawResults?.resultSet.rows && a.rawResults.resultSet.rows.length > 0))
```
robertbrignull, line 380:
```suggestion
          .filter(a => a.interpretedResults.length || a.rawResults?.resultSet?.rows?.length)
```
If we're willing to use the fact that `0` and `undefined` are both falsy then I think this is equivalent and a bit shorter. I have no preference whether we do this or keep it as it as though. What's there already is perfectly fine.

robertbrignull (approved):
Thanks for digging into this. This change looks good to me.

PR #1549, opened 2022-09-28
Implement skipped repositories tabs

robertbrignull:
Implements the "skipped repository" tabs.
Screenshot 2022-09-28 at 11 15 18
Screenshot 2022-09-28 at 11 15 11

Both the "no access" and "no database" tabs are implemented by one component since the only difference is the text.

I've tried to add stories for new components but I've been having trouble getting them to show up in storybook. Would be interesting to see if anyone else can reproduce this or can see what's going wrong here and what I've not configured correctly.

## Checklist


extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -72,10 +72,13 @@ export interface VariantAnalysisSkippedRepositories {
 
 export interface VariantAnalysisSkippedRepositoryGroup {
   repositoryCount: number,
-  repositories: Array<{
-    id?: number,
-    fullName: string
-  }>
+  repositories: Array<VariantAnalysisSkippedRepository>,
```
koesie10, line 75: I think the `Array` syntax makes sense for more complicated types, but for simple arrays I think it's easier to read it as a simple array:

```suggestion
  repositories: VariantAnalysisSkippedRepository[],
```

See also this TypeScript ESLint rule: https://typescript-eslint.io/rules/array-type/#array-simple

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisSkippedRepositoriesTab.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react';
+import styled from 'styled-components';
+import { VariantAnalysisSkippedRepositoryGroup } from '../../remote-queries/shared/variant-analysis';
+import { Alert } from '../common';
+import { VariantAnalysisSkippedRepositoryRow } from './VariantAnalysisSkippedRepositoryRow';
+
+export type SkippedRepositoriesReason = 'no_access' | 'no_database';
```
koesie10, line 7: An alternative for this pattern would be to pass the alert title and message into the component. This would decouple the alert from the display of the rows and make it easier to re-use this component.

This would change it from:

```tsx
<VariantAnalysisSkippedRepositoriesTab
  reason='no_access'
  skippedRepositoryGroup={notFoundRepos} />
```

to:

```tsx
<VariantAnalysisSkippedRepositoriesTab
  alertTitle="No access"
  alertMessage="The following repositories could not be scanned because you do not have read access."
  skippedRepositoryGroup={notFoundRepos} />
```

That would remove quite a few lines from this component which are now only used to reduce the number of props that's passed into the component.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisSkippedRepositoriesTab.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export const VariantAnalysisSkippedRepositoriesTab = ({
+  reason,
+  skippedRepositoryGroup,
+}: VariantAnalysisSkippedRepositoriesTabProps) => {
+  return (
+    <Container>
+      {getSkipReasonAlert(reason, skippedRepositoryGroup)}
+      {skippedRepositoryGroup.repositories.map((repo) =>
+        <VariantAnalysisSkippedRepositoryRow key={`repo/${repo.fullName}`} repository={repo} />
```
koesie10, line 67: The `key` only needs to be unique within its parent, so just using `repo.fullName` would be enough:

```suggestion
        <VariantAnalysisSkippedRepositoryRow key={repo.fullName} repository={repo} />
```

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisSkippedRepositoryRow.tsx
```diff
@@ -0,0 +1,48 @@
...
+const Row = styled.div`
+  display: flex;
+  flex-direction: row;
+  gap: 0.5em;
+  align-items: center;
+`;
+
+const ChevronIcon = styled.span`
+  color: var(--vscode-descriptionForeground);
+`;
```
koesie10, line 20: Instead of using a `span`, you should be able to use the `Codicon` component:

```suggestion
const ChevronIcon = styled(Codicon)`
  color: var(--vscode-descriptionForeground);
`;
```

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisSkippedRepositoryRow.tsx
```diff
@@ -0,0 +1,48 @@
...
+  return <PrivacyText>{text}</PrivacyText>;
+}
+
+export const VariantAnalysisSkippedRepositoryRow = ({
+  repository,
+}: VariantAnalysisSkippedRepositoryRowProps) => {
+  return (
+    <Row>
+      <VSCodeCheckbox />
+      <ChevronIcon className="codicon codicon-chevron-right" />
```
koesie10, line 41: If we're using the `Codicon` component, this would change to:

```suggestion
      <ChevronIcon name="chevron-right" label="Expand" />
```

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisSkippedRepositoryRow.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const Row = styled.div`
+  display: flex;
+  flex-direction: row;
+  gap: 0.5em;
+  align-items: center;
+`;
+
+const ChevronIcon = styled.span`
+  color: var(--vscode-descriptionForeground);
```
koesie10, line 19: This item always seems to be disabled, so I think we should use `--vscode-disabledForeground` here.

extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisSkippedRepositoriesTab.tsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
```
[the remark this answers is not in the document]
robertbrignull: 🤦  Thanks for spotting that! At least I won't forget this next time 🤞

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
robertbrignull: Thanks. Sounds good to me 👍

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisSkippedRepositoriesTab.tsx
robertbrignull: Thanks that's a really good suggestion. Definitely helps to reduce the complexity of that component and the number of methods taking lots of arguments.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisSkippedRepositoriesTab.tsx
```diff
@@ -0,0 +1,71 @@
...
+) {
+  return (
+    <Alert
+      key='alert'
+      type='warning'
+      title={getSkipReasonAlertTitle(reason)}
+      message={getSkipReasonAlertMessage(reason, repos)}
+    />
+  );
+}
```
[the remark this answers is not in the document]
robertbrignull: I agree with you on the original state of the code, but in the transformation to the props, this function has now gained a tiny bit of complexity. It has one variable for the extra text when repositories are omitted which would be good to keep contained to this function. So I think I'll keep this function as it is. Let me know if you still feel it should be inlined.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisSkippedRepositoriesTab.tsx
robertbrignull: My reasoning here was only to make super sure it doesn't conflict with the key for the alert, which is `alert`. Should be fine since a repo name contains a slash, but I didn't want to rely on that or risk conflicts if another component is added to this container. So unless there's more of a benefit to simplifying it I'd prefer to keep it as it is.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisSkippedRepositoryRow.tsx
robertbrignull: I'm not really sure what to put as the label, because it doesn't actually expand. As far as I can tell from the designs, the chevron is there to make it look like the other pages, but there's no case where it could expand and show more for this view. It can certainly be "expand" for now though. Would also be good for this to match whatever it used in the scanned repos tab.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisSkippedRepositoriesTab.tsx
koesie10: I'm happy to leave it as-is, but I don't think you need the `key` on the alert at all since it's not part of the list. React only uses the keys for change detection in a `map` call/list. Even when the component is a sibling of a `map` call, it shouldn't be necessary to add the `key`.

koesie10 (approved):

PR #1550, opened 2022-09-29
Restore variant analysis view on restart of VSCode

koesie10:
This implements persistence for the variant analysis webview, allowing the webview panel to be restored when VSCode is restarted. It's probably easier to add this now than to try to add it later.

The basic idea is that there are no real differences when opening the webview for the first time. However, when VSCode is restarted it will use the `VariantAnalysisViewSerializer` to restore the webview panel. In our case this means recreating the `VariantAnalysisView`.

To fully test this, I've also added a mock variant analysis ID as the state of the webview. This value is now randomly generated when calling the `codeQL.mockVariantAnalysisView` command. This allows us to test opening multiple webviews and that the webviews are restored with the correct state.

See: https://code.visualstudio.com/api/extension-guides/webview#persistence

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts
@@ -1,20 +1,38 @@
-import { ViewColumn } from 'vscode';
+import { ExtensionContext, ViewColumn } from 'vscode';
 import { AbstractWebview, WebviewPanelConfig } from '../abstract-webview';
-import { WebviewMessage } from '../interface-utils';
 import { logger } from '../logging';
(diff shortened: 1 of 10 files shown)


extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts
```diff
@@ -1,28 +1,212 @@
...
+
+        void logger.log('Variant analysis view loaded');
+
+        await this.postMessage({
+          t: 'setVariantAnalysis',
+          variantAnalysis: this.getVariantAnalysis(),
+        });
+
+        break;
+      case 'stopVariantAnalysis':
```
koesie10, line 56: This is not yet used but is required for TypeScript, otherwise it will complain about the `assertNever` branch (since its a union of only 1 type, so equal to the 1 type)

Issue #1556, opened 2022-09-29
Newly opened results view is not opening to the side


aeisenberg:
**Describe the bug**
When a (non-remote) results view is opened, it is supposed to open beside the current editor. However, it is opening in the same editor column as the current editor. This seems to be a regression since this behaviour is different from the past.

PR #1557, opened 2022-09-29
Ensure results view is opened in column beside 

aeisenberg:
The results view will always open next to the current editor.

This restores previous behaviour of the results view. Question: Do we want our other views to have the same behaviour? 

Previously, our users specifically asked for the results view to open in a way that would not block the currently active editor.

Fixes https://github.com/github/vscode-codeql/issues/1556

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/abstract-webview.ts
+++ b/extensions/ql-vscode/src/abstract-webview.ts
@@ -46,7 +46,7 @@ export abstract class AbstractWebview<ToMessage extends WebviewMessage, FromMess
       this.panel = Window.createWebviewPanel(
         config.viewId,
         config.title,
-        { viewColumn: ViewColumn.Active, preserveFocus: true },
+        { viewColumn: config.viewColumn, preserveFocus: config.preserveFocus },
         {
           enableScripts: true,
           enableFindWidget: true,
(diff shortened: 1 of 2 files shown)


aeisenberg:
@adityasharad, this should address the issue you were having earlier with the results view not activating.

angelapwen:
I tested this out and I'm getting the following error on loading the extension:

Screen Shot 2022-09-29 at 1 15 41 PM

I also no longer see the eval log viewer. 

I see the viewer and error on `main`, though, so perhaps it's to do with forcing this view while it doesn't have anything to load into it yet?

aeisenberg:
Hmm...I can't reproduce that error. Since you're seeing it on main, it's unlikely to be related to this PR. 

Can you try doing a clean build of the extension and then launch again?

angelapwen:
> Hmm...I can't reproduce that error. Since you're seeing it on main, it's unlikely to be related to this PR.
> 
> Can you try doing a clean build of the extension and then launch again?

Sorry, what I meant was that I wasn't seeing it on main. I'll do the clean build and check again.

angelapwen:
No longer seeing the error on a clean build 👍

aeisenberg:
As I'm thinking more, most likely this behaviour is not desirable for remote queries since they are run in a different context. Local queries are usually faster and more iterative. The results view for local queries will open automatically after the query completes. This can be annoying if it blocks the tab that you're working on. All other results views appear only when explicitly clicked, so blocking the previously active editor is not a big deal (might actually be desirable).

If you're ok with this, can I get a ✅ ?

angelapwen (approved):
Reasoning makes sense, 🚢

aeisenberg:
Hmmm...no changelog for this PR. I will add one.

PR #1558, opened 2022-09-30
Remove mentions of lgtm.com from README

jf205:
LGTM.com is due to be deprecated (https://github.com/orgs/community/discussions/29534) so this PR updates the readme for the new 'Download database from GitHub' feature.

See also https://github.com/github/vscode-codeql-starter/pull/132.

### DIFF
--- a/extensions/ql-vscode/README.md
+++ b/extensions/ql-vscode/README.md
@@ -22,7 +22,7 @@ For information about other configurations, see the separate [CodeQL help](https
 
 ### Quick start: Using CodeQL
 
-1. [Import a database from LGTM](#importing-a-database-from-lgtm).
+1. [Import a database from GitHub](#importing-a-database-from-github).
 1. [Run a query](#running-a-query).
 
 ---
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/README.md
```diff
@@ -73,18 +73,19 @@ If you're using your own clone of the CodeQL standard libraries, you can do a `g
...
-1. Copy the link to that project, for example `https://lgtm.com/projects/g/apache/kafka`.
-1. In VS Code, open the Command Palette and choose the **CodeQL: Download Database from LGTM** command.
+1. Find a project that you're interested in on GitHub.com, for example [Apache Kafka](https://github.com/apache/kafka).
+1. Copy the link to that project, for example `https://github.com/apache/kafka`.
+1. In VS Code, open the Command Palette and choose the **CodeQL: Download Database from GitHub** command.
 1. Paste the link you copied earlier.
 1. Select the language for the database you want to download (only required if the project has databases for multiple languages).
 1. Once the CodeQL database has been imported, it is displayed in the Databases view.
 
+For more information, see [Choosing a database](https://codeql.github.com/docs/codeql-for-visual-studio-code/analyzing-your-projects/#choosing-a-database) on codeql.github.com.
```
aeisenberg, line 87: Do you think there should be some guidance if the repository doesn't have a database? Also, is there an easy way for users to determine which repos already has one?

aeisenberg: Nice. In the future, we can make this better. Being able to search for repos with dbs would be great. I've already spent lots of time looking for a suitable db to download. If I could issue a search, and find repos with dbs that match certain features, it would be helpful.

aeisenberg (approved):
I think this is good as-is. CodeQL docs can be updated separately.

PR #1559, opened 2022-09-30
Download variant analysis results

elenatanasoiu:

## What
Follow-up from [this PR](https://github.com/github/vscode-codeql/pull/1545) where we introduced a monitor to check on the status of the variant analysis. 

We'd like to auto-download results from the variant analysis as they become ready. 

## Why
Based on a previous [discussion](https://github.com/github/vscode-codeql/pull/1545#discussion_r982591809) while implementing the variant analysis monitor, we decided that we'd like to trigger separate commands for actually downloading the results rather than performing it in the same process as the monitor.

The reason for this is that it would delay the polling and isn't really connected to what the monitor is supposed to do.

## Future
In a separate PR, we'll also change the monitor and manager in order to emit events to the UI. These will be the various states of a download.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
@@ -1,8 +1,18 @@
-import { CancellationToken, commands, ExtensionContext } from 'vscode';
+import * as ghApiClient from './gh-api/gh-api-client';
+import * as path from 'path';
+import * as fs from 'fs-extra';
+import { CancellationToken, ExtensionContext } from 'vscode';
 import { DisposableObject } from '../pure/disposable-object';
 import { Logger } from '../logging';
+import { Credentials } from '../authentication';
 import { VariantAnalysisMonitor } from './variant-analysis-monitor';
+import {
+  VariantAnalysis as VariantAnalysisApiResponse,
+  VariantAnalysisRepoTask,
+  VariantAnalysisScannedRepository as ApiVariantAnalysisScannedRepository
+} from './gh-api/variant-analysis';
 import { VariantAnalysis } from './shared/variant-analysis';
+import { getErrorMessage } from '../pure/helpers-pure';
 
 export class VariantAnalysisManager extends DisposableObject {
   private readonly variantAnalysisMonitor: VariantAnalysisMonitor;
@@ -20,5 +30,52 @@ export class VariantAnalysisManager extends DisposableObject {
     cancellationToken: CancellationToken
   ): Promise<void> {
     await this.variantAnalysisMonitor.monitorVariantAnalysis(variantAnalysis, cancellationToken);
+  }
+
+  public async autoDownloadVariantAnalysisResult(
+    scannedRepo: ApiVariantAnalysisScannedRepository,
+    variantAnalysisSummary: VariantAnalysisApiResponse,
+    cancellationToken: CancellationToken
+  ): Promise<void> {
+
+    const credentials = await Credentials.initialize(this.ctx);
+    if (!credentials) { throw Error('Error authenticating with GitHub'); }
+
+    if (cancellationToken && cancellationToken.isCancellationRequested) {
+      return;
+    }
+
+    let repoTask: VariantAnalysisRepoTask;
+    try {
+      repoTask = await ghApiClient.getVariantAnalysisRepo(
+        credentials,
+        variantAnalysisSummary.controller_repo.id,
+        variantAnalysisSummary.id,
+        scannedRepo.repository.id
+      );
+    }
+    catch (e) { throw new Error(`Could not download the results for variant analysis with id: ${variantAnalysisSummary.id}. Error: ${getErrorMessage(e)}`); }
+
+    if (repoTask.artifact_url) {
+      const resultDirectory = path.join(
+        this.ctx.globalStorageUri.fsPath,
+        'variant-analyses',
+        `${variantAnalysisSummary.id}`,
+        scannedRepo.repository.full_name
+      );
+
+      const storagePath = path.join(
+        resultDirectory,
+        scannedRepo.repository.full_name
+      );
+
+      const result = await ghApiClient.getVariantAnalysisRepoResult(
+        credentials,
+        repoTask.artifact_url
+      );
+
+      fs.mkdirSync(resultDirectory, { recursive: true });
+      await fs.writeFile(storagePath, JSON.stringify(result, null, 2), 'utf8');
+    }
   }
 }
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
@@ -1,0 +1,24 @@
+import { CancellationToken, commands, ExtensionContext } from 'vscode';
+import { DisposableObject } from '../pure/disposable-object';
+import { Logger } from '../logging';
+import { VariantAnalysisMonitor } from './variant-analysis-monitor';
+import { VariantAnalysis } from './shared/variant-analysis';
+
+export class VariantAnalysisManager extends DisposableObject {
+  private readonly variantAnalysisMonitor: VariantAnalysisMonitor;
+
+  constructor(
+    private readonly ctx: ExtensionContext,
+    logger: Logger,
+  ) {
+    super();
+    this.variantAnalysisMonitor = new VariantAnalysisMonitor(ctx, logger);
+  }
+
+  public async monitorVariantAnalysis(
+    variantAnalysis: VariantAnalysis,
+    cancellationToken: CancellationToken
+  ): Promise<void> {
+    await this.variantAnalysisMonitor.monitorVariantAnalysis(variantAnalysis, cancellationToken);
+  }
+}
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
@@ -1,4 +1,4 @@
-import { ExtensionContext, CancellationToken } from 'vscode';
+import { ExtensionContext, CancellationToken, commands } from 'vscode';
 import { Credentials } from '../authentication';
 import { Logger } from '../logging';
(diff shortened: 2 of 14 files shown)


extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-monitor.test.ts
```diff
@@ -93,13 +91,12 @@ describe('Variant Analysis Monitor', async function() {
...
         const result = await variantAnalysisMonitor.monitorVariantAnalysis(variantAnalysis, cancellationToken);
-        variantAnalysis = result.variantAnalysis;
 
         expect(mockGetVariantAnalysis.calledOnce).to.be.true;
         expect(result.status).to.eql('Failed');
         expect(result.error).to.eql(`Variant Analysis has failed: ${mockFailedApiResponse.failure_reason}`);
-        expect(variantAnalysis.status).to.equal(VariantAnalysisStatus.Failed);
-        expect(variantAnalysis.failureReason).to.equal(processFailureReason(mockFailedApiResponse.failure_reason as VariantAnalysisFailureReason));
+        expect(result?.variantAnalysis?.status).to.equal(VariantAnalysisStatus.Failed);
+        expect(result?.variantAnalysis?.failureReason).to.equal(processFailureReason(mockFailedApiResponse.failure_reason as VariantAnalysisFailureReason));
```
koesie10, line 99: I don't think the `result` can be nullable here, so I think it would be clearer to just add the safe navigation operator only to the `result.variantAnalysis`

extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/skipped-repositories.ts
```diff
@@ -0,0 +1,30 @@
...
+import {
+  VariantAnalysisSkippedRepositories,
+  VariantAnalysisSkippedRepositoryGroup
+} from '../../../../remote-queries/shared/variant-analysis';
+
+export function createMockSkippedRepos(): VariantAnalysisSkippedRepositories {
+  return {
+    accessMismatchRepos: createMockSkippedRepoGroup(),
+    noCodeqlDbRepos: createMockSkippedRepoGroup(),
+    notFoundRepos: createMockSkippedRepoGroup(),
```
koesie10, line 11: Should we ensure that this matches the actual data we receive and make the `notFoundRepos` mock separate/ The `notFoundRepos` does not have an `id` or `private`, while all the other groups will always have both.

extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/variant-analysis.ts
```diff
@@ -0,0 +1,34 @@
...
+import { createMockScannedRepos } from './scanned-repositories';
+import { createMockSkippedRepos } from './skipped-repositories';
+
+export function createMockVariantAnalysis(
+  status: VariantAnalysisStatus = VariantAnalysisStatus.InProgress,
+  scannedRepos: VariantAnalysisScannedRepository[] = createMockScannedRepos(),
+  skippedRepos: VariantAnalysisSkippedRepositories = createMockSkippedRepos()
+): VariantAnalysis {
+  const variantAnalysis: VariantAnalysis = {
+    id: 123,
```
koesie10, line 17: Would it make sense to also use `faker.datatype.number()` here (and for the other IDs) like in the repository mock?

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -21,4 +31,51 @@ export class VariantAnalysisManager extends DisposableObject {
...
+        scannedRepo.repository.full_name
+      );
+
+      const result = await ghApiClient.getVariantAnalysisRepoResult(
+        credentials,
+        repoTask.artifact_url
+      );
+
+      fs.mkdirSync(resultDirectory, { recursive: true });
+      await fs.writeFile(storagePath, JSON.stringify(result, null, 2), 'utf8');
```
koesie10, line 78: What will this actually write to disk? [From the action](https://github.com/github/codeql-variant-analysis-action/blob/0e889fce2f0d6fff82a150d83d36efa12d6b781a/src/query.ts#L170-L187), it seems like we are uploading a ZIP file, so would this write out a binary string encoded as JSON?

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
```diff
@@ -0,0 +1,156 @@
...
+  beforeEach(async () => {
+    sandbox = sinon.createSandbox();
+    sandbox.stub(logger, 'log');
+    sandbox.stub(config, 'isVariantAnalysisLiveResultsEnabled').returns(false);
+    sandbox.stub(fs, 'mkdirSync');
+    sandbox.stub(fs, 'writeFile');
+
+    cancellationToken = {
+      isCancellationRequested: false
+    } as unknown as CancellationToken;
```
koesie10, line 38: Since this is running in VSCode, could we use a real cancellation token instead using the [`CancellationTokenSource`](https://code.visualstudio.com/api/references/vscode-api#CancellationTokenSource)?

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
```diff
@@ -64,6 +64,7 @@ export class VariantAnalysisMonitor {
       if (variantAnalysisSummary.scanned_repositories) {
         variantAnalysisSummary.scanned_repositories.forEach(scannedRepo => {
           if (!scannedReposDownloaded.includes(scannedRepo.repository.id) && scannedRepo.analysis_status === 'succeeded') {
+            void commands.executeCommand('codeQL.autoDownloadVariantAnalysisResult', scannedRepo);
```
koesie10, line 67: This only seems to pass the `scannedRepo` to the command. Should this also be passing the `variantAnalysisSummary` (since that is one of the arguments of this command), or will that automatically be included?

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-monitor.test.ts
elenatanasoiu: You're right. I've addressed this [here](https://github.com/github/vscode-codeql/pull/1559/commits/644a83d6d8fe9e23ddb982b84fdcec5117440a78).

extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/skipped-repositories.ts
elenatanasoiu: Ah, I wonder why this wasn't caught by our type system. I've added a separate method to create this: https://github.com/github/vscode-codeql/pull/1559/commits/7dc5eebcc140ae4be86412cb32464cdfcf14c616

extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/variant-analysis.ts
elenatanasoiu: Makes sense to me. Have added it [here](https://github.com/github/vscode-codeql/pull/1559/commits/f56f017a8401b56d4a45971f7ad03a52d4ac5472).

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
elenatanasoiu: Thanks! I actually wasn't sure where to check what type of file the result was. 

I don't want this PR to hold up more work and it's going to take a bit more effort to rejig it to handle zip files, so I've written up a separate issue for this.

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
elenatanasoiu: Sure! It does seem a bit shady to use `unknown`. Have added the change [here](https://github.com/github/vscode-codeql/pull/1559/commits/b82cd8b6f4a5875ba46e2165e965c37a83bfa50e).

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
elenatanasoiu: That is totally a mistake. Thanks for catching it! We unfortunately don't write tests for these commands at the moment and I admit I am totally reliant on some test screaming at me about the right amount of params. Thanks for catching it!

extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/skipped-repositories.ts
koesie10: It seems like we don't have a different type for the not found repos in the `shared` `VariantAnalysis`, while we do have a separate type in the `gh-api` `VariantAnalysis`. It might make sense to create a separate type for it, but that's independent of this PR.

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
koesie10: Thanks! Does it make sense to not mock the cancellation token at all? Instead of mocking the cancellation token source, can we just call `new CancellationTokenSource()`?

koesie10 (approved):
There's still one outstanding comment about creating a new `CancellationTokenSource` instead of mocking it, but that can also be done in a follow-up PR.

elenatanasoiu:
Thanks @koesie10! I've created an issue for that as well.

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
elenatanasoiu: Ah I see what you mean! I'll take a look at this in a separate issue.

PR #1566, opened 2022-10-04
Emit `variantAnalysisAdded` event

elenatanasoiu:
When we first submit the variant analysis for processing, we'd like to update
the query history panel.

At the moment we're just adding the setup for triggering the event. In a future
PR we'll consume this event and change the query history panel accordingly.

In order for this second part to happen we will need to introduce a new 
`VariantAnalysisHistoryItem` type which will massage the data we get from the 
API into a type which the QueryHistory panel can consume.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -467,15 +467,15 @@ async function activateWithInstalledDistribution(
   const localQueryResultsView = new ResultsView(ctx, dbm, cliServer, queryServerLogger, labelProvider);
   ctx.subscriptions.push(localQueryResultsView);
 
-  void logger.log('Initializing remote queries manager.');
-  const rqm = new RemoteQueriesManager(ctx, cliServer, queryStorageDir, logger);
-  ctx.subscriptions.push(rqm);
-
   void logger.log('Initializing variant analysis manager.');
   const variantAnalysisStorageDir = path.join(ctx.globalStorageUri.fsPath, 'variant-analyses');
   await fs.ensureDir(variantAnalysisStorageDir);
   const variantAnalysisManager = new VariantAnalysisManager(ctx, cliServer, variantAnalysisStorageDir, logger);
   ctx.subscriptions.push(variantAnalysisManager);
+
+  void logger.log('Initializing remote queries manager.');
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/extension.ts
```diff
@@ -906,6 +906,7 @@ async function activateWithInstalledDistribution(
       variantAnalysis: VariantAnalysis,
       token: CancellationToken
     ) => {
+      variantAnalysisManager.variantAnalysisAddedEventEmitter.fire({ variantAnalysis });
```
koesie10, line 909: I'm not sure whether it makes sense to emit this event in the command, rather than in a method of the variant analysis manager. I would expect this command to also be called when VSCode is restarted and that should probably not emit this event (the variant analysis should have been added to the query history already at that point).

I think it would be clearer to fire this event from within `run-remote-query.ts` by calling a method on the variant analysis manager.

elenatanasoiu: Thanks for the advice, I hadn't taken into account what happens when VSCode is restarted. Have moved the event firing into `run-remote-query.ts` as you suggested.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -273,6 +275,9 @@ export async function runRemoteQuery(
 
       const processedVariantAnalysis = processVariantAnalysis(variantAnalysisSubmission, variantAnalysisResponse);
 
+      const variantAnalysisManager = new VariantAnalysisManager(ctx, logger);
```
koesie10, line 278: I think we should try to re-use the variant analysis manager that is created when the extension is created. This would fire the event, but no listeners can actually be attached to it. Passing in the `variantAnalysisManager` instead of the `ctx` should be enough.

Right now that would involve passing the `VariantAnalysisManager` to the `RemoteQueriesManager`, but I think it's fine for now. We should refactor/remove the `RemoteQueriesManager` in the future to remove this dependency anyway.

elenatanasoiu: Thanks @koesie10 ! I've made the change for this. 🙇‍♀️

koesie10 (approved):
LGTM. There's one warning given by CodeQL, but I think this will also be resolved if you merge in `main`.

PR #1568, opened 2022-10-05
Add commands for navigation of alerts

asgerf:
[Some time ago](https://github.com/github/vscode-codeql/pull/175) I added commands for navigating the steps on a path, but didn't add commands for navigating between alerts - this PR does that.

This PR adds the commands:
- `CodeQL: Navigate Up in Local Result Viewer`
- `CodeQL: Navigate Down in Local Result Viewer`
- `CodeQL: Navigate Left in Local Result Viewer`
- `CodeQL: Navigate Right in Local Result Viewer`

The intent is that you can bind keyboard shortcuts to these commands to navigate alerts with the keyboard.

When viewing 'raw' results, such as from quick-eval, the alert-navigation commands move between rows and columns.

When viewing alerts, the left/right commands instead move into or out of nesting levels of tree view:
- Alerts
  - Paths
    - Path steps

There are no commands for clicking the individual links in an interpreted alert message, but you can sort of achieve this by viewing the `#select` result set as a raw result table.

## Checklist


extensions/ql-vscode/src/interface.ts
```diff
@@ -154,6 +154,18 @@ export class ResultsView extends AbstractWebview<IntoResultsViewMsg, FromResults
         this.navigatePathStep.bind(this, -1)
       )
     );
+    this.push(
+      commandRunner(
+        'codeQLQueryResults.nextAlert',
+        this.navigateAlert.bind(this, 1)
```
aeisenberg, line 160: Instead of making this a number, can you make this an enum:

```
enum NavDirection {
   Next = 1,
   Prev = -1
}
```

And you can change the messages above as well. You can put this in `interface-types.ts` and have it shared with the results view.

extensions/ql-vscode/src/pure/result-keys.ts
```diff
@@ -64,7 +79,7 @@ export function getPathNode(sarif: sarif.Log, key: PathNode): sarif.Location | u
 /**
  * Returns true if the two keys are both `undefined` or contain the same set of indices.
  */
-export function equals(key1: PathNode | undefined, key2: PathNode | undefined): boolean {
+export function equals(key1: Partial<PathNode> | undefined, key2: Partial<PathNode> | undefined): boolean {
```
aeisenberg, line 82: I don't think this function is actually used.

aeisenberg: Same with `getPathNode` and `getPath` above.

extensions/ql-vscode/src/pure/result-keys.ts
```diff
@@ -1,33 +1,48 @@
...
+export type ResultKey = Result | Path | PathNode;
+
 /** Alias for `undefined` but more readable in some cases */
 export const none: PathNode | undefined = undefined;
 
 /**
  * Looks up a specific result in a result set.
  */
-export function getResult(sarif: sarif.Log, key: Result): sarif.Result | undefined {
+export function getResult(sarif: sarif.Log, key: Result | Path | PathNode): sarif.Result | undefined {
```
aeisenberg, line 45: Unused. You can delete this function.

extensions/ql-vscode/src/pure/result-keys.ts
```diff
@@ -37,7 +52,7 @@ export function getResult(sarif: sarif.Log, key: Result): sarif.Result | undefin
 /**
  * Looks up a specific path in a result set.
  */
-export function getPath(sarif: sarif.Log, key: Path): sarif.ThreadFlow | undefined {
+export function getPath(sarif: sarif.Log, key: Path | PathNode): sarif.ThreadFlow | undefined {
```
aeisenberg, line 55: Unused.

extensions/ql-vscode/src/view/results/raw-results-table.tsx
```diff
@@ -58,4 +76,53 @@ export class RawTable extends React.Component<RawTableProps, Record<string, neve
...
+      case 'navigateAlert': {
+        this.setState(prevState => {
+          const numberOfAlerts = this.props.resultSet.rows.length;
+          if (numberOfAlerts === 0) {
+            return prevState;
+          }
+          const currentRow = prevState.selectedItem?.row;
+          const nextRow = currentRow === undefined
+            ? 0
+            : (currentRow + event.direction);
```
aeisenberg, line 91: The direction _happens_ to be a number that tells you how far to move in a given direction. I'd rather this be explicit. Either use an enum for direction (as I suggest above) and convert that into 1  / -1, or explicitly use the term `distance` and make it clear that you can accept a number.

aeisenberg (commented):
This is great! Thanks for creating this PR. I think there are several functions in `result-keys.ts` that can be removed. And I have not tried this myself, but I will soon.

extensions/ql-vscode/src/pure/result-keys.ts
asgerf: `getResult` is called from `getPath`, which is used.

extensions/ql-vscode/src/pure/result-keys.ts
asgerf: The type is needed from both call sites to `getPath` - this is due to the change in the `Path` type.

asgerf:
Thanks for the quick review @aeisenberg! I wasn't expecting reviews while the PR is in draft state, as I'd like to get some more feedback on the UI, but I'm happy to iterate on it if you're OK with the scope of the PR changing along the way.

In particular, I'm not 100% convinced these commands are actually right. I feels a little off to have two different set of key bindings for moving up/down in different parts of the alert view.

It's possible we want to change the commands to be more general "move left/right/up/down". Like the arrow keys in a common tree-view, with the caveat that moving up/down shouldn't leave the current level (moving down past the last path step should not skip over to the next path - that would be annoying).

RasmusWL:
> When viewing 'raw' results, the alert-navigation commands move between rows (raw results cannot have paths, so the path step commands do nothing). This view is a table with an arbitrary number of columns, each of which could be a link to a source location. In principle there should be a way to navigate sideways here, but for now it just remembers which column you clicked last time, and will keep showing you that column when switching rows via the command.

Nice :muscle: 

> moving down past the last path step should not skip over to the next path - that would be annoying

When thinking about this, I feel like it would work OK, but without having tried it I guess it's hard to really say. Is your feeling about annoyance based on having tried it and it being bad, or from your intuitive feeling?

asgerf:
I've pushed another commit that changes the commands to left/right/up/down behaviour. On my machine I bound the keys to cmd+shift+{I,J,K,L} in a WASD-like pattern. Personally I like this a bit better, but I'd be very interested in feedback about how this works for others.

The internal command ID for up/down is `previousPathStep`/`nextPathStep` so as to not break existing keybindings. I couldn't find a way to deprecate commands in vscode/

extensions/ql-vscode/src/pure/result-keys.ts
asgerf: After some updates, `getPathNode` is actually used now, but I removed the other unused parts.

extensions/ql-vscode/src/interface.ts
asgerf: There is `enum NavigationDirection` now

extensions/ql-vscode/src/view/results/raw-results-table.tsx
asgerf: After the change to up/down/left/right navigation this is more explicit

asgerf:
After using this for a few days I noticed some quirky behaviour when stepping down while a node was expanded. I pushed another commit to collapse the previous node when stepping up or down.

Perhaps it's best if I keep using it for a few more days before moving ahead.

RasmusWL:
I've tried out the navigation in this PR, and overall I'm a big fan. Especially for quick evaling predicates, being able to move left/right for columns is amazing!

However, for `path-problem`s I'm not personally 100% on board with the required left/right movement. I was always pressing DOWN when on the path level (I only looked at alerts with a single path), and then got annoyed when it didn't work :smile:

Suggestion: When you expand an alert, could you be taken directly to the first path-step of the first path? -- at least if there is only one path.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Fix a bug where it was not possible to add a database folder if the folder name starts with `db-`. [#1565](https://github.com/github/vscode-codeql/pull/1565)
+- Added commands for navigating up, down, left, or right in the result viewer. Previously there were only commands for moving up and down the currently-selected path. We suggest binding keyboard shortcuts to these commands, for navigating the result viewer using the keyboard.
```
aeisenberg, line 6:
```suggestion
- Add commands for navigating up, down, left, or right in the result viewer. Previously there were only commands for moving up and down the currently-selected path. We suggest binding keyboard shortcuts to these commands, for navigating the result viewer using the keyboard. [#1568](https://github.com/github/vscode-codeql/pull/1568)
```

extensions/ql-vscode/src/pure/result-keys.ts
```diff
@@ -1,33 +1,45 @@
...
-/** Alias for `undefined` but more readable in some cases */
-export const none: PathNode | undefined = undefined;
+export type ResultKey = Result | Path | PathNode;
 
 /**
  * Looks up a specific result in a result set.
  */
-export function getResult(sarif: sarif.Log, key: Result): sarif.Result | undefined {
+export function getResult(sarif: sarif.Log, key: Result | Path | PathNode): sarif.Result | undefined {
   if (sarif.runs.length === 0) return undefined;
```
aeisenberg, line 43: Minor: The body of this function can be replaced with this:

```
  return sarif.runs?.[0]?.results?.[key.resultIndex];
```

It's really outside the scope of this PR, but it's a minor readability improvement if you want to add it.

extensions/ql-vscode/src/pure/result-keys.ts
aeisenberg: Sorry...missed that the first time.

extensions/ql-vscode/src/view/results/alert-table.tsx
```diff
@@ -14,37 +14,40 @@ import {
...
   constructor(props: PathTableProps) {
     super(props);
-    this.state = { expanded: {}, selectedPathNode: undefined };
+    this.state = { expanded: new Set<string>(), selectedItem: undefined };
     this.handleNavigationEvent = this.handleNavigationEvent.bind(this);
   }
 
   /**
-   * Given a list of `indices`, toggle the first, and if we 'open' the
+   * Given a list of `ids`, toggle the first, and if we 'open' the
```
aeisenberg, line 35: Minor:

```suggestion
   * Given a list of `keys`, toggle the first, and if we 'open' the
```

extensions/ql-vscode/src/view/results/alert-table.tsx
```diff
@@ -302,34 +306,103 @@ export class PathTable extends React.Component<PathTableProps, PathTableState> {
...
+          return prevState; // Path does not exist
+        }
+        jumpLocation = undefined; // When selecting a 'path', don't jump anywhere.
+      } else {
+        jumpLocation = Keys.getResult(data, key)?.locations?.[0];
+        if (jumpLocation == null) {
+          return prevState; // Path step does not exist.
+        }
+      }
+      if (jumpLocation !== undefined) {
```
aeisenberg, line 332: Is there a semantic difference between null and undefined here? I see you are using `== null` and `!= null` above, but `!== undefined` here. I personally prefer just using `if (jumpLocation) ...` or `if (!jumpLocation) ...`, which I think is clearer than using obscure JS semantics that `undefined == null`. 

This is just a preference of mine, so it's not a requirement to move forward with this PR.

aeisenberg (commented):
That's really neat. It works great for me.  I have a handful of small comments. 

If we are not going to bind the new commands to key shortcuts by default, then we should update the README with some instructions on how to do it manually. Perhaps it would fit after the `### Running a query` section.

extensions/ql-vscode/src/view/results/alert-table.tsx
asgerf: I've updated all the checks to consistently check for `undefined` only.

This seems to be the convention here, as there are >200 occurrences of `=== undefined` or `!== undefined` and only 16 of `== null` and `!= null`.

extensions/ql-vscode/package.json
```diff
@@ -595,19 +595,19 @@
...
       },
       {
         "command": "codeQLQueryResults.right",
-        "title": "CodeQL: Navigate right in result viewer"
+        "title": "CodeQL: Navigate Right in Result Viewer"
       },
       {
         "command": "codeQLQueryResults.left",
-        "title": "CodeQL: Navigate left in result viewer"
+        "title": "CodeQL: Navigate Left in Result Viewer"
```
aeisenberg, line 610: We also have the MRVA result  viewer, which this change doesn't support.

```suggestion
        "title": "CodeQL: Navigate Down in Local Result Viewer"
      },
      {
        "command": "codeQLQueryResults.up",
        "title": "CodeQL: Navigate Up in Local Result Viewer"
      },
      {
        "command": "codeQLQueryResults.right",
        "title": "CodeQL: Navigate Right in Local Result Viewer"
      },
      {
        "command": "codeQLQueryResults.left",
        "title": "CodeQL: Navigate Left in Local Result Viewer"
```

aeisenberg (commented):
Minor change in command name, and then it looks good.

asgerf:
Thanks for the feedback @RasmusWL 

> However, for path-problems I'm not personally 100% on board with the required left/right movement. I was always pressing DOWN when on the path level (I only looked at alerts with a single path), and then got annoyed when it didn't work 😄

Yeah the same thing happened to me, and I also tried various ways of skipping over the 'path' layer. But it never felt right. It generates that uncanny feeling where the UI tries to be really smart, but then becomes less predictable and less consistent with other UI components. My take is that the current version is something you can get used to, and then it stops being annoying, whereas the other one will always feel a little off.

asgerf:
I updated the PR description to match what the current behavior.

RasmusWL:
> > However, for path-problems I'm not personally 100% on board with the required left/right movement. I was always pressing DOWN when on the path level (I only looked at alerts with a single path), and then got annoyed when it didn't work smile
> 
> Yeah the same thing happened to me, and I also tried various ways of skipping over the 'path' layer. But it never felt right. It generates that uncanny feeling where the UI tries to be really smart, but then becomes less predictable and less consistent with other UI components. My take is that the current version is something you can get used to, and then it stops being annoying, whereas the other one will always feel a little off.

I want this feature more than I want to discuss it, so although I'm not 100% convinced I will learn this, I say we ship it :ship:

aeisenberg (approved):
I'm feeling good about this. I think this will be a nice change. I still think we can add some more documentation for this. We can either expand the readme, or put it into the codeql extension docs on the codeql microsite. 

But, that could be done later. Nice work.

Feel free to merge when you're ready.

asgerf:
Thanks @aeisenberg! It seems I'm not authorized to merge in this repo. Can you trigger the merge?

PR #1570, opened 2022-10-06
Add variant analysis results manager

koesie10:
This adds a new variant analysis results manager which is responsible for downloading and loading variant analysis results to/from the filesystem. It is essentially the `AnalysesResultsManager` modified to suit the variant analysis results.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -457,9 +457,15 @@ async function activateWithInstalledDistribution(
   const localQueryResultsView = new ResultsView(ctx, dbm, cliServer, queryServerLogger, labelProvider);
   ctx.subscriptions.push(localQueryResultsView);
 
-  void logger.log('Initializing variant analysis manager.');
+  void logger.log('Initializing remote queries manager.');
   const rqm = new RemoteQueriesManager(ctx, cliServer, queryStorageDir, logger);
   ctx.subscriptions.push(rqm);
+
+  void logger.log('Initializing variant analysis manager.');
+  const variantAnalysisStorageDir = path.join(ctx.globalStorageUri.fsPath, 'variant-analyses');
+  await fs.ensureDir(variantAnalysisStorageDir);
+  const variantAnalysisManager = new VariantAnalysisManager(ctx, cliServer, variantAnalysisStorageDir, logger);
+  ctx.subscriptions.push(variantAnalysisManager);
 
   void logger.log('Initializing query history.');
   const qhm = new QueryHistoryManager(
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/extension.ts
```diff
@@ -457,10 +457,16 @@ async function activateWithInstalledDistribution(
...
   ctx.subscriptions.push(localQueryResultsView);
 
-  void logger.log('Initializing variant analysis manager.');
+  void logger.log('Initializing remote queries manager.');
   const rqm = new RemoteQueriesManager(ctx, cliServer, queryStorageDir, logger);
   ctx.subscriptions.push(rqm);
 
+  void logger.log('Initializing variant analysis manager.');
+  const variantAnalysisStorageDir = path.join(ctx.globalStorageUri.fsPath, 'variant-analyses');
+  await fs.ensureDir(variantAnalysisStorageDir);
```
elenatanasoiu, line 466: Since we're building the storage path off of `ctx`, would we be able to move this logic into a method on the `VariantAnalysisManager` instead of passing in the `storagePath` as a param? 

I realize this `extension.ts` file already does this sort of setup for other commands but it's not very clean or readable. We might also accidentally overwrite some variable here and the extension will just happily go with that.

extensions/ql-vscode/src/remote-queries/variant-analysis-results-manager.ts
```diff
@@ -0,0 +1,178 @@
...
+  public async loadResults(
+    variantAnalysisId: number,
+    repoTask: VariantAnalysisRepoTask
+  ): Promise<VariantAnalysisScannedRepositoryResult> {
+    const result = this.cachedResults.get(createCacheKey(variantAnalysisId, repoTask));
+    if (result) {
+      return result;
+    }
+
+    return this.loadResultsIntoMemory(variantAnalysisId, repoTask);
```
elenatanasoiu, line 79: You could shorten this using the [`??` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator):

```
return result ?? this.loadResultsIntoMemory(variantAnalysisId, repoTask);
```

elenatanasoiu (approved):
Left two suggestions, but looks good 👍

extensions/ql-vscode/src/extension.ts
koesie10: I agree with you that this extension file is already doing too much, but unfortunately moving this logic to the constructor of `VariantAnalysisManager` (or even `VariantAnalysisResultsManager`) is not really possible as-is. In a constructor, it's not possible to use `await`, so we wouldn't be able to call `fs.ensureDir`.

We could also skip the call to `fs.ensureDir` since the `VariantAnalysisResultsManager` should be able to work without the directory existing, but I'm not sure whether there's a reason these calls were added in the first place.

elenatanasoiu: That's fair, thanks for looking at it! :)

elenatanasoiu (approved):

PR #1574, opened 2022-10-07
Handle `viewLoaded` message

koesie10:
This will handle the `viewLoaded` message by sending the variant analysis to the webview. It will also clean up some things that are not relevant anymore after this change.

It's probably easiest to review this PR commit-by-commit.

## Checklist


extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts
```diff
@@ -103,152 +93,20 @@ export class VariantAnalysisView extends AbstractWebview<ToVariantAnalysisMessag
...
-      },
-    };
+  protected async onWebViewLoaded() {
+    super.onWebViewLoaded();
+
+    void logger.log('Variant analysis view loaded');
+
+    const variantAnalysis = await this.manager.getVariantAnalysis(this.variantAnalysisId);
+
+    if (!variantAnalysis) {
```
charisk, line 103: Should we show a notification to the user to tell them something went wrong?

extensions/ql-vscode/package.json
```diff
@@ -314,10 +314,6 @@
         "command": "codeQL.exportVariantAnalysisResults",
         "title": "CodeQL: Export Variant Analysis Results"
       },
-      {
```
charisk, line 317: Why does this not provide value any more? It would allow someone (e.g. Design) to load the view without having actually ran a variant analysis. Is it not possible to keep this in a way that it loads a mock variant analysis?

charisk (commented):
Looking really good, just a couple of comments.

extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts
koesie10: I think that would indeed be useful, I'll add a `showAndLogWarningMessage` call here.

extensions/ql-vscode/package.json
koesie10: As it is currently, we would need to implement special handling for a mock view in the `VariantAnalysisView`. The mock view also only implements 1 specific variant analysis, so I think using #1577 would be more useful for someone to view a variant analysis, or using Storybook. I would be happy to add it back, but that will result in some additional code being added to `VariantAnalysisView` for handling a mock view (especially in `onWebViewLoaded`), or there would be some specific code for it in `extension.ts`.

charisk: Let's leave it then - forgot about Storybook!

charisk (approved):

PR #1575, opened 2022-10-10
Reset VSCode configuration between tests

koesie10:
This will introduce a new helper that can be used to set config values in VSCode. Instead of the previous behaviour where it would retain the value after the tests had finished (resulting in e.g. the `codeQL.variantAnalysis.controllerRepo` setting being changed when you run the CLI tests from VSCode's Run and Debug menu), this will restore the value set by the user after finishing all tests. It will also set the value of specified config settings to a known value before each test.

It's probably easiest to review this PR commit-by-commit.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/vscode-tests/test-config.ts
+++ b/extensions/ql-vscode/src/vscode-tests/test-config.ts
@@ -1,5 +1,5 @@
 import { ConfigurationTarget } from 'vscode';
-import { CUSTOM_CODEQL_PATH_SETTING, InspectionResult, REMOTE_CONTROLLER_REPO, REMOTE_REPO_LISTS, Setting } from '../config';
+import { ALL_SETTINGS, InspectionResult, Setting } from '../config';
 
 class TestSetting<T> {
   private initialSettingState: InspectionResult<T> | undefined;
@@ -59,24 +59,27 @@ class TestSetting<T> {
   }
 }
 
-export const testConfig = {
-  remoteControllerRepo: new TestSetting<string>(REMOTE_CONTROLLER_REPO),
-  remoteRepoLists: new TestSetting<Record<string, string[]>>(REMOTE_REPO_LISTS),
-  cliExecutablePath: new TestSetting<string>(CUSTOM_CODEQL_PATH_SETTING),
+// The test settings are all settings in ALL_SETTINGS which don't have any children
+const TEST_SETTINGS = ALL_SETTINGS
+  .filter(setting => ALL_SETTINGS.filter(s => s.parent === setting).length === 0)
+  .map(setting => new TestSetting(setting));
+
+export const getTestSetting = (setting: Setting): TestSetting<unknown> | undefined => {
+  return TEST_SETTINGS.find(testSetting => testSetting.setting === setting);
 };
 
 export const testConfigHelper = async (mocha: Mocha) => {
   // Read in all current settings
-  await Promise.all(Object.values(testConfig).map(setting => setting.initialSetup()));
+  await Promise.all(TEST_SETTINGS.map(setting => setting.initialSetup()));
 
   mocha.rootHooks({
     async beforeEach() {
       // Reset the settings to their initial values before each test
-      await Promise.all(Object.values(testConfig).map(setting => setting.setup()));
+      await Promise.all(TEST_SETTINGS.map(setting => setting.setup()));
     },
     async afterAll() {
       // Restore all settings to their default values after each test suite
-      await Promise.all(Object.values(testConfig).map(setting => setting.restoreToInitialValues()));
+      await Promise.all(TEST_SETTINGS.map(setting => setting.restoreToInitialValues()));
     }
   });
 };
--- a/extensions/ql-vscode/src/vscode-tests/test-config.ts
+++ b/extensions/ql-vscode/src/vscode-tests/test-config.ts
@@ -1,0 +1,93 @@
+import { ConfigurationTarget, workspace } from 'vscode';
+import { CUSTOM_CODEQL_PATH_SETTING, REMOTE_CONTROLLER_REPO, REMOTE_REPO_LISTS, Setting } from '../config';
+
+class TestSetting<T> {
+  private settingState: {
+    globalValue?: T;
+    workspaceValue?: T,
+    workspaceFolderValue?: T,
+  } | undefined;
+
+  constructor(
+    public readonly setting: Setting,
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/vscode-tests/test-config.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  private get section() {
+    if (this.setting.parent === undefined) {
+      throw new Error('Cannot get the value of a root setting.');
+    }
+    return workspace.getConfiguration(this.setting.parent.qualifiedName);
+  }
+}
+
+export const testConfig = {
```
robertbrignull, line 73: This is going a bit further than the other suggestions, but I wondered if it would be possible to:
- Avoid specifying a list of setting names here:
  - It creates duplication as we have to remember to add any new settings to another location.
  - It also means we don't reset all settings unless we manually write all setting names here.
- Avoid tests having to know about and use the test config:
  - It makes test code different from production code.
  - If some test code calls production code that sets config the normal way, there's a chance this might not get reset. It depends on if that setting is listed here, which it might not be if the test code isn't setting it through the test config.

Instead, could we programatically determine the list of all settings? I think we can. Again I had a go at it just to make sure it was possible. You can see my attempt at https://github.com/github/vscode-codeql/commit/d0eca7e01d61b3191755002abb2cbb120b4b3d9d. That also reverts some changes from this PR that are no longer necessary.

It's worth noting that `contributes.configuration.properties` doesn't include every setting, because putting a setting here makes it public. Therefore some settings such as the canary flag and live results flag are deliberately not in this list. You can still do `getValue` to get that value of such a setting, but you can't do `updateValue` as then vscode complains the setting isn't defined. This means for these setting we have to use [stubs in the tests](https://github.com/github/vscode-codeql/blob/5644206777d6942610189290d6721e20ce3a84f0/extensions/ql-vscode/src/vscode-tests/no-workspace/run-queries.test.ts#L22) but this is fine because stubs are restored after each test anyway.

What do you think?

robertbrignull: Update...  it doesn't seem to be working and I'm not sure why. Values aren't being reset between tests. I've been using d3d6c4143ae07f3ff9ffc6bda13a9d8acca96ace to test it.

We could have a look into it, either together or separately, if you think it's worth pursuing.

extensions/ql-vscode/src/vscode-tests/test-config.ts
```diff
@@ -0,0 +1,93 @@
...
+  public async restoreToInitialValues() {
+    const state = this.section.inspect(this.setting.name);
+
+    // We need to check the state of the setting before we restore it. This is less important for the global
+    // configuration target, but the workspace/workspace folder configuration might not even exist. If they
+    // don't exist, VSCode will error when trying to write the new value (even if that value is undefined).
+    if (state?.globalValue !== this.settingState?.globalValue) {
+      await this.set(this.settingState?.globalValue, ConfigurationTarget.Global);
+    }
+    if (state?.workspaceValue !== this.settingState?.workspaceValue) {
```
robertbrignull, line 57: Should this and the check for `workspaceFolderValue` use the same condition as in `initialsetup`? I think either should be fine as they'll have the same effect and both protect against the workspace configuration not existing, having the code be identical could be a nice safety feature.
```suggestion
    if (this.settingState?.workspaceValue !== undefined) {
```

robertbrignull (commented):
I like this. It seemed a little complicated at first, but after playing around a bit I agree with the approach and I can't see a better way.

I have a few ideas about how to make it more robust and less prone to user error as we add or change settings. Forgive me for writing a bunch of code around this but I felt it was the easiest way to communicate what I meant and to try stuff out to make sure it works before I suggest it. Feel free to dispute anything if you think it's not an improvement or not a good idea.

extensions/ql-vscode/src/vscode-tests/test-config.ts
```diff
@@ -0,0 +1,93 @@
...
+      await this.set(this.settingState?.workspaceValue, ConfigurationTarget.Workspace);
+    }
+    if (state?.workspaceFolderValue !== this.settingState?.workspaceFolderValue) {
+      await this.set(this.settingState?.workspaceFolderValue, ConfigurationTarget.WorkspaceFolder);
+    }
+  }
+
+  private get section() {
```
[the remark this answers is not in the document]
koesie10: I do think it's much better to keep this logic centralized in the `Setting`, but didn't really want to conflate the test and normal methods. I think just adding the `inspect` method should be fine, so I'll cherry pick your commit. Thanks!

extensions/ql-vscode/src/vscode-tests/test-config.ts
koesie10: I did think about automatically adding all settings. I didn't think about using the `package.json` for it since that doesn't include the canary and live results flags which are the config settings we're now setting most often in tests.

An alternative solution I did think about is keeping track of all created `Setting` objects and using that to keep a list of all used settings. That would mean the extension has some logic in it purely for tests, but this is fairly minimal (it could be as simple as `ALL_SETTINGS.push(this)` in the `Setting` constructor). What do you think about doing it in this way?

extensions/ql-vscode/src/vscode-tests/test-config.ts
koesie10: I agree that it would be really useful to have identical conditions. However, I don't think this results in identical behavior. For example, if a test is run in a workspace without a `codeQL.canary` setting and a test adds this setting to the workspace, the current behavior would be to remove the `codeQL.canary` flag at the end of the test (since the `state?.workspaceValue === true` and `this.settingState?.workspaceValue === undefined`). With the suggested change, it would not reset the value (since `this.settingState?.workspaceValue === undefined`). So, I don't think this change is safe to make.

robertbrignull: Ah I see what you mean. Thanks for explaining. Then I agree let's keep it as it is.

extensions/ql-vscode/src/vscode-tests/test-config.ts
robertbrignull: I think adding a line to the `Setting` constructor that adds it to some list would be ok. I think it would be worth trying it and seeing how well it works in practice.

robertbrignull (approved):
LGTM

Thanks for talking about different options and trying them out.

PR #1578, opened 2022-10-11
Fix freezing of live results view

koesie10:
It's probably easiest to review this PR commit-by-commit; these also contain some more detailed information about why certain things are necessary. In summary, on every rerender we were registering a new message listener. This would cause every message sent by the extension to trigger N+1 rerenders for the Nth render. The renders would look as follows:

1. 1 message listener, new listener registered. Total rerenders: 1
2. 2 message listeners, new listener registered. Total rerenders: 2
3. 3 message listeners, new listener registered. Total rerenders: 3
4. 4 message listeners, new listener registered. Total rerenders: 4

This would continue forever and would sometimes result in 10's of thousands of renders happening. This caused the live results view to be unresponsive until no more messages were being sent.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/variant-analysis/VariantAnalysis.tsx
+++ b/extensions/ql-vscode/src/view/variant-analysis/VariantAnalysis.tsx
@@ -65,7 +65,7 @@ export function VariantAnalysis({
   const [repoResults, setRepoResults] = useState<VariantAnalysisScannedRepositoryResult[]>(initialRepoResults);
 
   useEffect(() => {
-    window.addEventListener('message', (evt: MessageEvent) => {
+    const listener = (evt: MessageEvent) => {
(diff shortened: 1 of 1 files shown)


charisk (approved):
Good catch! I guess this is technically a problem in other views as well?

koesie10:
Yes, this would also be a problem in other views. However, those other views only receive 1 or 2 messages, so those don't run into the same problem. I'll open another PR to fix it for the compare and remote queries views (the results view does [handle it correctly](https://github.com/github/vscode-codeql/blob/dc6ae6cc39e5f44f0ba60d68a51f54470529c1f9/extensions/ql-vscode/src/view/results/results.tsx#L302-L311)).

aeisenberg:
Thanks for catching this. We'll have to do the same for our other views.

PR #1584, opened 2022-10-12
Download variant analysis results in batches

elenatanasoiu:
In the existing remote queries implementation, we make use of batching when we download results. 

For the `variantAnalysisMonitor`, [we spin off](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts#L76) the download part into its own command so that it doesn't hold up the monitor. 

Let's improve this further by batching download requests. 

---- 

## Questions / Observations

I'm not quite sure how to test that we're triggering batches of 3 downloads at a time. The `AnalysisResultsManager` (which also performs batch downloads) doesn't really have any example tests I could borrow from (or any tests for that matter 😓 ). 

Just as a sanity check, I've added an extra test to check we're only calling the download command once. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
@@ -78,7 +78,7 @@ export class VariantAnalysisMonitor extends DisposableObject {
 
       if (variantAnalysisSummary.scanned_repositories) {
         variantAnalysisSummary.scanned_repositories.forEach(scannedRepo => {
-          if (!scannedReposDownloaded.includes(scannedRepo.repository.id) && scannedRepo.analysis_status === 'succeeded') {
+          if (this.shouldDownload(scannedRepo, scannedReposDownloaded)) {
             this.scheduleForDownload(scannedRepo, variantAnalysisSummary);
             void commands.executeCommand('codeQL.autoDownloadVariantAnalysisResult', scannedRepo, variantAnalysisSummary);
             scannedReposDownloaded.push(scannedRepo.repository.id);
@@ -101,6 +101,13 @@ export class VariantAnalysisMonitor extends DisposableObject {
     variantAnalysisSummary: VariantAnalysisApiResponse
   ) {
     void commands.executeCommand('codeQL.autoDownloadVariantAnalysisResult', scannedRepo, variantAnalysisSummary);
+  }
+
+  private shouldDownload(
+    scannedRepo: VariantAnalysisScannedRepository,
+    alreadyDownloaded: number[]
+  ): boolean {
+    return (!alreadyDownloaded.includes(scannedRepo.repository.id) && scannedRepo.analysis_status === 'succeeded');
   }
 
   private async sleep(ms: number) {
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
```diff
@@ -91,6 +86,57 @@ export class VariantAnalysisMonitor extends DisposableObject {
...
+      const batch = repoResultsToDownload.slice(i, i + batchSize);
+
+      const nwos = batch.map(s => s.repository.full_name).join(', ');
+      const batchTasks = batch.map(scannedRepo => this.scheduleForDownload(scannedRepo, variantAnalysisSummary));
+
+      void this.logger.log(`Downloading batch ${Math.floor(i / batchSize) + 1} of ${numOfBatches} (${nwos})`);
+
+      await Promise.allSettled(batchTasks);
+
+      batch.map(scannedRepo => downloadedRepos.push(scannedRepo.repository.id));
```
koesie10, line 134: I think this can be simplified to:

```suggestion
      downloadRepos.push(...batch.map(scannedRepo => scannedRepo.repository.id));
```

This will also remove the `map` call which we're not using (if that's what we wanted to do, it should probably have been `forEach`).

koesie10 (commented):
Maybe I'm missing something, but from the code it seems like this will schedule the downloads in batches, but not actually limit the number of concurrent downloads. Previously, we would do something like this when 10 repos completed at the same time:

1. Schedule 10 repos

Now, we will do something like this:

1. Schedule 3 repos
2. Schedule 3 repos
3. Schedule 3 repos
4. Schedule 1 repo

However, I don't think there will be a large difference in the execution pattern; both will download 10 repos concurrently since we're not waiting for any of the downloads to complete.

elenatanasoiu:
Thanks @koesie10 ! I've made some changes to use the `p-limit` package you recommended. I ran into some issues with the latest version of that package (which are detailed in the commit message) so had to use a slightly older version that's compatible with Node 16. Let me know what you think!

elenatanasoiu:
I've gone the route of triggering one command and having a method on the `variantAnalysisManager` which makes batch calls to `autoDownloadVariantAnalysisResult` using `p-limit`. 

The method is creatively named `autoDownloadVariantAnalysisResults` : https://github.com/github/vscode-codeql/pull/1584/commits/0fe5e966e4cc924c20d4b40c163ded9d57bfcf78

There's quite a lot of noise in the diff but the end state is:
- before: call a command for each download request which will call `autoDownloadVariantAnalysisResult` in manager
- after: call a command with __all__ the repos to download, trigger `autoDownloadVariantAnalysisResults` in the manager. This in turn will use `p-limit` to trigger `autoDownloadVariantAnalysisResult` in batches.

elenatanasoiu:
As @koesie pointed out, this is still not quite what we want as we're rate limiting per monitor iteration rather than per variant analysis. So this is proving trickier than I expected initially!

I'm going to try a different approach using `p-queue` and `globalStorage` which would create a global queue for these download requests. 

Closing this.

PR #1588, opened 2022-10-12
Add ability to open the query text

koesie10:
This makes it possible to view the query text of a variant analysis. It adds the `queryText` property to the variant analysis and implements all necessary components to open a virtual file.

It's probably easiest to review this commit-by-commit.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
+++ b/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
@@ -8,7 +8,7 @@ export interface VariantAnalysis {
     name: string,
     filePath: string,
     language: VariantAnalysisQueryLanguage,
-    queryText: string,
+    text: string,
   },
   databases: {
     repositories?: string[],
@@ -114,7 +114,7 @@ export interface VariantAnalysisSubmission {
     name: string,
     filePath: string,
     language: VariantAnalysisQueryLanguage,
-    queryText: string,
+    text: string,
 
     // Base64 encoded query pack.
     pack: string,
--- a/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
+++ b/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
@@ -7,7 +7,8 @@ export interface VariantAnalysis {
   query: {
     name: string,
     filePath: string,
-    language: VariantAnalysisQueryLanguage
+    language: VariantAnalysisQueryLanguage,
+    queryText: string,
   },
   databases: {
     repositories?: string[],
@@ -113,6 +114,7 @@ export interface VariantAnalysisSubmission {
     name: string,
     filePath: string,
     language: VariantAnalysisQueryLanguage,
+    queryText: string,
 
     // Base64 encoded query pack.
     pack: string,
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-content-provider.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-content-provider.ts
@@ -10,14 +10,14 @@ export const createVariantAnalysisContentProvider = (variantAnalysisManager: Var
 
     const variantAnalysisIdString = params.get('variantAnalysisId');
     if (!variantAnalysisIdString) {
-      void showAndLogWarningMessage('No variant analysis ID provided');
+      void showAndLogWarningMessage('Unable to show query text. No variant analysis ID provided.');
       return undefined;
     }
     const variantAnalysisId = parseInt(variantAnalysisIdString);
 
     const variantAnalysis = await variantAnalysisManager.getVariantAnalysis(variantAnalysisId);
     if (!variantAnalysis) {
-      void showAndLogWarningMessage('No variant analysis found');
+      void showAndLogWarningMessage('Unable to show query text. No variant analysis found.');
       return undefined;
     }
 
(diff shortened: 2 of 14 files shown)


extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -7,7 +7,8 @@ export interface VariantAnalysis {
   query: {
     name: string,
     filePath: string,
-    language: VariantAnalysisQueryLanguage
+    language: VariantAnalysisQueryLanguage,
+    queryText: string,
```
charisk, line 11: Shall we call this `text` instead? We're inside the `query` object and we haven't prefixed other properties with that.

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -113,6 +114,7 @@ export interface VariantAnalysisSubmission {
     name: string,
     filePath: string,
     language: VariantAnalysisQueryLanguage,
+    queryText: string,
```
charisk, line 117: Why is this repeated here?

koesie10: To construct the `VariantAnalysis`, we initially use the `VariantAnalysisSubmission`. Since the value is not in the returned API response, that means it must be added to the `VariantAnalysisSubmission`

extensions/ql-vscode/src/remote-queries/variant-analysis-content-provider.ts
```diff
@@ -0,0 +1,26 @@
...
+import { SHOW_QUERY_TEXT_MSG } from '../query-history';
+import { VariantAnalysisManager } from './variant-analysis-manager';
+
+export const createVariantAnalysisContentProvider = (variantAnalysisManager: VariantAnalysisManager): TextDocumentContentProvider => ({
+  async provideTextDocumentContent(uri: Uri): Promise<string | undefined> {
+    const params = new URLSearchParams(uri.query);
+
+    const variantAnalysisIdString = params.get('variantAnalysisId');
+    if (!variantAnalysisIdString) {
+      void showAndLogWarningMessage('No variant analysis ID provided');
```
robertbrignull, line 13: Maybe make these error messages slightly more verbose 🤷 
Otherwise I'm not sure it'll be clear what they mean when they pop up.
```suggestion
      void showAndLogWarningMessage('Unable to show query text. No variant analysis ID provided.');
```

extensions/ql-vscode/src/remote-queries/variant-analysis-content-provider.ts
```diff
@@ -0,0 +1,26 @@
...
+      return undefined;
+    }
+    const variantAnalysisId = parseInt(variantAnalysisIdString);
+
+    const variantAnalysis = await variantAnalysisManager.getVariantAnalysis(variantAnalysisId);
+    if (!variantAnalysis) {
+      void showAndLogWarningMessage('No variant analysis found');
```
robertbrignull, line 20:
```suggestion
      void showAndLogWarningMessage('Unable to show query text. No variant analysis found.');
```

extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts
```diff
@@ -130,4 +134,29 @@ export class VariantAnalysisView extends AbstractWebview<ToVariantAnalysisMessag
...
+
+    const filename = variantAnalysis.query.filePath;
+
+    try {
+      const params = new URLSearchParams({
+        variantAnalysisId: variantAnalysis.id.toString(),
+      });
+      const uri = Uri.from({
+        scheme: 'codeql-variant-analysis',
+        path: filename,
```
robertbrignull, line 153: Is this path used for anything? I can't see it being read anywhere in`createVariantAnalysisContentProvider`. I suppose a URI needs to have a path, but I was wondering if it needs to be the query file path or if it could be a fixed path.

extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts
```diff
@@ -130,4 +134,29 @@ export class VariantAnalysisView extends AbstractWebview<ToVariantAnalysisMessag
...
+        query: params.toString(),
+      });
+      const doc = await workspace.openTextDocument(uri);
+      await Window.showTextDocument(doc, { preview: false });
+    } catch (error) {
+      void showAndLogWarningMessage('Could not open query text');
```
robertbrignull, line 159: Just to make the error match the other one earlier in this function. It could be useful having the errors be different so we can tell them apart in bug reports, but if that's the plan let's make them more descriptive, so for example the earlier error would be something like "Could not open variant analysis query text. Variant analysis not found."
```suggestion
      void showAndLogWarningMessage('Could not open variant analysis query text');
```

robertbrignull (commented):
A few suggestions around error messages but that's all. Otherwise looks good to me and well done for working out how to implement the `TextDocumentContentProvider`.

extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts
koesie10: It seems like VSCode uses this path for displaying the filename, so that's why I made it match the filename of the query. Remote queries will always use a fixed filename (`query-text.ql`), but I think it might be useful to also show the actual query filename.

robertbrignull: Ah, makes sense. Thanks for explaining.

robertbrignull (approved):
LGTM

I tried it out locally too and it works well

PR #1599, opened 2022-10-13
Download variant analysis results in batches - take two

elenatanasoiu:
Previous attempt: https://github.com/github/vscode-codeql/pull/1584

In the existing remote queries implementation, we make use of batching when we download results. 

For the `variantAnalysisMonitor`, [we spin off](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts#L76) the download part into its own command so that it doesn't hold up the monitor. 

Let's improve this further by batching download requests. This uses `p-queue` to add download tasks to a queue. We've configured the queue to only run three concurrent download requests at a time. The queue will wait for these to complete before starting new ones. 

Because the queue is stored on the `variantAnalysisManager`, we're able to keep adding new download tasks to it every time the monitor polls the API and finds new results are available. Any new tasks will have to wait for any existing tasks to be completed first.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -952,13 +952,9 @@ async function activateWithInstalledDistribution(
       token: CancellationToken
     ) => {
 
-      const input = [
-        variantAnalysisManager.queue.add(async () => {
-          await variantAnalysisManager.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
-        })
-      ];
-
-      await Promise.all(input);
+      await variantAnalysisManager.queue.add(async () => {
+        await variantAnalysisManager.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
+      });
     })
   );
 
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -951,10 +951,7 @@ async function activateWithInstalledDistribution(
       variantAnalysisSummary: VariantAnalysisApiResponse,
       token: CancellationToken
     ) => {
-
-      await variantAnalysisManager.queue.add(async () => {
-        await variantAnalysisManager.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
-      });
+      await variantAnalysisManager.enqueueDownload(scannedRepo, variantAnalysisSummary, token);
     })
   );
 
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -951,7 +951,14 @@ async function activateWithInstalledDistribution(
       variantAnalysisSummary: VariantAnalysisApiResponse,
       token: CancellationToken
     ) => {
-      await variantAnalysisManager.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
+
+      const input = [
+        variantAnalysisManager.queue.add(async () => {
+          await variantAnalysisManager.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
+        })
+      ];
+
+      await Promise.all(input);
     })
   );
 
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
@@ -22,6 +22,7 @@ import { VariantAnalysisResultsManager } from './variant-analysis-results-manage
 import { CodeQLCliServer } from '../cli';
 import { getControllerRepo } from './run-remote-query';
 import { processUpdatedVariantAnalysis } from './variant-analysis-processor';
+import PQueue from 'p-queue';
 
 export class VariantAnalysisManager extends DisposableObject implements VariantAnalysisViewManager<VariantAnalysisView> {
   private readonly _onVariantAnalysisAdded = this.push(new EventEmitter<VariantAnalysis>());
@@ -31,6 +32,8 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
   private readonly variantAnalysisResultsManager: VariantAnalysisResultsManager;
   private readonly variantAnalyses = new Map<number, VariantAnalysis>();
   private readonly views = new Map<number, VariantAnalysisView>();
+  private static readonly maxConcurrentTasks = 3;
+  public queue: PQueue;
 
   constructor(
     private readonly ctx: ExtensionContext,
@@ -44,6 +47,8 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
 
     this.variantAnalysisResultsManager = this.push(new VariantAnalysisResultsManager(cliServer, storagePath, logger));
     this.variantAnalysisResultsManager.onResultLoaded(this.onRepoResultLoaded.bind(this));
+
+    this.queue = new PQueue({ concurrency: VariantAnalysisManager.maxConcurrentTasks });
   }
 
   public async showView(variantAnalysisId: number): Promise<void> {
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
@@ -169,9 +169,7 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
     variantAnalysisSummary: VariantAnalysisApiResponse,
     token: CancellationToken
   ): Promise<void> {
-    await this.queue.add(async () => {
-      await this.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
-    });
+    await this.queue.add(() => this.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token));
   }
 
   public downloadsQueueSize(): number {
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
@@ -33,7 +33,7 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
   private readonly variantAnalyses = new Map<number, VariantAnalysis>();
   private readonly views = new Map<number, VariantAnalysisView>();
   private static readonly maxConcurrentDownloads = 3;
-  public queue: PQueue;
+  private readonly queue = new PQueue({ concurrency: VariantAnalysisManager.maxConcurrentDownloads });
 
   constructor(
     private readonly ctx: ExtensionContext,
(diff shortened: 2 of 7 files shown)


elenatanasoiu:
The CI failure seems to be unrelated to my changes as it tests a different part of the codebase: `src/vscode-tests/minimal-workspace/databases.test.ts:58:23`

elenatanasoiu:
Looks like the CI failure has been addressed https://github.com/github/vscode-codeql/commit/0885a22984c1b29f4e8768065cc2cc51d8e69fa9

elenatanasoiu:
Looks like there's still [one flakey test](https://github.com/github/vscode-codeql/actions/runs/3263709318/jobs/5363494535) that's unrelated to this PR:
```
"before each" hook: beforeEach for "should avoid creating a quick query"
```

extensions/ql-vscode/src/extension.ts
```diff
@@ -951,7 +951,14 @@ async function activateWithInstalledDistribution(
...
     ) => {
-      await variantAnalysisManager.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
+
+      const input = [
+        variantAnalysisManager.queue.add(async () => {
+          await variantAnalysisManager.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
+        })
+      ];
+
+      await Promise.all(input);
```
koesie10, line 961: I think you should be able to await the call to `queue.add` instead of wrapping it in an array and using `Promise.all`.

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -31,6 +32,8 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
   private readonly variantAnalysisResultsManager: VariantAnalysisResultsManager;
   private readonly variantAnalyses = new Map<number, VariantAnalysis>();
   private readonly views = new Map<number, VariantAnalysisView>();
+  private static readonly maxConcurrentTasks = 3;
+  public queue: PQueue;
```
koesie10, line 36: I think it might be better to keep this private and expose a method that will queue a new variant analysis download. That way, we can change the implementation of this method without needing to change all consumers (even though that's currently only in two files).

We should also be able to make this `readonly` and create the queue inline, so something like this:

```suggestion
  private readonly queue = = new PQueue({ concurrency: VariantAnalysisManager.maxConcurrentTasks });
```

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
```diff
@@ -149,6 +149,41 @@ describe('Variant Analysis Manager', async function() {
...
+        const getResultsSpy = sandbox.spy(variantAnalysisManager, 'autoDownloadVariantAnalysisResult');
+
+        const input = [
+          variantAnalysisManager.queue.add(async () => {
+            await variantAnalysisManager.autoDownloadVariantAnalysisResult(
+              scannedRepos[0],
+              variantAnalysis,
+              cancellationTokenSource.token
+            );
+          }),
```
koesie10, line 163: You shouldn't need to create a new `async` function which awaits the call, it should be possible to skip the `await` since the `autoDownloadVariantAnalysisResult` creates a promise anyway:

```suggestion
          variantAnalysisManager.queue.add(variantAnalysisManager.autoDownloadVariantAnalysisResult(
            scannedRepos[0],
            variantAnalysis,
            cancellationTokenSource.token
          )),
```

koesie10 (commented):
I'm not sure how to best test this. One idea I have is to mock the `autoDownloadVariantAnalysisResult` call and use a different implementation that just stores the time it started, sleep for a bit and then store the time it finished. We might then be able to compare the times and check that only three of these times overlap at once. I'm not completely sure how feasible it is or whether it would add value, since we should be able to assume that `p-queue` functions correctly.

extensions/ql-vscode/src/extension.ts
```diff
@@ -951,7 +951,14 @@ async function activateWithInstalledDistribution(
       variantAnalysisSummary: VariantAnalysisApiResponse,
       token: CancellationToken
     ) => {
-      await variantAnalysisManager.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
+
+      const input = [
+        variantAnalysisManager.queue.add(async () => {
```
robertbrignull, line 956: Is it worth it / better to move this implementation to a new method on `VariantAnalysisManager`? It would be a small function but the benefits would be:
- Keeping `extension.ts` as simple as possible
- Keeping knowledge of `queue` internal
  - Sadly the field still have to be `public` so it can be accessed from the tests, but that's fine

robertbrignull: I see this is similar to a [suggestion by Koen](https://github.com/github/vscode-codeql/pull/1599/files#r996838279) too

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -31,6 +32,8 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
   private readonly variantAnalysisResultsManager: VariantAnalysisResultsManager;
   private readonly variantAnalyses = new Map<number, VariantAnalysis>();
   private readonly views = new Map<number, VariantAnalysisView>();
+  private static readonly maxConcurrentTasks = 3;
```
robertbrignull, line 35: Is the intention that this queue handles more than just downloads? I wasn't expecting so but the generic naming implies this could be case.

I'd suggest rename to `maxConcurrentDownloads` and `downloadQueue` to make it more specific and clear what the fields are for.

I doubt we'll ever want to reuse these for more generic tasks, but if so we can rename the fields again at that point.

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
```diff
@@ -94,6 +89,46 @@ export class VariantAnalysisMonitor extends DisposableObject {
...
+    variantAnalysisSummary: VariantAnalysisApiResponse
+  ) {
+    void commands.executeCommand('codeQL.autoDownloadVariantAnalysisResult', scannedRepo, variantAnalysisSummary);
+  }
+
+  private shouldDownload(
+    scannedRepo: VariantAnalysisScannedRepository,
+    alreadyDownloaded: number[]
+  ): boolean {
+    return (!alreadyDownloaded.includes(scannedRepo.repository.id) && scannedRepo.analysis_status === 'succeeded');
```
robertbrignull, line 103:
```suggestion
    return !alreadyDownloaded.includes(scannedRepo.repository.id) && scannedRepo.analysis_status === 'succeeded';
```
Unless I'm mistaken, the braces are unnecessary. And I'm not sure they add to readability.

robertbrignull (approved):
I'm also not sure the best way to test this but @koesie10's suggestion sounds good. So we use a mocked/modified version that logs timestamps on starting/finishing a mocked "download". Then we manually look at the logs and check everything is being limited to the expected concurrency.

robertbrignull:
I think this approach looks good and should work just fine. All the comments we've had are stylistic or minor improvements.

extensions/ql-vscode/src/extension.ts
```diff
@@ -952,13 +952,9 @@ async function activateWithInstalledDistribution(
...
-      const input = [
-        variantAnalysisManager.queue.add(async () => {
-          await variantAnalysisManager.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
-        })
-      ];
-
-      await Promise.all(input);
+      await variantAnalysisManager.queue.add(async () => {
+        await variantAnalysisManager.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
+      });
```
elenatanasoiu: I initially experimented with:

```
await variantAnalysisManager.queue.add(variantAnalysisManager.autoDownloadVariantAnalysisResult(scannedRepo, variantAnalysisSummary, token);
});
```

but `queue.add()` expects to be passed a function as a parameter, rather than a promise.

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
elenatanasoiu: Have made this private: https://github.com/github/vscode-codeql/pull/1599/commits/1f4e1f27ae622f5b8231ae30618cb4f8957f1751

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
elenatanasoiu: I did try this, but the `.add()` method expects a function rather than a promise. 

Screenshot 2022-10-17 at 14 39 30

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
elenatanasoiu: Renamed: https://github.com/github/vscode-codeql/pull/1599/commits/f2fe1063d93aa41d5eb7d23d61d1edd99f677aae

extensions/ql-vscode/src/remote-queries/variant-analysis-monitor.ts
elenatanasoiu: Removed: https://github.com/github/vscode-codeql/pull/1599/commits/7d2bae1f1b61c9234d5839701205d8e41797c9ba

elenatanasoiu:
On the testing front, I had a breakthrough. I was initially unsure of how to stub `autoDownloadVariantAnalysisResult ` on the `variantAnalysisManager` since you don't have access to it via the monitor. HOWEVER, you can access it via the extension (which we load in the tests). 

So I was able to finally check that the download method is called correctly for each repo result: https://github.com/github/vscode-codeql/pull/1599/commits/bf7509e3df43e2517a75b1dd2322d06c82348877

As a stretch goal I would've really liked to test that if new repos appear, we only attempt to download the new results rather than the whole list. 

However, I was unable to stub the API to return different responses. Here's a rough draft of my test:
```
describe('when there are successfully scanned repos', async () => {
	beforeEach(async function() {
	  scannedRepos = createMockScannedRepos(['pending', 'pending', 'in_progress', 'in_progress', 'succeeded', 'succeeded', 'succeeded']);

	  const newScannedRepos = createMockScannedRepos(['succeeded', 'succeeded', 'in_progress']);
	  const newList = [...scannedRepos, ...newScannedRepos];

	  mockApiResponse = createMockApiResponse('completed', scannedRepos);
	  const mockApiResponse2 = createMockApiResponse('completed', newList);
	  
	  mockGetVariantAnalysis = sandbox.stub(ghApiClient, 'getVariantAnalysis')
	    .onFirstCall().resolves(mockApiResponse)
	    .onSecondCall().resolves(mockApiResponse2);

	  succeededRepos = newList.filter(r => r.analysis_status === 'succeeded');
	});

...

	it('should only try to download each repo once', async () => {
      await variantAnalysisMonitor.monitorVariantAnalysis(variantAnalysis, cancellationTokenSource.token);

      expect(mockGetDownloadResult).to.have.callCount(5);
    });
});		
``` 

I'm open to suggestions but I think this can be part of a the work we have for taking another look at testing for the monitor.

koesie10 (approved):

robertbrignull (approved):
LGTM too

PR #1601, opened 2022-10-14
Add display of duration and completion status

koesie10:
This uses some new fields on the API to show the duration and completion status of a variant analysis. See the individual commits for more details.

## Checklist


extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisHeader.tsx
```diff
@@ -63,6 +59,16 @@ export const VariantAnalysisHeader = ({
     return getSkippedRepoCount(variantAnalysis.skippedRepos) > 0;
   }, [variantAnalysis.skippedRepos]);
 
+  const duration = useMemo(() => {
```
charisk, line 62: Can we have some tests for this please?

extensions/ql-vscode/src/view/variant-analysis/__tests__/VariantAnalysisStats.spec.tsx
```diff
@@ -100,4 +101,22 @@ describe(VariantAnalysisStats.name, () => {
     expect(screen.getByText('Succeeded')).toBeInTheDocument();
     expect(screen.queryByText('Succeeded warnings')).not.toBeInTheDocument();
   });
+
+  it('renders the duration when it is less than a second', () => {
```
charisk, line 105: Is it worth adding one more test case for when the duration shouldn't be rendered (i.e. when `completedAt` isn't available).?

charisk (commented):
LGTM, just a suggestion for an extra test case.

charisk (approved):

extensions/ql-vscode/src/view/variant-analysis/__tests__/VariantAnalysisStats.spec.tsx
koesie10: I'll add a test case for it, but it's somewhat hard to test this since there's not a clearly correct text that should be shown. `-` will be shown when there is no duration, but this will also be shown when there are no results. I've now added some simple test cases that test that the correct number of `-` are shown, but this might be somewhat less robust again changes in this component.

PR #1594, opened 2022-10-13
Typo fixes in query history methods/properties

shati-patel:
I'm making some changes to the query history, and spotted a few minor typos. I've pulled these out to this separate PR, to avoid cluttering up the "actual" changes later! Lots of changed files, but no actual impact 🤞🏽 

(PS: I double-checked the currently-open PR, and this shouldn't conflict with anything there) 

## Checklist

- [x] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [x] Issues have been created for any UI or other user-facing changes made by this pull request.
- [x] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #1602, opened 2022-10-14
Use a method to handle results without sucess data

alexet:
This fixes a bug when loading old results where we think they weren't successful. This is a bit awkward due to having to create a method as the old getter would no longer work.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-results.ts
+++ b/extensions/ql-vscode/src/query-results.ts
@@ -101,6 +101,14 @@ export class CompletedQueryInfo implements QueryWithResults {
     this.resultCount = value;
   }
 
+  ranSucessfully(): boolean {
+    if(this.successful !== undefined) {
+      return this.successful; 
+    } else {
+      return this.result.resultType == legacyMessages.QueryResultType.SUCCESS;
+    }
+  }
+
   get statusString(): string {
     if (this.message) {
       return this.message;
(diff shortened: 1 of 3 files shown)


charisk:
I believe that another way to fix this would be to patch the data when reading it from the query history (see https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/query-serialization.ts#L29). That way we don't have to deal with the complexity further up.

extensions/ql-vscode/src/query-results.ts
```diff
@@ -101,6 +101,14 @@ export class CompletedQueryInfo implements QueryWithResults {
     this.resultCount = value;
   }
 
+  ranSucessfully(): boolean {
+    if(this.successful !== undefined) {
+      return this.successful; 
+    } else {
+      return this.result.resultType == legacyMessages.QueryResultType.SUCCESS;
```
aeisenberg, line 108: `===`

aeisenberg (commented):
Minor suggestion and then good to go.

aeisenberg:
Actually, I like @charis's idea better.

aeisenberg:
Problem is different from the solution here. There was a typo fix that was merged right around the same time as this and changed the serialization code. https://github.com/github/vscode-codeql/pull/1594 I am going to close this PR and work on a different fix.

PR #1604, opened 2022-10-14
Make bad join order warning threshold configurable

dbartol:
The threshold at which the bad join order detection reports a warning was previously hard-coded to 50. Initial feedback from internal QL developers suggests that this is too high, and should be configurable in any case. I've made it configurable via the `codeQL.logInsights.joinOrderWarningThreshold` setting, leaving the default at 50. Once we get more feedback about what a better default value is, I'll update the default.

## Checklist


extensions/ql-vscode/src/log-insights/join-order.ts
```diff
@@ -454,7 +452,11 @@ class JoinOrderScanner implements EvaluationLogScanner {
 }
 
 export class JoinOrderScannerProvider implements EvaluationLogScannerProvider {
+  constructor(private readonly getThreshdold: () => number) {
+  }
+
   public createScanner(problemReporter: EvaluationLogProblemReporter): EvaluationLogScanner {
-    return new JoinOrderScanner(problemReporter, DEFAULT_WARNING_THRESHOLD);
+    const threshold = this.getThreshdold();
+    return new JoinOrderScanner(problemReporter, threshold);
```
aeisenberg, line 460: So, this means that once the logger is created, the threshold can't be changed for that current run. I think that's fine.

aeisenberg (approved):

PR #1607, opened 2022-10-14
`loadPersistedState` should happen outside of dbm constructor

aeisenberg:
Also, add stub to logger in tests.

This fixes some occasionally failing tests on main.


angelapwen (approved):
Thanks for finding this!

PR #1611, opened 2022-10-14
Test cleanups

aeisenberg:
- Avoid Installing `xvfb` since it is already available.
- Ensure `supportsNewQueryServer()` takes the CLI version into account
- Always run the new query server tests on v2.11.1 and later
- Avoid printing directory contents in `run-remote-query-tests`
- Run tests with `--disable-workspace-trust` to avoid a non-fatal error
  being thrown from the dialog service.
- Ensure the exit code of the extension host while running integration
    1. tests is the exit code of the actual process. Otherwise, there is
    2. a possibility that an error exit code is swallowed up and ignored.
- Remove a duplicate unhandledRejection handler.
- Attempt to handle mysterious Exit code 7 from windows

Also, I tried to remove all those pesky errors in the logs like:

> [2094:1017/235357.424002:ERROR:bus.cc(398)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")

I was following advice from here, but I can't get it working.

- https://github.com/microsoft/vscode-test/issues/127
- https://github.com/electron/electron/issues/31981


aeisenberg:
This is ready for review now. I think CI is now passing regularly. Some of the changes might be controversial, so please have a review @github/code-scanning-secexp-reviewers and @github/codeql-vscode-reviewers.

extensions/ql-vscode/src/config.ts
```diff
@@ -11,9 +11,10 @@ export class Setting {
   name: string;
   parent?: Setting;
 
-  constructor(name: string, parent?: Setting) {
+  constructor(name: string, parent?: Setting, public readonly isHidden = false) {
```
robertbrignull, line 14: What's the definition of a hidden config setting?

I think the set that are visible to users are the ones listed under `contributes.configuration.properties` in the `package.json`. I believe this list is:
```
codeQL.cli.executablePath
codeQL.logInsights.joinOrderWarningThreshold
codeQL.queryHistory.format
codeQL.queryHistory.ttl
codeQL.resultsDisplay.pageSize
codeQL.runningQueries.autoSave
codeQL.runningQueries.cacheSize
codeQL.runningQueries.customLogDirectory
codeQL.runningQueries.debug
codeQL.runningQueries.maxPaths
codeQL.runningQueries.maxQueries
codeQL.runningQueries.memory
codeQL.runningQueries.numberOfThreads
codeQL.runningQueries.quickEvalCodelens
codeQL.runningQueries.saveCache
codeQL.runningQueries.timeout
codeQL.runningTests.additionalTestArguments
codeQL.runningTests.numberOfThreads
codeQL.telemetry.enableTelemetry
codeQL.telemetry.logTelemetry
codeQL.variantAnalysis.controllerRepo
codeQL.variantAnalysis.repositoryLists
```

And the list of all settings defined in this file is:
```
codeQL.cli.executablePath
codeQL.logInsights.joinOrderWarningThreshold
codeQL.queryHistory.format
codeQL.queryHistory.ttl
codeQL.resultsDisplay.pageSize
codeQL.runningQueries.autoSave
codeQL.runningQueries.cacheSize
codeQL.runningQueries.customLogDirectory
codeQL.runningQueries.debug
codeQL.runningQueries.maxPaths
codeQL.runningQueries.maxQueries
codeQL.runningQueries.memory
codeQL.runningQueries.numberOfThreads
codeQL.runningQueries.quickEvalCodelens
codeQL.runningQueries.saveCache
codeQL.runningQueries.timeout
codeQL.runningTests.additionalTestArguments
codeQL.runningTests.numberOfThreads
codeQL.telemetry.enableTelemetry
codeQL.telemetry.logTelemetry
codeQL.variantAnalysis.controllerRepo
codeQL.variantAnalysis.repositoryLists
```

So by this definition the set of "hidden" settings is:
```
codeQL.astViewer.disableCache
codeQL.canary
codeQL.canaryQueryServer
codeQL.cli.includePrerelease
codeQL.cli.personalAccessToken
codeQL.variantAnalysis.actionBranch
codeQL.variantAnalysis.liveResults
codeQL.variantAnalysis.repositoryListsPath
telemetry.enableTelemetry
```

extensions/ql-vscode/src/vscode-tests/test-config.ts
```diff
@@ -79,7 +82,10 @@ export const testConfigHelper = async (mocha: Mocha) => {
     },
     async afterAll() {
       // Restore all settings to their default values after each test suite
-      await Promise.all(TEST_SETTINGS.map(setting => setting.restoreToInitialValues()));
+      // Only do this outside of CI since the sometimes hangs on CI.
```
robertbrignull, line 85: Why does this hang in the `afterAll` but not in the `beforeEach`?

robertbrignull (commented):
Thank you for putting in the effort to make the tests runs more reliably and cleanly! ❤️ 

I don't have context on most of the changes here, but I've had a look through those bits I do understand.

extensions/ql-vscode/src/config.ts
aeisenberg: Thank you for a detailed look here. You're right. I should be making the settings above "hidden". Now that I think about it, I can do this more precisely and compare the settings defined in the config module with those defined in package.json. This way, we won't need to remember to mark config settings as hidden. It will happen programmatically.

extensions/ql-vscode/src/vscode-tests/test-config.ts
aeisenberg: I'm not exactly sure why. It looks like by the time the `afterAll` the workspace is already shut down and errors are thrown when you try to write settings. This seems to have been causing sporadic issues with the cli-integration tests on windows. The error stack traces are helpfully(!) hidden unless you run vscode with verbose logging.

robertbrignull (approved):

PR #1614, opened 2022-10-17
Get query ID for query history items (incl `VariantAnalysisHistoryItem`)

shati-patel:
Adds a new `getQueryId` helper function, which returns the ID for any `QueryHistoryInfo` item, including the new variant analysis case.

See internal linked issue for context and related work! 

## Checklist

N/A - internal only 🦔


extensions/ql-vscode/src/query-history-info.ts
```diff
@@ -17,3 +17,16 @@ export function getRawQueryName(item: QueryHistoryInfo): string {
       assertNever(item);
   }
 }
+
+export function getQueryId(item: QueryHistoryInfo): string {
```
charisk, line 21: Can we name this `getQueryHistoryItemId` instead? I think `queryId` isn't quite correct - we're just looking for something that identifies and separates between items in the query history.

extensions/ql-vscode/src/query-history-info.ts
```diff
@@ -17,3 +17,16 @@ export function getRawQueryName(item: QueryHistoryInfo): string {
...
+  switch (item.t) {
+    case 'local':
+      return item.initialInfo.id;
+    case 'remote':
+      return item.queryId;
+    case 'variant-analysis':
+      return item.variantAnalysis.id.toString();
```
charisk, line 28: Appreciate this was kind of my idea but I'm wondering if this is the right id to be using now 🤔 (sorry!)

I think it's okay because both local and remote queries use `nanoid`, but if we had a different item that had an id that is a number (like variant analysis), then this wouldn't be unique enough.

Perhaps we need to prefix with `variant-analysis-${item.variantAnalysis.id}`? Or we could generate one with `nanoid` and store it as part of the history item.

I'm leaning towards the latter (have an id that specifically identifies the history item, and generating that id through `nanoid`). Probably best to get team consensus on this one.

robertbrignull: Just trying to understand the purpose of this ID. So does this not ID need to be consistent at all between different loadings of the extension, or do we store the generated ID when we serialise the query history?

The definition of `createQueryId` from `remote-queries-manager.ts` says
> Generates a unique id for this query, suitable for determining the storage location for the downloaded query artifacts.

Assuming I understand correctly how this ID is used, the only "benefit" I can see to using `variant-analysis-${item.variantAnalysis.id}` is that we would re-use the same storage location directory for downloaded artifacts. I'm not sure if this is strictly even a benefit or rather a potential bug, and it's certainly not very impactful.

So I also think a `nanoid` is the right option.

shati-patel: Thanks! I'll change this to have a new top-level `queryId` property using `nanoid` 👍🏽

charisk: We also use this id to identify items in the query history (although it looks like we use type + id so it doesn't really have to be unique across all items) e.g. https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/query-history.ts#L722.

Let's go with `nanoid`! Can we call it `historyItemId` or something like that?

extensions/ql-vscode/src/query-history.ts
```diff
@@ -603,6 +604,7 @@ export class QueryHistoryManager extends DisposableObject {
         t: 'variant-analysis',
         status: QueryStatus.InProgress,
         completed: false,
+        queryId: nanoid(),
```
shati-patel, line 607: Does this look like the right place to create the `queryId`? (as in, in the `onVariantAnalysisAdded` event)

charisk: Yep!

robertbrignull: Creating the ID here looks fine to me too.

The remote queries version creates the ID one step and includes it in the `NewQueryEvent`, but I don't think this is necessarily better and I don't think it makes a difference.

It's also something we can easily change in the future so it's very low risk.

robertbrignull (approved):
LGTM

extensions/ql-vscode/src/query-history.ts
```diff
@@ -1068,19 +1070,13 @@ export class QueryHistoryManager extends DisposableObject {
       queryText: encodeURIComponent(await this.getQueryText(finalSingleItem)),
     });
 
-    if (finalSingleItem.t === 'variant-analysis') {
-      // TODO
-    } else {
-      const queryId = finalSingleItem.t === 'local'
-        ? finalSingleItem.initialInfo.id
-        : finalSingleItem.queryId;
+    const queryId = getQueryHistoryItemId(finalSingleItem);
```
charisk, line 1073: `historyItemId`? :)

charisk (approved):
Thanks for sorting this out!

PR #1620, opened 2022-10-18
QueryHistory: Consume event when variant analysis status is updated

elenatanasoiu:
Update the state of a query history item when the status of the variant analysis has changed.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -614,18 +614,20 @@ export class QueryHistoryManager extends DisposableObject {
     });
 
     const variantAnalysisStatusUpdateSubscription = this.variantAnalysisManager.onVariantAnalysisStatusUpdated(async (variantAnalysis) => {
-      const item = this.treeDataProvider.allHistory.find(i => i.t === 'variant-analysis' && i.variantAnalysis.id === variantAnalysis.id);
+      const items = this.treeDataProvider.allHistory.filter(i => i.t === 'variant-analysis' && i.variantAnalysis.id === variantAnalysis.id);
       const status = variantAnalysisStatusToQueryStatus(variantAnalysis.status);
 
-      if (item) {
-        const variantAnalysisHistoryItem = item as VariantAnalysisHistoryItem;
-        variantAnalysisHistoryItem.status = status;
-        variantAnalysisHistoryItem.failureReason = variantAnalysis.failureReason;
-        variantAnalysisHistoryItem.resultCount = getTotalResultCount(variantAnalysis.scannedRepos);
-        variantAnalysisHistoryItem.variantAnalysis = variantAnalysis;
-        if (status === QueryStatus.Completed) {
-          variantAnalysisHistoryItem.completed = true;
-        }
+      if (items.length > 0) {
+        items.forEach(async (item) => {
+          const variantAnalysisHistoryItem = item as VariantAnalysisHistoryItem;
+          variantAnalysisHistoryItem.status = status;
+          variantAnalysisHistoryItem.failureReason = variantAnalysis.failureReason;
+          variantAnalysisHistoryItem.resultCount = getTotalResultCount(variantAnalysis.scannedRepos);
+          variantAnalysisHistoryItem.variantAnalysis = variantAnalysis;
+          if (status === QueryStatus.Completed) {
+            variantAnalysisHistoryItem.completed = true;
+          }
+        });
         await this.refreshTreeView();
       } else {
         void logger.log('Variant analysis status update event received for unknown variant analysis');
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -622,6 +622,7 @@ export class QueryHistoryManager extends DisposableObject {
         variantAnalysisHistoryItem.status = status;
         variantAnalysisHistoryItem.failureReason = variantAnalysis.failureReason;
         variantAnalysisHistoryItem.resultCount = getTotalResultCount(variantAnalysis.scannedRepos);
+        variantAnalysisHistoryItem.variantAnalysis = variantAnalysis;
         if (status === QueryStatus.Completed) {
           variantAnalysisHistoryItem.completed = true;
         }
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -614,7 +614,7 @@ export class QueryHistoryManager extends DisposableObject {
     });
 
     const variantAnalysisStatusUpdateSubscription = this.variantAnalysisManager.onVariantAnalysisStatusUpdated(async (variantAnalysis) => {
-      const item = this.treeDataProvider.allHistory.find(i => i.t === 'variant-analysis' && i.historyItemId === variantAnalysis.id.toString());
+      const item = this.treeDataProvider.allHistory.find(i => i.t === 'variant-analysis' && i.variantAnalysis.id === variantAnalysis.id);
       const status = variantAnalysisStatusToQueryStatus(variantAnalysis.status);
 
       if (item) {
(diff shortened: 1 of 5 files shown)


elenatanasoiu:
Not sure why the commit is marked as failed, but CI is ✅

extensions/ql-vscode/src/query-history.ts
```diff
@@ -611,7 +613,26 @@ export class QueryHistoryManager extends DisposableObject {
       await this.refreshTreeView();
     });
 
+    const variantAnalysisStatusUpdateSubscription = this.variantAnalysisManager.onVariantAnalysisStatusUpdated(async (variantAnalysis) => {
+      const item = this.treeDataProvider.allHistory.find(i => i.t === 'variant-analysis' && i.historyItemId === variantAnalysis.id.toString());
+      const status = variantAnalysisStatusToQueryStatus(variantAnalysis.status);
+
+      if (item) {
+        const variantAnalysisHistoryItem = item as VariantAnalysisHistoryItem;
```
koesie10, line 621: Do we also need to update the `variantAnalysis` property of the `item`?

extensions/ql-vscode/src/query-status.ts
```diff
@@ -1,5 +1,23 @@
...
+export function variantAnalysisStatusToQueryStatus(status: VariantAnalysisStatus): QueryStatus {
+  switch (status) {
+    case VariantAnalysisStatus.Succeeded:
+      return QueryStatus.Completed;
+    case VariantAnalysisStatus.Failed:
+      return QueryStatus.Failed;
+    case VariantAnalysisStatus.InProgress:
+      return QueryStatus.InProgress;
+    case VariantAnalysisStatus.Canceled:
+      return QueryStatus.Completed;
```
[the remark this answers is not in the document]
elenatanasoiu: We did discuss this during the mobbing session and ended up deciding that `Canceled` would be `Completed`.

I'm fine with either option tbh. 

@robertbrignull, @norascheuch, @shati-patel any strong opinions?

extensions/ql-vscode/src/query-history.ts
elenatanasoiu: Right you are @koesie10. I find it confusing that we're storing both the `variantAnalysis` and properties from it (like `failureReason`).

I think the lack of a test here really shows how easy it is to miss things when the code isn't being exercised anywhere. 

It's unfortunate these methods are private so we can't add unit tests.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -611,7 +613,27 @@ export class QueryHistoryManager extends DisposableObject {
       await this.refreshTreeView();
     });
 
+    const variantAnalysisStatusUpdateSubscription = this.variantAnalysisManager.onVariantAnalysisStatusUpdated(async (variantAnalysis) => {
+      const item = this.treeDataProvider.allHistory.find(i => i.t === 'variant-analysis' && i.historyItemId === variantAnalysis.id.toString());
```
koesie10, line 617: Sorry for missing this earlier, but it seems like the `historyItemId` is actually a `nanoid`, so this would never match any history items. Instead, I think we'd need to check `i.variantAnalysis.id`.

elenatanasoiu: Hmmm, I was under the impression that we decided to go with a `nanoid` specifically to be consistent with the way we identify all the other types of query history items: https://github.com/github/vscode-codeql/pull/1614#discussion_r996891436

koesie10: I think that's correct, but that does mean we need to do something like this instead:

```suggestion
      const item = this.treeDataProvider.allHistory.find(i => i.t === 'variant-analysis' && i.variantAnalysis.id === variantAnalysis.id);
```

Essentially, the `historyItemId` is completely independent of the variant analysis and cannot be used to find the matching variant analysis. That also means that we don't have a 1-to-1 relationship between a variant analysis and a query history item. It might even be the case that there can be multiple history items for the same variant analysis (at least in theory), so it might even make sense to make this a `filter` call and update all matching history items.

extensions/ql-vscode/src/query-status.ts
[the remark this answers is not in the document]
shati-patel: I think our argument for `Canceled` was that it's a conscious user action, as in, they have successfully halted the variant analysis when they wanted to 😅 Assuming at least one of the repo tasks succeeded, there will be some completed results to display.

`Failed` sounded more like "something went wrong".

(I can't see any specific designs for the canceled case though, so I'm not sure how we need to indicate this in the view/history 🤔 Perhaps as "completed" but with a warning message, rather than an error.)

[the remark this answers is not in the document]
robertbrignull: I agree with @shati-patel, that was our reasoning from the meeting.

I suppose we potentially also have the option to add a `Cancelled` option to `QueryStatus`. Showing as "completed but with a warning" sounds like a good compromise to me.

We could just go with either option for now and raise it as a point for design to look at and decide what to do in this case?

extensions/ql-vscode/src/query-history.ts
elenatanasoiu: Alrighty, I've changed this.

elenatanasoiu: Have also changed this to use `filter`: https://github.com/github/vscode-codeql/pull/1620/commits/3a7f2000622e65cacf489a90515c57c9a923df2a

extensions/ql-vscode/src/query-history.ts
```diff
@@ -611,7 +613,29 @@ export class QueryHistoryManager extends DisposableObject {
...
+        items.forEach(async (item) => {
+          const variantAnalysisHistoryItem = item as VariantAnalysisHistoryItem;
+          variantAnalysisHistoryItem.status = status;
+          variantAnalysisHistoryItem.failureReason = variantAnalysis.failureReason;
+          variantAnalysisHistoryItem.resultCount = getTotalResultCount(variantAnalysis.scannedRepos);
+          variantAnalysisHistoryItem.variantAnalysis = variantAnalysis;
+          if (status === QueryStatus.Completed) {
+            variantAnalysisHistoryItem.completed = true;
+          }
+          await this.refreshTreeView();
```
robertbrignull, line 630: Thanks for making this work on all items except just one. In normal operation I'd hope there should just be one history item per variant analysis, but this does seem like a nice way of making things work in case that invariant is breached.

However, should we move this call to `refreshTreeView` to be after the loop?

extensions/ql-vscode/src/query-history.ts
robertbrignull: If there's duplicate data between the variant analysis and the other fields, I think we have the option to remove the extra fields. When we need those values we can calculate them from the variant anlaysis.

Would be best done as a separate PR though. Could you could open an issue to look into this?

robertbrignull (commented):
Generally looks good to me, minus one point and ongoing discussion about statuses.

extensions/ql-vscode/src/query-history.ts
robertbrignull: Just a note that it's not out of the picture to change the `historyItemId` to be something determined from the variant analysis ID instead of a random ID. The thinking about a "1-to-1 relationship between a variant analysis and a query history item" does make that seem a bit more attractive. But I don't see it as hugely urgent so we can leave things as they are for now and in the next couple of weeks see if this choice would make things better overall. How does that sound?

extensions/ql-vscode/src/query-history.ts
elenatanasoiu: Have added an issue 👍

extensions/ql-vscode/src/query-history.ts
elenatanasoiu: Sounds good, I've written up an issue for this as well! 

:all-the-issues-emoji:

extensions/ql-vscode/src/query-history.ts
elenatanasoiu: Have moved the call. Thanks!

elenatanasoiu:
I've had to fix some merge conflicts to account for https://github.com/github/vscode-codeql/pull/1621

elenatanasoiu:
Aaand now CI is timing out on a CLI test :/

robertbrignull (approved):

robertbrignull:
> Aaand now CI is timing out on a CLI test :/

Not sure if it's exactly the same, but I've seen timeouts when I've awaited something I shouldn't do. In my case it was accidentally awaiting a log statement, which you shouldn't do because it waits for the user to dismiss the notification. So that's something to check, though I can't see anything of that sort in this PR.

PR #1623, opened 2022-10-18
Fix missing DIL for new query server

alexet:
This makes the new queryserver write the DIL to the DIL path which should make it work. It also fixes the error message to report the DIL location.

I can't get extension development to work currently so I can't test this so whoever reviews this should test it.

### DIFF
--- a/extensions/ql-vscode/src/run-queries-shared.ts
+++ b/extensions/ql-vscode/src/run-queries-shared.ts
@@ -198,7 +198,7 @@ export class QueryEvaluationInfo {
         // This could be from the new query server
         // in which case we expect the qlo to be missing so we should ignore it
         throw new Error(
-          `DIL was not found. ${this.dilPath}`
+          `DIL was not found. Expected location: '${this.dilPath}'`
         );
       } else {
         throw new Error(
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -198,7 +198,7 @@ export class QueryEvaluationInfo {
         // This could be from the new query server
         // in which case we expect the qlo to be missing so we should ignore it
         throw new Error(
-          `DIL was not found. ${compiledQuery}`
+          `DIL was not found. ${this.dilPath}`
```
aeisenberg, line 201:
```suggestion
          `DIL was not found. Expected location: '${this.dilPath}'`
```

hmakholm:
I suspect a CLI change will also be necessary, since `QueryServer2Command` never calls `CompilationCommon.setIncludeDILinQLO(true)`.

alexet:
Yeah Henning is correct. I have managed to test this and it tells the query server to produce DIL but the new query server produces an empty file.

aeisenberg (approved):
This looks right to me.

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -198,7 +198,7 @@ export class QueryEvaluationInfo {
         // This could be from the new query server
         // in which case we expect the qlo to be missing so we should ignore it
```
aeisenberg, line 199: This comment is no longer relevant.

alexet: No it is a still correct. The point was that if we are using the new query server we don't expect a qlo so we shouldn't report that the qlo is missing and report that the DIL is missing. If we are still using the old queryserver then a missing qlo is likely to be the problem.

PR #1625, opened 2022-10-18
A couple of small changes around the new query server

aeisenberg:
@alexet These are minor cleanups to your PR #1623.

### DIFF
--- a/extensions/ql-vscode/src/run-queries-shared.ts
+++ b/extensions/ql-vscode/src/run-queries-shared.ts
@@ -192,7 +192,7 @@ export class QueryEvaluationInfo {
     if (await this.hasDil()) {
       return this.dilPath;
     }
-    const compiledQuery = path.join(this.querySaveDir, 'compiledQuery.qlo');
+    const compiledQuery = this.compileQueryPath;
     if (!(await fs.pathExists(compiledQuery))) {
       if (await cliServer.cliConstraints.supportsNewQueryServer()) {
         // This could be from the new query server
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -79,9 +79,6 @@ export class QueryEvaluationInfo {
     return path.join(this.querySaveDir, 'results.dil');
   }
 
-  /**
-   * Get the path that the compiled query is if it exists. Note that it only exists when using the legacy query server.
```
alexet, line 83: Why did you remove this comment?

alexet (approved):
Github seems confused. I miised that this was against my PR originally sorry. Other than the comment which I think is still true this LGTM.

extensions/ql-vscode/src/run-queries-shared.ts
aeisenberg: I can add it back. I had originally thought that the new query server would be working similarly to the old one and create a qlo file, but I understand now that it will create the dil file directly.

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -79,9 +79,6 @@ export class QueryEvaluationInfo {
...
-   */
   get compileQueryPath() {
```
aeisenberg, line 82:
```suggestion
  /**
   * Get the path that the compiled query is if it exists. Note that it only exists when using the legacy query server.
   */
    get compileQueryPath() {
```

PR #1628, opened 2022-10-19
Move variant analysis tests out of cli-integration directory

elenatanasoiu:
We've added some tests to the cli-integration folder which don't require
the CLI to be run successfully.  Let's move them in the no-workspace
folder.

The three files are:
- `cli-integration/remote-queries/variant-analysis-manager.test.ts`
- `cli-integration/remote-queries/variant-analysis-monitor.test.ts`
- `cli-integration/remote-queries/variant-analysis-results-manager.test`

We're also moving the `variant-analysis-process.test.ts` file into
`tests/pure-tests/remote-queries` as the processor contains helper methods
and doesn't depend on the CLI. 

While we're here, we're also documenting what the cli-integration tests are for 
since the name is confusing.

## Checklist


elenatanasoiu:
The CI failure is not related to this PR and fails on main as well:
```
1) "before each" hook: beforeEach for "should avoid re-downloading an analysis result"
```

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-results-manager.test.ts
```diff
@@ -10,7 +10,7 @@ import * as path from 'path';
 import { VariantAnalysisResultsManager } from '../../../remote-queries/variant-analysis-results-manager';
 import { createMockVariantAnalysisRepoTask } from '../../factories/remote-queries/gh-api/variant-analysis-repo-task';
 import { CodeQLCliServer } from '../../../cli';
-import { storagePath } from '../global.helper';
+import { storagePath } from '../../cli-integration/global.helper';
```
robertbrignull, line 13: Should we still be relying on this file from the `cli-integration` directory, or should that file be moved to be somewhere more generic? Although I'm also aware we don't want to bloat the PR unless necessary.

CONTRIBUTING.md
```diff
@@ -91,29 +91,58 @@ Alternatively, you can start Storybook inside of VSCode. There is a VSCode launc
...
+
+The CLI integration tests require an instance of the CodeQL CLI to run so they will require some extra setup steps. When adding new tests to our test suite, please be mindful of whether they need to be in the cli-integration folder as that requires extra setup when running. If the tests don't depend on the CLI, they are better suited to being a VSCode integration test.
+
+#### Running the tests
+
+##### 1. From the terminal
+
+First move into the `extensions/ql-vscode` directory. Then, depending on which tests you want to run, use the appropriate command to run the tests:
+
+* Unit tests: `npm run test`
```
robertbrignull, line 111: Should this be `test:unit`?
```suggestion
* Unit tests: `npm run test:unit`
```

The plain `test` script is defined as `"test": "npm-run-all -p test:*"` so that would run all the tests and take a long time.

robertbrignull (commented):
Thanks for adding more documentation and cleaning the tests up!  ❤️

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-results-manager.test.ts
elenatanasoiu: There are still three files in the cli-integration folder that depend on that helper: 

- `extensions/ql-vscode/src/vscode-tests/cli-integration/databases.test.ts`
- `extensions/ql-vscode/src/vscode-tests/cli-integration/new-query.test.ts`
- `extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts`

CONTRIBUTING.md
elenatanasoiu: Updated! Thanks!

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-results-manager.test.ts
koesie10: I don't think the `storagePath` is set outside of the `cli-integration` tests (it's set in a helper that is only loaded by the `cli-integration` tests), so this would be `undefined` in these tests. Where are the results actually stored when running these tests now?

elenatanasoiu:
Taking a look at this, but running the `no-workspace` tests via VSCode while switching branches is proving to be difficult. It seems to keep the state of the previous branch. 

Running the tests via the terminal works fine. 

I'm less convinced we should be telling people in the `README.md` to try to run the no-workspace tests from VSCode.

elenatanasoiu:
Have rebased to account for https://github.com/github/vscode-codeql/pull/1631

elenatanasoiu:
This is proving a bit more complicated so I've pulled out the README change into a small PR: https://github.com/github/vscode-codeql/pull/1635

elenatanasoiu:
Thanks for the reviews @koesie10 and @robertbrignull. I've had to change the way we mock some things in these tests now that they're in the `no-workspace` folder, including the storage path. 

More details in the commit message here: https://github.com/github/vscode-codeql/pull/1628/commits/ca3bf049a48aca4318d0c83583f8ce8912d47c27

This is now ready for review again.

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-results-manager.test.ts
elenatanasoiu: It was actually failing even earlier, when it was trying to mock the extension. I've had to change the way we stub things (including the storage path): https://github.com/github/vscode-codeql/commit/ca3bf049a48aca4318d0c83583f8ce8912d47c27 and https://github.com/github/vscode-codeql/pull/1628/commits/37bd51cd9dbdff602af90cdc5632137dc91f3301.

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-manager.test.ts
```diff
@@ -44,9 +44,21 @@ describe('Variant Analysis Manager', async function() {
...
+      cli = createMockCliServer({
+        bqrsInfo: [{ 'result-sets': [{ name: 'result-set-1' }, { name: 'result-set-2' }] }],
+        bqrsDecode: [{
+          columns: [{ kind: 'NotString' }, { kind: 'String' }],
+          tuples: [['a', 'b'], ['c', 'd']],
+          next: 1
+        }, {
+          columns: [{ kind: 'String' }, { kind: 'NotString' }, { kind: 'StillNotString' }],
+          tuples: [['a', 'b', 'c']]
+        }]
```
koesie10, line 57: It seems like these mock operations are not actually used in these tests. Would it make sense to switch the order of the arguments and make the `mockOperations` optional, so that this can be simplified to `cli = createMockCliServer(sandbox)`?

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-monitor.test.ts
```diff
@@ -18,35 +18,55 @@ import { processFailureReason } from '../../../remote-queries/variant-analysis-p
...
+          next: 1
+        }, {
+          columns: [{ kind: 'String' }, { kind: 'NotString' }, { kind: 'StillNotString' }],
+          tuples: [['a', 'b', 'c']]
+        }]
+      }, sandbox);
+      const storagePath = path.join(__dirname);
+
+      variantAnalysisMonitor = new VariantAnalysisMonitor(ctx);
+      variantAnalysisManager = new VariantAnalysisManager(ctx, cli, storagePath, logger);
```
koesie10, line 65: I'm not sure what exactly we're testing on this instance. The `VariantAnalysisMonitor` doesn't have a reference to the `VariantAnalysisManager` at all, nor to this specific `variantAnalysisManager`. We only have expectations on `expect(mockGetDownloadResult).to.not.have.been.called` and don't have the equivalent for `expect(mockGetDownloadResult).to.have.been.called` so wouldn't notice if it doesn't call this at all. Would it make sense to completely remove this and the expectations?

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-results-manager.test.ts
```diff
@@ -33,8 +32,18 @@ describe(VariantAnalysisResultsManager.name, function() {
...
+        bqrsDecode: [{
+          columns: [{ kind: 'NotString' }, { kind: 'String' }],
+          tuples: [['a', 'b'], ['c', 'd']],
+          next: 1
+        }, {
+          columns: [{ kind: 'String' }, { kind: 'NotString' }, { kind: 'StillNotString' }],
+          tuples: [['a', 'b', 'c']]
+        }]
+      }, sandbox);
+      storagePath = path.join(__dirname);
```
koesie10, line 46: The `path.join` doesn't really seem to do anything here. Can we just use `__dirname` or even an empty storage path if it's not going to be used anyway?

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-manager.test.ts
```diff
@@ -44,9 +44,21 @@ describe('Variant Analysis Manager', async function() {
     variantAnalysis = createMockApiResponse('in_progress', scannedRepos);
 
     try {
-      const extension = await extensions.getExtension<CodeQLExtensionInterface | Record<string, never>>('GitHub.vscode-codeql')!.activate();
-      cli = extension.cliServer;
-      variantAnalysisManager = new VariantAnalysisManager(extension.ctx, cli, storagePath, logger);
+      const ctx = createMockExtensionContext();
+      cli = createMockCliServer({
```
robertbrignull, line 48: Sorry if this is a silly question, but if we're having to provide a mock CLI server, does that indicate that these tests should be staying in the CLI-integration test suite?

koesie10: I don't think so, but @elenatanasoiu please correct me if I'm wrong. The `cli-integration` tests will also actually call the cli, but in these specific tests we're not actually calling the cli. The only place where the cli would be called from these tests is if we're decoding BQRS results, but that isn't something we're testing from these tests.

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-manager.test.ts
elenatanasoiu: I've changed this to make the `mockOperations` argument optional: https://github.com/github/vscode-codeql/pull/1628/commits/4341354c45447f0a9996f9bdaf151e4287c61ab8 & https://github.com/github/vscode-codeql/pull/1628/commits/91523bdadc8423f63b4382bd01f3a574ac1eb630

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-monitor.test.ts
elenatanasoiu: Yeah, in retrospect keeping the setup for the CLI and variant analysis manager was wishful thinking (e.g. could we still use these in the future for integration tests?). But at the moment they're not used properly so I've removed them: https://github.com/github/vscode-codeql/pull/1628/commits/ad19d0d06de2217c7df7bef33a77acd2fc117a7a

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-results-manager.test.ts
elenatanasoiu: I've realized I forgot to move the dummy zip file we use for these tests into the `no-workspace` folder. I've now moved that file into `no-workspace/data/remote-queries` and set this dummy storage path to point to that folder for consistency: https://github.com/github/vscode-codeql/pull/1628/commits/da011c12b311811e9e3fb17a9b875749bfffc7e7 & https://github.com/github/vscode-codeql/pull/1628#:~:text=Set%20storagePath%20to%20point%20at%20data/remote%2Dqueries

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-manager.test.ts
elenatanasoiu: Yeah, as @koesie said, we don't actually need the CLI for our tests. However, both the `variantAnalysisManager` and `variantAnalysisResultsManager` tests require some sort of CLI instance when they're first initialised so we need a dummy version of it without actually using it.

robertbrignull:
Just a note I'll leave the remaining review of this up to @koesie10. But do let me know if you'd like me to actively review.

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/factories/cli.ts
```diff
@@ -0,0 +1,29 @@
+import { CodeQLCliServer } from '../../../../cli';
+
+export function createMockCliServer(
+  sandbox: sinon.SinonSandbox,
+  mockOperations: Record<string, any[]> = {}
+): CodeQLCliServer {
+  if (mockOperations == {}) {
```
koesie10, line 7: I don't think this will work since objects are reference types, so even `{} == {}` will be `false`. Instead, you can add the defaults as part of the function arguments or make the default `undefined` (although then you will also need to change the argument type).

However, I don't think we are even using these operations. So, would it also work to simply have an empty default (i.e. no operations are mocked)?

extensions/ql-vscode/src/vscode-tests/no-workspace/remote-queries/variant-analysis-manager.test.ts
```diff
@@ -44,9 +44,11 @@ describe('Variant Analysis Manager', async function() {
     variantAnalysis = createMockApiResponse('in_progress', scannedRepos);
 
     try {
-      const extension = await extensions.getExtension<CodeQLExtensionInterface | Record<string, never>>('GitHub.vscode-codeql')!.activate();
-      cli = extension.cliServer;
-      variantAnalysisManager = new VariantAnalysisManager(extension.ctx, cli, storagePath, logger);
+      const ctx = createMockExtensionContext();
+      cli = createMockCliServer(sandbox);
+      const storagePath = path.join(__dirname);
```
koesie10, line 49: Do we need this storage path to be set to an existing path or can we just use an empty string instead? I don't think we're actually storing any results in these tests

charisk:
It looks like no-workspace means mocking out the extension, which from my chat with @elenatanasoiu is not exactly what we want for these tests, so moving to minimal-workspace might make more sense.

elenatanasoiu:
Thanks @charisk. While pairing together, we realized that in order to be able to write some more tests for the monitor and add more integration tests to the query history, we are going to need to be able to listen to API calls. 

Unfortunately, moving the tests to the `no-workspace` folder means we are using a very basic version of the extension object which doesn't allow us to trigger commands. This in turn means we don't get as far as triggering API calls from those commands so it limits our ability [to write tests](https://github.com/github/vscode-codeql/pull/1599#issuecomment-1281027345) (especially integration tests that check how multiple components interact).

We decided that moving these tests into the `minimal-workspace` folder is a good compromise since we can make use of 
[a real extension object](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/vscode-tests/minimal-workspace/activation.test.ts#L8) we have set up there instead of trying to make this dummy one work. 

I'm going to close this PR and open a new one.

PR #1629, opened 2022-10-19
Initial setup around GitHub API mock server

charisk:
Adds the msw package and some initial structure.

## Checklist
N/A - internal only.

### DIFF
--- a/extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
+++ b/extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
@@ -1,0 +1,19 @@
+export function startServer(): void {
+  // TODO: Enable HTTP interception.
+}
+
+export function stopServer(): void {
+  // Disable HTTP interception.
+}
+
+export function loadScenario(): void {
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
```diff
@@ -0,0 +1,19 @@
+export function startServer(): void {
```
koesie10, line 1: I'm fine with merging it like this, but do you think it would make sense to create a class for this instead so we can more easily track state (and potentially test it)? For example, it would be really useful to keep track of whether the server is listening.

koesie10 (approved):

extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
charisk: Yeah might as well do it now (I thought it'll likely be a class but didn't want to do it before we had to)!

koesie10 (approved):

PR #1631, opened 2022-10-19
Create `timestamp` file in variant analysis results directory

shati-patel:
(Split out from #1622)
- Makes the result storage directory accessible from the variant analysis manager
- Creates a `timestamp` file in that directory

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
+++ b/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
@@ -280,6 +280,8 @@ export async function runRemoteQuery(
 
       variantAnalysisManager.onVariantAnalysisSubmitted(processedVariantAnalysis);
 
+      await variantAnalysisManager.prepareStorageDirectory(processedVariantAnalysis.id);
+
       void logger.log(`Variant analysis:\n${JSON.stringify(processedVariantAnalysis, null, 2)}`);
 
       void showAndLogInformationMessage(`Variant analysis ${processedVariantAnalysis.query.name} submitted for processing`);
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
@@ -176,6 +176,10 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
     return this.queue.pending;
   }
 
+  public getVariantAnalysisStorageLocation(variantAnalysisId: number): string {
+    return this.variantAnalysisResultsManager.getStorageDirectory(variantAnalysisId);
+  }
+
   public async promptOpenVariantAnalysis() {
     const credentials = await Credentials.initialize(this.ctx);
     if (!credentials) { throw Error('Error authenticating with GitHub'); }
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
@@ -23,6 +23,7 @@ import { CodeQLCliServer } from '../cli';
 import { getControllerRepo } from './run-remote-query';
 import { processUpdatedVariantAnalysis } from './variant-analysis-processor';
 import PQueue from 'p-queue';
+import { createTimestampFile } from '../helpers';
 
 export class VariantAnalysisManager extends DisposableObject implements VariantAnalysisViewManager<VariantAnalysisView> {
   private readonly _onVariantAnalysisAdded = this.push(new EventEmitter<VariantAnalysis>());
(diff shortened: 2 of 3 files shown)


extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -176,6 +177,20 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
...
+    return this.variantAnalysisResultsManager.getStorageDirectory(variantAnalysisId);
+  }
+
+  /**
+   * Prepares a directory for storing results for a variant analysis.
+   * This directory contains a timestamp file, which will be
+   * used by the query history manager to determine when the directory
+   * should be deleted.
+   */
+  public async prepareStorageDirectory(variantAnalysisId: number): Promise<void> {
```
shati-patel, line 190: Follows the example of the remote queries manager:
https://github.com/github/vscode-codeql/blob/424520613e4d20526255e71514debb57fe553ede/extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts#L285-L287

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -280,6 +280,8 @@ export async function runRemoteQuery(
 
       variantAnalysisManager.onVariantAnalysisSubmitted(processedVariantAnalysis);
 
+      await variantAnalysisManager.prepareStorageDirectory(processedVariantAnalysis.id);
```
shati-patel, line 283: I wasn't sure where to do this 🤔 Just after the variant analysis is submitted seemed like a good place, but I'm open to suggestions!

charisk: I think that's a good time to do this. However, I think it'd be nicer if we do inside `variantAnalysisManager.onVariantAnalysisSubmitted()`

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -176,6 +177,20 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
     return this.queue.pending;
   }
 
+  public getVariantAnalysisStorageLocation(variantAnalysisId: number): string {
+    return this.variantAnalysisResultsManager.getStorageDirectory(variantAnalysisId);
```
charisk, line 181: Sorry I wasn't very clear with my comment in the other PR 🐱  

In my opinion, the responsibility and knowledge around where data is stored for a variant analysis, should live in the `VariantAnalysisManager` (not the `VariantAnalysisResultsManager`). If `VariantAnalysisResultsManager` needs it as well, it should be passed down to it in the relevant functions.

My reasoning for this is that we don't just store results, we store any information relevant to a variant analysis. Also, ideally the rest of the extension (e.g. query history) should interact with the variant analysis domain via the `VariantAnalysisManager` (which you've covered here! I'm just calling out how I think information should flow)

robertbrignull: So both the `VariantAnalysisManager` and `VariantAnalysisResultsManager` store the `storagePath` as a field? The results manager still needs access to that value, unless it retains a pointer back to the `VariantAnalysisManager`, and the circular dependency would seem like a bigger smell to me.

Ideally neither of these fields would be public, if we can expose methods like `getVariantAnalysisStorageLocation` that return the data that outside callers actually want.

If I've understood correctly then that approach sounds good to me.

charisk: > So both the VariantAnalysisManager and VariantAnalysisResultsManager store the storagePath as a field?

Nope, just `VariantAnalysisManager` will have storagePath (and knowledge about how exactly to store variant analysis information). It will then pass that information to the `VariantAnalysisResultsManager` whenever it needs it. See https://github.com/github/vscode-codeql/pull/1631/commits/43de90f03d832ecae49906c7a1ca8d38970cad5e.

Does that make sense?

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
shati-patel: Done in https://github.com/github/vscode-codeql/pull/1631/commits/6b7ebf543c903da865b84a03fd724a2fb0d5e886!

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-results-manager.test.ts
```diff
@@ -15,7 +15,9 @@ import { faker } from '@faker-js/faker';
 import * as ghApiClient from '../../../remote-queries/gh-api/gh-api-client';
 import { VariantAnalysisRepoTask } from '../../../remote-queries/gh-api/variant-analysis';
 
-describe(VariantAnalysisResultsManager.name, () => {
+describe(VariantAnalysisResultsManager.name, function() {
+  this.timeout(10000);
```
charisk, line 19: Let's revisit this timeout once the tests move out of cli-integration.

charisk (approved):
LGTM!

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
robertbrignull: Ok, I see what you mean now. That looks fine. 👍

PR #1634, opened 2022-10-19
Add recording of mock scenarios

koesie10:
This adds three commands for recording mock scenarios:
- "CodeQL Mock GitHub API Server: Start Scenario Recording"
- "CodeQL Mock GitHub API Server: Save Scenario"
- "CodeQL Mock GitHub API Server: Cancel Scenario"

The first one will only be visible when the setting is enabled and you are not yet recording. The latter two will only be visible when recording. On save, it will ask for a scenario save path and a name and will write all files to this directory.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -645,6 +645,18 @@
         "command": "codeQL.gotoQL",
         "title": "CodeQL: Go to QL Code",
         "enablement": "codeql.hasQLSource"
+      },
+      {
+        "command": "codeQL.mockGitHubApiServer.startRecording",
+        "title": "CodeQL Mock GitHub API Server: Start Scenario Recording"
+      },
+      {
+        "command": "codeQL.mockGitHubApiServer.saveScenario",
+        "title": "CodeQL Mock GitHub API Server: Save Scenario"
+      },
+      {
+        "command": "codeQL.mockGitHubApiServer.cancelRecording",
+        "title": "CodeQL Mock GitHub API Server: Cancel Scenario"
       }
     ],
     "menus": {
@@ -1104,6 +1116,18 @@
         {
           "command": "codeQLTests.showOutputDifferences",
           "when": "false"
+        },
+        {
+          "command": "codeQL.mockGitHubApiServer.startRecording",
+          "when": "config.codeQL.variantAnalysis.mockGitHubApiServer && !codeQL.mockGitHubApiServer.recording"
+        },
+        {
+          "command": "codeQL.mockGitHubApiServer.saveScenario",
+          "when": "codeQL.mockGitHubApiServer.recording"
+        },
+        {
+          "command": "codeQL.mockGitHubApiServer.cancelRecording",
+          "when": "codeQL.mockGitHubApiServer.recording"
         }
       ],
       "editor/context": [
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -656,7 +656,7 @@
       },
       {
         "command": "codeQLMockGitHubApiServer.cancelRecording",
-        "title": "CodeQL Mock GitHub API Server: Cancel Scenario"
+        "title": "CodeQL Mock GitHub API Server: Cancel Scenario Recording"
       }
     ],
     "menus": {
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -647,15 +647,15 @@
         "enablement": "codeql.hasQLSource"
       },
       {
-        "command": "codeQL.mockGitHubApiServer.startRecording",
+        "command": "codeQLMockGitHubApiServer.startRecording",
         "title": "CodeQL Mock GitHub API Server: Start Scenario Recording"
       },
       {
-        "command": "codeQL.mockGitHubApiServer.saveScenario",
+        "command": "codeQLMockGitHubApiServer.saveScenario",
         "title": "CodeQL Mock GitHub API Server: Save Scenario"
       },
       {
-        "command": "codeQL.mockGitHubApiServer.cancelRecording",
+        "command": "codeQLMockGitHubApiServer.cancelRecording",
         "title": "CodeQL Mock GitHub API Server: Cancel Scenario"
       }
     ],
@@ -1118,16 +1118,16 @@
           "when": "false"
         },
         {
-          "command": "codeQL.mockGitHubApiServer.startRecording",
-          "when": "config.codeQL.variantAnalysis.mockGitHubApiServer && !codeQL.mockGitHubApiServer.recording"
+          "command": "codeQLMockGitHubApiServer.startRecording",
+          "when": "config.codeQL.variantAnalysis.mockGitHubApiServer && !codeQLMockGitHubApiServer.recording"
         },
         {
-          "command": "codeQL.mockGitHubApiServer.saveScenario",
-          "when": "codeQL.mockGitHubApiServer.recording"
+          "command": "codeQLMockGitHubApiServer.saveScenario",
+          "when": "codeQLMockGitHubApiServer.recording"
         },
         {
-          "command": "codeQL.mockGitHubApiServer.cancelRecording",
-          "when": "codeQL.mockGitHubApiServer.recording"
+          "command": "codeQLMockGitHubApiServer.cancelRecording",
+          "when": "codeQLMockGitHubApiServer.recording"
         }
       ],
       "editor/context": [
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -1123,11 +1123,11 @@
         },
         {
           "command": "codeQLMockGitHubApiServer.saveScenario",
-          "when": "codeQLMockGitHubApiServer.recording"
+          "when": "config.codeQL.variantAnalysis.mockGitHubApiServer && codeQLMockGitHubApiServer.recording"
         },
         {
           "command": "codeQLMockGitHubApiServer.cancelRecording",
-          "when": "codeQLMockGitHubApiServer.recording"
+          "when": "config.codeQL.variantAnalysis.mockGitHubApiServer && codeQLMockGitHubApiServer.recording"
         }
       ],
       "editor/context": [
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -647,16 +647,16 @@
         "enablement": "codeql.hasQLSource"
       },
       {
-        "command": "codeQLMockGitHubApiServer.startRecording",
-        "title": "CodeQL Mock GitHub API Server: Start Scenario Recording"
+        "command": "codeQL.mockGitHubApiServer.startRecording",
+        "title": "CodeQL: Mock GitHub API Server: Start Scenario Recording"
       },
       {
-        "command": "codeQLMockGitHubApiServer.saveScenario",
-        "title": "CodeQL Mock GitHub API Server: Save Scenario"
+        "command": "codeQL.mockGitHubApiServer.saveScenario",
+        "title": "CodeQL: Mock GitHub API Server: Save Scenario"
       },
       {
-        "command": "codeQLMockGitHubApiServer.cancelRecording",
-        "title": "CodeQL Mock GitHub API Server: Cancel Scenario Recording"
+        "command": "codeQL.mockGitHubApiServer.cancelRecording",
+        "title": "CodeQL: Mock GitHub API Server: Cancel Scenario Recording"
       }
     ],
     "menus": {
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
```diff
@@ -33,8 +49,98 @@ export class MockGitHubApiServer {
...
+    await window.showInformationMessage('Recording scenario. To save the scenario, use the "CodeQL Mock GitHub API Server: Save Scenario" command.');
+  }
+
+  public async saveScenario(): Promise<void> {
+    const scenariosDirectory = await this.getScenariosDirectory();
+    if (!scenariosDirectory) {
+      return;
+    }
+
+    await commands.executeCommand('setContext', 'codeQL.mockGitHubApiServer.recording', false);
```
charisk, line 70: Let's add a comment to explain why we're doing this as it's not immediately obvious. E.g. `set a value in the context to track whether we're recording so we can decide whether commands around recording should be visible`

extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
```diff
@@ -33,8 +49,98 @@ export class MockGitHubApiServer {
...
+    const name = await window.showInputBox({
+      title: 'Save scenario',
+      prompt: 'Enter a name for the scenario.',
+      placeHolder: 'successful-run',
+    });
+    if (!name) {
+      return;
+    }
+
+    const filepath = await this.recorder.save(scenariosDirectory, name);
```
charisk, line 93: Minor `filepath` -> `filePath`

extensions/ql-vscode/src/mocks/recorder.ts
```diff
@@ -0,0 +1,176 @@
...
+
+import { DisposableObject } from '../pure/disposable-object';
+
+import { GitHubApiRequest, RequestKind } from './gh-api-request';
+
+export class Recorder extends DisposableObject {
+  private readonly allRequests = new Map<string, MockedRequest>();
+  private currentRecordedScenario: GitHubApiRequest[] = [];
+
+  private _recording = false;
```
charisk, line 16: This is a personal preference thing so feel free to ignore (although I'd like it to be a linting rule at some point): I think we should avoid prefixing variables with `_`. I don't think it really brings anything other than making people think the variable is special in some way.

extensions/ql-vscode/src/mocks/recorder.ts
```diff
@@ -0,0 +1,176 @@
...
+    super();
+    this.onRequestStart = this.onRequestStart.bind(this);
+    this.onResponseBypass = this.onResponseBypass.bind(this);
+  }
+
+  public get isRecording(): boolean {
+    return this._recording;
+  }
+
+  public get scenarioRequestCount(): number {
```
charisk, line 30: (Optional) Would it make sense to instead expose something like `anyRequestsRecorded` instead?

extensions/ql-vscode/src/mocks/recorder.ts
```diff
@@ -0,0 +1,176 @@
...
+    this.server.events.removeListener('request:start', this.onRequestStart);
+    this.server.events.removeListener('response:bypass', this.onResponseBypass);
+  }
+
+  public clear() {
+    this.currentRecordedScenario = [];
+    this.allRequests.clear();
+  }
+
+  public async save(scenariosDirectory: string, name: string): Promise<string> {
```
charisk, line 63: How about calling it `scenariosPath` instead of `scenariosDirectory`?

extensions/ql-vscode/package.json
```diff
@@ -645,6 +645,18 @@
...
+        "command": "codeQL.mockGitHubApiServer.startRecording",
+        "title": "CodeQL Mock GitHub API Server: Start Scenario Recording"
+      },
+      {
+        "command": "codeQL.mockGitHubApiServer.saveScenario",
+        "title": "CodeQL Mock GitHub API Server: Save Scenario"
+      },
+      {
+        "command": "codeQL.mockGitHubApiServer.cancelRecording",
+        "title": "CodeQL Mock GitHub API Server: Cancel Scenario"
```
charisk, line 659:
```suggestion
        "title": "CodeQL Mock GitHub API Server: Cancel Scenario Recording"
```

To match the starting command :)

extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
```diff
@@ -33,8 +49,98 @@ export class MockGitHubApiServer {
...
+  public async recordScenario(): Promise<void> {
+    if (this.recorder.isRecording) {
+      void window.showErrorMessage('A scenario is already being recorded. Use the "Save Scenario" or "Cancel Scenario" commands to finish recording.');
+      return;
+    }
+
+    this.recorder.start();
+    await commands.executeCommand('setContext', 'codeQL.mockGitHubApiServer.recording', true);
+
+    await window.showInformationMessage('Recording scenario. To save the scenario, use the "CodeQL Mock GitHub API Server: Save Scenario" command.');
```
charisk, line 61: Does this look better if we add some new lines? I'm a bit surprised it's not multi-line.

image

charisk (commented):
Great work! I'm glad my hacky PoC code didn't confuse you 😹 

Is this working well for you in its current state? I got 

image

This was after I:
- Started recording
- Ran a variant analysis

In terms of code, I think it's generally looking good 👌 most of my comments are pretty minor, except maybe one around the scenarios directory path.

extensions/ql-vscode/src/config.ts
```diff
@@ -449,3 +449,9 @@ export class MockGitHubApiConfigListener extends ConfigListener implements MockG
     return !!MOCK_GH_API_SERVER.getValue<boolean>();
   }
 }
+
+const MOCK_GH_API_SERVER_SCENARIOS_PATH = new Setting('mockGitHubApiServerScenariosPath', REMOTE_QUERIES_SETTING);
```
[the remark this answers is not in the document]
koesie10: Unfortunately, it's really hard to actually hardcode the path since we don't know where the `scenarios` directory is located. We can't really rely on the location of the extension since you can also run it outside of development mode and enable it in the production extension. In that case, we still want you to allow saving the files to the correct scenarios directory.

One thing we might be able to do is to check `ctx.extensionMode === ExtensionMode.Development` and only hardcode the path in that case (perhaps only if the path exists?).

extensions/ql-vscode/src/mocks/recorder.ts
koesie10: I agree with you that in general we shouldn't prefix these with `_` but in this case it's done because we want the variable to be read-only outside of the class, but read-write inside of the class. I think choosing two different names for the public getter and private variable would make things even more confusing, hence why I chose to prefix it like this.

extensions/ql-vscode/src/config.ts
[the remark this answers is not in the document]
charisk: I think it's okay to not provide a supported path outside of development mode. We're not intending to use this functionality outside of dev. 

> One thing we might be able to do is to check ctx.extensionMode === ExtensionMode.Development and only hardcode the path in that case (perhaps only if the path exists?).

This sounds great! I wasn't aware of `extensionMode`. It could even be: try and use hard-coded path and if it doesn't exist, ask for a path to use.

extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
```diff
@@ -33,8 +49,98 @@ export class MockGitHubApiServer {
...
-  public recordScenario(): void {
-    // TODO: Implement logic to record a new scenario to a directory.
+  public async recordScenario(): Promise<void> {
+    if (this.recorder.isRecording) {
+      void window.showErrorMessage('A scenario is already being recorded. Use the "Save Scenario" or "Cancel Scenario" commands to finish recording.');
+      return;
+    }
+
+    this.recorder.start();
+    await commands.executeCommand('setContext', 'codeQL.mockGitHubApiServer.recording', true);
```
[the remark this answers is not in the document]
koesie10: We aren't prefixing the commands or the `recorder`/`mock-gh-api-server` with variant analysis, so I don't think it's really necessary to do so here for consistency.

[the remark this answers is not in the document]
charisk: The config setting is inside variant analysis though https://github.com/github/vscode-codeql/blob/koesie10/record-scenario/extensions/ql-vscode/src/config.ts#L435-L436

extensions/ql-vscode/src/mocks/recorder.ts
charisk: But you have provided two different names - `isRecording` and `_recording`. Apologies if I'm missing something obvious!

extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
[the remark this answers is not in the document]
charisk: I'm leaning towards moving the config out of variant analysis 🤔 thoughts? It feels like we're somewhere between at the minute - would be good to choose to either scope this to variant analysis or not.

extensions/ql-vscode/src/mocks/recorder.ts
koesie10: Sorry about that! I renamed the method but didn't rename the variable. I've now aligned them so they match.

extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
koesie10: It seems like adding newlines doesn't actually change this since the notification is still collapsed. I'm not sure why it's collapsed, it might be because it doesn't have an action button? We could add a "Cancel" button to it just to make sure it's expanded but that feels a bit hacky.

extensions/ql-vscode/src/config.ts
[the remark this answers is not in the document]
koesie10: I've now made the simplest possible change which is using the hard-coded path if it exists and using the logic that was there before otherwise. We can change it to do something else when the path doesn't exist in a follow-up PR if necessary.

koesie10:
> Is this working well for you in its current state? I got
> 
> image
> 
> This was after I:
> 
> * Started recording
> * Ran a variant analysis

I tried it before I rebased and it worked fine then. After that, I changed it to assume the server was started when the config setting was set since we have a config listener. However, it seems like the server didn't get started when the config setting was already set, so https://github.com/github/vscode-codeql/pull/1634/commits/74ae5a7fdc55340887088a6566079b44226f85c2 should fix that and it should now work properly.

extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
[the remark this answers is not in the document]
charisk: Ok let's go with that then! Feel free to leave things as is in your PR and we'll have another issue/PR to tidy up.

extensions/ql-vscode/package.json
```diff
@@ -647,15 +647,15 @@
         "enablement": "codeql.hasQLSource"
       },
       {
-        "command": "codeQL.mockGitHubApiServer.startRecording",
+        "command": "codeQLMockGitHubApiServer.startRecording",
```
charisk, line 650: I'm not sure I agree with this approach.. I think we still want all the commands provided by our extension to be prefixed with `CodeQL: `. Why have we gone against it?

extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
charisk: That's a shame. You're right that it feels a bit hacky.

extensions/ql-vscode/src/mocks/recorder.ts
```diff
@@ -0,0 +1,176 @@
...
+  const repoDownloadMatch = url.match(/objects-origin\.githubusercontent\.com\/codeql-query-console\/codeql-variant-analysis-repo-tasks\/\d+\/(?<repositoryId>\d+)/);
+  if (repoDownloadMatch?.groups?.repositoryId) {
+    return {
+      request: {
+        kind: RequestKind.GetVariantAnalysisRepoResult,
+        repositoryId: parseInt(repoDownloadMatch.groups.repositoryId, 10),
+      },
+      response: {
+        status,
+        body: body as unknown as ArrayBuffer,
```
charisk, line 170: Perhaps something to address separately but I'm wondering if should do something a bit nicer and store things in an actual zip/sarif/bqrs file. What do you think?

koesie10: I think that would indeed be very useful because this will result in a JSON string with binary data, which is a really inefficient way of storing this. I left it out of this PR since I think it will be quite some additional work, but I think a follow-up PR to address this would definitely be useful!

extensions/ql-vscode/package.json
charisk: I'd rather we do that if that's okay. Then we know that everything starts with `CodeQL:`.

extensions/ql-vscode/src/mocks/recorder.ts
charisk: I'll create an issue that covers this.

extensions/ql-vscode/src/mocks/mock-gh-api-server.ts
```diff
@@ -33,17 +52,122 @@ export class MockGitHubApiServer {
...
+  }
+
+  private async getScenariosPath(): Promise<string | undefined> {
+    const scenariosPath = getMockGitHubApiServerScenariosPath();
+    if (scenariosPath) {
+      return scenariosPath;
+    }
+
+    if (this.ctx.extensionMode === ExtensionMode.Development) {
+      const developmentScenariosPath = Uri.joinPath(this.ctx.extensionUri, 'src/mocks/scenarios').toString();
```
charisk, line 134: I think you need the `fsPath`:

```suggestion
      const developmentScenariosPath = Uri.joinPath(this.ctx.extensionUri, 'src/mocks/scenarios').fsPath.toString();
```

charisk (approved):
🚀

Issue #1489, opened 2022-08-31
Restarting vscode while the extension is starting up can remove databases.


alexet:
**Describe the bug**
Restarting the extension before all databases have been reloaded means we lose the databases that haven't been loaded yet. 

**Version**

CodeQL extension version: 1.6.11 
CodeQL CLI version: 2.10.3 
Platform: linux x64

**To reproduce**

- Start vscode with the extension and add some large databases.
- Restart vscode.
- Close it while some but not all databases have been loaded.
- Reopen it and notice that the databases are missing.

**Expected behavior**

The databases that weren't loaded were loaded the next time.

PR #1638, opened 2022-10-20
Fix bug where dbs are lost on restart

aeisenberg:
If the workspace is restarted while databases are being loaded, this change prevents any from being lost.

The bug was that each time a database was added when rehydrating a db from persisted state on startup, the persisted db list was being updated. Instead of updating the list each time we add a db, on restart, instead update the persisted list only after all are added.

Note that we need to update the persisted list after reading it in since the act of rehydrating a database _may_ change its persisted state. For example, the primary language of the database may be initialized if it was not able to be determined originally.

Fixes https://github.com/github/vscode-codeql/issues/1489

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Fix a bug where persisted databases may be lost if VS code is restarted while databases are being initialized. [#1638](https://github.com/github/vscode-codeql/pull/1638)
+
 ## 1.7.2 - 14 October 2022
 
 - Fix a bug where results created in older versions were thought to be unsuccessful. [#1605](https://github.com/github/vscode-codeql/pull/1605)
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Fix a bug where persisted databases may be lost if VS code is restarted while databases are being initialized. [#1638](https://github.com/github/vscode-codeql/pull/1638)
```
edoardopirovano, line 5:
```suggestion
- Fix a bug where databases may be lost if VS Code is restarted while the extension is being started up. [#1638](https://github.com/github/vscode-codeql/pull/1638)
```

I think this describes the impact in a way the user can understand a bit better as the average user may not know when databases are being initialized.

aeisenberg: Thanks. That sounds better.

edoardopirovano (approved):

PR #1639, opened 2022-10-21
Implement query history label for variant analysis items

shati-patel:
Display the number of repositories that a variant analysis has been run on:
image

There are a few tangential changes: https://github.com/github/vscode-codeql/pull/1639/commits/add5417a422ec767e6ee475c68c33472542d7076 moves the `pluralize` method to the "pure" helpers file, so that we can avoid importing `vscode` in tests.

## Checklist

N/A - internal only 🦎 


extensions/ql-vscode/src/query-history-info.ts
```diff
@@ -43,3 +44,22 @@ export function getQueryText(item: QueryHistoryInfo): string {
...
+  if (item.t === 'remote') {
+      // Return the number of repositories queried if available. Otherwise, use the controller repository name.
+      const repositoryCount = item.remoteQuery.repositoryCount;
+  
+      if (repositoryCount) {
+        return pluralize(repositoryCount, 'repository', 'repositories');
+      }
+      return `${item.remoteQuery.controllerRepository.owner}/${item.remoteQuery.controllerRepository.name}`;
+    } else if (item.t === 'variant-analysis') {
+      const totalScannedRepositoryCount = item.variantAnalysis.scannedRepos?.length ?? 0;
```
shati-patel, line 58: Note: When we first trigger the variant analysis, `scannedRepos` might be undefined so we currently show `0` in this case. It looks slightly weird to display `0/0 repositories`, but this should update as soon as we start monitoring/updating the variant analysis.

image


------

Additional details, in case people are interested 😅 

Other options we considered for when `scannedRepos` is undefined:
- Controller repo name. The variant analysis only saves the controller repo ID, so we'd have to do some re-jigging to get the name.
- Something generic like `remote repositories`. This could work, but might not be any less confusing than `0/0 repositories`.
- Empty string. Annoyingly, the `on` is hard-coded in the label ([here](https://github.com/github/vscode-codeql/blob/4e5abee2ea2dcd437c0c58c844abca77dfd99be3/extensions/ql-vscode/package.json#L236)), so it would be some effort to adjust this and make the label make sense in all cases.

Other thoughts/suggestions welcome of course!

extensions/ql-vscode/src/pure/helpers-pure.ts
```diff
@@ -55,3 +55,11 @@ export function getErrorStack(e: any) {
 export function asError(e: any): Error {
   return e instanceof Error ? e : new Error(String(e));
 }
+
+/**
+ * Pluralizes a word.
+ * Example: Returns "N repository" if N is one, "N repositories" otherwise.
+ */
+export function pluralize(numItems: number | undefined, singular: string, plural: string): string {
```
charisk, line 63: `helpers-pure` is a generic bucket that I'd personally like to see us gradually move away from. How about we move this function to a new file/module e.g. `word`?

extensions/ql-vscode/test/pure-tests/query-history-info.test.ts
```diff
@@ -80,4 +90,24 @@ describe('Query history info', () => {
       expect(queryText).to.equal(variantAnalysisHistoryItem.variantAnalysis.query.text);
     });
   });
+
+  describe('buildRepoLabel', () => {
+    it('should build the repo label for remote query history items', () => {
```
charisk, line 95: Shall we cover a few more cases? E.g. remote query with no repo count, variant analysis with 1 repository.

extensions/ql-vscode/src/query-history-info.ts
charisk: Seems like something we should run past Product/Design or stick with the same approach that we have for remote queries (which I think is your first bullet point). I don't think we should block the PR though as it's taking as a step further.

extensions/ql-vscode/src/helpers.ts
```diff
@@ -590,5 +590,5 @@ export async function* walkDirectory(dir: string): AsyncIterableIterator<string>
  * Example: Returns "N repository" if N is one, "N repositories" otherwise.
  */
 export function pluralize(numItems: number | undefined, singular: string, plural: string): string {
```
charisk: This function could do with some unit tests. They'd nicely document the expected behaviour too.

Also I if `numItems` is 0, then this will return e.g. `0 repository`. Is this definitely what we want (sounds a bit off grammatically to me)? Previously we'd return an empty string.

charisk (commented):
Looking good! Thanks for tidying up the helper utils too.

extensions/ql-vscode/src/helpers.ts
shati-patel: It should return `0 repositories`, which sounds okay to me? 🤔  But you're right that it could do with some tests! Incoming...

robertbrignull: We deliberately changed it from the empty string because it resulted in a label like `0/` instead of `0/0 repositories`. I agree with Shati, the implementation looks correct now. But as said, tests will make this more clear.

extensions/ql-vscode/src/pure/helpers-pure.ts
shati-patel: Good point! Moved in https://github.com/github/vscode-codeql/pull/1639/commits/c85338d11a433eb0a69ad4677e66cef54f847f2d 🦞

extensions/ql-vscode/test/pure-tests/query-history-info.test.ts
shati-patel: Added more cases in https://github.com/github/vscode-codeql/pull/1639/commits/593ca574978de92385a8b4dce6e2d6d3b05a507e

extensions/ql-vscode/src/query-history-info.ts
shati-patel: Makes sense! I've left it as-is for now, but happy to make some follow-up changes later 🙇🏽‍♀️

charisk (approved):

PR #1641, opened 2022-10-21
Add loading of mock scenarios

charisk:
Adds a command to load a mock scenario for the mock GitHub API server to replay.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/mocks/request-handlers.ts
+++ b/extensions/ql-vscode/src/mocks/request-handlers.ts
@@ -1,7 +1,14 @@
 import * as path from 'path';
 import * as fs from 'fs-extra';
 import { DefaultBodyType, MockedRequest, rest, RestHandler } from 'msw';
-import { GetVariantAnalysisRepoRequest, GetVariantAnalysisRepoResultRequest, GitHubApiRequest, RequestKind } from './gh-api-request';
+import {
+  GitHubApiRequest,
+  isGetRepoRequest,
+  isGetVariantAnalysisRepoRequest,
+  isGetVariantAnalysisRepoResultRequest,
+  isGetVariantAnalysisRequest,
+  isSubmitVariantAnalysisRequest
+} from './gh-api-request';
 
 const baseUrl = 'https://api.github.com';
 
@@ -41,7 +48,7 @@ async function readRequestFiles(scenarioDirPath: string): Promise<GitHubApiReque
 }
 
 function createGetRepoRequestHandler(requests: GitHubApiRequest[]): RequestHandler {
-  const getRepoRequests = requests.filter(r => r.request.kind === RequestKind.GetRepo);
+  const getRepoRequests = requests.filter(isGetRepoRequest);
 
   if (getRepoRequests.length > 1) {
     throw Error('More than one get repo request found');
@@ -58,7 +65,7 @@ function createGetRepoRequestHandler(requests: GitHubApiRequest[]): RequestHandl
 }
 
 function createSubmitVariantAnalysisRequestHandler(requests: GitHubApiRequest[]): RequestHandler {
-  const submitVariantAnalysisRequests = requests.filter(r => r.request.kind === RequestKind.SubmitVariantAnalysis);
+  const submitVariantAnalysisRequests = requests.filter(isSubmitVariantAnalysisRequest);
 
   if (submitVariantAnalysisRequests.length > 1) {
     throw Error('More than one submit variant analysis request found');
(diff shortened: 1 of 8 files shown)


extensions/ql-vscode/src/pure/files.ts
```diff
@@ -28,3 +28,25 @@ export async function gatherQlFiles(paths: string[]): Promise<[string[], boolean
   }
   return [Array.from(gatheredUris), dirFound];
 }
+
+/**
+ * Lists the names of directories inside the given path.
+ * @param path The path to the directory to read.
+ * @returns the names of the directories inside the given path.
+ */
+export async function getDirectoryNamesInsidePath(path: string): Promise<string[]> {
```
koesie10, line 37: This seems like something we can create a test for. Do you think it would make sense to add a test for this in `files.test.ts`?

extensions/ql-vscode/src/mocks/request-handlers.ts
```diff
@@ -0,0 +1,137 @@
...
+function createGetRepoRequestHandler(requests: GitHubApiRequest[]): RequestHandler {
+  const getRepoRequests = requests.filter(r => r.request.kind === RequestKind.GetRepo);
+
+  if (getRepoRequests.length > 1) {
+    throw Error('More than one get repo request found');
+  }
+
+  const getRepoRequest = getRepoRequests[0];
+
+  return rest.get(`${baseUrl}/repos/*`, (_req, res, ctx) => {
```
koesie10, line 52: I'm worried that this might result in matching too many URLs if we ever change the implementation. Can we make this a more specific match, either using the regex of the recorder or by using something like `${baseUrl}/repos/:owner/:name`?

extensions/ql-vscode/src/mocks/request-handlers.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    return res(
+      ctx.status(request.response.status),
+      ctx.json(request.response.body),
+    );
+  });
+}
+
+function createGetVariantAnalysisRepoRequestHandlers(requests: GitHubApiRequest[]): RequestHandler[] {
+  const getVariantAnalysisRepoRequests = requests.filter(r => r.request.kind === RequestKind.GetVariantAnalysisRepo) as GetVariantAnalysisRepoRequest[];
```
koesie10, line 100: Instead of casting the result, we should also be able to use type guards. See e.g. [this](https://www.skovy.dev/blog/typescript-filter-array-with-type-guard) on how it can be used. I'm also happy with leaving it like this, I don't think it will provide a lot of value here.

extensions/ql-vscode/src/mocks/request-handlers.ts
```diff
@@ -0,0 +1,137 @@
...
+          ctx.status(request.response.status),
+          ctx.json(request.response.body),
+        );
+      });
+
+    handlers.push(handler);
+  }
+
+  return handlers;
+}
```
koesie10, line 117: I think this might be better suited with a `map` call on the `getVariantAnalysisRepoRequests`:

```suggestion
  return getVariantAnalysisRepoRequests.map(request => rest.get(
    `${baseUrl}/repositories/:controllerRepoId/code-scanning/codeql/variant-analyses/:variantAnalysisId/repositories/${request.request.repositoryId}`,
    (_req, res, ctx) => {
      return res(
        ctx.status(request.response.status),
        ctx.json(request.response.body),
      );
    }));
}
```

extensions/ql-vscode/src/mocks/request-handlers.ts
```diff
@@ -0,0 +1,137 @@
...
+
+function createGetVariantAnalysisRepoResultRequestHandlers(requests: GitHubApiRequest[]): RequestHandler[] {
+  const getVariantAnalysisRepoResultRequests = requests.filter(r => r.request.kind === RequestKind.GetVariantAnalysisRepoResult) as GetVariantAnalysisRepoResultRequest[];
+
+  const handlers: RequestHandler[] = [];
+  for (const request of getVariantAnalysisRepoResultRequests) {
+    const handler = rest.get(
+      `https://objects-origin.githubusercontent.com/codeql-query-console/codeql-variant-analysis-repo-tasks/:variantAnalysisId/${request.request.repositoryId}/*`,
```
koesie10, line 125: Do you think there's a way to re-use the regexes between the recorder and these request handlers? It now seems like we could easily change one of the two without changing the other.

koesie10 (commented):
Looks good overall and a simple test worked.

My biggest concern is not being able to stop a scenario. Should we unload the request handlers when we start a new recording, and also have a command to unload them? I'm fine with implementing that in a follow-up PR as well.

extensions/ql-vscode/src/mocks/request-handlers.ts
charisk: Yeah it didn't feel like it was worth it. But perhaps I should just stop being lazy :D

extensions/ql-vscode/src/mocks/request-handlers.ts
charisk: I've kind of been in two minds about this. I'm wondering if we can do something between gh-api-client, scenario recorded and scenario loader. I'll raise an issue to cover it.

extensions/ql-vscode/src/pure/files.ts
charisk: I've added some very simple tests - we could do with something a bit more advanced/extensive but I think we might want to tidy up the test data dirs first.

koesie10 (approved):

PR #1642, opened 2022-10-21
Add linter for scenario files

koesie10:
This adds a linter for JSON scenario files which will validate the JSON files in the scenarios directory against the TypeScript types. It will convert the TypeScript types to JSON schema to simplify this process.

Unfortunately, this will not currently allow adding scenarios with failing requests since the types do not allow this. Rather than removing this validation, we should fix the types. This can be done in a follow-up PR.

## Checklist

### DIFF
--- a/extensions/ql-vscode/scripts/lint-scenarios.ts
+++ b/extensions/ql-vscode/scripts/lint-scenarios.ts
@@ -1,0 +1,79 @@
+import * as fs from 'fs-extra';
+import * as path from 'path';
+
+import Ajv from 'ajv';
+import * as tsj from 'ts-json-schema-generator';
+
+const extensionDirectory = path.resolve(__dirname, '..');
+const rootDirectory = path.resolve(extensionDirectory, '../..');
+const scenariosDirectory = path.resolve(extensionDirectory, 'src/mocks/scenarios');
+
+const debug = process.env.RUNNER_DEBUG || process.argv.includes('--debug');
+
+async function lintScenarios() {
+  const schema = tsj.createGenerator({
+    path: path.resolve(extensionDirectory, 'src/mocks/gh-api-request.ts'),
+    tsconfig: path.resolve(extensionDirectory, 'tsconfig.json'),
+    type: 'GitHubApiRequest',
+    skipTypeCheck: true,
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/scripts/lint-scenarios.ts
```diff
@@ -0,0 +1,79 @@
...
+const scenariosDirectory = path.resolve(extensionDirectory, 'src/mocks/scenarios');
+
+const debug = process.env.RUNNER_DEBUG || process.argv.includes('--debug');
+
+async function lintScenarios() {
+  const schema = tsj.createGenerator({
+    path: path.resolve(extensionDirectory, 'src/mocks/gh-api-request.ts'),
+    tsconfig: path.resolve(extensionDirectory, 'tsconfig.json'),
+    type: 'GitHubApiRequest',
+    skipTypeCheck: true,
```
charisk, line 18: How come we're skipping the type check?

charisk (approved):
Looks great! 

Shall we add something in CONTRIBUTING to explain a bit more what scenarios are etc?

extensions/ql-vscode/scripts/lint-scenarios.ts
koesie10: This will make the linting faster since this will skip the TypeScript type checks. The type checks are already done by other parts of the pipeline (like when building the extension), so I didn't think it was really necessary to do the type checks when doing this linting.

koesie10:
> Shall we add something in CONTRIBUTING to explain a bit more what scenarios are etc?

I assumed that would be part of the documentation issue, which would also explain how to record and load scenarios, etc.

extensions/ql-vscode/scripts/lint-scenarios.ts
charisk: Ah yep, that makes sense!

charisk:
> > Shall we add something in CONTRIBUTING to explain a bit more what scenarios are etc?
> 
> I assumed that would be part of the documentation issue, which would also explain how to record and load scenarios, etc.

I may have forgotten I created that on Friday 😂

PR #1644, opened 2022-10-21
Update mock GitHub API request models to support failures

charisk:
This will allow us to simulate failed requests.

## Checklist
N/A - internal only

### DIFF
--- a/extensions/ql-vscode/src/mocks/gh-api-request.ts
+++ b/extensions/ql-vscode/src/mocks/gh-api-request.ts
@@ -12,13 +12,17 @@ export enum RequestKind {
   GetVariantAnalysisRepoResult = 'getVariantAnalysisRepoResult',
 }
 
+export interface BasicErorResponse {
+  message: string;
+}
+
 export interface GetRepoRequest {
   request: {
     kind: RequestKind.GetRepo
   },
   response: {
     status: number,
-    body: Repository
+    body: Repository | BasicErorResponse | undefined
   }
 }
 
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/mocks/gh-api-request.ts
```diff
@@ -12,13 +12,17 @@ export enum RequestKind {
...
+}
+
 export interface GetRepoRequest {
   request: {
     kind: RequestKind.GetRepo
   },
   response: {
     status: number,
-    body: Repository
+    body: Repository | BasicErorResponse | undefined
```
koesie10, line 25: Instead of using `undefined`, we might want to make the `body` property optional. Those two should be equivalent, but I think that will be clearer when comparing the JSON files to these types since the property will be missing from the JSON files.

```suggestion
    body?: Repository | BasicErorResponse
```

extensions/ql-vscode/src/mocks/gh-api-request.ts
```diff
@@ -28,7 +32,7 @@ export interface SubmitVariantAnalysisRequest {
   },
   response: {
     status: number,
-    body: VariantAnalysis
+    body: VariantAnalysis | BasicErorResponse | undefined
   }
```
koesie10, line 36: I think it would have been really nice to use something like:

```ts
  response: {
    status: 201,
    body: VariantAnalysis
  } | {
    status: number,
    body?: BasicErorResponse
  }
```

However, this doesn't actually seem to be validated properly by #1642, so I don't think it's really worth it.

koesie10 (approved):

extensions/ql-vscode/src/mocks/gh-api-request.ts
charisk: That's a good shout! Made the changes.

PR #1651, opened 2022-10-25
Remove historyItemId for variant analyses

charisk:
The query history item id is only used in [one place](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/query-history.ts#L1114-L1116) and in general we're not relying on uniqueness of query history items.

This PR removes `historyItemId` from variant analysis items. As part of that it also renames `getQueryHistoryItemId` to `getQueryId` as it now seems like a slightly more appropriate name. Open to suggestions though!

## Checklist
N/A - internal only.

### DIFF
--- a/extensions/ql-vscode/src/query-history-info.ts
+++ b/extensions/ql-vscode/src/query-history-info.ts
@@ -21,7 +21,7 @@ export function getRawQueryName(item: QueryHistoryInfo): string {
 }
 
 /**
- * Gets an identifier for the query of history item which could be
+ * Gets an identifier for the query history item which could be
  * a local/remote query or a variant analysis. This id isn't guaranteed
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/query-history-info.ts
```diff
@@ -20,14 +20,21 @@ export function getRawQueryName(item: QueryHistoryInfo): string {
   }
 }
 
-export function getQueryHistoryItemId(item: QueryHistoryInfo): string {
+/**
+ * Gets an identifier for the query of history item which could be
```
shati-patel, line 24: teeny typo? :)
```suggestion
 * Gets an identifier for the query history item which could be
```

shati-patel (approved):
LGTM!

PR #1653, opened 2022-10-25
Allow multiple query history items for a single variant analysis

charisk:
We decided to allow having multiple query history items for the same variant analysis. This PR ensures that all places in the code deal with this possibility. 

## Checklist
N/A - internal only.

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -635,10 +635,10 @@ export class QueryHistoryManager extends DisposableObject {
     });
 
     const variantAnalysisRemovedSubscription = this.variantAnalysisManager.onVariantAnalysisRemoved(async (variantAnalysis) => {
-      const item = this.treeDataProvider.allHistory.find(i => i.t === 'variant-analysis' && i.variantAnalysis.id === variantAnalysis.id);
-      if (item) {
+      const items = this.treeDataProvider.allHistory.filter(i => i.t === 'variant-analysis' && i.variantAnalysis.id === variantAnalysis.id);
+      items.forEach(async (item) => {
         await this.removeRemoteQuery(item as RemoteQueryHistoryItem);
-      }
+      });
     });
 
     this.push(variantAnalysisAddedSubscription);


koesie10 (approved):

extensions/ql-vscode/src/query-history.ts
```diff
@@ -635,10 +635,10 @@ export class QueryHistoryManager extends DisposableObject {
     });
 
     const variantAnalysisRemovedSubscription = this.variantAnalysisManager.onVariantAnalysisRemoved(async (variantAnalysis) => {
-      const item = this.treeDataProvider.allHistory.find(i => i.t === 'variant-analysis' && i.variantAnalysis.id === variantAnalysis.id);
-      if (item) {
+      const items = this.treeDataProvider.allHistory.filter(i => i.t === 'variant-analysis' && i.variantAnalysis.id === variantAnalysis.id);
+      items.forEach(async (item) => {
```
robertbrignull, line 639: Does `forEach` wait for the promise to complete? My searching on the internet implies it doesn't. So here we're firing off promises but then not waiting for them.

To wait for the promises we'd have to do something like:
```typescript
Promise.all(items.map(item => this.removeRemoteQuery(item as RemoteQueryHistoryItem)));
```

Alternatively use a `for (const item in items)` rather than a `forEach`.

robertbrignull: Preface: I don't know how much detail to go into here, so I apologise if this is too much. I'm happy to talk in person too if that would be easier.

---

I have to disagree. `forEach` is not equivalent to a normal `for` loop when it comes to async functions. This will have implications for ordering of operations and for handling of errors.

The ordering of operations is largely unimportant here. We're in an async command here already, and it doesn't matter when `removeRemoteQuery` gets called so long as it is called. The code here does still get executed.

The problem here is errors. Unfortunately `forEach` doesn't do anything with the return value of the function it is given. In the case of an async function the return value is a promise, because that's all async functions are. It's just a function that returns a promise.  And here that promise is being thrown away.

The way that async functions return errors is through the returned promise, and not through exceptions. In terms of error handling:
- `await` is syntactic sugar for "if the promise resolved successfully then return the value, or if the promise resolved unsuccessfully then throw an error",
- `async` means "if the function returns a value then convert that to a successful promise, or if the function threw an exception then convert that to an unsuccessful promise".

The above is great and works well but you have to follow it all the way to the top level of the promise chain. Here we are using `await` in the inner function but then an exception here gets converted to a failed promise, which then gets thrown away by `forEach`. The result of this is that if `removeRemoteQuery` were to fail then the error would be swallowed and not reported.

Something that may may make it easier to see is to note that
```typescript
async (item) => {
  await this.removeRemoteQuery(item as RemoteQueryHistoryItem);
}
```
is equivalent to just
```typescript
this.removeRemoteQuery(item as RemoteQueryHistoryItem)
```
The first version is wrapping it in another promise but it's not changing the behaviour.

---

A final point is that in the [https://github.com/github/codeql-variant-analysis-action](github/codeql-variant-analysis-action) repo we have more eslint rules enabled and this line would actually get caught by two of them. Sadly neither of these rules have great documentation, but it may be helpful to read them. See:
- https://www.npmjs.com/package/eslint-plugin-no-async-foreach
- https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-misused-promises.md

robertbrignull: Post-face, I believe this is how error handling works with promises. If it turns out I'm wrong here I'm very sorry about that and for making this into a big thing.

robertbrignull: @elenatanasoiu already opened an issue for this and other cases. But adding an eslint rule sounds like a great idea. I wonder if we could look at adding https://github.com/github/eslint-plugin-github?

PR #1654, opened 2022-10-25
Implement `openQueryResults` for variant analysis items

shati-patel:
Makes it possible to open the variant analysis results view by single-clicking a query history item: 

open results view from query history

See internal issue for context/related work 🔍 

## Checklist

N/A - internal only 🦈 

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
@@ -73,18 +73,17 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
   }
 
   public async showView(variantAnalysisId: number): Promise<void> {
-    try{
-      if (!this.views.has(variantAnalysisId)) {
-        // The view will register itself with the manager, so we don't need to do anything here.
-        this.push(new VariantAnalysisView(this.ctx, variantAnalysisId, this));
-      }
-
-      const variantAnalysisView = this.views.get(variantAnalysisId)!;
-      await variantAnalysisView.openView();
-      return;
-    } catch (e) {
-      void showAndLogErrorMessage(`Could not open the results for variant analysis with id: ${variantAnalysisId}. Error: ${getErrorMessage(e)}`);
+    if (!this.variantAnalyses.get(variantAnalysisId)) {
+      void showAndLogErrorMessage(`No variant analysis found with id: ${variantAnalysisId}.`);
     }
+    if (!this.views.has(variantAnalysisId)) {
+      // The view will register itself with the manager, so we don't need to do anything here.
+      this.push(new VariantAnalysisView(this.ctx, variantAnalysisId, this));
+    }
+
+    const variantAnalysisView = this.views.get(variantAnalysisId)!;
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -224,6 +224,18 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
     await createTimestampFile(this.getVariantAnalysisStorageLocation(variantAnalysisId));
   }
 
+  public async openVariantAnalysisResults(variantAnalysisId: number): Promise<void> {
+    try {
+      const variantAnalysis = this.variantAnalyses.get(variantAnalysisId);
+      if (!variantAnalysis) {
+        void showAndLogErrorMessage(`No variant analysis with id: ${variantAnalysisId}`);
+      }
+      void commands.executeCommand('codeQL.openVariantAnalysisView', variantAnalysisId);
```
charisk, line 233: Is there a reason we're not calling `this.showView()` directly instead of executing a command?

Also, since there is a command to show a view, I'm wondering if it makes sense to execute that command directly from query history, rather than calling a function on the `VariantAnalysisManager`. What do you think? 

We could add some of the nice things you added around defensiveness to the `showView()` function.

shati-patel: > Is there a reason we're not calling `this.showView()` directly instead of executing a command?

No good reason 😅 I was following the example of [`openRemoteQueryResults`](https://github.com/github/vscode-codeql/blob/c26d786a1cac98ba4f6bf22e65133fdbba60e38d/extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts#L103), but I agree that it would be more sensible to just call `showView` directly! 

Some of that defensiveness [already exists in the view](https://github.com/github/vscode-codeql/blob/6a636baa2139a7fac8e706d98feafce9bc41233e/extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts#L109-L112) anyway, so I definitely didn't need to duplicate it 🤦🏽 

Clean-up incoming! 🧹

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -73,14 +73,18 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
   }
 
   public async showView(variantAnalysisId: number): Promise<void> {
-    if (!this.views.has(variantAnalysisId)) {
-      // The view will register itself with the manager, so we don't need to do anything here.
-      this.push(new VariantAnalysisView(this.ctx, variantAnalysisId, this));
+    try{
```
charisk, line 76: I'd personally avoid the big try catch and aim to eliminate errors before calling the functions inside the try/catch. 

The only thing I can think of that would blow up is the existence of variant analysis. I'd check that the variant analysis with the given id exists (`if (!this.variantAnalyses.get(variantAnalysisId)) {...}`)

charisk (approved):
LGTM, just a suggestion

charisk (approved):

PR #1655, opened 2022-10-25
Implement `handleRemoveHistoryItem` for variant analysis history items

elenatanasoiu:

We had previously added a no-op placeholder for when we attempt
to remove a variant analysis from our query history.

This adds the implementation by:
- removing the item from the query history
- cleaning up any existing result files attached to the variant analysis

NB: For remote queries, we would store all their results in a single folder.
For variant analyses, we store results per repo. The folder names are 
generated using a `cacheKey` which is built from the variant analysis ID and
the repo name. The resulting `cacheKeys` are stored in `cachedResults`.

Since we need to know the repo name when we want to delete a results 
folder,  we've had to pass in the full variant analysis object to the manager 
and call `cacheResults.delete()` for each of its scanned repos.

Also we're adding tests, because we need more of them. ❤️ 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -52,6 +52,7 @@ import { QueryWithResults } from './run-queries-shared';
 import { QueryRunner } from './queryRunner';
 import { VariantAnalysisManager } from './remote-queries/variant-analysis-manager';
 import { nanoid } from 'nanoid';
+import { VariantAnalysisHistoryItem } from './remote-queries/variant-analysis-history-item';
 
 /**
  * query-history.ts
@@ -614,7 +615,7 @@ export class QueryHistoryManager extends DisposableObject {
     const variantAnalysisRemovedSubscription = this.variantAnalysisManager.onVariantAnalysisRemoved(async (variantAnalysis) => {
       const item = this.treeDataProvider.allHistory.find(i => i.t === 'variant-analysis' && i.variantAnalysis.id === variantAnalysis.id);
       if (item) {
-        await this.removeRemoteQuery(item as RemoteQueryHistoryItem);
+        await this.removeVariantAnalysis(item as VariantAnalysisHistoryItem);
       }
     });
 
@@ -764,7 +765,7 @@ export class QueryHistoryManager extends DisposableObject {
       } else if (item.t === 'remote') {
         await this.removeRemoteQuery(item);
       } else if (item.t === 'variant-analysis') {
-        // TODO
+        await this.removeVariantAnalysis(item);
       } else {
         assertNever(item);
       }
@@ -788,6 +789,18 @@ export class QueryHistoryManager extends DisposableObject {
     }
 
     await this.remoteQueriesManager.removeRemoteQuery(item.queryId);
+  }
+
+  private async removeVariantAnalysis(item: VariantAnalysisHistoryItem): Promise<void> {
+    // Remote queries can be removed locally, but not remotely.
+    // The user must cancel the query on GitHub Actions explicitly.
+    this.treeDataProvider.remove(item);
+    void logger.log(`Deleted ${this.labelProvider.getLabel(item)}.`);
+    if (item.status === QueryStatus.InProgress) {
+      void logger.log('The variant analysis is still running on GitHub Actions. To cancel there, you must go to the workflow run in your browser.');
+    }
+
+    await this.variantAnalysisManager.removeVariantAnalysis(item.variantAnalysis);
   }
 
   async handleSortByName() {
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/query-history.ts
```diff
@@ -635,7 +636,7 @@ export class QueryHistoryManager extends DisposableObject {
     const variantAnalysisRemovedSubscription = this.variantAnalysisManager.onVariantAnalysisRemoved(async (variantAnalysis) => {
       const items = this.treeDataProvider.allHistory.filter(i => i.t === 'variant-analysis' && i.variantAnalysis.id === variantAnalysis.id);
```
norascheuch, line 637: Just wondering about it: can there be multiple items? With the decision to remove the historyId there should always only be one history item per variant analysis?

extensions/ql-vscode/src/query-history.ts
```diff
@@ -812,6 +813,18 @@ export class QueryHistoryManager extends DisposableObject {
     await this.remoteQueriesManager.removeRemoteQuery(item.queryId);
   }
 
+  private async removeVariantAnalysis(item: VariantAnalysisHistoryItem): Promise<void> {
+    // Remote queries can be removed locally, but not remotely.
```
norascheuch, line 817:
```suggestion
    // We can remove a variant analysis locally, but not remotely.
```

extensions/ql-vscode/src/remote-queries/variant-analysis-results-manager.ts
```diff
@@ -179,6 +179,19 @@ export class VariantAnalysisResultsManager extends DisposableObject {
     return `https://github.com/${fullName}/blob/${sha}`;
   }
 
+  public removeAnalysesResults(variantAnalysis: VariantAnalysis) {
```
norascheuch, line 182:
```suggestion
  public removeAnalysisResults(variantAnalysis: VariantAnalysis) {
```

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -72,6 +72,17 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
     }
   }
 
+  public async removeVariantAnalysis(variantAnalysis: VariantAnalysis) {
+    this.variantAnalysisResultsManager.removeAnalysesResults(variantAnalysis);
```
norascheuch, line 76: I think this should be named `removeAnalysisResults` as it's removing multiple results for only one analysis?

norascheuch: I started suggesting the change but it has a couple of occurrences so it's easier to do it in the editor

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -103,6 +114,14 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
...
   }
 
+  public getVariantAnalyses(): Map<number, VariantAnalysis> {
+    return this.variantAnalyses;
+  }
+
+  public async setVariantAnalysis(variantAnalysis: VariantAnalysis): Promise<void> {
+    this.variantAnalyses.set(variantAnalysis.id, variantAnalysis);
+  }
+
```
norascheuch, line 124: These both are only created for tests, right? .. 😬

norascheuch (commented):
Left a couple of questions and comments. I don't feel good with the test changes and methods created just for tests but I suppose it's not possible to write good tests otherwise?..

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -103,6 +114,14 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
     return this.variantAnalyses.get(variantAnalysisId);
   }
 
+  public getVariantAnalyses(): Map<number, VariantAnalysis> {
```
charisk, line 117: This is quite risky/leaky. It gives any part of the codebase access to get and edit variant analyses. If we _have to_ expose them, we should try and protect against random changes to the data by returning a copy of the real data. Perhaps using [immutablejs](https://immutable-js.com/) which is already used in the codebase.

elenatanasoiu: I've removed this public method and replaced it with a smaller one which just returns the size: https://github.com/github/vscode-codeql/pull/1655/commits/51b547b411d8afd09eaa1f3f87d18c43b61970d7

elenatanasoiu:
Closing in favour of https://github.com/github/vscode-codeql/pull/1656

I'm going to split this into smaller PRs so it doesn't block us.

PR #1656, opened 2022-10-26
Implement `handleRemoveHistoryItem` for variant analysis history items - take two

elenatanasoiu:
Previous attempt: https://github.com/github/vscode-codeql/pull/1655

At the moment we have a no-op placeholder for when we try to remove 
a variant analysis from our query history.

This adds the actual implementation by:
- removing the item from the query history list
- cleaning up any existing result files attached to the variant analysis

NB: For remote queries, we would store all their results in a single folder.
For variant analyses, we store results per repo. The folder names are 
generated using a `cacheKey` which is built from the variant analysis ID and
the repo name. The resulting `cacheKeys` are stored in `cachedResults`.

Since we need to know the repo name when we want to delete a results 
folder,  we've had to pass in the full variant analysis object to the manager 
and call `cacheResults.delete()` for each of its scanned repos.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -635,7 +635,7 @@ export class QueryHistoryManager extends DisposableObject {
     const variantAnalysisRemovedSubscription = this.variantAnalysisManager.onVariantAnalysisRemoved(async (variantAnalysis) => {
       const items = this.treeDataProvider.allHistory.filter(i => i.t === 'variant-analysis' && i.variantAnalysis.id === variantAnalysis.id);
       items.forEach(async (item) => {
-        await this.removeRemoteQuery(item as RemoteQueryHistoryItem);
+        await this.removeVariantAnalysis(item as VariantAnalysisHistoryItem);
       });
     });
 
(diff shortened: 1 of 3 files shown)


elenatanasoiu:
Unblocks https://github.com/github/vscode-codeql/pull/1650

extensions/ql-vscode/src/query-history.ts
```diff
@@ -812,6 +812,18 @@ export class QueryHistoryManager extends DisposableObject {
     await this.remoteQueriesManager.removeRemoteQuery(item.queryId);
   }
 
+  private async removeVariantAnalysis(item: VariantAnalysisHistoryItem): Promise<void> {
+    // We can remove a Variant Analysis locally, but not remotely.
+    // The user must cancel the query on GitHub Actions explicitly.
+    this.treeDataProvider.remove(item);
+    void logger.log(`Deleted ${this.labelProvider.getLabel(item)}.`);
+    if (item.status === QueryStatus.InProgress) {
+      void logger.log('The variant analysis is still running on GitHub Actions. To cancel there, you must go to the workflow run in your browser.');
```
koesie10, line 821: Should this be an informational notification rather than only a log message? We could then even provide a link to the Actions workflow.

elenatanasoiu: Thanks @koesie10! I've added a small issue for this.

PR #1662, opened 2022-10-26
Update languageserver-client

alexet:
This updates `languageserver-client` allowing for inlay hints (among other things) if codeql supports them. There were a few minor breaking changes so theu are fixed. I also upgraded the jsonrpc dependency so it matches the version to avoid the extra copy.

This does however have a higher vscode dependency (1.67 which is April 2022) so I think I need to bump that as well (I was slightly surprised that npm didn't complain before I did that). However if we wanted to make an upgrade warning that affected the current version then we would need to do it in a release before this.

## Checklist

- [x] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request. (No user visible changes)
- [x] Issues have been created for any UI or other user-facing changes made by this pull request. (No user visible changes)


aeisenberg (commented):
Nice. I'm guessing that this doesn't need a corresponding bump in the CLI for any LSP changes there?

Also, we should be careful about breaking users on old versions of vscode. I did a quick look in our telemetry and didn't find anyone on 1.67 or earlier, but our telemetry is incomplete. So, it's probably fine just to issue a error or warning for users on older versions with instructions that they need to either downgrade their extension version or upgrade their vscode version.

Is that something easy to do?

aeisenberg:
OK. Looks like we are going to do a vscode release tomorrow West Coast time. 

Would you be able to get a PR for whatever changes we need to warn users about vscode version requirements changing? (I'm not really sure what's required, otherwise I could probably do it myself.) And we can make sure that the PR is merged before we do the release tomorrow.

cc: @angelapwen (since I think you will be managing the release.)

angelapwen:
> This does nothing until the corresponding CLI code is released, I just needed it for testing and it wasn't blocked on anything. Therefore we aren't really in a rush and this shouldn't block the next release.

Just to clarify, does this mean that I can release later today without this PR merged, and we can release post-Universe with this PR + user warning?

aeisenberg:
Something like this? https://github.com/github/vscode-codeql/pull/1674

aeisenberg:
@angelapwen I think you can release today. It's not a requirement to get this PR in for this current release.

alexet:
Ok now we have the warning in, I think this PR should be Ok again, I had quite a painful merge.

Also when I tried to commit in codespaces I got issues with the pre commit hook so this was commited and pushed with `--no-verify`

aeisenberg:
Thanks for updating.

> Also when I tried to commit in codespaces I got issues with the pre commit hook so this was commited and pushed with `--no-verify`

This shouldn't be a problem. If your push has any issues, it will be caught in one of the checks.

aeisenberg (commented):
This looks great (but I haven't tried this out locally).

Can you add a changelog entry?

alexet:
I have added a changelog entry now.

aeisenberg (approved):

PR #1665, opened 2022-10-27
Add some basic integration tests for MRVA

koesie10:
This adds some basic integration tests for MRVA using the GitHub mock API server. It only does basic assertions and still needs to stub some things because it is quite hard to properly test things since VSCode does not expose an API to e.g. answer quick pick pop-ups.

I'm not sure how useful these integration tests will actually be in practice, but they do at least ensure that we are able to successfully submit a variant analysis.

## Checklist


extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis.integration.test.ts
```diff
@@ -0,0 +1,142 @@
...
+
+async function showQlDocument(name: string): Promise<TextDocument> {
+  const folderPath = workspace.workspaceFolders![0].uri.fsPath;
+  const documentPath = path.resolve(folderPath, name);
+  const document = await workspace.openTextDocument(documentPath);
+  await window.showTextDocument(document!);
+  return document;
+}
+
+describe('Variant Analysis Integration', function() {
```
charisk, line 37: I guess these are tests around variant analysis submission. What do you think?

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis.integration.test.ts
```diff
@@ -0,0 +1,142 @@
...
+
+const server = setupServer();
+
+before(() => server.listen());
+
+afterEach(() => server.resetHandlers());
+
+after(() => server.close());
+
+async function loadScenario(scenarioName: string) {
```
charisk, line 23: Assume you'll refactor to use the new mock server :)

charisk (commented):
Thanks for exploring this! 

I think it's worth getting these tests in as variant analysis submission tests that happen to use the mock API rather than suggesting they're special in some way (aka "variant analysis integration tests"). What do you think? Appreciate I'm probably the one that suggested the current term in the first place 😬

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis.integration.test.ts
koesie10: Yes, that makes sense. The reason for calling it an integration test is that they are not linked to a specific class, so if I had just named the file `variant-analysis.test.ts` I'd expect it to test a file `variant-analysis.ts`. To differentiate, I added `.integration.test.ts` to ensure that the difference is clear. What do you think?

charisk: I think once we start talking about integration of things it makes sense to not have a 1-to-1 mapping between tests and classes. We're testing the integration between multiple things to prove out a feature or part of it. In this case we're testing variant analysis submission. 

So perhaps `variant-analysis-submission-integration.test.ts`?

charisk (approved):
LGTM!

PR #1666, opened 2022-10-27
Update `variantAnalyses` and view when "rehydrating" variant analyses

shati-patel:
This should fix a bug with opening variant analysis results from the query history. See internal issue!

## Checklist

N/A - internal only 🍌 

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
@@ -61,10 +61,14 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
       // In this case, the variant analysis was deleted from disk, most likely because
       // it was purged by another workspace.
       this._onVariantAnalysisRemoved.fire(variantAnalysis);
-    } else if (status === QueryStatus.InProgress) {
-      // In this case, last time we checked, the query was still in progress.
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -61,10 +61,14 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
...
-      // We need to setup the monitor to check for completion.
-      await commands.executeCommand('codeQL.monitorVariantAnalysis', variantAnalysis);
+    } else {
+      this.variantAnalyses.set(variantAnalysis.id, variantAnalysis);
+      await this.getView(variantAnalysis.id)?.updateView(variantAnalysis);
+      if (status === QueryStatus.InProgress) {
+        // In this case, last time we checked, the query was still in progress.
+        // We need to setup the monitor to check for completion.
+        await commands.executeCommand('codeQL.monitorVariantAnalysis', variantAnalysis);
+      }
```
koesie10, line 71: Is it true that we only need to do this when the variant analysis was still in progress? I believe it can also happen that the variant analysis has completed, but not all results have been downloaded yet. I don't think it's something we need to solve in this PR, but it's something to be aware of.

koesie10 (approved):

PR #1672, opened 2022-10-27
When rehydrating, always trigger a monitoring command unless the variant analysis is fully complete

robertbrignull:
The point of this PR is to trigger a monitoring command more often when we rehydrate a variant analysis. This should help with handling cases like when the analysis is complete but not all results are downloaded.

My end goal here is that it's always safe to just fire off a monitoring command and it'll do the right thing and get the variant analysis moving. This PR doesn't cover making the monitoring command handle all possible states perfectly, but I think in the currently implementation it won't be anything bad, so I think it's ok to consider this PR on its own and I'll look at opening a separate PR to make the monitoring command do the right thing in all cases.

This is the largest piece of code I've written in the extension on my own to date, and the tests are a bit complicated and I'm not sure I've done them correctly. Please don't hold back and do help me to do things properly.

Questions  / concerns I have right now are:
- I haven't run the tests locally yet, because it's the end of the day and I'm not sure I've ever run the CLI tests locally. If they all fail I'll look that them tomorrow.
- I made `isVariantAnalysisRepoDownloaded` public. This seems fine to me but please let me know if not. Because I made it public I added some simple tests of it. Was this the right thing to do?
- I don't like that `isVariantAnalysisComplete` is public but it seemed the best solution for testing, so for the tests of rehydrating we can stub `isVariantAnalysisComplete` and then have separate simpler tests of just that method. Otherwise there are a lot of different cases to cover and it didn't seem sensible doing the whole rehydrating in all of them.
- It feels like `isVariantAnalysisComplete` should not be in `VariantAnalysisManager`, but I'm not sure where to put it otherwise and also not sure it can be pulled out because it depends on the results manager.
- In the tests I've nested my `describe`s very heavily and tried to test only one thing at a time. Is this good style or not helpful?

## Checklist


extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -151,6 +151,14 @@ export function hasRepoScanCompleted(repo: VariantAnalysisScannedRepository): bo
   return isCompletedAnalysisRepoStatus(repo.analysisStatus);
 }
 
+/**
+ * @param repo
+ * @returns whether the repo scan is complete and has results that can be downloaded
+ */
+export function repoScanHasResults(repo: VariantAnalysisScannedRepository): boolean {
+  return repo.analysisStatus === VariantAnalysisRepoStatus.Succeeded && (repo.resultCount || 0) > 0;
```
koesie10, line 159:
```suggestion
  return repo.analysisStatus === VariantAnalysisRepoStatus.Succeeded && (repo.resultCount ?? 0) > 0;
```

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
```diff
@@ -151,6 +151,14 @@ export function hasRepoScanCompleted(repo: VariantAnalysisScannedRepository): bo
   return isCompletedAnalysisRepoStatus(repo.analysisStatus);
 }
 
+/**
+ * @param repo
+ * @returns whether the repo scan is complete and has results that can be downloaded
```
koesie10, line 156: I don't think we're actually skipping the upload when there are no results, so we will always have results that can be downloaded. I think it's fine to skip the download, but I don't think we're skipping the download currently.

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
```diff
@@ -156,13 +158,177 @@ describe('Variant Analysis Manager', async function() {
...
+    let onVariantAnalysisRemovedListener: Disposable;
+    let monitorVariantAnalysisCommandCalled = false;
+
+    beforeEach(() => {
+      variantAnalysis = createMockVariantAnalysis();
+
+      variantAnalysisRemovedFired = false;
+      onVariantAnalysisRemovedListener = variantAnalysisManager.onVariantAnalysisRemoved(() => {
+        variantAnalysisRemovedFired = true;
+      });
```
koesie10, line 183: If you use a `sinon.stub` instead of using an arrow function, you should be able to assert that the function was called using [`sinon.assert`](https://sinonjs.org/releases/latest/assertions/)

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
```diff
@@ -156,13 +158,177 @@ describe('Variant Analysis Manager', async function() {
...
+
+      monitorVariantAnalysisCommandCalled = false;
+      sandbox.stub(commands, 'executeCommand').callsFake((command: string) => {
+        if (command === 'codeQL.monitorVariantAnalysis') {
+          monitorVariantAnalysisCommandCalled = true;
+        }
+        return Promise.resolve();
+      });
```
koesie10, line 191: Same here, using `sinon.stub` you can assert that this was called with a specific argument

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
```diff
@@ -156,13 +158,177 @@ describe('Variant Analysis Manager', async function() {
...
+    });
+
+    afterEach(() => {
+      onVariantAnalysisRemovedListener.dispose();
```
koesie10, line 195: I think it's fine to skip this step, I don't think we're doing it anywhere else and the variant analysis manager is recreated in every test without being disposed, so the event emitter itself is also not being disposed.

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -56,22 +59,42 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
...
+  private async isVariantAnalysisRepoComplete(variantAnalysisId: number, repo: VariantAnalysisScannedRepository): Promise<boolean> {
+    if (!hasRepoScanCompleted(repo)) {
+      return false;
+    } else if (!repoScanHasResults(repo)) {
+      return true;
+    } else {
+      const storageLocation = this.getVariantAnalysisStorageLocation(variantAnalysisId);
+      return await this.variantAnalysisResultsManager.isVariantAnalysisRepoDownloaded(storageLocation, repo.repository.fullName);
+    }
+  }
```
koesie10, line 96: Since this doesn't depend on any instance variables, I think you can extract this to a function which is outside of the class. We could have a simple function like `isVariantAnalysisComplete(variantAnalysisResultsManager: VariantAnalysisResultsManager, variantAnalysisStorageLocation: string, variantAnalysis: VariantAnalysis)` which can be called from anywhere. This would also solve the problem of this being unnecessarily `public` since we can just `export` it and not think about its visibility.

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -234,7 +257,7 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
    * used by the query history manager to determine when the directory
    * should be deleted.
    */
-  private async prepareStorageDirectory(variantAnalysisId: number): Promise<void> {
+  public async prepareStorageDirectory(variantAnalysisId: number): Promise<void> {
```
koesie10, line 260: I don't really think this should be public since it's quite specific to the manager. I'm not sure how we can improve this though, but since this is public to be used in tests I think we can just copy the code to the tests.

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
charisk: Agree with @koesie10 - this function can be extracted and unit tested/mocked independently. 

If we want to avoid passing in the `VariantAnalysisResultsManager`, we could instead pass in a function that helps decide whether a repo has downloads.

extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts
robertbrignull: Interesting. I agree we always upload an artefact, regardless of results. Skipping the download when there's no results makes sense to me as an optimisation, but I agree it's not strictly necessary.

For now I think I'd prefer to fit in with the existing behaviour, so that means we don't check if the number of results is non-zero. We can make that change later on if we want that as an optimisation, but I don't want us doing different behaviour in different places.

I also noticed that we could change `shouldDownload` from the `variant-analysis-monitor.ts` to call this method and that would let us show more logic.

robertbrignull: On trying to reuse the method it doesn't work because `shouldDownload` operates on a `gh_api` version of `VariantAnalysis`. I personally feel we should keep those API types as localised as possible and avoid them leaking around the extension, but I don't want to change this is this PR as it extends to the download command as well.

I'll leave this for now but open an issue so people can give their opinions there.

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
robertbrignull: Thanks. That's very handy to know.

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
robertbrignull: Thanks for the advice. That makes sense to me, to extract the method and pass in a function that checks whether the repo has downloads or not. This way we can keep the logic pure and contained to `variant-analysis.ts` (unless you have suggestions for somewhere better to put it). And we can test the logic of this method and the logic of rehydrating separately.

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
robertbrignull: If you think it's better to keep this method private but duplicate the implementation in the tests, I'm happy to go with that approach.

robertbrignull:
Had to do a slightly complex merge resolution. The main conflicts were in `extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-results-manager.test.ts`.

I also ended up doing https://github.com/github/vscode-codeql/pull/1672/commits/fef3159a58a5040c192bceb8341ad12aee0836cc because once I'd merge `main` I wasn't able to commit the changes otherwise.

I was getting
```
> git commit
husky > pre-commit (node v16.17.0)

> vscode-codeql@1.7.4 format-staged
> lint-staged

✔ Preparing...
⚠ Running tasks...
  ✔ Running tasks for ./**/*.{json,css,scss}
  ❯ Running tasks for ./**/*.{ts,tsx}
    ✔ tsfmt -r
    ✖ eslint --fix [FAILED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up...

✖ eslint --fix:

/Users/robertbrignull/github/vscode-codeql/extensions/ql-vscode/scripts/fix-scenario-file-numbering.ts
  0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: scripts/fix-scenario-file-numbering.ts.
The file must be included in at least one of the projects provided

✖ 1 problem (1 error, 0 warnings)

husky > pre-commit hook failed (add --no-verify to bypass)
```

robertbrignull:
> I also ended up doing [fef3159](https://github.com/github/vscode-codeql/commit/fef3159a58a5040c192bceb8341ad12aee0836cc) because once I'd merge `main` I wasn't able to commit the changes otherwise.

As discussed on slack, I'll not include this change in this PR. We'll address it separately.

koesie10 (approved):
LGTM, thanks for making these changes!

I believe the failing tests are due to missing an `await` on the `isVariantAnalysisComplete` calls in `extensions/ql-vscode/test/pure-tests/variant-analysis.test.ts`.

robertbrignull:
Ok. I've now run all three test suites locally and they all pass. That had better be it 😦

robertbrignull:
Yay I think this is good to go, though I will wait until tomorrow because it may make sense to merge https://github.com/github/vscode-codeql/pull/1687 first otherwise we risk being unable to load the extension cleanly for all MRVA developers.

PR #1674, opened 2022-10-28
Assert VSCode version is high enough

aeisenberg:
Simple change to verify that the vscode version is high enough to support some future features. 

Before bumping the `engine` property of package.json, we will need to first bump the min version in extension.ts, then release that bumped version. Once that extension version is available, people on older versions will get warnings that they need to upgrade. _Then_ we can release a new version of vscode that requires a new minimum version.

Update: Added a way to disable version checks

Version checks are re-enabled whenever the version of vscode changes.
This is because the user would have needed to manually update their
vscode version in order to get this new version. And another failing
version check would mean there is a newer version that needs to be
downloaded.

## Checklist


henrymercer (approved):

henrymercer:
> Question: Will this be annoying to some users who are forced to be on an older version? An ideal implementation would only warn once, and remember that there was a warning.

Do we have any telemetry information about which VS Code version our users are on?

aeisenberg:
The telemetry is not easy to access at the moment, but it looks like about 4.5% of the users who are sending telemetry are below v1.67 (and would be affected by this change).

I think the key thing to do here is to only ask them to upgrade once in a while (perhaps whenever they upgrade the extension). I'll work on that later.

aeisenberg:
Better approach will be to continue to show this message on every restart, and also have a "Don't show this again" button.

PR #1681, opened 2022-10-31
Set variant analysis view title to query name

koesie10:
This sets the title of the variant analysis view to `{queryName} - CodeQL Query Results`. It does so by retrieving the variant analysis in the `getPanelConfig` method and updating the title whenever the variant analysis is updated.

It's probably easiest to review this commit-by-commit. The actual change for setting the view title is quite small, but we first need to allow `getPanelConfig` to return a promise.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/abstract-webview.ts
+++ b/extensions/ql-vscode/src/abstract-webview.ts
@@ -38,7 +38,7 @@ export abstract class AbstractWebview<ToMessage extends WebviewMessage, FromMess
   public async restoreView(panel: WebviewPanel): Promise<void> {
     this.panel = panel;
     const config = await this.getPanelConfig();
-    await this.setupPanel(panel, config);
+    this.setupPanel(panel, config);
   }
 
   protected get isShowingPanel() {
--- a/extensions/ql-vscode/src/abstract-webview.ts
+++ b/extensions/ql-vscode/src/abstract-webview.ts
@@ -5,7 +5,7 @@ import {
   ViewColumn,
   Uri,
   WebviewPanelOptions,
-  WebviewOptions
+  WebviewOptions,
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/abstract-webview.ts
```diff
@@ -35,20 +37,36 @@ export abstract class AbstractWebview<ToMessage extends WebviewMessage, FromMess
 
   public async restoreView(panel: WebviewPanel): Promise<void> {
     this.panel = panel;
-    this.setupPanel(panel);
+    const config = await this.getPanelConfig();
+    await this.setupPanel(panel, config);
```
robertbrignull, line 41:
```suggestion
   this.setupPanel(panel, config);
```
It looks like `setupPanel` returns `void` rather than a promise.

extensions/ql-vscode/src/abstract-webview.ts
```diff
@@ -101,7 +122,7 @@ export abstract class AbstractWebview<ToMessage extends WebviewMessage, FromMess
     );
   }
 
-  protected abstract getPanelConfig(): WebviewPanelConfig;
+  protected abstract getPanelConfig(): WebviewPanelConfig | Promise<WebviewPanelConfig>;
```
robertbrignull, line 125: I suppose this works because calling `await` on something that's not a promise is a no-op, so you can safely call `await` on either of these return types and end up with the panel config object.

I'm not 100% sure the flexibility of this is worth the mental complexity, but it looks like it works so I'm not objecting.

robertbrignull (approved):
One suggested change but otherwise this LGTM.

I haven't been able to verify it manually because I'm having trouble running my extension, but this change looks fine.

PR #1683, opened 2022-10-31
Fix duplicate variant analysis results downloads

koesie10:
See the two separate commits; this also fixes an issue in the view when the repo task had succeeded but the results were not yet downloaded.

## Checklist


extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -176,6 +199,10 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
     variantAnalysisSummary: VariantAnalysisApiResponse,
     cancellationToken: CancellationToken
   ): Promise<void> {
+    if (this.repoStates.get(variantAnalysisSummary.id)?.[scannedRepo.repository.id]?.downloadStatus === VariantAnalysisScannedRepositoryDownloadStatus.Succeeded) {
```
robertbrignull, line 206: Thinking about the high-level approach, my main thought is why do we need to store extra state? I would have thought that we can check if a repo result has been downloaded by looking to see if the relevant file exists on disk, because when we download results we store them to a consistent location.

I would view an approach based on existing code and file system state as preferable to introducing new state that has to be kept up to date and consistent.

What do you think?

koesie10: I thought so too initially, but I don't think that would actually work. I think there are several problems with that approach:
- If we check whether the directory exists, it could be that the download hasn't even started yet (e.g. it could have ended [here](https://github.com/github/vscode-codeql/blob/f9553d7033778564989453075c5f3cadf67fb060/extensions/ql-vscode/src/remote-queries/variant-analysis-results-manager.ts#L68))
- If we check whether the `results.zip` file exists, it could mean that we haven't unzipped the file yet (e.g. [here](https://github.com/github/vscode-codeql/blob/f9553d7033778564989453075c5f3cadf67fb060/extensions/ql-vscode/src/remote-queries/variant-analysis-results-manager.ts#L74))
- That means we'd need to do two checks for the actual unzipped file: `results.sarif` and `results.bqrs` (or detect this based on the query type, which we don't currently store)
- Even if the `results.sarif` file exists, I believe it can happen that the file is not written completely and is corrupt. This can of course also happen with the `repo_states.json` file, but that would just re-trigger the download of all repositories. We don't have a way to force the redownload of a corrupted extracted file.
- Lastly, I believe it could be quite slow to do all these checks. For example, if we have 1,000 repositories, we'd need to do 1,000 to 2,000 file existence checks. I believe this will be relatively fast on Linux, but I'm not sure how performant it would be on Windows. Having a single file to check makes it much more performant and will result in faster time to first download.

I'd also be happy to implement this by checking the existence of a `results.sarif`/`results.bqrs` file, but we'd still need to check all repo states at start-up or when the view is loaded since the view also needs to know whether the results are downloaded. Without it, it doesn't know whether it can actually allow the user to expand the repo row to show results.

robertbrignull: I agree we'd have to be careful about partially written files. Could we work around that by writing a special file once the download is fully complete. So once everything is done we create `success.stamp` or whatever. Since we're just creating the file afterwards that should be atomic enough.

Do we need this system to be able to account for multiple downloads running at once? I support that would only happen if you're running two separate VSCode instances against the same workspace and that shouldn't happen. If we don't have to account for that, then if you find some files there but not the "success" file then that indicates it's corrupt but it should be safe to delete everything and start downloading again.

The point about it being slow is harder to get around. Is it possible to get estimates of how slow it is on windows? Might have to enlist the help of `@shati-patel` since she has a windows machine.

koesie10: I'm not I understand what the advantage of creating a `success.stamp` file would be compared to the approach where we only need to keep track of a single file per variant analysis.

I don't think we need to support multiple VSCode instances in the same workspace, so it don't think we need to account for any race conditions of that kind or the on-disk state being out-of-sync with the in-memory state.

I did just do some benchmarks on a Windows laptop I own using this script:
<details>
<summary><code>scripts/benchmark-file-reads.ts</code></summary>

```ts
import * as path from 'path';
import { performance } from 'perf_hooks';
import * as tmp from 'tmp-promise';
import { faker } from '@faker-js/faker';

async function run() {
  let directory: string;
  if (process.argv.length >= 3) {
    directory = path.resolve(process.argv[2]);
  } else {
    directory = (await tmp.dir({ prefix: 'benchmark_file_reads_', keep: false, unsafeCleanup: true })).path;
  }

  console.log(`Creating files in ${directory}`);

  const repositoryNames = [];

  for (let i = 0; i < 100_000; i++) {
    const repositoryName = `${faker.random.word()}/${faker.random.word()}-${i}`;
    repositoryNames.push(repositoryName);

    await fs.outputJson(path.join(directory, repositoryName, 'success.stamp'), {
      repositoryName: repositoryNames[i],
    });
  }

  console.log(`${repositoryNames.length} files created`);

  const start = performance.now();

  for (const repositoryName of repositoryNames) {
    await fs.pathExists(path.join(directory, repositoryName, 'success.stamp'));
  }

  const end = performance.now();

  console.log(`Time taken: ${end - start}ms`);
}

run().catch(e => {
  console.error(e);
  process.exit(2);
});
```
</details>

The results are as follows:

| **System**  | **1000 files** | **100,000 files** |
|-------------|----------------|-------------------|
| MacBook Pro | 12 ms          | 1,046 ms          |
| Windows     | 110 ms         | 11,231 ms         |
| Codespaces  | 49 ms          | 3,884 ms          |

This is probably also a best case scenario since the files are still in cache and all systems are using SSDs, I expect it to slow down by another order of magnitude when using HDDs and cold startup.

robertbrignull: Thanks for discussing the various options and doing that investigation into filesystem performance. I'm coming around to the plan of having the `repo_states.json` file. I agree it seems like the best solution in terms of scalability.

My understanding is that the main danger is still programming bugs due to having to keep the state updated in response to multiple different events (e.g. rehydrating, running a new query, starting/finishing a download) but hopefully we can cover all that effectively with tests. On that point, could you add some tests that the `repo_states.json` file is read, created, and updated correctly? That could be new tests or adding assertions into existing tests that the file state is correct after each operation.

koesie10: I have added tests to the variant analysis manager tests which test that the repo states are read and saved correctly. It should be ready for another review now.

PR #1687, opened 2022-10-31
Move loading query history to after registering commands

robertbrignull:
We have a problem that rehydrating the query history can lead to triggering the `codeQL.monitorVariantAnalysis` command.  This is a problem because we start loading the query history before we register the `codeQL.monitorVariantAnalysis` command with the extension. This leads to the following error:
```
Activating extension 'GitHub.vscode-codeql' failed: command 'codeQL.monitorVariantAnalysis' not found
```

So this PR moves the query history loading to be right at the end of the extension-loading process. I've confirmed locally that doing this allows me to open the extension when there is a variant analysis in progress.

Is there any problem of moving this to be the last thing in the method? @aeisenberg, do you have any thoughts?

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -525,9 +525,6 @@ async function activateWithInstalledDistribution(
   ctx.subscriptions.push(logScannerService);
   ctx.subscriptions.push(logScannerService.scanners.registerLogScannerProvider(new JoinOrderScannerProvider(() => joinOrderWarningThreshold())));
 
-  void logger.log('Reading query history');
-  await qhm.readQueryHistory();
-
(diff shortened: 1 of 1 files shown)


charisk (approved):

robertbrignull:
We're going to merge this now to unblock other work and fix this bug for us developers. We're pretty confident that this is the right fix, but would still appreciate thoughts from @aeisenberg once he's back from holiday.

We think this is safe to merge because we don't plan to do any release extensions in the next couple of weeks, and if we do we can make it a patch on top of the last release instead of including this change. So if this change is bad it should only affect internal users.

aeisenberg:
👋🏼 A little late on this, but it doesn't look like a problem to me.

PR #1689, opened 2022-11-01
Move `createGist` to `gh-api-client`

koesie10:
The `createGist` function was part of `gh-actions-api-client`, while it didn't actually involve anything related to the GitHub Actions API. This moves it to the non-Actions-specific `gh-api-client` module.

Another candidate for moving to `gh-api-client` is `getRepositoriesMetadata`, but that one is a bit more involved since it uses `showAndLogErrorMessage`, so depends on the `vscode` module. This means it would not be possible to test in the "pure" tests and we would need to move all our `gh-actions-api` tests to the integration tests. It will not be used for variant analysis queries anymore, so I don't think it's worth moving or refactoring to not depend on `vscode`.

## Checklist


robertbrignull (approved):
LGTM

I agree not worth moving the other method now if it's problematic since we'll be removing it in the future.

PR #1691, opened 2022-11-01
Add database configuration store

shati-patel:
Adds a database configuration file, which will contain state for the new (hidden) database panel introduced in https://github.com/github/vscode-codeql/pull/1682, e.g. what repo lists you've added, what repos you want to run against...

We've added a new `DatabaseConfigStore` class, which reads/modifies the config file (`dbconfig.json`). Nothing else should be able to modify the config, so we return a clone whenever anything calls `getConfig()`.*
*<sub>note: we tried using `Object.freeze()` but that didn't give us the appropriate "deep" clone.</sub>

In a future PR we'll set up a "watcher" that checks for changes to the config.

See internal linked issue for more details! 🔗 

## Checklist

N/A - internal only 🍜

### DIFF
--- a/extensions/ql-vscode/src/databases/database-config.ts
+++ b/extensions/ql-vscode/src/databases/database-config.ts
@@ -1,0 +1,29 @@
+// Contains models for the data we want to store in the database config
+
+export interface DatabaseConfig {
+  remote: RemoteDatabaseConfig;
+}
+
+export interface RemoteDatabaseConfig {
+  repositoryLists: RemoteRepositoryList[];
+  owners: string[];
+  repositories: string[];
+}
+
+export interface RemoteRepositoryList {
+  name: string;
+  repositories: string[];
+}
+
+export function cloneDatabaseConfig(config: DatabaseConfig): DatabaseConfig {
+  return {
+    remote: {
+      repositoryLists: config.remote.repositoryLists.map((list) => ({
+        name: list.name,
+        repositories: list.repositories.slice(),
+      })),
+      owners: [...config.remote.owners],
+      repositories: [...config.remote.repositories],
+    }
+  };
+}
--- a/extensions/ql-vscode/src/databases/db-config-store.ts
+++ b/extensions/ql-vscode/src/databases/db-config-store.ts
@@ -10,7 +10,7 @@ export class DbConfigStore {
   public constructor(workspaceStoragePath: string) {
     this.configPath = path.join(workspaceStoragePath, 'dbconfig.json');
 
-    this.config = emptyConfig;
+    this.config = this.createEmptyConfig();
   }
 
   public async initialize(): Promise<void> {
@@ -24,8 +24,7 @@ export class DbConfigStore {
 
   private async loadConfig(): Promise<void> {
     if (!await fs.pathExists(this.configPath)) {
-      await fs.createFile(this.configPath);
-      await fs.writeJSON(this.configPath, emptyConfig, { spaces: 2 });
+      await fs.writeJSON(this.configPath, this.createEmptyConfig(), { spaces: 2 });
     }
 
     await this.readConfig();
@@ -34,12 +33,14 @@ export class DbConfigStore {
   private async readConfig(): Promise<void> {
     this.config = await fs.readJSON(this.configPath);
   }
-}
 
-const emptyConfig: DbConfig = {
-  remote: {
-    repositoryLists: [],
-    owners: [],
-    repositories: [],
+  private createEmptyConfig(): DbConfig {
+    return {
+      remote: {
+        repositoryLists: [],
+        owners: [],
+        repositories: [],
+      }
+    };
   }
-};
+}
--- a/extensions/ql-vscode/src/databases/db-config-store.ts
+++ b/extensions/ql-vscode/src/databases/db-config-store.ts
@@ -1,11 +1,11 @@
 import * as fs from 'fs-extra';
 import * as path from 'path';
-import { cloneDatabaseConfig, DatabaseConfig } from './database-config';
+import { cloneDbConfig, DbConfig } from './db-config';
 
-export class DatabaseConfigStore {
+export class DbConfigStore {
   private readonly configPath: string;
 
-  private config: DatabaseConfig;
+  private config: DbConfig;
 
   public constructor(workspaceStoragePath: string) {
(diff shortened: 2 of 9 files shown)


extensions/ql-vscode/src/databases/database-config.ts
```diff
@@ -0,0 +1,29 @@
...
+  return {
+    remote: {
+      repositoryLists: config.remote.repositoryLists.map((list) => ({
+        name: list.name,
+        repositories: list.repositories.slice(),
+      })),
+      owners: [...config.remote.owners],
+      repositories: [...config.remote.repositories],
+    }
+  };
```
aeisenberg, line 28: Possibly, this is more robust. The advantage is that if you change the shape of `DatabaseConfig` to add or remove properties, this function will not need to change the disadvantage is that any non-standard objects will lose their prototype.

```suggestion
  return JSON.parse(JSON.stringify(config));
```

charisk: I generally avoid doing that because it's considered slow. 

I think the shape will change a little bit while we implement the feature but it won't change often because it's something that will be visible to the users.

extensions/ql-vscode/src/databases/db-config.ts
```diff
@@ -0,0 +1,29 @@
...
+  name: string;
+  repositories: string[];
+}
+
+export function cloneDbConfig(config: DbConfig): DbConfig {
+  return {
+    remote: {
+      repositoryLists: config.remote.repositoryLists.map((list) => ({
+        name: list.name,
+        repositories: list.repositories.slice(),
```
robertbrignull, line 23: Out of interest, what's the reason for doing `list.repositories.slice()` here but doing `[...config.remote.owners]` later on?

extensions/ql-vscode/src/databases/db-config-store.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  public getConfig(): DbConfig {
+    // Clone the config so that it's not modified outside of this class.
+    return cloneDbConfig(this.config);
+  }
+
+  private async loadConfig(): Promise<void> {
+    if (!await fs.pathExists(this.configPath)) {
+      await fs.createFile(this.configPath);
```
koesie10, line 27: Do we need this call to `createFile`? I would expect `writeJSON` to also create the file.

extensions/ql-vscode/src/databases/db-config-store.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export class DbConfigStore {
+  private readonly configPath: string;
+
+  private config: DbConfig;
+
+  public constructor(workspaceStoragePath: string) {
+    this.configPath = path.join(workspaceStoragePath, 'dbconfig.json');
+
+    this.config = emptyConfig;
```
koesie10, line 13: Do we also need to clone this config? If we now create multiple instances of `DbConfigStore`, they would have a reference to the same config object and would see changes made by the other instance (and the values in `emptyConfig` would also be modified).

It probably won't be a problem since we always load in a new object in `readConfig`, but I still think it's good practice to avoid having the same reference for every instance.

extensions/ql-vscode/src/databases/db-config-store.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  public async initialize(): Promise<void> {
```
norascheuch, line 16: Question: why do we have an extra initialize method for this, why not put this in the constructor? 
I wonder because by using the `initialize` method someone using the class needs to remember to call it after creating a new instance, we increase complexity. Will we ever use the class without having run what's in `initialize` before?

norascheuch (commented):
One question otherwise lgtm!

extensions/ql-vscode/src/databases/db-config.ts
charisk: No good reason - updated to use spread everywhere. In theory I think `slice` might be faster in our case but I don't think it matters here, so any would do.

extensions/ql-vscode/src/databases/db-config-store.ts
charisk: Oh I didn't realise `writeJSON` creates the file too - that's handy. Removed `createFile`.

extensions/ql-vscode/src/databases/db-config-store.ts
charisk: That's a good shout. Updated to have a createEmptyConfig() function instead of an object.

extensions/ql-vscode/src/databases/db-config-store.ts
charisk: It's a good point about having to remember to call `initialize()` but unfortunately constructors can't be async and we don't want to block execution while we read the config.

extensions/ql-vscode/test/pure-tests/databases/db-config-store.test.ts
```diff
@@ -0,0 +1,57 @@
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import { DbConfigStore } from '../../../src/databases/db-config-store';
+import { expect } from 'chai';
+
+
+describe('db config store', async () => {
+  const workspaceStoragePath = path.join(__dirname, 'test-workspace');
+  const testStoragePath = path.join(__dirname, 'data');
```
norascheuch, line 9: Should we use a copy of this for every test that we can remove in an afterEach, so that tests cannot influence each other?

extensions/ql-vscode/src/databases/db-config-store.ts
norascheuch: Ah, you mentioned this, now I remember 🤦‍♀️ 🙈. All async methods can be replaced by methods from fs-extra I think?

charisk: We could use sync methods but that would mean blocking things rather than doing it asynchronously. Does that make sense?

extensions/ql-vscode/test/pure-tests/databases/db-config-store.test.ts
charisk: There are two dirs - I've updated the two variables to hopefully make more sense!

norascheuch (approved):

PR #1693, opened 2022-11-01
Fix indentation in exported markdown results

shati-patel:
This PR fixes the markdown export for remote query results. Previously, paths with more than 100 steps had their code snippets badly indented, which caused the markdown to render badly on GitHub Gists. 

I've changed the indentation to match the length of the list number, and updated tests to check this variable indentation size.

## Checklist

N/A - this is a private beta feature. See internal issue for more details!

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
+++ b/extensions/ql-vscode/src/remote-queries/remote-queries-markdown-generation.ts
@@ -205,6 +205,7 @@ function generateMarkdownForPathResults(
     const stepCount = codeFlow.threadFlows.length;
     const title = `Path with ${stepCount} steps`;
     for (let i = 0; i < stepCount; i++) {
+      const listNumber = i + 1;
       const threadFlow = codeFlow.threadFlows[i];
       const link = createMarkdownRemoteFileRef(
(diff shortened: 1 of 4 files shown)


koesie10 (approved):
LGTM

Do you think it might make sense to use code blocks using ``` as a future improvement?

robertbrignull:
> Do you think it might make sense to use code blocks using ``` as a future improvement?

I believe the reason we don't use ``` is because we want to add styling within the code block to highlight a specific bit. The code blocks don't allow you to add extra styling within them, though they do give other benefits like syntax highlighting and avoiding markdown indentation weirdness like this.

shati-patel:
Thanks for the review, @koesie10  ✨ 

> > Do you think it might make sense to use code blocks using ``` as a future improvement?
> 
> I believe the reason we don't use ``` is because we want to add styling within the code block to highlight a specific bit. The code blocks don't allow you to add extra styling within them, though they do give other benefits like syntax highlighting and avoiding markdown indentation weirdness like this.

Indeed! We can't add HTML styling (such as `<strong>` tags) in a regular markdown code block 😢 Slightly annoying, but it was the best compromise we could find.

PR #1694, opened 2022-11-01
Introduce command to run no-workspace tests separately

elenatanasoiu:
This takes down the feedback loop from 5-10 seconds to half a second since we're not running through the setup for minimal workspace tests.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -1258,6 +1258,7 @@
     "test:view": "jest",
     "preintegration": "rm -rf ./out/vscode-tests && gulp",
     "integration": "node ./out/vscode-tests/run-integration-tests.js no-workspace,minimal-workspace",
+    "no-workspace": "node ./out/vscode-tests/run-integration-tests.js no-workspace",
     "cli-integration": "npm run preintegration && node ./out/vscode-tests/run-integration-tests.js cli-integration",
     "update-vscode": "node ./node_modules/vscode/bin/install",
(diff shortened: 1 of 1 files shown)


elenatanasoiu:
Thanks @aeisenberg!

extensions/ql-vscode/package.json
```diff
@@ -1258,6 +1258,7 @@
     "test:view": "jest",
     "preintegration": "rm -rf ./out/vscode-tests && gulp",
     "integration": "node ./out/vscode-tests/run-integration-tests.js no-workspace,minimal-workspace",
+    "no-workspace": "node ./out/vscode-tests/run-integration-tests.js no-workspace",
```
robertbrignull, line 1261: Sadly I've discovered why this is so much faster than running `integration`, and that's because it's skipping `preintegration`. I didn't know this was a feature of `npm` until now but have a read of https://docs.npmjs.com/cli/v8/using-npm/scripts#pre--post-scripts. This means that `no-workspace` is different from running `integration` because it won't run `rm -rf ./out/vscode-tests && gulp` and this is what actually takes the majority of the time. For me running all of `integration` takes 164s, but just running `preintegration` takes 90s.

So for right now I think we need to add this back. What do you think? I've opened https://github.com/github/vscode-codeql/pull/1696 which follows on from this so take a look there.

If we want to make the tests faster I think we sadly have to make `gulp` faster, or deliberately decide to skip it and run the risk of running tests on outdated javascript files.

PR #1695, opened 2022-11-01
Fix flakey test

elenatanasoiu:
This test has been quite flakey due to timeout errors:

```
1) "before each" hook: beforeEach for "should avoid creating a quick query":
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at listOnTimeout (node:internal/timers:559:17)
      at processTimers (node:internal/timers:502:7)
```

Since this is an async test, let's attempt to make all file operations use their asynchronous counterparts and see if this allows our test to finish on time.

Normally I'd consider increasing the timeout but the test itself doesn't load massive amounts of data so the culprit seems to be an unresolved promise.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
@@ -5,6 +5,7 @@ import * as path from 'path';
 import * as fs from 'fs-extra';
 import { expect } from 'chai';
 import * as yaml from 'js-yaml';
+import { faker } from '@faker-js/faker';
 
 import { DatabaseItem, DatabaseManager } from '../../databases';
 import { CodeQLExtensionInterface } from '../../extension';
@@ -15,7 +16,6 @@ import { skipIfNoCodeQL } from '../ensureCli';
 import { tmpDir } from '../../helpers';
 import { createInitialQueryInfo } from '../../run-queries-shared';
 import { QueryRunner } from '../../queryRunner';
-
 
 /**
  * Integration tests for queries
@@ -41,7 +41,6 @@ describe('Queries', function() {
   let oldQlpackLockFile: string; // codeql v2.6.3 and earlier
   let qlFile: string;
 
-
   beforeEach(async () => {
     sandbox = sinon.createSandbox();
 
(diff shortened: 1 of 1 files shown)


aeisenberg:
The error happens in the `beforeEach` block, and so the current changes will not address that problem. 

I'm pretty sure that the hanging happens in the calls to [`safeDel`](https://github.com/github/vscode-codeql/blob/075a0d562589ab7f458989f4fcbb8d5f530fec46/extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts#L65).

I've had problems with this test before. Maybe the best solution is to generate new values for `qlFile` and `qlpackFile` for each test run. And assume the temp files will be cleaned up after the test process finishes (true for linux/mac, not true for windows).

extensions/ql-vscode/src/vscode-tests/cli-integration/queries.test.ts
```diff
@@ -53,18 +52,14 @@ describe('Queries', function() {
         qs = extension.qs;
         cli.quiet = true;
         ctx = extension.ctx;
-        qlpackFile = `${ctx.storageUri?.fsPath}/quick-queries/qlpack.yml`;
+        qlpackFile = `${ctx.storageUri?.fsPath}/quick-queries/qlpack-${faker.name}.yml`;
         qlpackLockFile = `${ctx.storageUri?.fsPath}/quick-queries/codeql-pack.lock.yml`;
         oldQlpackLockFile = `${ctx.storageUri?.fsPath}/quick-queries/qlpack.lock.yml`;
-        qlFile = `${ctx.storageUri?.fsPath}/quick-queries/quick-query.ql`;
+        qlFile = `${ctx.storageUri?.fsPath}/quick-queries/quick-query-${faker.name}.ql`;
```
aeisenberg, line 58: Unfortunately, this won't work since we expect a hard coded name for the quick query file.

I think we need to update the `ctx.storageUri` for each test and keep the rest of the path the same.

elenatanasoiu: Thanks for addressing this @aeisenberg. Closing this for now. If there are more flakey tests that pop-up I'll have another look. 👀

PR #1696, opened 2022-11-02
Run test setup before integration tests

robertbrignull:
Follows on from https://github.com/github/vscode-codeql/pull/1694. See comment on that PR.
cc @elenatanasoiu, @aeisenberg 

Sadly I've discovered why running `no-workspace` from that PR is so much faster than running `integration`, and that's because it's skipping `preintegration`. I didn't know this was a feature of `npm` until now but have a read of https://docs.npmjs.com/cli/v8/using-npm/scripts#pre--post-scripts. This means that `no-workspace` is different from running `integration` because it won't run `rm -rf ./out/vscode-tests && gulp` and this is what actually takes the majority of the time. For me running all of `integration` takes 164s, but just running `preintegration` takes 90s.

I've also added a `minimal-workspace` version and adjusted the names to fit the existing pattern a bit better. I can change the names back if you don't like them, so just let me know.

If we want to make the tests faster I think we sadly have to make `gulp` faster, or deliberately decide to skip it and run the risk of running tests on outdated javascript files.

## Checklist


elenatanasoiu (approved):
That's a shame. Thanks for looking into it more closely ❤️

PR #1697, opened 2022-11-02
Don't show "open on github" link when we don't yet have anything to show

robertbrignull:
We were trying to show the "Open variant analysis on github" link before we knew what the actions workflow run ID was, and therefore we were generating links like https://github.com/dsp-testing/qc-controller/actions/runs/undefined

See the individual commits on this PR. The first does some drive-by recfactoring and then the second commit actually implements the change in behaviour. I've tested it manually and it seems to be working as expected.

I went with just pulling `getIconPath` and `getContextValue` out to methods in `query-history.ts` because they unfortunately have references to instance variables and to imports from `vscode`. Otherwise I would have moved them to `query-history-info.ts`, but maybe we can do that in the future. For this PR I just wanted to make an improvement and not get stuck on it.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -199,7 +199,13 @@ export class HistoryTreeDataProvider extends DisposableObject implements TreeDat
   private async getContextValue(element: QueryHistoryInfo): Promise<string> {
     switch (element.status) {
       case QueryStatus.InProgress:
-        return element.t === 'local' ? 'inProgressResultsItem' : 'inProgressRemoteResultsItem';
+        if (element.t === 'local') {
+          return 'inProgressResultsItem';
+        } else if (element.t === 'variant-analysis' && element.variantAnalysis.actionsWorkflowRunId === undefined) {
+          return 'pendingRemoteResultsItem';
+        } else {
+          return 'inProgressRemoteResultsItem';
+        }
       case QueryStatus.Completed:
         if (element.t === 'local') {
           const hasResults = await element.completedQuery?.query.hasInterpretedResults();
--- a/extensions/ql-vscode/src/query-history.ts
+++ b/extensions/ql-vscode/src/query-history.ts
@@ -173,33 +173,47 @@ export class HistoryTreeDataProvider extends DisposableObject implements TreeDat
 
     // Populate the icon and the context value. We use the context value to
     // control which commands are visible in the context menu.
-    let hasResults;
+    treeItem.iconPath = this.getIconPath(element);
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/query-history.ts
```diff
@@ -173,33 +173,47 @@ export class HistoryTreeDataProvider extends DisposableObject implements TreeDat
...
+          return this.remoteSuccessIconPath;
+        }
+      case QueryStatus.Failed:
+        return this.failedIconPath;
+      default:
+        assertNever(element.status);
+    }
+  }
+
+  private async getContextValue(element: QueryHistoryInfo): Promise<string> {
```
koesie10, line 199: Can we extract this method out of the class so it can be tested separately? It doesn't seem like it uses `this` anywhere.

koesie10 (approved):

extensions/ql-vscode/src/query-history.ts
robertbrignull: You're right it's only `getIconPath` that relies on `this`, so we could pull out just `getContextValue`. I suppose doing it for one method is better than for zero methods.

robertbrignull: Ok, I've tried and I'm struggling quite a lot. The problem is that `getContextValue` depends on `element.completedQuery?.query.hasInterpretedResults()` and this calls into `QueryEvaluationInfo` and generally hits two problems:

The real implementation of the method relies on disk state, and the query history info tests aren't really set up for that. We could work around that passing in a helper method to `getContextValue`. It's a bit ugly but it would work. However there are deeper problems that than.

We don't do a good job of creating the right types in our tests. In this case `createMockLocalQueryInfo` doesn't actually create a `LocalQueryInfo`. It casts a random object to that type, but in reality `LocalQueryInfo` is a class so you can't just do this. This means when we call `element.status` it returns undefined because although `element.t === 'local'` it is not an instance of the `LocalQueryInfo` type, so the `status` method is not defined. I had a look at whether I could instantiate the class easily but it looks quite complicated. We do do some of this in `query-history.test.ts` but at this point things are getting very complicated and wide-ranging.

I'd like to merge this change as it is and then look at improving things separately.

elenatanasoiu: Fwiw, I'm going to introduce an improved version of the factory which does create a proper object. I need it in order to make sorting tests work. So you might be able to do this after that's merged.

elenatanasoiu: This has now been merged: https://github.com/github/vscode-codeql/pull/1710

PR #1699, opened 2022-11-02
Set up a watcher that checks for changes to the new db config

shati-patel:
Uses [chokidar](https://www.npmjs.com/package/chokidar) to watch for changes to the `dbconfig.json` file. 
Note: This watcher currently isn't used anywhere, except in tests.

(paired with @norascheuch and @charisk)

## Checklist

N/A - internal only 🐸 

### DIFF
--- a/extensions/ql-vscode/src/databases/db-config-store.ts
+++ b/extensions/ql-vscode/src/databases/db-config-store.ts
@@ -36,9 +36,13 @@ export class DbConfigStore {
     this.config = await fs.readJSON(this.configPath);
   }
 
+  private readConfigSync(): void {
+    this.config = fs.readJSONSync(this.configPath);
+  }
+
   private watchConfig(): void {
-    chokidar.watch(this.configPath).on('change', async () => {
-      await this.readConfig();
+    chokidar.watch(this.configPath).on('change', () => {
+      this.readConfigSync();
     });
   }
 
--- a/extensions/ql-vscode/src/databases/db-config-store.ts
+++ b/extensions/ql-vscode/src/databases/db-config-store.ts
@@ -2,21 +2,29 @@ import * as fs from 'fs-extra';
 import * as path from 'path';
 import { cloneDbConfig, DbConfig } from './db-config';
 import * as chokidar from 'chokidar';
+import { DisposableObject } from '../pure/disposable-object';
 
-export class DbConfigStore {
+export class DbConfigStore extends DisposableObject {
   private readonly configPath: string;
 
   private config: DbConfig;
+  private configWatcher: chokidar.FSWatcher | undefined;
 
   public constructor(workspaceStoragePath: string) {
+    super();
     this.configPath = path.join(workspaceStoragePath, 'dbconfig.json');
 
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/test/pure-tests/databases/db-config-store.test.ts
```diff
@@ -54,4 +55,43 @@ describe('db config store', async () => {
...
+    expect(reRetrievedConfig.remote.repositoryLists[0]).to.deep.equal({
+      'name': 'repoList2',
+      'repositories': ['foo/bar2', 'foo/baz2']
+    });
+  });
 });
+
+async function sleep(ms: number) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
```
shati-patel, line 97: Note: Once https://github.com/github/vscode-codeql/pull/1700 is merged, we can import `sleep` instead of redefining it here.

extensions/ql-vscode/src/databases/db-config-store.ts
```diff
@@ -34,6 +36,16 @@ export class DbConfigStore {
     this.config = await fs.readJSON(this.configPath);
   }
 
+  private readConfigSync(): void {
+    this.config = fs.readJSONSync(this.configPath);
+  }
+
+  private watchConfig(): void {
+    chokidar.watch(this.configPath).on('change', () => {
```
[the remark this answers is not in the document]
charisk: Done

extensions/ql-vscode/src/databases/db-config-store.ts
```diff
@@ -2,23 +2,31 @@ import * as fs from 'fs-extra';
...
   }
 
   public async initialize(): Promise<void> {
     await this.loadConfig();
     this.watchConfig();
   }
 
+  public dispose(): void {
+    this.configWatcher?.unwatch(this.configPath);
+  }
```
elenatanasoiu, line 28: Is it worth adding a little test to check we call dispose correctly?

There are some examples in our [telemetry tests](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/vscode-tests/no-workspace/telemetry.test.ts#L133).

elenatanasoiu (approved):
Left a suggestion, but it doesn't have to happen in this PR if it's holding you up.

extensions/ql-vscode/src/databases/db-config-store.ts
charisk: Hmm that's an interesting idea! 

I think there are two types of tests we could potentially add:
- One that tests that the `dispose` function is actually clearing out the watcher. I think that would require some `proxyquire` stuff 🤔 and adding more usage of `proxyquire` feels like adding tech debt since we want to move away from it soon. 
- One further up the chain (so can't write it yet) that uses `DbConfigStore` and ensures that `dispose` is called on it at the right time (like the example you linked).

So I think I'll just note this in an issue somewhere for now and come back to it later on. If we've not moved away from `proxyquire` by then, we might just have to use it.

Thanks for the feedback!

PR #1702, opened 2022-11-02
Don't run gulp for integration test commands

elenatanasoiu:
[We initially introduced][1] a `npm run no-workspace` command because it would run in half a second instead of 5-10 seconds as it didn't include the setup steps for minimal workspace.

[We looked a bit more into it][2] and realised we also need to run `rm -rf ./out/vscode-tests && gulp` when we run that command. This adds back the latency.

However, we already have `npm run watch` running in the background when we run our tests, so instead of running `gulp` on every test run we can introduce a `watcher` to check whether our test files need updating. This is much faster.

We're also removing the need to run `rm -rf` on every test command. In CI, we are guaranteed to have a clean VM so we don't need to remove any files. Locally, we want to avoid running `rm -rf` since that takes a lot of time. We'll instead rely on our new watcher to update files.

NB: While there's still a danger that you might forget to run `npm run watch`, we think the massive speed up for running test commands locally is worth it as we add more and more tests to this extension. To draw attention to this, we're updating the `CONTRIBUTING.md` file.

[1]: https://github.com/github/vscode-codeql/pull/1694
[2]: https://github.com/github/vscode-codeql/pull/1696

## Checklist


aeisenberg:
Will this delete non-javascript test artifacts like projects and data files? And does watch copy them back over?

extensions/ql-vscode/gulpfile.ts/tests.ts
```diff
@@ -6,6 +6,10 @@ export function copyTestData() {
   return Promise.resolve();
 }
 
+export function watchTestData() {
+  return gulp.watch(['src/vscode-tests/no-workspace/data/**/*', 'src/vscode-tests/minimal-workspace/data/**/*'], copyTestData);
```
aeisenberg, line 10: Maybe we should add a reminder in the contributing docs that all test data (sample projects, config files, etc) must go in the `data` folder.

And maybe this is a slight simplification that will make it easier to read:

```suggestion
  return gulp.watch('src/vscode-tests/*/data/**/*', copyTestData);
```

extensions/ql-vscode/gulpfile.ts/tests.ts
```diff
@@ -6,6 +6,10 @@ export function copyTestData() {
   return Promise.resolve();
 }
 
+export function watchTestData() {
```
aeisenberg, line 9: And looking at `copyTestData` now, I think there is a problem. `copyNoWorkspaceData` and `copyCliIntegrationData` return promises that only resolve when the copying completes, but `copyTestData` doesn't await them. It's probably working fine now since there aren't that many files to copy, but it may break in the future if there are some big files that take a long time.

This should work:

```
export function copyTestData() {
  return Promise.all([
    copyNoWorkspaceData(),
    copyCliIntegrationData()
  ]);
}
```

CONTRIBUTING.md
```diff
@@ -102,11 +102,15 @@ We have several types of tests:
 
 The CLI integration tests require an instance of the CodeQL CLI to run so they will require some extra setup steps. When adding new tests to our test suite, please be mindful of whether they need to be in the cli-integration folder. If the tests don't depend on the CLI, they are better suited to being a VSCode integration test.
 
+Any test data you're using (sample projects, config files, etc.) must go in a `src/vscode-tests/*/data` directory. When you run the tests, the test runner will copy the data directory to `out/vscode-tests/*/data`.
+
 #### Running the tests
 
+You will need to have `npm run watch` running in the background.
```
koesie10, line 109: Do we need to specify that you need to have run `npm run build` at least once? IIRC `npm run watch` doesn't copy over the `node_modules`.

koesie10 (approved):

PR #1703, opened 2022-11-02
Make hasInterpretedResults not be async

robertbrignull:
A tiny improvement I thought of when doing https://github.com/github/vscode-codeql/pull/1697 but didn't / forgot to include in that PR.

This is something I did when trying to pull out the `getContextValue` method and mock things, which didn't turn out well (see comment on that PR) but this small bit could be nice and could make things easier to mock and test in the future.

## Checklist


elenatanasoiu (approved):

robertbrignull:
Thanks for the review but I've changed my mind about whether changing async to sync is actually an improvement. I now don't think it's worth it. It either makes no difference or it removes flexibility. Also if we do ever want to do this change in the future it's very easy, so I don't think it's worth merging this now.

PR #1688, opened 2022-10-31
Make query history tests work with remote queries & variant analyses

elenatanasoiu:
🚨 Please review this PR commit-by-commit.

This PR holds part of the clean-up work to allow us to write more tests for the `QueryHistoryManager` and `HistoryTreeDataProvider` by re-organising tests to make it clear what they're checking and removing
some duplication in the process. This unfortunately results in a lot of noisy diffs as I shuffle things into 
`describe` blocks.

We're also changing part of these tests to work with `RemoteQueryHistoryItem`s and 
`VariantAnalysisHistoryItem`s, instead of just testing local query history items.

The query history now has increased test coverage in terms of how items of different types interact
together, specifically:
- handling an item being clicked / double clicked
- ~~removing and selecting an item~~ this has been reverted, see https://github.com/github/vscode-codeql/pull/1688/commits/5a4015900f1d638781aa6b78dd070bf4d35235de for details 
- showing query results for different item types
- handling single / multi selection

In a future PR we'll also look at:
- sorting results of different types and 
- removing & re-selecting an item

Disclaimer: I don't think these tests are perfect and there's definitely still some duplication between the 
tests in `query-history.test.ts` and the individual history item test files, but they're a step in the right 
direction as they untangle some things that are annoying to deal with in terms of writing/changing 
query history tests.

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #1710, opened 2022-11-03
Make sorting tests work with remote queries & variant analysis history items

elenatanasoiu:
🚨 Please review this commit-by-commit.

This is a follow-up to https://github.com/github/vscode-codeql/pull/1688 where we first started converting our query history tests to check all types of history items instead of just local query ones. 

In order to test sorting with local queries, remote queries and variant analyses, we've had to make a few changes:

1. Allow `createMockVariantAnalysis` and `createMockVariantAnalysisHistoryItem` to receive named params so that we can set `resultCount` and `executionStartTime`. We need this in order to test sorting by result and date.
2. We do the same for `createMockRemoteQueryHistoryItem`.
3. We're combining the two existing factory methods for creating a `LocalQueryInfo` into a single `createMockLocalQueryInfo` method.
4. For `createMockLocalQueryInfo` we're now returning a real object instead of typecasting something else. Not having a real object [blocked us][1] from being able to understand which fields we need in our tests and how they interact together.
5. Finally, we change our sorting tests to check all types of history item types, instead of just local queries.

[1]: https://github.com/github/vscode-codeql/pull/1697#discussion_r1011990685

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/variant-analysis.ts
+++ b/extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/variant-analysis.ts
@@ -10,11 +10,17 @@ import { createMockScannedRepos } from './scanned-repositories';
 import { createMockSkippedRepos } from './skipped-repositories';
 import { createMockRepository } from './repository';
 
-export function createMockVariantAnalysis(
-  status: VariantAnalysisStatus = VariantAnalysisStatus.InProgress,
-  scannedRepos: VariantAnalysisScannedRepository[] = createMockScannedRepos(),
-  skippedRepos: VariantAnalysisSkippedRepositories = createMockSkippedRepos()
-): VariantAnalysis {
+export function createMockVariantAnalysis({
+  status = VariantAnalysisStatus.InProgress,
+  scannedRepos = createMockScannedRepos(),
+  skippedRepos = createMockSkippedRepos(),
+  executionStartTime = faker.datatype.number()
+}: {
+  status?: VariantAnalysisStatus,
+  scannedRepos?: VariantAnalysisScannedRepository[],
+  skippedRepos?: VariantAnalysisSkippedRepositories,
+  executionStartTime?: number | undefined
+}): VariantAnalysis {
   const variantAnalysis: VariantAnalysis = {
     id: faker.datatype.number(),
     controllerRepo: {
@@ -32,7 +38,7 @@ export function createMockVariantAnalysis(
     databases: {
       repositories: ['1', '2', '3'],
     },
-    executionStartTime: faker.datatype.number(),
+    executionStartTime,
     createdAt: faker.date.recent().toISOString(),
     updatedAt: faker.date.recent().toISOString(),
     status: status,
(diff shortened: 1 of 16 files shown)


extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/variant-analysis.ts
```diff
@@ -10,11 +10,17 @@ import { createMockScannedRepos } from './scanned-repositories';
...
+  status = VariantAnalysisStatus.InProgress,
+  scannedRepos = createMockScannedRepos(),
+  skippedRepos = createMockSkippedRepos(),
+  executionStartTime = faker.datatype.number()
+}: {
+  status?: VariantAnalysisStatus,
+  scannedRepos?: VariantAnalysisScannedRepository[],
+  skippedRepos?: VariantAnalysisSkippedRepositories,
+  executionStartTime?: number | undefined
+}): VariantAnalysis {
```
koesie10, line 23: Thanks for making this change, this makes it much easier to use!

An alternative to this would be to allow passing in all properties using something like this:

```ts
export function createMockVariantAnalysis(data: Partial<VariantAnalysis>): VariantAnalysis {
  return {
    status: VariantAnalysisStatus.InProgress,
    // ... other defaults
    ...data,
  };
}
```

What do you think?

koesie10 (approved):
LGTM, thanks for the multiple commits, those really make it easier to review!

extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/variant-analysis.ts
elenatanasoiu: Nice! I gave it a shot but I think it would involve declaring these params:
```
scannedRepos = createMockScannedRepos(),
skippedRepos = createMockSkippedRepos(),
```
outside of the method call and passing them in. That would require a lot more refactoring and I'm a bit refactored out. 😄 

I considered adding it any way:
```
export function createMockVariantAnalysis({
  status = VariantAnalysisStatus.InProgress,
  scannedRepos = createMockScannedRepos(),
  skippedRepos = createMockSkippedRepos(),
  executionStartTime = faker.datatype.number(),
  data,
}: {
  status?: VariantAnalysisStatus,
  scannedRepos?: VariantAnalysisScannedRepository[],
  skippedRepos?: VariantAnalysisSkippedRepositories,
  executionStartTime?: number | undefined,
  data?: Partial<VariantAnalysis>
}): VariantAnalysis {
  return {
    // all the existing stuff
    ...data
  }
}
```

but that seems like it would defeat the point of it.

elenatanasoiu:
Thanks for the review @koesie10 !

extensions/ql-vscode/src/vscode-tests/factories/remote-queries/shared/variant-analysis.ts
elenatanasoiu: Ok I've thought this a bit more. I'll try in my next PR to see if we could adapt it.

PR #1716, opened 2022-11-04
Remove `gh-api` usage from variant analysis code

koesie10:
This removes all usages of the `gh-api` types from the variant analysis code by replacing it by the same types defined in `shared`.

This is a breaking change for the query history since the files serialized to disk now also change. However, since this is still behind a feature flag the change should be safe to make now.

## Checklist


elenatanasoiu (approved):

robertbrignull:
> This is a breaking change for the query history since the files serialized to disk now also change. However, since this is still behind a feature flag the change should be safe to make now.

I agree this is fine now since live results is still behind a feature flag. Long term I wonder if we've made a mistake by putting the `VariantAnalysis` object into the serialized query history. It means every change we make to that object breaks serialization. I'll open an issue internally to prompt us to think about it more.

PR #1720, opened 2022-11-07
Add cancelling of variant analysis to view

koesie10:
This implements the "Stop query" button on the view. It moves some of the logic of actually cancelling the variant analysis to the manager instead of being in the query history to allow better re-use of the code.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
@@ -22,8 +22,9 @@ import { VariantAnalysisResultsManager } from './variant-analysis-results-manage
 import { getControllerRepo } from './run-remote-query';
 import { processUpdatedVariantAnalysis, processVariantAnalysisRepositoryTask } from './variant-analysis-processor';
 import PQueue from 'p-queue';
-import { createTimestampFile, showAndLogErrorMessage } from '../helpers';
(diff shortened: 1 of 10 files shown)


extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -281,6 +282,25 @@ export class VariantAnalysisManager extends DisposableObject implements VariantA
...
   }
 
+  public async cancelVariantAnalysis(variantAnalysisId: number) {
+    const variantAnalysis = this.variantAnalyses.get(variantAnalysisId);
+    if (!variantAnalysis) {
+      throw new Error(`No variant analysis with id: ${variantAnalysisId}`);
+    }
+
+    if (!variantAnalysis.actionsWorkflowRunId) {
+      throw new Error(`No workflow run id for variant analysis ${variantAnalysis.query.name}`);
```
robertbrignull, line 292:
```suggestion
      throw new Error(`No workflow run id for variant analysis ${variantAnalysis.id}`);
```
Is it better to print the ID or the query name? I think in other places we use the ID to identify the variant analysis, even in error messages that may be visible to the user. This error hopefully shouldn't ever trigger unless there's a bug in the code, because we shouldn't show the option to cancel it if there's no actions workflow run ID yet.

robertbrignull (approved):
One comment but otherwise LGTM and worked when I tried it

PR #1721, opened 2022-11-07
Run all unit tests on Linux/macOS

koesie10:
It seems like the expansion of the test files pattern is different between Windows and Linux/macOS. This fixes it by allowing Mocha to expand the glob pattern rather than the shell which should fix the inconsistency.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -1278,7 +1278,7 @@
     "watch:webpack": "gulp watchView",
     "watch:files": "gulp watchTestData",
     "test": "npm-run-all -p test:*",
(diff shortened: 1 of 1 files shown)


aeisenberg (approved):
This makes sense to me. Thanks for noticing and fixing the problem. I think the problem is that the globstar operator isn't recognized in linux by default. A special environment variable needs to be set to enable it. So, `test/pure-tests/**/*.ts` will be interpreted by the shell as an empty set of files since no files match the glob pattern. But `'test/pure-tests/**/*.ts'` is interpreted as the raw string, which gets sent as-is to mocha which performs its own glob resolution.

charisk (approved):
LGTM! It works locally and I can see all the tests running in the Action checks.

PR #1735, opened 2022-11-08
Contextual queries: Support running when the library pack is in the package cache

adityasharad:
If the library pack containing the AST query does not have a lock file, it is likely to be in the package cache, not
a checkout of the CodeQL repo.
In this case, use `codeql pack resolve-dependencies` to create a temporary lock file, and `codeql pack install`
to install the dependencies of this library pack.

This allows the CLI to resolve the library path and dependencies for the AST query before running it, in the use case where the standard library packs are installed in the package cache rather than present in the workspace. Otherwise, in the absence of a lock file, `codeql resolve library-path` finds the standard library pack that contains the AST query, but not the `codeql/ssa` shared pack that this library pack depends on.

### Reviewer notes
- I've tried to keep this change minimal so that it doesn't affect the usual use case where the `codeql` repo is present as a workspace folder, containing library packs with lockfiles, and the AST queries are obtained from there.
  - Expanded to include find references and jump to definition, not just the AST viewer.
  - Tested in the usual starter workspace, and in a workspace where the library pack is in the package cache and not the workspace.
- This didn't work consistently when first run in a debug extension, which makes me think there may be race conditions where we attempt to run the query before the lock file is ready on disk. Is it possible to wait for the two pack-related CLI commands to complete? Suggestions welcome.
  - Fixed by review suggestion of clearing the CLI server pack cache.

## Checklist

- [N/A] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [N/A] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -970,9 +970,9 @@ export class CodeQLCliServer implements Disposable {
     }
   }
 
-  async packResolveDependencies(dir: string, mode: 'use-lock'): Promise<{ [pack: string]: string }> {
-    const args = ['--mode', mode, dir];
-    const results: { [pack: string]: string } = await this.runJsonCodeQlCliCommand(['pack', 'resolve-dependencies'], args, 'Resolving pack dependencies');
+  async packResolveDependencies(dir: string): Promise<{ [pack: string]: string }> {
+    // Uses the default `--mode use-lock`, which creates the lock file if it doesn't exist.
+    const results: { [pack: string]: string } = await this.runJsonCodeQlCliCommand(['pack', 'resolve-dependencies'], [dir], 'Resolving pack dependencies');
     return results;
   }
 
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -970,6 +970,12 @@ export class CodeQLCliServer implements Disposable {
     }
   }
 
+  async packResolveDependencies(dir: string, mode: 'use-lock'): Promise<{ [pack: string]: string }> {
+    const args = ['--mode', mode, dir];
+    const results: { [pack: string]: string } = await this.runJsonCodeQlCliCommand(['pack', 'resolve-dependencies'], args, 'Resolving pack dependencies');
+    return results;
+  }
+
   async generateDil(qloFile: string, outFile: string): Promise<void> {
     const extraArgs = await this.cliConstraints.supportsDecompileDil()
       ? ['--kind', 'dil', '-o', outFile, qloFile]
--- a/extensions/ql-vscode/src/contextual/queryResolver.ts
+++ b/extensions/ql-vscode/src/contextual/queryResolver.ts
@@ -1,6 +1,7 @@
 import * as fs from 'fs-extra';
 import * as yaml from 'js-yaml';
 import * as tmp from 'tmp-promise';
+import * as path from 'path';
 
 import * as helpers from '../helpers';
 import {
@@ -12,6 +13,11 @@ import {
 import { CodeQLCliServer } from '../cli';
 import { DatabaseItem } from '../databases';
 import { QlPacksForLanguage } from '../helpers';
+import { logger } from '../logging';
+import { createInitialQueryInfo } from '../run-queries-shared';
+import { CancellationToken, Uri } from 'vscode';
+import { ProgressCallback } from '../commandRunner';
+import { QueryRunner } from '../queryRunner';
 
 export async function qlpackOfDatabase(cli: CodeQLCliServer, db: DatabaseItem): Promise<QlPacksForLanguage> {
   if (db.contents === undefined) {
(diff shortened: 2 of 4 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -970,6 +970,12 @@ export class CodeQLCliServer implements Disposable {
     }
   }
 
+  async packResolveDependencies(dir: string, mode: 'use-lock'): Promise<{ [pack: string]: string }> {
```
aeisenberg, line 973: I think this is what you want:

```suggestion
  async packResolveDependencies(dir: string, mode = 'use-lock'): Promise<{ [pack: string]: string }> {
```

extensions/ql-vscode/src/contextual/templateProvider.ts
```diff
@@ -199,6 +211,31 @@ export class TemplatePrintAstProvider {
...
+    // The AST viewer queries now live within the standard library packs.
+    // This simplifies distribution (you don't need the standard query pack to use the AST viewer),
+    // but if the library pack doesn't have a lockfile, we won't be able to find
+    // other pack dependencies of the library pack.
+
+    // Work out the enclosing pack.
+    const packContents = await this.cli.packPacklist(query, false);
+    const packFilePath = packContents.find((p) => ['codeql-pack.yml', 'qlpack.yml'].includes(path.basename(p)));
+    if (packFilePath === undefined) {
+      // Should not happen; we already resolved this query.
```
aeisenberg, line 223: Is this always true? What happens if you try to run a query in a directory that's outside of a qlpack?

extensions/ql-vscode/src/contextual/templateProvider.ts
```diff
@@ -199,6 +211,31 @@ export class TemplatePrintAstProvider {
...
+      throw new Error(`Could not find a CodeQL pack file for the pack enclosing the query ${query}`);
+    }
+    const packPath = path.dirname(packFilePath);
+    const lockFilePath = packContents.find((p) => ['codeql-pack.lock.yml', 'qlpack.lock.yml'].includes(path.basename(p)));
+    if (!lockFilePath) {
+      // No lock file, likely because this library pack is in the package cache.
+      // Create a lock file so that we can resolve dependencies and library path
+      // for the AST query.
+      void logger.log(`Library pack ${packPath} is missing a lock file; creating a temporary lock file`);
+      await this.cli.packResolveDependencies(packPath, 'use-lock');
```
aeisenberg, line 233: `use-lock` is the default. No need to specify it. This mode will use a lock file if it exists or create one if it doesn't.

extensions/ql-vscode/src/contextual/templateProvider.ts
```diff
@@ -199,6 +211,31 @@ export class TemplatePrintAstProvider {
...
+      // Install dependencies.
+      void logger.log(`Installing package dependencies for library pack ${packPath}`);
+      await this.cli.packInstall(packPath);
```
aeisenberg, line 236: Likely you need to clear the query server cache before running. That probably explains why it wasn't working for you earlier. The cache is cleared every 2 seconds or so. 

```suggestion
      await this.cli.clearCache();
      await this.cli.packInstall(packPath);
```

If you don't clear the cache, then it will be using an old qlpack instance that will not know about the new lock file.

extensions/ql-vscode/src/contextual/templateProvider.ts
```diff
@@ -208,20 +245,34 @@ export class TemplatePrintAstProvider {
...
+      token,
+      templates
+    );
+
+    if (!lockFilePath) {
+      // Clean up the temporary lock file we created.
+      const tempLockFilePath = path.resolve(packPath, 'codeql-pack.lock.yml');
+      void logger.log(`Deleting temporary package lock file at ${tempLockFilePath}`);
+      // It's fine if the file doesn't exist.
+      fs.rmSync(path.resolve(packPath, 'codeql-pack.lock.yml'), { force: true });
```
aeisenberg, line 263: Can you make this function async and awaited?

aeisenberg (commented):
You will need to do something similar for find declarations and find references, but that is probably less important.

extensions/ql-vscode/src/contextual/templateProvider.ts
adityasharad: Good point. I had `verify` before but realised I didn't need it.

extensions/ql-vscode/src/contextual/templateProvider.ts
adityasharad: Not generally true, but I hope it is always true for our own contextual queries.

extensions/ql-vscode/src/contextual/templateProvider.ts
adityasharad: I had no idea. Thank you!

extensions/ql-vscode/src/contextual/queryResolver.ts
```diff
@@ -104,3 +110,67 @@ export async function resolveQueries(cli: CodeQLCliServer, qlpacks: QlPacksForLa
...
+  void logger.log(`Running contextual query ${query}; results will be stored in ${queryStorageDir}`);
+  const queryResult = await qs.compileAndRunQueryAgainstDatabase(
+    db,
+    initialInfo,
+    queryStorageDir,
+    progress,
+    token,
+    templates
+  );
+  if (!lockFilePath) {
```
aeisenberg, line 172: Oh...this confused me for a minute. I know it's unchanged from before, but maybe you can change the return value from `resolveContextualQuery` to have a boolean called something like `createdTempLockFile` or something.

extensions/ql-vscode/src/contextual/queryResolver.ts
```diff
@@ -104,3 +110,67 @@ export async function resolveQueries(cli: CodeQLCliServer, qlpacks: QlPacksForLa
...
+    await cli.packInstall(packPath);
+  }
+  return { packPath, lockFilePath };
+}
+
+async function removeTemporaryLockFile(packPath: string) {
+  const tempLockFilePath = path.resolve(packPath, 'codeql-pack.lock.yml');
+  void logger.log(`Deleting temporary package lock file at ${tempLockFilePath}`);
+  // It's fine if the file doesn't exist.
+  await fs.promises.rm(path.resolve(packPath, 'codeql-pack.lock.yml'), { force: true });
```
aeisenberg, line 150: Minor: this should work.

```suggestion
  await fs.rm(path.resolve(packPath, 'codeql-pack.lock.yml'), { force: true });
```

aeisenberg (approved):

PR #1746, opened 2022-11-11
Change remote queries to test against submitted data

koesie10:
The remote queries tests were testing the data on the filesystem, rather than the data submitted to the server. This required using a `dryRun` parameter to prevent deleting the temporary directory, while we can actually just test against the submitted data.

This will create an in-memory filesystem of the submitted query pack by un-tar-gz'ing the query pack into memory and using that to test the existence of certain files.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -1380,6 +1380,7 @@
     "@types/sinon-chai": "~3.2.3",
     "@types/stream-chain": "~2.0.1",
     "@types/stream-json": "~1.7.1",
+    "@types/tar-stream": "^2.2.2",
     "@types/through2": "^2.0.36",
     "@types/tmp": "^0.1.0",
     "@types/unzipper": "~0.10.1",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -1421,6 +1421,7 @@
     "proxyquire": "~2.1.3",
     "sinon": "~14.0.0",
     "sinon-chai": "~3.5.0",
+    "tar-stream": "^2.2.0",
     "through2": "^4.0.2",
     "ts-jest": "^29.0.1",
     "ts-json-schema-generator": "^1.1.2",
--- a/extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts
+++ b/extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts
@@ -2,7 +2,6 @@ import { assert, expect } from 'chai';
 import * as path from 'path';
 import * as sinon from 'sinon';
 import { CancellationTokenSource, ExtensionContext, extensions, QuickPickItem, Uri, window } from 'vscode';
-import * as fs from 'fs-extra';
 import * as os from 'os';
 import * as yaml from 'js-yaml';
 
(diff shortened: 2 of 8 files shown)


extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts
```diff
@@ -353,8 +328,8 @@ describe('Remote queries', function() {
     });
   });
 
-  function verifyQlPack(qlpackPath: string, queryPath: string, packVersion: string, pathSerializationBroken: boolean) {
-    const qlPack = yaml.load(fs.readFileSync(qlpackPath, 'utf8')) as QlPack;
+  function verifyQlPack(queryPath: string, contents: Buffer, packVersion: string, pathSerializationBroken: boolean) {
```
charisk, line 331: Should `queryPath` be `queryPackPath`?

extensions/ql-vscode/src/vscode-tests/utils/bundled-pack-helpers.ts
```diff
@@ -0,0 +1,76 @@
...
+import { pipeline } from 'stream/promises';
+import { createGunzip } from 'zlib';
+
+export interface QueryPackFS {
+  fileExists: (name: string) => boolean;
+  fileContents: (name: string) => Buffer;
+  directoryContents: (name: string) => string[];
+}
+
+export const readBundledPack = async (base64Pack: string): Promise<QueryPackFS> => {
```
charisk, line 13: (totally minor and optional) I think we've generally followed the `export function()` style so might be good to stick to that.

extensions/ql-vscode/package.json
```diff
@@ -1380,6 +1380,7 @@
     "@types/sinon-chai": "~3.2.3",
     "@types/stream-chain": "~2.0.1",
     "@types/stream-json": "~1.7.1",
+    "@types/tar-stream": "^2.2.2",
```
charisk, line 1383: It looks like we're just bringing in the types module, do we need the actual module too? 🤔

charisk (approved):
Looking good, thanks for improving tests!

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts
koesie10: I don't think so, this is not the path to the query pack, but the path to the query which is mentioned within the query pack. For example, it would be a YAML file like:

```yaml
name: codeql-remote/query
version: 0.0.0
dependencies:
   codeql/javascript-all: '*'
library: false
defaultSuite:
  - description: Query suite for variant analysis
  - query: in-pack.ql
```

The last `in-pack.ql` is the `queryPath`.

extensions/ql-vscode/package.json
koesie10: Good point! Although it's already included as a transitive dependency, I'll also add it as a dependency itself too.

extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts
charisk: Oh I see! Thanks!

PR #1748, opened 2022-11-11
Reduce nesting in `runRemoteQuery`

koesie10:
Now that we do not have a dry run mode, we can create and clean up the temporary directory in the same function. This allows us to remove the complete `try`..`finally` block inside `runRemoteQuery` and move it to a much more local spot.

It's easiest to review this by hiding whitespace, this should only show the actual changes.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
+++ b/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
@@ -181,7 +181,6 @@ export async function prepareRemoteQueryRun(
   cliServer: cli.CodeQLCliServer,
   credentials: Credentials,
   uri: Uri | undefined,
-  queryPackDir: string,
   progress: ProgressCallback,
   token: CancellationToken,
 ) {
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -225,7 +224,17 @@ export async function prepareRemoteQueryRun(
     throw new UserCancellationException('Cancelled');
   }
 
-  const { base64Pack, language } = await generateQueryPack(cliServer, queryFile, queryPackDir);
+  const { remoteQueryDir, queryPackDir } = await createRemoteQueriesTempDirectory();
+
+  let pack: Awaited<ReturnType<typeof generateQueryPack>>;
```
charisk, line 229: This is fine but I personally generally avoid using this sort of type and go with defining a type for the return type of the function. E.g. have a `QueryPack` interface and use that in the `generateQueryPack` signature as well as the definition of the `pack` variable.

What do you think of updating `generateQueryPack` to return a pre-defined t

charisk (approved):

PR #1749, opened 2022-11-11
Split flows for running remote queries

koesie10:
This removes the `runRemoteQuery` method and instead moves all logic specific to remote queries/variant analysis to the remote queries manager and variant analysis manager respectively. This will make it easier to completely remove the remote queries manager in the future.

## Checklist


extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
```diff
@@ -119,21 +119,47 @@ export class RemoteQueriesManager extends DisposableObject {
...
-      this.remoteQueryAddedEventEmitter.fire({ queryId, query });
-      void commands.executeCommand('codeQL.monitorRemoteQuery', queryId, query);
-    }
+    await this.prepareStorageDirectory(queryId);
+    await this.storeJsonFile(queryId, 'query.json', query);
+
+    this.remoteQueryAddedEventEmitter.fire({ queryId, query });
+    void commands.executeCommand('codeQL.monitorRemoteQuery', queryId, query);
+
+    return query;
```
charisk, line 162: Seems like this is only used in tests - is that correct? I think we should generally avoid having some code just for our tests but appreciate that would probably require a bunch more changes to avoid it.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -262,98 +257,7 @@ export async function prepareRemoteQueryRun(
...
-      language,
-      repositoryCount);
-
-    // don't return the path because it has been deleted
-    return { query: remoteQuery };
-  }
-}
-
-async function runRemoteQueriesApiRequest(
+export async function runRemoteQueriesApiRequest(
```
charisk, line 260: Can this and `buildRemoteQueryEntity` be moved to the `RemoteQueriesManager` too?

charisk (approved):

extensions/ql-vscode/src/remote-queries/remote-queries-manager.ts
koesie10: Thanks for mentioning this! I completely agree that we should not be doing this and I don't think it'll be a lot of work to remove (I can instead check that the `monitorRemoteQuery` command has been called), so I'll remove it now rather than leaving it in.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
koesie10: Yes, these can be moved out of this file. I'd prefer to not place them into the `RemoteQueriesManager` itself since it doesn't depend on the state of that class, but I'll move the `buildRemoteQueryEntity` into `remote-query.ts` and create a new file for this method in `remote-queries-api.ts`.

PR #1750, opened 2022-11-11
Add exporting of results from the context menu

koesie10:
This will add exporting of results from the context menu for variant analyses.

Please review this commit-by-commit: the first commit is in preparation for the second commit. Also see the commit messages for more context on these changes.

## Checklist


extensions/ql-vscode/src/remote-queries/export-results.ts
```diff
@@ -58,32 +67,111 @@ export async function exportRemoteQueryResults(
...
+        yield [repo, {
+          variantAnalysisId: variantAnalysis.id,
+          repositoryId: repo.repository.id,
+        }];
+        continue;
+      }
+
+      let result: VariantAnalysisScannedRepositoryResult;
+
+      if (!variantAnalysisManager.areResultsLoaded(variantAnalysis.id, repo.repository.fullName)) {
```
robertbrignull, line 131: Should this whole `if` block be a method on `VariantAnalysisManager` so outside code can see a cleaner interface and doesn't need to know it has to check `areResultsLoaded` first?

extensions/ql-vscode/src/remote-queries/export-results.ts
```diff
@@ -1,41 +1,50 @@
...
- * Exports the results of the currently-selected remote query.
+ * Exports the results of the currently-selected remote query or variant analysis.
  */
 export async function exportSelectedRemoteQueryResults(queryHistoryManager: QueryHistoryManager): Promise<void> {
   const queryHistoryItem = queryHistoryManager.getCurrentQueryHistoryItem();
   if (!queryHistoryItem || queryHistoryItem.t === 'local') {
     throw new Error('No variant analysis results currently open. To open results, click an item in the query history view.');
   }
 
-  if (!queryHistoryItem.completed) {
```
elenatanasoiu: Don't we still want to perform this check for remote queries?

extensions/ql-vscode/src/remote-queries/export-results.ts
koesie10: 👍, I've added a new `options` argument to the `loadResults` method, so all of this is now contained in the variant analysis results manager.

extensions/ql-vscode/src/query-history.ts
```diff
@@ -1267,8 +1267,22 @@ export class QueryHistoryManager extends DisposableObject {
...
+    if (!this.assertSingleQuery(finalMultiSelect) || !finalSingleItem) {
+      return;
+    }
+
+    // Remote queries and variant analysis only
+    if (finalSingleItem.t === 'remote') {
+      await commands.executeCommand('codeQL.exportRemoteQueryResults', finalSingleItem.queryId);
+    } else if (finalSingleItem.t === 'variant-analysis') {
+      await commands.executeCommand('codeQL.exportVariantAnalysisResults', finalSingleItem.variantAnalysis.id);
+    }
```
elenatanasoiu, line 1285: Is there room for a quick unit test [like we have](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/vscode-tests/no-workspace/query-history.test.ts#L402-L410) for `handleCancel`?

extensions/ql-vscode/src/remote-queries/export-results.ts
koesie10: Yes, we do still need it for the remote queries. However, I felt it was redundant to do it twice, so this is now only checked in [the command we call from this method](https://github.com/github/vscode-codeql/blob/e9681bc5468d846a1549f7e4abb1afbdd5f3a649/extensions/ql-vscode/src/remote-queries/export-results.ts#L62-L64). This should result in the same behaviour for the user.

extensions/ql-vscode/src/query-history.ts
koesie10: Thanks for the pointer! I've added unit tests for both this method and the `handleCopyRepoList` which does basically the same thing.

extensions/ql-vscode/src/remote-queries/export-results.ts
elenatanasoiu: Ah I see!

extensions/ql-vscode/src/query-history.ts
elenatanasoiu: This is excellent! 🙌  Thanks @koesie10 !

elenatanasoiu (approved):

PR #1751, opened 2022-11-11
Add support for local dbs in config

shati-patel:
Adds support for local databases the config for the new databases panel. Note: this only changes the config - we don't yet render the local items (that's a follow-up issue).

See internal issue for context!

## Checklist

N/A - internal only 🖌️ 

### DIFF
--- a/extensions/ql-vscode/src/databases/db-config.ts
+++ b/extensions/ql-vscode/src/databases/db-config.ts
@@ -30,7 +30,7 @@ export interface LocalDatabase {
   name: string;
   dateAdded: number;
   language: string;
-  path: string;
+  storagePath: string;
 }
 
 export function cloneDbConfig(config: DbConfig): DbConfig {
@@ -46,19 +46,9 @@ export function cloneDbConfig(config: DbConfig): DbConfig {
     local: {
       lists: config.local.lists.map((list) => ({
         name: list.name,
-        databases: list.databases.map((db) => ({
-          name: db.name,
-          dateAdded: db.dateAdded,
-          language: db.language,
-          path: db.path,
-        })),
+        databases: list.databases.map((db) => ({ ...db })),
       })),
-      databases: config.local.databases.map((db) => ({
-        name: db.name,
-        dateAdded: db.dateAdded,
-        language: db.language,
-        path: db.path,
-      })),
+      databases: config.local.databases.map((db) => ({ ...db })),
     },
   };
 }
--- a/extensions/ql-vscode/src/databases/db-config.ts
+++ b/extensions/ql-vscode/src/databases/db-config.ts
@@ -2,6 +2,7 @@
 
 export interface DbConfig {
   remote: RemoteDbConfig;
+  local: LocalDbConfig;
 }
 
 export interface RemoteDbConfig {
@@ -15,6 +16,23 @@ export interface RemoteRepositoryList {
   repositories: string[];
 }
 
+export interface LocalDbConfig {
+  lists: LocalRepositoryList[];
+  databases: LocalDatabase[];
+}
+
+export interface LocalRepositoryList {
+  name: string;
+  databases: LocalDatabase[];
+}
+
+export interface LocalDatabase {
+  name: string;
+  dateAdded: number;
+  language: string;
+  path: string;
+}
+
 export function cloneDbConfig(config: DbConfig): DbConfig {
   return {
     remote: {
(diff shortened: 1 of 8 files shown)


extensions/ql-vscode/src/databases/db-config.ts
```diff
@@ -15,6 +16,23 @@ export interface RemoteRepositoryList {
...
+export interface LocalRepositoryList {
+  name: string;
+  databases: LocalDatabase[];
+}
+
+export interface LocalDatabase {
+  name: string;
+  dateAdded: number;
+  language: string;
+  path: string;
```
charisk, line 33: How about `storagePath` instead?

extensions/ql-vscode/src/databases/db-config.ts
```diff
@@ -24,6 +42,23 @@ export function cloneDbConfig(config: DbConfig): DbConfig {
       })),
       owners: [...config.remote.owners],
       repositories: [...config.remote.repositories],
-    }
+    },
+    local: {
+      lists: config.local.lists.map((list) => ({
+        name: list.name,
+        databases: list.databases.map((db) => ({
```
charisk, line 49: I think this would work too:
```ts
databases: list.databases.map((db) => ({ ...db }));
```

extensions/ql-vscode/test/pure-tests/databases/data/workspace-databases.json
```diff
@@ -8,5 +8,39 @@
...
+  "local": {
+    "lists": [
+      {
+        "name": "localList1",
+        "databases": [
+          {
+            "name": "foo/bar",
+            "dateAdded": 1668096745193,
+            "language": "go",
+            "path": "Uri to db in workspace storage"
```
charisk, line 21: I guess it could just be the file system path rather than a URI? What do we store currently?

extensions/ql-vscode/test/pure-tests/databases/db-config-validator.test.ts
```diff
@@ -25,9 +25,10 @@ describe('db config validation', async () => {
 
     const validationOutput = configValidator.validate(dbConfig);
 
-    expect(validationOutput).to.have.length(2);
+    expect(validationOutput).to.have.length(3);
```
charisk, line 28: Shall we check the last validation error too? Or put this back to 2 and add local stuff to dbConfig on L13

extensions/ql-vscode/workspace-databases-schema.json
```diff
@@ -44,6 +44,72 @@
...
+                  "type": "object",
+                  "properties": {
+                    "name": {
+                      "type": "string"
+                    },
+                    "dateAdded": {
+                      "type": "number"
+                    },
+                    "language": {
+                      "type": "string"
```
charisk, line 71: We could force to only allow certain values to be set here e.g. javascript etc. 

You can do that using `enum` in JSON schema e.g. `"enum": ["javascript", "..."]`

charisk (commented):
Looking good so far, just some small comments!

extensions/ql-vscode/test/pure-tests/databases/data/workspace-databases.json
shati-patel: We currently store a `Uri`: https://github.com/github/vscode-codeql/blob/7e8fa5c5fe39f89273d817ae09b74aac0bb4a5fb/extensions/ql-vscode/src/databases.ts#L206

We could change to a file system path, but I wonder if that would make it harder to integrate with existing stuff?

charisk: Hmm `Uri` is a VS Code thing so I'm thinking we should stick to filesystem paths. We can easily translate to `vscode.Uri` by doing `vscode.Uri.file('/path/to/dir')`.

extensions/ql-vscode/workspace-databases-schema.json
shati-patel: I was wondering about this too, but I'm currently not actually sure what languages to include 😅 E.g. if someone's working on CodeQL support for a new language, or doing something with a custom version of the CLI, we don't necessarily know what language their database is for 🤔 

I'm thinking that we'd check the language anyway (when validating against the dbscheme), so it might be fine to keep as `string`? 

(I might be overthinking this though 😆 )

extensions/ql-vscode/test/pure-tests/databases/data/workspace-databases.json
shati-patel: Ah, good point. Will change that 👍🏽

shati-patel:
I'm going to add some more local cases to the tests, and then I'll take this out of draft :)

extensions/ql-vscode/workspace-databases-schema.json
charisk: Does the extension "jus work" if someone is working on support for a new language?

shati-patel: > Does the extension "just work" if someone is working on support for a new language?

Yep, language support doesn't depend on the extension (unless we've hard-coded languages somewhere), so people should be able to use custom CLI or databases without issue (at least locally)!

charisk: Hmm okay let's leave it as a string for now but we might need to revisit.

charisk (approved):
LGTM!

PR #1753, opened 2022-11-11
Add copying of repository list for variant analyses

koesie10:
This adds the ability to copy the repository list for variant analyses from the context menu in the query history.

## Checklist


extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
```diff
@@ -695,4 +700,121 @@ describe('Variant Analysis Manager', async function() {
       });
     });
   });
+
+  describe('copyRepoListToClipboard', async () => {
```
charisk, line 704: Can these tests be in no-workspace or minimal-workspace at all?

koesie10: Yes, I think they could be moved to `minimal-workspace`. However, I don't really see what the advantage would be of splitting these tests over multiple files and test suites. I think it would actually make it harder where something is being tested, although I would also be happy to move them over if there are advantages to that.

charisk: My understanding is that CLI tests are slower (and maybe flakier?) because they have to set up the whole CLI. We generally need to tidy up around the area so I'm happy for us to leave things where they are for now, but I was curious about whether there was a reason we chose to put them here. I guess the reason is that other `VariantAnalysisManager` tests are in CLI tests.

charisk (approved):

PR #1754, opened 2022-11-11
Add event emitter abstraction

charisk:
Currently, classes/modules that interact with events (raising or catching) have to import [Event](https://code.visualstudio.com/api/references/vscode-api#Event%3CT%3E)/[EventEmitter](https://code.visualstudio.com/api/references/vscode-api#EventEmitter%3CT%3E) which couples them to the VS Code API and as a consequence it means we're no longer able to write unit tests. It pushes us towards writing slow integration tests that require the whole extension host. 

This PR introduces some custom abstractions that we can use instead of bringing in the VS Code API. 

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/common/events.ts
+++ b/extensions/ql-vscode/src/common/events.ts
@@ -1,7 +1,7 @@
 import { Disposable } from '../pure/disposable-object';
 
 export interface AppEvent<T> {
-  (listener: (event: T) => any): Disposable;
+  (listener: (event: T) => void): Disposable;
 }
 
 export interface AppEventEmitter<T> {
--- a/extensions/ql-vscode/src/common/events.ts
+++ b/extensions/ql-vscode/src/common/events.ts
@@ -1,0 +1,10 @@
+import { Disposable } from '../pure/disposable-object';
+
+export interface AppEvent<T> {
+  (listener: (event: T) => any): Disposable;
+}
+
+export interface AppEventEmitter<T> {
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/common/events.ts
```diff
@@ -0,0 +1,10 @@
+import { Disposable } from '../pure/disposable-object';
+
+export interface AppEvent<T> {
+  (listener: (event: T) => any): Disposable;
```
koesie10, line 4: Can we make the return type of the `listener` either `void` or `unknown`? In theory it shouldn't return anything because we won't be using the result anyway, but I think it's good to make it explicit and explicitly require you to check the type.

charisk: Yeah I can't see why not!

koesie10 (approved):

adityasharad:
`DisposableObject` exists for a similar reason. Should we move all these abstractions into a common location, either `common/vscode` or `pure`?

charisk:
> `DisposableObject` exists for a similar reason. Should we move all these abstractions into a common location, either `common/vscode` or `pure`?

My plan/hope is to eventually move away from the term "pure" as it's a little bit confusing (makes me think of [pure functions](https://en.wikipedia.org/wiki/Pure_function)) and have `/vscode` in the location/name. I wanted to use this area as an example to make the case for it a bit more too, so expect to see/hear about it!

PR #1756, opened 2022-11-14
Add extension and storage paths to App container

charisk:
Added some information typically retrieved from the extension context, to our new App container and updated the `DbConfigStore` to use this new App container.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/common/vscode/vscode-app.ts
+++ b/extensions/ql-vscode/src/common/vscode/vscode-app.ts
@@ -5,18 +5,20 @@ import { VSCodeAppEventEmitter } from './events';
 
 export class ExtensionApp implements App {
   public constructor(
-    public readonly extensionPath: string,
-    public readonly globalStoragePath: string,
-    public readonly workspaceStoragePath?: string
+    public readonly extensionContext: vscode.ExtensionContext
   ) {
   }
 
-  public static createFromExtensionContext(extensionContext: vscode.ExtensionContext): ExtensionApp {
-    return new ExtensionApp(
-      extensionContext.extensionPath,
-      extensionContext.globalStorageUri.fsPath,
-      extensionContext.storageUri?.fsPath
-    );
+  public get extensionPath(): string {
+    return this.extensionContext.extensionPath;
+  }
+
+  public get globalStoragePath(): string {
+    return this.extensionContext.globalStorageUri.fsPath;
+  }
+
+  public get workspaceStoragePath(): string | undefined {
+    return this.extensionContext.storageUri?.fsPath;
   }
 
   public createEventEmitter<T>(): AppEventEmitter<T> {
--- a/extensions/ql-vscode/src/common/vscode/vscode-app.ts
+++ b/extensions/ql-vscode/src/common/vscode/vscode-app.ts
@@ -11,7 +11,7 @@ export class ExtensionApp implements App {
   ) {
   }
 
-  public static CreateFromExtensionContext(extensionContext: vscode.ExtensionContext): ExtensionApp {
+  public static createFromExtensionContext(extensionContext: vscode.ExtensionContext): ExtensionApp {
     return new ExtensionApp(
       extensionContext.extensionPath,
       extensionContext.globalStorageUri.fsPath,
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/common/vscode/vscode-app.ts
```diff
@@ -1,8 +1,24 @@
...
 
 export class ExtensionApp implements App {
+  public constructor(
+    public readonly extensionPath: string,
+    public readonly globalStoragePath: string,
+    public readonly workspaceStoragePath?: string
+  ) {
+  }
+
+  public static CreateFromExtensionContext(extensionContext: vscode.ExtensionContext): ExtensionApp {
```
koesie10, line 14:
```suggestion
  public static createFromExtensionContext(extensionContext: vscode.ExtensionContext): ExtensionApp {
```

extensions/ql-vscode/src/common/vscode/vscode-app.ts
```diff
@@ -1,8 +1,24 @@
+import * as vscode from 'vscode';
 import { App } from '../app';
 import { AppEventEmitter } from '../events';
 import { VSCodeAppEventEmitter } from './events';
 
 export class ExtensionApp implements App {
+  public constructor(
+    public readonly extensionPath: string,
+    public readonly globalStoragePath: string,
+    public readonly workspaceStoragePath?: string
```
koesie10, line 10: Instead of making these properties, can we make them getters and save the extension context as a property instead? Something like this:

```ts
constructor(public readonly extensionContext: vscode.ExtensionContext) {}

public get extensionPath(): string {
  return this.extensionContext.extensionPath;
}

public get globalStoragePath(): string {
  return this.extensionContext.globalStorageUri.fsPath;
}

public get workspaceStoragePath(): string | undefined {
  return this.extensionContext.storageUri?.fsPath;
}
```

I think this will make it more resilient to changes in how VSCode determines the paths. Right now, they [seem to be computed](https://github.com/microsoft/vscode/blob/c9fd4761bf159592d1611d7a1b42fd5565120358/src/vs/workbench/api/common/extHostStoragePaths.ts#L41) at extension load and never change, but since [it's a getter](https://github.com/microsoft/vscode/blob/c9fd4761bf159592d1611d7a1b42fd5565120358/src/vs/workbench/api/common/extHostExtensionService.ts#L528), it could be changed to return a different value depending on changes we don't control.

extensions/ql-vscode/src/databases/db-module.ts
```diff
@@ -21,9 +22,9 @@ export class DbModule extends DisposableObject {
 
     void logger.log('Initializing database module');
 
-    const storagePath = extensionContext.storageUri?.fsPath || extensionContext.globalStorageUri.fsPath;
-    const extensionPath = extensionContext.extensionPath;
-    const dbConfigStore = new DbConfigStore(storagePath, extensionPath);
+    const app = ExtensionApp.CreateFromExtensionContext(extensionContext);
```
koesie10, line 25: Would it make sense to create this in the `extension.ts` file instead and re-use the same instance for all classes that will eventually use this? That might also make it easier to test this class without having an extension context.

extensions/ql-vscode/src/databases/db-config-store.ts
```diff
@@ -14,17 +15,16 @@ export class DbConfigStore extends DisposableObject {
...
   private configWatcher: chokidar.FSWatcher | undefined;
 
-  public constructor(
-    workspaceStoragePath: string,
-    extensionPath: string) {
+  public constructor(app: App) {
     super();
 
-    this.configPath = path.join(workspaceStoragePath, 'workspace-databases.json');
+    const storagePath = app.workspaceStoragePath || app.globalStoragePath;
```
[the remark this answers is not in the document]
charisk: I'm not sure how much this is used yet. Shall we wait and see? I wouldn't object to it but I didn't know what to call the function and that usually makes me want to wait a bit :)

extensions/ql-vscode/src/common/vscode/vscode-app.ts
charisk: Thanks this is a good suggestion! I've made some changes.

koesie10 (approved):

PR #1761, opened 2022-11-14
Add selected to config

norascheuch:
To display selected Databases in the panel we first have to add the attribute to the workspace-databases.json file.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases/db-config.ts
+++ b/extensions/ql-vscode/src/databases/db-config.ts
@@ -2,7 +2,7 @@
 
 export interface DbConfig {
   databases: DbConfigDatabases;
-  selected?: SelectedDb;
+  selected?: SelectedDbItem;
 }
 
 export interface DbConfigDatabases {
@@ -10,9 +10,14 @@ export interface DbConfigDatabases {
   local: LocalDbConfig;
 }
 
-export interface SelectedDb {
-  kind: string;
-  path: string;
+export interface SelectedDbItem {
+  kind: SelectedDbKind;
+  value: string;
+}
+
+export enum SelectedDbKind {
+  ConfigDefined = 'configDefined',
+  RemoteSystemDefinedList = 'remoteSystemDefinedList',
 }
 
 export interface RemoteDbConfig {
@@ -64,7 +69,7 @@ export function cloneDbConfig(config: DbConfig): DbConfig {
     },
     selected: config.selected ? {
       kind: config.selected.kind,
-      path: config.selected.path
+      value: config.selected.value,
     } : undefined
   };
 }
--- a/extensions/ql-vscode/src/databases/db-config.ts
+++ b/extensions/ql-vscode/src/databases/db-config.ts
@@ -11,11 +11,11 @@ export interface DbConfigDatabases {
 }
 
 export interface SelectedDbItem {
-  kind: SelectedDbKind;
+  kind: SelectedDbItemKind;
   value: string;
 }
 
-export enum SelectedDbKind {
+export enum SelectedDbItemKind {
   ConfigDefined = 'configDefined',
   RemoteSystemDefinedList = 'remoteSystemDefinedList',
 }
--- a/extensions/ql-vscode/src/databases/db-config.ts
+++ b/extensions/ql-vscode/src/databases/db-config.ts
@@ -1,8 +1,18 @@
 // Contains models for the data we want to store in the database config
 
 export interface DbConfig {
+  databases: DbConfigDatabases;
+  selected?: SelectedDb;
+}
+
+export interface DbConfigDatabases {
   remote: RemoteDbConfig;
   local: LocalDbConfig;
+}
+
+export interface SelectedDb {
+  kind: string;
+  path: string;
 }
 
 export interface RemoteDbConfig {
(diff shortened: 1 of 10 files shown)


extensions/ql-vscode/src/databases/db-config.ts
```diff
@@ -1,10 +1,20 @@
...
+  selected?: SelectedDb;
+}
+
+export interface DbConfigDatabases {
   remote: RemoteDbConfig;
   local: LocalDbConfig;
 }
 
+export interface SelectedDb {
+  kind: string;
```
charisk, line 14: Let's add some more specific types here rather than just strings e.g. `kind` can be an enum with two possible values.

extensions/ql-vscode/src/databases/db-config.ts
```diff
@@ -1,10 +1,20 @@
...
+  path: string;
```
charisk, line 15: Since this can either be a path or a value I think we decided on the term `value`. Let's also add more specific types for this (happy to pair if it helps).

extensions/ql-vscode/test/pure-tests/databases/db-config-store.test.ts
```diff
@@ -17,39 +17,37 @@ describe('db config store', async () => {
   });
 
   it('should create a new config if one does not exist', async () => {
-    const configPath = path.join(tempWorkspaceStoragePath, 'workspace-databases.json');
-
     const configStore = new DbConfigStore(tempWorkspaceStoragePath, extensionPath);
     await configStore.initialize();
 
-    expect(await fs.pathExists(configPath)).to.be.true;
```
charisk, line 25: Why have we removed this check? I think it's good to check the file is actually created.

extensions/ql-vscode/test/pure-tests/databases/data/workspace-databases.json
```diff
@@ -1,46 +1,52 @@
...
-    ],
-    "databases": [
-      {
-        "name": "example-db",
-        "dateAdded": 1668096927267,
-        "language": "ruby",
-        "storagePath": "/path/to/database/"
-      }
-    ]
+  "selected": {
```
charisk, line 48: I think it would be nice to have some some representative/real data.

charisk (commented):
Great to see this getting implemented! Just some initial comments from me.

extensions/ql-vscode/test/pure-tests/databases/db-config-validator.test.ts
```diff
@@ -25,10 +25,9 @@ describe('db config validation', async () => {
 
     const validationOutput = configValidator.validate(dbConfig);
 
-    expect(validationOutput).to.have.length(3);
+    expect(validationOutput).to.have.length(2);
```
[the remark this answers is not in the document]
norascheuch: No, this is because the main missing object is 'databases' now, therefore only the checks on line 138-139 are returning errors.

extensions/ql-vscode/src/databases/db-config.ts
norascheuch: Ok, I wasn't sure how to exactly flesh this out in the later PRs, that's why I left it as open as possible

extensions/ql-vscode/test/pure-tests/databases/db-config-store.test.ts
norascheuch: Good catch, thanks! Will revert it back

extensions/ql-vscode/src/databases/db-config.ts
```diff
@@ -1,10 +1,20 @@
...
+  databases: DbConfigDatabases;
+  selected?: SelectedDb;
+}
+
+export interface DbConfigDatabases {
   remote: RemoteDbConfig;
   local: LocalDbConfig;
 }
 
+export interface SelectedDb {
```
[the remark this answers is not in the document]
norascheuch: For me, I think of a selection of pralines or goods when reading `selection`, definitely of a set of things. For me `SelectedDb` is the most descriptive of what it does but 🤷‍♀️

extensions/ql-vscode/src/databases/db-config.ts
```diff
@@ -1,10 +1,25 @@
...
   remote: RemoteDbConfig;
   local: LocalDbConfig;
 }
 
+export interface SelectedDbItem {
+  kind: SelectedDbKind;
+  value: string;
+}
+
+export enum SelectedDbKind {
```
charisk, line 18: I think this should be `SelectedDbItemKind` since we've gone with `SelectedDbItem`.

extensions/ql-vscode/src/databases/db-config.ts
```diff
@@ -1,10 +1,25 @@
...
+}
+
+export interface DbConfigDatabases {
   remote: RemoteDbConfig;
   local: LocalDbConfig;
 }
 
+export interface SelectedDbItem {
+  kind: SelectedDbKind;
+  value: string;
```
charisk, line 15: For me this feels like it's not going to be clear what to enter in `value` to users that want to edit the file directly.

I think I'd prefer we went with something like:
```json
{
  "selected": {
    "kind": "configDefined",
    "jsonPath": "json/path/to/item"
  }
}

{
  "selected": {
    "kind": "remoteSystemDefinedList",
    "listName": "top_10"
  }
}
```

or with a bit more nesting
```json
{
  "selected": {
    "kind": "configDefined",
    "value" : {
       "jsonPath": "json/path/to/item"
     }
  }
}
```

(I personally prefer the first one)

But if you like we could address that in a separate PR.

extensions/ql-vscode/test/pure-tests/databases/db-config-validator.test.ts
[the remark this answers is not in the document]
norascheuch: Can't say I care about it. I copied adjusting it the way Shati did in a different PR but 🤷‍♀️  will revert and so as you suggest.

extensions/ql-vscode/src/databases/db-config.ts
norascheuch: 👍

charisk (approved):

PR #1765, opened 2022-11-15
Use selected repositories for copying repo lists

koesie10:
This will use the selected repositories to limit which repositories are included in the copied repo list. If there are both selected repositories and a search filter (on the full name), the search filter will be ignored and the selected repositories will be used in full.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/pure/variant-analysis-filter-sort.ts
+++ b/extensions/ql-vscode/src/pure/variant-analysis-filter-sort.ts
@@ -76,22 +76,26 @@ export function compareWithResults(filterSortState: RepositoriesFilterSortState 
   };
 }
 
-function isFilterOnRepositoryIds(filterSortState: RepositoriesFilterSortState | RepositoriesFilterSortStateWithIds | undefined): filterSortState is RepositoriesFilterSortStateWithIds & Required<Pick<RepositoriesFilterSortStateWithIds, 'repositoryIds'>> {
-  return !!filterSortState && 'repositoryIds' in filterSortState && filterSortState.repositoryIds !== undefined && filterSortState.repositoryIds.length > 0;
-}
-
-// These define the behavior for undefined input values
-export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[], filterSortState: RepositoriesFilterSortState | RepositoriesFilterSortStateWithIds | undefined): T[];
-export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortState | RepositoriesFilterSortStateWithIds | undefined): T[] | undefined;
-
-export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortState | RepositoriesFilterSortStateWithIds | undefined): T[] | undefined {
+export function filterAndSortRepositoriesWithResultsByName<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortState | undefined): T[] | undefined {
   if (!repositories) {
     return undefined;
   }
 
-  const filteredRepositories = isFilterOnRepositoryIds(filterSortState) ?
-    repositories.filter(repo => filterSortState.repositoryIds.includes(repo.repository.id)) :
-    repositories.filter(repo => matchesFilter(repo.repository, filterSortState));
+  return repositories
+    .filter(repo => matchesFilter(repo.repository, filterSortState))
+    .sort(compareWithResults(filterSortState));
+}
 
-  return filteredRepositories.sort(compareWithResults(filterSortState));
+export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortStateWithIds | undefined): T[] | undefined {
+  if (!repositories) {
+    return undefined;
+  }
+
+  if (filterSortState?.repositoryIds && filterSortState.repositoryIds.length > 0) {
+    return repositories
+      .filter(repo => filterSortState.repositoryIds?.includes(repo.repository.id))
+      .sort(compareWithResults(filterSortState));
+  }
+
+  return filterAndSortRepositoriesWithResultsByName(repositories, filterSortState);
 }
--- a/extensions/ql-vscode/src/pure/variant-analysis-filter-sort.ts
+++ b/extensions/ql-vscode/src/pure/variant-analysis-filter-sort.ts
@@ -13,6 +13,10 @@ export type RepositoriesFilterSortState = {
   sortKey: SortKey;
 }
 
+export type RepositoriesFilterSortStateWithIds = RepositoriesFilterSortState & {
+  repositoryIds?: number[];
+}
+
 export const defaultFilterSortState: RepositoriesFilterSortState = {
   searchValue: '',
   sortKey: SortKey.Name,
@@ -52,7 +56,7 @@ export function compareRepository(filterSortState: RepositoriesFilterSortState |
 }
 
 type SortableResult = {
-  repository: SortableRepository;
+  repository: SortableRepository & Pick<Repository, 'id'>;
   resultCount?: number;
 }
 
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/pure/variant-analysis-filter-sort.ts
```diff
@@ -13,6 +13,10 @@ export type RepositoriesFilterSortState = {
   sortKey: SortKey;
 }
 
+export type RepositoriesFilterSortStateWithIds = RepositoriesFilterSortState & {
+  repositoryIds?: number[];
+}
```
elenatanasoiu, line 18: Was there a reason we couldn't add this field to the original `RepositoriesFilterSortState` ? At the moment the extra type is making all the method definitions quite verbose while the actual implementation is short.

koesie10: We could it add it to the original type as well, but I think that would make it less clear that this property should not be used in the view. The view should explicitly not use the `repositoryIds` property because it would mean that once you select a repository using a checkbox, all other repositories would be hidden. By making this a separate type, we can ensure that the front-end doesn't use this field, rather than it being something implicit.

elenatanasoiu: Gotcha. 

> The view should explicitly not use the `repositoryIds` property because it would mean that once you select a repository using a checkbox, all other repositories would be hidden.

Can you expand on this a bit? Which view are we talking about?

koesie10: Sure! The view is the variant analysis view, which keeps track of the state [here](https://github.com/github/vscode-codeql/blob/e0230375e36ee23fd1a0378506ff212a09ccf86f/extensions/ql-vscode/src/view/variant-analysis/VariantAnalysis.tsx#L61-L62). As you can see, there are two different states we keep track of: the sort and filter state and the selected repositories. In the view, the selected repositories state is only used for keeping track of which repositories have their checkbox checked and for sending it to the extension.

If we combine the two types, there would also be a `repositoryIds` on the `filterSortState` which might be enticing to use in the view as well, instead of this separate state. However, this state is actually used for filtering which repositories are visible [here](https://github.com/github/vscode-codeql/blob/73ff057df00ed82d00556c74a139469add776723/extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisAnalyzedRepos.tsx#L58). So, if you start using this state for keeping track of selected checkboxes, you would find that once you select a checkbox, all other repositories disappear.

We can of course still combine the types, but it feels somewhat weird to me that there's a field on the type used by the view which actually "breaks" the selection if we started using it.

I was able to somewhat simplify the type checks so we now only have 1 method for checking the types, rather than two.

extensions/ql-vscode/src/pure/variant-analysis-filter-sort.ts
```diff
@@ -72,16 +76,22 @@ export function compareWithResults(filterSortState: RepositoriesFilterSortState
...
-export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[], filterSortState: RepositoriesFilterSortState | undefined): T[];
-export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortState | undefined): T[] | undefined;
+export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[], filterSortState: RepositoriesFilterSortState | RepositoriesFilterSortStateWithIds | undefined): T[];
+export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortState | RepositoriesFilterSortStateWithIds | undefined): T[] | undefined;
 
-export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortState | undefined): T[] | undefined {
+export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortState | RepositoriesFilterSortStateWithIds | undefined): T[] | undefined {
```
elenatanasoiu, line 87: Simplifying the check is useful. 👍 

I seem to remember having this conversation before about not being able to split up this method into two.

Just for the sake of the argument, if we were able to split it, we could have:
- one method to be used by the view (without `repositoryIds`)
- one for everywhere else

E.g. 

```
  export function viewFilterAndSortRepositories<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortState | undefined): T[] | undefined {
    if (!repositories) {
      return undefined;
    }

    return repositories
      .filter(repo => filterSortState.repositoryIds.includes(repo.repository.id))
      .sort(compareWithResults(filterSortState));
  }

  export function filterAndSortRepositories<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortStateWithIds | undefined): T[] | undefined {
    if (!repositories) {
      return undefined;
    }

    return repositories
      .filter(repo => matchesFilter(repo.repository, filterSortState));
      .sort(compareWithResults(filterSortState));
  }

```

I think the reason this is getting quite intertwined is because the singular method is doing too much so every time we need to touch it, it's going to become harder to change. With smaller methods we'd release ourselves a little bit from that burden.

elenatanasoiu: Ok! So I think there's two problems here:

1. We want to make it crystal clear that we shouldn't be using `repositoryIds` in view filtering. 
2. When we do want to filter by repo id, we should only be doing that when there are `repositoryIds` set. Otherwise we should default to the way we do sorting like in the view (by repo name).

I think you can do two things:
1. Leave things as they are as it's doing the job, but maybe shorten the type check from 

```ts
!!filterSortState && 'repositoryIds' in filterSortState && filterSortState.repositoryIds !== undefined && filterSortState.repositoryIds.length > 0
```

to 

```ts
!!filterSortState?.repositoryIds && filterSortState.repositoryIds.length > 0
```

I ** believe ** from testing in the console that those are equivalent.

2. To solve your first problem of making it clear which version should be used in views you could still define a separate `filterAndSortByRepoName` method and replace the call to it in the view:

```ts
export function filterAndSortByRepoName<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortState | undefined): T[] | undefined {
  if (!repositories) {
    return undefined;
  }

  return repositories
    .filter(repo => matchesFilter(repo.repository, filterSortState))
    .sort(compareWithResults(filterSortState));
}
```

Then you could incorporate it into your existing method to make it clear that we fallback to sorting by repo name with repo ids are missing.

```ts
export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortStateWithIds | undefined): T[] | undefined {
  if (!repositories) {
    return undefined;
  }

  if (isFilterOnRepositoryIds(filterSortState)) {
    return repositories
      .filter(repo => filterSortState.repositoryIds.includes(repo.repository.id))
      .sort(compareWithResults(filterSortState));
  } else {
    filterAndSortByRepoName(repositories, filterSortState);
  }
}
```

extensions/ql-vscode/src/pure/variant-analysis-filter-sort.ts
```diff
@@ -76,22 +76,26 @@ export function compareWithResults(filterSortState: RepositoriesFilterSortState
...
-export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortState | RepositoriesFilterSortStateWithIds | undefined): T[] | undefined {
+export function filterAndSortRepositoriesWithResults<T extends SortableResult>(repositories: T[] | undefined, filterSortState: RepositoriesFilterSortStateWithIds | undefined): T[] | undefined {
   if (!repositories) {
     return undefined;
   }
 
-  const filteredRepositories = isFilterOnRepositoryIds(filterSortState) ?
-    repositories.filter(repo => filterSortState.repositoryIds.includes(repo.repository.id)) :
-    repositories.filter(repo => matchesFilter(repo.repository, filterSortState));
+  if (filterSortState?.repositoryIds && filterSortState.repositoryIds.length > 0) {
```
elenatanasoiu, line 94: ⭐

elenatanasoiu (approved):
Het lijkt me goed!

PR #1774, opened 2022-11-18
Allow synthetic variant analysis packs to handle `${workspace}`

aeisenberg:
`${workspace}` references are new in CLI version 2.11.3. These mean that
the version depended upon in a pack must be the version available in the
current codeql workspace.

When generating a variant analysis pack, however, we copy the target
query and generate a synthetic pack with the original dependencies.
This breaks workspace references since the synthetic pack is no longer
in the same workspace.

A simple workaround is to replace `${workspace}` with `*` references.

I wanted to ensure the tests exercise this feature, so I altered one of our test qlpacks to use the `${workspace}` reference. This will work on CLI v2.11.3 and later. For earlier versions, I needed to change that reference to a `*` reference.

I also fixed a problem I was having locally where the CLI tests were timing out. By bumping the timeout to 20s in the `testConfigHelper`, I got around that error.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ artifacts/
 # Visual Studio workspace state
 .vs/
 
-# Rush files
-/common/temp/**
-package-deps.json
-**/.rush/temp
+# CodeQL metadata
+.cache/
+.codeql/
--- a/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
+++ b/extensions/ql-vscode/src/remote-queries/run-remote-query.ts
@@ -21,7 +21,7 @@ import {
 import { ProgressCallback, UserCancellationException } from "../commandRunner";
 import { RequestError } from "@octokit/types/dist-types";
 import { QueryMetadata } from "../pure/interface-types";
-import { REPO_REGEX } from "../pure/helpers-pure";
+import { getErrorMessage, REPO_REGEX } from "../pure/helpers-pure";
 import * as ghApiClient from "./gh-api/gh-api-client";
 import {
   getRepositorySelection,
@@ -99,6 +99,8 @@ async function generateQueryPack(
 
     void logger.log(`Copied ${copiedCount} files to ${queryPackDir}`);
 
+    await fixPackFile(queryPackDir, packRelativePath);
+
     language = await findLanguage(cliServer, Uri.file(targetQueryFileName));
   } else {
     // open popup to ask for language if not already hardcoded
@@ -115,6 +117,7 @@ async function generateQueryPack(
       dependencies: {
         [`codeql/${language}-all`]: "*",
       },
+      defaultSuite: generateDefaultSuite(packRelativePath),
     };
     await fs.writeFile(
       path.join(queryPackDir, "qlpack.yml"),
@@ -124,8 +127,6 @@ async function generateQueryPack(
   if (!language) {
     throw new UserCancellationException("Could not determine language.");
   }
-
-  await ensureNameAndSuite(queryPackDir, packRelativePath);
 
   // Clear the cliServer cache so that the previous qlpack text is purged from the CLI.
   await cliServer.clearCache();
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -16,7 +16,6 @@ artifacts/
 # Visual Studio workspace state
 .vs/
 
-# Rush files
-/common/temp/**
-package-deps.json
-**/.rush/temp
```
angelapwen, line 22: What are these files and why do we not need them in the .gitignore anymore? 🤔

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
```diff
@@ -298,33 +299,52 @@ export async function prepareRemoteQueryRun(
...
+  // update pack name
   qlpack.name = QUERY_PACK_NAME;
 
-  qlpack.defaultSuite = [
+  // update default suite
+  delete qlpack.defaultSuiteFile;
+  qlpack.defaultSuite = generateDefaultSuite(packRelativePath);
+
+  // remove any ${workspace} version references
+  for (const [key, value] of Object.entries(qlpack.dependencies || {})) {
```
angelapwen, line 330: Also curious: why the `|| {}`?

angelapwen (commented):
Looks like a test is failing, but I'll be around to review any changes/release changes.

.gitignore
aeisenberg: A few years ago, we were using [rush](https://rushjs.io/) instead of npm. For our purposes, rush was overkill, so we switched to npm. We just never removed these ignore files.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
aeisenberg: Just in case `dependencies` is undefined. It's unlikely, but I guess someone _could_ be running a nolang query.

aeisenberg:
Of course...`${workspace}` is not recognized by CLI versions < 2.11.3.

extensions/ql-vscode/src/remote-queries/run-remote-query.ts
angelapwen: Oh, I see, because `Object.entries()` doesn't gracefully handle the case where the input is undefined. Thanks!

aeisenberg:
I think I found out why the CLI tests are failing locally. I made a change to https://github.com/aeisenberg/vscode-codeql/blob/aeisenberg/fix-mrva-packs/extensions/ql-vscode/src/vscode-tests/test-config.ts#L110 to add a timeout. Also, the `github.vscode-pull-request-github` extension was causing a 403 error. Disabling the extension seems to help.

aeisenberg:
@angelapwen would you mind taking another look? I think the tests are passing now. I had to do a bit of magic to get the tests working. I'll explain better in the PR description.

extensions/ql-vscode/src/vscode-tests/cli-integration/data-remote-qlpack/qlpack.yml
```diff
@@ -1,4 +1,4 @@
 name: github/remote-query-pack
 version: 0.0.0
 dependencies:
-  codeql/javascript-all: '*'
+  codeql/javascript-all: '${workspace}'
```
angelapwen, line 4: I don't know these tests well, but do you think you could add a comment here to explain that the `'${workspace}'` bit is important to ensure tests pass? Not sure that it's clear from the name of the test/directory.

extensions/ql-vscode/src/vscode-tests/cli-integration/global.helper.ts
```diff
@@ -122,3 +125,34 @@ export async function cleanDatabases(databaseManager: DatabaseManager) {
     await commands.executeCommand("codeQLDatabases.removeDatabase", item);
   }
 }
+
+export async function fixWorkspaceReferences(
```
angelapwen, line 129: Also wonder if you could add a comment here to describe why someone would use this

extensions/ql-vscode/src/vscode-tests/cli-integration/global.helper.ts
```diff
@@ -122,3 +125,34 @@ export async function cleanDatabases(databaseManager: DatabaseManager) {
...
+    );
+    const originalDeps = { ...qlpack.dependencies };
+    removeWorkspaceRefs(qlpack);
+    fs.writeFileSync(qlpackFileWithWorkspaceRefs, yaml.dump(qlpack));
+    return originalDeps;
+  }
+  return undefined;
+}
+
+export async function restoreWorkspaceReferences(
```
angelapwen, line 146: Same as above (comment description)

angelapwen (approved):
Looks good to me 👍 just some requests for code comments 😄

extensions/ql-vscode/src/vscode-tests/cli-integration/global.helper.ts
aeisenberg: Done

extensions/ql-vscode/src/vscode-tests/cli-integration/global.helper.ts
aeisenberg: Done

PR #1787, opened 2022-11-24
Remove deprecated fast compilation option

kaspersv:
Fast compilation has been deprecated. This PR removes the options associated with fast compilation.


aeisenberg (commented):
Does anything need to change for the new query server?

kaspersv:
No, the new query server does not support the fast compilation option in the first place. Since the old query server is already deprecated, let us leave these few lines of code in the legacy query server client, until the entire legacy query server client can be deleted.

PR #1788, opened 2022-11-24
Remove unsupported version constraints

koesie10:
This removes all CodeQL CLI version constraints for unsupported CLI versions (< 2.7.6). The oldest supported CLI version is 2.7.6 since GHES 3.4 recommends using CodeQL CLI 2.7.6.

I've also removed a constraint on version 2.7.1, but I'm not sure if we still support all 2.7.x versions. 

Please double check that I made the correct transformations in call-sites, especially with regards to if conditions being `true`/`false`.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Renamed command "CodeQL: Run Query" to "CodeQL: Run Query on Selected Dababase".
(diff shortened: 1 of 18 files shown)


elenatanasoiu (approved):

koesie10:
@aeisenberg Would you be able to confirm that we're not supporting the older 2.7 minor versions anymore?

aeisenberg:
Yes. That's exactly right. 2.7.6 is the earliest version we need to support.

aeisenberg (approved):
Nice! It's great to be deleting all this old code.

One thing to consider is maybe we want to have a check during startup in `extension.ts` where we throw an error if the CLI version is < 2.7.6. Doing this or not should not block this PR.

PR #1793, opened 2022-11-25
Add debugging support for Jest integration tests

koesie10:
Since we are launching a completely different process for the extension tests than the process that is launched by VSCode, we need to add some special handling for the debugging.

This will let the extension host/VSCode expose a debugging port, which VSCode will then connect to. This is "less desirable than letting the bootloader do its thing", but a packaged VSCode application does not allow using the bootloader (`NODE_OPTIONS=--require=...`). Therefore, we have to fallback to this option.

See: https://github.com/microsoft/vscode-js-debug/blob/47c60558ec31902f42c255abb9b460078df02f9d/src/configuration.ts#L405-L411

## Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -72,13 +72,10 @@
         "--projects",
         "out/vscode-tests/no-workspace"
       ],
-      "stopOnEntry": false,
       "sourceMaps": true,
-      "outFiles": [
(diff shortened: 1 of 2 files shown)


charisk (approved):
Legend!

PR #1801, opened 2022-11-28
Add retries of CLI integration tests

koesie10:
This will patch `jest-runner-vscode` to retry tests. See the separate commits for more information about each step.

## Checklist

### DIFF
--- a/extensions/ql-vscode/patches/jest-runner-vscode+3.0.1.patch
+++ b/extensions/ql-vscode/patches/jest-runner-vscode+3.0.1.patch
@@ -41,20 +41,44 @@ index 8657ace..4d35409 100644
 -export default function runVSCode({ vscodePath, args, jestArgs, env, tests, globalConfig, filterOutput, onStart, onResult, onFailure, ipc, quiet, }: RunVSCodeOptions): Promise<void>;
 +export default function runVSCode(options: RunVSCodeOptions): Promise<void>;
 diff --git a/node_modules/jest-runner-vscode/dist/run-vscode.js b/node_modules/jest-runner-vscode/dist/run-vscode.js
-index 5d8e513..6edf99a 100644
+index 5d8e513..cacbc42 100644
 --- a/node_modules/jest-runner-vscode/dist/run-vscode.js
 +++ b/node_modules/jest-runner-vscode/dist/run-vscode.js
-@@ -5,7 +5,8 @@ var __importDefault = (this && this.__importDefault) || function (mod) {
+@@ -5,8 +5,18 @@ var __importDefault = (this && this.__importDefault) || function (mod) {
  Object.defineProperty(exports, "__esModule", { value: true });
  const child_process_1 = __importDefault(require("child_process"));
  const console_1 = __importDefault(require("console"));
 -async function runVSCode({ vscodePath, args, jestArgs, env, tests, globalConfig, filterOutput, onStart, onResult, onFailure, ipc, quiet, }) {
+-    return await new Promise(resolve => {
++const fs_1 = __importDefault(require("fs"));
++const path_1 = __importDefault(require("path"));
++const os_1 = __importDefault(require("os"));
 +async function runVSCode(options) {
 +    const { vscodePath, args, jestArgs, env, tests, globalConfig, filterOutput, onStart, onResult, onFailure, ipc, quiet, attempt, maxRetries, } = options;
-     return await new Promise(resolve => {
++    const tempUserDir = await fs_1.default.promises.mkdtemp(path_1.default.resolve(os_1.default.tmpdir(), 'jest-runner-vscode-user-data-'));
++    return await new Promise(promiseResolve => {
++        const resolve = () => {
++            fs_1.default.rm(tempUserDir, { recursive: true }, () => {
++                promiseResolve();
++            });
++        };
          const useStdErr = globalConfig.json || globalConfig.useStderr;
          const log = useStdErr
-@@ -101,11 +102,31 @@ async function runVSCode({ vscodePath, args, jestArgs, env, tests, globalConfig,
+             ? console_1.default.error.bind(console_1.default)
+@@ -82,7 +92,11 @@ async function runVSCode({ vscodePath, args, jestArgs, env, tests, globalConfig,
+         ipc.server.on('stdout', onStdout);
+         ipc.server.on('stderr', onStderr);
+         ipc.server.on('error', onError);
+-        const vscode = child_process_1.default.spawn(vscodePath, args, { env: environment });
++        const launchArgs = args;
++        if (!hasArg('user-data-dir', launchArgs)) {
++            launchArgs.push(`--user-data-dir=${tempUserDir}`);
++        }
++        const vscode = child_process_1.default.spawn(vscodePath, launchArgs, { env: environment });
+         if (!silent && !filterOutput) {
+             vscode.stdout.pipe(process.stdout);
+             vscode.stderr.pipe(process.stderr);
+@@ -101,11 +115,31 @@ async function runVSCode({ vscodePath, args, jestArgs, env, tests, globalConfig,
              const message = `VS Code exited with exit code ${exit}`;
              if (typeof code !== 'number' || code !== 0) {
                  silent || quiet || console_1.default.error(message);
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/vscode-tests/cli-integration/jest-runner-vscode.config.js
```diff
@@ -25,6 +25,7 @@ const config = {
     ...baseConfig.extensionTestsEnv,
     INTEGRATION_TEST_MODE: "true",
   },
+  retries: 3,
```
elenatanasoiu, line 28: Is there any guarantee that 3 retries will resolve the flakiness?

I know it's a lot to ask after so much effort doing the jest migration, but are we able to pinpoint what the failure reasons are at the moment for windows?

extensions/ql-vscode/patches/jest-runner-vscode+3.0.1.patch
```diff
@@ -41,20 +41,44 @@ index 8657ace..4d35409 100644
...
 +async function runVSCode(options) {
 +    const { vscodePath, args, jestArgs, env, tests, globalConfig, filterOutput, onStart, onResult, onFailure, ipc, quiet, attempt, maxRetries, } = options;
-     return await new Promise(resolve => {
++    const tempUserDir = await fs_1.default.promises.mkdtemp(path_1.default.resolve(os_1.default.tmpdir(), 'jest-runner-vscode-user-data-'));
++    return await new Promise(promiseResolve => {
++        const resolve = () => {
++            fs_1.default.rm(tempUserDir, { recursive: true }, () => {
++                promiseResolve();
++            });
++        };
```
aeisenberg, line 64: Reading a diff of a diff is making my brain hurt. 😅

extensions/ql-vscode/src/vscode-tests/jest-runner-vscode.config.base.js
```diff
@@ -11,7 +11,6 @@ const config = {
   launchArgs: [
     "--disable-gpu",
     "--extensions-dir=" + path.join(rootDir, ".vscode-test", "extensions"),
```
aeisenberg, line 13: Should we be disabling potentially conflicting extensions here as well? There was a list of them when running with mocha. Not sure if these extensions are being explicitly disabled any more.

aeisenberg (commented):
In my experience with mocha, running the CLI tests on retries didn't work well. In fact, the retries usually masked the original problem. For the CLI tests, if there was a failure on the first try, the retries would _always_ fail due to some state being left around (though I never dug too deeply into what it was). Perhaps the `--new-window` command will work, but please make sure that running a retry will have the possibility of being successful.

One suggestion is to inject a known error into CLI integration tests, push it up to CI and make sure that error (and only that error) happens for each retry.

koesie10:
Thanks for the reviews!

I've found three common errors on Windows, for which retries do seem to work:
- "VSCode exited with exit code 1": I don't really know why, but VSCode occasionally seems to exit without running all tests. Usually, restarting VSCode and trying again works.
- "No test result returned": Again, not sure why this happens, but sometimes VSCode exits without reporting all test results. Usually, restarting VSCode and trying again works.
- Timeouts on the `legacy-query.test.ts`. For this one, retrying also usually works. I'm hesitant to increase the timeout since it's already set to 60 seconds per test/hook, so increasing it further might just prolong the time it takes until failure.

I'm not 100% confident that 3 retries will always be enough for resolving the flakiness, but this should at least reduce it to a somewhat reasonable level. Before this PR, we would sometimes get 4 failing CLI integration tests on Windows in a row, which is really frustrating to deal with and takes a lot of time just to get the tests to pass.

extensions/ql-vscode/src/vscode-tests/jest-runner-vscode.config.base.js
koesie10: The extensions are being disabled in the configuration file specific to the `cli-integration` tests:

https://github.com/github/vscode-codeql/blob/7ba90275a8c71f7e06142035f07cc762af7765fb/extensions/ql-vscode/src/vscode-tests/cli-integration/jest-runner-vscode.config.js#L14-L19

We are also disabling all extensions in the configuration for the `no-workspace` and `minimal-workspace` tests:

https://github.com/github/vscode-codeql/blob/7ba90275a8c71f7e06142035f07cc762af7765fb/extensions/ql-vscode/src/vscode-tests/no-workspace/jest-runner-vscode.config.js#L6

This should match the previous Mocha setup in terms of which extensions are enabled/disabled.

elenatanasoiu:
Thanks for writing up the error cases. From what you've described, I think there's three possibilities:

1. We're not loading a dependency.
2. We're loading too many dependencies and they get in the way of each other.

Given that the tests do eventually pass when VSCode is restarted, I'm inclined to think it's the second option. Something hangs when VSCode is loaded and gets in the way of tests being run successfully. When you restart, this process is killed so it unblocks everything else. 

Are you able to make the exit message more verbose when it does quit unexpectedly? E.g. spit out the entire log?

Also, is it always the same version of windows that fails?

koesie10:
> We're loading too many dependencies and they get in the way of each other.

This is probably part of the problem. We are not bundling the extension (yet), so VSCode needs to load thousands of files to even load the extension. IO on Windows is a lot slower than on Linux, so this can take hundreds of milliseconds to seconds in the worst cases. In subsequent retries, it might be that the filesystem cache is simply "hot" and it takes less time to load all files.

> Are you able to make the exit message more verbose when it does quit unexpectedly? E.g. spit out the entire log?

I don't think it's actually related to our tests, running tests in VSCode just seems to be somewhat unstable on Windows. We are already logging all information we have, since we only receive the exit code of the process and the extension host logs are already being logged. See for example [this run](https://github.com/github/vscode-codeql/actions/runs/3549822604/jobs/5962614719) where you can see that there are some logs of the extension host, but nothing specifically saying why VSCode exited.

You can see in that run that we're also running two instances of VSCode. That should also be fixed by this PR, but won't fix the flakiness.

> Also, is it always the same version of windows that fails?

No, I don't think we can narrow it down to a specific CLI version.

elenatanasoiu:
> No, I don't think we can narrow it down to a specific CLI version.

Ah, I see. I thought it was a version of windows specifically used for GitHub Actions. But it's the CLI version. Gotcha. 

Bundling the extension does sounds promising.

extensions/ql-vscode/patches/jest-runner-vscode+3.0.1.patch
```diff
@@ -41,7 +41,7 @@ index 8657ace..4d35409 100644
 -export default function runVSCode({ vscodePath, args, jestArgs, env, tests, globalConfig, filterOutput, onStart, onResult, onFailure, ipc, quiet, }: RunVSCodeOptions): Promise<void>;
 +export default function runVSCode(options: RunVSCodeOptions): Promise<void>;
```
elenatanasoiu, line 42: On adding changes to `jest-runner-vscode`, was there a reason to prefer a patch directly in the project over forking the package?

elenatanasoiu:
It looks like running tests before Jest takes around 12-14 mins while running tests with jest + retry version takes 14 mins so there doesn't seem to be a massive time difference there.

extensions/ql-vscode/patches/jest-runner-vscode+3.0.1.patch
koesie10: The original idea was to upstream the changes (there's a PR open for the earlier patch), but it doesn't seem likely that the maintainer is actively checking PRs. I think we can definitely fork the package as a future improvement, although that makes it somewhat harder to make these sorts of changes quickly. We would also need to change the package to publish to our own npm package, and I'm not sure how involved it would be to publish it under the `@github` scope.

I'll create an issue for investigating whether this is something we can do, although I think it's better to keep it in the project for now so the Jest migration can be merged into `main`.

aeisenberg (approved):
Thanks for the hard work on this. I know fixing these kinds of flaky tests is tedious and time consuming. 

@elenatanasoiu are there specific CLI tests that are failing on specific versions of the CLI?

elenatanasoiu:
They seem random:

- https://github.com/github/vscode-codeql/actions/runs/3564060734/jobs/5987575631
- https://github.com/github/vscode-codeql/actions/runs/3548377296/jobs/5962450301
- https://github.com/github/vscode-codeql/actions/runs/3548035823/jobs/5958815336

Also some ubuntu ones seem to flake, but I'm not sure that's related: 
- https://github.com/github/vscode-codeql/actions/runs/3548030220/jobs/5958801291
- https://github.com/github/vscode-codeql/actions/runs/3563927710/jobs/5987289645

PR #1802, opened 2022-11-28
Show correct query filename on Windows

koesie10:
On Windows, we were showing the full path to the query, rather than just the filename. This is because the `path` package being imported was actually `path-browserify` which only claims support for POSIX. Since Windows uses backslashes rather than forward slashes for paths, this resulted in the full path being shown.

This creates a new `basename` function that works on both POSIX and Windows by detecting whether a POSIX or Windows path is given. This ensures that the correct path is shown on Windows, and will also ensure that we show the correct path on Linux if the user has opened a variant analysis that was originally created on Windows.

## Checklist


elenatanasoiu (approved):
Nice, I love that the test use cases cover combinations of `\` and `/`.

PR #1804, opened 2022-11-29
Stop variant analysis monitor after removing the variant analysis from the query history

koesie10:
This will stop the variant analysis monitor from monitoring when a variant analysis is removed from the query history. Since the variant analysis monitor cannot depend on the variant analysis manager (this would create a circular dependency), a function is passed into the variant analysis monitor for checking whether the variant analysis should be cancelled.

This commit will also ensure that even if a variant analysis comes in through the `onVariantAnalysisChange` callback, it won't be added to the variant analysis map of the manager.

## Checklist


extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/variant-analysis-manager.test.ts
```diff
@@ -780,7 +780,12 @@ describe("Variant Analysis Manager", async function () {
         });
 
         it("should remove variant analysis", async () => {
-          await variantAnalysisManager.onVariantAnalysisUpdated(
+          pathExistsStub
+            .withArgs(
+              path.join(storagePath, dummyVariantAnalysis.id.toString()),
+            )
+            .resolves(true);
+          await variantAnalysisManager.rehydrateVariantAnalysis(
```
elenatanasoiu, line 788: Why do we need to add this step?

koesie10: Previously, it would store the variant analysis in the `variantAnalyses` map when it was changed. Now, we only store it in the map if it's changed **and** it's already in the map. So, we need to make sure it's still in the map. This was done with the call to `onVariantAnalysisUpdated` before, but now you need to actually fully rehydrate the variant analysis.

robertbrignull (approved):
LGTM, assuming @elenatanasoiu is also happy

elenatanasoiu (approved):

PR #1808, opened 2022-11-29
Don't show pending state when analysis has failed

robertbrignull:
If a variant analysis fails because there are no repos to query (or the repos don't have databases) then it doesn't get as far as creating an actions workflow. This means that using the presence of that field alone is problematic for determining the status of a variant analysis.

This PR will avoid cases like:
Screenshot 2022-11-25 at 16 29 11

## Checklist


robertbrignull:
I briefly forgot tests exist, but I'll try to write some that cover this.

koesie10 (approved):

PR #1814, opened 2022-11-30
Adjust for new GitHub Api response

norascheuch:
Creating a new MRVA will lead to different error responses which are now reflected in our MockApi tests.


robertbrignull (approved):
LGTM (once the API change is merged)

PR #1815, opened 2022-11-30
Add theme provider when rendering code flows

robertbrignull:
In our push to remove the dependency on Primer we accidentally broke the rendering of paths. This PR adds in a `ThemeProvider` which is what's needed to get it working again. Although this is a slight step in the wrong direction it's probably the best thing to do right now before the release, and then we'll look at doing it properly later on as we continue with the Primer migration.

Also I added some tests as this component. The tests would actually have caught the error when render, so at least this means we won't break it again next time.

## Checklist


robertbrignull:
I've tried it with the old view too and it seems to work. I see code paths and I don't see any errors on the screen or in the terminal. So I think it's fine. Someone may want to verify to be sure.

charisk (approved):

PR #1818, opened 2022-12-01
Close serialized webview when another view already exists

koesie10:
When deserializing a webview, it could happen that a view was already manually opened by the user before the webview was deserialized. This would result in duplicate webview tabs, which is not supported by the manager.

This will close the webview that is being deserialized and focus on the existing view. This should ensure that we never have duplicate loaded webview tabs. There could still be duplicate webview tabs if there are non-deserialized tabs, but once it is opened it should be closed automatically.

## Checklist


shati-patel (approved):
Looks good to me! (Tested locally too - fixes the original bug 🐛 🙌🏽  )

PR #1819, opened 2022-12-01
Fix Gist title in result export

koesie10:
The Gist title in the result export didn't take into account the actual number of exported repositories, it only used the scanned, unfiltered, repositories in the variant analysis. This switches it to use the actual exported repositories for determining the result and repository counts.

This is somewhat more complicated than we'd expect it to be since the results are being read in async, so we need to switch the order of operations and store some additional information for being able to compute this information. However, this also makes the code somewhat easier to understand since the summary file is now being created in only 1 location, rather than being split between a method and a for-loop.

## Checklist


shati-patel (approved):
Nice one! Thanks for fixing this (and making the "generate summary" code nicer in the process 💪🏽 )

PR #1826, opened 2022-12-02
Check variant analysis status in isVariantAnalysisComplete

robertbrignull:
This allows us to get out of a state where we think that a variant analysis is still in progress but all of its repo tasks are complete. It is possible for the API to return this state in a temporary edge case when the workflow has just finished, but its not permanent and the API will soon start saying the variant analysis is complete. So in this case we want to continue polling, and if VSCode starts up we want to poll for updated state.

## Checklist


koesie10 (approved):
LGTM, thanks for figuring out that this could be happening!

robertbrignull:
I've managed to reproduce this state a couple of times when the analysis has just finished. I shut down VSCode at that exact moment and it correctly started up the monitor once I started VScode again.

PR #1827, opened 2022-12-02
Limit SARIF code snippet size

koesie10:
This adds a new filtering on SARIF code snippets for very large code snippets (defined as 8KB or more). If less than 1% of such a snippet is highlighted, it will not include the code snippet in the analysed results, and it will thus not be shown in the UI.

This is to avoid very large SARIF files that can cause the extension host to crash when the analysis results are send to the UI. I don't think any of these snippets would ever be useful to show, so it should be fine to just not include them.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/sarif-processing.ts
+++ b/extensions/ql-vscode/src/remote-queries/sarif-processing.ts
@@ -1,5 +1,6 @@
 import * as sarif from "sarif";
 import {
+  parseHighlightedLine,
   parseSarifPlainTextMessage,
   parseSarifRegion,
 } from "../pure/sarif-utils";
@@ -14,6 +15,11 @@ import {
   CodeSnippet,
   HighlightedRegion,
 } from "./shared/analysis-result";
+
+// A line of more than 8k characters is probably generated.
+const CODE_SNIPPET_LARGE_LINE_SIZE_LIMIT = 8192;
+// If less than 1% of the line is highlighted, we consider it a small snippet.
+const CODE_SNIPPET_HIGHLIGHTED_REGION_MINIMUM_PERCENTAGE = 0.01;
 
 const defaultSeverity = "Warning";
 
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -15,6 +16,11 @@ import {
   HighlightedRegion,
 } from "./shared/analysis-result";
 
+// A line of more than 8k characters is probably generated.
+const CODE_SNIPPET_LARGE_LINE_SIZE_LIMIT = 8192;
+// If less than 1% of the line is highlighted, we consider it a small snippet.
+const CODE_SNIPPET_HIGHLIGHTED_REGION_MINIMUM_PERCENTAGE = 0.01;
```
robertbrignull, line 22:
```suggestion
const CODE_SNIPPET_HIGHLIGHTED_REGION_MINIMUM_RATIO = 0.01;
```
or
```suggestion
const CODE_SNIPPET_HIGHLIGHTED_REGION_MINIMUM_PERCENTAGE = 1;
```

Calling this "percentage" when the value it contains is out of 1.0 rather than out of 100 could potentially cause confusion in the future to someone who just goes by the variable name. I don't have a preference between changing name or changing the value.

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -163,17 +169,45 @@ export function tryGetRule(
 }
 
 function getCodeSnippet(
+  contextRegion?: sarif.Region,
```
robertbrignull, line 172: I spent a long time looking at the change to ordering of these parameter and whether we'd need to update the call-sites. I now agree it's a no-op because when we do `actualRegion = contextRegion ?? region` below the name change is consistent there. So I believe this is all fine, but it's a little confusing at first glance when reviewing.

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -163,17 +169,45 @@ export function tryGetRule(
...
+
+    const highlightedRegion = parseSarifRegion(region);
+
+    const highlightedLines = code.map((line, index) => {
+      return parseHighlightedLine(line, startLine + index, highlightedRegion);
+    });
+
+    const highlightedCharactersCount = highlightedLines
+      .map((line) => line.highlightedSection.length)
+      .reduce((a, b) => a + b, 0);
```
robertbrignull, line 199:
```suggestion
    const highlightedCharactersCount = highlightedLines.reduce(
      (total, line) => total + line.highlightedSection.length,
      0,
    );
```
Minor point but we could avoid the `map` step to avoid iterating the list twice. I don't think it makes the code less clear in this case.

extensions/ql-vscode/src/remote-queries/sarif-processing.ts
```diff
@@ -163,17 +169,45 @@ export function tryGetRule(
...
+
+    const highlightedPercentage = highlightedCharactersCount / text.length;
```
robertbrignull, line 201: Just another note on variable naming. The value here isn't a percentage because it's out of 1 rather than out of 100, so it could be confusing.

robertbrignull (approved):
Thanks for implementing this. I agree with the logic about region sizes and this should hopefully catch cases of generated or minimized code. I think we can certainly go with this as a first attempt and adjust the logic if we find other cases we want to filter out, so I'm not too worried about trying to make the logic absolutely perfect in this PR.

I have a few comments about variable naming, but the actual logic in the code looks correct to me 👍

robertbrignull:
Sorry just saw your other comment. Tests for this would be great too!

robertbrignull (approved):
Thanks for the updates and tests

PR #1835, opened 2022-12-05
Hide analyzed panel when it's empty

koesie10:
This will hide the "Analyzed" panel when there are no scanned repos and it's completely empty.

When all three panels are empty, this will also hide the search bar and filters, and will skip rendering anything for the panels.

Before:

Screenshot 2022-12-05 at 12 07 27

After:

Screenshot 2022-12-05 at 12 07 46

Screenshot 2022-12-05 at 14 03 33

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysis.stories.tsx
+++ b/extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysis.stories.tsx
@@ -385,3 +385,19 @@ Failed.args = {
   repoStates,
   repoResults,
 };
+
+export const FailedInternal = Template.bind({});
+FailedInternal.args = {
+  variantAnalysis: {
+    ...variantAnalysis,
+    status: VariantAnalysisStatus.Failed,
+    failureReason: VariantAnalysisFailureReason.InternalError,
+    completedAt: new Date(
+      new Date(variantAnalysis.createdAt).getTime() + 100_000,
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/view/jest.config.ts
```diff
@@ -171,7 +171,7 @@ const config: Config = {
     "^.+\\.tsx?$": [
       "ts-jest",
       {
-        tsconfig: "src/view/tsconfig.spec.json",
+        tsconfig: "<rootDir>/tsconfig.spec.json",
```
robertbrignull, line 174: Just wondering why does this change need to happen for this PR? Could this be moved to a separate PR so it's not mixed in with unrelated changes?

koesie10:
Thanks for the review!

> For the third screenshot where there are no panels to display, could we add the "no repositories to analyse" text? I worry it's not clear why nothing is being displayed, and I worry people would think the page is broken.
> 
> I'm not sure it's possible to hit this case since the API won't let you create a variant analysis with no scanned or skipped repos, but I appreciate the UI still needs to do something in this case.

I have updated the PR so we now show the warnings that are normally also shown. I have updated the screenshot in the PR description. I don't believe we can get into this state normally, but could it be that we get in this state when the processing job fails at some point?

robertbrignull (approved):
LGTM 👍

PR #1841, opened 2022-12-06
Fix incorrect await for monitor

koesie10:
When rehydrating remote queries, we were awaiting the monitoring command. Since this command may take minutes to hours to complete, it seems like this would block the extension from loading. This is the same issue as in https://github.com/github/vscode-codeql/pull/1698, but for remote queries instead of variant analyses.

## Checklist


charisk:
Any idea why this hasn't been an issue? Seems a bit strange.

koesie10:
> Any idea why this hasn't been an issue? Seems a bit strange.

I'm not sure, but I suspect not many people are closing VSCode while their remote query hasn't completed yet, or only opening it again after the remote query has completed. This would only be a big problem if the remote query was still in progress when re-opening VSCode.

charisk (approved):

PR #1847, opened 2022-12-07
Run type checking on all directories in CI

koesie10:
This commit adds a new step to the CI workflow that runs type checking on all directories containing `tsconfig.json` files, using `find` and `xargs`. Unfortunately, this does not work on Windows, so on Windows it's not possible to run all of these type checks locally.

There are some changes to imports to ensure that the typechecks actually pass since these were actually giving errors before.

## Checklist


extensions/ql-vscode/gulpfile.ts/tsconfig.json
```diff
@@ -16,8 +16,7 @@
     "noImplicitReturns": true,
     "experimentalDecorators": true,
     "noUnusedLocals": true,
-    "noUnusedParameters": true,
-    "esModuleInterop": true
```
elenatanasoiu, line 20: Why are we dropping this?

koesie10: The `tsconfig.json` inside the `gulpfile.ts` directory doesn't actually seem to be used by Gulp, it uses the root `tsconfig.json` instead. So, we need to use the same settings for this `tsconfig.json` as for the root `tsconfig.json` for these configuration values specifically, since `esModuleInterop = true` is incompatible with `esModuleInterop = false`. Before this change, the type checks on some files inside `gulpfile.ts` were failing, but Gulp did run the code successfully.

elenatanasoiu: Is it possible to get rid of this file instead?

koesie10: Unfortunately, I don't think we can. The root `tsconfig.json` doesn't include the `gulpfile.ts` directory, and if it did we would be shipping these files to users.

However, we can simplify the file by making it extend from the root `tsconfig.json` and just overriding the settings which decide which files are included. That should also make sure the settings are always the same, so it results in the same behavior.

elenatanasoiu (approved):
👍

PR #1848, opened 2022-12-07
Initialize DbManager when newQueryRun is enabled

norascheuch:
We need to enable checks to only initialize the DbManager when the respective feature flags are turned on.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -38,6 +38,7 @@ import {
   CliConfigListener,
   DistributionConfigListener,
   isCanary,
+  isNewQueryRunExperienceEnabled,
   isVariantAnalysisLiveResultsEnabled,
   joinOrderWarningThreshold,
   MAX_QUERIES,
@@ -138,6 +139,7 @@ import { VariantAnalysisResultsManager } from "./remote-queries/variant-analysis
 import { initializeDbModule } from "./databases/db-module";
 import { ExtensionApp } from "./common/vscode/vscode-app";
 import { RepositoriesFilterSortStateWithIds } from "./pure/variant-analysis-filter-sort";
+import { AppMode } from "./common/app";
 
 /**
  * extension.ts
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -101,7 +101,7 @@ export class VariantAnalysisManager
     private readonly cliServer: CodeQLCliServer,
     private readonly storagePath: string,
     private readonly variantAnalysisResultsManager: VariantAnalysisResultsManager,
-    private readonly dbManager: DbManager,
+    private readonly dbManager?: DbManager, // the dbManager is only needed when the newQueryRunExperience is enabled
```
charisk, line 104: I'd personally skip the comment here since we have the same in the caller. Up to you though!

extensions/ql-vscode/src/extension.ts
```diff
@@ -623,9 +625,15 @@ async function activateWithInstalledDistribution(
 
   void extLogger.log("Initializing variant analysis manager.");
 
-  const app = new ExtensionApp(ctx);
-  const dbModule = await initializeDbModule(app);
-  ctx.subscriptions.push(dbModule);
+  // We only want to initialize the new db panel when the newQueryRunExperience is enabled
+  let dbModule;
+  if (isCanary() && isNewQueryRunExperienceEnabled()) {
+    const app = new ExtensionApp(ctx);
```
charisk, line 631: (Optional) This line can be moved up a bit (and actually it should move up - we should always create an `app`).

charisk (approved):
Thanks for making these changes. I think they solve the issue so happy to merge (I have some optional suggestions).

This means that we're doing the checks twice which isn't great, but I'm happy for us to clean that up later on.

extensions/ql-vscode/src/extension.ts
norascheuch: Why? We don't need it anywhere else in that file I think

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
norascheuch: I added it so that when removing the flag all places appear that need to be touched when searching for it 🤷‍♀️

extensions/ql-vscode/src/extension.ts
charisk: It's not currently used anywhere else because it's a new concept. `app` is a container of information that is meant to be used across the codebase. There's is nothing "database" specific about `app`.

norascheuch: Ok makes sense!

PR #1851, opened 2022-12-07
DB panel: "Add list" command

shati-patel:
Adds a UI button to add a new (empty) list to the DB panel:

https://user-images.githubusercontent.com/42641846/206245834-230e2860-b3d3-4ca2-975e-e02b0d466f41.mp4

Still TODO:
- Add some tests
- Check whether `config` is defined before showing the input box. At the moment, you enter a list name and then might get an error because the config wasn't yet loaded. (One option would be to put the input box inside `db-config-store.ts` function, but ideally we don't want to import stuff from the VS Code API in that file.) 
- Add validation (tracked in a follow-up issue)

## Checklist

N/A - internal only 🙈 


extensions/ql-vscode/src/databases/db-manager.ts
```diff
@@ -73,4 +73,8 @@ export class DbManager {
 
     await this.dbConfigStore.updateExpandedState(newExpandedItems);
   }
+
+  public async addNewList(listName: string): Promise<void> {
```
charisk, line 77: Shall we name this `addNewRemoteList`?

extensions/ql-vscode/src/vscode-tests/minimal-workspace/databases/db-panel.test.ts
```diff
@@ -531,6 +531,66 @@ describe("db panel", () => {
...
+    await saveDbConfig(dbConfig);
+
+    const dbTreeItems = await dbTreeDataProvider.getChildren();
+
+    expect(dbTreeItems).toBeTruthy();
+    const items = dbTreeItems!;
+
+    const remoteRootNode = items[0];
+    const remoteUserDefinedLists = remoteRootNode.children.filter(
+      (c) => c.dbItem?.kind === DbItemKind.RemoteUserDefinedList,
```
charisk, line 568: I think you can use `isRemoteUserDefinedListDbItem` here instead of having to check the `kind`

charisk (approved):
Nice work!

extensions/ql-vscode/src/vscode-tests/minimal-workspace/databases/db-panel.test.ts
shati-patel: That could work! It's still a bit messy, because `isRemoteUserDefinedListDbItem` needs `dbItem` to not be undefined. I've left this as-is for now (since the rest of the tests also check the `kind`).

PR #1866, opened 2022-12-13
Hide DB panel UI actions when config is broken/undefined

shati-patel:
Hides the "Add list" button when the DB config contains errors. Note: We'll need to add this functionality to the upcoming "Add database" button too (see #1870)

I wanted to add a test to check that the context value for `codeQLDatabasesExperimental.configError` had been set to `true`/`false` as appropriate. Annoyingly, I couldn't find a way to access context values in a test 🤔 

## Checklist

N/A - internal only 🦝

- [x] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [x] Issues have been created for any UI or other user-facing changes made by this pull request.
- [x] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #1870, opened 2022-12-14
Create "Add database" button in DB panel 

norascheuch:
To add new databases to the new DB panel we want to have an "Add database" button (using the ["add" icon](https://microsoft.github.io/vscode-codicons/dist/codicon.html)), which lets a user add a new database or an owner item through the UI.

## Checklist


extensions/ql-vscode/src/databases/github-nwo.ts
```diff
@@ -28,7 +28,10 @@ export function looksLikeGithubRepo(
  * @param githubUrl The GitHub repository URL
  * @return The corresponding NWO, or undefined if the URL is not valid
  */
-export function convertGitHubUrlToNwo(githubUrl: string): string | undefined {
+export function convertGitHubUrlToIdentifier(
```
norascheuch, line 31: Is it a good idea to let this method handle both case? We would have some duplication if we create two methods.. Like this we will have to put it in a different file or rename the file though.

shati-patel: I don't have any strong opinions either way 😅 I think this is okay, if you could update function description and return value above? 

Also, this isn't something you added in this PR, but the name "convert ..." feels a bit unclear. Perhaps we could rename this function to something like "get owner (or identifier) from URL"?

extensions/ql-vscode/src/databases/ui/db-panel.ts
```diff
@@ -57,6 +76,77 @@ export class DbPanel extends DisposableObject {
...
+        alwaysShow: true,
+        kind: "repo",
+      },
+      {
+        label: "$(organization) All repositories of a GitHub org or owner",
+        detail:
+          "Add a remote list of databases from a GitHub organization or owner",
+        alwaysShow: true,
+        kind: "owner",
+      },
```
shati-patel, line 94: (not blocking) 
As we discussed offline, we might want to reconsider the wording around "adding databases". 🤔  But I'm also happy to get this in (since it matches the current design) and think about this a bit more later! ❤️

shati-patel (commented):
Looks good! And works well locally too 🎉  

Nothing much to add at the moment, but I'll leave @charisk or others to take a look too (since we paired on bits of this 🙇🏽‍♀️)

extensions/ql-vscode/package.json
```diff
@@ -778,6 +784,11 @@
           "when": "view == codeQLDatabasesExperimental",
           "group": "navigation"
         },
+        {
+          "command": "codeQLDatabasesExperimental.addNewDatabase",
+          "when": "view == codeQLDatabasesExperimental",
```
shati-patel, line 789: Once you've rebased/merged to include the changes from https://github.com/github/vscode-codeql/pull/1866, you can add an [extra condition](https://github.com/github/vscode-codeql/pull/1866/files#diff-be4c5a31505df7de81fa4900e84e66c04d1a6bca9b3da50f8061d2cb2bfbbde1R783) which will hide the button if the config is undefined! 😊

PR #1871, opened 2022-12-14
Workflow to automatically bump cli version if different from the current version

tjgurwara99:
A new CodeQL CLI gets released (roughly) every 2 weeks[^1], which requires some minor updates to the extension. At the time of writing, whoever first determines that a new version of CLI has been released manually updates the version number in some places within this repository. This PR aims to reduce the manual step of updating occurrences of the old CLI version from select files so that they remain up to date for our integration tests. Note that the removal of support of CLI version is a scenario where this newly added script might cause some issues and I'm open to gathering ideas on how to tackle the time when we have to drop support for certain versions.

NOTE for maintainers: This script also requires a newly generated PAT for it to create a PULL request to update the workflow file. Therefore the Bump CLI version workflow is expected to break until a new secret `WORKFLOW_TOKEN`  is added to the repository which would have as its value a PAT with the `workflow` permission.

[^1]: In general, there's a minor version bump every quarter, which requires an extra step: remove old unsupported version of the CLI from our testing matrix. (We stop testing once we no longer support a version of GHES that shipped with that CLI. That means we only need to support 4 minor versions back, since that's how many GHES releases we support.)

## Checklist


scripts/replace-cli-version.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+VERSIONS=$(gh api -H "Accept: application/vnd.github+json" /repos/github/codeql-cli-binaries/releases | jq '.[].tag_name' | head -2)
```
aeisenberg, line 3: Minor: Using the `-r` (raw) option can simplify things:

```suggestion
VERSIONS=$(gh api -H "Accept: application/vnd.github+json" /repos/github/codeql-cli-binaries/releases | jq -r '.[].tag_name' | head -2)
```

scripts/replace-cli-version.sh
```diff
@@ -0,0 +1,9 @@
...
+
+LATEST_VERSION=$(echo $VERSIONS | awk '{ print $1 }' | sed "s/\"//g")
+PREVIOUS_VERSION=$(echo $VERSIONS | awk '{ print $2 }' | sed "s/\"//g")
```
aeisenberg, line 6: With the `-r` option above:

```suggestion
LATEST_VERSION=$(echo $VERSIONS | awk '{ print $1 }')
PREVIOUS_VERSION=$(echo $VERSIONS | awk '{ print $2 }')
```

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,44 @@
...
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+          token: ${{ secrets.WORKFLOW_TOKEN }}
```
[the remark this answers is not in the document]
aeisenberg: And it looks like this is causing the workflow to fail since `WORKFLOW_TOKEN` is not defined.

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,44 @@
...
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git checkout -b bump-cli
+          if [[ `git status --porcelain` ]]; then
+            git add .
+            git commit -m "automatically bump cli version"
+            git push --set-upstream origin bump-cli
+            git push
+            gh pr create --title "Bump CLI Version for integration tests" --body ""
```
aeisenberg, line 43: Can you add something to the body to that specifies what the old and new versions are?

Maybe this:

```suggestion
            gh pr create --title "Bump CLI Version to $LATEST_VERSION for integration tests" --body "Bumps rom $PREVIOUS_VERSION to $LATEST_VERSION." 
```

And then you will need to export the env vars defined in the script.

scripts/replace-cli-version.sh
aeisenberg: Also, see above...can you export these variables so they can be used in the PR title and body?

aeisenberg (commented):
Also, just realized there is a problem if this workflow runs multiple times. If there is an open PR generated by this workflow, and the workflow runs again, trying to create a new PR, it will fail. I have an example of how we handle this in other repos.

aeisenberg:
Here is an action from a private repo that you (@tjgurwara99) should have access to. We use this to create a PR for a particular branch and close any other PRs for that branch if they are open. It might be overkill for this case, but you can still use this or copy pieces of this for this PR. https://github.com/github/semmle-code/blob/main/.github/actions/create-unique-pr/action.yml

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,44 @@
...
+          GH_TOKEN: ${{ secrets.WORKFLOW_TOKEN }}
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
+          git checkout -b bump-cli
+          if [[ `git status --porcelain` ]]; then
+            git add .
+            git commit -m "automatically bump cli version"
+            git push --set-upstream origin bump-cli
+            git push
```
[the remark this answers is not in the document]
tjgurwara99: No, its not necessary - I forgot to remove it in the debug step 😂

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+          token: ${{ secrets.WORKFLOW_TOKEN }}
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -qq jq
+          sudo apt-get install gh
```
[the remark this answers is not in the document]
tjgurwara99: Nice! I didn't know

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,44 @@
+name: Bump CLI version
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  schedule:
+  - cron: 0 0 */14 * * # run every 14 days
```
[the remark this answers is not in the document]
tjgurwara99: Yeah it's a no-op if the version hasn't changed. I had the `workflow_dispatch` as trigger but then I was confused since I thought it may make things weird because I wasn't sure how we wanted this to be executed - I also thought of creating chats-ops but then went against that idea since that would probably be something long term. I assumed a lot of things when I tried experimenting 😂

adityasharad:
High-level comment: I wonder if keeping a separate JSON file with the list of CLI versions would help. The various tests could load that file to determine what CLI versions to test against. Then your PR would only need to modify that file (and you can skip creating the PR if there is no diff). This saves you the logic of having to `sed`-replace the old and new versions within various workflows and test scripts, and allows us to update the entire supported version range, rather than just the last two versions.

.github/workflows/bump-cli.yml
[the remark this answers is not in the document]
aeisenberg: Manual triggers via the actions page is probably fine. This is how we trigger many of our internal release workflows.

scripts/replace-cli-version.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+VERSIONS=$(gh api -H "Accept: application/vnd.github+json" /repos/github/codeql-cli-binaries/releases | jq '.[].tag_name' | head -2)
+
+# we are exporting these variables so that we can access these variables in the workflow
+export LATEST_VERSION=$(echo $VERSIONS | awk '{ print $1 }' | sed "s/\"//g")
+export PREVIOUS_VERSION=$(echo $VERSIONS | awk '{ print $2 }' | sed "s/\"//g")
+
```
aeisenberg, line 8: You need to export these variables for later steps:

```suggestion
echo "LATEST_VERSION=$LATEST_VERSION" >> $GITHUB_ENV
echo "PREVIOUS_VERSION=$PREVIOUS_VERSION" >> $GITHUB_ENV
```

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,32 @@
+name: Bump CLI version
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+  schedule:
```
aeisenberg, line 6: Actually, let's make sure this runs when this particular file changes.

```suggestion
  pull:
    paths:
      - .github/actions/create-pr/action.yml
  schedule:
```

.github/actions/create-pr/action.yml
```diff
@@ -0,0 +1,83 @@
...
+          if [[ $? -ne 0 ]]; then
+            throw "Failed to fetch existing PRs."
+          fi
+          PR_NUMBERS=$(echo $PR_JSON | jq '. | length')
+          if [[ PR_NUMBERS -ne 0 ]]; then 
+            echo "Found existing open PR: $PR_NUMBERS"
+          else
+            gh pr create --head "$HEAD_BRANCH" --base "$BASE_BRANCH" --title "$TITLE" --body "$BODY" --assignee ${{ github.actor }}
+            if [[ $? -ne 0 ]]; then
+              echo "Failed to create new PR."
```
aeisenberg, line 80:
```suggestion
              echo "Failed to create new PR."
              exit 1
```

.github/workflows/main.yml
```diff
@@ -190,7 +190,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        version: ['v2.7.6', 'v2.8.5', 'v2.9.4', 'v2.10.5', 'v2.11.6', 'nightly']
+        version: ['v2.7.6', 'v2.8.5', 'v2.9.4', 'v2.10.5', 'v2.11.5', 'nightly']
```
aeisenberg, line 193: Mistake? Also. below.

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Bump CLI
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          scripts/replace-cli-version.sh
+      - name: Configure actors
+        shell: bash
+        run: |
+          git config --global user.email "github-actions@github.com"
+          git config --global user.name "github-actions[bot]"
```
aeisenberg, line 29: This can be removed since it's already in the action.

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,39 @@
+name: Bump CLI version
+on:
+  workflow_dispatch:
+  push:
```
aeisenberg, line 4: Can you make this `pull` so it can run during a pull request?

.github/workflows/main.yml
tjgurwara99: This is for testing - I was testing it in my fork - hence a lot of rebasing. Sorry for all the noise its creating for you 😓

.github/workflows/bump-cli.yml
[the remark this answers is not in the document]
tjgurwara99: If we clone the repo without this token, the `push` step (in the create-pr action) gets rejected by the remote because GitHub refuses to accept any changes to `workflow` files unless we specifically provide a token that allows it.  For example [here](https://github.com/tjgurwara99/vscode-codeql/actions/runs/3699789862/jobs/6267519174) in my forks workflow, you can see in the logs

```
+ git push --force origin bump-cli
[67](https://github.com/tjgurwara99/vscode-codeql/actions/runs/3699789862/jobs/6267519174#step:4:70)
To https://github.com/tjgurwara99/vscode-codeql
[68](https://github.com/tjgurwara99/vscode-codeql/actions/runs/3699789862/jobs/6267519174#step:4:71)
 ! [remote rejected] bump-cli -> bump-cli (refusing to allow a GitHub App to create or update workflow `.github/workflows/main.yml` without `workflows` permission)
[69](https://github.com/tjgurwara99/vscode-codeql/actions/runs/3699789862/jobs/6267519174#step:4:72)
error: failed to push some refs to 'https://github.com/tjgurwara99/vscode-codeql'
[70](https://github.com/tjgurwara99/vscode-codeql/actions/runs/3699789862/jobs/6267519174#step:4:73)
Error: Process completed with exit code 1.
```

tjgurwara99:
Now its ready I think, the [PR here](https://github.com/tjgurwara99/vscode-codeql/pull/9) is an example of how this works 😄 - for this to work we need the secret `WORKFLOW_TOKEN` with the `workflow` permissions.

.github/workflows/bump-cli.yml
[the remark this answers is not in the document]
adityasharad: Although this is technically possible by using one of our bot PATs, I would recommend against it. So far we have avoided storing any PATs in the secrets for public CodeQL repos, which avoids any risk of those tokens being compromised even if we accidentally include an overly permissive workflow that contributors can run on PRs.

Could you try either:
- Giving the `workflows` permission to the `GITHUB_TOKEN` (if that's possible)
- Avoiding changes to the workflow file itself in the automatically generated PR, instead modifying a separate file that records the version numbers and gets parsed by the first job in the generated workflow to determine the matrix (we have some prior art on this if that's useful)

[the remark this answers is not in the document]
koesie10: I think we could also use a different approach:
- We can use a GitHub App for creating the commit and the PR. This is not trivial since it requires using the GraphQL API to create commits, but I can create an action for doing this.
- Store the required secrets (app ID, installation ID and private key) in environment secrets. We can restrict the environment to only allow running it on the `main` branch, so it will not be possible for contributors to run the workflow without their change being in `main`. By using a GitHub app, we are also able to lock down which repositories it can access and which scopes it has. This is also possible using fine-grained access tokens, but as an admin on this repo you probably still have more access than the GitHub app.

I've already implemented something like this in the past, so I'll try to extract that into a GitHub Action which we can use here to better show what I have in mind.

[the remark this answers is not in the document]
koesie10: @tjgurwara99 I've created a PR on your fork for using a GitHub App: https://github.com/tjgurwara99/vscode-codeql/pull/10. It uses an action on my personal account (https://github.com/koesie10/create-app-pr-action), so it might be that we need to follow the release process for open source releases if we want to have this in the GitHub organization.

We would probably also need to follow a process for creating a GitHub Actions app. I've placed more information about what exactly we would need in the PR.

charisk:
> High-level comment: I wonder if keeping a separate JSON file with the list of CLI versions would help. The various tests could load that file to determine what CLI versions to test against. Then your PR would only need to modify that file (and you can skip creating the PR if there is no diff). This saves you the logic of having to `sed`-replace the old and new versions within various workflows and test scripts, and allows us to update the entire supported version range, rather than just the last two versions.

Would that also solve the auth issues and keep things as simple as what we do in https://github.com/github/vscode-codeql/blob/main/.github/workflows/release.yml#L91-L128?

.github/workflows/bump-cli.yml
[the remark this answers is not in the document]
tjgurwara99: > Giving the workflows permission to the GITHUB_TOKEN (if that's possible)

That isn't possible according to the documentation I think. Let me know if I missed something in the docs though 😅 - I have a habit of skim reading docs and jumping to conclusions. 

> Avoiding changes to the workflow file itself in the automatically generated PR, instead modifying a separate file that records the version numbers and gets parsed by the first job in the generated workflow to determine the matrix (we have some prior art on this if that's useful)

I would personally prefer that actually. The way I see it, if we don't need to modify anything in the workflow file, then the default `GH_TOKEN` suffices for us. @adityasharad you mentioned there is prior art to doing using external data in actions, could you point me to it? I would much prefer that approach 😄 

> @tjgurwara99 I've created a PR on your fork for using a GitHub App: https://github.com/tjgurwara99/vscode-codeql/pull/10. It uses an action on my personal account (https://github.com/koesie10/create-app-pr-action), so it might be that we need to follow the release process for open source releases if we want to have this in the GitHub organization.

Thanks @koesie10, I have left some comments on the PR - though I'm reluctant to use an app because, well there are more things that we own if we create it. I would much rather do this cleanly in place than create an extra repository that we'd have to own and most likely forget about. Though again, if external actions are allowed then I'm more than open to it. Though it still needs some credentials and I'd argue that they are equally as risky as PAT exposure (though maybe less so in theory). So I'm still in favour of not having any new secrets - I'll try to concoct something later today and re-review your custom action (I reviewed it briefly but need to give it a thorough review 😄). Let me know what are people's thoughts on having either btw.

adityasharad:
> > High-level comment: I wonder if keeping a separate JSON file with the list of CLI versions would help. The various tests could load that file to determine what CLI versions to test against. Then your PR would only need to modify that file (and you can skip creating the PR if there is no diff). This saves you the logic of having to `sed`-replace the old and new versions within various workflows and test scripts, and allows us to update the entire supported version range, rather than just the last two versions.
> 
> Would that also solve the auth issues and keep things as simple as what we do in https://github.com/github/vscode-codeql/blob/main/.github/workflows/release.yml#L91-L128?

Yes I believe so.

charisk:
> > > High-level comment: I wonder if keeping a separate JSON file with the list of CLI versions would help. The various tests could load that file to determine what CLI versions to test against. Then your PR would only need to modify that file (and you can skip creating the PR if there is no diff). This saves you the logic of having to `sed`-replace the old and new versions within various workflows and test scripts, and allows us to update the entire supported version range, rather than just the last two versions.
> > 
> > 
> > Would that also solve the auth issues and keep things as simple as what we do in https://github.com/github/vscode-codeql/blob/main/.github/workflows/release.yml#L91-L128?
> 
> Yes I believe so.

Seems like the right approach to me. Keeps things simple.

tjgurwara99:
I think everything is set now, now we just need to provide the default token with the ability to write in the repository settings. I tried it in my fork and it seems to be working but I don't have the permission to do that in this repo. @aeisenberg @adityasharad could you review this PR again 🙏🏼 

Screenshot 2022-12-16 at 13 41 05

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [main]
+    paths:
+      - .github/actions/create-pr/action.yml
+  schedule:
+  - cron: 0 0 */14 * * # run every 14 days
+
+jobs:
```
adityasharad, line 11: Add a `permissions` block to this job or workflow, with `contents: write` and `pull-requests: write` permissions. That should be enough to allow the workflow to create a PR _when run from the parent repo_. It will still not allow the workflow to create a PR when run from a fork, which I think is preferable.

With this change, I think we can avoid changing the setting to grant read+write permissions to workflows on PRs from forks.

adityasharad (commented):
Looks promising. A few further suggestions.

.github/workflows/bump-cli.yml
tjgurwara99: Oh ok. I did try that but it failed, I thought maybe it still needed that permission - didn't know it was because it was from the fork, it got rejected. That's great, thanks!

scripts/replace-cli-version.sh
```diff
@@ -0,0 +1,13 @@
...
+# we are exporting these variables so that we can access these variables in the workflow
+LATEST_VERSION=$(echo $VERSIONS | awk '{ print $1 }')
+PREVIOUS_VERSION=$(echo $VERSIONS | awk '{ print $2 }')
+
+echo "LATEST_VERSION=$LATEST_VERSION" >> $GITHUB_ENV
+echo "PREVIOUS_VERSION=$PREVIOUS_VERSION" >> $GITHUB_ENV
+
+sed -i "s/$PREVIOUS_VERSION/$LATEST_VERSION/g" supported_cli_versions.json
+sed -i "s/$PREVIOUS_VERSION/$LATEST_VERSION/g" extensions/ql-vscode/src/vscode-tests/ensureCli.ts
```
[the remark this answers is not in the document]
tjgurwara99: It can probably ready from that, I didn't touch it because I've been able to run the project directly in my local 😂 which was why I asked yesterday to help me with the setup but didn't get time 😓

.github/workflows/bump-cli.yml
adityasharad: It still won't work on your fork, which is ok. The `write` permission in the workflow gets downgraded to `read` on forks, thanks to the repo setting you noticed. It's safer to have read-only permissions on forks than trying to reason about whether we are granting the correct set of permissions to contributor PRs from forks.
 
If you're making automation changes in future it may be easier to use a branch on the parent repo (otherwise forks are totally fine).

scripts/replace-cli-version.sh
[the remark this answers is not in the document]
aeisenberg: Happy to help if you have setup problems.

supported_cli_versions.json
```diff
@@ -0,0 +1,8 @@
+[
+  "v2.7.6",
```
aeisenberg, line 2: Nice!

Can you make sure that the most recent, non-nightly version is first in the list? This will make it easier if we want to grab the latest version in other parts of the code. Just need to grab the first element of the array.

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,38 @@
...
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 1
+      - name: Bump CLI
+        env:
+          GH_TOKEN: ${{ github.token }}
```
aeisenberg, line 26: I might be mistaken, but I think this is what you need to do:

```suggestion
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: |
+          scripts/replace-cli-version.sh
+      - name: Commit, Push and Open a PR
+        uses: ./.github/actions/create-pr
+        with:
+          token: ${{ github.token }}
```
aeisenberg, line 32: Also, here:

```suggestion
          token: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/bump-cli.yml
tjgurwara99: The priority level for `gh` cli is `GH_TOKEN` then `GITHUB_TOKEN` I think, but both work; can change it if you'd like 😄 

```sh
> gh help environment
GH_TOKEN, GITHUB_TOKEN (in order of precedence): an authentication token for github.com
API requests. Setting this avoids being prompted to authenticate and takes precedence over
previously stored credentials.
...
```

.github/workflows/bump-cli.yml
tjgurwara99: I will change it 😄 Also, just wondering if there is any difference between the two? I thought the action context should make these equivalent 😅

adityasharad: They are indeed functionally equivalent, according to https://docs.github.com/en/actions/security-guides/automatic-token-authentication. I personally prefer using the `secrets.` syntax since it is more obvious that it's an encrypted secret.

.github/workflows/bump-cli.yml
aeisenberg: Thanks...TIL.

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,38 @@
...
+          fetch-depth: 1
+      - name: Bump CLI
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          scripts/replace-cli-version.sh
+      - name: Commit, Push and Open a PR
+        uses: ./.github/actions/create-pr
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
aeisenberg, line 32: Looks like the token doesn't have permission to push. I would think that `contents: write` should be enough.

.github/actions/create-pr/action.yml
```diff
@@ -0,0 +1,82 @@
...
+      run: |
+        git config --global user.email "github-actions@github.com"
+        git config --global user.name "github-actions[bot]"
+    - name: Commit, Push and Open PR
+      shell: bash
+      env:
+        COMMIT_MESSAGE: ${{ inputs.commit-message }}
+        HEAD_BRANCH: ${{ inputs.head-branch }}
+        BASE_BRANCH: ${{ inputs.base-branch }}
+        GH_TOKEN: ${{ inputs.token }}
```
aeisenberg, line 54: I wonder if this needs to change to `GITHUB_TOKEN` so that the git commands can work.

.github/workflows/bump-cli.yml
adityasharad: It's the right permission, but it won't work while this PR runs from a fork (by design). All write permissions get downgraded to read on forks.

.github/actions/create-pr/action.yml
tjgurwara99: I think, the checkout step has already taken care of the ambiguity. Both `GH_TOKEN` and `GITHUB_TOKEN` work 😄 - I also think there might actually be no need for this since the `GH_TOKEN` is not being used in this action at all...

tjgurwara99: Yes, but this is after the checkout which sets up the whole `gh` cli stuff right?

adityasharad: I don't think so. I believe the `checkout` step sets up Git, and uses the token information to set up Git remotes and credentials, but does not do `gh auth` for the CLI. I believe the CLI will still be expecting a token in `GH/GITHUB_TOKEN` and will attempt interactive auth if that's missing (which you don't want in CI).

tjgurwara99: Cool, I'll keep it as it is then 😄

scripts/replace-cli-version.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+VERSIONS=$(gh api -H "Accept: application/vnd.github+json" /repos/github/codeql-cli-binaries/releases | jq -r '.[].tag_name' | head -2)
+
+# we are exporting these variables so that we can access these variables in the workflow
+LATEST_VERSION=$(echo $VERSIONS | awk '{ print $1 }')
+PREVIOUS_VERSION=$(echo $VERSIONS | awk '{ print $2 }')
+
```
aeisenberg, line 8: When we do a minor version bump, we need to add the new version instead of replacing it.

Doing something like this can check if the minor version has changed:

```shell
LATEST_MINOR_VERSION="$(echo $LATEST_VERSION | awk 'BEGIN{FS=".";}{print $1$2}')"
PREVIOUS_MINOR_VERSION="$(echo $PREVIOUS_VERSION | awk 'BEGIN{FS=".";}{print $1$2}')"

if [ $LATEST_MINOR_VERSION != $PREVIOUS_MINOR_VERSION ];
then
   # handle case where minor version is changed
else
  # handle case where patch version is changed
fi
```

aeisenberg (commented):
We need to handle the case where the minor version has changed. Short term, it's ok just to fail this workflow, but it would be nice to handle it gracefully.

scripts/replace-cli-version.sh
tjgurwara99: So the list of versions in `supported_cli_versions.json` is expanding every time we have a minor version bump? For example, if we have the following in the file

```json
[
  "v2.11.5",
  "v2.7.6",
  "v2.8.5",
  "v2.9.4",
  "v2.10.5",
  "nightly"
]
```
and a version `v2.11.6` has been released. Do we want to add that like the following?

```json
[
  "v2.11.6", // this is the new version that got added
  "v2.11.5",
  "v2.7.6",
  "v2.8.5",
  "v2.9.4",
  "v2.10.5",
  "nightly"
]
```

shati-patel: `v2.11.5` -> `v2.11.6` would be a patch version bump, so in that case we want to replace!

The minor version bump would be `v2.12.0` in this case, so we want to add it:

```
[
  "v2.12.0", // this is the new minor version that got added
  "v2.11.5",
  "v2.7.6",
  "v2.8.5",
  "v2.9.4",
  "v2.10.5",
  "nightly"
]
```

(Or even, we could add the new minor version _and_ remove the oldest one (i.e. `v.2.7.6`).)

koesie10: I don't think we can actually remove 2.7.6. The removal depends on the support GHES versions and GHES 3.3 is still supported. GHES 3.3 [recommends](https://docs.github.com/en/enterprise-server@3.3/admin/code-security/managing-github-advanced-security-for-your-enterprise/configuring-code-scanning-for-your-appliance#provisioning-the-actions-for-code-scanning) CodeQL 2.7.6, so we can't remove that version yet. In general, I don't think we can remove a version automatically because it's not strictly aligned with the CodeQL CLI releases.

shati-patel: GHES 3.3 is only supported for another 6 days (2023-01-01), so it kind of aligns with the CodeQL CLI (give or take 1 week 😄). But we don't have to automate the removal (or at least not in this PR) if we want to keep that separate/manual! 🤔

tjgurwara99: Lol, I always get confused with semver 😂 For some reason I thought number after the second dot was minor release 😅 

I think the removal of a version should be done through a separate workflow/action - from what I can tell, there doesn't seem to be an automated way of telling the which version should be removed. I also think that given a manual trigger with the some inputs (like for example in the image below) and manually adding the version to remove would be better unless we want to completely automate it which may not be as simple or quick 😄 Let me know though whichever way is preferred and I can think on what I can do. PS: I think this issue was for the patch release being released every two weeks or so, so this might benefit from a separate issue which I can take (this PR has been open for a while now - I want to close it 😂 of course if it meets the acceptable standards 😉)

aeisenberg: Good points @tjgurwara99. I expect that this workflow will always be triggered manually since the timing of when to run this workflow depends on a bunch of human decisions.

I'm personally fine with focusing on patch releases for now and working on a new PR to add minor version bumps. I think we can avoid creating a second workflow for minor version bumps and handle that with a new set of inputs. 

And removing releases when a GHES version drops out of support is probably something that should be done manually since it also means we can remove certain feature flags that only pertain to the older version.

.github/workflows/bump-cli.yml
```diff
@@ -0,0 +1,38 @@
+name: Bump CLI version
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: [main]
+    paths:
+      - .github/actions/create-pr/action.yml
```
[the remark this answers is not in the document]
aeisenberg: I think this is the last change that needs to be done before we can merge.

aeisenberg (commented):
Also, please keep track of the follow-up changes for this PR. Nice work.

scripts/replace-cli-version.sh
tjgurwara99: I've added the [issue](https://github.com/github/code-scanning/issues/8597) and assigned it to myself 😄 will get to it whenever I get time

scripts/replace-cli-version.sh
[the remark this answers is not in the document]
tjgurwara99: Added an [issue](https://github.com/github/code-scanning/issues/8598) to track this.

shati-patel:
> I don't know why the workflow `Build Extension / Test` is failing - anyone has any insights on this?

Tempted to say it's just flaky 😅 Have you tried re-running?

supported_cli_versions.json
```diff
@@ -0,0 +1,8 @@
+[
+  "v2.11.5",
```
shati-patel, line 2:
```suggestion
  "v2.11.6",
```

(update since this PR was opened!)

shati-patel: This should hopefully force the required `CLI Test (v2.11.6)` checks to run

aeisenberg (approved):
Great!

Once this is merged, can you do a trial of the new workflow to see how it goes? I'm not sure if you have any examples elsewhere.

tjgurwara99:
> Once this is merged, can you do a trial of the new workflow to see how it goes?

Sure, if by trial you mean sending a workflow dispatch 😄 

> I'm not sure if you have any examples elsewhere.

The example run was in my fork [here](https://github.com/tjgurwara99/vscode-codeql/pull/12). I'm fairly confident that it would work once the workflow token changes from `read` permission to `write` permission but I can test it properly in this repo once its merged.

Also, I don't have the permission to merge this PR, could you merge it please?

aeisenberg:
Sure. I can merge. The latest test failures seem unrelated, but concerning.

aeisenberg:
Hmmm...they seem like flakes.

PR #1892, opened 2022-12-20
Simplify the credentials class, and clear up impossible error cases

robertbrignull:
The main thing being changed here, is that we remove the illusion that we prompt the user for authorization lazily, and instead we always fetch credentials and construct an authenticated octokit instance when calling `Credentials.initialize()`. I have a suspicion that a lot of the code in `Credentials` was unreachable, because the `octokit` field couldn't actually be undefined.

It's possible I am missing something here and the `Credentials` class really does need to be the way it is. If this is the case I'd like to know, and I'm happy to close or amend this PR as appropriate.

Secondly, and this is what tipped me off to this, we had a few places where we were doing
```typescript
if (!credentials) {
  throw Error("Error authenticating with GitHub");
}
```
but the problem is that `credentials` can never be undefined so this is dead code. The code makes it seem like `credentials.octokit` could be undefined at this point, but I think even this is impossible as I haven't been able to trigger it. I think this misunderstanding comes from the way that the `Credentials` class is written, so I hope this PR makes it clearer.

When removing the above exceptions I also remove some tests that were attempting to cover "when the credentials are invalid", but in reality I think these tests were not covering any code paths that could be hit in production. If there really are no credentials, because the user declined to authenticate the application, then `vscode.authentication.getSession` rejects, and thus we get an exception from the vscode library code. This is perfectly good, and I don't think we need these tests to cover it.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/authentication.ts
+++ b/extensions/ql-vscode/src/authentication.ts
@@ -34,7 +34,6 @@ export class Credentials {
   ): Promise<Credentials> {
     const c = new Credentials();
     c.registerListeners(context);
-    c.octokit = await c.createOctokit(false);
     return c;
   }
 
@@ -76,7 +75,7 @@ export class Credentials {
     context.subscriptions.push(
       vscode.authentication.onDidChangeSessions(async (e) => {
         if (e.provider.id === GITHUB_AUTH_PROVIDER_ID) {
-          this.octokit = await this.createOctokit(false);
+          this.octokit = undefined;
         }
       }),
     );
--- a/extensions/ql-vscode/src/authentication.ts
+++ b/extensions/ql-vscode/src/authentication.ts
@@ -13,42 +13,38 @@ const SCOPES = ["repo", "gist"];
  * Handles authentication to GitHub, using the VS Code [authentication API](https://code.visualstudio.com/api/references/vscode-api#authentication).
  */
 export class Credentials {
+  /**
+   * A specific octokit to return, otherwise a new authenticated octokit will be created when needed.
+   */
   private octokit: Octokit.Octokit | undefined;
 
   // Explicitly make the constructor private, so that we can't accidentally call the constructor from outside the class
   // without also initializing the class.
-  // eslint-disable-next-line @typescript-eslint/no-empty-function
-  private constructor() {}
+  private constructor(octokit?: Octokit.Octokit) {
+    this.octokit = octokit;
+  }
 
   /**
-   * Initializes an instance of credentials with an octokit instance.
+   * Initializes a Credentials instance. This will generate octokit instances
+   * authenticated as the user. If there is not already an authenticated GitHub
+   * session availabeT then the user will be prompted to log in.
    *
-   * Do not call this method until you know you actually need an instance of credentials.
-   * since calling this method will require the user to log in.
-   *
-   * @param context The extension context.
    * @returns An instance of credentials.
    */
-  static async initialize(
-    context: vscode.ExtensionContext,
-  ): Promise<Credentials> {
-    const c = new Credentials();
-    c.registerListeners(context);
-    return c;
+  static async initialize(): Promise<Credentials> {
+    return new Credentials();
   }
 
   /**
    * Initializes an instance of credentials with an octokit instance using
-   * a token from the user's GitHub account. This method is meant to be
-   * used non-interactive environments such as tests.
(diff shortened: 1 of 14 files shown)


robertbrignull:
> One of the requirements of the GitHub database download is that you're able to do so without being logged in. By dropping this parameter, you suddenly need to authenticate before being able to download any database from GitHub.

I think that's handled a different way so it's not affected by this change. In `promptImportGithubDatabase` it takes a `credentials: Credentials | undefined` and if not given credentials then it constructs its own octokit. This is what happens in the non-canary case.

> This does change the time at which you're asked for credentials. Before, by passing `createIfNone: false` in the `Credentials.initialize` call, we would only be creating an authenticated Octokit instance if you're authenticated. If you're not authenticated, this would not prompt the user to login. That would mean you would be asked to login when we try to retrieve an Octokit instance, but now you're asked when we initialize the credentials. This might not be a problem, but it does mean you're asked to login much earlier.

I agree it does move the point where the user is prompted a little earlier, though it'll still be within the same command so the user may not notice. But I'm looking into rewriting this class again so we can keep the behaviour of only prompting the user when calling `getOcotkit`.

I've pushed some new changes which go in a different direction. It basically removes the caching of the octokit instance and creates a new one each time. I realise this is potentially quite a big change but I argue it doesn't change behaviour from the user's point of view, and I don't think the benefit of the caching is worth the code complexity. I also realise that putting more controversial changes into this PR isn't helped it get merged, but I'm not too worried about the time pressure. I could split this into two PRs if that would be better.

I've made the changes across a large number of commits and each one explains the individual step and the reasons for it, so I hope it makes sense. But if you don't agree just say and I'm happy to discuss.

I've been playing with the `codeQL.authenticateToGitHub` command to test this, and I think it's behaving exactly as before 🤞

koesie10 (approved):
Thanks for simplifying this even further! I was testing the database download while logged out, but I was in canary mode and forgot that you'd still be required to login when in canary mode. I tested again and agree that it doesn't really change any user-visible behavior.

I am wondering if we can simplify this even further. The `initializeWithToken` method is used in only one test and this could be replaced by mocking the credentials class, like we do in all other tests. Then we'd be left with a class only containing the `getOctokit` method without any instance variables, which should be easier to reason about. We could then even replace it by a simple function, but that would make mocking it quite a bit harder and require a lot of changes.

I think the PR is good as-is, and potentially removing the `initializeWithToken` method can be done in a follow-up PR.

extensions/ql-vscode/src/authentication.ts
```diff
@@ -13,102 +13,59 @@ const SCOPES = ["repo", "gist"];
...
-  private constructor() {}
+  private constructor(octokit?: Octokit.Octokit) {
+    this.octokit = octokit;
+  }
 
   /**
-   * Initializes an instance of credentials with an octokit instance.
+   * Initializes a Credentials instance. This will generate octokit instances
+   * authenticated as the user. If there is not already an authenticated GitHub
+   * session availabeT then the user will be prompted to log in.
```
norascheuch, line 30:
```suggestion
   * session available  the user will be prompted to log in.
```

extensions/ql-vscode/src/authentication.ts
```diff
@@ -13,102 +13,59 @@ const SCOPES = ["repo", "gist"];
...
-    return c;
+  static async initialize(): Promise<Credentials> {
+    return new Credentials();
   }
 
   /**
    * Initializes an instance of credentials with an octokit instance using
-   * a token from the user's GitHub account. This method is meant to be
-   * used non-interactive environments such as tests.
+   * a specific known token. This method is meant to be used non-interactive
```
norascheuch, line 40:
```suggestion
   * a specific known token. This method is meant to be used in non-interactive
```

robertbrignull:
Thank you @koesie10 and @norascheuch for the reviews!

> I am wondering if we can simplify this even further. The `initializeWithToken` method is used in only one test and this could be replaced by mocking the credentials class, like we do in all other tests. Then we'd be left with a class only containing the `getOctokit` method without any instance variables, which should be easier to reason about. We could then even replace it by a simple function, but that would make mocking it quite a bit harder and require a lot of changes.

I've also been thinking this. After this PR I'm planning to look at how we mock the credentials in tests, the hope is to make it more type safe and avoid mocking the `initialize` method. As part of that I can remove the `initializeWithToken` method as well.

PR #1893, opened 2022-12-20
Stop user from adding a db or owner with the same name

charisk:
Follows a similar pattern to the validation checks around db lists.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/databases/config/db-config-store.ts
+++ b/extensions/ql-vscode/src/databases/config/db-config-store.ts
@@ -99,6 +99,12 @@ export class DbConfigStore extends DisposableObject {
       throw Error("Cannot add remote repo if config is not loaded");
     }
 
+    if (this.doesRemoteDbExist(repoNwo)) {
+      throw Error(
+        `A remote repository with the name '${repoNwo}' already exists`,
+      );
+    }
+
     const config: DbConfig = cloneDbConfig(this.config);
     config.databases.remote.repositories.push(repoNwo);
 
@@ -108,6 +114,10 @@ export class DbConfigStore extends DisposableObject {
   public async addRemoteOwner(owner: string): Promise<void> {
     if (!this.config) {
       throw Error("Cannot add remote owner if config is not loaded");
+    }
+
+    if (this.doesRemoteOwnerExist(owner)) {
+      throw Error(`A remote owner with the name '${owner}' already exists`);
     }
 
     const config: DbConfig = cloneDbConfig(this.config);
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/databases/db-manager.ts
```diff
@@ -76,10 +76,24 @@ export class DbManager {
   }
 
   public async addNewRemoteRepo(nwo: string): Promise<void> {
+    if (nwo === "") {
+      throw new Error("Repository name cannot be empty");
+    }
+    if (this.dbConfigStore.doesRemoteDbExist(nwo)) {
+      throw new Error(`The repository '${nwo}' already exists`);
+    }
```
norascheuch, line 84: We do this check twice now - is it really necessary? I think it would be better to decide who should take care of the validation and do the check only once.

extensions/ql-vscode/src/databases/db-manager.ts
```diff
@@ -76,10 +76,24 @@ export class DbManager {
...
     await this.dbConfigStore.addRemoteRepo(nwo);
   }
 
   public async addNewRemoteOwner(owner: string): Promise<void> {
+    if (owner === "") {
+      throw Error("Owner name cannot be empty");
+    }
+    if (this.dbConfigStore.doesRemoteOwnerExist(owner)) {
+      throw Error(`The owner '${owner}' already exists`);
+    }
```
norascheuch, line 95: We do this check twice now - is it really necessary? I think it would be better to decide who should take care of the validation and do the check only once.

extensions/ql-vscode/src/databases/config/db-config-store.ts
```diff
@@ -144,6 +154,32 @@ export class DbConfigStore extends DisposableObject {
...
+      throw Error(
+        "Cannot check remote database existence if config is not loaded",
+      );
+    }
+
+    if (listName) {
+      return this.config.databases.remote.repositoryLists.some(
+        (l) => l.name === listName && l.repositories.includes(dbName),
+      );
+    }
```
norascheuch, line 168: Optional: I'm wondering if we can solve this in a more general way:

```
this.configErrors = this.configValidator.validate(config);
if (this.configErrors.length > 0) {
  this.configErrors.forEach((error) => {
    throw Error(error.message);
  });
}
```

and use it for alle checks? Our error messages would be a bit more general - I checked for adding a new repo and it looks good (albeit we could add an 'Error: ' at the beginning of the message):

image

I think it would clean up the file and make it easier to read as well as make sure we catch all errors.

norascheuch (approved):
Thanks for making the change! I have two optional suggestions.

extensions/ql-vscode/src/databases/config/db-config-store.ts
charisk: Thanks for thinking about readability! 

We don't want to throw an exception in this case, we just want to check if something already exists and make a decision in the UI based on that. I think the fact that I'm throwing an exception further up is confusing (and wrong) so I'll change it as I mention in https://github.com/github/vscode-codeql/pull/1893/files#r1054201240.

Essentially, I don't think we should have unhandled exceptions for things the user has done "wrong".

norascheuch: 🤔 Hm then I'm misunderstanding something, as we're throwing errors everywhere in this file and they are shown in the UI..

extensions/ql-vscode/src/databases/db-manager.ts
norascheuch: I don't understand this, check three times and throw errors three times? :o

extensions/ql-vscode/src/databases/config/db-config-store.ts
charisk: The db-config-store adds checks and throws error to avoid the config becoming invalid.

However, in reality we _should_ never reach those errors. The UI should never try to add a new list if it already exists. It should first check if it exists and then either show an error to the user or add the list. Note that I say _should_ never because a) we could introduce bugs in our UI code, so this is another line of defence and b) timing could in theory cause this (e.g. if a user does some UI actions + edits the db config file directly, at the same time)

norascheuch: 👍 Ok, sounds good, thanks for the explanation!

PR #1895, opened 2022-12-21
Handle db validation errors gracefully

charisk:
We have a mixture of validation checks and unhandled exceptions in various places. This PR aims to tidy that up so that:
- Validation errors are handled gracefully in the UI (`DbPanel`), using error notifications rather than triggering the global error handler.
- The `DbConfigStore` throws exceptions in the unlikely event that a caller has asked it to do something that would get the config in an invalid state (e.g. to add a list with an empt name).
- The `DbManager` should just pass through without doing the same checks. 

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/databases/ui/db-panel.ts
+++ b/extensions/ql-vscode/src/databases/ui/db-panel.ts
@@ -128,10 +128,12 @@ export class DbPanel extends DisposableObject {
     const nwo = getNwoFromGitHubUrl(repoName) || repoName;
     if (!isValidGitHubNwo(nwo)) {
       void showAndLogErrorMessage(`Invalid GitHub repository: ${repoName}`);
+      return;
     }
 
     if (this.dbManager.doesRemoteRepoExist(nwo)) {
       void showAndLogErrorMessage(`The repository '${nwo}' already exists`);
+      return;
     }
 
     await this.dbManager.addNewRemoteRepo(nwo);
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/databases/ui/db-panel.ts
```diff
@@ -145,7 +150,11 @@ export class DbPanel extends DisposableObject {
 
     const owner = getOwnerFromGitHubUrl(ownerName) || ownerName;
     if (!isValidGitHubOwner(owner)) {
-      throw new Error(`Invalid user or organization: ${owner}`);
+      void showAndLogErrorMessage(`Invalid user or organization: ${owner}`);
+    }
+
+    if (this.dbManager.doesRemoteOwnerExist(owner)) {
+      void showAndLogErrorMessage(`The owner '${owner}' already exists`);
```
shati-patel, line 157: From testing: when I add a duplicate (for repos/owners/lists), I now get two error messages 🙈 

image

Can we only show one of them?

extensions/ql-vscode/src/databases/ui/db-panel.ts
```diff
@@ -145,7 +150,11 @@ export class DbPanel extends DisposableObject {
 
     const owner = getOwnerFromGitHubUrl(ownerName) || ownerName;
     if (!isValidGitHubOwner(owner)) {
-      throw new Error(`Invalid user or organization: ${owner}`);
+      void showAndLogErrorMessage(`Invalid user or organization: ${owner}`);
```
shati-patel, line 153: From testing: When I enter a bad owner name, I get this error, but the bad name still gets added to the config (causing it to error).

image

shati-patel (commented):
Thanks 🧹 A couple of initial comments, just from poking around locally!

norascheuch (approved):
Looks good, thanks for cleaning this up!

extensions/ql-vscode/src/databases/ui/db-panel.ts
charisk: 🤦 I forgot to `return` after showing notifications.

extensions/ql-vscode/src/databases/ui/db-panel.ts
charisk: This should be fixed now.

shati-patel (approved):
Looks good now, thanks for the fixes 😸

PR #1896, opened 2022-12-21
Update dependencies in workflow files

shati-patel:
I'd expect dependabot to update dependencies in our actions workflows (specifically, I noticed that we're using an old version of https://github.com/peter-evans/create-pull-request in our [release workflow](https://github.com/github/vscode-codeql/blob/082d4b8c78236c82dbeab7ad1377e1c2d740fa71/.github/workflows/release.yml#L119), which has an outdated version of `node` and outdated `set-output` syntax. See warnings [here](https://github.com/github/vscode-codeql/actions/runs/3685687906).)

It looks like our dependabot configuration for `github-actions` was wrong: we need to set the directory to `"/"` instead of `".github"`. (Found this in the docs [here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directory)!)

## Checklist

No user impact

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,7 @@ updates:
       - dependency-name: "*"
         update-types: ["version-update:semver-minor", "version-update:semver-patch"]
   - package-ecosystem: "github-actions"
-    directory: ".github"
+    directory: "/"
     schedule:
       interval: "weekly"
       day: "thursday" # Thursday is arbitrary


charisk (approved):
Great find!

aeisenberg:
Thanks for fixing! It would be nice to use `gh pr create` instead of the third party action. We added that action before `gh` was even a thing. Using the CLI should be simpler.

charisk:
> Thanks for fixing! It would be nice to use `gh pr create` instead of the third party action. We added that action before `gh` was even a thing. Using the CLI should be simpler.

Thanks @aeisenberg for the background context! I've raised an issue to cover this in our internal repo.

PR #1900, opened 2022-12-21
Add new repositories to a highlighted user defined list

norascheuch:
When a remote user defined list is highlighted when pressing the 'Add new Database'-Button we only want to show the option to add a new repository (instead of also adding a new owner) and add the new database to the highlighted list.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases/config/db-config-store.ts
+++ b/extensions/ql-vscode/src/databases/config/db-config-store.ts
@@ -114,11 +114,14 @@ export class DbConfigStore extends DisposableObject {
 
     const config: DbConfig = cloneDbConfig(this.config);
     if (parentList) {
-      config.databases.remote.repositoryLists.forEach((list) => {
-        if (list.name === parentList) {
-          list.repositories.push(repoNwo);
-        }
-      });
+      const parent = config.databases.remote.repositoryLists.find(
+        (list) => list.name === parentList,
+      );
+      if (!parent) {
+        throw Error(`Cannot find parent list '${parentList}'`);
+      } else {
+        parent.repositories.push(repoNwo);
+      }
     } else {
       config.databases.remote.repositories.push(repoNwo);
     }
--- a/extensions/ql-vscode/src/databases/config/db-config-store.ts
+++ b/extensions/ql-vscode/src/databases/config/db-config-store.ts
@@ -94,7 +94,10 @@ export class DbConfigStore extends DisposableObject {
     await this.writeConfig(config);
   }
 
-  public async addRemoteRepo(repoNwo: string): Promise<void> {
+  public async addRemoteRepo(
+    repoNwo: string,
+    parentList?: string,
+  ): Promise<void> {
     if (!this.config) {
       throw Error("Cannot add remote repo if config is not loaded");
     }
@@ -110,8 +113,15 @@ export class DbConfigStore extends DisposableObject {
     }
 
     const config: DbConfig = cloneDbConfig(this.config);
-    config.databases.remote.repositories.push(repoNwo);
-
+    if (parentList) {
+      config.databases.remote.repositoryLists.forEach((list) => {
+        if (list.name === parentList) {
+          list.repositories.push(repoNwo);
+        }
+      });
+    } else {
+      config.databases.remote.repositories.push(repoNwo);
+    }
     await this.writeConfig(config);
   }
 
--- a/extensions/ql-vscode/src/databases/db-manager.ts
+++ b/extensions/ql-vscode/src/databases/db-manager.ts
@@ -79,13 +79,6 @@ export class DbManager {
     nwo: string,
     parentList?: string,
   ): Promise<void> {
-    if (nwo === "") {
-      throw new Error("Repository name cannot be empty");
-    }
-    if (this.dbConfigStore.doesRemoteDbExist(nwo)) {
-      throw new Error(`The repository '${nwo}' already exists`);
-    }
-
     await this.dbConfigStore.addRemoteRepo(nwo, parentList);
   }
 
--- a/extensions/ql-vscode/src/databases/db-manager.ts
+++ b/extensions/ql-vscode/src/databases/db-manager.ts
@@ -75,8 +75,18 @@ export class DbManager {
     await this.dbConfigStore.updateExpandedState(newExpandedItems);
   }
 
-  public async addNewRemoteRepo(nwo: string): Promise<void> {
-    await this.dbConfigStore.addRemoteRepo(nwo);
+  public async addNewRemoteRepo(
+    nwo: string,
+    parentList?: string,
+  ): Promise<void> {
+    if (nwo === "") {
+      throw new Error("Repository name cannot be empty");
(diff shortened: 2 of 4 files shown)


extensions/ql-vscode/src/vscode-tests/minimal-workspace/databases/db-panel.test.ts
```diff
@@ -443,6 +443,91 @@ describe("db panel", () => {
...
+      const repo1 = remoteRootNode.children.find(
+        (c) =>
+          c.dbItem?.kind === DbItemKind.RemoteRepo &&
+          c.dbItem?.repoFullName === "owner1/repo1",
+      );
+
+      expect(remoteRepos.length).toBe(1);
+      expect(remoteRepos[0]).toBe(repo1);
+
+      await dbManager.addNewRemoteRepo("owner2/repo2");
```
norascheuch, line 472: These tests were copied from `describe("addNewList",`. I'm concerned we're calling the `dbManager` directly instead of using the `db-panel` (as it's a db-panel test).

Is it possible that we should move these tests into the dbManager-test-file?..

extensions/ql-vscode/src/vscode-tests/minimal-workspace/databases/db-panel.test.ts
```diff
@@ -443,6 +443,91 @@ describe("db panel", () => {
...
+      const list1 = remoteRootNode.children.find(
+        (c) =>
+          c.dbItem?.kind === DbItemKind.RemoteUserDefinedList &&
+          c.dbItem?.listName === "my-list-1",
+      );
+
+      expect(remoteUserDefinedLists.length).toBe(1);
+      expect(remoteUserDefinedLists[0]).toBe(list1);
+
+      await dbManager.addNewRemoteRepo("owner2/repo2", "my-list-1");
```
norascheuch, line 514: See my comment above. Also: to call the `db-panel` directly we need to stub the `private async getHighlightedDbItem()`-method on the panel. Any ideas how we can achieve that?

extensions/ql-vscode/src/databases/ui/db-panel.ts
```diff
@@ -82,40 +82,49 @@ export class DbPanel extends DisposableObject {
...
-    }
-    if (databaseKind.kind === "repo") {
-      await this.addNewRemoteRepo();
-    } else if (databaseKind.kind === "owner") {
-      await this.addNewRemoteOwner();
+    const highlightedItem = await this.getHighlightedDbItem();
+
+    if (highlightedItem?.kind === DbItemKind.RemoteUserDefinedList) {
+      await this.addNewRemoteRepo(highlightedItem.listName);
+    } else {
```
shati-patel, line 89: From the (internal) issue: 
> If a list or item inside a list is highlighted, add it to that list

So I think that means something like this? (i.e. also checking if the highlighted item is inside a list)
```suggestion
    if (highlightedItem?.kind === DbItemKind.RemoteUserDefinedList) {
      await this.addNewRemoteRepo(highlightedItem.listName);
    } else if (
      highlightedItem?.kind === DbItemKind.RemoteRepo &&
      highlightedItem.parentListName
    ) {
      await this.addNewRemoteRepo(highlightedItem.parentListName);
    } else {
```

extensions/ql-vscode/src/vscode-tests/minimal-workspace/databases/db-panel.test.ts
shati-patel: I think we weren't able to test/stub the `db-panel` methods directly, so that's why we're testing the `db-manager` methods instead 😬

extensions/ql-vscode/src/vscode-tests/minimal-workspace/databases/db-panel.test.ts
shati-patel: You're right: this test file doesn't actually test the panel, rather the manager, config store, tree data provider, and the interactions between them 🤔 

It probably makes sense to restructure/rename this test file, but that might be a question for a future PR 😅 Happy to keep this as-is for now.

shati-patel (commented):
Great! Thanks for adding this ⚡  One comment, and I tried answering your questions (though you might want a more informed opinion 😅)

extensions/ql-vscode/src/databases/ui/db-panel.ts
norascheuch: Oh yes, didn't think about that! Will add the case!

extensions/ql-vscode/src/databases/db-manager.ts
```diff
@@ -75,8 +75,18 @@ export class DbManager {
     await this.dbConfigStore.updateExpandedState(newExpandedItems);
   }
 
-  public async addNewRemoteRepo(nwo: string): Promise<void> {
-    await this.dbConfigStore.addRemoteRepo(nwo);
+  public async addNewRemoteRepo(
+    nwo: string,
+    parentList?: string,
+  ): Promise<void> {
+    if (nwo === "") {
```
charisk, line 82: I think these checks should happen in the `DbConfigStore` if they aren't already, and the `DbManager` should just pass through to that. That's what we did elsewhere.

extensions/ql-vscode/src/databases/config/db-config-store.ts
```diff
@@ -110,8 +113,15 @@ export class DbConfigStore extends DisposableObject {
     }
 
     const config: DbConfig = cloneDbConfig(this.config);
-    config.databases.remote.repositories.push(repoNwo);
-
+    if (parentList) {
+      config.databases.remote.repositoryLists.forEach((list) => {
```
charisk, line 117: This is fine, but it will keep looping even after it found the list, which is a little strange. I don't expect it to be a problem because we don't anticipate large numbers of lists.

We could instead find the list with the given name, and update its `repositories`.

charisk, line 117: Shall we also check whether the parent list exists at all and throw an error if it doesn't?

charisk (approved):
Good stuff, just some suggestions!

extensions/ql-vscode/src/databases/db-manager.ts
norascheuch: Hm, this must've sneaked in during a rebase. Great..

extensions/ql-vscode/src/databases/config/db-config-store.ts
norascheuch: Hm, in which case could it not exist?.. We can check for it though.

norascheuch: 👍

charisk: It's one of those things that _should never_ happen.

charisk (approved):
LGTM! 🎉

shati-patel (approved):
Not sure if you were waiting for another review from me - all looks good :)

PR #1902, opened 2022-12-21
Add workspace state to app container

charisk:
We'd like to be able to set the workspace state without necessarily depending on the VS Code API. In this PR we add a `workspaceState` to the `App` container.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/common/memento.ts
+++ b/extensions/ql-vscode/src/common/memento.ts
@@ -1,0 +1,44 @@
+/**
+ * A memento represents a storage utility. It can store and retrieve
+ * values.
+ *
+ * It is an interface used by the VS Code API. We replicate it here
+ * to avoid the dependency to the VS Code API.
+ */
+export interface Memento {
+  /**
+   * Returns the stored keys.
+   *
+   * @return The stored keys.
+   */
+  keys(): readonly string[];
+
+  /**
+   * Return a value.
+   *
+   * @param key A string.
+   * @return The stored value or `undefined`.
+   */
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/common/memento.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * A memento represents a storage utility. It can store and retrieve
+ * values.
+ *
+ * It is an interface used by the VS Code API. We replicate it here
+ * to avoid the dependency to the VS Code API.
```
shati-patel, line 6: More of a general rambly question: if we upgrade our VS Code version and the "real" `Memento` interface changes, how will we know to update our copy too? I'm trying to think where the type errors will show up 🤔 

(I'm guessing we'll get type errors anywhere where we use the real `Memento`, and then we'll have to remember to update it here?)

shati-patel (approved):
🚢

extensions/ql-vscode/src/common/memento.ts
charisk: Yes, correct. If it's a breaking change, we'll see type errors where we do `extensionContext.workspaceState` expecting to return our own `Memento`. If it's new functionality, we'll need to remember to update the interface. But perhaps we'll just do it on demand.

It's not great having to do this, but it works. This is what we've done with `Disposable` elsewhere. Since these are foundational pieces I don't expect much churn (famous last words :D).

shati-patel: Makes sense, thanks!

PR #1903, opened 2022-12-22
Introduce idea of centrally registered credentials

robertbrignull:
I'm aware that this PR is quite a big move in a certain direction and it's not something we've discussed as a team. I'm very open to discussing other options and I won't be insulted if this idea doesn't make it. I'm opening this PR as a potential solution and a talking point.

The point of this PR is to improve how we mock `getOctokit` in tests. What we do is:
- Introduce a new `extensions/ql-vscode/src/pure/authentication.ts` file that can be accessed from production code and tests. This file exports two methods:
  - `getOctokit` which can be called from any location and will automatically use the correct credentials for the environment.
  - `registerCredentials` which sets the credentials to be used for subsequent calls to `getOctokit`. This is called once when initialising the extension, as well as in test setup.
- Introduce a new `extensions/ql-vscode/src/vscode-tests/factories/authentication.ts` file which contains multiple helpful ways of constructing credentials for tests.

The benefits are:
- We aren't mocking `Credentials.initialize`. It felt like using too much internal knowledge of the class, and it felt bad referencing code that is part of the VSCode implementation.
- If a test fails to set up mocking, then it'll fail with a helpful error message instead of trying to prompt the user for credentials or making a real unauthenticated HTTP request.
  - Also if a test fails to remove the mocking, then the next test that tries to set up the mocking should fail and we'll notice.
- Avoid having to pass `Credentials` object around methods, when it's a stateless object and there's only one implementation that we want to use in production.

## Checklist


extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -297,10 +294,7 @@ export class DatabaseUI extends DisposableObject {
       commandRunnerWithProgress(
         "codeQLDatabases.chooseDatabaseGithub",
         async (progress: ProgressCallback, token: CancellationToken) => {
-          const credentials = isCanary()
```
robertbrignull, line 300: Note that this canary check has been moved into `promptImportGithubDatabase`, so there shouldn't be any behaviour change and we shouldn't be forcing non-canary users to log in.

robertbrignull:
I've changed the return type of `registerCredentials` to be a `Disposable`, and also moved it to the `/common` directory instead of `/pure`.

robertbrignull:
I had a discussion with @charisk today and we're going to have a go at incorporating this into the App Container

robertbrignull:
I/we will have another go at implementing this, but I'll close this PR and we'll make a new one.

PR #1909, opened 2022-12-23
Move test data closer to the test that is using them

charisk:
Moved `data` and `data2` directories that are only used by the `file.test.ts` into a directory specific to those tests.

## Checklist
N/A:


koesie10 (changes requested):
The `test/data` directory is also used by the [minimal-workspace](https://github.com/github/vscode-codeql/blob/24a7348fc21262032c8e50128325c5eada811860/extensions/ql-vscode/src/vscode-tests/minimal-workspace/jest-runner-vscode.config.js#L14) and [cli-integration](https://github.com/github/vscode-codeql/blob/24a7348fc21262032c8e50128325c5eada811860/extensions/ql-vscode/src/vscode-tests/cli-integration/jest-runner-vscode.config.js#L20) tests, so I don't think we can actually move these without affecting those tests.

Because those two suites are also using them, `test/data` is probably the best path they can be in. I don't think we should move them to the `pure-tests` directory.

charisk:
> The `test/data` directory is also used by the [minimal-workspace](https://github.com/github/vscode-codeql/blob/24a7348fc21262032c8e50128325c5eada811860/extensions/ql-vscode/src/vscode-tests/minimal-workspace/jest-runner-vscode.config.js#L14) and [cli-integration](https://github.com/github/vscode-codeql/blob/24a7348fc21262032c8e50128325c5eada811860/extensions/ql-vscode/src/vscode-tests/cli-integration/jest-runner-vscode.config.js#L20) tests, so I don't think we can actually move these without affecting those tests.
> 
> Because those two suites are also using them, `test/data` is probably the best path they can be in. I don't think we should move them to the `pure-tests` directory.

You're right, I only just realised - didn't run integration tests locally.

PR #1911, opened 2022-12-23
Switch bundler from ESBuild to Webpack

koesie10:
In terms of keeping down the number of dependencies and different tools we are using, this switches the bundler for the main extension module from ESBuild to Webpack to keep it in line with the bundler used for the webview.

:warning: This is based on #1799 and is just for testing whether this also works (for now).

## Checklist


koesie10:
For now, let's see how well ESBuild works. If we run into any problems or want to unify the two builds, we can think about switching this to Webpack.

PR #1921, opened 2023-01-03
Add context menu command to open repo/owner on GitHub

shati-patel:
Adds an "Open on GitHub" context menu action for remote repos/owners in the DB panel:
image

## Checklist

N/A—internal only 🍝

### DIFF
--- a/extensions/ql-vscode/src/databases/ui/db-panel.ts
+++ b/extensions/ql-vscode/src/databases/ui/db-panel.ts
@@ -1,7 +1,9 @@
 import {
+  commands,
   QuickPickItem,
   TreeView,
   TreeViewExpansionEvent,
+  Uri,
   window,
   workspace,
 } from "vscode";
@@ -18,6 +20,7 @@ import { DbItem, DbItemKind, DbListKind, remoteDbKinds } from "../db-item";
 import { DbManager } from "../db-manager";
 import { DbTreeDataProvider } from "./db-tree-data-provider";
 import { DbTreeViewItem } from "./db-tree-view-item";
+import { getGitHubUrl } from "./db-tree-view-item-action";
 
 export interface RemoteDatabaseQuickPickItem extends QuickPickItem {
   kind: string;
@@ -81,6 +84,12 @@ export class DbPanel extends DisposableObject {
       commandRunner(
         "codeQLDatabasesExperimental.setSelectedItemContextMenu",
         (treeViewItem: DbTreeViewItem) => this.setSelectedItem(treeViewItem),
+      ),
+    );
+    this.push(
+      commandRunner(
+        "codeQLDatabasesExperimental.openOnGitHubContextMenu",
+        (treeViewItem: DbTreeViewItem) => this.openOnGitHub(treeViewItem),
       ),
     );
   }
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/databases/ui/db-panel.ts
```diff
@@ -290,4 +299,18 @@ export class DbPanel extends DisposableObject {
...
+      throw new Error("Unable to open on GitHub. Please select a valid item.");
+    }
+    const githubUrl = getGitHubUrl(treeViewItem.dbItem);
+    if (!githubUrl) {
+      throw new Error(
+        "Unable to open on GitHub. Please select a remote repository or owner.",
+      );
+    }
+
+    await commands.executeCommand("vscode.open", Uri.parse(githubUrl));
```
shati-patel, line 314: FYI, I haven't added any kind of validation for whether this is a valid link, so people will get a 404 if their repo or owner isn't accessible.

I think this is okay: we check that the repo/owner is of the right "shape" in the config, and then we check whether it's accessible when running a variant analysis.

extensions/ql-vscode/src/databases/ui/db-tree-view-item-action.ts
```diff
@@ -59,3 +59,14 @@ function canBeRenamed(dbItem: DbItem): boolean {
 function canBeOpenedOnGitHub(dbItem: DbItem): boolean {
   return dbItemKindsThatCanBeOpenedOnGitHub.includes(dbItem.kind);
 }
+
+export function getGitHubUrl(dbItem: DbItem): string | undefined {
```
shati-patel, line 63: I _could_ have put this logic directly in `db-panel.ts`, but this made it slightly easier to test 😄

norascheuch: Since this method method is called get `Url` we could return an url directly (instead of a string), so that the caller doesn't have to care about conversion. 

It might make sense (not sure) to put it in `extensions/ql-vscode/src/common/github-url-identifier-helper.ts`?

extensions/ql-vscode/src/databases/ui/db-panel.ts
norascheuch, line 314: Instead of using Uri from vscode we could use URL (new URL('string')).

charisk (approved):
LGTM!

extensions/ql-vscode/src/databases/ui/db-tree-view-item-action.ts
charisk: WRT location, I think where it is makes sense - it's a function that helps us with db tree view item actions. It's not generic enough to be in `common` since it's specific to `DbItem`.

extensions/ql-vscode/src/databases/ui/db-panel.ts
shati-patel: I like this! But annoyingly it looks like `vscode.open` expects a Uri or string, so can't use URL 😿 

```
Invalid argument 'uriOrString' when running 'vscode.open', received: https://github.com/codeql (codeQLDatabasesExperimental.openOnGitHubContextMenu)
```

extensions/ql-vscode/src/databases/ui/db-tree-view-item-action.ts
shati-patel: > Since this method is called get `Url` we could return an url directly (instead of a string), so that the caller doesn't have to care about conversion.

After a bit of [experimentation](https://github.com/github/vscode-codeql/pull/1921#discussion_r1061332949), turns out the `vscode.open` command wanted a Uri or string, so I've kept this as-is!

PR #1925, opened 2023-01-04
Add remove context menu action

norascheuch:
Adds a remove context menu action.

image

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases/config/db-config-store.ts
+++ b/extensions/ql-vscode/src/databases/config/db-config-store.ts
@@ -7,7 +7,9 @@ import {
   renameLocalList,
   renameRemoteList,
   SelectedDbItem,
+  SelectedDbItemKind,
 } from "./db-config";
+import * as lodash from "lodash";
 import * as chokidar from "chokidar";
 import { DisposableObject, DisposeHandler } from "../../pure/disposable-object";
 import { DbConfigValidator } from "./db-config-validator";
@@ -113,6 +115,7 @@ export class DbConfigStore extends DisposableObject {
           );
         break;
       case DbItemKind.LocalDatabase:
+        // TODO: Remove databases from Disk once implemented
         if (dbItem.parentListName) {
           const parent = config.databases.local.lists.find(
             (list) => list.name === dbItem.parentListName,
@@ -158,14 +161,25 @@ export class DbConfigStore extends DisposableObject {
     }
 
     // Remove item from selected
-    const mappedItem = mapDbItemToSelectedDbItem(dbItem);
-    if (
-      selectedItem &&
-      JSON.stringify(mappedItem) === JSON.stringify(selectedItem)
-    ) {
-      config.selected = undefined;
+    const removedItem = mapDbItemToSelectedDbItem(dbItem);
+    if (selectedItem) {
+      // if removedItem has a parentList, check if parentList is selectedItem
+      if (
+        removedItem &&
+        (removedItem.kind === SelectedDbItemKind.LocalUserDefinedList ||
+          removedItem.kind === SelectedDbItemKind.RemoteUserDefinedList)
+      ) {
+        if (
+          (selectedItem.kind === SelectedDbItemKind.LocalDatabase ||
+            selectedItem.kind === SelectedDbItemKind.RemoteRepository) &&
+          removedItem.listName === selectedItem.listName
+        ) {
+          config.selected = undefined;
+        }
+      } else if (lodash.isEqual(removedItem, selectedItem)) {
+        config.selected = undefined;
+      }
     }
-
     await this.writeConfig(config);
   }
 
--- a/extensions/ql-vscode/src/databases/config/db-config-store.ts
+++ b/extensions/ql-vscode/src/databases/config/db-config-store.ts
@@ -22,7 +22,10 @@ import {
   LocalDatabaseDbItem,
   LocalListDbItem,
   RemoteUserDefinedListDbItem,
+  DbItem,
+  DbItemKind,
 } from "../db-item";
+import { mapDbItemToSelectedDbItem } from "../db-item-selection";
 
 export class DbConfigStore extends DisposableObject {
   public readonly onDidChangeConfig: AppEvent<void>;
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/databases/config/db-config-store.ts
```diff
@@ -76,6 +78,84 @@ export class DbConfigStore extends DisposableObject {
...
+        break;
+      default:
+        throw Error(`Type '${dbItem.kind}' cannot be removed`);
+    }
+
+    // Remove item from selected
+    const mappedItem = mapDbItemToSelectedDbItem(dbItem);
+    if (
+      selectedItem &&
+      JSON.stringify(mappedItem) === JSON.stringify(selectedItem)
```
norascheuch, line 151: Open to suggestions!! This is a very ugly solution that I would fortify with writing tests for all cases - I experimented with other options but got errors and didn't want write 100 lines of code just to compare these objects. 

Also: if this check fails and the selected item is not removed it will go unnoticed in most cases and if the edge case happens that a user runs a query directly after removing the selected item they will get an error which should be a minor inconvenience only.

extensions/ql-vscode/src/databases/config/db-config-store.ts
```diff
@@ -76,6 +78,84 @@ export class DbConfigStore extends DisposableObject {
...
+          (list) => list.name !== dbItem.listName,
+        );
+        break;
+      case DbItemKind.RemoteUserDefinedList:
+        config.databases.remote.repositoryLists =
+          config.databases.remote.repositoryLists.filter(
+            (list) => list.name !== dbItem.listName,
+          );
+        break;
+      case DbItemKind.LocalDatabase:
```
shati-patel, line 102: Once we properly support local databases, will we need to think about removing DBs on disk as well? If so, maybe add a comment about that 🤔 (This is very much a future concern though, so feel free to ignore!)

extensions/ql-vscode/src/databases/config/db-config-store.ts
```diff
@@ -76,6 +78,84 @@ export class DbConfigStore extends DisposableObject {
...
+      default:
+        throw Error(`Type '${dbItem.kind}' cannot be removed`);
+    }
+
+    // Remove item from selected
+    const mappedItem = mapDbItemToSelectedDbItem(dbItem);
+    if (
+      selectedItem &&
+      JSON.stringify(mappedItem) === JSON.stringify(selectedItem)
+    ) {
```
shati-patel, line 152: I think we also need to set `selected` to `undefined` when a database's surrounding parent list is removed!

extensions/ql-vscode/src/databases/config/db-config-store.ts
```diff
@@ -76,6 +78,84 @@ export class DbConfigStore extends DisposableObject {
...
+  public async removeDbItem(dbItem: DbItem): Promise<void> {
+    if (!this.config) {
+      throw Error("Cannot remove item if config is not loaded");
+    }
+
+    const config: DbConfig = cloneDbConfig(this.config);
+    const selectedItem: SelectedDbItem | undefined = config.selected;
+
+    // Remove item from databases
+    switch (dbItem.kind) {
```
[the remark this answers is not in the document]
shati-patel: It's possible we initially put those methods in the db-manager because it was easier for writing tests? But I don't have much of an opinion here—@charisk might have a better idea!

shati-patel (commented):
Looks good! I've left one actual comment, and then a few rambly thoughts 💭 😆

extensions/ql-vscode/src/databases/config/db-config-store.ts
charisk: Are we trying to be too generic? If the code was in the various switch cases, the shape of the `SelectedDbItem` would be known and easy to compare I think.

I'm having to do similar checks in the area I'm working on so we could go with something for now and circle back to it too if it helps.

extensions/ql-vscode/src/databases/config/db-config-store.ts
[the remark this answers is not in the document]
charisk: The existing flow (that is followed by `addNewList` etc.) was intentional and I would say I have a preference for it, but I'm happy to be convinced that this new approach is better. Here's my thinking about the current set up:
- I see the `DbManager` as the entity that coordinates things/deals with domain logic. It interacts with the config store but also knows about other stuff like the expanded state (which I don't see being cleaned up in this PR?), deletion of actual db files (in the future), etc.
- In the `DbConfigStore` we have to provide targeted functions for some stuff (e.g. `doesLocalListExist`) so it seems like having things like `addNewList` fits that pattern nicely.
Happy to have a chat if it helps!

extensions/ql-vscode/src/databases/db-item-selection.ts
```diff
@@ -80,15 +80,15 @@ export function mapDbItemToSelectedDbItem(
     case DbItemKind.LocalDatabase:
       return {
         kind: SelectedDbItemKind.LocalDatabase,
-        listName: dbItem?.parentListName,
         databaseName: dbItem.databaseName,
+        listName: dbItem?.parentListName,
```
charisk, line 84: Is there a reason we've moved these around?

norascheuch: Yes, that's the reason why `JSON.stringify(mappedItem) === JSON.stringify(selectedItem)` is so ugly: it depends on order. There doesn't seem to be an easy way to compare Javascript objects, (or sort them), so that's what we got..

norascheuch: this is not needed anymore but I'll leave it so that it's the same order everywhere

extensions/ql-vscode/src/databases/config/db-config-store.ts
norascheuch: I found good solution using a library we already have.

norascheuch:
Unit tests will be handled [here](https://github.com/github/code-scanning/issues/8455)

extensions/ql-vscode/src/databases/ui/db-panel.ts
```diff
@@ -363,6 +369,15 @@ export class DbPanel extends DisposableObject {
...
   }
 
+  private async removeItem(treeViewItem: DbTreeViewItem): Promise<void> {
+    if (treeViewItem.dbItem === undefined) {
+      throw new Error(
+        "Not a removable database item. Please select a valid item.",
+      );
+    }
+    await this.dbManager.removeDbItem(treeViewItem.dbItem);
+  }
```
charisk, line 379: Shall we also clean expanded items like we said? `updateDbItemExpandedState` now does that.

extensions/ql-vscode/src/databases/config/db-config-store.ts
```diff
@@ -88,6 +93,96 @@ export class DbConfigStore extends DisposableObject {
...
+        );
+        break;
+      case DbItemKind.RemoteUserDefinedList:
+        config.databases.remote.repositoryLists =
+          config.databases.remote.repositoryLists.filter(
+            (list) => list.name !== dbItem.listName,
+          );
+        break;
+      case DbItemKind.LocalDatabase:
+        // TODO: Remove databases from Disk once implemented
```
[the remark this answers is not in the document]
norascheuch: 👍

extensions/ql-vscode/src/databases/config/db-config-store.ts
```diff
@@ -88,6 +93,96 @@ export class DbConfigStore extends DisposableObject {
     await this.writeConfig(config);
   }
 
+  public async removeDbItem(dbItem: DbItem): Promise<void> {
+    if (!this.config) {
+      throw Error("Cannot remove item if config is not loaded");
+    }
+
+    const config: DbConfig = cloneDbConfig(this.config);
```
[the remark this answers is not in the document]
norascheuch: 👍

charisk (approved):
LGTM! Some stuff aren't ideal from a readability perspective, but we've discussed offline some refactoring that we could do to tidy things up.

Thank you for the hard work on this!

PR #1933, opened 2023-01-06
Ensure remote queries can handle codeql-pack.yml files

aeisenberg:
This updates one of our integration tests so that it uses `codeql-pack.yml` instead of `qlpack.yml`.


charisk (approved):
Change LGTM - would be cool if we could cover with a test too.

aeisenberg:
> Change LGTM - would be cool if we could cover with a test too.

I did add a test for this. One of the query packs in the test folder now uses `codeql-pack.yml` and the other uses `qlpack.yml`.

charisk (approved):
Ship it!

PR #1943, opened 2023-01-09
Avoid race condition requesting telemetry permissions

robertbrignull:
I think this is a race condition, because `setTelemetryRequested` is setting `telemetry-request-viewed` to false but `requestTelemetryPermission` is checking `telemetry-request-viewed`. So this could get the value false or it could get whatever the existing value is.

My understanding of the intention of this code is that if the user has canary mode enabled but not telemetry then we always want to prompt them to enable telemetry and we want to ignore if they've already been prompted. So therefore the intention is to set `telemetry-request-viewed` to false first and the use of `Promise.all` seems like a bug.

It's been this way all the way since 292e695646186d5b2d700aac1fc8dcdd0fbf08f2 when telemetry was first added.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/telemetry.ts
+++ b/extensions/ql-vscode/src/telemetry.ts
@@ -95,10 +95,8 @@ export class TelemetryListener extends ConfigListener {
       CANARY_FEATURES.getValue() &&
       !ENABLE_TELEMETRY.getValue()
     ) {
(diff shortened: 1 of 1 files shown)


aeisenberg (approved):
Nice catch! I think you're right. Thanks for fixing this.

PR #1945, opened 2023-01-10
Use io-ts for validating GitHub API responses

koesie10:
This introduces [`io-ts`](https://github.com/gcanti/io-ts) for validating GitHub API responses. Instead of creating interfaces and directly casting the JSON responses to these interfaces after parsing, this will create a "schema"/"type" which will validate that the JSON response matches what we expect. This ensures that we do not work with invalid data.

This does not add this for all API requests/responses, only for those related to variant analyses. I'm not sure if there are any other places where we may want to introduce this, but it seems like the only other place where the GitHub API is used is for database fetching (with just a single API response we could validate) and in remote queries (which will be removed soon).

Whether we even want to do this is also up for discussion; it makes it harder to understand the types at a glance since you need to be familiar with how `io-ts` works and it's not trivial to mix required and optional properties.

Screenshot 2023-01-10 at 13 27 05

## Checklist


koesie10:
We decided to not go ahead with this approach.

PR #1948, opened 2023-01-10
Add db manager tests for "removeDbItem" functionality

shati-patel:
[This follows the pattern of the rename tests from #1939.]

Follow-up to #1925, to add some "integration" tests. These test the integration between the db manager and db config store. (Don't require integration with the VS Code API, which is why they are in the "unit tests" folder.)

I've added a test for each of the removable db items, and checked that it's correctly removed from the config:
- remote user-defined list
- remote owner
- remote repo
- local list
- local db 

## Checklist

N/A—internal only 🎩

### DIFF
--- a/extensions/ql-vscode/test/unit-tests/databases/db-manager.test.ts
+++ b/extensions/ql-vscode/test/unit-tests/databases/db-manager.test.ts
@@ -78,9 +78,10 @@ describe("db manager", () => {
 
       await saveDbConfig(dbConfig);
 
-      const dbItem = getRemoteUserDefinedListDbItem("my-list-1");
+      const remoteListDbItems = getRemoteUserDefinedListDbItems("my-list-1");
+      expect(remoteListDbItems.length).toEqual(1);
 
-      await dbManager.renameList(dbItem, "my-list-2");
+      await dbManager.renameList(remoteListDbItems[0], "my-list-2");
 
       const dbConfigFileContents = await readDbConfigDirectly();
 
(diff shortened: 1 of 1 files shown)


shati-patel:
I _think_ the test failures are just general flakiness 🤔 😭

extensions/ql-vscode/test/unit-tests/databases/db-manager.test.ts
```diff
@@ -219,6 +388,34 @@ describe("db manager", () => {
     return localDbItems[0];
   }
 
+  function getRemoteRepoDbItem(
```
norascheuch, line 391: I'm note sure about this method and `getRemoteOwnerDbItem`.

It should test that the items are written to the config and can be retrieved using the `dbManager.getDbItems()` right? In that case I think it would be good to make the expectation explicit and let the current method return `repoDbItems` so that all expectations can be seen in the main test.

norascheuch (approved):
Looks good! One small comment otherwise 🚀

extensions/ql-vscode/test/unit-tests/databases/db-manager.test.ts
shati-patel: For those methods, I copied the same approach as for the existing ones (`getRemoteUserDefinedListDbItem` etc). But I kind of agree that it would be nice to have the expectations more visible in the main test 🔍 

Will reshuffle those slightly! 🙏🏽

norascheuch: Thanks!

shati-patel:
(on pause while we fix the async test issues!)

shati-patel:
> (on pause while we fix the async test issues!)

Rebased to pull in the changes from #1951. I'll merge this once CI passes 🤞🏽

PR #1952, opened 2023-01-11
Convert opening variant analysis logs to a command

robertbrignull:
Moves the work for opening the logs for a variant analysis into a command. The primary motivation here is to get telemetry logged for it. The alternative would be outputting telemetry explicitly, but using a command seems cleaner, and makes the code more reusable.

This only affects when opening the logs from the variant analysis results webview. The query history has its own implementation. We could look at unifying these implementations but I think we should wait until the remote queries code has been deleted first.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
@@ -19,6 +19,7 @@ import { DisposableObject } from "../pure/disposable-object";
 import { Credentials } from "../authentication";
 import { VariantAnalysisMonitor } from "./variant-analysis-monitor";
 import {
+  getActionsWorkflowRunUrl,
   isVariantAnalysisComplete,
   parseVariantAnalysisQueryLanguage,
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
```diff
@@ -585,6 +586,20 @@ export class VariantAnalysisManager
     await cancelVariantAnalysis(credentials, variantAnalysis);
   }
 
+  public async openVariantAnalysisLogs(variantAnalysisId: number) {
```
charisk, line 589: We could add an integration test for this function, but it doesn't have to be part of this PR.

charisk (approved):
LGTM!

extensions/ql-vscode/src/remote-queries/variant-analysis-manager.ts
robertbrignull: Good point. I'll look at doing that in a followup

PR #1953, opened 2023-01-11
Report telemetry on actions taken in the UI

robertbrignull:
When the user makes certain actions in the UI, output telemetry for it. This is done by sending a message from the webview back to the extension backend, which can then output telemetry in the normal way. We introduce a new telemetry event name of `ui-interaction`.

A large source of telemetry is from react state changes. If the data shown in the UI changes then there must have been a react state change, so we can use `useState` calls as an indicator of an event we want to emit telemetry for. Not all state changes should emit telemetry events, but only the ones that are the direct result of a user action. To make it easier, I've added a `useStateWithTelemetry` method that wraps `useState` and allows emitting telemetry whenever the value changes.

Another source of useful telemetry is when the user clicks a link. To cover these we have to identify the links and add `onClick` handlers to them.

I've manually tried to click every button and perform every action I can in the MRVA results view, and I think this now sends telemetry events for all actions.

## Checklist


extensions/ql-vscode/src/view/common/Telemetry.ts
```diff
@@ -0,0 +1,56 @@
+import * as React from "react";
```
koesie10, line 1: Can we rename this file to `telemetry.ts`? The only files which are upper-cased contain React components, so for utility functions or hooks I'd expect to see a lower-case filename.

Usually, I'd also prefer the file to be named the same as the hook (so `useStateWithTelemetry.ts`), but since this also contains another function I don't think that makes sense in this case.

extensions/ql-vscode/src/view/common/FileCodeSnippet/CodeSnippetMessage.tsx
```diff
@@ -65,6 +66,7 @@ export const CodeSnippetMessage = ({
               return (
                 <LocationLink
                   key={index}
+                  onClick={() => sendTelemetry("alert-message-link")}
```
koesie10, line 69: Can we extract this function to be outside of the component? By having it here, the `onClick` value will change on every re-render, which could result in a much larger part of the tree being re-rendered. I don't think that would happen here, but I think it's good practice to always keep that in mind.

There are some other files where you use this same pattern, I think those can also be extracted.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysis.tsx
```diff
@@ -60,11 +61,16 @@ export function VariantAnalysis({
...
+  const [selectedRepositoryIds, setSelectedRepositoryIds] =
+    useStateWithTelemetry<number[]>(
+      [],
+      "variant-analysis-selected-repository-ids",
+    );
   const [filterSortState, setFilterSortState] =
-    useState<RepositoriesFilterSortState>(defaultFilterSortState);
+    useStateWithTelemetry<RepositoriesFilterSortState>(
+      defaultFilterSortState,
+      "variant-analysis-filter-sort-state",
```
koesie10, line 72: This will potentially result in a lot of these events being send. I believe that if you start searching for "github", you will get 6 events, 1 for every keystroke. I'm not sure if that's the intention here.

koesie10 (commented):
I have some small suggestions, but in general it looks good. I clicked around the MRVA UI and agree with you that this should cover all events.

extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts
```diff
@@ -149,6 +150,9 @@ export class VariantAnalysisView
       case "openLogs":
         await this.openLogs();
         break;
+      case "telemetry":
```
charisk, line 153: We'd have to add this to every view. Can we add it in some generic way in `AbstractWebview` instead? (e.g. providing an implementation at the base that is called from all views)? This can be a separate change.

extensions/ql-vscode/src/view/common/Telemetry.ts
```diff
@@ -0,0 +1,56 @@
...
+export function useStateWithTelemetry<S>(
+  defaultValue: S | (() => S),
+  telemetryAction: string,
+): [S, React.Dispatch<React.SetStateAction<S>>];
+export function useStateWithTelemetry<S>(
+  defaultValue: S | (() => S),
+  telemetryAction: string,
+  filterTelemetryOnValue: (value: S) => boolean,
+): [S, React.Dispatch<S>];
+export function useStateWithTelemetry<S>(
```
robertbrignull, line 25: I realise I didn't explain why `useStateWithTelemetry` is so complicated. I wanted to add the ability to only output telemetry for certain value changes, which mostly means only when the user opens something and not when they close it. This isn't absolutely essential but it seems sensible to me to limit the telemetry to only the things we're most interested in. We can always change or removing the filtering in the future if we want to.

However, this led to problems accepting `React.SetStateAction<S>` which is `S | (prevValue: S) => S`. We need to pass the function to the `setter` and there's not a way I can find to get what the new value is. This means we can't pass the value to our filter function to know if we should emit telemetry or not.

The way around this that I've gone for is to have two different signatures, which shows that you can only use `React.SetStateAction<S>` when not using a filter function.

Do people think this is sensible and necessary, or is it over the top and just makes things complicated?

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysis.tsx
robertbrignull: That's not really the intention. Perhaps it would be worth adding a debounce here or something else to limit the number of telemetry events emitted. I don't think we want this to affect the call to `useState`, since it's currently working ok and I don't want to introduce more complexity unnecessarily, but we could have the telemetry be rate limited.

extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts
robertbrignull: I'll have a look into it. I agree we could add this to `AbstractWebview` and therefore have it be implemented for all views for free. This will also protect us against adding telemetry to a shared component and accidentally having it send a message that the view doesn't understand.

robertbrignull: I've been experimenting with a few different ways of doing this and I can't get the types to work out. I'd like it so that `FromVariantAnalysisMessage` doesn't have to include the `TelemetryMessage` type, and also not have the webview subclass have to call a super method (or at least have it be enforced). I can't get it all to work.

The types just don't like it, because when doing `SomeTemplatedType | TelemetryMessage` it can't tell that the `t` field is unique, and so doing the switch to guarantee enumerating all types doesn't work.

Also, it seems that if a superclass has an abstract method, then the subclass can implement it with different types for the parameters and it doesn't seem to complain 🤷 , which isn't helpful.

My plan currently is to leave it for this PR and look at it again in a followup, where we can get input from people on the various options without delaying this PR.

robertbrignull: I've been experimenting with a few different ways of doing this and I can't get the types to work out. I'd like it so that `FromVariantAnalysisMessage` doesn't have to include the `TelemetryMessage` type, and also not have the webview subclass have to call a super method (or at least have it be enforced). I can't get it all to work.

The types just don't like it, because when doing `SomeTemplatedType | TelemetryMessage` it can't tell that the `t` field is unique, and so doing the switch to guarantee enumerating all types doesn't work.

Also, it seems that if a superclass has an abstract method, then the subclass can implement it with different types for the parameters and it doesn't seem to complain 🤷 , which isn't helpful.

My plan currently is to leave it for this PR and look at it again in a followup, where we can get input from people on the various options without delaying this PR.

koesie10: > This will also protect us against adding telemetry to a shared component and accidentally having it send a message that the view doesn't understand.

I think we could already have this problem. The views in `common` are also used by the remote queries view, so the remote queries view will throw an error when the `telemetry` message is sent. So, I think we need to need to at least add handling of this message to the remote queries view.

extensions/ql-vscode/src/view/common/Telemetry.ts
koesie10: I think we can simplify the call signature by using a different method of sending the telemetry event. Instead of sending the telemetry event when calling the setter, we can send the telemetry event when the value has actually been updated. That means we'd use `useEffect` on the value instead of wrapping `useState`.

However, that will not necessarily make the implementation simpler since we now need to deal with the initial render and the other values in the `useEffect` callback changing, so we'd probably need to use `useRef` to keep track of the previous value. Something like this:

```ts
export function useStateWithTelemetry<S>(
  defaultValue: S | (() => S),
  telemetryAction: string,
  filterTelemetryOnValue?: (value: S) => boolean,
): [S, React.Dispatch<React.SetStateAction<S>>] {
  const previousValue = useRef(defaultValue);
  const [value, setter] = useState<S>(defaultValue);

  useEffect(() => {
    if (value === previousValue.current) {
      return;
    }
    previousValue.current = value;

    if (filterTelemetryOnValue && !filterTelemetryOnValue(value)) {
      return;
    }

    sendTelemetry(telemetryAction);
  }, [telemetryAction, filterTelemetryOnValue, value, previousValue]);

  return [value, setter];
}
```

I haven't actually tested this, so I'm not sure whether this works in practice. Since we're now not actually using the setter, we could even simplify this to a different hook which doesn't need to wrap `useState` at all and you'd be able to use `useState` as-is. Something like this:

```ts
export function useTelemetryOnChange<S>(
  value: S,
  telemetryAction: string,
  filterTelemetryOnValue?: (value: S) => boolean,
): void {
  const previousValue = useRef(value);

  useEffect(() => {
    if (value === previousValue.current) {
      return;
    }
    previousValue.current = value;

    if (filterTelemetryOnValue && !filterTelemetryOnValue(value)) {
      return;
    }

    sendTelemetry(telemetryAction);
  }, [telemetryAction, filterTelemetryOnValue, value, previousValue]);
}
```

Again, I haven't actually tested this, but I believe the initial `value` that is passed in should ensure it's not called on first render.

extensions/ql-vscode/src/remote-queries/variant-analysis-view.ts
robertbrignull: I was semi-intentionally ignoring the remote queries view since people shouldn't be using that anymore, but you're right maybe we do need to support that here until we actually remove it for good.

robertbrignull: I've added the telemetry message to the remote queries view too.

extensions/ql-vscode/src/view/common/Telemetry.ts
robertbrignull: Thanks for taking a look here and giving ideas. I think both of these are great and should work. The second option where we're not wrapping `useState` is also very convenient and will make it easier to add in more options and use it more places. This will also make it easier to implement the debounce idea from earlier too, to avoid repeated events during filtering.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysis.tsx
robertbrignull: I've now implemented this as an option to `useTelemetryOnChange`.

robertbrignull:
@koesie10, @charisk, thank you both for the review. I believe I have now addressed all comments so this is ready for a re-review.

It was mentioned earlier today maybe we don't have to send an extension message and we could emit telemetry events from within the webview. I don't know if that's possible but I'm happy to explore it in a followup PR. It won't change the bulk of this work as sending the extension messages isn't the hard bit.

extensions/ql-vscode/src/telemetry.ts
```diff
@@ -168,6 +168,23 @@ export class TelemetryListener extends ConfigListener {
     );
   }
 
+  sendUIInteraction(name: string) {
+    if (!this.reporter) {
+      return;
+    }
+
+    const isCanary = (!!CANARY_FEATURES.getValue<boolean>()).toString();
```
aeisenberg, line 176: Minor: This is already implemented: https://github.com/github/vscode-codeql/blob/koesie10/bundle-extension/extensions/ql-vscode/src/config.ts#L443-L446

extensions/ql-vscode/src/view/common/telemetry.ts
```diff
@@ -0,0 +1,60 @@
...
+  options?: {
+    filterTelemetryOnValue?: (value: S) => boolean;
+    debounceTimeoutMillis?: number;
+  },
+) {
+  const previousValue = useRef(value);
+  const filterTelemetryOnValue = options?.filterTelemetryOnValue;
+  const debounceTimeoutMillis = options?.debounceTimeoutMillis;
+
+  const sendTelemetryFunc = useMemo<() => void>(() => {
```
aeisenberg, line 27: So, the idea here is to ensure that if this effect gets invoked multiple times within the debounce period, earlier data will be overwritten with newer data?

It looks like there may be a problem. If `debounceTimeoutMillis` is set to (say) 100ms, and this is invoked every 99ms (forever), then it looks like the event will never be sent. Am I reading that right?

I _think_ what we want to do is avoid clearing the `timer`, and just make sure that each time we call this effect, then we update the `telemetryAction` to the latest action. This way, the event will always be sent at most every `debounceTimeoutMillis` and always with the latest `telemetryAction`.

extensions/ql-vscode/src/view/variant-analysis/RepoRow.tsx
```diff
@@ -157,6 +158,8 @@ const isExpandableContentLoaded = (
   return resultsLoaded;
 };
 
+const filterRepoRowExpandedTelemetry = (v: boolean) => v === true;
```
aeisenberg, line 161: If `v` is a boolean, then isn't this equivalent? Otherwise, maybe the type for `v` isn't 100% correct?

```suggestion
const filterRepoRowExpandedTelemetry = (v: boolean) => v;
```

extensions/ql-vscode/src/view/common/telemetry.ts
aeisenberg: Maybe this is a theoretical problem and in practice we'll never hit it.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysis.tsx
```diff
@@ -63,8 +61,16 @@ export function VariantAnalysis({
   const [selectedRepositoryIds, setSelectedRepositoryIds] = useState<number[]>(
     [],
   );
-  const [filterSortState, setFilterSortState] =
-    useState<RepositoriesFilterSortState>(defaultFilterSortState);
+  useTelemetryOnChange(
+    selectedRepositoryIds,
+    "variant-analysis-selected-repository-ids",
```
aeisenberg, line 66: Can you explain why this usage has no `debounceTimeoutMillis`, but the one below does?

extensions/ql-vscode/src/view/common/telemetry.ts
```diff
@@ -0,0 +1,60 @@
...
+  value: S,
+  telemetryAction: string,
+  options?: {
+    filterTelemetryOnValue?: (value: S) => boolean;
+    debounceTimeoutMillis?: number;
+  },
+) {
+  const previousValue = useRef(value);
+  const filterTelemetryOnValue = options?.filterTelemetryOnValue;
+  const debounceTimeoutMillis = options?.debounceTimeoutMillis;
```
koesie10, line 25: I think this can be simplified slightly:

```suggestion
  {
    filterTelemetryOnValue,
    debounceTimeoutMillis,
  }: {
    filterTelemetryOnValue?: (value: S) => boolean;
    debounceTimeoutMillis?: number;
  } = {},
) {
  const previousValue = useRef(value);
```

That should also allow you to use `@param` on the names of the properties, rather than on `options`.

extensions/ql-vscode/src/view/common/telemetry.ts
koesie10: I think the current behavior makes more sense. If the user doesn't stop typing, we wouldn't want to send the telemetry event yet because they haven't completed their action. I'd expect us to be interested in how many distinct searches a user does, rather than that they took 5 seconds to construct their search.

Alternatively, we could have something like a `maxWait` parameter [like the `use-debounce` library has](https://github.com/xnimorz/use-debounce#cancel-maxwait-and-memoization). If the debounce is 1 second and the max wait is 10 seconds, this would send the event 1 second after the last action, or after 10 seconds, whichever comes first.

koesie10 (approved):

extensions/ql-vscode/src/view/common/telemetry.ts
robertbrignull: > So, the idea here is to ensure that if this effect gets invoked multiple times within the debounce period, earlier data will be overwritten with newer data?

Yes, the point here is that if the state value changes multiple times within the debounce period then it will not send a telemetry event, and it'll instead send it later once the value hasn't changed for the required period. The idea is that if a user interacts with a feature like the search bar that changes state on every keypress, we only want to send one (or at most a few) telemetry events instead of one for every keypress. Effectively we want it to count as one use of the feature rather than many uses.

We're using `useMemo` because we need the `sendTelemetryFunc` function to stay the same function object, so that we can keep track of the `NodeJS.Timeout` variable.

> It looks like there may be a problem. If `debounceTimeoutMillis` is set to (say) 100ms, and this is invoked every 99ms (forever), then it looks like the event will never be sent. Am I reading that right?

You're right that if the event keeps happening within the debounce period then the telemetry will never be sent. I think that's unlikely since that requires the user to make an infinite number of actions. These aren't responding to automated events, so it's unlikely that it'll happen like this where there's never a long enough gap between events. We're also only dealing with user-interaction telemetry so if it doesn't send it's also not that bad. In general I'm not concerned about it too much.

> I _think_ what we want to do is avoid clearing the `timer`, and just make sure that each time we call this effect, then we update the `telemetryAction` to the latest action. This way, the event will always be sent at most every `debounceTimeoutMillis` and always with the latest `telemetryAction`.

I'm afraid I'm not quite following the suggestion here. Could you show an example of what you mean, or explain further?

extensions/ql-vscode/src/view/variant-analysis/RepoRow.tsx
robertbrignull: Good point. This could definitely be simpler 👍

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysis.tsx
robertbrignull: Good point. This one probably should have a debounce too, as the user could click a lot of checkboxes in quick succession.

extensions/ql-vscode/src/view/common/telemetry.ts
robertbrignull: Nice. That is cleaner.

extensions/ql-vscode/src/view/common/telemetry.ts
robertbrignull: > I'd expect us to be interested in how many distinct searches a user does, rather than that they took 5 seconds to construct their search.

I agree we ideally only want to send one telemetry event per search, regardless of if the user takes a while to type it out. That's the reason why it's got the debounce time of 1s which is quite long. I actually considered making it even longer, but settled on 1s. Sending a few events is not the end of the world, so long as we don't send one for every single keypress.

> Alternatively, we could have something like a maxWait parameter

That does make sense to me as an extra addition, but I'd wait to see if we need it first, so I'm hesitant to add it right now. It's certainly something we could look at in the future though if we think the telemetry we're getting isn't correct or good enough.

aeisenberg: Thanks for the explanation and discussion. I think this all makes sense to me right now.

aeisenberg:
I just saw this comment:

> It was mentioned earlier today maybe we don't have to send an extension message and we could emit telemetry events from within the webview. I don't know if that's possible but I'm happy to explore it in a followup PR. It won't change the bulk of this work as sending the extension messages isn't the hard bit.

It would generally be possible to do this, but we'd need to change the CSP when we construct the webview. In the past, we've been careful to ensure that the webview only communicates with the extension. Maybe it's something worth reconsidering. However, I think you'd need to use a different library to send the telemetry events. This would mean that the structure of webview events would be different from extension events. This wouldn't be terrible, but would require some extra work in application insights

aeisenberg (approved):

robertbrignull:
> I just saw this comment:
> 
> > It was mentioned earlier today maybe we don't have to send an extension message and we could emit telemetry events from within the webview. I don't know if that's possible but I'm happy to explore it in a followup PR. It won't change the bulk of this work as sending the extension messages isn't the hard bit.
> 
> It would generally be possible to do this, but we'd need to change the CSP when we construct the webview. In the past, we've been careful to ensure that the webview only communicates with the extension. Maybe it's something worth reconsidering. However, I think you'd need to use a different library to send the telemetry events. This would mean that the structure of webview events would be different from extension events. This wouldn't be terrible, but would require some extra work in application insights.

Thanks for explaining. That's really helpful! Sounds to me like maybe too much effort and downsides of having two implementations, unless sending the extension messages turns out to be a big problem. Right now I don't think the volume of extension messages or the implementation of the message is too bad.

aeisenberg:
Yes, and another benefit of using a single mechanism to send telemetry is that the package we are using right now batches multiple telemetry events into a single request.

PR #1963, opened 2023-01-13
Use Webpack in production mode for releases

koesie10:
This will set the `mode` of Webpack to `production` for release builds. It will also stop inlining the sourcemap and instead produce a separate file which is excluded by `.vscodeignore`.

In terms of the bundled extension, this will add 1 file (`out/webview.js.LICENSE.txt`). It decreases the size of the VSIX file from 4.28MB to 1.77MB.

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,6 +59,12 @@ jobs:
           name: vscode-codeql-extension
           path: artifacts
 
+      - name: Upload source maps
+        uses: actions/upload-artifact@v3
+        with:
+          name: vscode-codeql-sourcemaps
+          path: dist/vscode-codeql/out/*.map
+
         # TODO Run tests, or check that a test run on the same branch succeeded.
 
       - name: Create release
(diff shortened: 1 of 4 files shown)


charisk (approved):
The change LGTM (although I've not tested at all).

I'd be keen to hear from previous folks (e.g. @aeisenberg) why we were using dev mode. Perhaps there was a reason, not just tech debt?

aeisenberg:
The main reason was for debugging. When a user has an error, they can send their stack traces to us and we can more easily debug. We found it quite helpful in the past, but of course it adds to the size of the extension. It's up to you all what you want to do.

koesie10:
> The main reason was for debugging. When a user has an error, they can send their stack traces to us and we can more easily debug. We found it quite helpful in the past, but of course it adds to the size of the extension. It's up to you all what you want to do.

Thanks for adding some context! I've pushed a commit that will upload the sourcemaps on release builds. This should still allow us to reconstruct the stack trace, even if we're building in production mode.

charisk (approved):

PR #1964, opened 2023-01-13
Use stdin for supplying auth to CodeQL

koesie10:
This will supply the GitHub access token to certain CodeQL CLI commands such that private packages can be resolved. It will only do so if the user has an existing auth session. If they don't, they will now get a prompt to login. However, this will only happen for commands which actually use authentication, which is limited to packaging commands.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -1,3 +1,4 @@
+import { EOL } from "os";
 import { spawn } from "child-process-promise";
 import * as child_process from "child_process";
 import { readFile } from "fs-extra";
@@ -26,6 +27,7 @@ import { Logger, ProgressReporter } from "./common";
 import { CompilationMessage } from "./pure/legacy-messages";
 import { sarifParser } from "./sarif-parser";
 import { dbSchemeToLanguage, walkDirectory } from "./helpers";
+import { Credentials } from "./authentication";
 
 /**
  * The version of the SARIF format that we are using.
@@ -155,6 +157,10 @@ interface BqrsDecodeOptions {
   /** The entity names to retrieve from the bqrs file. Default is url, string */
   entities?: string[];
 }
+
+export type OnLineCallback = (
+  line: string,
+) => Promise<string | undefined> | string | undefined;
 
 /**
  * This class manages a cli server started by `codeql execute cli-server` to
@@ -304,6 +310,7 @@ export class CodeQLCliServer implements Disposable {
     command: string[],
     commandArgs: string[],
     description: string,
+    onLine?: OnLineCallback,
   ): Promise<string> {
     const stderrBuffers: Buffer[] = [];
     if (this.commandInProcess) {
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -559,6 +589,44 @@ export class CodeQLCliServer implements Disposable {
...
+    commandArgs: string[],
+    description: string,
+    addFormat = true,
+    progressReporter?: ProgressReporter,
+  ): Promise<OutputType> {
+    const credentials = await Credentials.initialize();
+
+    const accessToken = await credentials.getExistingAccessToken();
+
+    const extraArgs = accessToken ? ["--github-auth-stdin"] : [];
```
aeisenberg, line 612: Why get the accessToken here at all if it's not going to be used below? Should it be the case that if we have an accessToken already, then we use it below and don't prompt?

aeisenberg: Though, I wonder if this will be a poor user experience. Most users will only be downloading public packages. They shouldn't need to log in for that. 

Do you think it would be better if:

1. Try to get an existing token.
2. If found, use that.
3. If not found, use no token (as before)
4. If we get an error because the package is private (and no token was provided) prompt the user to log in.
5. If we get an error because the package is private (and a token was provided) clearly explain that this token doesn't have proper permissions so the user can re-authenticate with a different token.

extensions/ql-vscode/src/cli.ts
```diff
@@ -559,6 +589,44 @@ export class CodeQLCliServer implements Disposable {
...
+
+    return this.runJsonCodeQlCliCommand(
+      command,
+      [...extraArgs, ...commandArgs],
+      description,
+      addFormat,
+      progressReporter,
+      async (line) => {
+        if (line.startsWith("Enter value for --github-auth-stdin")) {
+          return credentials.getAccessToken();
```
aeisenberg, line 622: Is this what we want to do?

```suggestion
          return accessToken;
```

robertbrignull (commented):
Code looks ok to me. Code doing this kind of job of starting other processes is always quite complicated.

I'll let @aeisenberg complete his review and approve, but I'm happy whenever he's happy.

extensions/ql-vscode/src/cli.ts
aeisenberg: Ok. Thanks for the explanation. So, it looks like there are two race conditions that aren't handled:

1. User logs in after invoking this command, but before the command is run. In this case, the extension will assume that the user is not logged in (this is fine, I think).
2. User logs out after invoking this command, but before the command is run. In this case, the extension will assume the user is still logged in, try to get the credentials, but `credentials.getAccessToken()` returns undefined. In this case, will the command hang? Or will it fail since the value `undefined` is passed to stdin as the string "undefined"?

This race condition is probably rare enough that it's not worth spending too much effort on and when it does happen, it would probably be obvious to the user what happened.

PR #1970, opened 2023-01-16
Save scroll position before opening code paths

charisk:
This is so that the user doesn't lose their place in the results view.

https://user-images.githubusercontent.com/311693/212700041-573ea61d-582b-4a76-b07d-3237e957f74c.mov

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/view/common/CodePaths/CodePaths.tsx
+++ b/extensions/ql-vscode/src/view/common/CodePaths/CodePaths.tsx
@@ -31,13 +31,24 @@ export const CodePaths = ({
 }: CodePathsProps) => {
   const [isOpen, setIsOpen] = useState(false);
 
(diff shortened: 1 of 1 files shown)


robertbrignull:
Interesting! I thought this would be harder than this. This does seem to work, though I am getting other slightly odd effects.

I'm now getting behaviour where it scrolls to the top of the page when I expand a repo row. Do you also get this? I don't see it in your screen recording but I wonder if that's because you have one repo with lots of results rather than lots of repos, so it is already at the top when you expand the row.

Could you also explain why we're calling `window.scrollTo` when the paths overlay is opened, rather than when it's closed? Maybe it doesn't matter, but that would seem more instinctive to me so I'd like to understand.

charisk:
Apologies for leaving this open. I've not had time to look at it (I couldn't repro on my machine). Will mark it as draft for now.

PR #1973, opened 2023-01-16
Modifying the `bump-cli` workflow to append/replace version from `supported_cli_versions.json`

tjgurwara99:
During the development of `bump_cli.yml` workflow, we understood the requirement that we want to be able to append/prepend/remove versions from versions from `supported_cli_versions.json` file. Initially it was decided that we should modify the `bump_cli` workflow, however, the complication I had when doing that was the logic to handle the cases with inputs and without inputs becomes quite complicated and therefore I decided against modifying the `bump_cli` workflow. If however, the maintainers really want to just keep one workflow then I can modify this branch and go from there 😄. Ref [issue](https://github.com/github/code-scanning/issues/8597).

EDIT: Since the requirements are simplified to replacing and prepending version, we have combined the `bump-cli` workflow and `version-change` workflow into one workflow.


.github/workflows/version-change.yml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          fetch-depth: 1
+      - name: Remove Version
+        if: ${{ inputs.option == 'remove' }}
+        run: |
+          cat supported_cli_versions.json | jq 'del(.[] | select(. == "${{ inputs.version }}"))' > supported_cli_versions_temp.json
+      - name: Append Version
+        if: ${{ inputs.option == 'append' }}
+        run: |
+          cat supported_cli_versions.json | jq '. += ["${{ inputs.version }}"]' > supported_cli_versions_temp.json
```
aeisenberg, line 39: Are there situations where a maintainer will want to remove or append a version? In the rare case of there was a mistake, it's fine to manually edit the file. 

I'm suggesting this since I think this workflow file can be made simpler by removing the input and always prepending.

aeisenberg (commented):
And I think it can be made even simpler for maintainers to have only a single workflow with an option input. The two options could be "replace" and "prepend". You could add a second job to the bump-cli workflow that runs only when "prepend" is selected. This job could do something similar to what you already wrote.

.github/workflows/version-change.yml
tjgurwara99: Cool, I can remove the other options - I thought it might be easier to just add or remove versions whenever we needed 😄

.github/workflows/bump-cli.yml
```diff
@@ -23,10 +36,18 @@ jobs:
...
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           scripts/replace-cli-version.sh
+      - name: Prepend another version
+        if: ${{ inputs.option == 'prepend' }}
+        run: |
+          cat supported_cli_versions.json | jq '. |= ["${{ inputs.version }}"] + .' > supported_cli_versions_temp.json
+          mv supported_cli_versions_temp.json supported_cli_versions.json
+          echo "LATEST_VERSION=${{ inputs.version }}" >> $GITHUB_ENV
+          echo "PREVIOUS_VERSION=`jq -r '.[1]' supported_cli_versions.json`" >> $GITHUB_ENV
```
tjgurwara99, line 50: Let me know if I should move this into a separate script similar to the `scripts/replace-cli-versions.sh` 😄

.github/workflows/bump-cli.yml
```diff
@@ -1,6 +1,19 @@
...
+      option:
+        description: "Option"
+        required: true
+        default: 'replace'
+        type: choice
+        options:
+          - prepend
+          - replace
+      version:
+        description: "Version"
```
aeisenberg, line 14: I think more description is required here. How about:

```suggestion
        description: |
            The version to prepend to the supported versions file. This should be in the form: `vA.B.C`.
```

aeisenberg (commented):
Looks good. I tried it out here and behaved as I expected. Just a suggestion to update the description and then looks good.

https://github.com/github/vscode-codeql/pull/1979

aeisenberg (approved):

PR #1976, opened 2023-01-17
Emit telemetry for user interactions in the local results view

robertbrignull:
This PR adds telemetry for all UI interactions in the local results view.

We use the same telemetry message type that was used for the variant analysis results view, but we have to use a slightly different technique for sending the message when state changes. The local results view is not a modern react component and it uses `setState` rather than `useState`, which means we can't use `useEffect` here and instead have to add a call to the telemetry every time we call `setState`.

I've managed to trigger each of these telemetry events apart from `local-results-show-raw-results`. I can't quite work out the series of conditions that need to happen to be able to trigger this code. But I think if it ever does trigger then telemetry is reasonable.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/interface.ts
+++ b/extensions/ql-vscode/src/interface.ts
@@ -256,6 +256,7 @@ export class ResultsView extends AbstractWebview<
         }
         case "changeSort":
           await this.changeRawSortState(msg.resultSetName, msg.sortState);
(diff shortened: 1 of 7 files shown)


aeisenberg (approved):
Nice! 

The local results view should probably get an overhaul, but it's great that you were able to add telemetry without any major changes.

PR #1985, opened 2023-01-18
Upgrade telemetry library

robertbrignull:
Upgrades our telemetry library to a newer version. The library has also changed NPM package. There's a deprecation notice on the original package: https://www.npmjs.com/package/vscode-extension-telemetry

The benefit we get from this is builtin support for removing fields, so we can replace our existing hacky implementation. See https://github.com/microsoft/vscode-extension-telemetry/blob/v0.6.2/src/node/telemetryReporter.ts#L23 for the implementation which shows that it removes the fields from both the places we are interested in.

The downside is that the app insights client is no longer accessible on the reporter object, so we can't do our hacky `addTelemetryProcessor` to log the telemetry. I've been searching through the telemetry and app insights source code and I can't see a way to add a hook like this that has access to the data being sent. The best option I could come up with is to log the data before it's passed into app insights. This will change the data that gets logged to the extension log because it won't include all the data collected by app insights (e.g. machine information). This means it's less representative of the data that actually sent and so may be less useful to users who want to verify the data being sent. As far as I can tell this doesn't break our agreements in https://codeql.github.com/docs/codeql-for-visual-studio-code/about-telemetry-in-codeql-for-visual-studio-code/ but I'm not 100% sure on that.

Overall, I'm not sure if upgrading this dependency is a good idea or worth it. I'd be keen to get other opinions.

## Checklist


aeisenberg (commented):
Thanks for taking a look at this. It's unfortunate about the logging. My recollection is that the new version does a better job at sanitizing exception stack traces, making it possible for us to start collecting them. If that's the case, then this PR is a valuable change since it would unlock this feature. Otherwise, I don't think the benefits outweigh the costs.

robertbrignull:
Looking through https://github.com/microsoft/vscode-extension-telemetry/releases I don't see anything jumping out about improving sanitisation of error stack traces. The closest is about the addition of `replacementOptions`, which I now realise you implemented 😮 ✨ 

It's possible we get a benefit from using a newer version of https://github.com/microsoft/ApplicationInsights-node.js, but I don't think we're actually updating the version of that here. We still transitively depend on the same version.

aeisenberg:
Oh...apologies. I might be mistaken about stack traces with the new library version. It looks like there would need to be a new telemetry processor added to the applicationInsights instance in the IDE that could filter or anonymize stack traces. This is not even possible using the new version, making it even harder to use.

robertbrignull:
> It looks like there would need to be a new telemetry processor added to the applicationInsights instance in the IDE that could filter or anonymize stack traces.

This was my thought about how to implement it too. We may have to do something ourselves then to implement this.

robertbrignull:
I'm going to close this PR. We can revisit in the future if needed.

There is perhaps a downside that we are forced to upgrade in the future and we have to solve these same issues but under time pressure. That's the only downside I can see of not upgrading now.

PR #1989, opened 2023-01-18
Add credentials to App

robertbrignull:
An attempt at adding the `Credentials` class to the `App` so that we can better control how it is initialised in tests. This involves passing the `App` object to more classes, like the variant analysis manager, but this is expected as we convert more of the codebase to use the `App`. Surprisingly this PR actually turns out to be a net-reduction in lines of code 😄 

Before properly merging I can add some more documentation to the `Credentials` class, and there will likely be some test failures that I missed locally. My aim in making this draft PR is to get eyes from the rest of the team on the general approach.

## Checklist


extensions/ql-vscode/test/factories/authentication.ts
```diff
@@ -0,0 +1,38 @@
...
+    requestSpy ??
+    jest.fn(async () => {
+      throw new Error("Tried to make HTTP request but no stub was provided");
+    });
+  return makeTestOctokit({ request } as any);
+}
+
+export function testCredentialsWithUnauthenticatedOctokit(): Credentials {
+  return makeTestOctokit(new Octokit.Octokit({ retry }));
+}
```
elenatanasoiu, line 34: Can we get rid of this one? It's used in one place and the test is using a scenario so responses should be mocked.

extensions/ql-vscode/test/vscode-tests/no-workspace/remote-queries/gh-api/gh-actions-api-client.test.ts
```diff
@@ -95,7 +93,7 @@ describe("gh-actions-api-client real responses", () => {
       return;
```
elenatanasoiu, line 93: Branch here to check if this can go: https://github.com/github/vscode-codeql/pull/2000

extensions/ql-vscode/test/factories/authentication.ts
robertbrignull: I've combined the two methods together to just `testCredentialsWithRealOctokit` which now takes an _optional_ `token` parameter. I think that's a little cleaner than it was before.

robertbrignull:
@elenatanasoiu, this should be ready for review now. All I've done since we looked through it together is add documentation and adjust the test credentials factories. The rest of the test failures were just flakes that went away with retries.

elenatanasoiu (approved):

extensions/ql-vscode/src/authentication.ts
```diff
@@ -13,41 +14,12 @@ const SCOPES = ["repo", "gist", "read:packages"];
 /**
  * Handles authentication to GitHub, using the VS Code [authentication API](https://code.visualstudio.com/api/references/vscode-api#authentication).
  */
-export class Credentials {
+export class VSCodeCredentials implements Credentials {
```
koesie10, line 17: This file should probably live in `common/vscode/credentials.ts` since it's now part of the app and doesn't need to be accessed itself.

extensions/ql-vscode/src/common/authentication.ts
```diff
@@ -0,0 +1,34 @@
+import * as Octokit from "@octokit/rest";
+
+/**
+ * An interface providing methods for obtaining access tokens
+ * or an octokit instance for making HTTP requests.
+ */
+export interface Credentials {
```
koesie10, line 7: Would it make more sense for this file to be named `credentials.ts` rather than `authentication.ts`? I find that having a class named `Credentials` in a file named `authentication` makes it harder to find, especially if you're new to the codebase.

extensions/ql-vscode/test/factories/authentication.ts
```diff
@@ -0,0 +1,42 @@
+import { retry } from "@octokit/plugin-retry";
```
koesie10, line 1: Same comment for this file as the other ones: should this be named `credentials.ts`?

koesie10 (approved):
Thanks for making these changes! I have some very minor suggestions about the naming of some files, but this could also be done in a follow-up PR.

robertbrignull:
@koesie10, to answer all of your review comments. I'm planning to do another PR after this to deal with renaming classes and files. I agree with the suggestion to rename `authentication.ts` to match the name of the class, and it's also been suggested that we rename the `Credentials` class to something like `CredentialsProvider`. I also agree putting the vscode implementation into `common/vscode/` sounds good, so I may as well do that in the followup naming PR.

I hope all of this is ok with people.

PR #1992, opened 2023-01-20
Ensure cli using latest version from the supported_cli_versions.json

tjgurwara99:
closes [this issue](https://github.com/github/code-scanning/issues/8598)

The PR changes the bump-cli workflow to only edit the `supported_cli_versions.json` file and remove the need to edit `ensureCli.ts` which used to include a hardcoded string of the latest version. Now however, it takes the latest version from the `supported_cli_versions.json` (the first version in the array in `supported_cli_versions.json` is always going to be the latest version of the supported cli by design) removing the need to update the `ensureCli.ts` file 😄


extensions/ql-vscode/test/vscode-tests/ensureCli.ts
```diff
@@ -38,7 +39,7 @@ const _10MB = _1MB * 10;
 
 // CLI version to test. Hard code the latest as default. And be sure
 // to update the env if it is not otherwise set.
-const CLI_VERSION = process.env.CLI_VERSION || "v2.11.6";
+const CLI_VERSION = process.env.CLI_VERSION || data[0];
```
shati-patel, line 42: Minor: we could make this variable name more descriptive 😄 

```suggestion
const CLI_VERSION = process.env.CLI_VERSION || supportedCliVersions[0];
```

shati-patel (commented):
The `main` workflow also need an update: 
https://github.com/github/vscode-codeql/blob/b0168aa04a6763eb41897c6989933aef7e9a51ea/.github/workflows/main.yml#L194

Otherwise, LGTM with just some minor comments! (I tried running the CLI tests locally too, and all good 🎉 )

aeisenberg (approved):
LGTM

PR #1997, opened 2023-01-23
Make use of asError instead of reimplementing

robertbrignull:
This PR updates a few places that were doing their own `e instanceOf Error` check to instead use `asError` or `getErrorMessage` as appropriate.

Also updates the types of the helper methods to be more restrictive, as I believe using `unknown` instead of `any` where possible is generally best practice. I haven't gone any further than this with annotating things as `unknown` to avoid going down too much of a rabbit hole.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -73,7 +73,7 @@ import {
   showInformationMessageWithAction,
   tmpDir,
 } from "./helpers";
-import { assertNever, getErrorMessage } from "./pure/helpers-pure";
+import { asError, assertNever, getErrorMessage } from "./pure/helpers-pure";
 import { spawnIdeServer } from "./ide-server";
 import { ResultsView } from "./interface";
 import { WebviewReveal } from "./interface-utils";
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/extension.ts
```diff
@@ -771,9 +771,7 @@ async function activateWithInstalledDistribution(
         // Note we must update the query history view after showing results as the
         // display and sorting might depend on the number of results
       } catch (e) {
-        const err = asError(e);
-        err.message = `Error running query: ${err.message}`;
-        item.failureReason = err.message;
+        item.failureReason = `Error running query: ${getErrorMessage(e)}`;
```
koesie10, line 774: This changes behaviour for the thrown error since the error message now doesn't include `Error running query:`. The message from the error is probably shown to the user directly since this method is called by a command, so this might change user-facing behaviour as well (depending on if the thrown error is an instance of `Error`).

koesie10 (commented):
Thanks for cleaning this up!

extensions/ql-vscode/src/extension.ts
robertbrignull: Oh your right because `asError` can just return `e` instead of a new error, so actually this does edit the existing `e` which we later throw 🤯 

I'll revert this change and leave this file alone.

PR #2002, opened 2023-01-24
Move query history files to new folders

norascheuch:
## Checklist


norascheuch:
The failing test is a failed CodeQL check in a file where I had do adjust an import. I suggest to create an issue to check this separately.

extensions/ql-vscode/src/extension.ts
```diff
@@ -83,7 +83,7 @@ import {
   ProgressReporter,
   queryServerLogger,
 } from "./common";
-import { QueryHistoryManager } from "./query-history";
+import { QueryHistoryManager } from "./query-history/query-history";
```
koesie10, line 86: This might be a good candidate for creating an `index.ts` file in the `query-history` directory so we can just import from `./query-history` instead of `./query-history/query-history`, but that's something better done in a follow-up PR if we want to do this.

koesie10 (approved):

extensions/ql-vscode/src/extension.ts
norascheuch: 👍

PR #1998, opened 2023-01-23
Log more places in jest setup to help with debugging

elenatanasoiu:
Mobbed on debugging flakey tests with @norascheuch & @starcke.

We know our flakey test [starts failing after](https://github.com/github/vscode-codeql/actions/runs/3941406658/jobs/6743776714):

```
Downloading test database to /home/runner/work/vscode-codeql/vscode-codeql/extensions/ql-vscode/build/tests/db.zip
```

So let's add more logging and see if we can see which step is failing.

Since we don't run CI on branches, I've opened this as a draft PR.

PR #2003, opened 2023-01-24
Run integration tests sequentially

elenatanasoiu:
Currently we're running our jest tests in multiple parallel threads.

We can see that CI is experiencing a lot of flakiness where a test will randomly error with the following message:

```
Unable to write into user settings because the content of the file is newer.
```

These are failing only on CI so we believe at least some of the flakiness is caused by running our tests in a multi-threaded environment in CI.

So we experimented with running our jest tests in sequence via the `--runInBand` option: github/vscode-codeql#1998.

While this meant that our tests would take longer to run (15 mins vs. 4-6mins), most of them now passed on the first CI run:

- first run (12m48s): https://github.com/github/vscode-codeql/actions/runs/3997178977/jobs/6858164662
- another first run (11m58s): https://github.com/github/vscode-codeql/actions/runs/3997164504/jobs/6858136512

You can see more first-run examples on the commits in this PR: github/vscode-codeql#1998.

For comparison, the first CI run for tests running in parallel sees many more failures:
- first attempt (23m8s) - 10 failed jobs: https://github.com/github/vscode-codeql/actions/runs/3997630502/attempts/1
- second attempt (18m58s) - one failed job: https://github.com/github/vscode-codeql/actions/runs/3997630502/attempts/2
- third attempt - green (6m42s): https://github.com/github/vscode-codeql/actions/runs/3997630502/attempts/2

As you can probably tell, there is a notable reduction in flakiness which means we're able to re-run less jobs to get to a green state.

This can also reveal other sources of flakiness that we've perhaps been ignoring in the past.

So our proposal is for now to run tests in sequence and think about running them in parallel once we understand more about the way tests interact in a multi-threaded environment.

## Checklist


koesie10:
Theoretically, we should already be running all tests sequentially because [the runner has the `isSerial` option set](https://github.com/adalinesimonian/jest-runner-vscode/blob/0c98dc12adac536a200e836b77e186b8c2ba8326/packages/jest-runner-vscode/src/runner.ts#L17) and this should do [exactly the same thing as `--runInBand`](https://github.com/facebook/jest/blob/9fa2a463a0d5b1a7763ef4a28004db082b8bcf6e/packages/jest-core/src/TestScheduler.ts#L245). So, I think the actual change this is making [is the max worker count](https://github.com/facebook/jest/blob/9fa2a463a0d5b1a7763ef4a28004db082b8bcf6e/packages/jest-config/src/getMaxWorkers.ts#L27-L28). I'm not sure how the worker count works when `isSerial` is set, but that means both `--runInBand` and `-w 1` have the same effect. `--runInBand` is definitely the clearer option, so this change makes sense to me.

For enabling parallelization, we could do something like [this Gist shows](https://gist.github.com/imhoffd/b4ca0a94c2496d81303ebb00063a863d), essentially running a subset of tests in a matrix.

elenatanasoiu:
> Theoretically, we should already be running all tests sequentially because [the runner has the isSerial option set](https://github.com/adalinesimonian/jest-runner-vscode/blob/0c98dc12adac536a200e836b77e186b8c2ba8326/packages/jest-runner-vscode/src/runner.ts#L17) a

I see! That explains why trying to use `--maxWorkers` as an option threw an error about not being able to use `--runInBand` AND `--maxWorkers` at the same time. 

> So, I think the actual change this is making [is the max worker count](https://github.com/facebook/jest/blob/9fa2a463a0d5b1a7763ef4a28004db082b8bcf6e/packages/jest-config/src/getMaxWorkers.ts#L27-L28). I'm not sure how the worker count works when isSerial. 

[This](https://github.com/facebook/jest/blob/9fa2a463a0d5b1a7763ef4a28004db082b8bcf6e/packages/jest-runner/src/index.ts#L50) seems to indicate the `.serial` option triggers running tests in band so it's probably still setting workers to 1. 

> `--runInBand` is definitely the clearer option, so this change makes sense to me.

Hmmm, yeah but it doesn't get us closer to solving our flakiness in that case. Shame. 

I think adding this option would be just as unclear because we'll think removing it will run our tests in parallel.

elenatanasoiu:
Thanks for the review @koesie10 !

PR #2010, opened 2023-01-25
Introduce ShowAndLogOptions

robertbrignull:
The idea here is to standardise how our `showAndLogXMessage` methods behave. Currently they're each a little different. This PR should mean that the arguments for all the methods stay in sync, and also the default values are only defined in one place instead of multiple places.

I'm also possibly about to introduce another variant of `showAndLogErrorMessage` so I'll want to be able to easily make that consistent too.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/helpers.ts
+++ b/extensions/ql-vscode/src/helpers.ts
@@ -19,7 +19,7 @@ import {
 } from "vscode";
 import { CodeQLCliServer, QlpacksInfo } from "./cli";
 import { UserCancellationException } from "./commandRunner";
-import { extLogger } from "./common";
+import { extLogger, OutputChannelLogger } from "./common";
 import { QueryMetadata } from "./pure/interface-types";
 
 // Shared temporary folder for the extension.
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/helpers.ts
```diff
@@ -43,6 +43,12 @@ export const tmpDirDisposal = {
   },
 };
 
+interface ShowAndLogOptions {
+  outputLogger?: OutputChannelLogger;
+  items?: string[];
+  fullMessage?: string;
```
koesie10, line 49: Can we also add documentation for the properties here? We could even do this instead of in the method documentation, that avoids having to repeat the documentation in three different methods.

koesie10 (approved):

extensions/ql-vscode/src/helpers.ts
robertbrignull: Good idea. I'll move the documentation to the fields here instead of on the methods. I can still include a link to the type from the method documentation.

PR #2011, opened 2023-01-25
Enforce that promise rejection values must have unknown type

robertbrignull:
This PR adds the following eslint rule: https://github.com/cartant/eslint-plugin-etc/blob/main/docs/rules/no-implicit-any-catch.md

This enforces that whenever you use `catch` for a promise rejection value, the value is typed as `unknown`. This is safer than the default of `any` and will avoid us accidentally passing incorrectly typed values around.

There already exists https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables which does the same thing for `catch` statements for exceptions. We already have this enabled because we are using `strict` mode.

Unfortunately I had to add a whole new eslint plugin for this because I couldn't find another rule that implemented this. If there's a better rule to add I'm all for using that instead.

## Checklist

### DIFF
--- a/extensions/ql-vscode/.eslintrc.js
+++ b/extensions/ql-vscode/.eslintrc.js
@@ -7,7 +7,8 @@ module.exports = {
   },
   plugins: [
     "github",
-    "@typescript-eslint"
+    "@typescript-eslint",
+    "etc"
   ],
   env: {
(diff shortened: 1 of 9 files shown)


koesie10 (approved):
Thanks for adding this!

PR #2012, opened 2023-01-25
Use JSON schemas to validate query history items

elenatanasoiu:
Introduces:
- a separate set of types for the data we want to read from/write to disk (in `data-serialization/source-schema-types`)
- a `SchemaGenerator` which takes those types and converts them into JSON schemas (in `data-serialization/generated-schemas`). I've tried to make it configurable so that we can re-use it in other parts of the extension where we read/write data. That may or may not be overkill.
- a `JsonValidator` which reads the JSON schemas and checks whether provided data is valid.

We then use the JsonValidator in our `query-serialization.ts` code to validate remote queries and variant analyses when we read/write from disk. This decouples our code from our existing types as we're now using the types defined in `data-serialization/source-schema-types`.

We could expand this further by:
1. using this validator for local query history items
2. re-using this pattern in other parts of the code where we read/write data (e.g. database panel config, query results)
3. think about versioning

While these would be nice to have, the immediate issue we want to solve is decoupling our types from the data we read/write for query history so that we're able to make changes to our types without making the data we've already written to disk unreadable. See internal issue for details. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-serialization/generated-schemas/RemoteQueryHistoryItem.json
+++ b/extensions/ql-vscode/src/data-serialization/generated-schemas/RemoteQueryHistoryItem.json
@@ -1,0 +1,102 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$ref": "#/definitions/RemoteQueryHistoryItem",
+  "definitions": {
+    "RemoteQueryHistoryItem": {
+      "type": "object",
+      "properties": {
+        "t": {
+          "type": "string",
+          "const": "remote"
+        },
+        "failureReason": {
+          "type": "string"
+        },
+        "resultCount": {
+          "type": "number"
+        },
+        "status": {
+          "$ref": "#/definitions/QueryStatus"
+        },
+        "completed": {
+          "type": "boolean"
+        },
+        "queryId": {
+          "type": "string"
+        },
+        "remoteQuery": {
+          "type": "object",
+          "properties": {
+            "queryName": {
+              "type": "string"
+            },
+            "queryFilePath": {
+              "type": "string"
+            },
+            "queryText": {
+              "type": "string"
+            },
+            "language": {
+              "type": "string"
+            },
+            "controllerRepository": {
+              "type": "object",
+              "properties": {
+                "owner": {
+                  "type": "string"
+                },
+                "name": {
+                  "type": "string"
+                }
+              },
+              "required": [
+                "owner",
+                "name"
+              ],
+              "additionalProperties": false
+            },
+            "executionStartTime": {
+              "type": "number"
+            },
+            "actionsWorkflowRunId": {
+              "type": "number"
+            },
+            "repositoryCount": {
+              "type": "number"
+            }
+          },
+          "required": [
+            "queryName",
+            "queryFilePath",
+            "queryText",
+            "language",
+            "controllerRepository",
+            "executionStartTime",
+            "actionsWorkflowRunId",
+            "repositoryCount"
+          ],
+          "additionalProperties": false
+        },
+        "userSpecifiedLabel": {
+          "type": "string"
+        }
+      },
+      "required": [
+        "t",
+        "status",
+        "completed",
+        "queryId",
+        "remoteQuery"
+      ],
+      "additionalProperties": false
+    },
+    "QueryStatus": {
+      "type": "string",
+      "enum": [
+        "InProgress",
+        "Completed",
+        "Failed"
+      ]
+    }
+  }
+}
--- a/extensions/ql-vscode/src/data-serialization/json-validator.ts
+++ b/extensions/ql-vscode/src/data-serialization/json-validator.ts
@@ -1,0 +1,50 @@
+import Ajv from "ajv";
+import { JsonSchemaTypes } from "./json-schema-types";
+import * as fs from "fs";
+import { JSONSchema7 as JSONSchema } from "json-schema";
+
+export class JsonValidator {
+  ajv: Ajv;
+  schemaTypes: string[];
+  schemas: { [key: string]: JSONSchema };
+
+  constructor(schemaTypes?: string[]) {
+    this.ajv = new Ajv();
+    this.schemaTypes = schemaTypes || JsonSchemaTypes;
+    this.schemas = this.loadSchemas(this.schemaTypes);
+  }
+
+  public validate(object: unknown, typeName: string) {
+    const validate = this.createValidatorFunction(typeName);
+
+    const valid = validate(object);
+
+    if (!valid) {
+      throw new Error(
+        `Object does not match the "${typeName}" schema: ${validate.errors}`,
+      );
+    }
+    return object;
+  }
+
+  private createValidatorFunction(typeName: string) {
+    const schema = this.loadSchema(typeName);
+    return this.ajv.compile(schema);
+  }
+
+  private loadSchemas(schemaTypes: string[]): { [key: string]: JSONSchema } {
+    const schemaList: { [key: string]: JSONSchema } = {};
+
+    for (const typeName of schemaTypes) {
+      const schemaFilepath = `src/data-serialization/generated-schemas/${typeName}.json`;
+      schemaList[typeName] = JSON.parse(
+        fs.readFileSync(schemaFilepath, "utf-8"),
+      );
+    }
+    return schemaList;
+  }
+
+  private loadSchema(typeName: string): JSONSchema {
+    return this.schemas[typeName];
+  }
+}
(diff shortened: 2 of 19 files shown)


elenatanasoiu:
Oh dear... I'm unsure if the remaining failures are flakiness or something I caused with upgrading packages. I'll retry a few times.

extensions/ql-vscode/src/data-serialization/generated-schemas/RemoteQueryHistoryItem.json
```diff
@@ -0,0 +1,102 @@
+{
```
robertbrignull, line 1: We could use `.gitattributes` to have files in this directory be marked as generated. GitHub might work it out by itself but we can still give it a hint to make sure it gets it correct.
https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

extensions/ql-vscode/src/data-serialization/json-validator.ts
```diff
@@ -0,0 +1,50 @@
...
+  schemaTypes: string[];
+  schemas: { [key: string]: JSONSchema };
+
+  constructor(schemaTypes?: string[]) {
+    this.ajv = new Ajv();
+    this.schemaTypes = schemaTypes || JsonSchemaTypes;
+    this.schemas = this.loadSchemas(this.schemaTypes);
+  }
+
+  public validate(object: unknown, typeName: string) {
```
robertbrignull, line 17: I think if we try we can give this method a return type that's not `unknown` or `any`.

Take a look at https://github.com/github/codeql-variant-analysis-action/blob/main/src/json-validation.ts for inspiration. By using `SchemaTypes` as a map from keys to types, and then `Schema = keyof SchemaTypes` and `validators: Record<Schema, ValidateFunction>`, we were able to make the return type of `validateObject` be the correct object type. This means that there are no casts or objects typed as `any` outside of the validator function.

I'm also happy to help take a stab at it since it's quite arcane and complicated getting the type system to behave.

robertbrignull (commented):
Generally I like and am in favour of the direction this is going. I've left a few comments on different bits, but the general idea of generating json schema types from the typescript types is one I definitely vibe with quite well.

I realise I may be biased because I implemented https://github.com/github/codeql-variant-analysis-action/blob/60c86debfe88a754b40851a0da7661deeaa878cb/src/json-validation.ts and recommended that to you as a potential option. So we probably want to get eyes and approval from the rest of the team too.

I think that converting more bits of the codebase to use this system shouldn't be too hard.
- `DbConfigValidator` is already doing basically this. I believe the json schema is currently hand-written but maybe it could be generated instead.
- I don't know if `LocalQueryInfo` being a JavaScript class rather than just a TypeScript type is a problem or not. That could be something to investigate. Can we still generate a json schema for it and validate the objects.

extensions/ql-vscode/src/data-serialization/generated-schemas/RemoteQueryHistoryItem.json
elenatanasoiu: Don't we want to see when we add a new field to the schemas?

elenatanasoiu: I guess it would be enough to see the changes to associated types... I'll add it.

robertbrignull: Yeah you should be able to judge the change being made and how compatible it is by looking at the typescript types. They're still in the `data-serialization` directory too.

extensions/ql-vscode/src/data-serialization/json-validator.ts
```diff
@@ -0,0 +1,50 @@
...
+    const validate = this.createValidatorFunction(typeName);
```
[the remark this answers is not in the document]
elenatanasoiu: I've pushed up a new draft PR for this: https://github.com/github/vscode-codeql/pull/2020.

extensions/ql-vscode/src/data-serialization/json-validator.ts
elenatanasoiu: Ah yes, I did see that. I'll write a follow-up ticket to poke at improving this since this PR is getting mahoosive. Thanks for explaining it. 👍

elenatanasoiu:
> DbConfigValidator is already doing basically this. I believe the json schema is currently hand-written but maybe it could be generated instead.

That looks straightforward. We'd need to generate the schema automatically and swap in the JsonValidator.

> I don't know if LocalQueryInfo being a JavaScript class rather than just a TypeScript type is a problem or not. That could be something to investigate. Can we still generate a json schema for it and validate the objects.

ts-json-schema-generator does handle classes, but after taking a crack at moving all the parts for `LocalQueryInfo` I think there might be some edge case we could hit with all the scattered types and functions used to define properties. It's going to take a bit of mechanical work to translate it over into a serializable type, but it's doable.

extensions/ql-vscode/src/data-serialization/json-validator.ts
```diff
@@ -0,0 +1,67 @@
...
+    return this.ajv.compile(schema);
+  }
+
+  private loadSchemas(): { [key: string]: JSONSchema } {
+    const schemaList: { [key: string]: JSONSchema } = {};
+
+    for (const typeName of this.schemaTypes) {
+      const schemaFilepath = `src/data-serialization/generated-schemas/${typeName}.json`;
+      schemaList[typeName] = JSON.parse(
+        fs.readFileSync(schemaFilepath, "utf-8"),
```
aeisenberg, line 45: We probably want to avoid synchronous file reads in user-facing code. Maybe the schemas can be loaded asynchronously in a separate `init()` method, or as needed from `loadSchema`

extensions/ql-vscode/src/data-serialization/schema-generator.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    const generator = tjsg.createGenerator(config);
+    const schema = generator.createSchema(config.type);
+    return JSON.stringify(schema, null, 2);
+  }
+
+  public async writeSchema(typeName: string, schemaContents: string) {
+    const output_path = `${this.outputFilePath}/${typeName}.json`;
+
+    fs.writeFileSync(output_path, schemaContents);
```
aeisenberg, line 46: This function is only called in scripts, right? Then using `Sync` here is fine.

extensions/ql-vscode/src/data-serialization/json-validator.ts
elenatanasoiu: Thanks @aeisenberg! That's a good observation.

There's actually a follow up PR to this where we store the `JsonValidator` globally on the new `App` instance: https://github.com/github/vscode-codeql/pull/2020. 

This means we initialize the `JsonValidator` once in the extension (and perform all the schema file reading upfront) when the extension starts up, then re-use those schemas when we need to perform validation. 

So the file reading should only run once when the app is initialized rather than start reading files when a user performs a read/write action. 

Based on this, I think it might affect your original observation that we're reading files in user-facing code, but I might not be understanding all the sides to this so I'm interested to hear your interpretation.

If I move the file reading into an `init()` method, I'll still need to call it in the same place where I'm building the JsonValidator [when we initialize the app](https://github.com/github/vscode-codeql/pull/2020/commits/5fb787f3b19384c849dbfa818698c3d57d4057f8#diff-d75026577deda4d24fc20ee90b6db54add17aaefd2c2702a0d55a5fe04d6930bR13). 

`loadSchema()` is actually triggered when a user performs an action, so I think we would be performing file reading in user-facing code if we moved the file reading action in that method.

elenatanasoiu: I'm going to go ahead and merge this as-is, but if we discuss this further and decide it's still worth doing, I can open a a follow-up PR for it.

aeisenberg: Object constructors are not allowed to be `async`, so my suggestion was around moving the file reads outside of the constructor and put it into an `init` method so that it can be async. Startup time is already quite slow (especially if you have many databases in your workspace) and something we should probably look at some point. So, we should be careful about adding any more bottlenecks to startup.

Feel free to merge and I will look at your followup.

aeisenberg (approved):
There are some points that will be taken up in followup PRs and everything else looks good to me.

elenatanasoiu:
Thanks Andrew! In the end we had a mob session about this and decided to break this up into smaller PRs, with more focus on decoupling all the data types first before introducing validation. I'll take your advice into account when I get to adding back the JsonValidator.

PR #2019, opened 2023-01-27
Use tagged template for telemetry error redaction

koesie10:
This is an idea I had for the telemetry errors. Instead of using fixed strings, this will use [tagged templates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates) to redact any dynamic information in error messages. For example, this allows us to call the `showAndLogExceptionWithTelemetry` method as follows:

```ts
void showAndLogExceptionWithTelemetry(
  asError(e),
  redactableErrorMessage`Failed to choose and set database: ${getErrorMessage(
    e,
  )}`,
);
```

The `redactableErrorMessage` function creates a `RedactableErrorMessage` object which has methods for a `fullMessage` or a `redactedMessage`. This creates messages like:

```ts
console.log(message.fullMessage);
// Failed to choose and set database: Cannot read properties of undefined (reading 'chooseAndSetDatabase')

console.log(message.redactedMessage)
// Failed to choose and set database: [REDACTED]
```

We can safely send the `redactedMessage` to app insights since it only contains static strings. It's not possible to add dynamic strings without deliberate workarounds. There's even support for nested `redactableErrorMessage`, but that still doesn't allow dynamic data:

```ts
const nestedMessage =
  e instanceof TypeError
    ? redactableErrorMessage`This is a type error: ${getErrorMessage(e)}`
    : redactableErrorMessage`This is not a type error: ${getErrorMessage(
      e,
    )}`;

const message = redactableErrorMessage`Failed to choose and set database: ${nestedMessage}`;

console.log(message.fullMessage);
// Failed to choose and set database: This is a type error: Cannot read properties of undefined (reading 'chooseAndSetDatabase')

console.log(message.redactedMessage);
// Failed to choose and set database: This is a type error: [REDACTED]
```

A disadvantage of this approach is that we don't have clear identifiers for errors and you could end up with the same error message for different error locations. However, errors should be easy to distinguish by their stack trace.

## Checklist


robertbrignull:
I like the idea a lot. I hadn't ever heard of tagged templates, but they look pretty cool. Here they do exactly what we want by being able to provide a full version and a redacted version of the message.

The only part I think could be improved here is that passing the `Error` object seems a bit clunky, and when there is no error object and we're constructing one manually we have to [duplicate the message](https://github.com/github/vscode-codeql/blob/160a6cca76ae579e487ad1975fa30d5f6536cebd/extensions/ql-vscode/src/query-serialization.ts#L29). I wonder if we could make the `Error` object be passed to `redactableErrorMessage`, and maybe we can have one fewer argument on `showAndLogExceptionWithTelemetry` and avoid duplication of the error message.

I haven't explored the above ideas in depth yet, but I'm very happy to give it a go. My current plan is to merge this branch into `robertbrignull/logged_error_telemetry` and then continue building upon it.

koesie10:
> The only part I think could be improved here is that passing the `Error` object seems a bit clunky, and when there is no error object and we're constructing one manually we have to [duplicate the message](https://github.com/github/vscode-codeql/blob/160a6cca76ae579e487ad1975fa30d5f6536cebd/extensions/ql-vscode/src/query-serialization.ts#L29). I wonder if we could make the `Error` object be passed to `redactableErrorMessage`, and maybe we can have one fewer argument on `showAndLogExceptionWithTelemetry` and avoid duplication of the error message.

I agree with you that creating an error object and duplicating the message will be hard to keep in sync. I think it would be really useful to explore whether we can just use an `Error` object which is one of the values in the `RedactableErrorMessage`. It should be relatively easy to detect and would remove the duplication. For most errors where we're duplicating the message, we don't really have an `Error` object, so I'm not sure how to handle these cases. We can construct an `Error` object in the `showAndLogExceptionWithTelemetry` method, but that will create a somewhat wrong stack trace since it will also contain the `showAndLogExceptionWithTelemetry` method itself. However, this might be an acceptable compromise to remove this duplication.

robertbrignull:
I think constructing an `Error` object within `showAndLogExceptionWithTelemetry` is acceptable. This is the approach I was thinking of taking, if needed. The stack trace will then include that method but it'll still contain all the rest of the data, so I don't think it'll be too confusing.

PR #2027, opened 2023-01-31
Introduce command to set default Code Tour database

elenatanasoiu:
Makes this possible: https://github.com/github/codespaces-codeql/pull/4

We have a codespace template which houses our CodeQL tour: https://github.com/github/codespaces-codeql

This contains a repo with a default databases already loaded for the user so that they can start writing queries more quickly.

At the moment we're asking the user to manually right click on the database folder (`codeql-tutorial-database`) and set it as the current database.

We can take this one step further by defining a command that gets triggered when we arrive at the step for setting up the database.

The command (`codeQL.setDefaultTourDatabase`) will build the URI pointing to our preloaded database and set it as the current one.

We initially considered whether we can re-use the `setCurrentDatabase` command and pass the URI of the database from the codespace itself, but the URI would be hardcoded as:

```
file://0-62/workspaces/codespaces-codeql/codeql-tutorial-database
```

as we can only pass the codeTour extension a command and string parameters.

This would have been brittle as the filepath for a codespace might change in the future.

Instead we can define a custom tour command (`setDefaultTourDatabase`) to look at the current workspace folder and build the path to the database in the CodeQL extension.

We've made the command publicly visible so that we can execute it from our code tour. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases-ui.ts
+++ b/extensions/ql-vscode/src/databases-ui.ts
@@ -363,7 +363,7 @@ export class DatabaseUI extends DisposableObject {
     token: CancellationToken,
   ): Promise<void> => {
     try {
-      if (workspace.workspaceFolders === undefined) {
+      if (!workspace.workspaceFolders?.length) {
         throw new Error("No workspace folder is open.");
       } else {
         // This specifically refers to the database folder in
--- a/extensions/ql-vscode/src/databases-ui.ts
+++ b/extensions/ql-vscode/src/databases-ui.ts
@@ -12,6 +12,7 @@ import {
   CancellationToken,
   ThemeIcon,
   ThemeColor,
+  workspace,
 } from "vscode";
 import { pathExists, stat, readdir, remove } from "fs-extra";
 
@@ -220,6 +221,15 @@ export class DatabaseUI extends DisposableObject {
     );
     this.push(
       commandRunnerWithProgress(
+        "codeQL.setDefaultTourDatabase",
+        this.handleSetDefaultTourDatabase,
+        {
+          title: "Set Default Database for Codespace CodeQL Tour",
+        },
+      ),
+    );
+    this.push(
+      commandRunnerWithProgress(
         "codeQL.upgradeCurrentDatabase",
         this.handleUpgradeCurrentDatabase,
         {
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -348,6 +358,36 @@ export class DatabaseUI extends DisposableObject {
...
 
+  private handleSetDefaultTourDatabase = async (
+    progress: ProgressCallback,
+    token: CancellationToken,
+  ): Promise<void> => {
+    try {
+      if (workspace.workspaceFolders === undefined) {
+        throw new Error("No workspace folder is open.");
+      } else {
+        const codespaceRootFolderUri = workspace.workspaceFolders[0].uri;
```
shati-patel, line 369: I think it's fine for us to say `workspace.workspaceFolders[0]`, since we control the associated tutorial workspace and we know that this is the right folder. 

Would it be helpful to add a comment mentioning that this specifically refers to https://github.com/github/codespaces-codeql?

shati-patel (commented):
Looks good, but then we paired on this, so I'm happy for another pair of eyes 😅 👀

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -348,6 +358,36 @@ export class DatabaseUI extends DisposableObject {
     }
   };
 
+  private handleSetDefaultTourDatabase = async (
+    progress: ProgressCallback,
+    token: CancellationToken,
+  ): Promise<void> => {
+    try {
+      if (workspace.workspaceFolders === undefined) {
```
adityasharad, line 366: Might as well check it's non-empty too.
```suggestion
      if (!(workspace.workspaceFolders?.length)) {
```

shati-patel:
[As discussed offline too]
Latest commit (https://github.com/github/vscode-codeql/pull/2027/commits/e2c37c5e131e0ba774dc8dbb6be9a34fb06505fd) removes our new command from the `package.json` file. We don't want anyone to call `codeQL.setDefaultTourDatabase` manually, and I've double-checked that the code tour can still use it, so no need to define it "publicly"!

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -348,6 +358,36 @@ export class DatabaseUI extends DisposableObject {
...
+            progress,
+            token,
+            uri,
+          );
+        }
+        await this.databaseManager.setCurrentDatabaseItem(databaseItem);
+      }
+    } catch (e) {
+      // rethrow and let this be handled by default error handling.
+      throw new Error(`Could not set database: ${getErrorMessage(e)}`);
```
aeisenberg, line 387: Could the error message be more precise and helpful? Something like this perhaps:

```suggestion
      throw new Error(`Could not set the database for the Code Tour. Please make sure you are default workspace of the CodeQL starter workspace in codespaces: ${getErrorMessage(e)}`);
```

Or something like that. Maybe even include a link. Most likely, if someone is seeing this error, they're not actually in the codespace, or maybe the code space didn't open the right workspace.

extensions/ql-vscode/src/databases-ui.ts
```diff
@@ -384,7 +384,7 @@ export class DatabaseUI extends DisposableObject {
       }
     } catch (e) {
       // rethrow and let this be handled by default error handling.
-      throw new Error(`Could not set database: ${getErrorMessage(e)}`);
+      throw new Error(`Could not set the database for the Code Tour. Please make sure you are default workspace of the CodeQL starter workspace in codespaces: ${getErrorMessage(e)}`);
```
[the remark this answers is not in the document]
elenatanasoiu: Thanks! Have reworded and the linting issue is now fixed.

PR #2029, opened 2023-01-31
Added version to db config

charisk:
The version isn't used yet but will be used in the future.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/databases/config/db-config-store.ts
+++ b/extensions/ql-vscode/src/databases/config/db-config-store.ts
@@ -14,6 +14,7 @@ import {
   renameLocalList,
   renameRemoteList,
   SelectedDbItem,
+  SUPPORTED_DB_CONFIG_VERSION,
 } from "./db-config";
 import * as chokidar from "chokidar";
 import { DisposableObject, DisposeHandler } from "../../pure/disposable-object";
@@ -459,6 +460,7 @@ export class DbConfigStore extends DisposableObject {
 
   private createEmptyConfig(): DbConfig {
     return {
+      version: SUPPORTED_DB_CONFIG_VERSION,
       databases: {
         variantAnalysis: {
           repositoryLists: [],
--- a/extensions/ql-vscode/src/databases/config/db-config-store.ts
+++ b/extensions/ql-vscode/src/databases/config/db-config-store.ts
@@ -14,7 +14,7 @@ import {
   renameLocalList,
   renameRemoteList,
   SelectedDbItem,
-  SUPPORTED_DB_CONFIG_VERSION,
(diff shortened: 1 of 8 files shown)


extensions/ql-vscode/src/databases/config/db-config.ts
```diff
@@ -1,6 +1,10 @@
-// Contains models for the data we want to store in the database config
+// Contains models and consts for the data we want to store in the database config.
+// Changes to these models should be done carefully and account for backwards compatibility of data.
+
+export const SUPPORTED_DB_CONFIG_VERSION = 1;
```
norascheuch, line 4: Is the `SUPPORTED_DB_CONFIG_VERSION` the latest version, so that it can be used above in the `db-config-store` as `current version`, or does it show all supported versions and should be an array?

If it's the former we should name it something like `CURRENT_DB_CONFIG_VERSION` or `LATEST_etc`?

extensions/ql-vscode/src/databases/config/db-config-store.ts
```diff
@@ -459,6 +460,7 @@ export class DbConfigStore extends DisposableObject {
 
   private createEmptyConfig(): DbConfig {
     return {
+      version: SUPPORTED_DB_CONFIG_VERSION,
```
norascheuch, line 463: See comment below, I would expect an array of all supported versions if we use this constant name.

norascheuch (commented):
Looks good, I just have one question about the constant name and difference between all supported versions and the latest.

extensions/ql-vscode/src/databases/config/db-config.ts
charisk: I'm imagining that a version of the extension works with a specific version of the db config. If the data is using a different version, the config would need to be migrated to the supported version. I can see that the name may not be ideal as we'll potentially _support_ multiple versions though (by dealing with migrating the data).

`CURRENT_DB_CONFIG_VERSION` doesn't sound great to me either though as it makes me think of user state and that may not be constant. 

Perhaps we should just go with `DB_CONFIG_VERSION`?

shati-patel: A single specified version makes sense to me. (Like what we do with [query history](https://github.com/github/vscode-codeql/blob/c5494b6fd240e82cd436d25bdf73999ea67bf875/extensions/ql-vscode/src/query-serialization.ts#L132)) 

I think just `DB_CONFIG_VERSION` is fine, but no strong feelings! 😄

shati-patel (approved):
LGTM 🎈

norascheuch:
Thanks! `DB_CONFIG_VERSION` is perfect! :)

PR #2033, opened 2023-02-02
Mock config instead of writing settings files

koesie10:
In our tests, we were writing settings files to disk because we were using the VSCode configuration API which writes settings to files. This results in flaky tests because concurrency can cause the VSCode API to misbehave.

This will switch the tests to use a mocked API by default. For some tests the real implementation is used, but the large majority of tests is now using a mocked version which only keeps track of the configuration in memory. This makes it easier to reset the state between tests since we can just empty out the in-memory configuration.

## Checklist


extensions/ql-vscode/test/vscode-tests/cli-integration/jest.setup.ts
```diff
@@ -8,23 +8,24 @@ import { dirname } from "path";
...
-import { extensions, workspace } from "vscode";
+import { ConfigurationTarget, env, extensions, workspace } from "vscode";
+import { beforeEachAction } from "../test-config";
 
-import baseJestSetup from "../jest.setup";
-
-export default baseJestSetup;
+(env as any).openExternal = () => {
+  /**/
+};
```
aeisenberg, line 17: Should this be mocked instead of outright overwritten?

extensions/ql-vscode/test/vscode-tests/cli-integration/jest.setup.ts
```diff
@@ -78,6 +79,15 @@ beforeAll(async () => {
   await extensions.getExtension("GitHub.vscode-codeql")?.activate();
 });
 
+beforeEach(async () => {
+  await beforeEachAction();
+
+  await CUSTOM_CODEQL_PATH_SETTING.updateValue(
+    process.env.CLI_PATH,
+    ConfigurationTarget.Global,
```
aeisenberg, line 87: I don't think the tests should be messing with global settings here.

```suggestion
    ConfigurationTarget.Workspace,
```

extensions/ql-vscode/test/vscode-tests/test-config.ts
```diff
@@ -1,126 +1,225 @@
...
+  }
+  current[parts[parts.length - 1]] = value;
+}
+
+interface WorkspaceConfiguration {
+  scope?: ConfigurationTarget;
+
+  get<T>(section: string | undefined, key: string): T | undefined;
+  has(section: string | undefined, key: string): boolean;
+  update(section: string | undefined, key: string, value: unknown): void;
```
aeisenberg, line 39: Should be consistent with the vscode API that this is mocking.

```suggestion
  update(section: string | undefined, key: string, value: unknown): Thenable<void>;
```

extensions/ql-vscode/test/vscode-tests/test-config.ts
```diff
@@ -1,126 +1,225 @@
...
   }
 
-  public async setInitialTestValue(value: T | undefined) {
-    this.initialTestValue = value;
+  public update(
+    section: string | undefined,
+    key: string,
+    value: unknown,
+  ): void {
+    setIn(this.values, this.getKey(section, key), value);
```
aeisenberg, line 60: As mentioned above, this should be made async to mimic the original API.

extensions/ql-vscode/test/vscode-tests/test-config.ts
```diff
@@ -1,126 +1,225 @@
...
+  return pkg.contributes.configuration.properties;
+})();
+
+export const vsCodeGetConfiguration = workspace.getConfiguration;
+export let vscodeGetConfigurationMock: jest.SpiedFunction<
+  typeof workspace.getConfiguration
+>;
+
+export const beforeEachAction = async () => {
+  const defaultConfiguration = new DefaultConfiguration(packageConfiguration);
```
aeisenberg, line 163: Would it make more sense to have the default configuration be the _actual_ configuration of the system? I guess the benefit doing it this way is that we are guaranteed that the default config is consistent across all machines and tests.

extensions/ql-vscode/test/vscode-tests/test-config.ts
```diff
@@ -1,126 +1,225 @@
...
-  ) {}
+class InMemoryConfiguration implements WorkspaceConfiguration {
+  private readonly values: Record<string, unknown> = {};
 
-  public async get(): Promise<T | undefined> {
-    return this.setting.getValue();
+  public constructor(public readonly scope: ConfigurationTarget) {}
+
+  public get<T>(section: string | undefined, key: string): T | undefined {
+    return getIn(this.values, this.getKey(section, key)) as T | undefined;
```
aeisenberg, line 48: Instead of using `getIn` and `setIn`, would it be simpler to just have a simple dictionary that has keys being the entire configuration key? So instead of this:

```
 {
  codeql: {
    cli: {
      executablePath: '/foo/bar/codeql'
    }
  }
}
```

You just have:

```
{
  'codeQL.cli.executablePath': '/foo/bar/codeql'
}
```

What do we get by slicing up the object?

extensions/ql-vscode/test/vscode-tests/cli-integration/jest.setup.ts
koesie10: We can definitely mock this instead, that will make it easier to test that any calls to this function are being made. This was a copy of the previous implementation, but I'll also change that to be a mocked function.

extensions/ql-vscode/test/vscode-tests/cli-integration/jest.setup.ts
koesie10: 👍, this shouldn't actually change any configuration settings for the user because the `beforeEachAction` has already run at this point which mocks the configuration. However, I do agree that if this isn't mocked it should update the workspace configuration.

extensions/ql-vscode/test/vscode-tests/test-config.ts
koesie10: I don't really think it needs to be consistent. This API is already inconsistent with the VSCode API because it's missing the `configurationTarget` and `overrideInLanguage` arguments. The [actual mocked method](https://github.com/github/vscode-codeql/blob/cac0add7887a6caaab8bd2329e52a2b32cc923b1/extensions/ql-vscode/test/vscode-tests/test-config.ts#L193) itself is `async` and I don't think we gain anything from also making this `async`.

extensions/ql-vscode/test/vscode-tests/test-config.ts
koesie10: Exactly. I think it's better to make the config fully deterministic rather than also needing to figure out what the actual system configuration is.

extensions/ql-vscode/test/vscode-tests/test-config.ts
koesie10: The VSCode API also accepts calls like these:

```ts
workspace.getConfiguration("codeQL").update("cli", {
  "executablePath": "/foo/bar/codeql",
}, ConfigurationTarget.Global);
```

If we were storing everything by string key, we'd need to create the correct string keys and remove any string keys which are not passed, but do exist in the configuration.

We're not using any of these calls in the extension, but I think it still makes sense to support as much of the VSCode API as possible if it's not too much work. I think the `getIn` and `setIn` implementations are easier than an implementation by string keys which supports this.

aeisenberg:
Interesting the test failure. https://github.com/github/vscode-codeql/actions/runs/4083438944/jobs/7039427032#step:6:442 Is this transient?

koesie10:
> Interesting the test failure. https://github.com/github/vscode-codeql/actions/runs/4083438944/jobs/7039427032#step:6:442 Is this transient?

This was caused by the upgrade to VSCode 1.75.0 and has been fixed in #2041. I've merged in `main`, so it has been fixed now.

The latest commit also ran successfully in CI without any retries!

aeisenberg (approved):
Nice.

PR #2037, opened 2023-02-03
Add script for retrieving original source location

koesie10:
This adds a script that can be used for retrieving the original source location when given a location in the released extension. It will download the source map from the Actions workflow run of the release and use the `source-map` library to extract the original location.

Example input:
```shell
npx ts-node scripts/source-map.ts v1.7.8 "/Users/user/.vscode/extensions/github.vscode-codeql-1.7.8/out/extension.js:131164:13"
```

Output:
```
/Users/user/.vscode/extensions/github.vscode-codeql-1.7.8/src/cli.ts:281:12
```

## Checklist

### DIFF
--- a/extensions/ql-vscode/scripts/source-map.ts
+++ b/extensions/ql-vscode/scripts/source-map.ts
@@ -4,6 +4,7 @@
  * works with released extensions.
  *
  * Usage: npx ts-node scripts/source-map.ts <version-number> <filename>:<line>:<column>
+ * For example: npx ts-node scripts/source-map.ts v1.7.8 "/Users/user/.vscode/extensions/github.vscode-codeql-1.7.8/out/extension.js:131164:13"
  */
 
 import { spawnSync } from "child_process";
--- a/extensions/ql-vscode/scripts/source-map.ts
+++ b/extensions/ql-vscode/scripts/source-map.ts
@@ -1,0 +1,124 @@
+/**
+ * This scripts helps finding the original source file and line number for a
+ * given file and line number in the compiled extension. It currently only
+ * works with released extensions.
+ *
+ * Usage: npx ts-node scripts/source-map.ts <version-number> <filename>:<line>:<column>
+ */
+
+import { spawnSync } from "child_process";
+import { basename, resolve } from "path";
+import { pathExists, readJSON } from "fs-extra";
+import { SourceMapConsumer } from "source-map";
+
+if (process.argv.length !== 4) {
+  console.error(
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/scripts/source-map.ts
```diff
@@ -0,0 +1,124 @@
+/**
+ * This scripts helps finding the original source file and line number for a
+ * given file and line number in the compiled extension. It currently only
+ * works with released extensions.
+ *
+ * Usage: npx ts-node scripts/source-map.ts <version-number> <filename>:<line>:<column>
```
aeisenberg, line 6: To be clear, this is a stack frame entry directly copied from a stack trace? Could you maybe make that explicit in the docs with an example?

extensions/ql-vscode/scripts/source-map.ts
```diff
@@ -0,0 +1,124 @@
...
+    "..",
+    "artifacts",
+    "source-maps",
+    versionNumber,
+  );
+
+  if (!(await pathExists(sourceMapsDirectory))) {
+    console.log("Downloading source maps...");
+
+    const workflowRuns = runGhJSON<WorkflowRunListItem[]>([
```
aeisenberg, line 37: So, this will only work for as long as the workflow runs are still available from actions. Ideally, we can publish the source map as part of the release process and download the release artifact instead.

aeisenberg (commented):
I think it's ok to download the artifacts from the workflow run, but it's not an ideal solution. Does anyone else have some suggestions?

charisk:
> I think it's ok to download the artifacts from the workflow run, but it's not an ideal solution. Does anyone else have some suggestions?

I guess the other option is to have the artifact in the release artifacts. Has this been considered?

extensions/ql-vscode/scripts/source-map.ts
koesie10: Definitely! We'd probably need to create a ZIP file and upload it as a release asset. I think we can do this in a follow-up PR.

aeisenberg:
> I guess the other option is to have the artifact in the release artifacts. Has this been considered?

That was what I had in mind.

charisk:
> > I guess the other option is to have the artifact in the release artifacts. Has this been considered?
> 
> That was what I had in mind, but don't know if there are any downsides to this.

I can't think of one. It feels like it makes sense too.

aeisenberg (approved):
Thanks. From my perspective this looks good. And in the future, we may move to using release artifacts for the smap.

PR #2043, opened 2023-02-06
Enable eqeqeq ESLint rule and fix violations

charisk:
Enforce use of === and !== and update places where we're not. See https://eslint.org/docs/latest/rules/eqeqeq for details.

Please shout if there are reasons we should not do this.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/.eslintrc.js
+++ b/extensions/ql-vscode/.eslintrc.js
@@ -49,7 +49,6 @@ const baseConfig = {
     "@typescript-eslint/no-throw-literal": "error",
     "no-useless-escape": 0,
     camelcase: "off",
-    eqeqeq: "off",
     "escompat/no-regexp-lookbehind": "off",
     "etc/no-implicit-any-catch": "error",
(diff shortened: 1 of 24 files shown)


shati-patel (approved):
Seems very reasonable to me 👀 Though we can wait for a second opinion from someone with more historical context 😅

charisk:
> Seems very reasonable to me 👀 Though we can wait for a second opinion from someone with more historical context 😅

Yeah I'll leave it open for now.

robertbrignull (commented):
Sounds good to me too as an idea. I also don't have historical context and haven't yet looked in detail at each change, but I could do to see if it'll be behaviour changing or not.

koesie10 (approved):
LGTM too. I think the comparisons with `undefined` are the only ones that could potentially change behaviour because `null == undefined`, but according to the types these should never be `null`.

charisk:
Thanks everyone for the reviews. I'll go ahead and merge it. @aeisenberg please shout if you see an issue with the change!

PR #2054, opened 2023-02-07
Handle `codeql-pack.yml` files everywhere

koesie10:
This will add support for the `codeql-pack.yml` filename in all places where we currently support `qlpack.yml`. It centralizes the list of the supported filenames in a single place and a method that can figure out the correct filename to use.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/helpers.ts
+++ b/extensions/ql-vscode/src/helpers.ts
@@ -23,6 +23,7 @@ import { extLogger, OutputChannelLogger } from "./common";
 import { QueryMetadata } from "./pure/interface-types";
 import { telemetryListener } from "./telemetry";
 import { RedactableError } from "./pure/errors";
+import { getQlPackPath } from "./pure/ql";
 
 // Shared temporary folder for the extension.
 export const tmpDir = dirSync({
@@ -741,21 +742,4 @@ export async function* walkDirectory(
       yield entry;
     }
   }
-}
-
-export const QLPACK_FILENAMES = ["qlpack.yml", "codeql-pack.yml"];
-export const FALLBACK_QLPACK_FILENAME = QLPACK_FILENAMES[0];
-
-export async function getQlPackPath(
-  packRoot: string,
-): Promise<string | undefined> {
-  for (const filename of QLPACK_FILENAMES) {
-    const path = join(packRoot, filename);
-
-    if (await pathExists(path)) {
-      return path;
-    }
-  }
-
-  return undefined;
 }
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/helpers.ts
```diff
@@ -737,3 +742,20 @@ export async function* walkDirectory(
...
+): Promise<string | undefined> {
+  for (const filename of QLPACK_FILENAMES) {
+    const path = join(packRoot, filename);
+
+    if (await pathExists(path)) {
+      return path;
+    }
+  }
+
+  return undefined;
```
elenatanasoiu, line 760: Could we have a test for this new function?

extensions/ql-vscode/src/quick-query.ts
```diff
@@ -158,7 +160,13 @@ export async function displayQuickQuery(
...
 }
 
-async function checkShouldRewrite(qlPackFile: string, newDependency: string) {
+async function checkShouldRewrite(
+  qlPackFile: string | undefined,
+  newDependency: string,
+) {
+  if (!qlPackFile) {
+    return true;
+  }
```
elenatanasoiu, line 169: I see we don't have a test file for `quick-query` functions so this is a stretch. If the setup isn't terribly complicated, do you think it's worth adding one to test what this is doing? 

I do see we have a `queries.test.ts` file which I guess covers some of the quick-query functionality.

elenatanasoiu (commented):
Looks good! Thanks for doing this. I left some small suggestions but they're not blocking.

extensions/ql-vscode/src/helpers.ts
koesie10: 👍 , I've added a test for this function. I've also moved the function to the `pure` directory because this doesn't actually need VSCode and that makes it clearer that this can be a unit test.

extensions/ql-vscode/src/quick-query.ts
koesie10: I could add a test for this function, but I don't really know what input values it would actually receive. I'm not quite sure when the function should be returning `true` or `false` or what it's purpose is, so I'm hesitant to add tests based purely on its implementation, rather than on its intended behaviour.

aeisenberg: Quick query is a feature where the IDE generates a bare-bones qlpack in a temporary directory that is tailored to the current language. It is meant to be semi-ephemeral.

When you call the quick query command, the extension checks if the query query pack exists already and if it is for the correct language of the current database.

extensions/ql-vscode/test/unit-tests/pure/ql.test.ts
```diff
@@ -0,0 +1,48 @@
...
+  beforeEach(() => {
+    tmpDir = dirSync({
+      prefix: "queries_",
+      keep: false,
+      unsafeCleanup: true,
+    });
+  });
+
+  afterEach(() => {
+    tmpDir.removeCallback();
```
elenatanasoiu, line 19: Nice

elenatanasoiu (approved):

PR #2036, opened 2023-02-02
Introduce command for creating skeleton QL packs

elenatanasoiu:
We'd like to make it easier for a user going through the CodeQL Tour to write their queries.

To help them along, we can generate skeleton QL packs once we know which database they're using, instead of expecting them to know how to create this themselves.

We're then able to download the necessary dependencies for their CodeQL queries.

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #2055, opened 2023-02-07
Generate QL pack for CodeTour

elenatanasoiu:
Follow-up to https://github.com/github/vscode-codeql/pull/2036 where we set up the check for existing QL packs and offer to help the user download the packs.

This PR contains the "doing the thing" part --> downloading the QL packs. 

When you run the [codespace template](https://github.com/github/codespaces-codeql) for the first time, you will automatically have QL packs downloaded for you.

It creates:
- an `example.ql` file
- a `qlpack.yml` file
- a `codeql-pack.lock.yml` file

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -28,6 +28,7 @@ import { CompilationMessage } from "./pure/legacy-messages";
 import { sarifParser } from "./sarif-parser";
 import { dbSchemeToLanguage, walkDirectory } from "./helpers";
 import { App } from "./common/app";
+import { QueryLanguage } from "./qlpack-generator";
 
 /**
  * The version of the SARIF format that we are using.
@@ -1213,6 +1214,23 @@ export class CodeQLCliServer implements Disposable {
       ["resolve", "queries"],
       args,
       "Resolving queries",
+    );
+  }
+
+  /**
+   * Adds a list of QL library packs with optional version ranges as dependencies of
+   * the current package, and then installs them. This command modifies the qlpack.yml
+   * file of the current package. Formatting and comments will be removed.
+   * @param dir The directory where QL pack exists.
+   * @param language The language of the QL pack.
+   */
+  async packAdd(dir: string, queryLanguage: QueryLanguage) {
+    const args = ["--dir", dir];
+    args.push(`codeql/${queryLanguage}-all`);
+    return this.runJsonCodeQlCliCommandWithAuthentication(
+      ["pack", "add"],
+      args,
+      "Adding pack dependencies and installing them",
     );
   }
 
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -1218,8 +1218,8 @@ export class CodeQLCliServer implements Disposable {
   }
 
   /**
-   * Adds a list of QL library packs with optional version ranges as dependencies of
-   * the current package, and then installs them. This command modifies the qlpack.yml
+   * Adds a core language QL library pack for the given query language as a dependency
+   * of the current package, and then installs them. This command modifies the qlpack.yml
    * file of the current package. Formatting and comments will be removed.
    * @param dir The directory where QL pack exists.
    * @param language The language of the QL pack.
@@ -1230,7 +1230,7 @@ export class CodeQLCliServer implements Disposable {
     return this.runJsonCodeQlCliCommandWithAuthentication(
       ["pack", "add"],
       args,
-      "Adding and installing pack dependencies.",
+      `Adding and installing ${queryLanguage} pack dependency.`,
     );
   }
 
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -1230,7 +1230,7 @@ export class CodeQLCliServer implements Disposable {
     return this.runJsonCodeQlCliCommandWithAuthentication(
       ["pack", "add"],
       args,
-      "Adding pack dependencies and installing them",
+      "Adding and installing pack dependencies.",
     );
   }
 
--- a/extensions/ql-vscode/src/databases.ts
+++ b/extensions/ql-vscode/src/databases.ts
@@ -669,6 +669,7 @@ export class DatabaseManager extends DisposableObject {
         folderName,
         databaseItem.language as QueryLanguage,
         this.cli,
+        this.ctx.storageUri?.fsPath,
       );
       await qlPackGenerator.generate();
     } catch (e: unknown) {
--- a/extensions/ql-vscode/src/databases.ts
+++ b/extensions/ql-vscode/src/databases.ts
@@ -657,7 +657,7 @@ export class DatabaseManager extends DisposableObject {
     }
 
     const answer = await showBinaryChoiceDialog(
-      `We've noticed you don't have a QL pack downloaded to analyze this database. Can we set up a query pack for you?`,
+      `We've noticed you don't have a CodeQL pack available to analyze this database. Can we set up a query pack for you?`,
     );
 
     if (!answer) {
--- a/extensions/ql-vscode/src/databases.ts
+++ b/extensions/ql-vscode/src/databases.ts
@@ -26,6 +26,7 @@ import { QueryRunner } from "./queryRunner";
 import { pathsEqual } from "./pure/files";
 import { redactableError } from "./pure/errors";
 import { isCodespacesTemplate } from "./config";
+import { QlPackGenerator, QueryLanguage } from "./qlpack-generator";
 
 /**
  * databases.ts
@@ -655,9 +656,26 @@ export class DatabaseManager extends DisposableObject {
       return;
     }
 
-    await showBinaryChoiceDialog(
-      `We've noticed you don't have a QL pack downloaded to analyze this database. Can we set up a ${databaseItem.language} query pack for you`,
+    const answer = await showBinaryChoiceDialog(
+      `We've noticed you don't have a QL pack downloaded to analyze this database. Can we set up a query pack for you?`,
     );
+
+    if (!answer) {
+      return;
+    }
+
+    try {
+      const qlPackGenerator = new QlPackGenerator(
+        folderName,
+        databaseItem.language as QueryLanguage,
+        this.cli,
+      );
+      await qlPackGenerator.generate();
+    } catch (e: unknown) {
+      void this.logger.log(
+        `Could not create skeleton QL pack: ${getErrorMessage(e)}`,
+      );
+    }
   }
 
   private async reregisterDatabases(
--- a/extensions/ql-vscode/src/qlpack-generator.ts
+++ b/extensions/ql-vscode/src/qlpack-generator.ts
@@ -1,0 +1,85 @@
+import { mkdir, writeFile } from "fs-extra";
+import { dump } from "js-yaml";
+import { join } from "path";
+import { CodeQLCliServer } from "./cli";
+
+export type QueryLanguage =
+  | "csharp"
+  | "cpp"
+  | "go"
+  | "java"
+  | "javascript"
+  | "python"
+  | "ruby"
+  | "swift";
+
+export class QlPackGenerator {
+  private readonly qlpackName: string;
+  private readonly qlpackVersion: string;
+  private readonly header: string;
+  private readonly qlpackFileName: string;
+
+  constructor(
+    private readonly folderName: string,
+    private readonly queryLanguage: QueryLanguage,
+    private readonly cliServer: CodeQLCliServer,
+  ) {
+    this.qlpackName = `getting-started/codeql-extra-queries-${this.queryLanguage}`;
+    this.qlpackVersion = "1.0.0";
+    this.header = "# This is an automatically generated file.\n\n";
+
+    this.qlpackFileName = "qlpack.yml";
+  }
+
+  public async generate() {
+    await mkdir(this.folderName);
+
+    // create qlpack.yml
+    await this.createQlPackYaml();
+
+    // create example.ql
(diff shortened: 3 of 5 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -1216,6 +1217,23 @@ export class CodeQLCliServer implements Disposable {
...
+   * @param dir The directory where QL pack exists.
+   * @param language The language of the QL pack.
+   */
+  async packAdd(dir: string, queryLanguage: QueryLanguage) {
+    const args = ["--dir", dir];
+    args.push(`codeql/${queryLanguage}-all`);
+    return this.runJsonCodeQlCliCommandWithAuthentication(
+      ["pack", "add"],
+      args,
+      "Adding pack dependencies and installing them",
```
aeisenberg, line 1233: How about? 

```suggestion
      "Adding and installing pack dependencies.",
```

extensions/ql-vscode/src/databases.ts
```diff
@@ -655,9 +656,26 @@ export class DatabaseManager extends DisposableObject {
       return;
     }
 
-    await showBinaryChoiceDialog(
-      `We've noticed you don't have a QL pack downloaded to analyze this database. Can we set up a ${databaseItem.language} query pack for you`,
+    const answer = await showBinaryChoiceDialog(
+      `We've noticed you don't have a QL pack downloaded to analyze this database. Can we set up a query pack for you?`,
```
aeisenberg, line 660: The "official" term is _CodeQL pack_.

```suggestion
      `We've noticed you don't have a CodeQL pack available to analyze this database. Can we set up a query pack for you?`,
```

extensions/ql-vscode/src/qlpack-generator.ts
```diff
@@ -0,0 +1,85 @@
+import { mkdir, writeFile } from "fs-extra";
+import { dump } from "js-yaml";
+import { join } from "path";
+import { CodeQLCliServer } from "./cli";
+
+export type QueryLanguage =
```
aeisenberg, line 6: We already have multiple places where languages are hardcoded. It would be nice not to add another one. (Even better would be to consolidate them.)

I found 4(-ish) locations:

- https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/helpers.ts#L533
- https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/packaging.ts#L11
- https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/remote-queries/gh-api/variant-analysis.ts#L13
- https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts#L30

extensions/ql-vscode/src/qlpack-generator.ts
```diff
@@ -0,0 +1,85 @@
...
+      },
+    };
+
+    await writeFile(qlPackFile, this.header + dump(qlPackYml), "utf8");
+  }
+
+  private async createExampleQlFile() {
+    const exampleQlFile = join(this.folderName, "example.ql");
+
+    const exampleQl = `
```
aeisenberg, line 64: Does this query work for all languages? Maybe? Hopefully? I wonder if long term this is a template that should be provided by the language extractor since it could change in the future.

aeisenberg: Testing that these sample queries work for all languages will be difficult. It will be a poor user experience if the first time they try to generate a query, it has a compile error.

This is probably fine for now, but I don't know if this is a good long term solution.

extensions/ql-vscode/src/qlpack-generator.ts
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * @name Empty block
```
aeisenberg, line 66: Maybe nice to add a comment here that this is auto-generated and perhaps even a url where they can go to find more information?

extensions/ql-vscode/src/qlpack-generator.ts
adityasharad: This query won't work for Ruby (where you need to either `import codeql.ruby.AST` or use `Ast::Block`). I would start by keeping the placeholder query much simpler, rather than trying to use the language libraries:
```
import <lang>
select "Hello, world!"
```

In future we can maintain example query files within the extension repo, and then we can have CI test that they still compile. Or even better, but more complicated: download one of the [example query packs](https://github.com/orgs/codeql/packages?q=example&tab=packages&q=examples), find a named example query within that, and copy its contents here.

extensions/ql-vscode/src/qlpack-generator.ts
adityasharad: Agreed. Arguably in this case you don't need the enum. We should already have the logic to infer the language name from the database, and then we pass that same inferred language name to `packAdd` without checking it again.

extensions/ql-vscode/src/qlpack-generator.ts
aeisenberg: This is definitely safer. Maybe start with the simplest query and add a link in a comment to docs where you can learn about more sophisticated queries.

Future work to generate a more meaningful query.

extensions/ql-vscode/src/qlpack-generator.ts
elenatanasoiu: @aeisenberg Thanks for rounding up all the places where we repeat the language type. My intention is to open a follow-up PR to consolidate all these places. 👍 

>  We should already have the logic to infer the language name from the database

I was thinking the same, but actually the database allows any string to be set as a language as we don't do validation on what we store on that field (as far as I can tell). So you could end up generating a QL pack for an unknown/unsupported language. 

But I agree we could have one place to store info on the language. 👍

extensions/ql-vscode/src/qlpack-generator.ts
elenatanasoiu: I had a look at what the language extractor is (Robert explained it to me) and I understand that would mean creating a new CLI command to tap into it and then use that command to generate our example file.

I'll make a note of this idea as I think it's a bit more involved to add a new command.

In the mean time I've gone the route of using Aditya's suggestion and simplify the example query. 

Thanks for the input! Really useful! Please keep it coming.

extensions/ql-vscode/src/qlpack-generator.ts
elenatanasoiu: I did add one but was afraid the comment would be significant! I'll add it back.

elenatanasoiu:
Thanks for the reviews @aeisenberg & @adityasharad. I think this is ready for review again.

shati-patel (commented):
Code looks sensible, but (as mentioned offline): 
From testing locally, it looks like we're generating the QL pack in the wrong folder at the moment (something like `AppData\Local\Programs\Microsoft VS Code\codeql-custom-queries-javascript` instead of in the actual workspace!)

extensions/ql-vscode/src/qlpack-generator.ts
```diff
@@ -0,0 +1,84 @@
...
+  ) {
+    this.qlpackName = `getting-started/codeql-extra-queries-${this.queryLanguage}`;
+    this.qlpackVersion = "1.0.0";
+    this.header = "# This is an automatically generated file.\n\n";
+
+    this.qlpackFileName = "qlpack.yml";
+  }
+
+  public async generate() {
+    await mkdir(this.folderName);
```
shati-patel, line 35: For quick queries and adding database source archives, we use `workspace.updateWorkspaceFolders`. (E.g. [here](https://github.com/github/vscode-codeql/blob/465da69ecf41b58ef2050e66a23374e932105b23/extensions/ql-vscode/src/quick-query.ts#L43)).

We probably want something similar in this case 📂

extensions/ql-vscode/src/qlpack-generator.ts
adityasharad: The database contains a `*.dbscheme` file, and the VS Code extension infers the language name from the dbscheme file name (which follows certain naming conventions). Such a file can only exist if there was a valid CodeQL extractor to produce the database. So it is IMO valid to assume that the inferred language name is valid, without doing further enum-based validation.

In the rare case of working with a DB that was produced by an experimental extractor with no published library pack (i.e. someone has locally created the extractor and DB but there is no `codeql/<lang>-all` pack published yet), generating a starter CodeQL query pack using your new command will fail, which is perfectly acceptable.

aeisenberg: > My intention is to open a follow-up PR to consolidate all these places. 

Sounds good. 👍

extensions/ql-vscode/src/qlpack-generator.ts
elenatanasoiu: Thanks for thoroughly testing this! I've pushed up a commit to save the folder in the workspace instead of VS Code local storage.

extensions/ql-vscode/src/qlpack-generator.ts
```diff
@@ -29,6 +30,7 @@ export class QlPackGenerator {
     this.header = "# This is an automatically generated file.\n\n";
 
     this.qlpackFileName = "qlpack.yml";
+    this.sourcePath = join(__dirname, "..", this.folderName);
```
aeisenberg, line 33: `__dirname` refers to the directory that this file is in. Is that what we want here? I think `this.sourcePath` refers to the location where the qlpack is going to be created. It should probably be created in the workspace, not in the extension itself.

elenatanasoiu: Yup, currently looking into using `workspace.updateWorkspaceFolders` as Shati suggested above. Sorry, I should've updated the PR! It's been a busy day after ** waves at everything ** but I'm hoping that's the last piece of the puzzle.

extensions/ql-vscode/src/qlpack-generator.ts
```diff
@@ -44,8 +52,19 @@ export class QlPackGenerator {
     await this.createCodeqlPackLockYaml();
   }
 
+  private async createWorkspaceFolder() {
```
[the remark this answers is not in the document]
elenatanasoiu: At the moment this is going to run behind a feature flag in the codespace template workspace: https://github.com/github/codespaces-codeql/pull/5

This is a multi-root workspace. 

We have planned work for making this behaviour work with the extension in the same way as the codespace. So we'll need to make sure this would work with a single root workspace. But for now, this PR is only aiming to work with the multi-root workspace.

extensions/ql-vscode/test/vscode-tests/minimal-workspace/qlpack-generator.test.ts
```diff
@@ -1,10 +1,14 @@
 import { join } from "path";
-import { existsSync, rmdirSync } from "fs";
+import { existsSync, rmSync } from "fs";
```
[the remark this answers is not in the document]
elenatanasoiu: I gave this a shot but I had to fix another problem with tests in Windows not having the same path for the directory.

I ended up adding a temporary directory and using its `removeCallback()` function, which I've seen in a few places in our tests.

extensions/ql-vscode/test/vscode-tests/minimal-workspace/qlpack-generator.test.ts
```diff
@@ -32,13 +36,13 @@ describe("QlPackGenerator", () => {
...
       mockCli,
-      __dirname,
+      dir.name,
     );
   });
 
   afterEach(async () => {
     try {
-      rmSync(packFolderPath, { recursive: true });
+      dir.removeCallback();
```
aeisenberg, line 45: Yeah...this is probable even better than `del`.

extensions/ql-vscode/src/cli.ts
```diff
@@ -1216,6 +1217,23 @@ export class CodeQLCliServer implements Disposable {
...
+   * @param dir The directory where QL pack exists.
+   * @param language The language of the QL pack.
+   */
+  async packAdd(dir: string, queryLanguage: QueryLanguage) {
+    const args = ["--dir", dir];
+    args.push(`codeql/${queryLanguage}-all`);
+    return this.runJsonCodeQlCliCommandWithAuthentication(
+      ["pack", "add"],
+      args,
+      "Adding and installing pack dependencies.",
```
aeisenberg, line 1233: Minor: Maybe include the language in the message?

```suggestion
      `Adding and installing ${queryLanguage} pack dependency.`,
```

extensions/ql-vscode/src/cli.ts
```diff
@@ -1216,6 +1217,23 @@ export class CodeQLCliServer implements Disposable {
     );
   }
 
+  /**
+   * Adds a list of QL library packs with optional version ranges as dependencies of
```
aeisenberg, line 1221: Minor: the jsdoc isn't quite right.

```suggestion
   * Adds a core language QL library pack for the given query language as a dependency of
```

extensions/ql-vscode/src/qlpack-generator.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  private async createQlPackYaml() {
+    const qlPackFilePath = join(this.folderUri.fsPath, this.qlpackFileName);
+
+    const qlPackYml = {
+      name: this.qlpackName,
+      version: this.qlpackVersion,
+      dependencies: {
+        [`codeql/${this.queryLanguage}-all`]: "*",
```
aeisenberg, line 73: This line isn't actually necessary since you are calling `packAdd` below, which does the same thing (and more).

aeisenberg (commented):
A few small thing, and after that :shipit:.

shati-patel (approved):
Thanks! 🚢

PR #2065, opened 2023-02-10
Add ability to run query with data extensions

aeisenberg:
- Add a new config that toggles between using all data extensions or none.
- If using all data extensions, then before a query evaluation, run a `codeql resolve qlpacks` command with the new `--kind` option. This will return a list of extension packs in the workspace.
- Pass these packs to the cli before evaluation/
- This will only work with CLI v2.12.3 (not yet released) or later.
- Also include some CLI tests to ensure this works.

Note that the new integration tests will be skipped until 2.12.3 is released.

## Checklist


extensions/ql-vscode/src/cli.ts
```diff
@@ -1142,19 +1142,26 @@ export class CodeQLCliServer implements Disposable {
    *   the default CLI search path is used.
    * @returns A dictionary mapping qlpack name to the directory it comes from
    */
-  resolveQlpacks(
+  async resolveQlpacks(
     additionalPacks: string[],
-    searchPath?: string[],
+    extensionPacksOnly = false,
```
robertbrignull, line 1147: When `extensionPacksOnly` is false, does it include information about normal qlpacks and extension packs, or only non-extension qlpacks? I'm asking this because we only add `"--kind", "extension"` in the `extensionPacksOnly` case.

I'm wondering if "only" is misleading, and perhaps something like `packType: "qlpack" | "extensionPack"` would be more accurate.

robertbrignull (commented):
I have a couple of comments, but these changes make sense to me. I'm mostly reviewing this in isolation and I don't have much context yet of the whole data extensions work.

extensions/ql-vscode/src/cli.ts
aeisenberg: The default case is to return all packs. `--kind=extension` will return extension packs only. `--kind=query` for query packs and `--kind=library` will return library packs (note that extension packs are a variant of library packs and will be returned as well).

I can add comments to explain this better.

extensions/ql-vscode/test/vscode-tests/cli-integration/queries.test.ts
```diff
@@ -96,6 +98,59 @@ describeWithCodeQL()("Queries", () => {
     await cleanDatabases(databaseManager);
   });
 
+  describe("extension packs", () => {
+    skipIfTrue(qs.cliServer.cliConstraints.supportsQlpacksKind());
```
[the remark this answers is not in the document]
aeisenberg: Hmmm...I will have to rework this. Good catch.

aeisenberg:
Hmmm...linting is no longer working in the editor, so I keep missing linting errors.

extensions/ql-vscode/src/cli.ts
robertbrignull: Thanks for explaining. The parameter name and behaviour is fine then 👍

extensions/ql-vscode/src/cli.ts
```diff
@@ -1138,23 +1138,31 @@ export class CodeQLCliServer implements Disposable {
   /**
    * Gets information about available qlpacks
    * @param additionalPacks A list of directories to search for qlpacks before searching in `searchPath`.
```
robertbrignull, line 1140: Thanks for updating the doc comment. Sorry but `seachPath` is also mentioned in this line.

extensions/ql-vscode/test/vscode-tests/cli.ts
```diff
@@ -42,3 +42,11 @@ export function itWithCodeQL() {
 
   return it;
 }
+
+export function skipIfTrue(condition: () => Thenable<boolean>) {
+  return beforeEach(async () => {
+    if (await condition) {
```
robertbrignull, line 48: Should be
```suggestion
    if (await condition()) {
```
as otherwise this will always be true, assuming a function object is truthy.

Or convert `condition` back into a `Thenable<boolean>`. I don't think the extra layer of being function is necessary to make this work.

As an extra side point, what benefit does using `Thenable` give over using `Promise`? I'm curious to know which one we should use in general.

robertbrignull (commented):
A couple more tiny points but then this looks good to me. Thanks for explaining bits. The use of the CLI now makes sense to me.

extensions/ql-vscode/test/vscode-tests/cli.ts
aeisenberg: Actually, I don't think the function is really doing much. I'm moving to vanilla `beforeEach...pending`.

aeisenberg:
OK...all tests are passing, but the new cli integration test is not running since it relies on CLI version 2.12.3.

aeisenberg:
v2.11.3 is now available. I am taking this PR out of draft. @robertbrignull, would you be able to take another look?

robertbrignull (commented):
One final comment. Sorry I think it was missed last time.

Other than that this look good to me. I've done a quick test too that everything still works as before when the config option is not set and this won't change behaviour for users that don't opt-in.

extensions/ql-vscode/src/cli.ts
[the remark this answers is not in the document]
aeisenberg: Hmmm...this is actually technically correct, even though it might be a bit misleading. This is turning into one of the dark corners of pack resolution in the CLI.

The original idea was that people would use the `.config/codeql/config` file and hard code a `--search-path` option in the file that applies to every command anywhere on their machine. Then the workspace folders are added to the `--additional-packs` path.

While this is still true and possible, we don't recommend this way of working any more. 

So, that's a little background. And I think you're right that it is confusing and I will remove the line.

aeisenberg:
@robertbrignull now that the MRVA release is out, can you take another look? I'm hoping to merge this soon.

PR #2070, opened 2023-02-13
Remove remote queries manager

koesie10:
This removes the remote queries manager and any code that was only being called by the remote queries manager.

Based on #2069

## Checklist


extensions/ql-vscode/src/query-history/query-history-manager.ts
```diff
@@ -1547,9 +1476,11 @@ the file in the file explorer and dragging it into the workspace.`,
         false,
       );
     } else if (item.t === "remote") {
-      await this.remoteQueriesManager.openRemoteQueryResults(item.queryId);
+      // Do nothing. TODO: Remove when remote queries is no longer supported.
     } else if (item.t === "variant-analysis") {
       await this.variantAnalysisManager.showView(item.variantAnalysis.id);
+    } else {
+      assertNever(item);
```
elenatanasoiu, line 1483: We're not removing the branch for checking whether an item is `remote`. Why do we need to add the extra `else` here?

extensions/ql-vscode/src/remote-queries/remote-queries-view.ts
```diff
@@ -129,13 +108,13 @@ export class RemoteQueriesView extends AbstractWebview<
...
       additionalOptions: {
-        localResourceRoots: [Uri.file(this.analysesResultsManager.storagePath)],
+        localResourceRoots: [],
       },
     };
   }
 
   protected onPanelDispose(): void {
-    this.currentQueryId = undefined;
+    // No-op
```
elenatanasoiu: Why are we keeping this view around instead of removing it?

elenatanasoiu: Oh I see you've removed in the next commit. 👍

extensions/ql-vscode/src/query-history/query-history-manager.ts
koesie10: We don't necessarily need to, but this makes it consistent with most of the other `if-else` constructs. This will make it an error if we add a new possible `item.t` value. I'm happy to remove it, but I do think it adds some extra safety.

extensions/ql-vscode/src/remote-queries/remote-queries-view.ts
koesie10: Sorry, I should've specified that I added two commits to this PR.

extensions/ql-vscode/src/query-history/query-history-manager.ts
elenatanasoiu: I see! In which case it sounds like a good addition. Thanks for explaining.

elenatanasoiu (approved):

PR #2072, opened 2023-02-13
Inject extension pack dependencies into MRVA packs

aeisenberg:
If the user requests that extension packs be included in their MRVA run,
then do the following:

1. Search the workspace for all extension packs
2. Add each extension pack as an explicit and direct dependency on
   the generated pack.

It is ok to use `*` as a dependency since we are guaranteed that
exactly one version of each injected extension pack dependency is
available when the pack is being compiled.

If we find multiple paths to an extension pack of the same name, this
is an error since it is ambiguous which path to use.

Commit-by-commit review is recommended.

The first commit introduces a refactoring with no behavioural changes. Second commit introduces the actual change for this PR.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -1284,10 +1284,24 @@ export class CodeQLCliServer implements Disposable {
     );
   }
 
-  async packInstall(dir: string, forceUpdate = false) {
+  async packInstall(
+    dir: string,
+    { forceUpdate = false, workspaceFolders = [] as string[] } = {},
+  ) {
     const args = [dir];
     if (forceUpdate) {
       args.push("--mode", "update");
+    }
+    if (workspaceFolders?.length > 0) {
+      if (await this.cliConstraints.supportsAdditionalPacksInstall()) {
+        args.push(
+          // Allow prerelease packs from the ql submodule.
+          "--allow-prerelease",
+          // Allow the use of --additional-packs argument without issueing a warning
+          "--no-strict-mode",
+          ...this.getAdditionalPacksArg(workspaceFolders),
+        );
+      }
     }
     return this.runJsonCodeQlCliCommandWithAuthentication(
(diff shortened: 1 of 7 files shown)


aeisenberg:
Nightly build is failing because the tip of main does not have the fix in it that gets this PR working on 2.12.4. When 2.12.4 is finally released and the branch is merged back into main and a new nightly build is created, we can re-run the failing jobs and they should pass. 

While we are waiting for that, this PR is finally ready for review.

extensions/ql-vscode/test/vscode-tests/cli-integration/data-extension-pack/`
```diff
@@ -0,0 +1,12 @@
+extensions:
```
robertbrignull, line 1: Is the name of this file meant to be a back-tick?

aeisenberg: Oh...that's odd. This must have been accidentally committed. It can be removed. I'll fix it.

robertbrignull (commented):
It's slightly more complicated than using extension packs for local queries, but it looks reasonable to me. I'm not an expert on what each of the CodeQL CLI calls does, but if it works when tested that's good enough for me. As I understand it there should be no changes needed to the MRVA code in the API or workflow so this could work as it is now. I'd love to try it out, but I would need some advice on how to do that.

aeisenberg:
> As I understand it there should be no changes needed to the MRVA code in the API or workflow so this could work as it is now. 

That is correct.

> I'd love to try it out, but I would need some advice on how to do that.

I'm on vacation starting tomorrow, but I will DM you some instructions on how to try this out.

robertbrignull (approved):
We're fairly confident this won't break existing functionality while the extension packs flag is turned off, so I'm approving based on the testing that @aeisenberg has done. I'll also do my own testing later on, and report any problems if I find any.

PR #2074, opened 2023-02-14
Consolidate the way we define query languages

elenatanasoiu:
Suggested here: https://github.com/github/vscode-codeql/pull/2055#discussion_r1100491734

We already have multiple places where languages are hardcoded. It would be nice to consolidate them:

- https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/helpers.ts#L533
- https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/packaging.ts#L11
- https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/remote-queries/gh-api/variant-analysis.ts#L13
- https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/remote-queries/shared/variant-analysis.ts#L30

## Questions
1. I've noticed that [packaging.ts](https://github.com/github/vscode-codeql/pull/2074/commits/43b984e80795a4ed44484146d8e81fddd6e56e76#diff-fde2f27b0f599d4ee9705a6d5ff252d5adb6d2c6b65d412074197b1229064690R14-R28) doesn't mention CodeQL packs for Swift. I don't see them in the CodeQL docs either. Do we want to add them?
2. I've added `QueryLanguage` in a new `types` folder. Ideally it would live in some sort of shared folder. Would `common` work?

## Checklist


extensions/ql-vscode/src/packaging.ts
```diff
@@ -9,18 +9,23 @@ import { ProgressCallback, UserCancellationException } from "./commandRunner";
...
-  "codeql/csharp-queries",
-  "codeql/go-queries",
-  "codeql/java-queries",
-  "codeql/javascript-queries",
-  "codeql/python-queries",
-  "codeql/ruby-queries",
-  "codeql/csharp-solorigate-queries",
-  "codeql/javascript-experimental-atm-queries",
-];
+const PACKS_BY_QUERY_LANGUAGE = {
```
aeisenberg, line 14: This is definitely better to be using constants that define the different languages, but their use is still distributed throughout the code. Could you consolidate the different objects that contain the the languages in one area? This way, when we add new languages it will be easier to ensure that we've updated all the right blocks.

elenatanasoiu: I've opted to move this constant and the dbSchemeToLanguage constant into the same file as the type.

extensions/ql-vscode/src/types/query-language.ts
```diff
@@ -0,0 +1,37 @@
+export enum QueryLanguage {
```
shati-patel, line 1: As discussed separately, we could put this somewhere (in `common`?) instead of creating a new folder 📁 💥

shati-patel (commented):
One comment, looks good otherwise!

aeisenberg (approved):
Nice.

extensions/ql-vscode/test/factories/remote-queries/shared/variant-analysis-submission.ts
```diff
@@ -1,8 +1,6 @@
 import { faker } from "@faker-js/faker";
-import {
-  VariantAnalysisQueryLanguage,
-  VariantAnalysisSubmission,
-} from "../../../../src/remote-queries/shared/variant-analysis";
+import { VariantAnalysisSubmission } from "../../../../src/remote-queries/shared/variant-analysis";
+import { QueryLanguage } from "../../../../src/common/query-languagee";
```
aeisenberg, line 3: Oops...spleling.

elenatanasoiu:
Thanks for the review @aeisenberg !

PR #2075, opened 2023-02-14
Enable eslint rule to disallow awaiting a non-thenable type

robertbrignull:
Enables https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/await-thenable.md

I spotted that VS Code is giving suggestions for this error in the editor but not in CI. I guess it's not included in `@typescript-eslint/recommended`. Enabling this eslint rule means we can fix this case and avoid introducing more in the future.

I think this fix in `extension.ts` is a change of behaviour, because the [void operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void) evaluates the expression and then throws away the result and returns `undefined`. So previously the `chosenAction === installActionName` condition was always false.

## Checklist


aeisenberg (approved):

extensions/ql-vscode/src/extension.ts
```diff
@@ -462,7 +462,7 @@ export async function activate(
     ) {
       registerErrorStubs([checkForUpdatesCommand], (command) => async () => {
         const installActionName = "Install CodeQL CLI";
-        const chosenAction = await void showAndLogErrorMessage(
+        const chosenAction = await showAndLogErrorMessage(
```
aeisenberg, line 465: Interesting. I guess it was never possible to install the cli from this code path. I never tried this out, though so I don't really know.

Thanks for catching this.

robertbrignull:
Oh and the CI found more cases which didn't show up locally for whatever reason. I'll look into them tomorrow.

robertbrignull:
I've pushed more linting fixes. I believe all these extra changes are behaviour preserving.

robertbrignull:
Ok, third time's the charm. Had to tix another alert that had been introduced on `main` since I opened this PR 🤦

koesie10 (approved):

PR #2076, opened 2023-02-14
Add unhandled rejection listener

robertbrignull:
I need to check if this works as I believe it does, but I think this is along the right lines for catching unhandled errors and logging telemetry for them.

## Checklist


extensions/ql-vscode/src/extension.ts
```diff
@@ -1529,6 +1530,16 @@ async function activateWithInstalledDistribution(
   };
 }
 
+function addUnhandledRejectionListener() {
+  process.addListener("unhandledRejection", (error: unknown) => {
+    void showAndLogExceptionWithTelemetry(
+      redactableError(asError(error))`Unhandled error: ${getErrorMessage(
+        error,
+      )}`,
+    );
```
aeisenberg, line 1539: What happens if `showAndLogExceptionWithTelemetry` throws? It will be an `unhandledRejection`, I think. And this would cause a loop of this handler being called. Unlikely to happen, but if it happens, it will be bad.

I think it would be best to add a `.then(...)` handler here, with a catch argument that just logs to. Or, you can just be _really_ sure that `showAndLogExceptionWithTelemetry` will never throw.

robertbrignull: Good point. I think we can add a `catch` to `showAndLogExceptionWithTelemetry` which then logs in a much more simple way. Or alternatively we could just fully ignore it in this case. That wouldn't put us in a worse place than we were before this PR.

robertbrignull: Actually it seems NodeJS thought of this:
> Exceptions thrown from within the event handler will not be caught. Instead the process will exit with a non-zero exit code and the stack trace will be printed. This is to avoid infinite recursion.

From https://nodejs.org/api/process.html#event-uncaughtexception

extensions/ql-vscode/src/extension.ts
```diff
@@ -1529,6 +1530,35 @@ async function activateWithInstalledDistribution(
...
+function addUnhandledRejectionListener() {
+  const handler = (error: unknown) => {
+    const message = redactableError(
+      asError(error),
+    )`Unhandled error: ${getErrorMessage(error)}`;
+    // Add a catch so that showAndLogExceptionWithTelemetry fails, we avoid
+    // triggering "unhandledRejection" and avoid an infinite loop
+    showAndLogExceptionWithTelemetry(message).catch(
+      (telemetryError: unknown) => {
+        void extLogger.log(
```
robertbrignull, line 1542: To test this I tried throwing a test exception, and also added an error to `showAndLogExceptionWithTelemetry`. Annoying something weird is going on and the second error seems to be getting lost. For example it prints
```
Failed to send error telemetry: Cannot read properties of undefined (reading 't')
Unhandled error: Cannot read properties of undefined (reading 't')
```
even though the two messages should be different.

I'm therefore not sure whether to keep this or just log one of the messages.

aeisenberg (approved):
Nice

robertbrignull:
I've done some more manual testing of having or not having the rejection handler and throwing exceptions in various ways. I think it's all working as expected and won't cause problems. Also no telemetry is sent unless the config option is enabled, and that option is still currently off by default.

It does slightly change behaviour because some exceptions were previously lost and not shown to the user. If an exception happens in a command then it'll be shown to the user as an error, with the command name. If an exception happens within something like a `setTimeout` or an unhandled promise then it is not tied to the command in the same way and also not shown to the user.

I think this change of behaviour is acceptable. The existing logging is done by us, so our intention is to log exceptions. Therefore this PR makes the situation more consistent because now we're logging more exceptions that previously were lost.

PR #2080, opened 2023-02-15
Don't offer to create skeleton pack for default database in CodeTour

elenatanasoiu:
Once the user completes the CodeQL tour, they'll be able to add additional databases so they can continue to write queries on their own. For these databases, we want to offer to create the skeleton pack (implemented [here](https://github.com/github/vscode-codeql/pull/2055)). 

However, in Step 3 in the code tour, we also add a default database for you. At this point we don't need to create the skeleton pack, so we're disabling this functionality for the default database.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases-ui.ts
+++ b/extensions/ql-vscode/src/databases-ui.ts
@@ -379,11 +379,14 @@ export class DatabaseUI extends DisposableObject {
         );
 
         let databaseItem = this.databaseManager.findDatabaseItem(uri);
+        const isTutorialDatabase = true;
         if (databaseItem === undefined) {
           databaseItem = await this.databaseManager.openDatabase(
             progress,
             token,
             uri,
+            "CodeQL Tutorial Database",
+            isTutorialDatabase,
           );
         }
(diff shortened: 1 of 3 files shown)


aeisenberg (approved):
Is the reason why we don't need to create a pack for the default database because there already is a pack?

The code LGTM, but I would just like to understand more about why this is the case.

elenatanasoiu:
> Is the reason why we don't need to create a pack for the default database because there already is a pack?

@aeisenberg That's right, the pack for this db is committed into the repo [here](https://github.com/github/codespaces-codeql/commit/142228bd3eb90bd6990db30aec5bd3c87beae0d0). 

Although I'm not quite sure why we've added two `codeql-pack.yml` files in that commit. Maybe @adityasharad can enlighten us both? :)

adityasharad:
One was `codeql/tutorial-queries` (a skeleton query pack), the other was `codeql/tutorial` (a library pack with the actual tutorial library). Checking in `codeql/tutorial` was always a temporary measure; now we are moving towards removing the checked-in copy in favour of the published `codeql/tutorial` pack.

PR #2081, opened 2023-02-15
Don't send `--format` option for `codeql pack add` command

elenatanasoiu:
Follow-up to https://github.com/github/vscode-codeql/pull/2055.

Once we create a `qlpack.yml` file for you, we run the `codeql pack add` command to install all the dependencies for it.

For this [we added a wrapper](https://github.com/github/vscode-codeql/blob/95f46b3b3a69889532b504dbd8ccad3958da0f16/extensions/ql-vscode/src/cli.ts#L1227) around the `codeql pack add` command in the extension. 

By default, a `--format` option is added to the command when we call `runJsonCodeQlCliCommandWithAuthentication`.

However, `codeql pack add` doesn't support this option so we need to turn it off.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -1227,10 +1227,12 @@ export class CodeQLCliServer implements Disposable {
   async packAdd(dir: string, queryLanguage: QueryLanguage) {
     const args = ["--dir", dir];
     args.push(`codeql/${queryLanguage}-all`);
+    const addFormat = false;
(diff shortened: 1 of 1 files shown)


koesie10 (approved):
It would probably make sense to convert the arguments of the `runJsonCodeQlCliCommand` to an options object instead of having multiple optional positional arguments, but making it clear like this makes sense to me. If we do want to convert this to an options object, we should probably do it as part of a tech-debt issue.

elenatanasoiu:
Thanks for the suggestion @koesie10, have added an issue.

PR #2085, opened 2023-02-15
Fix Graphviz WASM module not loading for graph viewer

### LINKED ISSUE #2077: Graph viewer: blocked by CSP?
**Describe the bug**
Previously, running a graph query would render the results in the graph viewer. Now, the results view shows but no graph is rendered. Instead, the console shows this error:

```
webview.js:1696 Uncaught (in promise) RuntimeError: Aborted(CompileError: WebAssembly.instantiate(): Wasm code generation disallowed by embedder). Build with -sASSERTIONS for more info.
    at $ (webview.js:1696:87261)
    at webview.js:1696:130912
```

**Version**
CodeQL extension version: 1.7.8 
CodeQL CLI version: 2.12.2 
Platform: darwin x64

**To reproduce**
With a database selected, run a `@kind graph` query.

**Expected behavior**
The results view should render a graph.

**Additional context**
It seems that the CSP policy now disallows the wasm that `d3-graphviz` evaluates to render the graph. But I'm not super familiar with internals of this extension so I may be mistaken.

koesie10:
It seems that when we added the CSP policy to the webview, we did not take into account that `d3-graphviz` uses `@hpcc-js/wasm` to load Graphviz as a WASM module. This commit adds `'wasm-unsafe-eval'` to the CSP policy to allow this.

This also fixes a warning emitted by the module because we're not telling it we're not using WebWorkers.

Closes #2077

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/interface-utils.ts
+++ b/extensions/ql-vscode/src/interface-utils.ts
@@ -163,7 +163,7 @@ export function getHtmlForWebview(
   /*
    * Content security policy:
    * default-src: allow nothing by default.
-   * script-src: allow only the given script, using the nonce.
+   * script-src: allow the given script, using the nonce. also allow loading WebAssembly modules.
    * style-src: allow only the given stylesheet, using the nonce.
    * connect-src: only allow fetch calls to webview resource URIs
    * (this is used to load BQRS result files).
@@ -172,7 +172,7 @@ export function getHtmlForWebview(
 <html>
   <head>
     <meta http-equiv="Content-Security-Policy"
-          content="default-src 'none'; script-src 'nonce-${nonce}'; font-src ${fontSrc}; style-src ${styleSrc}; connect-src ${
+          content="default-src 'none'; script-src 'nonce-${nonce}' 'wasm-unsafe-eval'; font-src ${fontSrc}; style-src ${styleSrc}; connect-src ${
     webview.cspSource
   };">
         ${stylesheetsHtmlLines.join(`    ${EOL}`)}
--- a/extensions/ql-vscode/src/view/results/graph.tsx
+++ b/extensions/ql-vscode/src/view/results/graph.tsx
@@ -5,7 +5,7 @@ import {
   InterpretedResultSet,
   GraphInterpretationData,
 } from "../../pure/interface-types";
-import { graphviz } from "d3-graphviz";
+import { graphviz, GraphvizOptions } from "d3-graphviz";
 import { tryGetLocationFromString } from "../../pure/bqrs-utils";
 export type GraphProps = ResultTableProps & {
   resultSet: InterpretedResultSet<GraphInterpretationData>;
@@ -59,11 +59,12 @@ export class Graph extends React.Component<GraphProps> {
       return;
     }
 
-    const options = {
+    const options: GraphvizOptions = {
       fit: true,
       fade: false,
       growEnteringEdges: false,
       zoom: true,
+      useWorker: false,
     };
 
     const element = document.querySelector(`#${graphId}`);
@@ -77,8 +78,7 @@ export class Graph extends React.Component<GraphProps> {
     const borderColor = getComputedStyle(element).borderColor;
     let firstPolygon = true;
 
-    graphviz(`#${graphId}`)
-      .options(options)
+    graphviz(`#${graphId}`, options)
       .attributer(function (d) {
         if (d.tag === "a") {
           const url = d.attributes["xlink:href"] || d.attributes["href"];


koesie10:
> Two questions:
> 
> * Does `wasm-eval` work? I see that option is designed for Chrome apps and extensions so it may not work in Electron, but it's good to document why we can't use it.

I tried using `'wasm-eval'`, but it doesn't seem to work. I found the `'wasm-unsafe-eval'` [here](https://github.com/WebAssembly/content-security-policy/blob/dd75e5ba3d31aa50cda1216e7ae15170c72ce7c7/proposals/CSP.md#the-wasm-unsafe-eval-source-directive), and it seems like Chrome did change the name of the directive (see [the Electron update](https://github.com/electron/electron/pull/29751/files#diff-dc35ee05282a97cdeaa1472b7ed710ec622bb452d16c75723bb8da46766bd60fR356) and [the Chromium commit](https://chromium.googlesource.com/chromium/src/+/791c8335079bbe7abe7a1b7b6251317ec62b2563)). [The spec](https://w3c.github.io/webappsec-csp/#can-compile-wasm-bytes) also only lists the `'wasm-unsafe-eval'` as a valid option.

> * Is there a way to avoid `wasm-unsafe-eval` the same way we've avoided JS `unsafe-eval`, by only allowing WebAssembly that has the same nonce? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src suggests no such option but perhaps you have found something else in your investigation.

Unfortunately, [the spec](https://w3c.github.io/webappsec-csp/#can-compile-wasm-bytes) doesn't list this as a possibility:

> If source-list is non-null, and does not contain a [source expression](https://w3c.github.io/webappsec-csp/#source-expression) which is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the string "['unsafe-eval'](https://w3c.github.io/webappsec-csp/#grammardef-unsafe-eval)", and does not contain a [source expression](https://w3c.github.io/webappsec-csp/#source-expression) which is an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for the string "['wasm-unsafe-eval'](https://w3c.github.io/webappsec-csp/#grammardef-wasm-unsafe-eval)", then:
> 1. Let violation be the result of executing [§ 2.4.1 Create a violation object for global, policy, and directive](https://w3c.github.io/webappsec-csp/#create-violation-for-global) on global, policy, and "script-src".
> 2. Set violation’s [resource](https://w3c.github.io/webappsec-csp/#violation-resource) to "wasm-eval".
> ...

There's nothing about checking the nonce, so I don't think it's possible to require a specific nonce.

elenatanasoiu:
I've read through the links you've added (thanks for documenting it 👍 ). It's not clear to me what level of risk this change has. Would we be able to make a single exception for `d3-graphviz` instead of changing our CSP?

koesie10:
> I've read through the links you've added (thanks for documenting it 👍 ). It's not clear to me what level of risk this change has. Would we be able to make a single exception for `d3-graphviz` instead of changing our CSP?

I don't think we're able to do that, there doesn't seem to be a possibility to specify a nonce in the [recommendation](https://github.com/WebAssembly/content-security-policy/blob/dd75e5ba3d31aa50cda1216e7ae15170c72ce7c7/proposals/CSP.md) (see also `https://github.com/WebAssembly/content-security-policy/issues/37`). The recommendation does [list the risks](https://github.com/WebAssembly/content-security-policy/blob/dd75e5ba3d31aa50cda1216e7ae15170c72ce7c7/proposals/CSP.md#risk-analysis-of-webassembly-api) associated with enabling this option. I believe the main risk would be executing untrusted WASM modules, but for this it is necessary for some JavaScript to already be compromised. As long as all JavaScript we're loading is trusted, we shouldn't be able to load untrusted WebAssembly.

elenatanasoiu:
> As long as all JavaScript we're loading is trusted, we shouldn't be able to load untrusted WebAssembly.

Hmm, if we trusted that all our JS packages are safe, I think our particular department would be a bit obsolete. 

Would you be open to considering other JS packs that don't require a CSP change?

adityasharad:
Agree with Elena's suggestions. Or could we use a different CSP just for the graph viewer, e.g. by having a boolean `allowWasmUnsafeEval` parameter when constructing each CSP?

aeisenberg:
> The graph viewer is an unsupported part of the extension that seems to have been created as part of a hackathon (see #1111). I don't think we should invest any more time than necessary into the graph viewer. I'm not sure there are many other packages that allow us to do the same thing without rewriting large parts of the graph viewer itself.

This is correct and I agree. This feature is not officially supported and if it is too much of a support burden, we can turn it off.

There are internal users who rely on this feature. And it would be worth understanding how often and in what capacity they use it.

extensions/ql-vscode/src/abstract-webview.ts
```diff
@@ -116,6 +117,7 @@ export abstract class AbstractWebview<
       config.view,
       {
         allowInlineStyles: true,
+        allowWasmEval: config.allowWasmEval ?? false,
```
aeisenberg, line 120: Minor, optional: this is slightly more typesafe as it will coerce `allowWasmEval` to a bool.

```suggestion
        allowWasmEval: !!config.allowWasmEval,
```

adityasharad (approved):

extensions/ql-vscode/src/interface.ts
```diff
@@ -656,7 +658,8 @@ export class ResultsView extends AbstractWebview<
     }
     let data;
     let numTotalResults;
-    if (metadata?.kind === GRAPH_TABLE_NAME) {
+    // Graph results are only supported in canary mode because the graph viewer is unsupported
```
aeisenberg, line 661:
```suggestion
    // Graph results are only supported in canary mode because the graph viewer is not actively supported
```

aeisenberg (approved):
I haven't actually tried this out. The code looks good. A couple of comment suggestions and after that, I think it is fine. Thanks for working on this.

extensions/ql-vscode/src/interface-utils.ts
```diff
@@ -163,7 +166,7 @@ export function getHtmlForWebview(
   /*
    * Content security policy:
    * default-src: allow nothing by default.
-   * script-src: allow only the given script, using the nonce.
+   * script-src: allow the given script, using the nonce. also allow loading WebAssembly modules.
```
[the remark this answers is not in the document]
koesie10: The `'wasm-unsafe-eval'` string is actually part of `script-src`, so I've split it broken up the `script-src` into two parts: one for the script and one for wasm eval.

aeisenberg (approved):

PR #2090, opened 2023-02-16
Install dependencies for tutorial query in codespace

elenatanasoiu:
Required by this PR: https://github.com/github/codespaces-codeql/pull/6

When a user goes through the Code Tour, we select a dummy `csv` database
for them to get them up and running.

Once they complete the code tour and would like to continue writing
queries, they will need to add their own database.

After they do that, we check the language of their new database and
generate a skeleton QL pack for them so that they don't need to create
these files by hand. See [1] for details.

This skeleton pack folder will be called
`codeql-custom-queries-<language>` and it comes with its own example
query: `example.ql`.

When we try to run this example query, the query gets confused about
which `dbscheme` to pick, as it sees a `qlpack.yml` file in the new
skeleton pack folder, as well as one in the existing `tutorial-lib`
folder.

So we'll need to get rid of the `tutorial-lib` folder in order to make room
for new queries to be run once the tour is complete.

This PR introduces a `handleTourDependencies` step which will trigger a 
`codeql pack install` command in order to install real library dependencies 
for `tutorial-queries`, since we no longer have the dummy library in 
`tutorial-lib`.

[1]: https://github.com/github/vscode-codeql/pull/2055

### DIFF
--- a/extensions/ql-vscode/src/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/local-databases-ui.ts
@@ -390,6 +390,7 @@ export class DatabaseUI extends DisposableObject {
           );
         }
         await this.databaseManager.setCurrentDatabaseItem(databaseItem);
+        await this.handleTourDependencies();
       }
     } catch (e) {
       // rethrow and let this be handled by default error handling.
@@ -398,6 +399,23 @@ export class DatabaseUI extends DisposableObject {
           e,
         )}`,
       );
+    }
+  };
+
+  private handleTourDependencies = async (): Promise<void> => {
+    // find workspace root folder
+    if (!workspace.workspaceFolders?.length) {
+      throw new Error("No workspace folder is open.");
+    } else {
+      // find tutorial queries folder
+      const tutorialQueriesPath = Uri.parse(
+        `${workspace.workspaceFolders?.[0].uri}/tutorial-queries`,
+      );
+      const cli = this.queryServer?.cliServer;
+      if (!cli) {
+        throw new Error("No CLI server found");
+      }
+      await cli.packInstall(tutorialQueriesPath.fsPath);
     }
   };
 
--- a/extensions/ql-vscode/src/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/local-databases-ui.ts
@@ -407,7 +407,8 @@ export class DatabaseUI extends DisposableObject {
       throw new Error("No workspace folder is open.");
     } else {
       const tutorialQueriesPath = join(
-        `${workspace.workspaceFolders[0].uri.fsPath}/tutorial-queries`,
+        workspace.workspaceFolders[0].uri.fsPath,
+        "tutorial-queries",
       );
       const cli = this.queryServer?.cliServer;
       if (!cli) {
--- a/extensions/ql-vscode/src/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/local-databases-ui.ts
@@ -403,19 +403,17 @@ export class DatabaseUI extends DisposableObject {
   };
 
   private handleTourDependencies = async (): Promise<void> => {
-    // find workspace root folder
     if (!workspace.workspaceFolders?.length) {
       throw new Error("No workspace folder is open.");
     } else {
-      // find tutorial queries folder
-      const tutorialQueriesPath = Uri.parse(
-        `${workspace.workspaceFolders?.[0].uri}/tutorial-queries`,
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/local-databases-ui.ts
```diff
@@ -410,6 +411,23 @@ export class DatabaseUI extends DisposableObject {
...
 
+  private handleTourDependencies = async (): Promise<void> => {
+    // find workspace root folder
+    if (!workspace.workspaceFolders?.length) {
+      throw new Error("No workspace folder is open.");
+    } else {
+      // find tutorial queries folder
+      const tutorialQueriesPath = Uri.parse(
+        `${workspace.workspaceFolders?.[0].uri}/tutorial-queries`,
+      );
```
aeisenberg, line 422: Any reason why you are (re-)parsing this URI to add a path to it? Would this be better?

```suggestion
      const tutorialQueriesPath = path.join(workspace.workspaceFolders[0].uri, "tutorial-queries");
```

Also, no need for `?.` since you checked above that there is at least one element.

extensions/ql-vscode/src/local-databases-ui.ts
```diff
@@ -410,6 +411,23 @@ export class DatabaseUI extends DisposableObject {
...
+      const cli = this.queryServer?.cliServer;
+      if (!cli) {
+        throw new Error("No CLI server found");
+      }
+      await cli.packInstall(tutorialQueriesPath.fsPath);
```
aeisenberg, line 427: And if you do above, you can change to this:

```suggestion
      await cli.packInstall(tutorialQueriesPath);
```

aeisenberg (commented):
What you have now is probably fine, just I think it's a little simpler to avoid re-parsing the URI.

extensions/ql-vscode/src/local-databases-ui.ts
elenatanasoiu: You're absolutely right, I went a bit deep on debugging https://github.com/github/vscode-codeql/pull/2104 so I wanted to be extra sure file paths weren't the issue when I was seeing a large list of errors. When I was working on introducing skeleton packs, I had some trouble getting these paths to work on Windows/Linux/MacOS. 

I've switched back to using a `join` 👍  Thanks!

extensions/ql-vscode/src/local-databases-ui.ts
```diff
@@ -401,6 +402,21 @@ export class DatabaseUI extends DisposableObject {
     }
   };
 
+  private handleTourDependencies = async (): Promise<void> => {
+    if (!workspace.workspaceFolders?.length) {
+      throw new Error("No workspace folder is open.");
+    } else {
+      const tutorialQueriesPath = join(
+        `${workspace.workspaceFolders[0].uri.fsPath}/tutorial-queries`,
```
aeisenberg, line 410:
```suggestion
        workspace.workspaceFolders[0].uri.fsPath, "tutorial-queries",
```

elenatanasoiu: Oops! 🙈

adityasharad:
This looks reasonable, but it makes me wonder why we need the separate command. Could the tour call the existing `CodeQL: Install Pack Dependencies` command and specify `tutorial-queries` as the query pack whose dependencies we want to install?

elenatanasoiu:
@adityasharad Thanks. You're right, we don't need to do this in the extension since the tutorial folder is only present in the codespace. 

I've opened a new PR to run the install command: https://github.com/github/codespaces-codeql/pull/10

extensions/ql-vscode/src/local-databases-ui.ts
```diff
@@ -390,6 +390,7 @@ export class DatabaseUI extends DisposableObject {
           );
         }
         await this.databaseManager.setCurrentDatabaseItem(databaseItem);
+        await this.handleTourDependencies();
```
adityasharad, line 393: Minor: Arguably this should be called at a higher level, not as part of `handleSetDefaultTourDatabase`.

elenatanasoiu:
Spoke to @adityasharad about this. We couldn't get this to work with running the packInstall command: https://github.com/github/codespaces-codeql/pull/10

So I'm reopening this PR. We know this works so we can merge this and spend time separately thinking about how to run parameterised commands in the tour extension.

adityasharad (approved):
Approving as a temporary strategy, until we figure out how to pass directory URIs to commands via CodeTour.

PR #2092, opened 2023-02-17
Cache schema validator for db config

charisk:
Pre-compile the jSON schema used for validating the db config and cache the validation function.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/databases/config/db-config-validator.ts
+++ b/extensions/ql-vscode/src/databases/config/db-config-validator.ts
@@ -1,6 +1,6 @@
 import { readJsonSync } from "fs-extra";
 import { resolve } from "path";
-import Ajv from "ajv";
+import Ajv, { ValidateFunction } from "ajv";
 import { clearLocalDbConfig, DbConfig } from "./db-config";
 import { findDuplicateStrings } from "../../pure/text-utils";
 import {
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/databases/config/db-config-validator.ts
```diff
@@ -9,22 +9,22 @@ import {
 } from "../db-validation-errors";
 
 export class DbConfigValidator {
-  private readonly schema: any;
+  private readonly validateSchemaFn: ValidateFunction;
 
   constructor(extensionPath: string) {
     const schemaPath = resolve(extensionPath, "databases-schema.json");
-    this.schema = readJsonSync(schemaPath);
+    const schema = readJsonSync(schemaPath);
```
koesie10, line 16: Can we also "cache" the JSON file by importing it instead of reading it from disk? I believe you should be able to do something like `import * as schema from "../../../databases-schema.json";`

koesie10 (approved):
LGTM

extensions/ql-vscode/src/databases/config/db-config-validator.ts
charisk: Unfortunately not atm because the schema file is outside of `/src` so TS can't see it. We can potentially move it, but I'll leave that for another day.

PR #2098, opened 2023-02-20
Use options objects instead of multiple optional arguments in CLI methods

koesie10:
In some of the methods of the `CodeQLCliServer` class there we had multiple optional arguments, including default values. This made it quite hard to read and required specifying the default values for an argument if you only needed to change a later argument. This will switch them to have options objects which are easier to read and don't require repeating default values.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/cli.ts
+++ b/extensions/ql-vscode/src/cli.ts
@@ -125,14 +125,6 @@ export interface SourceInfo {
 export type ResolvedTests = string[];
 
 /**
- * Options for `codeql test run`.
- */
-export interface TestRunOptions {
-  cancellationToken?: CancellationToken;
-  logger?: Logger;
-}
-
-/**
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/cli.ts
```diff
@@ -485,8 +477,13 @@ export class CodeQLCliServer implements Disposable {
     command: string[],
     commandArgs: string[],
     description: string,
-    cancellationToken?: CancellationToken,
-    logger?: Logger,
+    {
```
charisk, line 480: The aim of the PR is to make things more readable so I'm going to pick on this a bit. Is this definitely more readable? Is it worth defining some type(s) that cover the different kind of options and their default values instead of inlining them in the function definitions, or are the functions just too different?

charisk (approved):

PR #2099, opened 2023-02-21
Starcke/qlpack gen rename

starcke:
The first commit changes the name of the generated pack from `qlpack.yml` to `codeql-pack.yml`.

The second commit is a single error message change and some comment changes. I noticed them when doing the change, but I can just revert that commit if we do not want to change it.

## Checklist

N/A.


norascheuch:
Regarding the changed comments: do we want to advertise both names?

starcke:
> Regarding the changed comments: do we want to advertise both names?

Not sure. I am not super familiar with the context for the change, but I found it odd to refer to a specific name when it could actually be another name. An alternative is to use a less concrete term like "codeql pack file" but often a concrete name is easier to understand.

norascheuch (approved):
Elena confirmed since both names are possible we can advertise them!

PR #2101, opened 2023-02-21
Add unique-command-use.ql

robertbrignull:
Introduces a new CodeQL query for detecting when a VS Code command is used in more than one location. By "location" we mean a UI element or being called from the code or being available in the command palette. The intention is to make our telemetry more useful, because in the telemetry we can't tell which place the command was called from. To remedy an alert, the command should effectively be duplicated with a new name but referencing the same implementation.

This is still a draft, to share the query and get thoughts. I'll come and add more comments, and I'm sure the CodeQL team will have some helpful advice.

## Checklist

### DIFF
--- a/.github/codeql/queries/unique-command-use.ql
+++ b/.github/codeql/queries/unique-command-use.ql
@@ -1,0 +1,93 @@
+/**
+ * @name A VS Code command should not be used in multiple locations
+ * @kind problem
+ * @problem.severity warning
+ * @id vscode-codeql/unique-command-use
+ * @description Using each VS Code command from only one location makes
+ * our telemtry more useful because we can differentiate more user
+ * interactions and know which features of the UI users are using.
+ */
+
+ import javascript
+
+ abstract class CommandUsage extends Locatable {
+   abstract string getCommandName();
+ 
+   predicate isUsedFromOtherPlace() {
+     exists(CommandUsage e | e != this and e.getCommandName() = this.getCommandName())
+   }
+ 
+   predicate isFirstUsage() {
+     forall(CommandUsage e | e.getCommandName() = this.getCommandName() |
+       e.getLocationOrdinal() >= this.getLocationOrdinal()
+     )
+   }
+ 
+   string getLocationOrdinal() {
+     result =
+       this.getFile().getRelativePath() + ":" + this.getLocation().getStartLine() + ":" +
+         this.getLocation().getStartColumn()
+   }
+ }
+ 
+ class CommandUsageCallExpr extends CommandUsage, CallExpr {
+   CommandUsageCallExpr() {
+     this.getCalleeName() = "executeCommand" and
+     this.getArgument(0).(StringLiteral).getValue().matches("%codeQL%") and
+     not this.getFile().getRelativePath().matches("extensions/ql-vscode/test/%")
+   }
+ 
+   override string getCommandName() { result = this.getArgument(0).(StringLiteral).getValue() }
+ }
+ 
+ class CommandUsagePackageJsonMenuItem extends CommandUsage, JsonObject {
+   CommandUsagePackageJsonMenuItem() {
+     this.getFile().getBaseName() = "package.json" and
+     exists(this.getPropValue("command")) and
+     exists(JsonObject topObject, string menuName |
+       not exists(topObject.getParent()) and
+       topObject
+           .getPropValue("contributes")
+           .getPropValue("menus")
+           .getPropValue(menuName)
+           .getElementValue(_) = this and
+       menuName != "commandPalette"
+     )
+   }
+ 
+   override string getCommandName() { result = this.getPropValue("command").getStringValue() }
+ }
+ 
+ predicate isDisabledInCommandPalette(string commandName) {
+   exists(JsonObject topObject, JsonObject commandPaletteObject |
+     not exists(topObject.getParent()) and
+     topObject
+         .getPropValue("contributes")
+         .getPropValue("menus")
+         .getPropValue("commandPalette")
+         .getElementValue(_) = commandPaletteObject and
+     commandPaletteObject.getPropValue("command").getStringValue() = commandName and
+     commandPaletteObject.getPropValue("when").getStringValue() = "false"
+   )
+ }
+ 
+ class CommandUsagePackageJsonCommandPalette extends CommandUsage, JsonObject {
+   CommandUsagePackageJsonCommandPalette() {
+     this.getFile().getBaseName() = "package.json" and
+     exists(this.getPropValue("command")) and
+     exists(JsonObject topObject |
+       not exists(topObject.getParent()) and
+       topObject.getPropValue("contributes").getPropValue("commands").getElementValue(_) = this
+     ) and
+     not isDisabledInCommandPalette(this.getPropValue("command").getStringValue())
+   }
+ 
+   override string getCommandName() { result = this.getPropValue("command").getStringValue() }
+ }
+ 
+ from CommandUsage e
+ where
+   e.isUsedFromOtherPlace() and
+   not e.isFirstUsage()
+ select e, "The " + e.getCommandName() + " command is used from another location"
+ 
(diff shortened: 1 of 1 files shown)


.github/codeql/queries/unique-command-use.ql
```diff
@@ -0,0 +1,93 @@
...
+   override string getCommandName() { result = this.getArgument(0).(StringLiteral).getValue() }
+ }
+ 
+ class CommandUsagePackageJsonMenuItem extends CommandUsage, JsonObject {
+   CommandUsagePackageJsonMenuItem() {
+     this.getFile().getBaseName() = "package.json" and
+     exists(this.getPropValue("command")) and
+     exists(JsonObject topObject, string menuName |
+       not exists(topObject.getParent()) and
+       topObject
```
koesie10, line 49: It seems like there's a `PackageJson` class in CodeQL. Can we use that here and in the other places where we're doing the same thing? Something like:

```suggestion
     exists(PackageJson packageJson, string menuName |
       packageJson
```

I'm not sure about this, but I think this would also further restrict this query: it's currently not checking that the `topObject` is from the `package.json` file.

aeisenberg: I think something like this would work.

.github/codeql/queries/unique-command-use.ql
```diff
@@ -0,0 +1,93 @@
+/**
+ * @name A VS Code command should not be used in multiple locations
+ * @kind problem
+ * @problem.severity warning
+ * @id vscode-codeql/unique-command-use
+ * @description Using each VS Code command from only one location makes
```
aeisenberg, line 6: So, the goal is to ensure we use different commands for different kinds of interactions even if they have the same handler? Eg- distinguishing between right-click->Run Query and CMD+Shift+P->Run Query?

And this is only around helping to understand which ways our users are interacting with the product?

.github/codeql/queries/unique-command-use.ql
```diff
@@ -0,0 +1,93 @@
...
+   string getLocationOrdinal() {
+     result =
+       this.getFile().getRelativePath() + ":" + this.getLocation().getStartLine() + ":" +
+         this.getLocation().getStartColumn()
+   }
+ }
+ 
+ class CommandUsageCallExpr extends CommandUsage, CallExpr {
+   CommandUsageCallExpr() {
+     this.getCalleeName() = "executeCommand" and
```
aeisenberg, line 35: This class doesn't prevent invocations of `executeCommand` from different kinds of interactions if the actual call to `executeCommand` is wrapped in a function called in different ways. I'm not sure how we would solve this, though.

aeisenberg (commented):
Overall, looks reasonable. A few questions inline.

.github/codeql/queries/unique-command-use.ql
```diff
@@ -0,0 +1,93 @@
...
+   }
+ 
+   override string getCommandName() { result = this.getPropValue("command").getStringValue() }
+ }
+ 
+ from CommandUsage e
+ where
+   e.isUsedFromOtherPlace() and
+   not e.isFirstUsage()
+ select e, "The " + e.getCommandName() + " command is used from another location"
```
adityasharad, line 92: Skipping the first usage is clever. I can also see it being confusing if you want to audit all the places that call a command and decide which one you want to preserve (which may not be the first). As a consumer of the alert, would it be useful to just create an alert for each use? Then you don't need `isFirstUsage`. 
You can also put the total count in the message, e.g. `count(CommandUsage use | use.getCommandName() = ...)`.

.github/codeql/queries/unique-command-use.ql
robertbrignull: Yes, your understanding is correct. The goal is to distinguish between different ways of performing the same action, so distinguishing between right-click->Run Query and CMD+Shift+P->Run Query would be an example because one is the editor context menu and the other is the command palette. So they should be registered in different places in package.json

.github/codeql/queries/unique-command-use.ql
robertbrignull: Good point we haven't actually asserted that `topObject` comes from package.json. And thanks for pointing out the `PackageJson` class too.

.github/codeql/queries/unique-command-use.ql
robertbrignull: You're right it doesn't protect again `executeCommand` being wrapped, but as far as we could tell that doesn't really happen. I'm hoping it won't be a big issue in practice 🤞 

I'm also not sure how to solve it in a better way, so taking this current approach may be all we can do realistically.

.github/codeql/queries/unique-command-use.ql
robertbrignull: The original aim of omitting the first usage was so that going from one usage to two usages only creates one alert rather than two. But you make a very good point that the one that gets alerted isn't necessarily the one you want to remove/update. If everyone is ok with it perhaps it would be best to just alert on every instance so it's easier to find the usages.

Including the number of usages in the message is also a good idea. If you go from two usages to three, would than count as 1 alert added, or would it count as two alerts removed and three alerts added 🤔

robertbrignull: After discussion with @norascheuch we have decided to go the other way and instead only alert once per command. So this means if there are multiple usages we have to pick one, and we decided to stick with this existing `isFirstUsage` implementation. The exact implementation doesn't matter too much and we can change it in the future. For right now we just want this query to be useful to us and we know that the location being alerted on may not be the place to change but you should look for other usages too.

To make this easier, I've changed the alert message to include the number of usages, and added some documentation to the query indicating that you should search for other usages when fixing.

robertbrignull:
@norascheuch, you can see the new alerts at https://github.com/github/vscode-codeql/pull/2101/checks?check_run_id=11630028878. Let me know how they look to you.

.github/codeql/queries/unique-command-use.ql
```diff
@@ -0,0 +1,134 @@
+/**
+ * @name A VS Code command should not be used in multiple locations
+ * @kind problem
+ * @problem.severity warning
+ * @id vscode-codeql/unique-command-use
+ * @description Using each VS Code command from only one location makes
+ * our telemtry more useful because we can differentiate more user
```
adityasharad, line 7: minor: some typos in this description

.github/codeql/queries/unique-command-use.ql
```diff
@@ -0,0 +1,134 @@
...
+    * and the usage given may not be the one that should be changed when fixing
+    * the alert.
+    */
+   CommandUsage getFirstUsage() {
+     result.getCommandName() = this and
+     forall(CommandUsage e | e.getCommandName() = this |
+       e.getLocationOrdinal() >= result.getLocationOrdinal()
+     )
+   }
+ }
```
adityasharad, line 43: QL suggestions that should preserve the meaning but hopefully make this more intuitive to maintain:
- Create a member predicate `CommandUsage getAUse() { result.getCommandName() = this}`
- `getNumberOfUsages` then becomes `count(this.getAUse())`
- `getFirstUsage` then becomes `result = max(CommandUsage use | use = this.getAUse() | use order by use.getLocationOrdinal())`
- Optional: drop `getLocationOrdinal` entirely and use `result = max(CommandUsage use | use = this.getAUse() | use order by use.getFile().getRelativePath(), use.getLocation().getStartLine(), use.getLocation().getStartColumn())`

adityasharad (commented):
Looks sensible. Some minor suggestions on description and QL structure.

.github/codeql/queries/unique-command-use.ql
robertbrignull: Thanks @adityasharad for the suggestions. That's a lot cleaner! ❤️

adityasharad (approved):

norascheuch (approved):
Alerts look good, thank you!

PR #2104, opened 2023-02-21
Fix `Unhandled error: Invalid arguments` bug

elenatanasoiu:
When we added an extra optional param `isTutorialDatabase` to `openDatabase` [1] , the tests all passed but there is at least one call [2] where this triggers an `Unhandled arguments` error.

image

This is because the method call doesn't know which optional param we're passing (`displayName` or `isTutorialDatabase`).

Let's revert back to the original method signature where we had just one optional parameter (the database's `displayName`) and set it to a special "CodeQL Tutorial Database" value (which we want to do anyway).

We can then use this to skip creating a skeleton pack for the tutorial database since it comes with a checked-in QL pack.

[1]: https://github.com/github/vscode-codeql/blob/5e51bb57f5f76b57768fb6a3543f153f5752c23c/extensions/ql-vscode/src/local-databases.ts#L610
[2]: https://github.com/github/vscode-codeql/blob/5e51bb57f5f76b57768fb6a3543f153f5752c23c/extensions/ql-vscode/src/databaseFetcher.ts#L257-L262

NB: Unfortunately the error message doesn't link to any further information or a stack trace but after testing with this change, the error is resolved.

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -37,6 +37,7 @@ import {
   CliConfigListener,
   DistributionConfigListener,
   isCanary,
+  isCodespacesTemplate,
   joinOrderWarningThreshold,
   MAX_QUERIES,
   QueryHistoryConfigListener,
@@ -234,7 +235,10 @@ export async function activate(
   const distributionConfigListener = new DistributionConfigListener();
   await initializeLogging(ctx);
   await initializeTelemetry(extension, ctx);
-  addUnhandledRejectionListener();
+  if (!isCodespacesTemplate()) {
+    addUnhandledRejectionListener();
+  }
+
   install();
 
   const codelensProvider = new QuickEvalCodeLensProvider();
--- a/extensions/ql-vscode/src/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/local-databases-ui.ts
@@ -380,12 +380,7 @@ export class DatabaseUI extends DisposableObject {
 
         let databaseItem = this.databaseManager.findDatabaseItem(uri);
         if (databaseItem === undefined) {
-          databaseItem = await this.databaseManager.openDatabase(
-            progress,
-            token,
-            uri,
-            "CodeQL Tutorial Database",
-          );
+          databaseItem = await this.openTutorialDatabase(progress, token, uri);
         }
         await this.databaseManager.setCurrentDatabaseItem(databaseItem);
       }
@@ -397,6 +392,22 @@ export class DatabaseUI extends DisposableObject {
         )}`,
       );
     }
+  };
+
+  // Opens the tutorial database for the CodeQL Tour.
+  // We set the database name as "CodeQL Tutorial Database" in purpose. This means
+  // we won't attempt to create a skeleton QL pack for this database.
+  openTutorialDatabase = async (
+    progress: ProgressCallback,
+    token: CancellationToken,
+    uri: Uri,
+  ): Promise<DatabaseItem> => {
+    return await this.databaseManager.openDatabase(
+      progress,
+      token,
+      uri,
+      "CodeQL Tutorial Database",
+    );
   };
 
   handleRemoveOrphanedDatabases = async (): Promise<void> => {
--- a/extensions/ql-vscode/src/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/local-databases-ui.ts
@@ -379,14 +379,12 @@ export class DatabaseUI extends DisposableObject {
         );
 
         let databaseItem = this.databaseManager.findDatabaseItem(uri);
-        const isTutorialDatabase = true;
         if (databaseItem === undefined) {
           databaseItem = await this.databaseManager.openDatabase(
             progress,
             token,
             uri,
             "CodeQL Tutorial Database",
-            isTutorialDatabase,
           );
         }
         await this.databaseManager.setCurrentDatabaseItem(databaseItem);
--- a/extensions/ql-vscode/src/local-databases.ts
+++ b/extensions/ql-vscode/src/local-databases.ts
@@ -607,7 +607,6 @@ export class DatabaseManager extends DisposableObject {
     token: vscode.CancellationToken,
     uri: vscode.Uri,
     displayName?: string,
-    isTutorialDatabase?: boolean,
   ): Promise<DatabaseItem> {
     const contents = await DatabaseResolver.resolveDatabaseContents(uri);
     // Ignore the source archive for QLTest databases by default.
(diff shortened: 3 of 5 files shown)


extensions/ql-vscode/src/local-databases-ui.ts
```diff
@@ -379,14 +379,12 @@ export class DatabaseUI extends DisposableObject {
         );
 
         let databaseItem = this.databaseManager.findDatabaseItem(uri);
-        const isTutorialDatabase = true;
         if (databaseItem === undefined) {
           databaseItem = await this.databaseManager.openDatabase(
             progress,
             token,
             uri,
             "CodeQL Tutorial Database",
```
norascheuch, line 387: I don't particularly like that we're matching on the DisplayName later in the local-databases.ts file, because someone might change it here without thinking about the consequences. Can we add a comment here making this dependency clear?

elenatanasoiu:
Chatted to @aeisenberg about this. It looks like the unhandled rejection listener will pick up errors from other extensions, not just our own, which is why it's not working with the code tour:

<details>
  <summary>Stack trace</summary>
```
stack trace: Error: Invalid arguments
    at new i (/Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/workbench/api/node/extensionHostProcess.js:94:15029)
    at apply (/Users/elenatanasoiu/.vscode/extensions/vsls-contrib.codetour-0.0.58/src/player/decorator.ts:97:23)
    at Array.map (<anonymous>)
    at Proxy.map (/Users/elenatanasoiu/.vscode/extensions/vsls-contrib.codetour-0.0.58/node_modules/mobx/lib/mobx.module.js:3381:30)
    at /Users/elenatanasoiu/.vscode/extensions/vsls-contrib.codetour-0.0.58/src/player/decorator.ts:96:43
    at Generator.next (<anonymous>)
    at s (/Users/elenatanasoiu/.vscode/extensions/vsls-contrib.codetour-0.0.58/dist/extension-node.js:16:122957)
```
</details>

What this means is that our extension will not play very well if the user has other extensions installed.

For now I'm disabling it for the code tour extension but we probably want to consider only logging these errors if they're coming from the ql-vscode extension.

extensions/ql-vscode/src/extension.ts
```diff
@@ -234,7 +235,10 @@ export async function activate(
   const distributionConfigListener = new DistributionConfigListener();
   await initializeLogging(ctx);
   await initializeTelemetry(extension, ctx);
-  addUnhandledRejectionListener();
+  if (!isCodespacesTemplate()) {
+    addUnhandledRejectionListener();
+  }
+
```
aeisenberg, line 241: I'm concerned about just disabling the unhandled rejection listener for the template. It means that none of our new users will get the benefits of it. I think we'd be better off inspecting the stack frames for references to our sources (as we discussed on our chat). 

My biggest concern is that we're going to display errors from other extensions (I think it's even possible for this to happen now, just no one has reported it yet).

aeisenberg (commented):
If this is blocking your progress, then I think we can go ahead with this change, but realize that this is technical debt that should be addressed later.

robertbrignull:
> This is because the method call doesn't know which optional param we're passing (`displayName` or `isTutorialDatabase`).

Could you explain more? I've never heard of this happening before, and it sounds like a serious bug in javascript if this is what's happening. Have you found any documentation / bug reports / anyone else who had the same error?

The closest I've found online is https://udn.realityripple.com/docs/Web/JavaScript/Reference/Errors/Typed_array_invalid_arguments which implies it could be a case of abusing some javascript builtin function and it's throwing this error. Unfortunately doesn't really help tracking down where it's happening if the exception didn't have a stack trace.

> NB: Unfortunately the error message doesn't link to any further information or a stack trace but after testing with this change, the error is resolved.

We could change the "unhandled error" listener to also log the stack trace. I think it currently only logs the message and sends the stack trace to the telemetry. It could be useful if it logs the stack trace as well so debugging error reports from users is easier.

robertbrignull:
> Chatted to `@aeisenberg` about this. It looks like the unhandled rejection listener will pick up errors from other extensions, not just our own, which is why it's not working with the code tour.

Hmm, I wasn't aware it did this. I thought / hoped that each extension would be its own node process and therefore we'd be isolated from each other and not able to break each other by changing global properties. This is potentially problematic. I agree we should make it filter the errors to only those that occur in our extension. That doesn't have to / shouldn't be this PR though. I'll open an issue to investigate it.

> This means that our extension might not play well if the user has other extensions installed.

It will show more errors to the user, but as far as I'm aware it won't change the behaviour of the code producing the errors. So in the case of this "invalid arguments" error I think it's happening regardless and we just aren't aware of it without the error listener. Happy to be convinced otherwise if there's evidence though.

> For now I'm disabling it for the code tour extension

Is the code tour a separate extension from the main CodeQL extension, or instead an optional feature/mode of the CodeQL extension? I'm not that aware of how the code tour in implemented. Probably doesn't affect what we do in this PR but I'm keen to understand.

robertbrignull:
I've opened https://github.com/github/vscode-codeql/pull/2107 which reverts the PR that introduced the listener. I've also opened an internal issue to investigate what's going on here regarding errors from other extensions.

elenatanasoiu:
> Could you explain more? I've never heard of this happening before, and it sounds like a serious bug in javascript if this is what's happening. Have you found any documentation / bug reports / anyone else who had the same error?

I think we can actually drop those changes related to optional parameters. I was guessing where the problem was at that point because I couldn't see a stack trace. I think the only thing to fix is to disable the listener for the code tour. 

> We could change the "unhandled error" listener to also log the stack trace. I think it currently only logs the message and sends the stack trace to the telemetry. It could be useful if it logs the stack trace as well so debugging error reports from users is easier.

You can actually see the stack trace in the extension. If you look closely, the error is in `.vscode/extensions/vsls-contrib.codetour-0.0.58/src/player/decorator.ts`. It would be great if we could make it clear the error is being thrown by the listener. 

Screenshot 2023-02-22 at 10 06 15

> I'll open an issue to investigate it.

Thanks!

> Happy to be convinced otherwise if there's evidence though.

It happens when you move through each of the steps in code tour so it looks like it's broken. But other than that, I'm not sure how visible this would be. I think it depends what other extensions you have installed. They would have the error but the CodeQL extension would be the one that would throw the error so it would look broken. 

> Is the code tour a separate extension from the main CodeQL extension, or instead an optional feature/mode of the CodeQL extension?

It's a separate extension: https://github.com/microsoft/codetour.

elenatanasoiu:
@robertbrignull Thanks for opening https://github.com/github/vscode-codeql/pull/2107! I'm going to close this since that fixes my problem.

PR #2106, opened 2023-02-21
Update skeleton query

elenatanasoiu:
We mark this generated query as `@kind problem`.

We'll need to change the query a bit to make it fit this type of ...
erm... kind.

This means the results view will be formatted to display the file name
next to each of the results.

We're also getting rid of any mentions of an empty block query since
that's no longer what it checks.


aeisenberg (approved):

elenatanasoiu:
Thanks @aeisenberg ! 💐

PR #2117, opened 2023-02-27
Split usage of runQuery and runQueryOnMultipleDatabases

norascheuch:
Since we're tracking user behaviour through command usage we need to split them by action.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -69,8 +69,6 @@
     "onCommand:codeQLVariantAnalysisRepositories.removeItemContextMenu",
     "onCommand:codeQL.quickQuery",
     "onCommand:codeQL.restartQueryServer",
-    "onCommand:codeql.runQueryContextEditor",
-    "onCommand:codeql.runQueryOnMultipleDatabasesContextEditor",
     "onWebviewPanel:resultsView",
     "onWebviewPanel:codeQL.variantAnalysis",
     "onFileSystem:codeql-zip-archive"
@@ -309,12 +307,12 @@
         "title": "CodeQL: Run Query on Selected Database"
       },
       {
-        "command": "codeQL.runQueryOnMultipleDatabases",
-        "title": "CodeQL: Run Query on Multiple Databases"
-      },
-      {
         "command": "codeQL.runQueryContextEditor",
         "title": "CodeQL: Run Query on Selected Database"
+      },
+      {
+        "command": "codeQL.runQueryOnMultipleDatabases",
+        "title": "CodeQL: Run Query on Multiple Databases"
       },
       {
         "command": "codeQL.runQueryOnMultipleDatabasesContextEditor",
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/extension.ts
```diff
@@ -871,6 +939,25 @@ async function activateWithInstalledDistribution(
...
+
+  // Since we are tracking extension usage through commands, this command mirrors the runQuery command
+  ctx.subscriptions.push(
+    commandRunnerWithProgress(
+      "codeQL.runQueryContextEditor",
+      async (
+        progress: ProgressCallback,
+        token: CancellationToken,
+        uri: Uri | undefined,
+      ) => await compileAndRunQuery(false, uri, progress, token, undefined),
```
robertbrignull, line 951:
```suggestion
      compileAndRunQueryOnMultipleDatabases,
```
This can be shortened to just pass in the reference to the method, instead of defining a new anonymous function with the same parameters. Should be just as type-safe but with less repetition.

extensions/ql-vscode/src/extension.ts
```diff
@@ -881,71 +968,22 @@ async function activateWithInstalledDistribution(
...
+  );
+  // Since we are tracking extension usage through commands, this command mirrors the runQueryOnMultipleDatabases command
+  ctx.subscriptions.push(
+    commandRunnerWithProgress(
+      "codeQL.runQueryOnMultipleDatabasesContextEditor",
+      async (
+        progress: ProgressCallback,
+        token: CancellationToken,
+        uri: Uri | undefined,
+      ) => await compileAndRunQueryOnMultipleDatabases(progress, token, uri),
```
robertbrignull, line 986:
```suggestion
      compileAndRunQueryOnMultipleDatabases,
```

extensions/ql-vscode/src/extension.ts
robertbrignull: Talked about this in person with @norascheuch, but I was totally wrong with this (and then next comment too), since here I'm suggesting replacing `compileAndRunQuery` with `compileAndRunQueryOnMultipleDatabases` 🤦 . And we can ignore the bit about using a reference to the method too.

extensions/ql-vscode/package.json
```diff
@@ -310,6 +312,14 @@
         "command": "codeQL.runQueryOnMultipleDatabases",
         "title": "CodeQL: Run Query on Multiple Databases"
       },
+      {
```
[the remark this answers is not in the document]
norascheuch: 👍

robertbrignull:
Ah, this PR is introducing 2 new alerts, because we haven't done a `where: false` to remove the new commands (e.g. `codeQL.runQueryContextEditor`) from the command palette.

It also doesn't appear to fix the existing two alerts, which I'm unsure about but I think that's just because the PR to add the query was merged after this PR was opened.

PR #2118, opened 2023-02-27
Clean up variant analyses from query history

charisk:
Updated query history clean up logic to also clean up variant analyses.

This required some minor refactoring - see individual commits.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/query-history/query-history-scrubber.ts
+++ b/extensions/ql-vscode/src/query-history/query-history-scrubber.ts
@@ -1,8 +1,9 @@
-import { pathExists, readdir, stat, remove, readFile } from "fs-extra";
+import { pathExists, stat, remove, readFile } from "fs-extra";
 import { EOL } from "os";
 import { join } from "path";
 import { Disposable, ExtensionContext } from "vscode";
 import { extLogger } from "../common";
+import { readDirFullPaths } from "../pure/files";
 import { QueryHistoryDirs } from "./query-history-dirs";
 import { QueryHistoryManager } from "./query-history-manager";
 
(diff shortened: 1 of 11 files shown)


extensions/ql-vscode/src/query-history/query-history-scrubber.ts
```diff
@@ -74,18 +76,33 @@ async function scrubQueries(
...
+        "Cleaning up query history directories. Removing old entries.",
+      );
+
+      if (!(await pathExists(queryHistoryDirs.localQueriesDirPath))) {
+        void extLogger.log(
+          `Cannot clean up query history directories. Local queries directory does not exist: ${queryHistoryDirs.localQueriesDirPath}`,
+        );
+        return;
+      }
+      if (!(await pathExists(queryHistoryDirs.variantAnalysesDirPath))) {
```
charisk, line 89: I think this is ok to do since https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/extension.ts#L628 is not checking for canary flags.

extensions/ql-vscode/src/query-history/query-history-scrubber.ts
```diff
@@ -74,18 +76,33 @@ async function scrubQueries(
...
-      if (!(await pathExists(queryDirectory))) {
+      void extLogger.log(
+        "Cleaning up query history directories. Removing old entries.",
+      );
+
+      if (!(await pathExists(queryHistoryDirs.localQueriesDirPath))) {
+        void extLogger.log(
+          `Cannot clean up query history directories. Local queries directory does not exist: ${queryHistoryDirs.localQueriesDirPath}`,
+        );
+        return;
```
robertbrignull, line 87: Do we want to return here? If the local query history directory doesn't exist we can still clean up the variant analysis directory. I suppose we do expect both dirs to exist, so this is still an unexpected error, but do we want to abort fully or try to continue.

robertbrignull (approved):
One question, but I think this is fine as is. It took me a while to understand what's going on, and the tests are still a bit daunting. So if you're also not too confident you may want to get a second review. I leave it up to you.

extensions/ql-vscode/src/query-history/query-history-scrubber.ts
charisk: I thought that since this is an exceptional case then it's okay to abort. I don't have a strong opinion though - happy to change it to be more resilient.

robertbrignull: You're right it should be an exceptional case that doesn't happen in practice. Also it's not like one of local or remote queries is a more central part of the extension, so it's not like a user would want to only use one and therefore want to run without one of the directories existing. And we recreate the directories on extension startup. So if this does happen it hopefully won't be a long term problem and will be fixed by restarting.

That's a lot of words, but in short I think it's fine as it is. Feel free to abort if any of the directories don't exist and keep the code simple.

charisk:
I'll leave this open until tomorrow in case @aeisenberg wants to have a look (no pressure though).

PR #2122, opened 2023-03-01
Remove `as unknown as QuickPickItem`

koesie10:
This is definitely not a perfect solution since we're essentially just moving the place where we're casting. However, because we have manually made the types similar, this provides some type assurances where there were none before. This also has the cast in only one place, which makes it easier to find and fix in the future.

## Checklist


elenatanasoiu (approved):
Looks like `mockedQuickPickItem` extends `QuickPickItem` directly rather than forcing the cast through `as unknown as` so it's definitely better in terms of type-checking (as far as I can tell). 

So I think that means we're addressing our linting issue with the same pattern of creating a mock object, even if it has some different mocking requirements.

PR #2125, opened 2023-03-01
Report unhandled errors, but only those that are from our extension

robertbrignull:
This is a second attempt at https://github.com/github/vscode-codeql/pull/2076. That PR worked a bit too well, in that it was picking up errors from other extensions and presenting them to the user. The fix for this that we discussed was to check the stack trace to see if it's from our extension or somewhere else.

I've implemented this by checking for `/extensions/ql-vscode/`. I couldn't think of a better string to check for. We might actually have trouble in the future if we move the extension to the top level of the repo. I'm not sure if that means we should look for a different approach now, or save that problem for if/when we do move the code in the repo.

I've tested this using https://github.com/robertbrignull/error-extension. I created a new extension (from a template) that provides commands that intentionally throw errors and reject promises. Using this I was able to reproduce the existing bad behaviour and then also show that we now don't report errors from other extensions.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -72,7 +72,12 @@ import {
   tmpDir,
   tmpDirDisposal,
 } from "./helpers";
-import { asError, assertNever, getErrorMessage } from "./pure/helpers-pure";
+import {
+  asError,
+  assertNever,
+  getErrorMessage,
+  getErrorStack,
+} from "./pure/helpers-pure";
 import { spawnIdeServer } from "./ide-server";
 import { ResultsView } from "./interface";
 import { WebviewReveal } from "./interface-utils";
@@ -235,6 +240,7 @@ export async function activate(
   const distributionConfigListener = new DistributionConfigListener();
   await initializeLogging(ctx);
   await initializeTelemetry(extension, ctx);
+  addUnhandledRejectionListener();
   install();
 
   const codelensProvider = new QuickEvalCodeLensProvider();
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/extension.ts
```diff
@@ -1557,6 +1563,53 @@ async function activateWithInstalledDistribution(
...
+    // report them in our telemetry.
+    //
+    // We check if an error is from this extension by looking for a specific
+    // path in the stack trace.
+    //
+    // The stack trace gets redacted before being sent as telemetry, but at
+    // this point in the code we have the full unredacted information.
+    const isFromThisExtension = getErrorStack(error).includes(
+      "/extensions/ql-vscode/",
+    );
```
koesie10, line 1580: Have you tested this using a packaged extension (VSIX)? I believe the stack trace in the packaged extension looks something like shown in [the source map script](https://github.com/github/vscode-codeql/blob/56d283f6d5f9c0ebd4e0e2857847ccf22549a8ad/extensions/ql-vscode/scripts/source-map.ts):

```
Error: Failed to find CodeQL distribution.
     at CodeQLCliServer.getCodeQlPath (/Users/user/.vscode/extensions/github.vscode-codeql-1.7.8/out/extension.js:131164:13)
     at CodeQLCliServer.launchProcess (/Users/user/.vscode/extensions/github.vscode-codeql-1.7.8/out/extension.js:131169:24)
     at CodeQLCliServer.runCodeQlCliInternal (/Users/user/.vscode/extensions/github.vscode-codeql-1.7.8/out/extension.js:131194:24)
     at CodeQLCliServer.runJsonCodeQlCliCommand (/Users/user/.vscode/extensions/github.vscode-codeql-1.7.8/out/extension.js:131330:20)
     at CodeQLCliServer.resolveRam (/Users/user/.vscode/extensions/github.vscode-codeql-1.7.8/out/extension.js:131455:12)
     at QueryServerClient2.startQueryServerImpl (/Users/user/.vscode/extensions/github.vscode-codeql-1.7.8/out/extension.js:138618:21)
```

This does not include `/extensions/ql-vscode` because when the extension is packaged, that part of the path is not included.

robertbrignull: Good point, I was testing with the development host rather than a packaged extension. If the paths do always look like the ones you've included here, that looks very promising for a robust way to detect extension code.

robertbrignull: I've rewritten this check to use `extension.extensionPath` instead, which will be much more accurate. It'll also avoid issues on windows because I accidentally used forward slashes in the previous implementation.

robertbrignull:
Thanks @koesie10 for the earlier review. It's ready for re-review now.

extensions/ql-vscode/src/extension.ts
```diff
@@ -1569,6 +1575,49 @@ async function activateWithInstalledDistribution(
...
+    // This listener will be triggered for errors from other extensions as
+    // well as errors from this extension. We don't want to flood the user
+    // with popups about errors from other extensions, and we don't want to
+    // report them in our telemetry.
+    //
+    // The stack trace gets redacted before being sent as telemetry, but at
+    // this point in the code we have the full unredacted information.
+    const isFromThisExtension =
+      extension && getErrorStack(error).includes(extension.extensionPath);
+
```
aeisenberg, line 1589: Nice...so this will handle the local development case as well as bundled extensions?

aeisenberg (approved):
Nice...LGTM assuming the answer to my question is "yes".

extensions/ql-vscode/src/extension.ts
robertbrignull: It does still work for local development, because the extension path still matches the location of the installed extension. In my case it's `/Users/robertbrignull/github/vscode-codeql/extensions/ql-vscode`.

koesie10 (approved):
LGTM

PR #2126, opened 2023-03-01
Hack to avoid CodeQL CLI v2.12.3

charisk:
(Paired with @robertbrignull)

CodeQL CLI v2.12.3 was released with a bug that causes the extension to fail so we force the extension to ignore it.

Raising the PR against main for now, but we might need to target a different branch.

The CHANGELOG will need to be updated.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/distribution.ts
+++ b/extensions/ql-vscode/src/distribution.ts
@@ -315,6 +315,15 @@ class ExtensionSpecificDistributionManager {
     const extensionSpecificRelease = this.getInstalledRelease();
     const latestRelease = await this.getLatestRelease();
 
+    // v2.12.3 was released with a bug that causes the extension to fail
+    // so we force the extension to ignore it.
+    if (
+      extensionSpecificRelease &&
+      extensionSpecificRelease.name === "v2.12.3"
+    ) {
+      return createUpdateAvailableResult(latestRelease);
+    }
+
     if (
       extensionSpecificRelease !== undefined &&
       codeQlPath !== undefined &&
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/distribution.ts
```diff
@@ -315,6 +315,15 @@ class ExtensionSpecificDistributionManager {
     const extensionSpecificRelease = this.getInstalledRelease();
     const latestRelease = await this.getLatestRelease();
 
+    // v2.12.3 was released with a bug that causes the extension to fail
+    // so we force the extension to ignore it.
+    if (
+      extensionSpecificRelease &&
+      extensionSpecificRelease.name === "v2.12.3"
```
aeisenberg, line 322: Minor:

```suggestion
      extensionSpecificRelease?.name === "v2.12.3"
```

aeisenberg: It _looks_ like at this point, we are guaranteed that 2.12.2 was not chosen since it is already being filtered out in `this.getLatestRelease()`. Is that true?

charisk: `extensionSpecificRelease` is installed version though I think.

robertbrignull: Do you mean `2.12.2` or `2.12.3`?

I agree it should be the case that `latestRelease` cannot be `2.12.3` since it should be considered incompatible and get filtered out. Even if you're already on `2.12.3` it'll report `2.12.2` as the latest.

aeisenberg: > Do you mean 2.12.2 or 2.12.3?

Yes. :)

aeisenberg (approved):
Happy with the manual testing.

aeisenberg:
This should probably be cherry-picked to a branch based off of the latest release.

PR #2132, opened 2023-03-02
Be able to send params to `installPackDependencies` command

elenatanasoiu:
We'd like to install QL pack dependencies from within a code tour step.

This involves triggering a `codeQL.installPackDependencies` command with a custom directory parameter. 

However, this command doesn't currently support receiving a directory as a param. We need to change this in order to unblock ourselves. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/packaging.ts
+++ b/extensions/ql-vscode/src/packaging.ts
@@ -4,12 +4,13 @@ import {
   showAndLogExceptionWithTelemetry,
   showAndLogInformationMessage,
 } from "./helpers";
-import { QuickPickItem, window } from "vscode";
+import { QuickPickItem, window, workspace } from "vscode";
 import { ProgressCallback, UserCancellationException } from "./commandRunner";
 import { extLogger } from "./common";
 import { asError, getErrorStack } from "./pure/helpers-pure";
 import { redactableError } from "./pure/errors";
 import { PACKS_BY_QUERY_LANGUAGE } from "./common/query-language";
+import { join, resolve } from "path";
 
 /**
  * Prompts user to choose packs to download, and downloads them.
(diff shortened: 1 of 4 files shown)


elenatanasoiu:
It appears I broke some tests 👀  Moving into draft.

extensions/ql-vscode/src/packaging.ts
```diff
@@ -4,12 +4,13 @@
   showAndLogExceptionWithTelemetry,
   showAndLogInformationMessage,
 } from "./helpers";
-import { QuickPickItem, window } from "vscode";
+import { QuickPickItem, Uri, window, workspace } from "vscode";
```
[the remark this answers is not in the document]
elenatanasoiu: Good bot, I can see that in the test failure.

elenatanasoiu:
I accidentally committed a change to set `process.env.TEST_CODEQL_PATH` to my local copy of codeql so I can run tests locally (different problem, env vars aren't loading for me locally). 

I then uncommitted, but I think the machines that previously ran my tests haven't wiped the env var change so they're blindly looking for a local copy of codeql. 

I'm hoping if I push a few more times they'll respawn so I can set the PR to ready for review.

koesie10:
> I accidentally committed a change to set `process.env.TEST_CODEQL_PATH` to my local copy of codeql so I can run tests locally (env vars aren't being set correctly for me).
> 
> I then uncommitted, but I think the machines that previously ran my tests haven't wiped the env var change so they're blindly looking for a local copy of codeql.
> 
> I'm hoping if I push a few more times they'll respawn so I can set the PR to ready for review.

I think the problem is that you're trying to spy on `cli` [here](https://github.com/github/vscode-codeql/blob/f39d29ec50e1c537902282e56b0dd773bb64260f/extensions/ql-vscode/test/vscode-tests/cli-integration/packaging.test.ts#L43-L45), while `cli` is only set [here](https://github.com/github/vscode-codeql/blob/f39d29ec50e1c537902282e56b0dd773bb64260f/extensions/ql-vscode/test/vscode-tests/cli-integration/packaging.test.ts#L53). So, you're essentially calling `jest.spyOn(undefined)` which it (understandably) doesn't accept. Does it work if you move the `packInstallSpy = ` to be below `cli = `?

elenatanasoiu:
Thanks @koesie10, I've pushed the change up to see if it works as I'm still unable to run tests locally at the moment and don't have time to debug today. 

(Maybe I need to add codeql to my PATH?)

elenatanasoiu:
Cool, I can see real test failures now, not "Ensure codeql is instaleld". Will fix up a bit later.

extensions/ql-vscode/src/packaging.ts
```diff
@@ -82,26 +83,44 @@ interface QLPackQuickPickItem extends QuickPickItem {
...
-    placeHolder: "Select packs to install dependencies for",
-    canPickMany: true,
-    ignoreFocusOut: true,
-  });
+
+  let packsToInstall: QLPackQuickPickItem[] | undefined;
+
+  if (relativeDirectoryPath) {
+    const dirPath = resolve(
+      join(workspace.rootPath || "", relativeDirectoryPath),
```
aeisenberg, line 98: There are some edge cases that I am not sure are covered here. 

If this is a multi-root workspace, what is the value of `workspace.rootPath`? What value should we even use? If `workspace.rootPath` is empty or undefined, does that necessarily mean that `relativeDirectoryPath` is absolute?

extensions/ql-vscode/src/packaging.ts
```diff
@@ -82,26 +83,44 @@ interface QLPackQuickPickItem extends QuickPickItem {
...
+      {
+        label: relativeDirectoryPath,
+        packRootDir: [dirPath],
+      },
+    ];
+  } else {
+    const workspacePacks = await cliServer.resolveQlpacks(
+      getOnDiskWorkspaceFolders(),
+    );
+    const quickPickItems = Object.entries(
```
aeisenberg, line 111: In the case of a code tour, do we want to install the packs automatically without a popup?

elenatanasoiu:
Paired with @shati-patel on installing pack dependencies during the prebuild stage (branch [here](https://github.com/github/codespaces-codeql/compare/prebuild#diff-3fc721e0ee2115f6c356be40355b519376b409cecf998cfc117b6e077b2ce02dR12)).

It seemed to work fine: https://github.com/github/codespaces-codeql/actions/runs/4324410039/jobs/7549200793#step:4:272 and I can see the resulting `codeql-pack.lock.yml` file when I fire up a codespace on that branch so this all works!

Happily, the prebuild step also makes this PR redundant because we don't need to call the `installPackDependencies` command anymore from the tour.

I could spend more time fixing this up, but I'd rather do the rest of the changes we need to get the tour into a good state. 

Thanks for reviewing @koesie10 & @aeisenberg !

PR #2135, opened 2023-03-03
Give the LanguageClient an easier id to use.

alexet:
The default id is the lowerase of the name. This is particually wordy and looks wierd.

This is only used for the lsp internal settings root which will now be things like `codeQL.lsp.trace.server`,  rather than `codeql langauge server.trace.server`. 

All these intenral settings are undocumented and for debugging on our side so changing them shouldn't matter.


aeisenberg (approved):
Sounds reasonable.

PR #2139, opened 2023-03-06
Add listeners for unhandled errors to web views

robertbrignull:
Listens for unhandled errors and rejected promises in the webviews. We then send a message with info on the error so we can display the error and potentially log telemetry on it.

There's a few points worth noting here:
- Each webview is isolated from others. If I register these listeners in one webview, they do not pick up errors from other views.
- Including an `Error` object in a message is problematic. It just doesn't seem to work. And it's not like it just loses the prototype bits but keeps the `message` and `stack` fields. Everything gets a bit messed up and you end up with an empty object on the other end. To work around this I've made it send over an error-like object in the message, and I've made the logging and telemetry code handle this type just like it does `Error` objects.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/pure/errors.ts
+++ b/extensions/ql-vscode/src/pure/errors.ts
@@ -1,6 +1,6 @@
 export class RedactableError extends Error {
   constructor(
-    cause: Error | undefined,
+    cause: ErrorLike | undefined,
     private readonly strings: TemplateStringsArray,
     private readonly values: unknown[],
   ) {
@@ -54,19 +54,35 @@ export function redactableError(
   ...values: unknown[]
 ): RedactableError;
 export function redactableError(
-  error: Error,
+  error: ErrorLike,
 ): (strings: TemplateStringsArray, ...values: unknown[]) => RedactableError;
 
 export function redactableError(
-  errorOrStrings: Error | TemplateStringsArray,
+  errorOrStrings: ErrorLike | TemplateStringsArray,
   ...values: unknown[]
 ):
   | ((strings: TemplateStringsArray, ...values: unknown[]) => RedactableError)
   | RedactableError {
-  if (errorOrStrings instanceof Error) {
+  if (isErrorLike(errorOrStrings)) {
     return (strings: TemplateStringsArray, ...values: unknown[]) =>
       new RedactableError(errorOrStrings, strings, values);
   } else {
     return new RedactableError(undefined, errorOrStrings, values);
   }
+}
+
+export interface ErrorLike {
+  message: string;
+  stack?: string;
+}
+
+function isErrorLike(error: any): error is ErrorLike {
+  if (
+    error.message !== undefined &&
+    typeof error.message === "string" &&
+    (error.stack === undefined || typeof error.stack === "string")
+  ) {
+    return true;
+  }
+  return false;
 }
--- a/extensions/ql-vscode/src/pure/interface-types.ts
+++ b/extensions/ql-vscode/src/pure/interface-types.ts
@@ -12,6 +12,7 @@ import {
   VariantAnalysisScannedRepositoryState,
 } from "../variant-analysis/shared/variant-analysis";
 import { RepositoriesFilterSortStateWithIds } from "./variant-analysis-filter-sort";
+import { ErrorLike } from "./errors";
 
 /**
  * This module contains types and code that are shared between
@@ -189,7 +190,8 @@ export type FromResultsViewMsg =
   | ViewLoadedMsg
   | ChangePage
   | OpenFileMsg
-  | TelemetryMessage;
+  | TelemetryMessage
+  | UnhandledErrorMessage;
 
 /**
  * Message from the results view to open a database source
@@ -291,7 +293,8 @@ export type FromCompareViewMessage =
   | ChangeCompareMessage
   | ViewSourceFileMsg
   | OpenQueryMessage
-  | TelemetryMessage;
+  | TelemetryMessage
+  | UnhandledErrorMessage;
 
 /**
  * Message from the compare view to request opening a query.
@@ -439,6 +442,11 @@ export interface TelemetryMessage {
   action: string;
 }
 
+export interface UnhandledErrorMessage {
+  t: "unhandledError";
+  error: ErrorLike;
+}
+
 export type ToVariantAnalysisMessage =
   | SetVariantAnalysisMessage
   | SetRepoResultsMessage
@@ -453,4 +461,5 @@ export type FromVariantAnalysisMessage =
   | ExportResultsMessage
   | OpenLogsMessage
   | CancelVariantAnalysisMessage
-  | TelemetryMessage;
+  | TelemetryMessage
+  | UnhandledErrorMessage;
--- a/extensions/ql-vscode/src/view/common/errors.ts
+++ b/extensions/ql-vscode/src/view/common/errors.ts
@@ -1,0 +1,51 @@
+import { useEffect } from "react";
+import { getErrorMessage, getErrorStack } from "../../pure/helpers-pure";
+import { vscode } from "../vscode-api";
+
+const unhandledErrorListener = (event: ErrorEvent) => {
+  vscode.postMessage({
+    t: "unhandledError",
+    error: {
+      message: getErrorMessage(event.error),
+      stack: getErrorStack(event.error),
+    },
+  });
+};
+
+const unhandledRejectionListener = (event: PromiseRejectionEvent) => {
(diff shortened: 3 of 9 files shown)


robertbrignull:
There's also a bug I'm still working out, but I decided to open the PR anyway while I look into it. We're getting duplicate `window.addEventListener("error"` events. It just seems to fire twice, and I'm not sure why. This doesn't happen with the unhandled rejection listener, and I don't believe it's because of different webviews interfering with each other.

robertbrignull:
Seems the error firing twice is due to react error boundaries. Possibly due to the default error boundary re-throwing. See https://github.com/facebook/react/issues/11499

I'm going to try implementing an error boundary and see how it behaves. That will only affect errors thrown when rendering. From reading the issue it seems we could still end up with duplicate errors from event handlers. We can decide if that's acceptable, or whether to try fixing it by remembering which errors we've seen previously.

It doesn't appear to show a duplicate popup, so the user won't notice that there are two errors. It'll only show up in our telemetry and make the numbers a bit inflated.

extensions/ql-vscode/src/pure/errors.ts
```diff
@@ -54,19 +54,35 @@ export function redactableError(
...
+
+export interface ErrorLike {
+  message: string;
+  stack?: string;
+}
+
+function isErrorLike(error: any): error is ErrorLike {
+  if (
+    error.message !== undefined &&
+    typeof error.message === "string" &&
```
aeisenberg, line 82: Minor:

```suggestion
    typeof error?.message === "string" &&
```

extensions/ql-vscode/src/pure/errors.ts
```diff
@@ -54,19 +54,35 @@ export function redactableError(
...
+    (error.stack === undefined || typeof error.stack === "string")
```
aeisenberg, line 83:
```suggestion
    (error.stack === "undefined" || typeof error.stack === "string")
```

extensions/ql-vscode/src/view/common/errors.ts
```diff
@@ -0,0 +1,31 @@
+import { getErrorMessage, getErrorStack } from "../../pure/helpers-pure";
+import { vscode } from "../vscode-api";
+
+const unhandledErrorListener = (event: ErrorEvent) => {
+  vscode.postMessage({
```
aeisenberg, line 5: Does the stack checking you implemented earlier to avoid printing stack traces for not-our-code also work for the webview? This is something to verify at least.

extensions/ql-vscode/src/pure/errors.ts
robertbrignull: Thinking about this more, I think we can simplify this to just `typeof error.message === "string"` and the `?` is unnecessary. That question mark is purely a typescript thing and is removed when transpiling to javascript, but the type of `error` here is `any` so there's no typescript type checking anyway. So we can just remove it. You can safely do `typeof undefined` and it'll return `"undefined"`.

extensions/ql-vscode/src/view/common/errors.ts
robertbrignull: I'll try to do a test with another extension to make sure, but I'm currently pretty confident that webviews are isolated and so the errors from one are not reported in another. I say this because the various webviews in our application are not interfering with each other. We get errors from a webview if and only if a listener has been registered in that particular view, and having another view open with a listener registered will not make it report errors from other open views.

robertbrignull:
I've pushed some new changes which include code for remembering which errors we're seen previously and avoiding reporting telemetry for them again.

Please let me know what you think about the approach or remember previous errors. I haven't seen a better way of avoiding the duplicate reporting, but I'm very open to other options.

I've done some investigation of using an error boundary, and I found that it did indeed stop the duplicate error reporting for errors during rendering but not for errors from event handlers. This is as we thought it would be, but it's nice to confirm it. I think error boundaries can be considered separately from this PR so I've opened an issue to track them.

extensions/ql-vscode/src/view/common/errors.ts
```diff
@@ -0,0 +1,54 @@
...
+      error: {
+        message: getErrorMessage(event.error),
+        stack: getErrorStack(event.error),
+      },
+    });
+  }
+};
+
+const unhandledRejectionListener = (event: PromiseRejectionEvent) => {
+  if (shouldReportError(event.reason)) {
```
robertbrignull, line 36: I'm not 100% sure but I think the problem of duplicate errors only happens for thrown errors and not rejected promises. So this check here may be unnecessary, but it also shouldn't hurt so I decided to keep it in. I'm not really sure about this though and I'm happy to remove it.

extensions/ql-vscode/src/pure/errors.ts
```diff
@@ -54,19 +54,34 @@ export function redactableError(
...
+
+export interface ErrorLike {
+  message: string;
+  stack?: string;
+}
+
+function isErrorLike(error: any): error is ErrorLike {
+  if (
+    typeof error.message === "string" &&
+    (typeof error.stack === "undefined" || typeof error.stack === "string")
```
koesie10, line 82: I'm not sure what the `typeof` for `undefined` adds here. Would this be equivalent?

```suggestion
    (error.stack === undefined || typeof error.stack === "string")
```

extensions/ql-vscode/src/pure/interface-types.ts
```diff
@@ -189,7 +190,8 @@ export type FromResultsViewMsg =
   | ViewLoadedMsg
   | ChangePage
   | OpenFileMsg
-  | TelemetryMessage;
+  | TelemetryMessage
+  | UnhandledErrorMessage;
```
koesie10, line 194: Should we perhaps be creating a separate type which contains all "common" messages. Currently this would be the `ViewLoadedMsg`, `TelemetryMessage`, and `UnhandledErrorMessage`. If we add a new view, these all need to be included, otherwise the extension might crash if any of these messages are sent anyway.

extensions/ql-vscode/src/view/common/errors.ts
```diff
@@ -0,0 +1,54 @@
...
+      t: "unhandledError",
+      error: {
+        message: getErrorMessage(event.error),
+        stack: getErrorStack(event.error),
+      },
+    });
+  }
+};
+
+const unhandledRejectionListener = (event: PromiseRejectionEvent) => {
```
koesie10, line 35: Should we be making a distinction between the unhandled errors and unhandled rejections? As far as I am aware, unhandled rejections don't cause any user-visible behaviour (except perhaps some data not loading), but unhandled errors might cause the complete view to disappear. It would be useful to make that distinction in the telemetry.

koesie10 (approved):
I think this approach makes sense, and if we can't filter out duplicate errors for event handlers anyway, then I think this behaviour is exactly what we need.

I have a few minor suggestions, but nothing blocking.

extensions/ql-vscode/src/pure/errors.ts
robertbrignull: I don't think it adds anything except then all the checks are "typeof". But I agree just checking directly against `undefined` is equivalent and more understandable. I added this earlier because of a suggestion from Andrew but I think I may have misunderstood what he was saying.

I'm pretty confident what you're suggesting here is correct and it's what I had originally in this PR, so I'll just go with this.

extensions/ql-vscode/src/pure/interface-types.ts
robertbrignull: Sounds good to me. Making a type for all the "command" messages should be fine, and then we include that type in all the types like `FromVariantAnalysisMessage`. Earlier when I first started this PR I tried to get `AbstractWebview` to handle these common messages automatically, but that proved too difficult and I couldn't get the types to work out. So each view still has to implement these messages and remember to include them in it's messages type, but at least by grouping them together it should be harder to miss one.

extensions/ql-vscode/src/view/common/errors.ts
robertbrignull: For now I'm going to leave them combined. I don't think it's quite as well defined as saying unhandled errors are always more important. Missing or incorrect data is also bad, and maybe worse if the user don't know the data they're looking at is incorrect. Unhandled errors can also happen outside of rendering.

Also, we can certainly split these up in the future, and doing it after this PR will be fine. The telemetry is quite schema-less and we only keep it for a short period, so changing the format is quite cheap to do.

robertbrignull:
I've done some more tests and this seems to be working correctly, so I think this is good to merge.

PR #2140, opened 2023-03-07
Save scroll position before opening code paths

charisk:
This is so that the user doesn't lose their place in the results view.

https://user-images.githubusercontent.com/311693/212700041-573ea61d-582b-4a76-b07d-3237e957f74c.mov

We do this by storing the scroll position in state when the show paths link gets clicked.

A [previous PR](https://github.com/github/vscode-codeql/pull/1970) attempted the same approach, but was setting the initial scroll position to (0,0) instead of undefined, so was causing a different bug.

Note that this is meant to be a quick workaround for the issue that has been reported by users. There are a few other issues around the area, but those won't be addressed by this PR. We're looking to change the UI/flow a bit anyway, so I don't want to spend too much time on this.  

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/view/common/CodePaths/CodePaths.tsx
+++ b/extensions/ql-vscode/src/view/common/CodePaths/CodePaths.tsx
@@ -37,13 +37,24 @@ export const CodePaths = ({
     filterTelemetryOnValue: filterIsOpenTelemetry,
   });
 
+  const [scrollPosition, setScrollPosition] = useState({ x: 0, y: 0 });
+  React.useEffect(() => {
+    window.scrollTo(scrollPosition.x, scrollPosition.y);
+  }, [scrollPosition]);
+
   const linkRef = useRef<HTMLAnchorElement>(null);
 
   const closeOverlay = () => setIsOpen(false);
 
   return (
     <>
-      <ShowPathsLink onClick={() => setIsOpen(true)} ref={linkRef}>
+      <ShowPathsLink
+        onClick={() => {
+          setScrollPosition({ x: window.scrollX, y: window.scrollY });
+          setIsOpen(true);
+        }}
+        ref={linkRef}
+      >
         Show paths
       </ShowPathsLink>
       {isOpen && (
--- a/extensions/ql-vscode/src/view/common/CodePaths/CodePaths.tsx
+++ b/extensions/ql-vscode/src/view/common/CodePaths/CodePaths.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { useRef, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import styled from "styled-components";
 import { VSCodeLink } from "@vscode/webview-ui-toolkit/react";
 
@@ -24,6 +24,11 @@ export type CodePathsProps = {
   severity: ResultSeverity;
 };
 
+type ScrollPosition = {
+  x: number;
+  y: number;
+};
+
 const filterIsOpenTelemetry = (v: boolean) => v;
 
 export const CodePaths = ({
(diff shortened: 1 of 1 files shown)


robertbrignull:
The behaviour is better, but there's still some oddities. I'll try to explain.

It correctly keeps the scroll positions of the results window while the paths popup is open. This is good. What's a little strange is that while the paths popup is open, the results window scroll bar is still shown and can still be manipulated. If you try to scroll while the paths popup is open, the behaviour is:
- If the paths popup can be scrolled then it'll scroll that. But there is no visible scroll bar for this popup. The scroll bar shown on the right is for the results windows and it doesn't move when scrolling the paths popup.
- If the paths pop can't be scrolled, either because it doesn't fill the window or it's already at the top/botton, then it'll scroll the hidden results window instead. The scroll bar on the right will move but nothing is visibly scrolling on screen, until you close the paths popup and realise that the results window is not how you left it.

Can we give the paths popup it's own visible scroll bar, and make it not possible the scroll the results window while the paths popup is open?

---

Another point I want to ask about it why does it appear that we call `window.scrollTo` when the paths popup is opened, rather than when it's closed? I'm not 100% saying this is wrong, but I'd like an explanation because it's not how I would expect. If the idea is to put the results window back how it was when the user last saw it, shouldn't we be setting the scroll value when the paths popup is closed? It appears like we're just setting the scroll value to the value it already has. Maybe I'm misunderstanding when the `useEffect` block executes. I'm just a little confused, but very happy to have it explained.

extensions/ql-vscode/src/view/common/CodePaths/CodePaths.tsx
```diff
@@ -24,6 +24,11 @@ export type CodePathsProps = {
   severity: ResultSeverity;
 };
 
+type ScollPosition = {
```
iamgiolaga, line 27:
```suggestion
type ScrollPosition = {
```
hey, just here to dome some open source, little typo :)

extensions/ql-vscode/src/view/common/CodePaths/CodePaths.tsx
```diff
@@ -37,13 +42,28 @@ export const CodePaths = ({
     filterTelemetryOnValue: filterIsOpenTelemetry,
   });
 
+  const [scrollPosition, setScrollPosition] = useState<
+    ScollPosition | undefined
+  >(undefined);
+  React.useEffect(() => {
```
iamgiolaga, line 48:
```suggestion
useEffect(() => {
```
and importing it from above, wdyt?

iamgiolaga: just an extra consideration from an external point of view: is the `useEffect` what you really need? In general I would use it if I need to run it when refreshing a page for example but here in vscode I don't know if the same concept exists and triggers the same.
If that's not the case, wouldn't be more suitable to use `useCallback` and call it `onClick` from below?

charisk: I believe `useEffect` is the right thing, since scroll position changes are a side effect of the component.

charisk:
> The behaviour is better, but there's still some oddities. I'll try to explain.

> It correctly keeps the scroll positions of the results window while the paths popup is open. This is good. What's a little strange is that while the paths popup is open, the results window scroll bar is still shown and can still be manipulated. If you try to scroll while the paths popup is open, the behaviour is:

> If the paths popup can be scrolled then it'll scroll that. But there is no visible scroll bar for this popup. The scroll bar shown on the right is for the results windows and it doesn't move when scrolling the paths popup.
If the paths pop can't be scrolled, either because it doesn't fill the window or it's already at the top/botton, then it'll scroll the hidden results window instead. The scroll bar on the right will move but nothing is visibly scrolling on screen, until you close the paths popup and realise that the results window is not how you left it.
Can we give the paths popup it's own visible scroll bar, and make it not possible the scroll the results window while the paths popup is open?

Discussed offline, these are existing issues that we won't be trying to fix just yet. See PR description for more details.

PR #2150, opened 2023-03-08
Do not copy `node_modules` when packaging

koesie10:
We are no longer including our dependencies in the VSIX package, so we can tell VSCE that we don't want it to look at dependencies using `--no-dependencies`. If we do this, VSCE doesn't require the `node_modules` directory anymore and we can skip that step, which will make building significantly faster.

I've confirmed that there are no changes between the two options by building the extension both without and with the change. This is the diff of the two outputs (using `diff -r`):

```diff
diff --color -r vscode-codeql-1.8.0-dev.2023.3.8.15.10.13/extension/package.json vscode-codeql-old/extension/package.json
7c7
<   "version": "1.8.0-dev.2023.3.8.15.10.13",
---
>   "version": "1.8.0-dev.2023.3.8.15.6.51",
diff --color -r vscode-codeql-1.8.0-dev.2023.3.8.15.10.13/extension.vsixmanifest vscode-codeql-old/extension.vsixmanifest
4c4
< 			<Identity Language="en-US" Id="vscode-codeql" Version="1.8.0-dev.2023.3.8.15.10.13" Publisher="GitHub" />
---
> 			<Identity Language="en-US" Id="vscode-codeql" Version="1.8.0-dev.2023.3.8.15.6.51" Publisher="GitHub" />
```

The only difference is the version number, which is expected.

## Checklist


extensions/ql-vscode/gulpfile.ts/package.ts
```diff
@@ -17,6 +17,7 @@ export async function packageExtension(): Promise<void> {
       "..",
       `${deployedPackage.name}-${deployedPackage.version}.vsix`,
     ),
+    "--no-dependencies",
```
aeisenberg, line 20: The docs are a little unclear as to what this option does. Does this mean that `node_modules` will be ignored when packaging the extension? And are we able to do this because we are now using esbuild, which does the bundling for us?

aeisenberg (commented):
Have you tried this out on a dev build to make sure it is still installable?

extensions/ql-vscode/gulpfile.ts/package.ts
aeisenberg: Thanks for the explanation.

aeisenberg (approved):

PR #2153, opened 2023-03-09
Remove unnecessary commands from `activationEvents`

norascheuch:
VSCode Docs: https://code.visualstudio.com/api/references/activation-events

From the docs: `Your extension becomes activated when the Activation Event happens.`. This means only commands that can be triggered from the command palette need to be listed in the `activationEvents`. All other events can only be executed when the extension is already activated.

## Checklist


robertbrignull:
We might also be able to remove `codeQLDatabases.chooseDatabase`. As far as I can tell that command doesn't exist. There a few `codeQLDatabases.chooseDatabaseX` for where X is Folder, Achive, Internet, Github. Those commands are also listed and marked as unnecessary, so it's not like there's anything special going on because it's a prefix. I think this is leftover from the command getting renamed or split up and now we can remove it.

extensions/ql-vscode/package.json
```diff
@@ -58,15 +53,6 @@
     "onCommand:codeQL.chooseDatabaseInternet",
     "onCommand:codeQL.chooseDatabaseGithub",
     "onCommand:codeQLDatabases.chooseDatabase",
```
shati-patel, line 55: Agree with @robertbrignull that `onCommand:codeQLDatabases.chooseDatabase` can be removed, since it doesn't exist 😄 

```suggestion
    "onCommand:codeQL.chooseDatabaseGithub",
```

shati-patel (approved):
Just one comment, otherwise LGTM!

--- 

Re: 
> According to the linting in my VS Code we can remove quite a few more. I see the ones currently included in this PR are just the ones related to variant analysis.
> 
> I think we can remove all of the indicated events? What do you think? Does your VS Code also give these warnings.

This is new from VS Code version 1.74.0, so let's not remove these yet (until we set that as our [earliest supported VS Code version](https://github.com/github/vscode-codeql/blob/2bd9c8a732fa1b199763ebf1b6401919f2297830/extensions/ql-vscode/package.json#L16)). We have an internal issue tracking this 💪🏽

robertbrignull:
I bow to @shati-patel here 🙇  so please go with her review.

extensions/ql-vscode/package.json
norascheuch: Yes, good catch @robertbrignull! :)

PR #2157, opened 2023-03-09
Commands registration

starcke:
This adds a command manager, which will enable us to make command execution well-typed.

The command manager itself is generic, and is mostly a wrapper around the types.

To implement the typed commands we introduce an explicit typing layer (in `extensions/ql-vscode/src/common/commands.ts`) and separate registration/implementation layer (`getCommands` and `allCommands` methods). This split was necessary to avoid dependency cycles where command execution would need to know all types possible (and if not split all implementations).

We put the command manager into the app for ease of use.

Rest of the changes should be fairly straightforward.

## Checklist


extensions/ql-vscode/src/common/commands.ts
```diff
@@ -0,0 +1,11 @@
+import { CommandManager } from "../packages/commands";
+
+export type ExtensionCommands = {
+  "codeQL.openVariantAnalysisLogs": (
+    variantAnalysisId: number,
+  ) => Promise<void>;
+};
+
+export type AllCommands = ExtensionCommands;
```
elenatanasoiu: This might be obvious, but what other types of commands besides variant analysis commands do we think we'd need? Should `ExtensionCommands` eventually have a less general name or do we intend to get rid of it?

extensions/ql-vscode/src/common/vscode/vscode-app.ts
```diff
@@ -6,14 +6,19 @@ import { AppEventEmitter } from "../events";
 import { extLogger, Logger } from "../logging";
 import { Memento } from "../memento";
 import { VSCodeAppEventEmitter } from "./events";
+import { ExtensionCommandManager } from "../commands";
+import { initializeVSCodeCommandManager } from "./commands";
 
 export class ExtensionApp implements App {
```
elenatanasoiu, line 12: Unrelated to this PR, but I wonder why we call this `ExtensionApp` when the file is called `vscode-app`.

extensions/ql-vscode/src/packages/commands/CommandManager.ts
```diff
@@ -1,9 +1,25 @@
...
+/**
+ * A command function is a completely untyped command.
+ */
 export type CommandFunction = (...args: any[]) => Promise<unknown>;
 
+/**
+ * The command manager basically takes a single input, the type
+ * of all the known commands. The second parameter is provided by
+ * default (and should not be needed by the caller) it is a
+ * technicality to allow the type system to look up commands.
```
elenatanasoiu, line 21: Thanks for clarifying 👍

elenatanasoiu (commented):
Left some comments, mostly for my own understanding. Great job at making this look clear!

extensions/ql-vscode/test/unit-tests/packages/commands/CommandManager.test.ts
```diff
@@ -1,8 +1,111 @@
...
+        return variantAnalysisId * 10;
+      },
+    };
+    const commandManager = new CommandManager<typeof commands>(
+      jest.fn(),
+      jest.fn(),
+    );
+
+    // @ts-expect-error wrong command name should give a type error
+    commandManager.registerCommand("abc", jest.fn());
```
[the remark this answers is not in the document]
starcke: It is a good question. I wasnt sure, but given how simple the test is I thought it actually a quite neat way of making sure it actually behaves as expected. E.g. if we accidentally had an `any` somewhere in the implementation, then it might suddenly not give us the safety we expect.

extensions/ql-vscode/src/extension.ts
```diff
@@ -1161,6 +1177,17 @@ async function activateWithInstalledDistribution(
       },
     ),
   );
+*/
+
+  const vsCommandRunner = initializeVSCodeCommandManager<AllCommands>();
```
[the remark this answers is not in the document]
starcke: I think the trouble we had without the wrapper function was that typescript wasnt clever enough to see that `ReturnType<Commands[CommandName]>` was actually a `Promise`, that is why we do the wrapping and unwrapping of the type here: `Promise<Awaited<ReturnType<Commands[CommandName]>>>`.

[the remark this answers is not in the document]
elenatanasoiu: Right, this is ... hard core.

extensions/ql-vscode/test/unit-tests/packages/commands/CommandManager.test.ts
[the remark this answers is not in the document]
elenatanasoiu: Yeah I guess if the point is for us to assure ourselves that our typing works as we expect it to, then it makes sense. But I would argue if we had to add tests on top of the type system, then why do we need the type system in the first place. It seems like tests are doing the validation for us. 

But this is getting philosophical, I still think what you're doing makes sense. It's just annoying that we can't trust it completely.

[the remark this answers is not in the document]
koesie10: I don't think this is really about testing the type system. We're basically testing that the types that we have defined are working and correctly give errors where we expect them to. This basically ensure that if we change the first parameter of `registerCommand` to be `string` instead of `CommandName`, we get this error. I don't think there's another way to test this, but we would like to have that assurance.

extensions/ql-vscode/src/common/commands.ts
koesie10: I think we could have at least the following command "groups":
- Variant analysis
- New database panel
- Old database panel
- Query history
- Local query
- Any other commands, like showing documentation

I think the last group would be the `ExtensionCommands`. I think we'd need to see which commands end up in that group to give it a better name.

extensions/ql-vscode/src/packages/commands/CommandManager.ts
koesie10: I think it's better to leave it as-is. It's not unused, it's just an optional parameter. In function arguments, we would also only add underscores if the argument is unused, rather than optional.

elenatanasoiu (approved):

extensions/ql-vscode/src/common/vscode/commands.ts
```diff
@@ -0,0 +1,25 @@
+import { commands } from "vscode";
+import { commandRunner } from "../../commandRunner";
+import { CommandFunction, CommandManager } from "../../packages/commands";
+
+/**
+ * Intializes a command manager for VSCode, wrapping the commandRunner
+ * and vscode.executeCommand.
+ */
+export function initializeVSCodeCommandManager<
```
charisk, line 9: (Minor and optional) `initialize` makes me think that there is some work to be done to initialize something, but I think that here we're just creating an object. Would `createVSCodeCommandManager` be a better name?

extensions/ql-vscode/src/common/vscode/commands.ts
```diff
@@ -0,0 +1,25 @@
...
+  Commands extends Record<string, CommandFunction>,
+>(): CommandManager<Commands> {
+  return new CommandManager(commandRunner, wrappedExecuteCommand);
+}
+
+async function wrappedExecuteCommand<
```
charisk, line 15: Why are we wrapping this? Is it just to satisfy TS and set the expected types? Might be worth adding a code comment to explain.

Also should it be named `wrapExecuteCommand` ?

extensions/ql-vscode/src/packages/commands/CommandManager.ts
```diff
@@ -1 +1,72 @@
...
+ */
+export class CommandManager<
+  Commands extends Record<string, CommandFunction>,
+  CommandName extends keyof Commands & string = keyof Commands & string,
+> implements Disposable
+{
+  // TODO: should this be a map?
```
charisk, line 28: Did you mean to leave this TODOs in?

extensions/ql-vscode/src/packages/commands/CommandManager.ts
```diff
@@ -1 +1,72 @@
...
+  // TODO: handle multiple command names
+  private commands: Disposable[] = [];
+
+  constructor(
+    private readonly commandRegister: <T extends CommandName>(
+      commandName: T,
+      definition: Commands[T],
```
charisk, line 35: Is this the "command function". I'm wondering if the terms are the same and whether we could just use a single term.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -123,6 +127,18 @@ export class VariantAnalysisManager
...
 
+  getCommands(): VariantAnalysisCommands {
+    return {
+      "codeQL.openVariantAnalysisLogs": async (variantAnalysisId: number) => {
+        await this.openVariantAnalysisLogs(variantAnalysisId);
+      },
+    };
+  }
+
+  get commandManager(): ExtensionCommandManager {
```
charisk, line 138: Can you explain why this is needed?

charisk (commented):
This is looking really great, and the sort of thing I had in mind! Great work!

I have a few comments but they're mainly around naming. I'm being extra picky just because this is something that will be so core to the overall codebase.

charisk:
One more thing to consider around naming (and this can perhaps be separate to the PR): should we abbreviate Command to `Cmd`? I'm not generally a fan of abbreviations but for standard things like this, I think it could make sense.

extensions/ql-vscode/src/common/commands.ts
```diff
@@ -0,0 +1,24 @@
...
+// Commands tied to variant analysis
+export type VariantAnalysisCommands = {
+  "codeQL.openVariantAnalysisLogs": (
+    variantAnalysisId: number,
+  ) => Promise<void>;
+};
+
+export type AllCommands = ExtensionCommands & VariantAnalysisCommands;
+
+export type ExtensionCommandManager = CommandManager<AllCommands>;
```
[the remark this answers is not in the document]
starcke: I think I prefer `App` over `Extension` as well, do you have any thoughts on just renaming that @koesie10 ?

extensions/ql-vscode/src/common/app.ts
```diff
@@ -15,6 +16,7 @@ export interface App {
   readonly workspaceStoragePath?: string;
   readonly workspaceState: Memento;
   readonly credentials: Credentials;
+  readonly commandManager: ExtensionCommandManager;
```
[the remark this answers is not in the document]
starcke: I love shorter names coming from go, so :+1: from me.

extensions/ql-vscode/src/common/commands.ts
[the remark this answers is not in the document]
koesie10: I think naming it `App` makes a lot of sense. I'm not sure about renaming `ExtensionCommands` to `AppCommands` though since `ExtensionCommands` doesn't actually contain all commands of the app, it only contains the commands registered in the `extension.ts` file. I think having `AllCommands` as `AppCommands` would make more sense, but I'll leave that up to @starcke to decide which one he prefers.

extensions/ql-vscode/src/packages/commands/CommandManager.ts
```diff
@@ -1 +1,72 @@
-export class CommandManager {}
+/**
+ * Contains a generic implementation of typed commands.
+ *
+ * This allows different parts of the extension to register commands with a certain type,
+ * and then allow other parts to call those commands in a well-typed manner.
+ */
+
+export interface Disposable {
```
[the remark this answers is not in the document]
koesie10: I think moving it to a separate file `Disposable.ts` could make sense. It's intended to duplicate the `vscode.Disposable`, so it's expected that they are basically the same.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
koesie10: This is needed to avoid any direct dependencies from the `VariantAnalysisView` to the `VariantAnalysisManager`; it's the implementation of the `VariantAnalysisViewManager` interface.

extensions/ql-vscode/src/common/commands.ts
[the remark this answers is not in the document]
starcke: `AppCommandManager` is good, but thinking about it I think both `ExtensionCommands` and `AppCommands` are a bit wrong as `VariantAnalysisCommands` are also commands for the extension/app. So I think something like `BaseCommands` for the commands that are not defined in any specific manager (like the variant analysis manager).

extensions/ql-vscode/src/common/vscode/commands.ts
starcke: Yeah the types are tricky to make work without a separate function as the `Promise` part does not behave in the most expected way. Which is also why we need the wrapping and unwrapping of `Promise<Awaited`. I'll add a comment.

> Also should it be named wrapExecuteCommand ?
:+1:

extensions/ql-vscode/src/packages/commands/CommandManager.ts
starcke: I cannot exactly remember why we needed this, @koesie10 do you have an answer here?

extensions/ql-vscode/src/packages/commands/CommandManager.ts
```diff
@@ -1 +1,72 @@
...
+    private readonly commandExecute: <T extends CommandName>(
+      commandName: T,
+      ...args: Parameters<Commands[T]>
+    ) => Promise<Awaited<ReturnType<Commands[T]>>>,
+  ) {}
+
+  /**
+   * Register a command with the specified name and implementation.
+   */
+  registerCommand<T extends CommandName>(
```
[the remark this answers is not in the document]
starcke: Yes! very good idea.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
charisk: Can't the view get the command manager (or the App container) passed in? Sorry if I'm missing something obvious! Also, this is not blocking.

starcke: I think that the variant analysis view only has access to the variant analysis manager and not the app (because the app is private), so it needs this to call the commands.

koesie10: I think we could also pass it in, but as @starcke points out we still need to have some way to get the command manager out of the variant analysis manager (in the [variant analysis view serializer](https://github.com/github/vscode-codeql/blob/39d91788dc00a8ed1f8b6daf011e3991680f2e05/extensions/ql-vscode/src/variant-analysis/variant-analysis-view-serializer.ts)), so we still wouldn't be able to get rid of this method. I'd prefer to keep the variant analysis view serializer as lightweight as possible so that we don't need to change the constructor parameters that often.

extensions/ql-vscode/src/packages/commands/CommandManager.ts
koesie10: Yes, this is the command function. My PoC originally had a way to register commands with progress using an object, so in that case it was really a definition. Now it's just a function, so renaming it to `commandFunction` or even `fn` would make sense to me.

extensions/ql-vscode/test/unit-tests/packages/commands/CommandManager.test.ts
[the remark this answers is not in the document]
elenatanasoiu: Yeah I understand that we want to see that it actually works as expected, which is why I didn't press it further. 

But I'll be honest, it still a bit weird to me to have to test Typescript. We're intentionally disabling typescript errors to make the test work. I understand we could miss something if we redefined the types, but we don't really do this kind of testing pattern for other Typescript constraints, and those could also start failing at any time. 

I would be inclined to trust the type system, the way we do for the rest of the codebase. But I understand if you want extra assurance.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
charisk: Hmm okay. It feels a bit off to me but we can circle back some other time if it actually becomes a problem.

extensions/ql-vscode/src/common/commands.ts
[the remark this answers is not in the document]
charisk: Thanks both!

extensions/ql-vscode/src/packages/commands/CommandManager.ts
charisk: Great! One less term 😄

charisk (approved):
LGTM! Thank you both for the changes!

starcke:
> One more thing to consider around naming (and this can perhaps be separate to the PR): should we abbreviate Command to Cmd? I'm not generally a fan of abbreviations but for standard things like this, I think it could make sense.

We decided to leave this for a followup.

PR #2162, opened 2023-03-09
Add back variant analysis tests

aeisenberg:
This commit adds back the variant analysis tests that were inadvertently deleted after this commit

https://github.com/github/vscode-codeql/blob/82ada54103f591228d9c61ac6b3c93c85d6ffd1c/extensions/ql-vscode/src/vscode-tests/cli-integration/remote-queries/run-remote-query.test.ts#L67

They are not identical to the removed tests. I refactored them so that there is a single test function invoked three times with different parameters.

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
@@ -269,7 +269,6 @@ describe("Variant Analysis Manager", () => {
       filesThatDoNotExist: string[];
     }) {
       const fileUri = getFile(queryPath);
-
       await variantAnalysisManager.runVariantAnalysis(
         fileUri,
         progress,
@@ -291,6 +290,7 @@ describe("Variant Analysis Manager", () => {
       filesThatExist.forEach((file) => {
         expect(packFS.fileExists(file)).toBe(true);
       });
+
       if (await cli.cliConstraints.supportsQlxRemote()) {
         qlxFilesThatExist.forEach((file) => {
           expect(packFS.fileExists(file)).toBe(true);
@@ -313,7 +313,7 @@ describe("Variant Analysis Manager", () => {
       const packFileName = packFS.fileExists("qlpack.yml")
         ? "qlpack.yml"
         : "codeql-pack.yml";
-      const qlpackContents: any = load(
+      const qlpackContents = load(
         packFS.fileContents(packFileName).toString("utf-8"),
       );
       expect(qlpackContents.name).toEqual("codeql-remote/query");
@@ -322,7 +322,7 @@ describe("Variant Analysis Manager", () => {
         "*",
       );
 
-      const qlpackLockContents: any = load(
+      const qlpackLockContents = load(
         packFS.fileContents("codeql-pack.lock.yml").toString("utf-8"),
       );
 
--- a/extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
@@ -19,7 +19,10 @@ import {
   storagePath,
 } from "../../global.helper";
 import { VariantAnalysisResultsManager } from "../../../../src/variant-analysis/variant-analysis-results-manager";
-import { VariantAnalysisStatus } from "../../../../src/variant-analysis/shared/variant-analysis";
+import {
+  VariantAnalysisStatus,
+  VariantAnalysisSubmission,
+} from "../../../../src/variant-analysis/shared/variant-analysis";
 import { VariantAnalysis as VariantAnalysisApiResponse } from "../../../../src/variant-analysis/gh-api/variant-analysis";
 import { createMockApiResponse } from "../../../factories/variant-analysis/gh-api/variant-analysis-api-response";
 import { UserCancellationException } from "../../../../src/commandRunner";
@@ -28,6 +31,9 @@ import { DbManager } from "../../../../src/databases/db-manager";
 import { ExtensionApp } from "../../../../src/common/vscode/vscode-app";
 import { DbConfigStore } from "../../../../src/databases/config/db-config-store";
 import { mockedQuickPickItem } from "../../utils/mocking.helpers";
+import { QueryLanguage } from "../../../../src/common/query-language";
+import { readBundledPack } from "../../utils/bundled-pack-helpers";
+import { load } from "js-yaml";
 
 // up to 3 minutes per test
 jest.setTimeout(3 * 60 * 1000);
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
```diff
@@ -202,5 +204,136 @@ describe("Variant Analysis Manager", () => {
...
+      filesThatExist,
+      qlxFilesThatExist,
+      filesThatDoNotExist,
+    }: {
+      queryPath: string;
+      filesThatExist: string[];
+      qlxFilesThatExist: string[];
+      filesThatDoNotExist: string[];
+    }) {
+      const fileUri = getFile(queryPath);
```
koesie10, line 271: The `getFile` function only seems to be used here, so would this make it clearer?

```suggestion
      const fileUri = Uri.file(join(baseDir, queryPath));
```

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
```diff
@@ -202,5 +204,136 @@ describe("Variant Analysis Manager", () => {
...
+      // depending on the cli version, we should have one of these files
+      expect(
+        packFS.fileExists("qlpack.lock.yml") ||
+          packFS.fileExists("codeql-pack.lock.yml"),
+      ).toBe(true);
+
+      const packFileName = packFS.fileExists("qlpack.yml")
+        ? "qlpack.yml"
+        : "codeql-pack.yml";
+      const qlpackContents: any = load(
```
koesie10, line 316: Do we need these `any` type annotations or is there a reason you've added them?

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
```diff
@@ -202,5 +204,136 @@ describe("Variant Analysis Manager", () => {
...
+        await doVariantAnalysisTest({
+          queryPath: "data-remote-qlpack-nested/subfolder/in-pack.ql",
+          filesThatExist: ["subfolder/in-pack.ql", "otherfolder/lib.qll"],
+          filesThatDoNotExist: ["subfolder/not-in-pack.ql"],
+          qlxFilesThatExist: ["subfolder/in-pack.qlx"],
+        });
+      });
+    });
+
+    async function doVariantAnalysisTest({
```
koesie10, line 260: An alternative for creating a function for this would be to use [`it.each`](https://jestjs.io/docs/api#each). That would avoid the need to create a function, which might make the tests easier to read (since all expectations are located within the `it`).

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
```diff
@@ -202,5 +204,136 @@ describe("Variant Analysis Manager", () => {
...
+      filesThatDoNotExist,
+    }: {
+      queryPath: string;
+      filesThatExist: string[];
+      qlxFilesThatExist: string[];
+      filesThatDoNotExist: string[];
+    }) {
+      const fileUri = getFile(queryPath);
+
+      await variantAnalysisManager.runVariantAnalysis(
```
koesie10, line 273: Since this is testing the query pack generation instead of specifically the variant analysis, would it make sense to make these tests of a new test suite for [`run-remote-query.ts`](https://github.com/github/vscode-codeql/blob/6f37f176e4a6aa474d2574ff632b3f5f2828da20/extensions/ql-vscode/src/variant-analysis/run-remote-query.ts) which tests the `prepareRemoteQueryRun` function? I think that makes the test easier since you don't need to mock anything, although you'll lose some of the end-to-end testing that we're doing now.

koesie10 (commented):
Thanks for working on this!

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
aeisenberg: This was from the previous tests. I can see if I can get rid of them.

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
aeisenberg: This might work, but I'll have to see since in a follow-up PR I'm adding a new test that requires a bit of work before running the assertions. 

Actually...I think I can get this done by passing in a closure that can be invoked at the start of the `.each` function.

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
aeisenberg: This is used throughout the test file.

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
aeisenberg: Hmmmm...I don't think this is a great approach:

1. it prevents you from adding custom names for each test
2. it is more cumbersome to run individual tests

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
aeisenberg: I started going down this path, but setting up the parameters for `prepareRemoteQueryRun` is no simpler than using the `variantAnalysisManager` already existing. Since the tests are already set up for this, we can keep things simpler by reusing existing test infrastructure.

aeisenberg:
Thanks for your review. It's ready for another look.

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
koesie10: Ah, sorry, I only checked the diff, I missed that it's also used by the existing tests.

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
koesie10: It's totally fine to not do it, it was just something we might be able to try. Both look like valid points, so let's just keep it like this for now.

koesie10 (approved):

PR #2165, opened 2023-03-13
Simpler `withProgress` calls

koesie10:
See the two separate commits. This will make `withProgress` easier to use by removing the required options argument and making it optional instead. This brings it in line with how `commandRunnerWithProgress` calls this function.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/commandRunner.ts
+++ b/extensions/ql-vscode/src/commandRunner.ts
@@ -43,21 +43,34 @@ export interface ProgressUpdate {
 export type ProgressCallback = (p: ProgressUpdate) => void;
 
 /**
+ * A task that reports progress.
+ *
+ * @param progress a progress handler function. Call this
+ * function with a `ProgressUpdate` instance in order to
+ * denote some progress being achieved on this task.
+ * @param token a cancellation token
+ */
+export type ProgressTask<R> = (
+  progress: ProgressCallback,
+  token: CancellationToken,
+) => Thenable<R>;
+
+/**
  * A task that handles command invocations from `commandRunner`
  * and includes a progress monitor.
  *
  *
  * Arguments passed to the command handler are passed along,
- * untouched to this `ProgressTask` instance.
+ * untouched to this `ProgressTaskWithArgs` instance.
  *
  * @param progress a progress handler function. Call this
  * function with a `ProgressUpdate` instance in order to
  * denote some progress being achieved on this task.
- * @param token a cencellation token
+ * @param token a cancellation token
  * @param args arguments passed to this task passed on from
  * `commands.registerCommand`.
  */
-export type ProgressTask<R> = (
+export type ProgressTaskWithArgs<R> = (
   progress: ProgressCallback,
   token: CancellationToken,
   ...args: any[]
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/contextual/templateProvider.ts
```diff
@@ -91,6 +86,11 @@ export class TemplateQueryDefinitionProvider implements DefinitionProvider {
           (src, _dest) => src === uriString,
         );
       },
+      {
+        location: ProgressLocation.Notification,
```
starcke, line 90: Could you remove the location from here (due to the default) or is there something I am missing?

extensions/ql-vscode/src/contextual/templateProvider.ts
```diff
@@ -154,6 +149,11 @@ export class TemplateQueryReferenceProvider implements ReferenceProvider {
...
```
starcke, line 153: As above.

extensions/ql-vscode/src/local-databases.ts
```diff
@@ -861,6 +858,9 @@ export class DatabaseManager extends DisposableObject {
 
         void this.logger.log("Finished loading persisted databases.");
       },
+      {
```
starcke, line 861: With everything being default, could we just not pass it here?

starcke (commented):
Looks ok to me (but probably good to get second eyes). I wasnt super sure on the defaults though.

extensions/ql-vscode/src/contextual/templateProvider.ts
koesie10: Yep, we can definitely remove these.

extensions/ql-vscode/src/commandRunner.ts
```diff
@@ -189,7 +197,9 @@ export function commandRunnerWithProgress<R>(
         ...progressOptions,
       };
 
-      return withProgress(progressOptionsWithDefaults, task, ...args);
+      return withProgress(progressOptionsWithDefaults, (progress, token) =>
+        task(progress, token, ...args),
```
elenatanasoiu: nice!

extensions/ql-vscode/src/commandRunner.ts
```diff
@@ -42,6 +42,11 @@ export interface ProgressUpdate {
 
 export type ProgressCallback = (p: ProgressUpdate) => void;
 
+// Make certain properties within a type optional
+type Optional<T, K extends keyof T> = Pick<Partial<T>, K> & Omit<T, K>;
+
+export type ProgressOptions = Optional<VSCodeProgressOptions, "location">;
```
elenatanasoiu, line 48: This is cool, but I'm wondering what the advantage is over having a union of two types: one with the location attribute and one without?

elenatanasoiu: I can see you want to use `VSCodeProgressOptions`, but it is harder to read than two types added together. 

🤔 Just wondering how to make this easier to understand when you first look at it.

koesie10: I didn't want to copy over the types of `VSCodeProgressOptions`, but you're right that this makes it harder to read. We can copy over the types of that type, but that means potentially some additional work when we update the `vscode` types, which might not be too bad for this type. What do you think?

elenatanasoiu: Yeah, `ProgressOptions` and `ProgressLocation` don't look that elaborate so it wouldn't be that much work to change. But we would need to remember to do it, which we probably won't unless we have solid testing around what we expect the progress UI to look like. 

Since I don't think we have testing at this level, I'm leaning towards just using the root types like you said.

PR #2167, opened 2023-03-13
Upgrade to React 18

koesie10:
This upgrades the webviews to use React 18. There should be no user-facing changes.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/webview.tsx
+++ b/extensions/ql-vscode/src/view/webview.tsx
@@ -1,4 +1,5 @@
-import { render as ReactDOM_render } from "react-dom";
+import * as React from "react";
+import { createRoot } from "react-dom/client";
 import { vscode } from "./vscode-api";
 
 import { WebviewDefinition } from "./webview-definition";
@@ -25,11 +26,11 @@ const render = () => {
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const view: WebviewDefinition = require(`./${viewName}/index.tsx`).default;
 
-  ReactDOM_render(
-    view.component,
-    document.getElementById("root"),
-    // Post a message to the extension when fully loaded.
-    () => vscode.postMessage({ t: "viewLoaded", viewName }),
+  createRoot(element).render(
+    // Post a message to the extension when fully loaded. See https://github.com/reactwg/react-18/discussions/5 ("What about the render callback?")
+    <div ref={() => vscode.postMessage({ t: "viewLoaded", viewName })}>
+      {view.component}
+    </div>,
   );
 };
 
--- a/extensions/ql-vscode/package-lock.json
+++ b/extensions/ql-vscode/package-lock.json
@@ -32,8 +32,8 @@
         "node-fetch": "~2.6.7",
         "p-queue": "^6.0.0",
         "path-browserify": "^1.0.1",
-        "react": "^17.0.2",
-        "react-dom": "^17.0.2",
+        "react": "^18.2.0",
+        "react-dom": "^18.2.0",
         "semver": "~7.3.2",
         "source-map": "^0.7.4",
         "source-map-support": "^0.5.21",
@@ -66,7 +66,7 @@
         "@storybook/react": "^6.5.10",
         "@storybook/testing-library": "^0.0.13",
         "@testing-library/jest-dom": "^5.16.5",
-        "@testing-library/react": "^12.1.5",
+        "@testing-library/react": "^14.0.0",
         "@testing-library/user-event": "^14.4.3",
         "@types/child-process-promise": "^2.2.1",
         "@types/classnames": "~2.2.9",
@@ -85,8 +85,8 @@
         "@types/nanoid": "^3.0.0",
         "@types/node": "^16.11.25",
         "@types/node-fetch": "~2.5.2",
-        "@types/react": "^17.0.2",
-        "@types/react-dom": "^17.0.2",
+        "@types/react": "^18.0.28",
+        "@types/react-dom": "^18.0.11",
         "@types/sarif": "~2.1.2",
         "@types/semver": "~7.2.0",
         "@types/stream-chain": "~2.0.1",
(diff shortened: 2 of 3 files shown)


charisk (approved):

koesie10:
Thanks for the review @charisk! I've had to make some other changes to make it fully work with React 18. The results view wasn't working due to how React is now batching state updates, so there were quite some changes in how the results view now receives some messages.

@aeisenberg Would you be able to review those changes as well? It's mostly about https://github.com/github/vscode-codeql/pull/2167/commits/36eb2cd2eaeb1e8312e4aea9ed171e793e0b8b4d. I don't believe there are any major behaviour changes, but we're not using `shouldKeepOldResultsWhileRendering` anymore since we're loading the results in the same `setState` call as where that gets set.

extensions/ql-vscode/src/view/webview.tsx
```diff
@@ -25,11 +26,11 @@ const render = () => {
...
-  ReactDOM_render(
-    view.component,
-    document.getElementById("root"),
-    // Post a message to the extension when fully loaded.
-    () => vscode.postMessage({ t: "viewLoaded", viewName }),
+  createRoot(element).render(
+    // Post a message to the extension when fully loaded. See https://github.com/reactwg/react-18/discussions/5 ("What about the render callback?")
+    <div ref={() => vscode.postMessage({ t: "viewLoaded", viewName })}>
+      {view.component}
+    </div>,
```
elenatanasoiu, line 33: Shouldn't we be changing `view.component` to accept callbacks and then do:

```
const rootElement = createRoot(element);
rootElement.render(<view.component callback={() => vscode.postMessage({ t: "viewLoaded", viewName })} >)
```

I think at least having the `rootElement` in it's own variable makes it slight easier to parse.

extensions/ql-vscode/src/view/webview.tsx
```diff
@@ -25,11 +26,11 @@ const render = () => {
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const view: WebviewDefinition = require(`./${viewName}/index.tsx`).default;
 
-  ReactDOM_render(
-    view.component,
-    document.getElementById("root"),
-    // Post a message to the extension when fully loaded.
-    () => vscode.postMessage({ t: "viewLoaded", viewName }),
+  createRoot(element).render(
+    // Post a message to the extension when fully loaded. See https://github.com/reactwg/react-18/discussions/5 ("What about the render callback?")
```
elenatanasoiu, line 30: I'm not sure we need the comment to stick around. We could have it in the commit message for people to find if they want to know why we made the change.

extensions/ql-vscode/src/view/results/results.tsx
```diff
@@ -164,76 +162,52 @@ export class ResultsApp extends React.Component<
 
   private updateStateWithNewResultsInfo(resultsInfo: ResultsInfo): void {
     this.setState((prevState) => {
-      const stateWithDisplayedResults = (
```
elenatanasoiu, line 167: Could we write tests for the behaviour before we made this change and then see if the same behaviour happens afterwards? That would mean we wouldn't need to guess if we've completely got it right.

elenatanasoiu:
Do we get any extra warnings now that we've upgraded to React 18? I'd be surprised if we have a clean upgrade.

extensions/ql-vscode/src/view/webview.tsx
koesie10: I think localising this behaviour to the `webview.tsx` file makes the rest of the components much clearer. I don't think the other components should be concerned with how the `viewLoaded` message gets send and adding an extra `div` shouldn't change anything about the layout.

I will extract the `createRoot` result into a separate variable, this also seems to be shown in [the React documentation](https://beta.reactjs.org/reference/react-dom/client/createRoot#createroot) with a name of `root`. I think `rootElement` could be confusing because you could confuse it with an HTML element. The type name of the return type of `createRoot` is `Root`, so calling it `root` like in the documentation seems like it would make sense.

extensions/ql-vscode/src/view/webview.tsx
koesie10: I left the comment here because this is quite a strange use of the `ref` property. `ref` is usually used to get a reference to the underlying HTML element, and we're using it in a completely different way. I can definitely remove it and we can rely on the Git history.

koesie10:
> Do we get any extra warnings now that we've upgraded to React 18? I'd be surprised if we have a clean upgrade.

We don't have any extra warnings since we're actually not using that many React features. What kind of warnings would you have expected to see?

extensions/ql-vscode/src/view/webview.tsx
elenatanasoiu: That's fair, I didn't clock from the link that `ref` is required now. Should we extract this div into an `App` component to more closely match the docs example?

elenatanasoiu:
I was expecting to see warnings when we run `npm run build` / `npm run watch`, during CI or when running tests locally. But if those are clean, that's great. I also see you''re enabling strict mode in a different PR so maybe that will show more warnings in the future.

extensions/ql-vscode/src/view/webview.tsx
elenatanasoiu: Sounds good, and yeah adding it to view.component doesn't quite make sense. Perhaps we can turn the extra `div` into an `App`-like component, like they have in the docs?

extensions/ql-vscode/src/view/results/results.tsx
elenatanasoiu: Is it hard because we would have to check visual components or because it's a timing issue?

koesie10: It's a timing issue, there's no real way to check the state in between the two `setState` calls, so we'd never get into the state that we'd want to test. We'd want to check what the view looks like in between [these two calls](https://github.com/github/vscode-codeql/blob/64d97aaf7e47f1a012d4564159193b767ff14eb0/extensions/ql-vscode/src/view/results/results.tsx#L93-L107) but they are called sequentially without any waiting, so there's no way to actually test that state.

elenatanasoiu: I'm gonna defer to @aeisenberg on this one. I would ask if we can get rid of `shouldKeepOldResultsWhileRendering` entirely if we're combining calls.

extensions/ql-vscode/src/view/jest.setup.ts
```diff
@@ -15,6 +15,11 @@ Object.defineProperty(window, "matchMedia", {
   })),
 });
 
+// Used by Primer React
+(window as any).CSS = {
+  supports: jest.fn().mockResolvedValue(false),
```
elenatanasoiu, line 20: What are we mocking here exactly? If this always returns false, are we saying we won't support Primer CSS? My primer-fu is failing me on this one. 

I can't see primer/css installed in package.json so I'm assuming it's a dependency of `primer/react` or some other `@primer/*` package. 

I'm also aware we want to move away from Primer entirely and it looks like only one component uses it, through `ThemeProvider`. I'm wondering if we could make a copy of the `ThemeProvider` component and delete primer entirely. One for later probably.

koesie10: We're mocking `CSS.supports` here to make these calls work: https://github.com/primer/react/blob/f0e0ec77f5d91c73ce22d10bc8d55a1a5ab2a9e6/src/PageLayout/useStickyPaneHeight.ts#L106-L113. We're essentially saying that we don't support any modern CSS features, see [the MDN documentation for `CSS.supports`](https://developer.mozilla.org/en-US/docs/Web/API/CSS/supports). This won't change any of the behaviour, it just makes it look like we're using an older browser (which `jest-dom` essentially is since it doesn't support CSS at all).

The `ThemeProvider` code doesn't look like it uses many other files from `@primer/react`: https://github.com/primer/react/blob/f0e0ec77f5d91c73ce22d10bc8d55a1a5ab2a9e6/src/ThemeProvider.tsx. However, we'd also need to copy over the `Overlay` component. This seems to have a lot of dependencies on other files in `@primer/react`, so we'd also need to copy over those: https://github.com/primer/react/blob/f0e0ec77f5d91c73ce22d10bc8d55a1a5ab2a9e6/src/Overlay.tsx

@charisk is working on a new CodePaths view that doesn't use the `Overlay` component, so we can get rid of Primer once that work is done. That would also allow us to get rid of this mock.

extensions/ql-vscode/src/view/results/results.tsx
elenatanasoiu: Looks like Andrew is on holiday so you might have to wait a bit with this one.

charisk: Just wanted to say that I wasn't expecting us to do a complete re-write of this view soon as the changes we'd like to make are mostly cosmetic. But I can be convinced otherwise.

charisk:
> Thanks for the review @charisk! I've had to make some other changes to make it fully work with React 18. The results view wasn't working due to how React is now batching state updates, so there were quite some changes in how the results view now receives some messages.

Thanks for the poke! The changes look okay to me, but as you say the logic is quite hard to follow. I've also tested them a bit manually and things seem ok to me.

elenatanasoiu:
Thanks for testing @charisk! I wasn't sure how to go about checking this. If we know what to expect in the behaviour change, I don't want to hold this up.

elenatanasoiu (approved):

extensions/ql-vscode/src/view/results/results.tsx
```diff
@@ -164,76 +162,52 @@ export class ResultsApp extends React.Component<
...
+          database: resultsInfo.database,
+          sortStates: this.getSortStates(resultsInfo),
+        };
+      } catch (e) {
+        let errorMessage: string;
+        if (e instanceof Error) {
+          errorMessage = e.message;
+        } else {
+          errorMessage = "Unknown error";
+        }
```
aeisenberg, line 206:
```suggestion
        const errorMessage = getErrorMessage(e);
```

extensions/ql-vscode/src/view/results/results.tsx
```diff
@@ -164,76 +162,52 @@ export class ResultsApp extends React.Component<
...
+          return {
+            displayedResults: {
+              resultsInfo: null,
+              results: null,
+              errorMessage: "No results to display",
+            },
+            isExpectingResultsUpdate: prevState.isExpectingResultsUpdate,
+            nextResultsInfo: resultsInfo,
+          };
+        }
```
aeisenberg, line 188: Are you missing a case here? I don't see line 191 from the original handled.

```
return stateWithDisplayedResults(prevState.displayedResults);
```

extensions/ql-vscode/src/view/results/results.tsx
aeisenberg: This method is overly complex and a simplification would be really nice. To be safe when we are doing a refactoring like this, it would be nice to add some pure regression unit tests to ensure that all the different cases are handled identically before and after.

Although nice, it's not necessary to spend the time to create more complex tests that check the view between the two `setState` calls. All that I think is necessary is that we ensure the method behaves the same before and after the refactoring. It's a complicated method, but at least the complexity is contained.

koesie10: I've created a really simple regression test in #2255. This does not test the loading state, but checks that we can at least render a SARIF file.

extensions/ql-vscode/src/view/results/results.tsx
koesie10: In this case, the missing case is correct. In the previous code, after the `setState` returned with `stateWithDisplayedResults(prevState.displayedResults)`, `this.loadResults()` would be called immediately. The only reason for storing the `nextResultsInfo` in the state is so that `loadResults` can use it. However, with concurrent updates the state seen in `loadResults` would not yet be the state set in this `setState` call. Therefore, I've combined this `setState` and the `loadResults` method. Therefore, this case won't happen anymore.

Thanks for the suggestion for the simplification. Since we don't need the third case, I couldn't use it as-is. However, I have split the flows differently so there's only 1 level of nesting rather than 2.

aeisenberg (commented):
This looks good now, but can you rebase on https://github.com/github/vscode-codeql/pull/2255 so we can be sure that the behaviour remains consistent?

PR #2169, opened 2023-03-13
Use `TeeLogger` instead of `additionalLogLocation:` option

dbartol:
This PR refactors the existing implemention of "side logs" to avoid the need to remove additional log locations once logging is complete. This simplifies some in-flight work on debugging context by removing the need for the result of `QueryRunner.compileAndRunQueryOnDatabase()` to be disposable.

### Code Changes
- Removed the `additionalLogLocation:` option to `Logger.log()`
- Removed `Logger.removeAdditionalLogLocation()`
- Created the new class `TeeLogger`, which wraps an existing `Logger` and also sends output to the specified file. The banner that was previously emitted when a particular side log path was first used is now emitted on the first message logged to `TeeLogger`
- Updated the query server clients to track `activeQueryLogger` instead of `activeQueryLogFile`.
- Created a `TeeLogger` for each query we compile/run
- Fixed up code that no longer needs to dispose query results.
- Updated tests

### DIFF
--- a/extensions/ql-vscode/src/common/logging/tee-logger.ts
+++ b/extensions/ql-vscode/src/common/logging/tee-logger.ts
@@ -1,5 +1,6 @@
 import { appendFile, ensureFile } from "fs-extra";
 import { isAbsolute } from "path";
+import { getErrorMessage } from "../../pure/helpers-pure";
 import { Logger, LogOptions } from "./logger";
 
 /**
@@ -11,6 +12,7 @@ import { Logger, LogOptions } from "./logger";
  */
 export class TeeLogger implements Logger {
   private emittedRedirectMessage = false;
+  private error = false;
 
   public constructor(
     private readonly logger: Logger,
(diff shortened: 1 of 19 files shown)


extensions/ql-vscode/src/common/logging/tee-logger.ts
```diff
@@ -0,0 +1,49 @@
...
+      await this.logger.log(msg);
+      await this.logger.log(separator);
+    }
+
+    const trailingNewline = options.trailingNewline ?? true;
+    await ensureFile(this.location);
+
+    await appendFile(this.location, message + (trailingNewline ? "\n" : ""), {
+      encoding: "utf8",
+    });
```
aeisenberg, line 41: Maybe wrap this in a try-catch block and print a reasonable message to the main log location if there's an error.

aeisenberg (commented):
Looks good. Suggestion for a minor improvement.

extensions/ql-vscode/src/common/logging/tee-logger.ts
dbartol: Done

aeisenberg (approved):

PR #2171, opened 2023-03-14
Convert run variant analysis command to new commands package

koesie10:
This replaces the run variant analysis commands by an implementation based on the new commands package. There should be no changes in behaviour.

Based on #2157

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/commands.ts
+++ b/extensions/ql-vscode/src/common/commands.ts
@@ -1,4 +1,5 @@
 import { CommandManager } from "../packages/commands";
+import type { Uri } from "vscode";
 
 /**
  * Contains type definitions for all commands used by the extension.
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/common/commands.ts
```diff
@@ -1,4 +1,5 @@
 import { CommandManager } from "../packages/commands";
+import type { Uri } from "vscode";
```
koesie10, line 2: I'd have preferred to not import from `vscode` in the `src/common` directory, but importing a type should be fine. We could also create an interface that defines the necessary properties to avoid importing from `vscode` here, but we'd need to find out which methods on `Uri` in command arguments are being used. For this command, it's just `fsPath`, but for others there may be more.

PR #2172, opened 2023-03-14
Add new data flow paths view (empty)

charisk:
Add a new webview that will be used to show the data flow paths, instead of the full screen overlay that we use currently. See linked internal issue for more details. 

This is currently not in use - it will be wired up in a follow-up PR.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
+++ b/extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
@@ -21,13 +21,10 @@ export class DataFlowPathsView extends AbstractWebview<
   }
 
   public async showDataFlows(dataFlowPaths: DataFlowPaths) {
-    if (!this.isShowingPanel) {
-      await this.getPanel();
-      await this.waitForPanelLoaded();
-    }
+    const panel = await this.getPanel();
+    panel.reveal(undefined, true);
 
-    // Focus on panel
-    this.panel?.reveal(undefined, true);
+    await this.waitForPanelLoaded();
 
     await this.postMessage({
       t: "setDataFlowPaths",
--- a/extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
+++ b/extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
@@ -1,0 +1,71 @@
+import { ExtensionContext, ViewColumn } from "vscode";
+import { AbstractWebview, WebviewPanelConfig } from "../abstract-webview";
+import { assertNever } from "../pure/helpers-pure";
+import { telemetryListener } from "../telemetry";
+import {
+  FromDataFlowPathsMessage,
+  ToDataFlowPathsMessage,
+} from "../pure/interface-types";
+import { DataFlowPaths } from "./shared/data-flow-paths";
+import { showAndLogExceptionWithTelemetry } from "../helpers";
+import { redactableError } from "../pure/errors";
+
+export class DataFlowPathsView extends AbstractWebview<
+  ToDataFlowPathsMessage,
+  FromDataFlowPathsMessage
+> {
+  public static readonly viewType = "codeQL.dataFlowPaths";
+
+  public constructor(ctx: ExtensionContext) {
+    super(ctx);
+  }
+
+  public async showDataFlows(dataFlowPaths: DataFlowPaths) {
+    if (!this.isShowingPanel) {
+      await this.getPanel();
+      await this.waitForPanelLoaded();
+    }
+
+    // Focus on panel
+    this.panel?.reveal(undefined, true);
+
+    await this.postMessage({
+      t: "setDataFlowPaths",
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public async showDataFlows(dataFlowPaths: DataFlowPaths) {
+    if (!this.isShowingPanel) {
+      await this.getPanel();
+      await this.waitForPanelLoaded();
+    }
+
+    // Focus on panel
+    this.panel?.reveal(undefined, true);
```
koesie10, line 28: This seems to be in a different order than the variant analysis view. Does this construct from the variant analysis view not work for this view or should the variant analysis view be changed to use this?

```suggestion
    const panel = await this.getPanel();
    panel.reveal(undefined, true);

    await this.waitForPanelLoaded();
```

extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
```diff
@@ -0,0 +1,62 @@
...
+    await this.postMessage({
+      t: "setDataFlowPaths",
+      dataFlowPaths,
+    });
+  }
+
+  protected async getPanelConfig(): Promise<WebviewPanelConfig> {
+    return {
+      viewId: DataFlowPathsView.viewType,
+      title: "Data Flow Paths",
```
koesie10, line 39: Will you be updating this in a future PR to include the name of the query and potentially the name of the database/repository?

extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
```diff
@@ -0,0 +1,62 @@
...
+  FromDataFlowPathsMessage,
+  ToDataFlowPathsMessage,
+} from "../pure/interface-types";
+import { DataFlowPaths } from "./shared/data-flow-paths";
+
+export class DataFlowPathsView extends AbstractWebview<
+  ToDataFlowPathsMessage,
+  FromDataFlowPathsMessage
+> {
+  public static readonly viewType = "codeQL.dataFlowPaths";
```
koesie10, line 15: Should opening this view also be added as an activation event?

It might not be necessary to add it since it can only be opened if another CodeQL view is already open, but perhaps it would be good for consistency?

extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
charisk: I think that the use case is just slightly different. With the data flow paths view, we'd want the callers to just say "show the paths" instead of having separate `openView` and `updateView` steps. We also only want to ever allow 1 data flow paths view to ever be open, rather than one for each set of data flow paths it renders.

extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
charisk: I don't believe so - we won't be including that information AFAIK. That could potentially change once Product/Design reviews though.

extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
koesie10: I think that is still possible with the logic from the variant analysis view. The `isShowingPanel` check is implicit when you call `getPanel`, and `waitForPanelLoaded` will just resolve immediately when the panel is already loaded. So, I think the behaviour is the same. I think it would be good to keep these checks consistent between the views.

extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
koesie10: Activation events are also used for opening a non-persisted view, so it would also be an activation event if you opened the view from somewhere else. However, as I mentioned I don't believe it's necessary for this view to be an activation event because another view will always have to be loaded already before you can open this one. However, all of our current views are in the activation events, even those that currently also have that behaviour (like `onWebviewPanel:resultsView`), so I think it would be good to add it for consistency.

extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
charisk: Oh sorry I didn't look at the diff properly. I thought you were suggesting having two functions - `openView` and `editView`. Made the change now.

extensions/ql-vscode/src/variant-analysis/data-flow-paths-view.ts
charisk: Seems odd to just do it for consistency but that's perhaps a tech-debt issue. Updated.

koesie10 (approved):

PR #2173, opened 2023-03-14
Add linting of jsdoc comments

robertbrignull:
Installs the [jsdoc eslint plugin](https://github.com/gajus/eslint-plugin-jsdoc). This plugin is recommended from https://eslint.org/blog/2018/11/jsdoc-end-of-life/ as the suggested replacement for eslint's builtin rules which were deprecated.

We have quite a bit of jsdoc in the repo and a lot of it is outdated and incorrect. I'm not arguing that every function needs to have a jsdoc comment, but if there is a jsdoc comment it should be correct. So I'm suggesting we disable the `jsdoc/require-jsdoc` rule, but keep basically everything else from the "recommended" set. I've also disabled the `jsdoc/require-param-type` rule because it seems unnecessary to duplicate the type information.

As of opening this PR, this generates 590 warnings. Fixing these will take time and effort so I don't want to make these errors, though that could be an eventual goal. Right now it'll just put a whole load of yellow squigglies in the IDE and hopefully we'll address them as we touch those areas of code.

Do people agree that getting rid of incorrect jsdoc is important enough to do this? And is this many warnings ok, or is it too annoying?

## Checklist

### DIFF
--- a/extensions/ql-vscode/.eslintrc.js
+++ b/extensions/ql-vscode/.eslintrc.js
@@ -14,7 +14,7 @@ const baseConfig = {
       resolve(__dirname, ".storybook/tsconfig.json"),
     ],
   },
-  plugins: ["github", "@typescript-eslint", "etc"],
+  plugins: ["github", "@typescript-eslint", "etc", "jsdoc"],
   env: {
     node: true,
     es6: true,
@@ -27,6 +27,7 @@ const baseConfig = {
     "plugin:jest-dom/recommended",
     "plugin:prettier/recommended",
     "plugin:@typescript-eslint/recommended",
+    "plugin:jsdoc/recommended",
   ],
   rules: {
     "@typescript-eslint/await-thenable": "error",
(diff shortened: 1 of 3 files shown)


robertbrignull:
Apparently warnings still fail the linting CI check. I assume we could work around that if we wanted to make these non-failing.

koesie10 (commented):
I'm not sure whether disabling just these rules will be enough. Just looking through the warnings, there still seem to be some rules which are incorrect or redundant:
- `jsdoc/require-returns-type`: This is redundant because TypeScript already has return type informatino
- `jsdoc/valid-types`: These seem to be incorrect, they do work in the IDE and this is the recommended syntax

I think some of the comments added to functions are not JSDoc, but are written in the JSDoc syntax. By changing their starting characters from `/**` to `/*`, we can probably already remove those warnings.

It seems like we can also autofix at least some of these warnings, which is probably a good idea to do before merging this. This can be done by running `npm run lint -- --fix`. After this, there are still 476 warnings remaining. After also disabling the rules mentioned above, there are 399 warnings remaining.

I must say that I'm not sure whether we want to fix all of these warnings. For example, there are a lot of warnings about missing param descriptions. It's quite obvious what `cliServer` is, and adding a description for that is not really necessary. In most cases, it seems like we've added descriptions for some parameters which are non-obvious, but don't necessarily need it for the other params.

I think a better approach might be to selectively enable rules, rather than disabling some rules. There should be warnings when we've added a param description for a param that doesn't exist anymore (which is probably `jsdoc/check-param-names`), and perhaps when we're using an invalid tag (`jsdoc/check-tag-names`). I think most of the other rules are unnecessary overhead, but I'm happy to hear other opinions.

PR #2176, opened 2023-03-15
Convert query history commands to typed commands

koesie10:
This converts the query history commands to the new typed commands package. There should be no behaviour changes.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -644,7 +644,7 @@
       },
       {
         "command": "codeQLQueryHistory.openOnGithub",
-        "title": "Open on GitHub"
+        "title": "View Logs"
       },
       {
         "command": "codeQLQueryHistory.copyRepoList",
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/package.json
```diff
@@ -644,7 +644,7 @@
       },
       {
         "command": "codeQLQueryHistory.openOnGithub",
-        "title": "View Logs"
+        "title": "Open on GitHub"
```
robertbrignull, line 647: I don't necessarily disagree with this change, but it's worth noting that this will be a user-visible change because it's visible in the query history context menu. Is there a desire to change the name of this menu option?

robertbrignull (commented):
One question, but otherwise LGTM. I find it pleasing that even though we're now listing all the commands in another place it's still a net reduction in lines, and much safer too.

extensions/ql-vscode/package.json
koesie10: Good point, I did this because this is one of the only commands where the command name doesn't match the title, but we'll probably want to keep it as-is. I'll revert the change.

PR #2185, opened 2023-03-17
Remove two more internal-only commands

robertbrignull:
Removes the `codeQL.cancelVariantAnalysis` and `codeQL.openVariantAnalysisQueryText` commands. I've done these two together in one PR since they both also use the approach from https://github.com/github/vscode-codeql/pull/2179.

Check the individual commits for changes relevant to each command. For each one we essentially inline the implementation so we call the method on the variant analysis manager directly instead of going through a command. The only case where we're not already inside a command is when triggered from the variant analysis view, so we add dedicated telemetry to cover that.

## Checklist


koesie10 (approved):
LGTM

PR #2188, opened 2023-03-20
Codespace: Open tutorial workspace on extension start

elenatanasoiu:
When opening https://github.com/github/codespaces-codeql/ in a codespace, it's easy to miss the prompt that tells you to open the tutorial.code-workspace file.

In fact people actively dismiss the alert to get it out of the way.

If you miss that prompt, you end up with a single-rooted workspace, which causes various other problems.

While there is an open issue to allow VS Code to open a default workspace [1], there doesn't seem to have been any progress on it in the last two years.

So we're taking matters into our own hands and forcing the extension to open the tutorial workspace, if it detects it.

This will only happen if the following three conditions are met:
- the `.tours` folder exists
- the `tutorial.code-workspace` file exists
- the `CODESPACES_TEMPLATE` setting hasn't been set

NB: the `CODESPACES_TEMPLATE` setting can only be found if the `tutorial.code-workspace` has already been opened. So it's a good indicator that we're in the folder, but the user has ignored the prompt.

[1]: https://github.com/microsoft/vscode-remote-release/issues/3665

## Testing

I've added more in-depth testing for the behaviour in the code, but it doesn't hurt to see it work as expected so I've tested this locally by:
- creating a `.vsix` file with this branch of the extension
- opening the `codespaces-codeql` repo but not the workspace
- installing the CodeQL extension from the `vsix` file
- clicking on the extension
- seeing that the workspace is reloaded and the tutorial workspace is opened 🎉 
- trying to add a database and seeing the skeleton pack getting generated (this means we're using codespace behaviour) 🎉 🎉 

Separately, I've also tested that this behaviour doesn't happen with the vscode-codeql-starter workspace. 🎉 🎉 🎉


extensions/ql-vscode/src/helpers.ts
```diff
@@ -266,6 +269,29 @@ export function isFolderAlreadyInWorkspace(folderName: string) {
...
+    );
+
+    const toursFolderPath = join(currentFolder, ".tours");
+
+    if (
+      existsSync(tutorialWorkspaceUri.fsPath) &&
+      existsSync(toursFolderPath) &&
+      !isCodespacesTemplate()
+    ) {
+      await commands.executeCommand("vscode.openFolder", tutorialWorkspaceUri);
```
elenatanasoiu, line 290: While this command is called `openFolder`, it actually accepts a workspace file URI and will open the workspace correctly. More details on this: https://github.com/Microsoft/vscode/issues/42610

elenatanasoiu:
Ideally we'd have this merged before the next extension release (Wednesday) so that we're able to test it in the codespace.

shati-patel (commented):
Thank you for the super helpful PR description & testing notes ❤️ 

The code generally looks good, though I'd like to have another play with this tomorrow before approving officially ✔️  A couple of things:

- There's a lag between the user opening VS Code, and the extension being activated + reloading the workspace. This is probably disruptive if the user is already clicking around and "doing stuff" when suddenly their window reloads. I'm not sure if we can avoid this, sadly, since the extension takes a while to activate...
- I haven't been able to get this working locally on Windows (a possible path issue?) The window reloads, but then hangs in a completely empty workspace. I get some errors in the debug console, which I've pasted below. Happy to try and investigate more!
  <details><summary>Details</summary>
  <p>

  ```
  rejected promise not handled within 1 second: Error: Invalid arguments
  extensionHostProcess.js:100
  stack trace: Error: Invalid arguments
      at new n (c:\Users\ShatiPatel\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.  js:94:15017)
      at apply (c:\Users\ShatiPatel\.vscode\extensions\vsls-contrib.codetour-0.0.58\src\player\decorator.ts:97:23)
      at Array.map (<anonymous>)
      at Proxy.map (c:\Users\ShatiPatel\.vscode\extensions\vsls-contrib.codetour-0.0.58\node_modules\mobx\lib\mobx.module.js:3381:30)
      at c:\Users\ShatiPatel\.vscode\extensions\vsls-contrib.codetour-0.0.58\src\player\decorator.ts:96:43
      at Generator.next (<anonymous>)
      at s (c:\Users\ShatiPatel\.vscode\extensions\vsls-contrib.codetour-0.0.58\dist\extension-node.js:16:122957)
  ```
  </p>
  </details>

extensions/ql-vscode/test/vscode-tests/no-workspace/helpers.test.ts
```diff
@@ -559,3 +569,103 @@ describe("isFolderAlreadyInWorkspace", () => {
...
+
+        // spy that we open the workspace file by calling the 'vscode.openFolder' command
+        const commandSpy = jest.spyOn(commands, "executeCommand");
+        commandSpy.mockImplementation(() => Promise.resolve());
+
+        await prepareCodeTour();
+
+        expect(commandSpy).toHaveBeenCalledWith(
+          "vscode.openFolder",
+          expect.anything(),
```
adityasharad, line 615: You could expect this to be the (system-specific) path to the workspace file right?

extensions/ql-vscode/src/helpers.ts
```diff
@@ -266,6 +269,36 @@ export function isFolderAlreadyInWorkspace(folderName: string) {
...
+    const tutorialWorkspacePath = join(
+      currentFolder,
+      "tutorial.code-workspace",
+    );
+    const toursFolderPath = join(currentFolder, ".tours");
+
+    if (
+      existsSync(tutorialWorkspacePath) &&
+      existsSync(toursFolderPath) &&
+      !isCodespacesTemplate()
```
adityasharad, line 289: Minor: comment here why we need this negated condition. The explanation you put in the PR description is helpful.

extensions/ql-vscode/src/helpers.ts
```diff
@@ -266,6 +269,36 @@ export function isFolderAlreadyInWorkspace(folderName: string) {
...
+    // We need this path to check that the file exists on windows
+    const tutorialWorkspacePath = join(
+      currentFolder,
+      "tutorial.code-workspace",
+    );
+    const toursFolderPath = join(currentFolder, ".tours");
+
+    if (
+      existsSync(tutorialWorkspacePath) &&
+      existsSync(toursFolderPath) &&
```
adityasharad, line 288: Could these be async? Otherwise you have a mix of sync and async operations in this async handler.

extensions/ql-vscode/src/helpers.ts
```diff
@@ -266,6 +269,36 @@ export function isFolderAlreadyInWorkspace(folderName: string) {
...
+      !isCodespacesTemplate()
+    ) {
+      const tutorialWorkspaceUri = Uri.parse(
+        join(
+          workspace.workspaceFolders[0].uri.fsPath,
+          "tutorial.code-workspace",
+        ),
+      );
+      await commands.executeCommand("vscode.openFolder", tutorialWorkspaceUri);
```
adityasharad, line 297: Minor: Consider adding some debug logging just before you issue the command, so it's easier to debug this sequence of events in future.

Also consider giving the user a notification message telling them we're about to restart. You can decide whether you want to notify and restart immediately, or whether you want to make this a blocking prompt where you restart only if the user says yes. The spontaneous restart has been one cause of user confusion in workshops when switching to a multiroot workspace file, so we could take the opportunity to reduce such confusion here.

adityasharad (commented):
Great idea making the extension handle this as a workaround. Minor suggestions only.

extensions/ql-vscode/src/helpers.ts
```diff
@@ -266,6 +269,36 @@ export function isFolderAlreadyInWorkspace(folderName: string) {
...
+    const toursFolderPath = join(currentFolder, ".tours");
+
+    if (
+      existsSync(tutorialWorkspacePath) &&
+      existsSync(toursFolderPath) &&
+      !isCodespacesTemplate()
+    ) {
+      const tutorialWorkspaceUri = Uri.parse(
+        join(
+          workspace.workspaceFolders[0].uri.fsPath,
```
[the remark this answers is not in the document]
elenatanasoiu: It should be, but unfortunately I've found the two behave differently for windows tests. It might be related to the error you found above. I'll look into it.

extensions/ql-vscode/src/extension.ts
```diff
@@ -344,6 +345,14 @@ export async function activate(
     codeQlExtension.variantAnalysisManager,
   );
 
+  try {
+    await prepareCodeTour();
```
koesie10, line 349: Could we move this call to be before activating the extension? It doesn't seem like this uses the extension or CLI, so by moving it up above the `installOrUpdateThenTryActivate` call, we can significantly reduce the lag between the user opening VS Code and reloading the workspace mentioned by @shati-patel.

elenatanasoiu: Ah nice! I'll give it a shot, thanks! :)

extensions/ql-vscode/src/helpers.ts
elenatanasoiu: It looks the async versions of `exists`, `access` and `stat` all expect something to happen in a callback, rather than returning a boolean so they're a bit unwieldy:

```
exists('/path/to/a/file', (e) => {
  console.log(e ? 'it exists' : 'no file found!');
});
```

Other example: https://github.com/github/vscode-codeql/blob/3c229d244e53eb4cd7df93abff1c48d6b2352a1f/extensions/ql-vscode/src/test-adapter.ts#L330

I'm happy to adapt it though, just not sure what we've done so far in the extension for these cases. There seem to be multiple solutions.

extensions/ql-vscode/src/helpers.ts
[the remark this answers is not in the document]
elenatanasoiu: Ended up using the whole path, which we already store: https://github.com/github/vscode-codeql/pull/2188/commits/737a1f5c370940876357548fdcfd59f966874ce6

extensions/ql-vscode/src/helpers.ts
elenatanasoiu: Have added some logging and a prompt. This opens immediately when you click on the extension:

Screenshot 2023-03-21 at 11 15 16

extensions/ql-vscode/src/extension.ts
elenatanasoiu: Alrighty, I've moved this up and it looks like the prompt to open the workspace appears immediately. 

I'm unable to test this on windows, but at least we know it still works.

@shati-patel are you willing to try it out one more time?

extensions/ql-vscode/test/vscode-tests/no-workspace/helpers.test.ts
elenatanasoiu: Added the specific URI we expect

extensions/ql-vscode/src/helpers.ts
elenatanasoiu: Added a comment.

elenatanasoiu:
Thanks for the feedback everyone! This is ready for a re-review.

extensions/ql-vscode/test/vscode-tests/no-workspace/helpers.test.ts
elenatanasoiu: Looks like the test becomes flakey if we do this, even though we're using `Uri.parse` in both situations: https://github.com/github/vscode-codeql/actions/runs/4478429334/jobs/7871178529#step:7:231

elenatanasoiu: In the end I used `expect.objectContaining()` to just compare the path property instead of the whole URI: https://github.com/github/vscode-codeql/pull/2188/commits/8acf3e96512e2bfdff9643d596a9ac7d947de590

extensions/ql-vscode/src/helpers.ts
elenatanasoiu: Thanks! This seems to work fine 👍

extensions/ql-vscode/test/vscode-tests/no-workspace/helpers.test.ts
elenatanasoiu: [Woot! Success!](https://github.com/github/vscode-codeql/pull/2188/commits/8fd9ebf2d89ce1835040bec24177d3fd39d8edc5). 

We're able to check the path now, I just needed to compare the parsed `fsPath` off the result of `Uri.parse`, rather than the raw file path before parsing. 

----
### Aside

`toEqualPath` would have been perfect here, if I was using the comparison outside of `toHaveBeenCalledWith` instead of a nested expectation. 

I don't think I'm able to do:

```
expect(commandSpy).toHaveBeenCalledWith(
          "vscode.openFolder",
          expect(something).toEqualPath(otherThing)
)
```

shati-patel (approved):
Nice! Thanks for the PR and the changes 💪🏽 As discussed, we can merge this and tidy up some of the issues afterwards. 

Main follow-ups: getting this working (or at least non-breaking 😅 ) on local Windows, and adjusting the code tour steps to make it run more smoothly in the https://github.com/github/codespaces-codeql/ template 🚀

elenatanasoiu:
Thanks @shati-patel !

shati-patel:
> We have been battling with this problem as well! Would it be possible to generalize this to a stand-alone extension, so anyone can specify a default code-workspace file?

Hi @chris-boson! Sorry to hear you're having the same problem 😞 It's a shame VS Code doesn't have this option by default yet 😢 

We don't really have the capacity to create/maintain a stand-alone extension, but you're welcome to do so (and copy this PR's approach, if it's helpful)! 💚

PR #2191, opened 2023-03-20
Convert all variant analysis commands to typed commands

robertbrignull:
Please check each individual commit because I've done one commit per command.

This did actually find some places where our command arguments were incorrect. For the `codeQL.monitorVariantAnalysis` and `codeQL.autoDownloadVariantAnalysisResult` commands we were accepting a cancellation token parameter, but this parameter was not being passed in practice when we were invoking the commands. This meant the type of the argument was incorrect, but the commands already accounted this by checking if the token was undefined and ignoring it if so. Therefore this should be a behaviour-preserving change because the cancellation token was in fact always undefined already.

For `codeQL.loadVariantAnalysisRepoResults` I chose to give it a return type because this means we can use the existing  `loadResults` method without having to define a new method that just ignores the return value. But I could do this if you think it would be better.

## Checklist


koesie10 (approved):
LGTM, I just have 1 very minor suggestion

PR #2196, opened 2023-03-21
Convert AST and CFG commands to typed commands

koesie10:
Please review this commit-by-commit. This converts the AST and CFG commands to typed commands. It also moves them to a separate file to reduce the size of `extension.ts`.

## Checklist


extensions/ql-vscode/src/extension.ts
```diff
@@ -795,6 +815,16 @@ async function activateWithInstalledDistribution(
     ...variantAnalysisManager.getCommands(),
     ...databaseUI.getCommands(),
     ...dbModule.getCommands(),
+    ...getAstCfgCommands({
```
robertbrignull, line 827: Hmm, this is a lot of arguments and it obscures things a bit. I think it's fine for this PR and I'm not suggesting anything right now, but it would be nice if we could make this clearer and clearer in the future.

robertbrignull (approved):

extensions/ql-vscode/src/extension.ts
koesie10: I completely agree. In the future we should probably create separate modules and then we'd be able to just pass any dependent modules to it, and create the AST viewer and template providers in the module itself. However, this is something for the future.

PR #2207, opened 2023-03-22
Start using app.commands.execute for all commands called from extension.ts

robertbrignull:
We want to replace all calls to `commands.executeCommand` with `app.commands.execute` so we can take advantage of the new typing. This PR converts the few calls from `extension.ts` to the new method. This is just a first small PR to test the approach and make sure we're happy.

I've introduced the separation between commands where we are providing the implementation, and commands that are provided for us but we still want to get strong typing.

If we're happy with the groundwork laid by this PR then we should be able to start converting more command usages.

## Checklist


extensions/ql-vscode/src/common/commands.ts
```diff
@@ -30,6 +30,13 @@ export type SingleSelectionCommandFunction<Item> = (
  * the implementation in the corresponding `getCommands` function.
  */
 
+// Builtin commands where the implementation is provided by VS Code and not by this extension.
+// See https://code.visualstudio.com/api/references/commands
+export type VSCodeCommands = {
```
charisk, line 35: What do you think about calling these `BuiltInVsCodeCommands` instead?

charisk (approved):
Looks reasonable to me!

extensions/ql-vscode/src/common/commands.ts
```diff
@@ -185,7 +192,8 @@ export type EvalLogViewerCommands = {
   "codeQLEvalLogViewer.clear": () => Promise<void>;
 };
 
-export type AllCommands = BaseCommands &
+// All commands where the implementation is provided by this extension.
+export type AllCodeQLCommands = BaseCommands &
```
koesie10, line 196: Should this perhaps be called `AllExtensionCommands` or just `ExtensionCommands`? I believe we're mostly using extension to refer to the extension (like in `ExtensionApp`), rather than CodeQL (since this can also refer to the CLI).

koesie10 (approved):

PR #2212, opened 2023-03-22
Replace app.executeCommand with app.commands.execute

robertbrignull:
There is a `executeCommand` method on `App` leftover from before we started the migration to typed commands. This PR removes it and replaces all usages with typed commands.

## Checklist


charisk (approved):
Thanks for tidying that up!

robertbrignull:
Thanks for reviewing. Sorry just had to fix some tests that were using that function.

extensions/ql-vscode/src/common/commands.ts
```diff
@@ -36,6 +36,7 @@ export type SingleSelectionCommandFunction<Item> = (
 // See https://code.visualstudio.com/api/references/commands
 export type BuiltInVsCodeCommands = {
   "markdown.showPreviewToSide": (uri: Uri) => Promise<void>;
+  setContext: (key: string, value: unknown) => Promise<void>;
```
koesie10, line 39: Should we limit the `key` to just `codeQL`? I don't think we have any use cases for using a non-CodeQL key, so something like this should work:

```suggestion
  setContext: (
    key: `${"codeql" | "codeQL"}${string}`,
    value: unknown,
  ) => Promise<void>;
```

koesie10 (approved):
LGTM, I just have 1 optional suggestion to limit the keys we can pass to `setContext`.

extensions/ql-vscode/src/common/commands.ts
robertbrignull: LGTM. All existing uses do indeed start with CodeQL and I agree we don't currently see a need to set other keys, so it would be good to avoid doing it accidentally.

PR #2217, opened 2023-03-23
Convert call sites to use typed commands (part 1)

robertbrignull:
This PR converts some places that call commands to use typed commands. This isn't everything but I'll try to open a few smaller PRs instead of one giant PR. That way it'll be easier to review and easier to adapt thoughts and ideas before getting too deep into things.

Each commit converts a single file, so if you want to review commit-by-commit then that is an option.

This PR involves:
- Documenting new builtin commands whenever we encounter a new one.
  - I'm working out parameter types from the usages we have in the codebase and from the [VS Code docs](https://code.visualstudio.com/api/references/commands).
- Passing either `App` or `AppCommandManager` to wherever it is needed.
  - When deciding between the two, my thought process so far has been that when we're constructing a long-lived class like `VariantAnalaysisManager` then give it the full `App`, but when it's a standalone function that triggers commands but doesn't do anything else then give it just the `AppCommandManager`.
  - I'm happy to do things differently if there's tangible benefit. It's also worth noting we can change arguments to functions easily, so none of these decisions will block progress in the future.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/commands.ts
+++ b/extensions/ql-vscode/src/common/commands.ts
@@ -48,6 +48,7 @@ export type BuiltInVsCodeCommands = {
     title?: string,
     columnOrOptions?: TextDocumentShowOptions,
   ) => Promise<void>;
+  "vscode.open": (uri: Uri) => Promise<void>;
   "vscode.openFolder": (uri: Uri) => Promise<void>;
 };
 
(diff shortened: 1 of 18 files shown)


robertbrignull:
Test should all be fixed now. Things compile for me locally. I've rebased the changes into the original commits, so you can see which changes necessitated which other changes.

extensions/ql-vscode/src/common/commands.ts
```diff
@@ -35,6 +35,7 @@ export type SingleSelectionCommandFunction<Item> = (
 // Builtin commands where the implementation is provided by VS Code and not by this extension.
 // See https://code.visualstudio.com/api/references/commands
 export type BuiltInVsCodeCommands = {
+  "codeQLDatabases.focus": () => Promise<void>;
```
koesie10, line 38: Perhaps add a comment for this command? It's not entirely obvious if you're just looking at the file that this is a command added by VS Code on custom views.

koesie10 (approved):

PR #2220, opened 2023-03-23
Add support for typed commands to CodeQL query

koesie10:
This adds support for detecting the `CommandManager.execute` method in the unique command use query.

This may not be the best way to implement this. There's a method `hasUnderlyingType` on `this.getReceiver().getType()`, but I couldn't really figure out how to get it recognize `CommandManager`. It might be possible if we can construct the type of `CommandManager`, but this will probably include the filepath to the `CommandManager` class, which might not neccessarily be something we want: moving the `CommandManager` class should not require updating the query. I'm very happy to hear other suggestions.

## Checklist

### DIFF
--- a/.github/codeql/queries/unique-command-use.ql
+++ b/.github/codeql/queries/unique-command-use.ql
@@ -72,6 +72,20 @@
  
    override string getCommandName() { result = this.getArgument(0).(StringLiteral).getValue() }
  }
+
+ /**
+  * A usage of a command from the typescript code, by calling `CommandManager.execute`.
+  */
+ class CommandUsageCommandManagerMethodCallExpr extends CommandUsage, MethodCallExpr {
+   CommandUsageCommandManagerMethodCallExpr() {
+     this.getCalleeName() = "execute" and
+     this.getReceiver().getType().unfold().(TypeReference).getTypeName().getName() = "CommandManager" and
+     this.getArgument(0).(StringLiteral).getValue().matches("%codeQL%") and
+     not this.getFile().getRelativePath().matches("extensions/ql-vscode/test/%")
+   }
+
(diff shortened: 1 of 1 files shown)


.github/codeql/queries/unique-command-use.ql
```diff
@@ -72,6 +72,20 @@
...
    override string getCommandName() { result = this.getArgument(0).(StringLiteral).getValue() }
  }
+
+ /**
+  * A usage of a command from the typescript code, by calling `CommandManager.execute`.
+  */
+ class CommandUsageCommandManagerMethodCallExpr extends CommandUsage, MethodCallExpr {
+   CommandUsageCommandManagerMethodCallExpr() {
+     this.getCalleeName() = "execute" and
+     this.getReceiver().getType().unfold().(TypeReference).getTypeName().getName() = "CommandManager" and
```
robertbrignull, line 82:
```suggestion
     this.getReceiver().getType().(TypeAliasReference).getAliasedType().(TypeReference).getTypeName().getName() = "CommandManager" and
```

I think this achieves the same effect. It's a bit more explicit about what kind of types we expect. Though you could argue we want to make it more robust to changes in the typescript code. I don't have any strong opinions about which direction we go.

robertbrignull (approved):
LGTM

I think this works and it finds calls to the command manager.

I've made one suggestion about how to determine the `CommandManager` type which you are free to use or not.

.github/codeql/queries/unique-command-use.ql
koesie10: I do think we want to ensure that if we make a type like `type CommandManagerWithFoo = CommandManager & { foo: string }`, we'd still want to detect it. I think that's what `unfold` allows us to do, but `(TypeAliasReference)` would lose that information.

PR #2222, opened 2023-03-23
Convert call sites to use typed commands (part 2): QueryHistoryManager

robertbrignull:
Following on from https://github.com/github/vscode-codeql/pull/2217, this PR converts some more places to call typed commands, focussing on the `QueryHistoryManager` class because that turned out to need some non-trivial refactoring to make it work.

The problem was because the tests used `tryOpenExternalFile = (QueryHistoryManager.prototype as any).tryOpenExternalFile;` to access and call the `tryOpenExternalFile` method but without actually constructing a `QueryHistoryManager` instance. Somehow this worked for calling the actual method, but it was a problem when we needed to check that it called `this.app.commands.execute` because we couldn't pass in the correct `app`.

To get around this I pulled the `tryOpenExternalFile` method out to a separate file so it's no longer defined on the `QueryHistoryManager` class. This then means we can test it separately and makes it possible to test when using typed commands.

I wasn't quite sure where to put the new file, so if anyone has suggestions I'm very open. I went with the `src/vscode-utils/` directory because it was nearby and it already existed. I think the tests still need to be in `no-workspace` because it calls `window.showTextDocument`, so that made me think that `common/vscode` wouldn't work out.

Each commit can be reviewed independently if desired.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/vscode-utils/external-files.ts
+++ b/extensions/ql-vscode/src/vscode-utils/external-files.ts
@@ -1,4 +1,5 @@
-import { commands, Uri, window } from "vscode";
+import { Uri, window } from "vscode";
+import { AppCommandManager } from "../common/commands";
 import {
   showAndLogExceptionWithTelemetry,
   showBinaryChoiceDialog,
@@ -6,7 +7,10 @@ import {
 import { redactableError } from "../pure/errors";
 import { asError, getErrorMessage, getErrorStack } from "../pure/helpers-pure";
 
-export async function tryOpenExternalFile(fileLocation: string) {
+export async function tryOpenExternalFile(
+  CommandManager: AppCommandManager,
+  fileLocation: string,
+) {
   const uri = Uri.file(fileLocation);
   try {
     await window.showTextDocument(uri, { preview: false });
@@ -25,7 +29,7 @@ the file in the file explorer and dragging it into the workspace.`,
       );
       if (res) {
         try {
-          await commands.executeCommand("revealFileInOS", uri);
+          await CommandManager.execute("revealFileInOS", uri);
         } catch (e) {
           void showAndLogExceptionWithTelemetry(
             redactableError(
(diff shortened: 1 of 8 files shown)


extensions/ql-vscode/src/vscode-utils/external-files.ts
```diff
@@ -0,0 +1,50 @@
...
+import { AppCommandManager } from "../common/commands";
+import {
+  showAndLogExceptionWithTelemetry,
+  showBinaryChoiceDialog,
+} from "../helpers";
+import { redactableError } from "../pure/errors";
+import { asError, getErrorMessage, getErrorStack } from "../pure/helpers-pure";
+
+export async function tryOpenExternalFile(
+  CommandManager: AppCommandManager,
```
koesie10, line 11:
```suggestion
  commandManager: AppCommandManager,
```

extensions/ql-vscode/src/vscode-utils/external-files.ts
```diff
@@ -0,0 +1,50 @@
...
+      const res = await showBinaryChoiceDialog(
+        `VS Code does not allow extensions to open files >50MB. This file
+exceeds that limit. Do you want to open it outside of VS Code?
+
+You can also try manually opening it inside VS Code by selecting
+the file in the file explorer and dragging it into the workspace.`,
+      );
+      if (res) {
+        try {
+          await CommandManager.execute("revealFileInOS", uri);
```
koesie10, line 32:
```suggestion
          await commandManager.execute("revealFileInOS", uri);
```

koesie10 (approved):
This approach makes sense to me, I think `vscode-utils` is a sensible location to move this method to.

extensions/ql-vscode/src/vscode-utils/external-files.ts
robertbrignull: Thanks. It keeps suggesting this capitalization and I accidentally accepted it this time.

PR #2059, opened 2023-02-10
Retry VSCode integration tests

koesie10:
This will call `jest.retryTimes` to retry the VSCode integration tests up to 3 times. This should help with the flaky tests.

Based on #2058

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #2227, opened 2023-03-23
Only retry tests in CI

elenatanasoiu:
We introduced this change to help with [reducing flakiness in CI][1].

This has a slightly different effect locally, where every failed test will output three times.

This in turn makes it harder to read, especially when you have multiple failing tests.

Since the original intent for this behaviour was to be used in CI, I'm proposing we disable it when the CI env variable isn't set.

I've opted to set it for all jobs involving tests, just for consistency. I'm happy to limit it to just the places where it's required.

[1]: https://github.com/github/vscode-codeql/pull/2059

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -199,6 +199,7 @@ jobs:
         run: echo "cli-versions=$(cat ./extensions/ql-vscode/supported_cli_versions.json | jq -rc)" >> $GITHUB_OUTPUT
     outputs:
       cli-versions: ${{ steps.set-variables.outputs.cli-versions }}
+
   cli-test:
     name: CLI Test
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -108,6 +108,8 @@ jobs:
   unit-test:
     name: Unit Test
     runs-on: ${{ matrix.os }}
+    env:
+      CI: true
```
koesie10, line 112: The `CI` definitions should be unnecessary, the `CI` environment variable is [set by default in GitHub Actions](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) (and most other CI system).

elenatanasoiu: Even better!

koesie10 (approved):

PR #2229, opened 2023-03-23
Convert call sites to use typed commands (part 3)

robertbrignull:
Follows on from earlier PRs and converts some more places to call typed commands.

The only non-trivial bit of this PR is around `VariantAnalysisMonitor`:
- This class already took credentials, but in this case I decided to instead provide a full app and moved it to the constructor instead of the `monitorVariantAnalysis` method.
- The tests were asserting that the `codeQL.autoDownloadVariantAnalysisResult` command was called, but then also asserting that the `VariantAnalysisManager. autoDownloadVariantAnalysisResult` method was called. This only works if it's actually executing commands rather than using a mock, and is also not very necessary. I decided to remove these and just check that the command was called, and this made the tests a lot simpler.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/mocks/vscode-mock-gh-api-server.ts
+++ b/extensions/ql-vscode/src/mocks/vscode-mock-gh-api-server.ts
@@ -1,12 +1,5 @@
 import { pathExists } from "fs-extra";
-import {
-  env,
-  ExtensionContext,
-  ExtensionMode,
-  QuickPickItem,
-  Uri,
-  window,
-} from "vscode";
+import { env, QuickPickItem, Uri, window } from "vscode";
 
 import {
   getMockGitHubApiServerScenariosPath,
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/mocks/vscode-mock-gh-api-server.ts
```diff
@@ -27,7 +27,10 @@ export class VSCodeMockGitHubApiServer extends DisposableObject {
   private readonly server: MockGitHubApiServer;
   private readonly config: MockGitHubApiConfigListener;
 
-  constructor(private readonly ctx: ExtensionContext) {
+  constructor(
+    private readonly ctx: ExtensionContext,
+    private readonly app: App,
```
koesie10, line 32: It feels strange to be passing in both the context and the app. I believe both of the properties that this class accesses on the context (`extensionMode` and `extensionUri`) also exist on the app as `mode` and `extensionPath`. I think it would be best to switch to those in this PR as well, even though the change is unrelated.

robertbrignull: That sounds good to me. I didn't really have it clear in my head what the point of `ExtensionContext` was and how it relates to the app, but I see now the app can replace the extension context in some places. The context is stored as a field on the app, so it can act as a platform-independent interface to those values.

I can take a look (not in this PR) to see if there are more places where we pass both types and if it could be simplified.

koesie10 (approved):

PR #2230, opened 2023-03-23
Use undefined instead of Record<string, never> type when unable to activate extension

robertbrignull:
The main aim of this PR is to make activating the extension more type safe and less error prone.

A problem I've had a few times in tests is unexpectedly getting an `undefined` value because the `extension` variable was not a `CodeQLExtensionInterface` like I thought but actually a `Record<string, never>` which we are using as an error type. This means the type system will happily let you access any field on it, but it may be undefined and you can't tell.

The changes to `extension.ts` should be limited only to changing the error type from `Record<string, never>` to `undefined`. I'm hoping not to make any major behavioural changes here. There are some small changes such as around `variantAnalysisViewSerializer.onExtensionLoaded`, but previously we were just passing an `undefined` into the method.

In the tests I introduce `getActivatedExtension` which encapsulates a very common piece of code and adds extra error handling. This means we can make a lot of tests safer, and they should fail with an error message explaining that the extension couldn't be activated, rather than something being undefined somewhere down the line.

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/cli-integration/packaging.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/cli-integration/packaging.test.ts
@@ -42,13 +42,7 @@ describe("Packaging commands", () => {
       .mockResolvedValue(undefined);
 
     const extension = await getActivatedExtension();
-    if ("cliServer" in extension) {
-      cli = extension.cliServer;
-    } else {
-      throw new Error(
-        "Extension not initialized. Make sure cli is downloaded and installed properly.",
-      );
-    }
+    cli = extension.cliServer;
   });
 
   it("should download all core query packs", async () => {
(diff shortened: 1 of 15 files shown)


extensions/ql-vscode/test/vscode-tests/cli-integration/packaging.test.ts
```diff
@@ -41,11 +41,7 @@ describe("Packaging commands", () => {
       .spyOn(helpers, "showAndLogInformationMessage")
       .mockResolvedValue(undefined);
 
-    const extension = await extensions
-      .getExtension<CodeQLExtensionInterface | Record<string, never>>(
-        "GitHub.vscode-codeql",
-      )!
-      .activate();
+    const extension = await getActivatedExtension();
     if ("cliServer" in extension) {
```
koesie10, line 45: Can we remove these checks now? It seems like the `cliServer` should always be present, otherwise `getActivatedExtension` would have thrown an error. It seems like some other tests have this same pattern as well.

koesie10 (approved):
Thanks for cleaning this up!

extensions/ql-vscode/test/vscode-tests/cli-integration/packaging.test.ts
robertbrignull: Good point! I somehow missed these checks, but they can definitely be removed. I'll also update the error message from `getActivatedExtension` to be more like this one here.

PR #2233, opened 2023-03-24
Convert cleanDatabases to call typed commands

robertbrignull:
The goal here is to remove a use of `commands.executeCommand` and instead use typed commands. The difference here from normal is that we actually need to call the command implementation, rather than providing a mocked implementation. This `cleanDatabases` method is used from a few tests (never from production code) so it's the effects of calling the command that we want.

I'm not 100% sure why we are using commands here at all. Another idea I explored was calling the underlying methods directly, and I pushed this up as [a branch](https://github.com/github/vscode-codeql/compare/robertbrignull/use_app_commands_4_alt). Unfortunately that ran into confusing errors saying `removeAllDatabases is not a function`, which I think means there's something screwy going on when constructing the `DatabaseManager` object but I was struggling to solve it. So I've opened this PR with a simpler but maybe less idealistic solution instead.

What do you think of this? Is this acceptable, or should I have another go at bugfixing that other branch (linked above)?

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/global.helper.ts
+++ b/extensions/ql-vscode/test/vscode-tests/global.helper.ts
@@ -1,11 +1,12 @@
 import { join } from "path";
 import { load, dump } from "js-yaml";
 import { realpathSync, readFileSync, writeFileSync } from "fs-extra";
-import { commands, extensions } from "vscode";
+import { extensions } from "vscode";
 import { DatabaseManager } from "../../src/local-databases";
 import { CodeQLCliServer } from "../../src/cli";
 import { removeWorkspaceRefs } from "../../src/variant-analysis/run-remote-query";
 import { CodeQLExtensionInterface } from "../../src/extension";
+import { createVSCodeCommandManager } from "../../src/common/vscode/commands";
 
 // This file contains helpers shared between tests that work with an activated extension.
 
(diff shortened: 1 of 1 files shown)


norascheuch (approved):

koesie10 (approved):
LGTM if we're not able to fix the error in the other branch. I was able to reproduce the issue in the other branch once, but on subsequent runs it worked without any changes. Could you perhaps create a PR for that branch as well just to see whether the Actions build succeeds?

robertbrignull:
I've opened the other branch as a PR: https://github.com/github/vscode-codeql/pull/2234

robertbrignull:
Hmm, the tests all pass on the other PR. That makes sense to me since I'd expect the error would only happen when there are databases to delete.

So unfortunately I'm not sure this helps us resolve the issue.

koesie10:
> Hmm, the tests all pass on the other PR. That makes sense to me since I'd expect the error would only happen when there are databases to delete.
> 
> So unfortunately I'm not sure this helps us resolve the issue.

Do you know how to reproduce the issue locally? The error message you mentioned in the description above (`removeAllDatabases is not a function`) happens before checking whether there are any databases, so I wouldn't expect it to depend on whether there are any databases to delete.

robertbrignull:
I think let's close this and go with https://github.com/github/vscode-codeql/pull/2234 instead, assuming the bug doesn't reappear on that PR.

PR #2234, opened 2023-03-24
(TEST PR) Convert cleanDatabases to not use commands

robertbrignull:
See https://github.com/github/vscode-codeql/pull/2233 for more context.

Just want to test this out running in CI on actions to see if it works.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/local-databases.ts
+++ b/extensions/ql-vscode/src/local-databases.ts
@@ -1001,6 +1001,15 @@ export class DatabaseManager extends DisposableObject {
     });
   }
 
+  public async removeAllDatabases(
+    progress: ProgressCallback,
+    token: vscode.CancellationToken,
+  ) {
+    for (const item of this.databaseItems) {
+      await this.removeDatabaseItem(progress, token, item);
+    }
+  }
+
   private async deregisterDatabase(
     progress: ProgressCallback,
     token: vscode.CancellationToken,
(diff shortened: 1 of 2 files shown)


alexet (dismissed):

koesie10 (approved):
LGTM, my integration tests pass locally as well.

I did get the error once (`TypeError: Cannot read properties of undefined (reading 'removeAllDatabases')`), but that is because I didn't have the `TEST_CODEQL_PATH` environment variable set. Perhaps we can just make the `removeAllDatabases` call optional by using `databaseManager?.removeAllDatabases`?

elenatanasoiu:
> but that is because I didn't have the TEST_CODEQL_PATH environment variable set

@koesie10 I was expecting [that env var to be set by the jest config](https://github.com/github/vscode-codeql/blob/23f783beefb7a2cdff842dccbfc41fcda9136f11/extensions/ql-vscode/test/vscode-tests/cli-integration/jest-runner-vscode.config.js#L23), but now I see the docs do explicitly tell you to set it. TIL.

elenatanasoiu:
That seems a bit harsh, since you need to know to set it in your bash profile/zsh config in order to persist it over multiple sessions.

robertbrignull:
> Perhaps we can just make the `removeAllDatabases` call optional by using `databaseManager?.removeAllDatabases`?

I'm not really sure exactly what this code is doing or how essential it is, so I was defaulting to trying to avoid changing behaviour and assuming it is important. I'd be prepared to merge as it is but then come back to it if it causes errors for people or we think we can make the call optional? Does that sound ok?

> but that is because I didn't have the TEST_CODEQL_PATH environment variable set

Interesting. I do have that set in the zsh config, so I don't know if that's what was causing the errors for me or not. Maybe it's not set when running from vscode. But I also can't reproduce the error when running in vscode either now, so 🤷

PR #2235, opened 2023-03-24
Convert call sites to use typed commands (part 5): VariantAnalysisView

robertbrignull:
Converts `VariantAnalysisView` to call typed commands.

This PR is simpler than I originally expected. I pulled this out to a separate PR because I thought it was going to be hard, but in the meantime someone has fixed the main issue which is that the app wasn't available when we constructed `VariantAnalysisViewSerializer`. But now we create the app much earlier in activation, so this PR is very simple.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -358,7 +358,10 @@ export async function activate(
     ),
   );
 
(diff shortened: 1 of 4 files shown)


koesie10 (approved):
LGTM. Ideally we wouldn't need to pass both `ctx` and `app` to the view, but this is a tech-debt issue relating to `AbstractWebview` requiring the `ctx`, so I don't think it's something we should fix now.

PR #2237, opened 2023-03-24
Convert call sites to use typed commands (part 6): createVSCodeCommandManager in tests

robertbrignull:
We have quite a few tests that trigger commands as part of them. They do this to call code that isn't accessible otherwise. The easiest way to convert these tests to use typed commands is to use `createVSCodeCommandManager` to get a real command manager than will trigger real commands. By doing this the the behaviour should be exactly the same except that the commands are type checked.

It would likely be better overall if these tests were able to call code without using commands. That would likely lead to better organised code. However, that is more of a long term goal.

Does this make sense? Or should we put in more effort to refactor the tests? Or something else entirely?

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/activated-extension/databases/db-panel.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/activated-extension/databases/db-panel.test.ts
@@ -16,12 +16,13 @@ import { createRemoteSystemDefinedListDbItem } from "../../../factories/db-item-
 import { DbConfigStore } from "../../../../src/databases/config/db-config-store";
 import { getActivatedExtension } from "../../global.helper";
 import { createVSCodeCommandManager } from "../../../../src/common/vscode/commands";
+import { AllCommands } from "../../../../src/common/commands";
 
 jest.setTimeout(60_000);
 
 describe("Db panel UI commands", () => {
   let storagePath: string;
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/test/vscode-tests/activated-extension/databases/db-panel.test.ts
```diff
@@ -15,11 +15,13 @@ import { createDbTreeViewItemSystemDefinedList } from "../../../../src/databases
 import { createRemoteSystemDefinedListDbItem } from "../../../factories/db-item-factories";
 import { DbConfigStore } from "../../../../src/databases/config/db-config-store";
 import { getActivatedExtension } from "../../global.helper";
+import { createVSCodeCommandManager } from "../../../../src/common/vscode/commands";
 
 jest.setTimeout(60_000);
 
 describe("Db panel UI commands", () => {
   let storagePath: string;
+  const commandManager = createVSCodeCommandManager();
```
koesie10, line 24: I believe we need to specify the type of the commands to ensure we actually have typing in tests as well:

```suggestion
  const commandManager = createVSCodeCommandManager<AllCommands>();
```

koesie10 (commented):
This approach makes sense to me, I don't think we should refactor all tests to call directly into the methods right now.

PR #2244, opened 2023-03-27
Refactor local query evaluation to prepare for debug adapter

dbartol:
_Review tip: This PR will probably make more sense if you step through the code. I suggest that you do so after you've read the PR description, but before you try to make sense of the diffs. Set a breakpoint at `compileAndRunQuery()`, select the `Run query against a single database...` command, and follow along._

This PR refactors our existing code path for local query evaluation so that it can be used by the CodeQL Debug Adapter, which will be added in an upcoming PR.

Our existing code path for local query evaluation goes through several layers of code, but there is no clear separation of concerns between those layers. As one extreme example, the `LocalQueryInfo` object, which handles the UI presentation of the query evaluation in the query history view, is passed all the way down the same function that directly invokes the query server. The same goes for the `InitialQueryInfo` object, which has a couple properties that only apply to the UI. The `QueryEvaluationInfo` object, also passed all the way down, mixed some very handy functions for determining the paths to various files in the query's output directory with a bunch of properties and functions used in the UI, and even serialized to disk for query history persistence. The same goes for `DatabaseItem`.

The separate code paths to handle the new query server vs. the legacy query server has quite a bit of duplicated and almost-duplicated code, which, due to the abovementioned comingling of concerns, was difficult to factor out.

The comingling of UI and non-UI responsibilities also makes it more difficult for the code paths that invoke the query server but do not populate the query history (example: AST generation) to evaluate a query, because they have to create extra UI-related state that will never be used.

The changes in this PR thoroughly refactor the whole code path to separate concerns. Specifically, the `QueryRunner` abstract class and everything below it are now focused on the mechanics of the actual evaluation via the query server, with no connection to the UI other than a progress callback and a `Logger` inteface. Everything else, including query history UI, timestamp file generation, and evaluator log summarization, has been hoisted out of `QueryRunner` and its implementations.

## Details

The two implementations of `QueryRunner` now focus solely on evaluating the query. In particular, the implementation of `compileAndRunQueryAgainstDatabaseCore()` for the new query server does nothing but set up the message to send to the query server, send the message, and then return the results. The implementation for the legacy query server does much more work, but only because the legacy query server interface requires it to. The evaluation is initialized by calling `createQueryRun()` on `QueryRunner`. This holds information that will be needed even before the evaluation completes, like the path to the evaluation's output directory. The actual evaluation is invoked by calling `evaluate()` on the `CoreQueryRun` object, which will evantually return the evaluation results.

The interface to `QueryRunner` now defines its own minimal interfaces to specify the arguments to the evaluation (query, database, additional search paths) and the results of the evaluation, rather than reusing the UI-related types from the higher layers.

I left the layout and functionality of most of the UI-related types (`LocalQueryInfo`, etc.) alone, because those are serialized directly to disk and deserialized from JSON by slapping a new prototype on the raw JSON object. We could do better than this, but improving this wasn't necessary for this PR.

I factored all the properties of `QueryEvaluationInfo` that just computed output file paths out into a new base class `QueryOutputDir`. This new class is now used throughout the lower-level code to find paths, without depending on other state that is only necessary for results interpretation.

The top-level command handling for the various `Run query...` commands had recently been factored out into a separate `local-queries.ts` file, instead of being a bunch of ugly nested functions within the extension initialization code (thanks, @koesie10!). I've converted that file to a full on `LocalQueries` service, in the pattern used by most of our other components. All of the UI-related code that used to be scattered throughout the query evaluation code path is now here. All of the UI setup that's done before query evaluation starts (picking an output directory, creating a cancellation token source, etc.) is done via `createLocalQueryRun()`, and all of the post-evaluation updates are done by calling `complete()` on the `LocalQueryRun` object. As the comment in the code indicates, this looks a little clunky as-is, but it will be necessary when we try to hook it up to the debug adapter.

The diffs in `local-queries.ts` look scarier than they are. I didn't actually change any of the commands except to put them into the new `LocalQueries` class. They all just go through `compileAndRunQuery()`, which is where the real changes are.

The AST and Go To Reference code now talks directly to the `QueryRunner` without having to manufacture the various UI-related state that it doesn't need.

I separated the existing `Logger` interface into `BaseLogger`, with just the `log()` function, and `Logger`, which adds the `show()` function. Code that only generates log messages needs only `BaseLogger`. The distinction isn't critical in this PR, but becomes more important when I add the debug adapter, which will need a `BaseLogger` that routes the messages over the Debug Adapter Protocol, but has no way to implement the `show()` function.

### DIFF
--- a/extensions/ql-vscode/src/contextual/queryResolver.ts
+++ b/extensions/ql-vscode/src/contextual/queryResolver.ts
@@ -13,11 +13,10 @@ import {
 import { KeyType, kindOfKeyType, nameOfKeyType, tagOfKeyType } from "./keyType";
 import { CodeQLCliServer } from "../cli";
 import { DatabaseItem } from "../local-databases";
-import { extLogger } from "../common";
-import { createInitialQueryInfo } from "../run-queries-shared";
-import { CancellationToken, Uri } from "vscode";
+import { extLogger, TeeLogger } from "../common";
+import { CancellationToken } from "vscode";
 import { ProgressCallback } from "../progress";
-import { QueryRunner } from "../queryRunner";
+import { CoreCompletedQuery, QueryRunner } from "../queryRunner";
 import { redactableError } from "../pure/errors";
 import { QLPACK_FILENAMES } from "../pure/ql";
 
@@ -169,32 +168,30 @@ export async function runContextualQuery(
   progress: ProgressCallback,
   token: CancellationToken,
   templates: Record<string, string>,
-) {
+): Promise<CoreCompletedQuery> {
   const { packPath, createdTempLockFile } = await resolveContextualQuery(
     cli,
     query,
   );
-  const initialInfo = await createInitialQueryInfo(
-    Uri.file(query),
-    {
-      name: db.name,
-      databaseUri: db.databaseUri.toString(),
-    },
+  const queryRun = qs.createQueryRun(
+    db.databaseUri.fsPath,
+    { queryPath: query, quickEvalPosition: undefined },
     false,
+    getOnDiskWorkspaceFolders(),
+    queryStorageDir,
+    undefined,
+    templates,
   );
   void extLogger.log(
-    `Running contextual query ${query}; results will be stored in ${queryStorageDir}`,
+    `Running contextual query ${query}; results will be stored in ${queryRun.outputDir}`,
   );
-  const queryResult = await qs.compileAndRunQueryAgainstDatabase(
-    db,
-    initialInfo,
-    queryStorageDir,
+  const results = await queryRun.evaluate(
     progress,
     token,
-    templates,
+    new TeeLogger(qs.logger, queryRun.outputDir.logPath),
   );
   if (createdTempLockFile) {
     await removeTemporaryLockFile(packPath);
   }
-  return queryResult;
+  return results;
 }
--- a/extensions/ql-vscode/src/contextual/queryResolver.ts
+++ b/extensions/ql-vscode/src/contextual/queryResolver.ts
@@ -183,7 +183,7 @@ export async function runContextualQuery(
     templates,
   );
   void extLogger.log(
-    `Running contextual query ${query}; results will be stored in ${queryRun.outputDir}`,
+    `Running contextual query ${query}; results will be stored in ${queryRun.outputDir.querySaveDir}`,
   );
   const results = await queryRun.evaluate(
     progress,
--- a/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
+++ b/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
@@ -27,7 +27,7 @@ import { redactableError } from "../pure/errors";
 import { CoreQueryResults, CoreQueryTarget } from "../queryRunner";
 import { Position } from "../pure/messages-shared";
 
-async function compileQuery(
+export async function compileQuery(
   qs: qsClient.QueryServerClient,
   program: messages.QlProgram,
   quickEvalPosition: Position | undefined,
@@ -185,7 +185,7 @@ export class QueryInProgress {
     readonly querySaveDir: string,
     readonly dbItemPath: string,
     databaseHasMetadataFile: boolean,
-    readonly queryDbscheme: string, // the dbscheme file the query expects, based on library path resolution
+    readonly queryDbscheme: string, // the dbscheme file the query expects, ba`sed on library path resolution
     readonly quickEvalPosition?: messages.Position,
     readonly metadata?: QueryMetadata,
     readonly templates?: Record<string, string>,
--- a/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
+++ b/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
@@ -292,7 +292,7 @@ function translateLegacyResult(
       break;
     case messages.QueryResultType.TIMEOUT:
       // This is the only legacy result type that doesn't exist for the new query server. Format the
-      // messasge here, and let the later code treat is as `OTHER_ERROR`.
+      // messasge here, and let the later code treat it as `OTHER_ERROR`.
       newResultType = newMessages.QueryResultType.OTHER_ERROR;
       newMessage = `timed out after ${Math.round(
         legacyResult.evaluationTime / 1000,
--- a/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
+++ b/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
@@ -9,11 +9,7 @@ import {
   DatabaseItem,
   DatabaseResolver,
 } from "../local-databases";
-import {
-  showAndLogExceptionWithTelemetry,
-  showAndLogWarningMessage,
-  upgradesTmpDir,
-} from "../helpers";
+import { showAndLogExceptionWithTelemetry, upgradesTmpDir } from "../helpers";
 import { ProgressCallback } from "../progress";
 import { QueryMetadata } from "../pure/interface-types";
 import { extLogger, Logger } from "../common";
@@ -141,11 +137,6 @@ async function runQuery(
   };
   try {
     await qs.sendRequest(messages.runQueries, params, token, progress);
-    if (qs.config.customLogDirectory) {
-      void showAndLogWarningMessage(
-        `Custom log directories are no longer supported. The "codeQL.runningQueries.customLogDirectory" setting is deprecated. Unset the setting to stop seeing this message. Query logs saved to ${logPath}.`,
-      );
-    }
   } finally {
     qs.unRegisterCallback(callbackId);
     if (
--- a/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
+++ b/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
@@ -292,7 +292,7 @@ function translateLegacyResult(
       break;
     case messages.QueryResultType.TIMEOUT:
       // This is the only legacy result type that doesn't exist for the new query server. Format the
-      // messasge here, and let the later code treat it as `OTHER_ERROR`.
+      // message here, and let the later code treat it as `OTHER_ERROR`.
       newResultType = newMessages.QueryResultType.OTHER_ERROR;
       newMessage = `timed out after ${Math.round(
         legacyResult.evaluationTime / 1000,
--- a/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
+++ b/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
@@ -445,9 +445,7 @@ export async function compileAndRunQueryAgainstDatabaseCore(
         void showAndLogExceptionWithTelemetry(error);
       }
 
-      return {
-        ...translateLegacyResult(result),
-      };
+      return translateLegacyResult(result);
     } else {
       // Error dialogs are limited in size and scrollability,
       // so we include a general description of the problem,
--- a/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
+++ b/extensions/ql-vscode/src/legacy-query-server/run-queries.ts
@@ -1,28 +1,173 @@
 import * as tmp from "tmp-promise";
-import { basename, join } from "path";
+import { basename } from "path";
 import { CancellationToken, Uri } from "vscode";
 import { LSPErrorCodes, ResponseError } from "vscode-languageclient";
 
 import * as cli from "../cli";
-import { DatabaseItem } from "../local-databases";
 import {
-  getOnDiskWorkspaceFolders,
-  showAndLogErrorMessage,
+  DatabaseContentsWithDbScheme,
+  DatabaseItem,
+  DatabaseResolver,
+} from "../local-databases";
+import {
   showAndLogExceptionWithTelemetry,
   showAndLogWarningMessage,
-  tryGetQueryMetadata,
   upgradesTmpDir,
 } from "../helpers";
 import { ProgressCallback } from "../progress";
 import { QueryMetadata } from "../pure/interface-types";
-import { extLogger, Logger, TeeLogger } from "../common";
+import { extLogger, Logger } from "../common";
 import * as messages from "../pure/legacy-messages";
-import { InitialQueryInfo, LocalQueryInfo } from "../query-results";
+import * as newMessages from "../pure/new-messages";
 import * as qsClient from "./queryserver-client";
 import { asError, getErrorMessage } from "../pure/helpers-pure";
 import { compileDatabaseUpgradeSequence } from "./upgrades";
-import { QueryEvaluationInfo, QueryWithResults } from "../run-queries-shared";
+import { QueryEvaluationInfo, QueryOutputDir } from "../run-queries-shared";
 import { redactableError } from "../pure/errors";
+import { CoreQueryResults, CoreQueryTarget } from "../queryRunner";
+import { Position } from "../pure/messages-shared";
+
+async function compileQuery(
+  qs: qsClient.QueryServerClient,
+  program: messages.QlProgram,
+  quickEvalPosition: Position | undefined,
+  outputDir: QueryOutputDir,
+  progress: ProgressCallback,
+  token: CancellationToken,
+  logger: Logger,
+): Promise<messages.CompilationMessage[]> {
+  let compiled: messages.CheckQueryResult | undefined;
+  try {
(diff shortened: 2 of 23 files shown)


koesie10 (commented):
I've only done a partial review of the code by looking at the diff, but haven't matched the old code to the new code. These are just some things that I spotted while looking through it.

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -120,9 +118,29 @@ export class QueryEvaluationInfo {
     return findQueryEvalLogEndSummaryFile(this.querySaveDir);
   }
 
+  get bqrsPath() {
+    return join(this.querySaveDir, "results.bqrs");
+  }
+}
+
+export class QueryEvaluationInfo extends QueryOutputDir {
```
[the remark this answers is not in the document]
dbartol: Normally I would do just that, but see the comment I just added for why we can't in this case.

extensions/ql-vscode/src/legacy-query-server/run-queries.ts
```diff
@@ -381,80 +390,71 @@ export async function compileAndRunQueryAgainstDatabase(
...
-      const message = formatLegacyMessage(result);
 
       return {
-        query: query.queryEvalInfo,
-        message,
-        result,
-        successful: result.resultType === messages.QueryResultType.SUCCESS,
-        logFileLocation: result.logFileLocation,
+        ...translateLegacyResult(result),
       };
```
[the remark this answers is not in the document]
dbartol: Oh, weird, must be an artifact of how my changes evolved. Fixed.

extensions/ql-vscode/src/legacy-query-server/run-queries.ts
```diff
@@ -286,34 +270,67 @@ async function compileNonDestructiveUpgrade(
...
+      break;
+    case messages.QueryResultType.CANCELLATION:
+      newResultType = newMessages.QueryResultType.CANCELLATION;
+      break;
+    case messages.QueryResultType.OOM:
+      newResultType = newMessages.QueryResultType.OOM;
+      break;
+    case messages.QueryResultType.TIMEOUT:
+      // This is the only legacy result type that doesn't exist for the new query server. Format the
+      // messasge here, and let the later code treat is as `OTHER_ERROR`.
```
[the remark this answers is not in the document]
dbartol: You meant to s/is/it/, right? Fixed.

extensions/ql-vscode/src/queryRunner.ts
```diff
@@ -54,4 +97,53 @@ export abstract class QueryRunner {
...
+    return {
+      queryTarget: query,
+      dbPath,
+      id: actualId,
+      outputDir,
+      evaluate: async (
+        progress: ProgressCallback,
+        token: CancellationToken,
+        logger: BaseLogger,
+      ): Promise<CoreCompletedQuery> => {
```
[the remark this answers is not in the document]
dbartol: `evaluate()` needs to use several of the parameters already passed to `createQueryRun()`, so one way or another `createQueryRun()` has to return a new object to hold that state. I could make `CoreQueryRun` into a class, and make the state capture explicit. Would that make things more clear overall? Or are you suggesting that `evaluate()` be a member of `QueryRunner` but take a `CoreQueryRun` as a parameter?

extensions/ql-vscode/src/local-queries.ts
```diff
@@ -7,384 +7,560 @@ import {
...
+
+  /**
+   * Updates the UI based on the results of the query evaluation. This creates the evaluator log
+   * summaries, updates the query history item for the evaluation with the results and evaluation
+   * time, and displays the results view.
+   *
+   * This function must be called when the evaluation completes, whether the evaluation was
+   * successful or not.
+   * */
+  public async complete(results: CoreQueryResults): Promise<void> {
```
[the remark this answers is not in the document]
dbartol: I was actually going for "abstract and generic":) My idea was that code that invokes a local query run just knows that it has to call `complete()` when it's done, and it's up to `complete()` to do whatever it is that the UI needs to do when the evaluation completes. Today, that's "save and show query results", but it's up to `local-queries.ts` to decide what to do. Think of it like an `onLocalRunCompleted` event, where the handler chooses to mark the local query item as completed and show the results view.

robertbrignull (commented):
I've done one pass through, but it's a big and complex PR so I haven't really understood everything. I've made some surface-level comments on things I noticed.

Overall I think it looks like a very good direction and I don't have any huge concerns. Decoupling the UI code from the backend execution code will be a massive help. It's just quite complex so it's hard to see the full effect of the changes.

extensions/ql-vscode/src/queryRunner.ts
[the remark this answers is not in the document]
robertbrignull: I also was going to make a comment asking whether the `evaluate` method needs to live on this interface. It's a bit of a shame that we then have to explicitly omit it when defining the `CoreCompletedQuery` interface.

I note there's only one implementation, but I see what you mean about it needing access to state available in `createQueryRun`.  But in the two places where we call `evaluate`, we also called `createQueryRun` earlier in that same method, so the state would still be there and could be passed into `evaluate` as well / instead.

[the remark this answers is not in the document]
dbartol: While it's not apparent in this PR, my motivation for having `CoreQueryRun` as an object with an `evaluate()` function, rather than having two different functions on `QueryRunner` that require several common parameters, was that the debug adapter needs to initialize an evaluation in one function, but then later await it in another function. A `CoreQueryRun` object seemed like the natural way to encapsulate everything the debug adapter would need to later evaluate the query. I'd rather not have to have the debugger capture multiple parameters from the `createQueryRun()` method just to pass them to a later `evaluateQueryRun()` method.

The real problem here is that I want a function that evaluates a query, which is asynchronous, and so would normally return a `Promise<CoreQueryResults>`. However, just _starting_ the evaluation produces some properties, like `outputDir`, that I'll need to use before the promise completes. I could just have a single `runQuery()` function on `QueryRunner`, and have it return something like `Promise<CoreQueryResult> & CoreQueryRun` (note that `CoreQueryRun` is not inside the angle brackets). I'm pretty sure I've seen this done somewhere before. I'm not sure if other developers would find it more or less odd than the separate `evaluate()` function I have now.

I suppose I could just replace the `evaluate` property in `CoreQueryRun` with `results: Promise<CoreQueryResults>`. That seems less confusing than anything I suggest above, and also less confusing than what I originally wrote.

extensions/ql-vscode/src/local-queries.ts
```diff
@@ -7,384 +7,560 @@ import {
...
+    public readonly cliServer: CodeQLCliServer,
+    private readonly databaseUI: DatabaseUI,
+    private readonly localQueryResultsView: ResultsView,
+    private readonly queryStorageDir: string,
+  ) {
+    super();
+  }
+
+  public getCommands(): LocalQueryCommands {
+    const runQuery = async (uri: Uri | undefined) =>
```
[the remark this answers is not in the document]
dbartol: Opened as separate PR, #2245, to avoid making this PR's diffs even more confusing.

extensions/ql-vscode/src/legacy-query-server/run-queries.ts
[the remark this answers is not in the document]
dbartol: Fixed (both)

extensions/ql-vscode/src/local-queries.ts
```diff
@@ -7,384 +7,560 @@ import {
...
+   */
+  public async createLocalQueryRun(
+    selectedQuery: SelectedQuery,
+    dbItem: DatabaseItem,
+    outputDir: QueryOutputDir,
+    tokenSource: CancellationTokenSource,
+  ): Promise<LocalQueryRun> {
+    await createTimestampFile(outputDir.querySaveDir);
+
+    if (this.queryRunner.customLogDirectory) {
```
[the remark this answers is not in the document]
dbartol: Good catch. I've removed redundant one in the legacy code path.

extensions/ql-vscode/src/contextual/queryResolver.ts
```diff
@@ -169,32 +168,30 @@ export async function runContextualQuery(
...
+    { queryPath: query, quickEvalPosition: undefined },
     false,
+    getOnDiskWorkspaceFolders(),
+    queryStorageDir,
+    undefined,
+    templates,
   );
   void extLogger.log(
-    `Running contextual query ${query}; results will be stored in ${queryStorageDir}`,
+    `Running contextual query ${query}; results will be stored in ${queryRun.outputDir}`,
```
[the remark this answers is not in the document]
dbartol: Fixed.

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -642,36 +587,113 @@ async function convertToQlPath(filePath: string): Promise<string> {
...
+    humanReadableSummary,
+    endSummary,
+    jsonSummary,
+    summarySymbols,
+  };
+}
+
+/**
+ * Calls the appropriate CLI command to generate a human-readable log summary.
+ * @param qs The query server client.
```
aeisenberg, line 656:
```suggestion
 * @param cliServer The cli server client.
```

extensions/ql-vscode/src/legacy-query-server/run-queries.ts
```diff
@@ -1,28 +1,164 @@
...
+  DatabaseResolver,
+} from "../local-databases";
+import { showAndLogExceptionWithTelemetry, upgradesTmpDir } from "../helpers";
 import { ProgressCallback } from "../progress";
 import { QueryMetadata } from "../pure/interface-types";
-import { extLogger, Logger, TeeLogger } from "../common";
+import { extLogger, Logger } from "../common";
 import * as messages from "../pure/legacy-messages";
-import { InitialQueryInfo, LocalQueryInfo } from "../query-results";
+import * as newMessages from "../pure/new-messages";
```
aeisenberg, line 17: Instead of importing from newMessages here, can you put the shared messages in `pure/messages-shared`?

extensions/ql-vscode/src/legacy-query-server/run-queries.ts
```diff
@@ -327,17 +335,15 @@ export async function compileAndRunQueryAgainstDatabase(
...
       `Query schema was ${querySchemaName}, but database schema was ${dbSchemaName}.`,
     );
     throw new Error(
       `The query ${basename(
-        initialInfo.queryPath,
-      )} cannot be run against the selected database (${
-        dbItem.name
-      }): their target languages are different. Please select a different database and try again.`,
+        query.queryPath,
+      )} cannot be run against the selected database: their target languages are different. Please select a different database and try again.`,
```
aeisenberg, line 346: Minor: Would be nice to continue using the database name here, but not sure if that value is available with this refactoring.

extensions/ql-vscode/src/local-queries.ts
```diff
@@ -7,384 +7,575 @@ import {
...
+        async (progress, token) =>
+          displayQuickQuery(
+            this.app,
+            this.cliServer,
+            this.databaseUI,
+            progress,
+            token,
+          ),
+        {
+          title: "Run Quick Query",
```
aeisenberg, line 357: Do we want this?

```suggestion
          title: "Run Quick Query",
          cancellable: true,
```

extensions/ql-vscode/src/local-queries.ts
```diff
@@ -7,384 +7,575 @@ import {
...
+        );
+
+        await localQueryRun.complete(results);
+
+        return results;
+      } catch (e) {
+        // It's odd that we have two different ways for a query evaluation to fail: by throwing an
+        // exception, and by returning a result with a failure code. This is how the code worked
+        // before the refactoring, so it's been preserved, but we should probably figure out how
+        // to unify both error handling paths.
```
aeisenberg, line 497: I think this is because it depends on where the error happens. If the error happens in the query server, then we get a result with a failure code. It can also happen that the error happens in the IDE, in which case, we need to handle the thrown exception.

It would be nice to figure out a way to unify.

extensions/ql-vscode/src/local-queries.ts
```diff
@@ -7,384 +7,575 @@ import {
...
-          qhm,
-          databaseUI,
-          localQueryResultsView,
-          queryStorageDir,
-          false,
-          uri,
-          progress,
-          token,
-          item.databaseItem,
+    // If possible, only show databases with the right language (otherwise show all databases).
```
aeisenberg, line 519: Is this comment incorrect? It looks like we show an error if there are no suitable databases found. I'm not sure we want to show all databases.

extensions/ql-vscode/src/local-queries.ts
```diff
@@ -7,384 +7,575 @@ import {
...
+        } catch (e) {
+          skippedDatabases.push(item.label);
+          errors.push(getErrorMessage(e));
+        }
+      }
+      if (skippedDatabases.length > 0) {
+        void extLogger.log(`Errors:\n${errors.join("\n")}`);
+        void showAndLogWarningMessage(
+          `The following databases were skipped:\n${skippedDatabases.join(
+            "\n",
```
aeisenberg, line 566: The popup box is not very large. I don't think it can handle more than 3-4 lines. Have you tried it when there are lots of skipped DBs? Maybe better to concatenate by a `,`.

extensions/ql-vscode/src/queryRunner.ts
aeisenberg, line 1: Not related to this PR, but it is strange that this is `queryRunner.ts` and there is a `query-runner.ts` just above. We should probably move our ts modules to `-` case.

extensions/ql-vscode/src/queryRunner.ts
```diff
@@ -54,4 +97,53 @@ export abstract class QueryRunner {
...
+   * query evaluation that is known even before evaluation starts, including the unique ID of the
+   * evaluation and the path to its output directory.
+   */
+  public createQueryRun(
+    dbPath: string,
+    query: CoreQueryTarget,
+    generateEvalLog: boolean,
+    additionalPacks: string[],
+    queryStorageDir: string,
+    id: string | undefined,
```
aeisenberg, line 113: Minor: is this cleaner than what you have? And then you can remove `actualId` below.

```suggestion
    id: string = `${basename(query.queryPath)}-${nanoid()}`,
```

aeisenberg (commented):
At a high level, this is a good refactoring. I added a bunch of comments. I see no major issues with this.

extensions/ql-vscode/src/legacy-query-server/run-queries.ts
dbartol: The only symbols we reference via `newMessages` are the `QueryResultType` status codes. We translate between the legacy ones and the new ones, and expose only the new ones via the `QueryRunner`.

extensions/ql-vscode/src/legacy-query-server/run-queries.ts
dbartol: If this were the new query server, I probably would have put more effort into that. Since it's just for the legacy server, it didn't seem worthwhile.

extensions/ql-vscode/src/local-queries.ts
dbartol: My first reponse was "yes, obvious oversight". However, looking at the implementation, the command seems to just show the Quick Query editor, rather than actually running the query. That doesn't seem like we need a "cancel" button.

extensions/ql-vscode/src/local-queries.ts
dbartol: I think it's probably easy enough to unify, but I didn't want to attempt that in this PR.

extensions/ql-vscode/src/local-queries.ts
dbartol: I think the `findLanguage()` call below can return `undefined` if the user cancels the prompt to select a language, and in that case the filtering by language (and the error message) won't happen.

extensions/ql-vscode/src/local-queries.ts
dbartol: This is not new code, although the diff view might make it look like it is. I only changed `compiledAndRunQuery()` and its callees.

aeisenberg (approved):
Minor comment reformatting, but looks good otherwise. I have not tried this out locally, so I assume this works for you.

aeisenberg (approved):

PR #2245, opened 2023-03-28
Move `LocalQueries` commands to be member functions

dbartol:
This PR implements a [review suggestion](https://github.com/github/vscode-codeql/pull/2244#discussion_r1150312991) from #2244. I've made it a separate, dependent PR to avoid making the original PR's diffs even worse.

### DIFF
--- a/extensions/ql-vscode/src/local-queries.ts
+++ b/extensions/ql-vscode/src/local-queries.ts
@@ -225,151 +225,154 @@ export class LocalQueries extends DisposableObject {
   }
 
   public getCommands(): LocalQueryCommands {
-    const runQuery = async (uri: Uri | undefined) =>
-      withProgress(
-        async (progress, token) => {
-          await this.compileAndRunQuery(false, uri, progress, token, undefined);
-        },
-        {
-          title: "Running query",
-          cancellable: true,
-        },
-      );
-
-    const runQueryOnMultipleDatabases = async (uri: Uri | undefined) =>
-      withProgress(
-        async (progress, token) =>
-          await this.compileAndRunQueryOnMultipleDatabases(
-            progress,
-            token,
-            uri,
-          ),
-        {
-          title: "Running query on selected databases",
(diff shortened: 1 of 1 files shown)


aeisenberg:
Just wanting to be clear. `[dbartol/debug-adapter](https://github.com/github/vscode-codeql/tree/dbartol/debug-adapter)` has no PR associated with it. Is this PR targeting the right branch?

extensions/ql-vscode/test/vscode-tests/no-workspace/contextual/astBuilder.test.ts
```diff
@@ -52,11 +53,12 @@ describe("AstBuilder", () => {
     const astBuilder = createAstBuilder();
     const roots = await astBuilder.getRoots();
 
+    const bqrsPath = path.normalize("/a/b/c/results.bqrs");
```
aeisenberg, line 56: Any understanding of why this test had to be changed?

dbartol:
> Just wanting to be clear. `[dbartol/debug-adapter](https://github.com/github/vscode-codeql/tree/dbartol/debug-adapter)` has no PR associated with it. Is this PR targeting the right branch?

Oh, yikes! That's the wrong target branch. I'll fix this in a few minutes.

extensions/ql-vscode/test/vscode-tests/no-workspace/contextual/astBuilder.test.ts
dbartol: I think that change went into the base PR, but the reason for the change was that previously, the test code passed in `"/a/b/c"` as the path to the BQRS file itself, which got passed through the implementation verbatim and came out the other end with forward slashes intact. With my changes, `"/a/b/c"` is treated as the query output directory, which goes through some `path` calls (`join`, etc.) before the BQRS path comes out the other end with platform-specific separators.

PR #2246, opened 2023-03-28
Move extension pack resolution into `LocalQueries`

dbartol:
This way the debug configuration will be able to control it, since the debug configuration won't go through `compileAndRunQueryInternal`


extensions/ql-vscode/src/contextual/queryResolver.ts
```diff
@@ -178,6 +178,7 @@ export async function runContextualQuery(
     { queryPath: query, quickEvalPosition: undefined },
     false,
     getOnDiskWorkspaceFolders(),
+    undefined,
```
aeisenberg, line 181: Do we want the possibility of running with extension packs here? The internal CFG query could give better results with extension packs.

extensions/ql-vscode/src/local-queries.ts
```diff
@@ -455,14 +455,20 @@ export class LocalQueries extends DisposableObject {
       throw new Error("Can't run query without a selected database");
     }
 
+    const additionalPacks = getOnDiskWorkspaceFolders();
+    const extensionPacks = (await this.cliServer.useExtensionPacks())
```
aeisenberg, line 459: If someone happens to be using the old query server, what's to prevent extension packs from being passed in?

extensions/ql-vscode/src/contextual/queryResolver.ts
dbartol: How are the internal CFG query's results affected by extension packs?

aeisenberg: Since the graph is for internal use only (and only barely supported), we can probably avoid adding extension packs until someone asks for it.

extensions/ql-vscode/src/legacy-query-server/run-queries.ts
```diff
@@ -308,12 +308,17 @@ export async function compileAndRunQueryAgainstDatabaseCore(
...
+  extensionPacks: string[] | undefined,
   outputDir: QueryOutputDir,
   progress: ProgressCallback,
   token: CancellationToken,
   templates: Record<string, string> | undefined,
   logger: Logger,
 ): Promise<CoreQueryResults> {
+  if (extensionPacks !== undefined && extensionPacks.length > 0) {
+    throw new Error("Legacy query server does not support extension packs.");
+  }
```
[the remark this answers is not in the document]
dbartol: A warning seems reasonable. Done.

extensions/ql-vscode/src/local-queries.ts
dbartol: Nothing prevents them from being passed in, but they'll get the ~~error~~ warning from the comment below. Also, since the old query server doesn't support data extensions at all, presumably any use of extension packs would either fail (if the query actually references an extensible predicate) or would be a no-op (if the query doesn't reference an extensible predicate), right?

extensions/ql-vscode/src/legacy-query-server/run-queries.ts
```diff
@@ -308,12 +312,19 @@ export async function compileAndRunQueryAgainstDatabaseCore(
...
+  extensionPacks: string[] | undefined,
   outputDir: QueryOutputDir,
   progress: ProgressCallback,
   token: CancellationToken,
   templates: Record<string, string> | undefined,
   logger: Logger,
 ): Promise<CoreQueryResults> {
+  if (extensionPacks !== undefined && extensionPacks.length > 0) {
+    await showAndLogWarningMessage(
+      "Legacy query server does not support extension packs.",
```
aeisenberg, line 324: Possibly add a message here to suggest changing the `extensionPacks` configuration property to `none`.

Though, so few people are using the legacy query server now that this is probably not a big deal.

aeisenberg (approved):
Consider updating the warning message, but it's fine either way.

extensions/ql-vscode/src/legacy-query-server/run-queries.ts
dbartol: I think we can probably leave it as-is. Once I add the debug configuration in my next PR, I'd just have to update the warning again with more detail about changing the `extensionPacks:` property in `launch.json`, so let's just hope that nobody who is still using the legacy query server is also on the bleeding edge of extension pack development:)

PR #2250, opened 2023-03-30
Adapt skeleton pack generation to work with new queries panel

elenatanasoiu:
Please review this PR commit-by-commit. 

In [a previous PR](https://github.com/github/vscode-codeql/pull/2036), we implemented a feature where if you add a new database, we generate a skeleton QL pack for you to work with the database. The skeleton pack would contain an example query file and would match the language of the database.

We've found it's hard for the user to discover this workflow in the `codespaces-codeql` repo. 

After some input from design, we want to reverse this flow by starting with generating the query file and skeleton pack first, and then downloading the database. This means the user immediately focuses on the query file instead of having to switch away from the extension and look for the query file in the explorer view. (see videos below for a visual illustration of what this all means)

This new flow should trigger via a command in the extension ("Create Query") which takes you through a wizard and generates everything.

When the new queries panel is ready, we can also hook up this command with the link suggesting how to create a query ("Create one to start"):

Screenshot 2023-03-15 at 11 56 43

Finally, we're making sure that both the `codespaces-codeql` and `vscode-codeql-starter` repos work as expected with this new command:

## codespaces-codeql repo

https://user-images.githubusercontent.com/1354439/228851055-0f3224ab-843e-4f55-8b44-0040957c67da.mov

## vscode-codeql-starter repo

https://user-images.githubusercontent.com/1354439/228851740-fc028334-daed-48df-a00c-2cfce7426d9c.mov


extensions/ql-vscode/src/extension.ts
```diff
@@ -857,6 +857,7 @@
     databaseUI,
     localQueryResultsView,
     queryStorageDir,
+    ctx,
```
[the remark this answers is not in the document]
elenatanasoiu: Good bot. This was already flagged in CI.

elenatanasoiu:
After rebasing off main the tests in this PR started failing due to timeout errors. I don't know if it's because we've moved to React 18 or something else, but I've added a commit to reduce the size of the `beforeEach` to make it take less time. Seems like this has done the trick.

elenatanasoiu:
This PR is ready for review again.

After rebasing off `main`, the tests in this PR [started failing](https://github.com/github/vscode-codeql/actions/runs/4599058629/jobs/8123952131) due to timeout errors:

```
thrown: "Exceeded timeout of 5000 ms for a hook.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

  11 | import fetch from "node-fetch";
  12 |
> 13 | beforeAll(async () => {
     | ^
  14 |   // ensure the test database is downloaded
  15 |   mkdirpSync(dirname(dbLoc));
  16 |   if (!existsSync(dbLoc)) {

at Object.<anonymous> (jest.setup.ts:13:1)
```

 I don't know if it's because we've moved to React 18 or something else, but I've added a commit to reduce the size of the `beforeEach` in the skeleton wizard tests. 
 
 Seems like this has done the trick.

extensions/ql-vscode/src/skeleton-query.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    const databaseItem = await this.databaseManager.digForDatabaseItem(
+      this.language,
+      databaseNwo,
+    );
+
+    if (databaseItem) {
+      // select the found database
+      await this.databaseManager.setCurrentDatabaseItem(
+        databaseItem as DatabaseItem,
```
koesie10:
```suggestion
        databaseItem,
```

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -468,23 +512,31 @@ export async function convertGithubNwoToDatabaseUrl(
 
     const languages = response.data.map((db: any) => db.language);
 
-    const language = await promptForLanguage(languages, progress);
-    if (!language) {
-      return;
+    if (language && languages.includes(language)) {
```
koesie10, line 515: Can we avoid duplicating the logic below by using something like:

```ts
if (!language || !languages.includes(language)) {
  language = await promptForLanguage(languages, progress);
  if (!language) {
    return
  }
}
```

extensions/ql-vscode/src/skeleton-query.ts
```diff
@@ -30,13 +31,14 @@ export class SkeletonQueryWizard {
...
     private readonly cliServer: CodeQLCliServer,
-    private readonly storagePath: string | undefined,
     private readonly progress: ProgressCallback,
     private readonly credentials: Credentials | undefined,
     private readonly extLogger: OutputChannelLogger,
     private readonly databaseManager: DatabaseManager,
     private readonly token: CancellationToken,
-  ) {}
+  ) {
+    this.storagePath = this.workoutStoragePath();
```
koesie10: It always seems fishy to me when we're calling instance methods in the constructor. The method doesn't depend on any instance variables, so can we move this out of the class and into a function?

extensions/ql-vscode/src/skeleton-query.ts
```diff
@@ -80,6 +82,18 @@ export class SkeletonQueryWizard {
     });
   }
 
+  private workoutStoragePath() {
```
koesie10: `workoutStoragePath` seems quite general. Would `getFirstStoragePath` work better as a name that describes what this is doing?

extensions/ql-vscode/src/skeleton-query.ts
```diff
@@ -91,7 +91,15 @@ export class SkeletonQueryWizard {
 
     const firstFolder = workspaceFolders[0];
 
-    return firstFolder.uri.path;
+    // For the vscode-codeql-starter repo, the first folder will be a ql pack
+    // so we need to get the parent folder
+    if (firstFolder.uri.path.includes("codeql-custom-queries")) {
```
koesie10: Can we make this work even better by doing this only after the language has been selected? It seems to me that the distinction between `folderName` and `storagePath` can be removed completely and we don't need this logic at all.

It should be possible to just have a single instance variable specifying the directory where the qlpack will be generated. For the codespace, this would be `<workspace>/codeql-custom-queries-${language}`. For the starter workspace, it would be the workspace directory path which matches the `codeql-custom-queries-${language}` name.

extensions/ql-vscode/test/vscode-tests/cli-integration/skeleton-query-wizard.test.ts
```diff
@@ -12,62 +12,55 @@ import { CancellationTokenSource } from "vscode-jsonrpc";
...
+  const credentials = testCredentialsWithStub();
+  const chosenLanguage = "ruby";
+  const mockDatabaseManager = mockedObject<DatabaseManager>({
+    setCurrentDatabaseItem: jest.fn(),
+    digForDatabaseItem: jest.fn(),
+  });
+  const mockCli = mockedObject<CodeQLCliServer>({
+    resolveLanguages: jest
+      .fn()
+      .mockResolvedValue([
```
koesie10, line 44: This mock will be reset when a new test is started, so I don't think you can move this out of the `beforeEach`. These calls are also not expensive, so I don't see why this would help with the timeout.

extensions/ql-vscode/src/common/commands.ts
```diff
@@ -96,6 +96,7 @@ export type LocalQueryCommands = {
   "codeQL.quickEvalContextEditor": (uri: Uri) => Promise<void>;
   "codeQL.codeLensQuickEval": (uri: Uri, range: Range) => Promise<void>;
   "codeQL.quickQuery": () => Promise<void>;
+  "codeQL.createSkeletonQuery": () => Promise<void>;
```
koesie10, line 99: Would it make sense for this to be a separate "module"? This doesn't seem like it's really related to local querying.

extensions/ql-vscode/package.json
```diff
@@ -361,6 +361,10 @@
         "command": "codeQL.quickQuery",
         "title": "CodeQL: Quick Query"
       },
+      {
+        "command": "codeQL.createSkeletonQuery",
+        "title": "CodeQL: Create Query"
```
koesie10, line 366: Should this command immediately be available to all users or should this be feature flagged?

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -0,0 +1,243 @@
...
+      this.fileName = await this.workoutNextFileName(this.folderName);
+      await qlPackGenerator.createExampleQlFile(this.fileName);
+    } catch (e: unknown) {
+      void this.extLogger.log(
+        `Could not create skeleton QL pack: ${getErrorMessage(e)}`,
+      );
+    }
+  }
+
+  private async workoutNextFileName(folderName: string): Promise<string> {
```
koesie10, line 174: I think this naming is quite confusing. I think something like `determineNextFileName` would be better.

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -0,0 +1,243 @@
...
+      throw new Error("Workspace storage path is undefined");
+    }
+
+    const folderUri = Uri.file(join(this.storagePath, folderName));
+    const files = await workspace.fs.readDirectory(folderUri);
+    const qlFiles = files.filter((file) =>
+      file[0].endsWith(".ql") ? true : false,
+    );
+
+    return `example${qlFiles.length + 1}.ql`;
```
koesie10, line 185: Should this be based on all the `.ql` files in the directory? It seems like this should only be based on the number of `example{n}.ql` files in the directory.

I also think the `+ 1` is unnecessary. If there's already an `example.ql` file in the directory, this would create `example2.ql` rather than `example1.ql`.

koesie10 (commented):
Thanks for splitting this into multiple commits, that made it easier to review. However, it seems like some of the comments are now in different places, so I hope they are all visible.

I haven't actually run this, so this review is just based on the code.

extensions/ql-vscode/src/local-databases.ts
```diff
@@ -897,6 +897,20 @@ export class DatabaseManager extends DisposableObject {
...
+  ): Promise<DatabaseItem | undefined> {
+    const dbItems = this.databaseItems || [];
+    const dbs = dbItems.filter(
+      (db) => db.language === language && db.name === databaseNwo,
+    );
+    if (dbs.length === 0) {
+      return undefined;
+    }
+    return dbs[0];
+  }
```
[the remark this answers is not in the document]
elenatanasoiu: I initially started out with defining the method in the wizard, but the `databaseItems` property on the `databaseManager` is private & readonly so when trying to pass it in, I'd get this error:

```
Argument of type 'readonly DatabaseItem[]' is not assignable to parameter of type 'DatabaseItem[]'.
  The type 'readonly DatabaseItem[]' is 'readonly' and cannot be assigned to the mutable type 'DatabaseItem[]'.
```

I agree that class is massive, but on the other hand I'm also calling `setCurrentDatabaseItem` on it as well, so I think it makes sense for it to deal with databaseItem logic. Also there are two other search methods in that class: `findDatabaseItem` (searches by Uri) and `findDatabaseItemBySourceArchive`.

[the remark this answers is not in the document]
elenatanasoiu: We could remove the `readonly` on the attribute, but I think it would be cooler if we moved that large class into its own file to make it more scannable. I'm happy to look at that in a separate PR.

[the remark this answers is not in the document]
koesie10: You should be able to change the type from `databaseItems: DatabaseItem[]` to `databaseItems: readonly DatabaseItem[]` since you aren't modifying the array or items. If that doesn't work, there's also `databaseItems: (readonly DatabaseItem)[]` which just sets the `readonly` on the items instead of on the array.

I'm not sure if I mentioned this in my initial review, but I think the naming of this method could also be clearer, something like `findDatabaseItemByNwo`.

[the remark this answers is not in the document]
elenatanasoiu: Yeah the naming is definitely a thing I struggled with. Cool, I can give it a shot (I didn't know you could do that! I can't think of a case where this would be a bad idea.)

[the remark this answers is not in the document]
elenatanasoiu: I'm going to settle for moving the search methods in the skeleton wizard file, because I'm not sure where helper methods like these would go if we had to define a separate file. Open to suggestions, if you have any ideas.

extensions/ql-vscode/src/databaseFetcher.ts
```diff
@@ -103,6 +107,48 @@ export async function promptImportGithubDatabase(
...
+/**
+ * Downloads a database from GitHub
+ *
+ * @param githubRepo the GitHub repository to download the database from
+ * @param databaseManager the DatabaseManager
+ * @param storagePath where to store the unzipped database.
+ * @param credentials the credentials to use to authenticate with GitHub
+ * @param progress the progress callback
+ * @param token the cancellation token
+ * @param cli the CodeQL CLI server
```
shati-patel, line 140: Minor: Should this include the `language` param as well?

shati-patel (commented):
In contrast to `@koesie10`, I haven't looked too deeply at the code yet, but found some issues whilst trying this locally. Added them here, but let me know if I can help with debugging 🪟

extensions/ql-vscode/src/local-databases.ts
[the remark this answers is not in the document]
elenatanasoiu: I've done the refactoring in a separate PR: https://github.com/github/vscode-codeql/pull/2292

extensions/ql-vscode/src/skeleton-query.ts
```diff
@@ -0,0 +1,218 @@
...
+import * as databaseFetcher from "./databaseFetcher";
+import { ProgressCallback } from "./progress";
+
+type QueryLanguagesToDatabaseMap = {
+  [id: string]: string;
+};
+
+export class SkeletonQueryWizard {
+  private language: string | undefined;
+  private folderName: string | undefined;
```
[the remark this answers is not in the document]
elenatanasoiu: I'm happy to convert this, but I'm not sure what the advantage would be in this case? 

As far as I understand, `getter` methods are a way to access private members of a class. `folderName` is already private and I don't really want to make this accessible from outside.

Also adding this code doesn't seem like it would improve anything:

```
  private _folderName: string | undefined;
  ...
  public get folderName() {
    return this._folderName;
  }
```

Happy to be told otherwise though, just not sure this is the right use case (as far as my understanding goes of this).

extensions/ql-vscode/src/skeleton-query.ts
```diff
@@ -0,0 +1,218 @@
...
+    private readonly credentials: Credentials | undefined,
+    private readonly extLogger: OutputChannelLogger,
+    private readonly databaseManager: DatabaseManager,
+    private readonly token: CancellationToken,
+  ) {}
+
+  public async execute() {
+    // show quick pick to choose language
+    await this.chooseLanguage();
+    if (!this.language) {
```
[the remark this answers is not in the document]
elenatanasoiu: Yes! I don't know why I didn't do that in the first place. This is much clearer.

extensions/ql-vscode/src/skeleton-query.ts
```diff
@@ -0,0 +1,218 @@
...
+      );
+    }
+  }
+
+  private async workoutNextFileName(folderName: string): Promise<string> {
+    if (this.storagePath === undefined) {
+      throw new Error("Workspace storage path is undefined");
+    }
+
+    const folderUri = Uri.file(join(this.storagePath, folderName));
```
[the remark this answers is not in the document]
elenatanasoiu: Same issue as before, `joinPath` would expect a `Uri` as the first param.

extensions/ql-vscode/src/skeleton-query.ts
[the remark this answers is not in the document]
koesie10: That is one way of using getters, but getters can also return dynamic values. I was thinking more along the lines of:

```ts
private get folderName() {
  return `codeql-custom-queries-${this.language}`;
}
```

This removes some instance state from the class, which would make this class easier to understand in my opinion.

extensions/ql-vscode/src/skeleton-query.ts
```diff
@@ -0,0 +1,218 @@
...
+import { askForLanguage, isFolderAlreadyInWorkspace } from "./helpers";
+import { getErrorMessage } from "./pure/helpers-pure";
+import { QlPackGenerator } from "./qlpack-generator";
+import { DatabaseItem, DatabaseManager } from "./local-databases";
+import * as databaseFetcher from "./databaseFetcher";
+import { ProgressCallback } from "./progress";
+
+type QueryLanguagesToDatabaseMap = {
+  [id: string]: string;
+};
```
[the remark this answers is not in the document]
elenatanasoiu: Ah yes! That's what I was looking for. Thanks!

I did start out with trying to make it use `QueryLanguage` but the dropdown we show the user (via `askForLanguage`) is going to return a `string`. So we'd need to some more refactoring in all the places we prompt the user for a language. I'm gonna write this up as a ticket for later.

extensions/ql-vscode/src/skeleton-query.ts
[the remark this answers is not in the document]
elenatanasoiu: Ah that makes more sense! Cool, I'll add it. Thanks!

extensions/ql-vscode/src/skeleton-query.ts
```diff
@@ -0,0 +1,218 @@
...
+      // download database based on language and select it
+      await this.downloadDatabase();
+    }
+
+    // open a query file
+    await this.openExampleFile();
+  }
+
+  private async openExampleFile() {
+    if (this.folderName === undefined || this.storagePath === undefined) {
```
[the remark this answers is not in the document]
elenatanasoiu: I'm not really sure it would make it more readable. We'd still need this check to see if they're undefined, when we call the `openExampleFile` method. But now we also have to pass in params to the method. 

We'd be inconsistent if we didn't also do this for `this.fileName` and also for the previous methods that use these values, so I think it might make it harder to read because we'd be getting some values from state and some from params. 

I might lack imagination, I admit. I think it would result in longer method calls.

extensions/ql-vscode/src/skeleton-query.ts
elenatanasoiu: I've moved the call to this method after the language selection and added some tests for it.

extensions/ql-vscode/test/vscode-tests/cli-integration/skeleton-query-wizard.test.ts
elenatanasoiu: My understanding was that because we were creating the object once, it would shorten the overall duration of the tests in the file. Since I went with increasing the timeout, I've moved the mock back in the forEach.

extensions/ql-vscode/package.json
elenatanasoiu: I've discussed this with our PM and we want this to be available.

extensions/ql-vscode/src/common/commands.ts
elenatanasoiu: I should've mentioned this in the commit message. My bad. I've discussed this with our PM and we want to make this the default way you create local queries, which is why we kept the name "CodeQL: Create Query". 

I'm happy to move it somewhere else, or alternatively we could just call it `codeQL.createQuery` or `codeQL.createLocalQuery`.

extensions/ql-vscode/test/vscode-tests/cli-integration/skeleton-query-wizard.test.ts
elenatanasoiu: I ended up doing this in the new refactoring PR, to avoid having to rebase that PR later: https://github.com/github/vscode-codeql/pull/2292/commits/d720bae20110630f0a51287041f8f58c4672be1e

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -0,0 +1,243 @@
...
+    if (skeletonPackAlreadyExists) {
+      // just create a new example query file in skeleton QL pack
+      await this.createExampleFile();
+      // select existing database for language
+      await this.selectExistingDatabase();
+    } else {
+      // generate a new skeleton QL pack with query file
+      await this.createQlPack();
+      // download database based on language and select it
+      await this.downloadDatabase();
```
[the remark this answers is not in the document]
elenatanasoiu: Thanks for catching this Shati! I think I'll have to dig into it a bit more. I'm gonna write up a ticket and try to fix this a in a follow-up PR, as this one is getting quite heavy on reviews + replies.

elenatanasoiu:
Thanks for the reviews @koesie10 & @shati-patel! I'm going to pair with @shati-patel tomorrow to look at the windows issues she's discovered. 

In the meantime, this PR is ready for review again.

extensions/ql-vscode/src/skeleton-query-wizard.ts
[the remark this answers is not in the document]
elenatanasoiu: Have added a fix here: https://github.com/github/vscode-codeql/pull/2298

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -0,0 +1,247 @@
...
+      // if the first folder is not a ql pack, then we are in a normal workspace
+      return firstFolder.uri.path;
+    }
+  }
+
+  private async chooseLanguage() {
+    this.progress({
+      message: "Choose language",
+      step: 1,
+      maxStep: 1,
```
koesie10, line 118: Why does this have a `maxStep` of 1? I would expect the `maxStep` to be consistent across all `this.progress` calls so that the user sees their progress increasing.

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  private async determineNextFileName(folderName: string): Promise<string> {
+    if (this.storagePath === undefined) {
+      throw new Error("Workspace storage path is undefined");
+    }
+
+    const folderUri = Uri.file(join(this.storagePath, folderName));
+    const files = await workspace.fs.readDirectory(folderUri);
+    const qlFiles = files.filter(([filename, _fileType]) =>
+      filename.match(/example[0-9]*.ql/),
```
koesie10, line 188:
```suggestion
      filename.match(/^example[0-9]*\.ql$/),
```

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -0,0 +1,247 @@
...
+    try {
+      void workspace.openTextDocument(queryFileUri).then((doc) => {
+        void Window.showTextDocument(doc);
+      });
+    } catch (e: unknown) {
+      void this.extLogger.log(
+        `Could not open example query file: ${getErrorMessage(e)}`,
+      );
+    }
+  }
```
koesie10, line 92: This doesn't actually `catch` the error since we're not awaiting the result. We need to use the `.catch` method on the promise to catch this error, or `await` the result. However, I don't think we want to `await` the result since then we wouldn't be using `void` and `.then`.

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -0,0 +1,247 @@
...
+import { CancellationToken, Uri, workspace, window as Window } from "vscode";
+import { CodeQLCliServer } from "./cli";
+import { OutputChannelLogger } from "./common";
+import { Credentials } from "./common/authentication";
+import { QueryLanguage } from "./common/query-language";
+import { askForLanguage, isFolderAlreadyInWorkspace } from "./helpers";
+import { getErrorMessage } from "./pure/helpers-pure";
+import { QlPackGenerator } from "./qlpack-generator";
+import { DatabaseManager } from "./local-databases";
+import * as databaseFetcher from "./databaseFetcher";
```
koesie10, line 11: Can we just import the methods we need here instead of importing the complete module? Right now, I'm not sure why it's different than the other imports.

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -0,0 +1,247 @@
...
+    if (skeletonPackAlreadyExists) {
+      // just create a new example query file in skeleton QL pack
+      await this.createExampleFile();
+      // select existing database for language
+      await this.selectExistingDatabase();
+    } else {
+      // generate a new skeleton QL pack with query file
+      await this.createQlPack();
+      // download database based on language and select it
+      await this.downloadDatabase();
```
shati-patel, line 67: It might be helpful to send some messages to the extension logger before/after each step, to help us (and users) with debugging.

shati-patel (approved):
Thanks! 🎉 I'm also keen to get this merged, since I'm struggling to review/keep track of the parallel PRs and changes 🙈 

I've left a bunch of comments/questions, but they are all non-blocking 💚 The main important thing for me is to (temporarily) hide the command, so that we can do more testing/debugging without disrupting users.

extensions/ql-vscode/src/skeleton-query-wizard.ts
elenatanasoiu: Thanks 👍 I've opened a new PR to address this: https://github.com/github/vscode-codeql/pull/2305

I admit, I just copied it without really looking into how the steps were marked.

extensions/ql-vscode/src/skeleton-query-wizard.ts
elenatanasoiu: Added here: https://github.com/github/vscode-codeql/pull/2305

extensions/ql-vscode/src/skeleton-query-wizard.ts
elenatanasoiu: I had to do this in order to [spy on methods for the databaseFetcher](https://github.com/github/vscode-codeql/pull/2250/files#diff-69022948abbb5354a1731c421934f96e1b96b53a2efc045bf40a3c7b78be1ea5R90):

```
downloadGitHubDatabaseSpy = jest
      .spyOn(databaseFetcher, "downloadGitHubDatabase")
      .mockResolvedValue(undefined);
```

I'm not sure if there's a better way.

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -0,0 +1,247 @@
...
+
+export const QUERY_LANGUAGE_TO_DATABASE_REPO: QueryLanguagesToDatabaseMap = {
+  cpp: "protocolbuffers/protobuf",
+  csharp: "dotnet/efcore",
+  go: "evanw/esbuild",
+  java: "google/guava",
+  javascript: "facebook/react",
+  python: "pallets/flask",
+  ruby: "rails/rails",
+  swift: "Alamofire/Alamofire",
```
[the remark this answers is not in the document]
elenatanasoiu: Is there a way to see the size of the databases? I'm not really sure how to investigate this at the moment.

extensions/ql-vscode/src/local-databases.ts
```diff
@@ -897,6 +897,20 @@ export class DatabaseManager extends DisposableObject {
...
   }
 
+  public async digForDatabaseItem(
+    language: string,
+    databaseNwo: string,
+  ): Promise<DatabaseItem | undefined> {
+    const dbItems = this.databaseItems || [];
+    const dbs = dbItems.filter(
+      (db) => db.language === language && db.name === databaseNwo,
+    );
```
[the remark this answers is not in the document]
elenatanasoiu: I think we settled on just the first database we find when this was requested. 

Your idea about filtering out databases with errors makes sense, I'm gonna look into this tomorrow.

extensions/ql-vscode/src/skeleton-query-wizard.ts
elenatanasoiu: Done here: https://github.com/github/vscode-codeql/pull/2305

extensions/ql-vscode/src/skeleton-query-wizard.ts
elenatanasoiu: Do we do this anywhere else? Just looking for examples.

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  private async selectExistingDatabase() {
+    if (this.language === undefined) {
+      throw new Error("Language is undefined");
+    }
+
+    if (this.storagePath === undefined) {
+      throw new Error("Workspace storage path is undefined");
+    }
```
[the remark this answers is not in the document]
elenatanasoiu: Added here: https://github.com/github/vscode-codeql/pull/2305

elenatanasoiu:
Thanks for the reviews @koesie10 & @shati-patel 🙇‍♀️ 

I'm gonna merge this and follow up with the remaining issues tomorrow.

extensions/ql-vscode/package.json
elenatanasoiu: Added here: https://github.com/github/vscode-codeql/pull/2305

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -0,0 +1,243 @@
...
+    );
+
+    if (skeletonPackAlreadyExists) {
+      // just create a new example query file in skeleton QL pack
+      await this.createExampleFile();
+      // select existing database for language
+      await this.selectExistingDatabase();
+    } else {
+      // generate a new skeleton QL pack with query file
+      await this.createQlPack();
```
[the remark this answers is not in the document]
elenatanasoiu: Thanks Shati. We want to treat the folder as a normal folder in the repo that you can commit and edit. 

So that means:
1. We can't save it in the extension storage, it needs to be in the workspace as a regular folder.
2. We should avoid confusing the user by also pushing the folder as a top level workspace folder (like you've discovered). 

So that means
- for the codespace repo: it needs to be created in the codespaces-codeql folder ✅ 
- for the starter repo: it needs to be a top level folder ✅ 
- we should open a follow-up PR to disable the "pushing as workspace folder" behaviour. 

The reason we were adding the folder to the workspace in the first place is because in the original flow we weren't interested in committing the folder so we went for storing it together with the database. That means the folder was invisible in the file explorer so we decided to make it visible by pushing it to the workspace as a top level folder. Hindsight is 20/20.

PR #2255, opened 2023-03-31
Add simple regression unit test for the results view

koesie10:
This adds a really simple regression unit test for the results view which checks that the results view can render a SARIF file. This is in preparation for the upgrade to React 18 to ensure that we don't break the basic functionality of the results view.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/results/__tests__/results.spec.tsx
+++ b/extensions/ql-vscode/src/view/results/__tests__/results.spec.tsx
@@ -1,0 +1,120 @@
+import * as React from "react";
+import { render as reactRender, screen } from "@testing-library/react";
+import { ResultsApp } from "../results";
+import {
+  Interpretation,
+  IntoResultsViewMsg,
+  SortDirection,
+} from "../../../pure/interface-types";
+import * as fs from "fs-extra";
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/view/results/__tests__/results.spec.tsx
```diff
@@ -0,0 +1,120 @@
...
+    window.dispatchEvent(
+      new MessageEvent("message", {
+        source: window,
+        origin: window.location.origin,
+        data: msg,
+      }),
+    );
+
+    // The event is dispatched asynchronously, so we need to wait for it to be handled.
+    await new Promise((resolve) => setTimeout(resolve, 0));
```
aeisenberg, line 34: This seems brittle to me. Are we guaranteed that the dispatch will complete before the next tick?

If this is indeed running in a DOM context, does it make sense to add an event handler that resolves the promise? If it seems to work reliably right now, then I'm fine to keep as it is, but this may be something we need to revisit.

aeisenberg (approved):
Thank you for implementing this. We may need to revisit how we await the event dispatch, but let's look into that only if we need to.

PR #2258, opened 2023-04-03
Upgrade Storybook for TS 5.0 compatibility

koesie10:
This upgrades Storybook to an alpha version to ensure we can start Storybook when using TypeScript 5.0. This is temporary until we can upgrade to Storybook 7 or to a released version of 6.5.17.

## Checklist


charisk (approved):
Why are we not updating to 7? Is it a lot of work/are there breaking changes for us?

koesie10:
> Why are we not updating to 7? Is it a lot of work/are there breaking changes for us?

I've done the work already (see https://github.com/github/vscode-codeql/pull/2259), but it seems to crash on start-up without any error message. We should upgrade to 7, but since it's only been released yesterday, they might need to iron out some bugs so it actually works as we would expect to.

PR #2265, opened 2023-04-04
Load existing data extension YAML in editor

koesie10:
This loads in the existing data extension YAML file for the selected database. It only supports the filename we save it to, and will not load it from any other data extension YAML files.

Based on #2264

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #2267, opened 2023-04-04
Add generating of flow model to data extension editor

koesie10:
This adds the automatic generation of sources/sinks/summary flows to the data extension editor using the flow model queries. This is based on the Python script available in the CodeQL repo.

See: https://github.com/github/codeql/blob/main/java/ql/src/utils/modelgenerator/GenerateFlowModel.py

Based on #2265

## Checklist


extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    const resultSet = bqrsInfo["result-sets"][0];
+
+    const decodedResults = await this.cli.bqrsDecode(bqrsPath, resultSet.name);
+
+    const results = decodedResults.tuples;
+
+    return results
+      .map((result): [string, ModeledMethod] | undefined => {
+        const row = result[0] as string;
```
charisk, line 77: Shall we check the type is a string before doing `as string`?

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    const resultSet = bqrsInfo["result-sets"][0];
+
+    const decodedResults = await this.cli.bqrsDecode(bqrsPath, resultSet.name);
+
+    const results = decodedResults.tuples;
+
+    return results
+      .map((result): [string, ModeledMethod] | undefined => {
```
charisk, line 76: The `[string, ModeledMethod]` is being used again and again. Is it time to name it and re-use it?

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
```diff
@@ -0,0 +1,146 @@
...
+      return undefined;
+    }
+
+    const resultSet = bqrsInfo["result-sets"][0];
+
+    const decodedResults = await this.cli.bqrsDecode(bqrsPath, resultSet.name);
+
+    const results = decodedResults.tuples;
+
+    return results
```
charisk, line 75: Is it worth extracting this into its own function with a good name?

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
```diff
@@ -0,0 +1,146 @@
...
+      .map((result): [string, ModeledMethod] | undefined => {
+        const row = result[0] as string;
+
+        return definition.readModeledMethod(row.split(";"));
+      })
+      .filter(
```
charisk, line 81: Shall we filter undefined before we map?

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
```diff
@@ -0,0 +1,146 @@
...
+    private readonly cli: CodeQLCliServer,
+    private readonly queryRunner: QueryRunner,
+    private readonly queryStorageDir: string,
+    private readonly qlDir: string,
+    private readonly databaseItem: DatabaseItem,
+    private readonly progress: ProgressCallback,
+    private readonly token: CancellationToken,
+  ) {}
+
+  async getAddsTo(
```
charisk, line 23: I don't quite understand the name of this function.

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
```diff
@@ -0,0 +1,146 @@
...
+import { join } from "path";
+import { QueryRunner } from "../queryRunner";
+import { CodeQLCliServer } from "../cli";
+import { extLogger, TeeLogger } from "../common";
+import { definitions } from "./yaml";
+import { ProgressCallback } from "../progress";
+import { getOnDiskWorkspaceFolders } from "../helpers";
+import { ModeledMethod, ModeledMethodType } from "./modeled-method";
+
+class FlowModelGenerator {
```
charisk, line 12: What's the reasoning creating a class for this? I don't see it's purpose considering we provide a function that takes all the dependencies/args, creates an instance of the function and executes 1 function on it. Apologies if I'm missing something obvious.

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
```diff
@@ -0,0 +1,146 @@
...
+        }),
+      this.token,
+      new TeeLogger(this.queryRunner.logger, queryRun.outputDir.logPath),
+    );
+
+    const bqrsPath = queryResult.outputDir.bqrsPath;
+
+    const bqrsInfo = await this.cli.bqrsInfo(bqrsPath);
+    if (bqrsInfo["result-sets"].length !== 1) {
+      void extLogger.log(
```
charisk, line 63: Should this be a warning for users?

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
```diff
@@ -0,0 +1,146 @@
...
+    private readonly databaseItem: DatabaseItem,
+    private readonly progress: ProgressCallback,
+    private readonly token: CancellationToken,
+  ) {}
+
+  async getAddsTo(
+    type: Exclude<ModeledMethodType, "none">,
+    queryName: string,
+    queryStep: number,
+  ): Promise<Array<[string, ModeledMethod]> | undefined> {
```
charisk, line 27: Can we just return an empty array (or blow up) instead of `undefined`?

charisk (commented):
I have quite a few comments but I want to mention that I appreciate that a lot of this work is at a prototype phase so I'm happy to discuss how far we want to take things.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -213,6 +222,85 @@ export class DataExtensionsEditorView extends AbstractWebview<
     }
   }
 
+  protected async generateExternalApi(): Promise<void> {
+    const tokenSource = new CancellationTokenSource();
+
+    const selectedDatabase = this.databaseManager.currentDatabaseItem;
+
+    const database = await promptImportGithubDatabase(
```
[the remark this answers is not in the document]
koesie10: The database that we're generating against is the database of the external library source code, rather than the program source code. So, in our example `dsp-testing/sql2o-example`, we are temporarily adding `dsp-testing/sql2o-import` which contains the source code of the sql2o library. Since the user is not trying to analyze or explore the library source code, we're only temporarily adding it and removing it.

The reason for adding it at all is to ensure that it gets registered in the query server.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -213,6 +222,85 @@ export class DataExtensionsEditorView extends AbstractWebview<
...
+    if (!database) {
+      await this.clearProgress();
+      void extLogger.log("No database chosen");
+
+      return;
+    }
+
+    await this.databaseManager.setCurrentDatabaseItem(selectedDatabase);
+
+    const workspaceFolder = workspace.workspaceFolders?.find(
```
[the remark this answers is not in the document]
koesie10: 👍, this is only temporary and will be removed in one of the follow-up issues, so I've just extracted it into a function in the same file.

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
koesie10: I don't think any `readModeledMethod` is returning `undefined` right now, so I've just removed `undefined` from the return type to simplify this code.

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
koesie10: The name was based on the YAML file which is generated, which has an `addsTo` section. It doesn't make sense anymore since we're now just generating modeled methods, so I'll rename it.

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
koesie10: The reasoning is just to ensure we don't need to pass around all of these dependencies, especially since the same function taking the dependencies is called four times. Putting it in a class encapsulates this nicely.

An alternative which avoids passing these dependencies four times would be creating a nested function that calls the actual function with the correct parameters, but that also doesn't help with readability.

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
koesie10: I think it's better to leave this until we can properly split out this function and unify the query running, which would happen after both this PR and #2271 are merged.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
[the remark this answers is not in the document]
charisk: Oh I see, thanks! Do you mind adding some code comments to cover this?

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
charisk: Sorry I hadn't spotted that we were calling it four times! I think we've generally used this pattern - creating a class that gets the dependencies in the constructor and then exposing a function to call, so I'm happy leaving it like that.

charisk (approved):
Thanks for addressing all the feedback! This is generally looking good to me. Just a couple of remaining but non blocking comments.

PR #2270, opened 2023-04-05
Repo States: new data model and mapping

norascheuch:
This PR introduces new data types for storing repo states and uses them to map between data and domain model. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/store/repo-states-data-types.ts
+++ b/extensions/ql-vscode/src/variant-analysis/store/repo-states-data-types.ts
@@ -1,7 +1,12 @@
-import { VariantAnalysisScannedRepositoryDownloadStatus } from "../shared/variant-analysis";
-
 export interface VariantAnalysisScannedRepositoryStateData {
   repositoryId: number;
-  downloadStatus: VariantAnalysisScannedRepositoryDownloadStatus;
+  downloadStatus: VariantAnalysisScannedRepositoryDownloadData;
   downloadPercentage?: number;
+}
+
+export enum VariantAnalysisScannedRepositoryDownloadData {
+  Pending = "pending",
+  InProgress = "inProgress",
+  Succeeded = "succeeded",
+  Failed = "failed",
 }
--- a/extensions/ql-vscode/src/variant-analysis/store/repo-states-data-types.ts
+++ b/extensions/ql-vscode/src/variant-analysis/store/repo-states-data-types.ts
@@ -1,0 +1,7 @@
+import { VariantAnalysisScannedRepositoryDownloadStatus } from "../shared/variant-analysis";
+
+export interface VariantAnalysisScannedRepositoryStateData {
+  repositoryId: number;
+  downloadStatus: VariantAnalysisScannedRepositoryDownloadStatus;
+  downloadPercentage?: number;
+}
--- a/extensions/ql-vscode/src/variant-analysis/store/repo-states-store.ts
+++ b/extensions/ql-vscode/src/variant-analysis/store/repo-states-store.ts
@@ -1,6 +1,8 @@
 import { outputJson, readJson } from "fs-extra";
 import { VariantAnalysisScannedRepositoryState } from "../shared/variant-analysis";
 import { VariantAnalysisScannedRepositoryStateData } from "./repo-states-data-types";
+import { mapRepoStateToData } from "./repo-states-to-data-mapper";
+import { mapRepoStateToDomain } from "./repo-states-to-domain-mapper";
 
 export const REPO_STATES_FILENAME = "repo_states.json";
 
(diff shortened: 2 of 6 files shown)


norascheuch:
I have two questions for the reviewer:
- should we add to the existing logic and catch an error when reading or writing from file?
- should we create a mapper for the enum as we do [here](https://github.com/github/vscode-codeql/blob/9015a57732d332e59ccdf961fdaa380c1348b931/extensions/ql-vscode/src/variant-analysis/variant-analysis-processor.ts#L191) ?

extensions/ql-vscode/src/variant-analysis/store/repo-states-store.ts
```diff
@@ -1,17 +1,39 @@
...
+  );
+
+  return await outputJson(storagePath, repoStatesData);
 }
 
 export async function readRepoStates(
   storagePath: string,
 ): Promise<Record<number, VariantAnalysisScannedRepositoryState>> {
-  return await readJson(storagePath);
+  const repoStatesData: Record<
```
robertbrignull, line 25: I'm wondering if we need some more error handling here. Now we're no longer just returning the raw data, but doing some processing here.

I think we can assume that if it does exist then it contains valid data, since we're leaving that concern for when we add JSON schema validation later on. But right now what do we do in case that the JSON file doesn't exist?

I've tried to read the documentation for this method and it's not clear to me how it behaves in cases of errors. My best guess by looking at the error handling the existing code has is that it returns `undefined`. It would be good to find some docs, or to test that manually and see.

robertbrignull:
> should we create a mapper for the enum

I would say yes, unless you want to do it in 2 PRs. But I think we do want a mapper for this type as well, particularly because of the enum. It should be a lot simpler to add than the query history types, so doing it in this PR shouldn't be infeasible.

extensions/ql-vscode/src/variant-analysis/store/repo-states-data-types.ts
```diff
@@ -0,0 +1,12 @@
+export interface VariantAnalysisScannedRepositoryStateData {
+  repositoryId: number;
+  downloadStatus: VariantAnalysisScannedRepositoryDownloadData;
+  downloadPercentage?: number;
+}
+
+export enum VariantAnalysisScannedRepositoryDownloadData {
```
robertbrignull, line 7: Nice! This looks good. It's good these "data types" files don't import any other files, and that shows we don't accidentally depend on a domain model.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -264,16 +264,16 @@ export class VariantAnalysisManager
...
-        this.repoStates.set(variantAnalysis.id, {});
+      const repoStatesFromDisk = await readRepoStates(
+        this.getRepoStatesStoragePath(variantAnalysis.id),
+      );
+
+      if (repoStatesFromDisk) {
+        this.repoStates.set(variantAnalysis.id, repoStatesFromDisk);
       }
 
+      this.repoStates.set(variantAnalysis.id, {});
```
robertbrignull, line 275: I like moving the catch to be inside `readRepoStates`. Unfortunately rewriting this changed the behaviour so now this will currently always overwrite the repo states with `{}`, which isn't what we want. 

We could put the case for when `repoStatesFromDisk` is undefined into an `else` block. Or we could use `|| {}` to supply this default value.

```suggestion
      this.repoStates.set(variantAnalysis.id, repoStatesFromDisk || {})
```

extensions/ql-vscode/src/variant-analysis/store/repo-states-to-data-mapper.ts
```diff
@@ -0,0 +1,35 @@
...
+    case VariantAnalysisScannedRepositoryDownloadStatus.Pending:
+      return VariantAnalysisScannedRepositoryDownloadData.Pending;
+    case VariantAnalysisScannedRepositoryDownloadStatus.InProgress:
+      return VariantAnalysisScannedRepositoryDownloadData.InProgress;
+    case VariantAnalysisScannedRepositoryDownloadStatus.Succeeded:
+      return VariantAnalysisScannedRepositoryDownloadData.Succeeded;
+    case VariantAnalysisScannedRepositoryDownloadStatus.Failed:
+      return VariantAnalysisScannedRepositoryDownloadData.Failed;
+    default:
+      return VariantAnalysisScannedRepositoryDownloadData.Pending;
```
robertbrignull, line 33:
```suggestion
      assertNever(downloadedStatus);
```

This is another good case for using `assertNever`.  When mapping from domain to data models, we don't have to worry about badly typed data and having to handle every possibility, so we can say this case will never happen and the type system will enforce it.

robertbrignull (commented):
A couple of suggestions / corrections, but this is looking very good generally and almost there. We can likely merge this afternoon.

robertbrignull (approved):

PR #2263, opened 2023-04-04
Show external API calls in data extensions editor

koesie10:
This updates the view of the data extensions editor to show a table of possible sources/sinks/flow summaries that can be edited. It's not yet possible to save the changes or load the existing file.

This is quite a big PR because it runs the query and shows the results. However, I don't think splitting this up further would really make this clearer. However, the extension and view parts can be reviewed relatively independently since the only communication between them is through two messages.

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #2271, opened 2023-04-05
Split and add tests for external API usages query

koesie10:
This extracts the methods `runQuery` and `getResults` out of the data extensions editor view and into a separate file, as suggested in [this comment](https://github.com/github/vscode-codeql/pull/2263/files/73bd6d696c214e37c7c0e5e5c605223bdb0db0e2#r1157363628).

It also adds tests for these functions. It does not use the "real" CLI integration, but instead mocks the CLI server and query runner.

To make mocking easier and require less type casting, I've narrowed some of the arguments of some other functions. They now use `Pick` to only require the properties they need.

## Checklist


extensions/ql-vscode/src/data-extensions-editor/external-api-usage-query.ts
```diff
@@ -0,0 +1,109 @@
...
+  );
+}
+
+export type GetResultsOptions = {
+  cliServer: Pick<CodeQLCliServer, "bqrsInfo" | "bqrsDecode">;
+  bqrsPath: string;
+  logger: Logger;
+};
+
+export async function getResults({
```
charisk, line 93: (Optional) I would call this `readQueryResults` or something like that.

extensions/ql-vscode/src/data-extensions-editor/external-api-usage-query.ts
```diff
@@ -0,0 +1,109 @@
...
+import { dump as dumpYaml } from "js-yaml";
+import { getOnDiskWorkspaceFolders } from "../helpers";
+import { Logger, TeeLogger } from "../common";
+import { CancellationToken } from "vscode";
+import { CodeQLCliServer } from "../cli";
+import { DatabaseItem } from "../local-databases";
+import { ProgressCallback } from "../progress";
+
+export type RunQueryOptions = {
+  cliServer: Pick<CodeQLCliServer, "resolveQlpacks" | "resolveQueriesInSuite">;
```
charisk, line 14: Does it make sense to define an interface for this and the other two fields where we use `Pick`? 

We might want to take this as tech debt instead of changing anything now (considering that we're at prototyping phase).

In general I think `Pick` can be really useful but it can hurt readability. To me it usually indicates fat types that could potentially be broken down.

charisk (approved):
Thanks for splitting that out and writing tests.

extensions/ql-vscode/src/data-extensions-editor/external-api-usage-query.ts
koesie10: I think this could indeed be seen as tech debt. The CLI server type is really large, and I chose to use `Pick` here so that we don't need to mock the entire CLI server object. Even if we do create an interface out of this, we'd still be relying on types from the CLI server (like `QlpacksInfo`), so I'm not sure creating a separate interface for this would even be beneficial without larger changes to those types in general.

PR #2279, opened 2023-04-06
Fix empty result view when switching between queries

koesie10:
When updating to React 18, we removed the loading step from the updating of the state of the result view since React would batch the updates anyway. However, this caused a bug where the result view would be empty when switching between queries. This is because the result view would retain the old selected result set name. This would not happen previously because React would re-render the view at least once, which would cause the result view to be unmounted and re-created.

This fixes it by resetting the selected result set if we can't find the result set in the new result sets. It also adds a test to ensure we don't have a regression in this behavior in the future.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/results/result-tables.tsx
+++ b/extensions/ql-vscode/src/view/results/result-tables.tsx
@@ -101,7 +101,7 @@ function getResultSets(
 ): ResultSet[] {
   const resultSets: ResultSet[] =
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore 2783
+    // @ts-ignore 2783 Avoid compilation error for overwriting the t property
     rawResultSets.map((rs) => ({ t: "RawResultSet", ...rs }));
 
   if (interpretation !== undefined) {
--- a/extensions/ql-vscode/src/view/results/result-tables.tsx
+++ b/extensions/ql-vscode/src/view/results/result-tables.tsx
@@ -79,7 +79,7 @@ function renderResultCountString(resultSet: ResultSet): JSX.Element {
 }
 
 function getInterpretedTableName(interpretation: Interpretation): string {
-  return interpretation?.data.t === "GraphInterpretationData"
+  return interpretation.data.t === "GraphInterpretationData"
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/view/results/result-tables.tsx
```diff
@@ -78,6 +78,52 @@ function renderResultCountString(resultSet: ResultSet): JSX.Element {
   );
 }
 
+function getInterpretedTableName(interpretation: Interpretation): string {
+  return interpretation?.data.t === "GraphInterpretationData"
```
aeisenberg, line 82: Minor: Is `?` necessary? According to the type information, `interpretation` can never be undefined.

```suggestion
  return interpretation.data.t === "GraphInterpretationData"
```

extensions/ql-vscode/src/view/results/result-tables.tsx
```diff
@@ -78,6 +78,52 @@ function renderResultCountString(resultSet: ResultSet): JSX.Element {
...
+    : parsedResultSets.resultSetNames;
+}
+
+function getResultSets(
+  rawResultSets: readonly ResultSet[],
+  interpretation: Interpretation | undefined,
+): ResultSet[] {
+  const resultSets: ResultSet[] =
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore 2783
```
aeisenberg, line 104: Can you add an explanation of what this comment is ignoring and why it is OK to ignore?

aeisenberg: Oh...looking at the git history, looks like *I* added this line 3 years ago. 😮‍💨

aeisenberg: How about this?

```suggestion
    // @ts-ignore 2783 Avoid compilation error for overwriting the t property
```

aeisenberg (commented):
Thanks for finding and fixing this! 

A handful of minor comments included.

dbartol:
Your description of the underlying bug explains several other strange behaviors that I observed but couldn't put into words.

aeisenberg (approved):

PR #2287, opened 2023-04-06
Introduce helpers for commands that operate on selections

robertbrignull:
This PR aims to improve our commands that work on selections of items from the UI. These commands all follow a pattern but currently there's nothing enforcing the arguments the commands take or how they process them. I've added new helper functions and converted all of the commands to use the helpers.

This should hopefully help in the following ways:
- Enforces the number and types of arguments accepted by the command, and it is not duplicated on each command.
- The command implementation accepts only the arguments it needs (either single item or array).
- Consistent error handling for when a command is called in a situation it isn't designed for.

I'm still open for suggestions of how to make this even better. We could definitely reduce duplication even more, or make things more streamlined or safer.

## Checklist


extensions/ql-vscode/test/vscode-tests/no-workspace/query-history/query-history-manager.test.ts
```diff
@@ -228,25 +220,6 @@ describe("QueryHistoryManager", () => {
...
-          secondItemClicked,
-        ]);
-
-        expect(localQueriesResultsViewStub.showResults).not.toHaveBeenCalled();
-        expect(variantAnalysisManagerStub.showView).not.toBeCalled();
-        expect(
-          queryHistoryManager.treeDataProvider.getCurrent(),
-        ).toBeUndefined();
-      });
-    });
```
elenatanasoiu, line 249: Has the behaviour changed or is there a different reason for removing this test?

I can see we've removed some of the tests that were checking for multiple items being passed, when the function should only receive a single item. I think for those cases, it makes sense for the tests to become redundant.

extensions/ql-vscode/src/common/selection-commands.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export function createMultiSelectionCommand<T extends NotArray>(
+  f: (argument: T[]) => Promise<void>,
+): SelectionCommand<T> {
+  return async (singleItem, multiSelect) => {
+    if (multiSelect !== undefined && multiSelect.length > 0) {
+      return f(multiSelect);
+    } else {
+      return f([singleItem]);
```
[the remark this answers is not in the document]
robertbrignull: As per the types and my experimentation, `singleItem` cannot be undefined. If you can find a way to trigger one of our commands and get `singleItem` to be undefined I'd be very interested to hear. Otherwise I'd prefer to keep our types and code simple.

extensions/ql-vscode/test/vscode-tests/no-workspace/query-history/query-history-manager.test.ts
robertbrignull: I'm deleting this test because it was covering the case of having multiple items selected. I don't quite understand why it's named "double click", but in practice it's calling `handleItemClicked` as if two items in the query history are selected, and then asserting it ignores it and does nothing. It's now no longer possible to pass two items to `handleItemClicked`, so we don't need this test anymore.

elenatanasoiu: > It's now no longer possible to pass two items to handleItemClicked, so we don't need this test anymore.

Ah, I see. Thanks for explaining!

extensions/ql-vscode/src/common/selection-commands.ts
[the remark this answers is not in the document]
elenatanasoiu: We seem to be checking whether a databaseItem is `undefined` in three places?

https://github.com/github/vscode-codeql/pull/2287/commits/40f309224457cc886a4e04e69e80cf25c1ef2e23#diff-b11b6f7e1d76e818f721c94ae740f046663c80d519c2f515f2f78344f90e6c78R518

https://github.com/github/vscode-codeql/pull/2287/commits/40f309224457cc886a4e04e69e80cf25c1ef2e23#diff-b11b6f7e1d76e818f721c94ae740f046663c80d519c2f515f2f78344f90e6c78R540

https://github.com/github/vscode-codeql/pull/2287/commits/40f309224457cc886a4e04e69e80cf25c1ef2e23#diff-b11b6f7e1d76e818f721c94ae740f046663c80d519c2f515f2f78344f90e6c78R564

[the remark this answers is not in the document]
robertbrignull: Thanks for pointing those out. The commands around upgrading databases are a little complicated because we kinda have two notions of being "selected". You can click to highlight entries in the list (i.e. the vscode tree view), and this state gets passed through to the `codeQLDatabases.upgradeDatabase` command if you select to upgrade the databases from the context menu. You can also click "make selected" on a database and this state is managed internally, and this is what is used for the `codeQL.upgradeCurrentDatabase` or `codeQL.runQuery` (a.k.a. "Run Query on Selected Database") commands.

Looking at each bit of code:
- https://github.com/github/vscode-codeql/commit/40f309224457cc886a4e04e69e80cf25c1ef2e23#diff-b11b6f7e1d76e818f721c94ae740f046663c80d519c2f515f2f78344f90e6c78R518
  - As mentioned above, the `codeQL.upgradeCurrentDatabase` command isn't really a "selection command" like the ones we're looking at in this PR. It's operating on the currently selected local database, but it's not a vscode tree view selection like the other cases in this PR. It's a selection that we manage ourselves, and thus you can trigger this command at any time to operate on the "current" database, and there may be a database selected or not.
- https://github.com/github/vscode-codeql/commit/40f309224457cc886a4e04e69e80cf25c1ef2e23#diff-b11b6f7e1d76e818f721c94ae740f046663c80d519c2f515f2f78344f90e6c78R540
  - This commit is from quite early in the PR and is doing some refactoring before we get into the rest of the PR. The code being linked to is already changed on the latest commit.
- https://github.com/github/vscode-codeql/commit/40f309224457cc886a4e04e69e80cf25c1ef2e23#diff-b11b6f7e1d76e818f721c94ae740f046663c80d519c2f515f2f78344f90e6c78R564
  - Thanks. This check is unnecessary. It's an oversight from me just wrapping the existing function implementation with `await Promise.all`. I've pushed a commit to remove this check.

extensions/ql-vscode/test/vscode-tests/no-workspace/query-history/query-history-manager.test.ts
elenatanasoiu: _I think_ the reason we added the test in the first place was because double clicking was causing problems and we fixed the issue. We wanted to make sure we account for the behaviour. It was kind of a superfluous test if we were already testing this in other places. Perhaps it was me not having a proper overview of the existing tests when this was introduced. Anyway, I agree we don't need this.

elenatanasoiu (approved):

PR #2291, opened 2023-04-10
Implement basic CodeQL debug adapter

dbartol:
This PR is the initial implementation of a [Debug Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/)-based debug engine for CodeQL. Implementing the DAP allows us to take advantage of all of VS Code's existing UI around debugging, including configuration via `launch.json`, the "Start Debugging" and "Start Without Debugging" commands, and the Debug Console. Later, we'll build further UI on top of this work, and perhaps add some advanced debugging features.

For now, our DAP implementation runs in the extension process, and invokes the query server via the same code that we were already using for "Run Query" and other local query commands. The existing commands do _not_ go through the debug adapter yet. Instead, I've added a few new commands under the canary flag, and I've enabled the regular VS Code "Start Debugging" (F5) and "Start Without Debugging" (Ctrl-F5) commands to work for CodeQL.

## Configuration
In "launch.json", there is a new available debug configuration type, named "codeql". Its configuration options are as follows:
- `query` - The path to the query to debug. This query will be treated as the root query of the compilation, _even when doing quick eval_. Thus, the quick eval selection can now be in a different file from the actual query, allowing debugging of queries that extend abstract class, instantiate parameterized modules, etc. If `query` is undefined, the currently active ".ql" file will be used, just like the "Run Query" command.
- `database` - The path to the target database directory. If `database` is undefined, the currently selected database from the database panel will be used, just like the "Run Query" command.
- `extensionPacks` - An array of the pack names (_not_ paths) of the extension packs to use. Each of these packs must be findable along the `additionalPacks` path. An empty array specifies that no extension packs are to be used. If `extensionPacks` is undefined, the default depends on the `codeQL.useExtensionPacks` setting: If true, then all extension packs in the current workspace will be used. If false, no extension packs will be used.
- `additionalPacks` - An array of paths to search for additional library packs. This is equivalent to the `--additional-packs` CLI option. If `additionalPacks` is undefined, the default is to use the list of root folders in the current workspace, as we already do for the other commands.
- `quickEval` - A boolean value to specify whether to perform a quickeval of the current selection instead of a regular query evaluation. Users will rarely set this explicitly in "launch.json". It exists primarily so that the new "CodeQL: Debug Selection" command can force it to true.

## Commands
The "CodeQL: Run Query" and "CodeQL: Quick Evaluation" commands work exactly as before. They ignore any debug configuration, and do not start a debug session. Later, once we're more confident in the debugger UX, we'll make all of those commands go through the debug configuration.

If a "codeql" debug configuration is selected on the Debug panel, hitting F5 ("Start Debugging") will launch a debug session with the active debug configuration. This will bring up the VS Code debug toolbar, and will evaluate the configured query against the configured database. Results are displayed in the results view as before, and the query history panel is updated. The most visible change is that all of the query server log output will now go to the Debug Console. The Debug Console is specific to that debug session, so you don't have to search through the query server output pane to make sure you're looking at the right query run.

The debug session will stay active after evaluation is complete. It can be terminated by clicking the stop (red square) button on the debug toolbar, or hitting Shift-F5.

Hitting Ctrl-F5 ("Start Without Debuggin") with a "codeql" debug configuration active will evaluate the query, display the results, and immediately terminate the debug session. This is most similar to the behavior of the existing "Run Query" command.

While a "codeql" debug session is active, the "CodeQL: Debug Selection" command will perform a quickeval of the current selection, _in the context of the active debug session_. If no debug session is active, but a "codeql" debug configuration is selected in the Debug panel, then this command launches a new debug session with that debug configuration to perform the quickeval.

While a "codeql" debug session is active, the "Continue" (F5) command will re-evaluate the entire query.

The "Step Into", "Steo Out", and "Step Over" commands don't do anything (yet).

## Code Changes
`package.json` has been updated to provide the "codeql" debug configuration type. Note that this is available even without the canary flag, but any attempt to start a CodeQL debug session will display an error stating that the feature is not yet available. It also adds the new commands.

`debug-configuration.ts` resolves the debug configuration from "launch.json" to the actual launch parameters needed by the debug engine. VS Code calls the `resolveDebugConfiguration()` function to fill in defaults before any `${...}` variables are resolved, and then calls `resolveDebugConfigurationWithSubstitutedVariables()` to give us a chance to finalize the launch configuration after variable substitution. It is in the latter function that we determine the quickeval context if needed, and validate that the user has specified a query and a database.

`debug-protocol.ts` defines a few custom messages we've added beyond the standard DAP messages. We've added one new request (to run a quickeval with a specific selection), and two events (one at the start of evaluation that reports the output directory and quickeval context, and one at the end of evaluation that reports the evaluation time and any error message).

`debug-session.ts` is the actual implementation of the DAP, using a helper library provided by VS Code. This is actually pretty simple, since it just invokes the query server to do the heavy lifting.

`debugger-factory.ts` implements the factory that creates `DebugSession` objects when VS Code starts a new debug session.

 `debugger-ui.ts` implements the new debugger commands, and handles the display of results after each successful evaluation.

`local-databases-ui.ts` doesn't really do anything new, but I did have to update how it handles progress when prompting the user to select a database. Before, it assumed that it always had a progress callback provided by the invoking command. However, when invoked during debug launch to fill in a missing `database` property on the configuration, it needs to create its own progress bar if it has to load a database.

`local-databases.ts` now separates the creation of a `DatabaseItem` from the registration of a `DatabaseItem` in the list of available databases. This allows a debug configuration to specify a path to a database that might not be present in the database manager for the current workspace.

`local-queries.ts` and `run-queries-shared.ts` got a bit of refactoring around how we compute query configuration. I've split up `determineSelectedQuery()` so that each configuration property (query, quickeval, etc.) can be computed independently. I also removed the assumption that the quickeval selection is in the same document as the actual query being compiled.

## Tests
The new debugger tests I've added exercise the new commands, and validate that we evaluation the correct files and selections, and get the correct number of results. This is all done via a `DebugController` class that turns the event-driven style of the debug adapter into an `async`/`await` based sequential style. The test issues debugger commands, then asserts that they caused the expected events.

I've also added new flavors to the existing query tests to perform the same evaluation while going through the debugger instead of `localQueries`. This also uses the `DebugController` when using the debugger.


extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -300,13 +437,25 @@ export class QLDebugSession extends LoggingDebugSession implements Disposable {
...
+    this.queryRun = undefined;
+  }
+
+  private terminateAndExit(): void {
     // Report the debugging session as terminated.
     this.sendEvent(new TerminatedEvent());
 
     // Report the debuggee as exited.
-    this.sendEvent(new ExitedEvent(result.resultType));
+    this.sendEvent(new ExitedEvent(this.lastResult!.resultType));
```
aeisenberg, line 569: I guess here, `!` is unavoidable.

extensions/ql-vscode/src/local-databases-ui.ts
```diff
@@ -754,31 +776,40 @@ export class DatabaseUI extends DisposableObject {
...
-        this.databaseManager,
-        this.storagePath,
-        progress,
-        token,
-        this.queryServer?.cliServer,
-      );
-    }
+    return await withInheritedProgress(
+      progress,
+      async (progress, token) => {
```
aeisenberg, line 788: I see that you're not inheriting and cancellation token.  I haven't looked closely, but I would have expected this to be the case.

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -450,40 +428,52 @@ export async function determineSelectedQuery(
...
+    throw new Error("Can't run quick evaluation without an active editor.");
+  }
+  // For Quick Evaluation, the selected position comes from the active editor, but it's possible
+  // that query itself was a different file. We need to validate the path of the file we're using
+  // for the QuickEval selection in case it was different.
+  validateQueryUri(editor.document.uri, true);
+  const quickEvalPosition = await getSelectedPosition(editor, range);
+  let quickEvalText: string;
+  if (!editor.selection?.isEmpty) {
+    quickEvalText = editor.document.getText(editor.selection);
```
aeisenberg, line 458: Maybe?

```suggestion
    quickEvalText = editor.document.getText(editor.selection).trim();
```

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -333,6 +392,50 @@ export class QLDebugSession extends LoggingDebugSession implements Disposable {
...
+    command: string,
+    response: CodeQLDebugProtocol.Response,
+    args: any,
+    request?: DebugProtocol.Request,
+  ): void {
+    switch (command) {
+      case "codeql-quickeval": {
+        this.quickEvalRequest(
+          response,
+          <CodeQLDebugProtocol.QuickEvalRequest["arguments"]>args,
```
aeisenberg: We're generally using this style for type casts. 

```suggestion
          args as CodeQLDebugProtocol.QuickEvalRequest["arguments"],
```

aeisenberg (commented):
That's a lot of code. I went through it commit by commit, so some of my comments may be outdated because of changes, but they might still be relevant.

I haven't tried this out, but will shortly.

aeisenberg:

> [GitHub.vscode-codeql]: Menu item references a command `codeQL.debug.quickEvalContextEditor` which is not defined in the 'commands' section.

aeisenberg:
```
      {
      },
      {
      },
```

aeisenberg:
Nice. I have it working. Do you have some easy examples of things that work differently in debug mode than with regular evaluation.

Minor thing: It would be nice if I saw some evidence of which launch configuration gave rise to a query result in the history view. Right now, there's no way to tell if something was debugged or if it was a regular launch. This is something to work on later.

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -0,0 +1,587 @@
...
+    if (this.tokenSource !== undefined) {
+      this.tokenSource!.dispose();
+      this.tokenSource = undefined;
+    }
+  }
+
+  private cancelEvaluation(): void {
+    if (this.tokenSource !== undefined) {
+      this.tokenSource.cancel();
+      this.disposeTokenSource();
```
koesie10, line 584: Would a single call to this method be enough instead of also calling `dispose` here?

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -0,0 +1,587 @@
...
+    this.sendEvent(new ExitedEvent(this.lastResult!.resultType));
+
+    this.state = "terminated";
+  }
+
+  private disposeTokenSource(): void {
+    if (this.tokenSource !== undefined) {
+      this.tokenSource!.dispose();
+      this.tokenSource = undefined;
+    }
```
koesie10, line 578:
```suggestion
    this.tokenSource?.dispose();
    this.tokenSource = undefined;
```

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -0,0 +1,587 @@
...
+  }
+
+  protected configurationDoneRequest(
+    response: Protocol.ConfigurationDoneResponse,
+    args: Protocol.ConfigurationDoneArguments,
+    request?: Protocol.Request,
+  ): void {
+    super.configurationDoneRequest(response, args, request);
+  }
```
koesie10, line 214: Do we need this method implementation? It seems to match `super.configurationDoneRequest` exactly.

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -0,0 +1,587 @@
...
+    command: string,
+    response: CodeQLProtocol.Response,
+    args: any,
+    request?: Protocol.Request,
+  ): void {
+    switch (command) {
+      case "codeql-quickeval": {
+        this.quickEvalRequest(
+          response,
+          <CodeQLProtocol.QuickEvalRequest["arguments"]>args,
```
koesie10, line 405:
```suggestion
          args as CodeQLProtocol.QuickEvalRequest["arguments"],
```

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -0,0 +1,587 @@
...
+          0,
+        );
+        progressStart.body.cancellable = true;
+        this.sendEvent(progressStart);
+        let result: CoreCompletedQuery;
+        try {
+          result = await this.queryRun.evaluate(
+            (p) => {
+              const progressUpdate = new ProgressUpdateEvent(
+                this.queryRun!.id,
```
koesie10, line 504: To avoid using `!` here, I would propose storing the query run in a local variable as well as storing it in the instance variable.

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -0,0 +1,587 @@
...
+                p.message,
+                (p.step * 100) / p.maxStep,
+              );
+              this.sendEvent(progressUpdate);
+            },
+            this.tokenSource!.token,
```
koesie10, line 510: Same here, we should store this in a local variable to avoid using `!`.

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -0,0 +1,587 @@
...
+            this.createLogger(),
+          );
+        } finally {
+          // Report the end of the progress
+          this.sendEvent(new ProgressEndEvent(this.queryRun!.id));
+        }
+        this.completeEvaluation(result);
+      } catch (e) {
+        const message = e instanceof Error ? e.message : "Unknown error";
```
koesie10, line 519: You can use `getErrorMessage` here to get the error message.

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -0,0 +1,587 @@
...
+        this.completeEvaluation({
+          resultType: QueryResultType.OTHER_ERROR,
+          message,
+          evaluationTime: 0,
+        });
+      }
+    } finally {
+      this.disposeTokenSource();
```
koesie10, line 527: Would calling `tokenSource.dispose()` on a local variable here work just as well?

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -0,0 +1,587 @@
...
+    // Report the debugging session as terminated.
+    this.sendEvent(new TerminatedEvent());
+
+    // Report the debuggee as exited.
+    this.sendEvent(new ExitedEvent(this.lastResult!.resultType));
+
+    this.state = "terminated";
+  }
+
+  private disposeTokenSource(): void {
```
koesie10, line 574: It would be nice if we could use something similar to `DisposableObject` here so you can call `track` at the start of the `evaluate` method and `disposeAndStopTracking` at the end of that method without needing to keep track of the `tokenSource` manually.

We can't use `DisposableObject` directly since we can only extend 1 class, but keeping a list of tracked objects might be useful to copy.

extensions/ql-vscode/src/debugger/debugger-factory.ts
```diff
@@ -0,0 +1,57 @@
...
+      throw new Error("The CodeQL debugger feature is not available yet.");
+    }
+    return new DebugAdapterInlineImplementation(
+      new QLDebugSession(this.queryStorageDir, this.queryRunner),
+    );
+  }
+
+  private handleOnDidStartDebugSession(session: DebugSession): void {
+    const config = session.configuration;
+    void config;
```
koesie10, line 55: Why do we need this? It seems like this method could also be empty.

extensions/ql-vscode/src/debugger/debugger-factory.ts
```diff
@@ -0,0 +1,57 @@
...
+    this.push(debug.registerDebugAdapterDescriptorFactory("codeql", this));
+    this.push(
+      debug.registerDebugConfigurationProvider(
+        "codeql",
+        new QLDebugConfigurationProvider(localQueries),
+        DebugConfigurationProviderTriggerKind.Dynamic,
+      ),
+    );
+
+    this.push(debug.onDidStartDebugSession(this.handleOnDidStartDebugSession));
```
koesie10, line 38: This method is not bound to `this`, so we wouldn't have access to `this` variables here:

```suggestion
    this.push(debug.onDidStartDebugSession(this.handleOnDidStartDebugSession.bind(this)));
```

extensions/ql-vscode/src/debugger/debugger-ui.ts
```diff
@@ -0,0 +1,245 @@
...
+    private readonly app: App,
+    private readonly localQueryResultsView: ResultsView,
+    private readonly localQueries: LocalQueries,
+    private readonly dbm: DatabaseManager,
+  ) {
+    super();
+
+    if (isCanary()) {
+      this.push(debug.registerDebugAdapterTrackerFactory("codeql", this));
+    }
```
koesie10, line 156: This means that the user needs to restart VS Code when enabling the canary flag. Should we be handling changes to the canary flag here or is this a short-lived feature flag?

extensions/ql-vscode/src/debugger/debugger-ui.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    return this.getTrackerForSession(session);
+  }
+
+  public async showResultsForCompletedQuery(
+    query: CompletedLocalQueryInfo,
+    forceReveal: WebviewReveal,
+  ): Promise<void> {
+    await this.localQueryResultsView.showResults(query, forceReveal, false);
+  }
```
koesie10, line 244: This method seems to be unused. Do we still need this to be here?

extensions/ql-vscode/test/vscode-tests/cli-integration/debug-controller.ts
koesie10, line 1: Can we move the debugger integration tests to a `debugger` subdirectory to make the separation clear?

extensions/ql-vscode/test/vscode-tests/cli-integration/debug-controller.ts
```diff
@@ -0,0 +1,432 @@
...
+
+  public createDebugAdapterTracker(
+    session: DebugSession,
+  ): ProviderResult<DebugAdapterTracker> {
+    return new Tracker(session, this);
+  }
+
+  public async createLaunchJson(config: QLDebugConfiguration): Promise<void> {
+    const launchJsonPath = join(
+      workspace.workspaceFolders![0].uri.fsPath,
```
koesie10, line 214:
```suggestion
     getOnDiskWorkspaceFolders()[0],
```

extensions/ql-vscode/test/vscode-tests/cli-integration/debugger.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  getActivatedExtension,
+} from "../global.helper";
+import { describeWithCodeQL } from "../cli";
+import { withDebugController } from "./debug-controller";
+import { CodeQLCliServer } from "../../../src/cli";
+import { QueryOutputDir } from "../../../src/run-queries-shared";
+import { createVSCodeCommandManager } from "../../../src/common/vscode/commands";
+import { AllCommands } from "../../../src/common/commands";
+
+jest.setTimeout(30_000);
```
koesie10, line 17: This can be removed after merging in `main`.

extensions/ql-vscode/package.json
```diff
@@ -1322,7 +1444,7 @@
       "editor/context": [
         {
           "command": "codeQL.runQueryContextEditor",
-          "when": "editorLangId == ql && resourceExtname == .ql"
+          "when": "editorLangId == ql && resourceExtname == .ql && !inDebugMode"
```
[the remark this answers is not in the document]
dbartol: Yeah, that's the existing "CodeQL: Run Query" command. I just changed the `when` clause to disable it when we're already debugging. Otherwise, the user is likely to select "Run Query", which will run the query outside the debugger and without debug context, rather than just "Continue".

Once we have the basic debugger support merged, we'll have to take a careful look at the UX around the commands, configuration, etc. What I've got now is intended to be the easiest-to-implement way to enable the new features, not necessarily the best UX.

extensions/ql-vscode/src/debugger/debug-configuration.ts
```diff
@@ -0,0 +1,78 @@
...
+    debugConfiguration: DebugConfiguration,
+    _token?: CancellationToken,
+  ): DebugConfiguration {
+    const qlConfiguration = <QLDebugConfiguration>debugConfiguration;
+
+    // Fill in defaults
+    const resultConfiguration: QLDebugConfiguration = {
+      ...qlConfiguration,
+      query: qlConfiguration.query ?? "${file}",
+      database: qlConfiguration.database ?? "${command:currentDatabase}",
```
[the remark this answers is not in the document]
dbartol: I think that has subtly different behavior. If `query` was specifically set to `undefined`, then your suggestion will preserve the `undefined` value, but mine will overwrite the `undefined` value with the `"${file}"` value. I'm pretty sure the current implementation is what we want for this particular case.

extensions/ql-vscode/src/debugger/debug-configuration.ts
```diff
@@ -0,0 +1,78 @@
...
+      await showAndLogErrorMessage(
+        "No database was specified in the debug configuration.",
+      );
+      return null;
+    }
+
+    const resultConfiguration: QLResolvedDebugConfiguration = {
+      ...qlConfiguration,
+      query: qlConfiguration.query,
+      database: qlConfiguration.database,
```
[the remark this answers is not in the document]
dbartol: They're copying the query and database paths from the input configuration to the result configuration, right?

extensions/ql-vscode/src/debugger/debug-session.ts
dbartol: I realized that I was only saving `lastResult` for the exit code (`resultType`), so I changed it to just save `resultType` with a default value.

dbartol:
Since there were a lot of comments about the various `!`s needed in the debug session code to handle properties that are only defined in certain states, I took another approach and factored that evaluation and related resource management out into a `RunningQuery` class, similar in spirit to the `CoreQueryRun` class except that it also handles the debugger-specific cancellation token source, logging, and progress. Now, the only state-dependent property on the session is `runningQuery`, which seems to make the code a bit simpler. I've resolved all of the `!`-related comments as no longer relevant, but of course please do comment on the new approach.

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -0,0 +1,324 @@
...
+
+  private async evaluate(
+    response: DebugProtocol.LaunchResponse,
+  ): Promise<void> {
+    // Send the response immediately. We'll send a "stopped" message when the evaluation is complete.
+    this.sendResponse(response);
+
+    const args = this.args!;
+
+    this.tokenSource = new CancellationTokenSource();
```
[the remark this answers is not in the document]
dbartol: That's true today. It's possible for a single debug session to support multiple running processes, but there's no motivation to do that now since it's easy enough to let VS Code launch a separate session for each query. If/when we move the debug adapter into the CLI, we'd probably want to implement the multi-launch support so that we can just have one instance of the adapter, running in the query server process.

extensions/ql-vscode/src/debugger/debugger-factory.ts
```diff
@@ -0,0 +1,57 @@
...
+  public createDebugAdapterDescriptor(
+    _session: DebugSession,
+    _executable: DebugAdapterExecutable | undefined,
+  ): ProviderResult<DebugAdapterDescriptor> {
+    if (useInlineImplementation) {
+      return new DebugAdapterInlineImplementation(
+        new QLDebugSession(this.queryStorageDir, this.queryRunner),
+      );
+    } else {
+      return new DebugAdapterServer(2112);
```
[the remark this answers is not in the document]
dbartol: I haven't implemented the out-of-proc debug adapter support, so I just removed that unused code path. If/when we do implement it, I expect that the query server will tell us which port it's listening on at startup.

extensions/ql-vscode/src/debugger/debugger-ui.ts
```diff
@@ -0,0 +1,167 @@
+import {
+  DebugAdapterTracker,
+  DebugAdapterTrackerFactory,
+  DebugSession,
+  debug,
+  //  window,
```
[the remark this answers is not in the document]
dbartol: Gone

extensions/ql-vscode/src/debugger/debugger-ui.ts
```diff
@@ -0,0 +1,167 @@
...
+    super();
+
+    this.push(debug.registerDebugAdapterTrackerFactory("codeql", this));
+  }
+
+  public createDebugAdapterTracker(
+    session: DebugSession,
+  ): DebugAdapterTracker | undefined {
+    if (session.type === "codeql") {
+      const tracker = new QLDebugAdapterTracker(
```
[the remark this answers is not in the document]
dbartol: Added comment about where this is being disposed.

extensions/ql-vscode/src/local-databases-ui.ts
dbartol: The `ProgressContext` contains both the progress callback and the token from the parent. If there's a parent context, we'll inherit its token. Otherwise, `withProgress` will create a new one if needed.

extensions/ql-vscode/src/debugger/debug-session.ts
dbartol: Fixed

dbartol:
> Nice. I have it working. Do you have some easy examples of things that work differently in debug mode than with regular evaluation.

Everything should work the same for now, except that using the debugger mode will respect configuration from `launch.json`. So, if you edit your `launch.json` to add, say, `database:` or `query:` paths, you'll use those instead of the current database and current file.

The main thing that works differently _because_ of configuration is QuickEval in a `.qll` file. If a query adds additional overrides of an `abstract` class, the QuickEval will pick them up in debug mode. One of the debugger tests tests this case specifically.

extensions/ql-vscode/src/debugger/debug-configuration.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  public async resolveDebugConfigurationWithSubstitutedVariables(
+    _folder: WorkspaceFolder | undefined,
+    debugConfiguration: DebugConfiguration,
+    _token?: CancellationToken,
+  ): Promise<DebugConfiguration | null> {
+    const qlConfiguration = <QLDebugConfiguration>debugConfiguration;
+    if (qlConfiguration.query === undefined) {
+      await showAndLogErrorMessage(
```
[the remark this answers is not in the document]
dbartol: Fixed both places in this file, plus one from my previous PR in `run-queries.ts` that had the same problem. Searching for `await showAndLog` in the rest of our codebase returns only one more hit, which appears to await it on purpose to consume the result.

extensions/ql-vscode/src/debugger/debug-configuration.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    // Fill in defaults here, instead of in `resolveDebugConfiguration`, to avoid the highly
+    // unusual case where one of the computed default values looks like a variable substitution.
+    const additionalPacks =
+      qlConfiguration.additionalPacks === undefined
+        ? getOnDiskWorkspaceFolders()
+        : typeof qlConfiguration.additionalPacks === "string"
+        ? [qlConfiguration.additionalPacks]
+        : qlConfiguration.additionalPacks;
```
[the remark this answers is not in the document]
dbartol: Done.

extensions/ql-vscode/src/debugger/debug-configuration.ts
```diff
@@ -0,0 +1,123 @@
...
+    // path.
+    const extensionPacks =
+      qlConfiguration.extensionPacks === undefined
+        ? await this.localQueries.getDefaultExtensionPacks(additionalPacks)
+        : typeof qlConfiguration.extensionPacks === "string"
+        ? [qlConfiguration.extensionPacks]
+        : qlConfiguration.extensionPacks;
+
+    const quickEval = qlConfiguration.quickEval ?? false;
+    validateQueryPath(qlConfiguration.query, quickEval);
```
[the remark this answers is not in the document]
dbartol: Interestingly, it pops up an OS-native message box with the error message, with a button to open the offending `launch.json` file. I've added a `try/catch` to just use `showAndLogErrorMessage()` and then return `null`. Note that if we return `null`, VS Code just opens `launch.json` automatically.

extensions/ql-vscode/src/debugger/debug-session.ts
dbartol: Nope. I must have added it to set a breakpoint more easily. Removed.

extensions/ql-vscode/src/debugger/debug-session.ts
dbartol: Fixed

extensions/ql-vscode/src/debugger/debug-session.ts
dbartol: I've added a new `RunningQuery` class just to encapsulate all of this stuff in a single `Disposable`. The result seems quite a bit cleaner than before.

extensions/ql-vscode/src/debugger/debugger-factory.ts
dbartol: That was just for debugging. Removed.

extensions/ql-vscode/src/debugger/debugger-factory.ts
dbartol: Removed because the handler didn't do anything anyway.

extensions/ql-vscode/src/debugger/debugger-ui.ts
dbartol: It turns out that I can just remove the canary check here. There's no way to conditionally register a new debugger type, so the "codeql" debugger type is always registered (by `package.json`). Instead, there's code to test the canary flag and show an error if the user manages to create a "codeql" debug configuration and try to launch it. Since the debug session will fail to start, there's no harm in registering the debug adapter tracker unconditionally.

extensions/ql-vscode/test/vscode-tests/cli-integration/debug-controller.ts
dbartol: Done

extensions/ql-vscode/test/vscode-tests/cli-integration/debugger.test.ts
dbartol: Done

dbartol:
OK, I think I've addressed all of the feedback from both @aeisenberg and @koesie10.

extensions/ql-vscode/src/debugger/debug-configuration.ts
```diff
@@ -35,6 +36,15 @@ export type QLDebugConfiguration = DebugConfiguration & QLDebugArgs;
 export type QLResolvedDebugConfiguration = DebugConfiguration &
   CodeQLProtocol.LaunchConfig;
 
+/** If the specified value is a single element, then turn it into an array containing that element. */
+function makeArray<T extends Exclude<any, any[]>>(value: T | T[]): T[] {
```
aeisenberg, line 40: Curious...why `extends Exclude<any, any[]>`?

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -131,21 +128,129 @@ const QUERY_THREAD_ID = 1;
...
+      };
+    }
+  }
+
+  /**
+   * Attempts to cancel the running evaluation.
+   */
+  public cancel(): void {
+    this.tokenSource.cancel();
+  }
```
aeisenberg, line 233: Do you think we should be `cancel`ing the RunningQuery when it is disposed? It might be that the RunningQuery is disposed, but the query itself is still running. This can cause flakes in tests.

extensions/ql-vscode/src/debugger/debug-session.ts
```diff
@@ -155,7 +260,9 @@ export class QLDebugSession extends LoggingDebugSession implements Disposable {
   }
 
   public dispose(): void {
-    this.cancelEvaluation();
+    if (this.runningQuery !== undefined) {
+      this.runningQuery.cancel();
+    }
```
aeisenberg, line 265: Oh....I see you did here exactly what I suggested above. Can you change this so `this.runningQuery` is disposed here and in the `RunningQuery` class itself, the dispose method calls cancel?

extensions/ql-vscode/src/debugger/debug-configuration.ts
dbartol: Otherwise `makeArray<number[]>([ 5, 6, 7 ])` wouldn't know whether to return the input or wrap it in an array without some deeper runtime type inspection.

extensions/ql-vscode/src/debugger/debug-session.ts
dbartol: I'm not sure it's OK to dispose the token source until we're sure the cancellation has actually been observed by the running query. The way I've implemented it, the running query (and its token source) are disposed after query completes (even if cancelled), and the debug session just "lets go" of the running query.

extensions/ql-vscode/src/debugger/debug-session.ts
dbartol: See below

extensions/ql-vscode/src/debugger/debug-configuration.ts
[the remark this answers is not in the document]
dbartol: I think you're looking at an older version of this file. The current version doesn't have the `...qlConfiguration`.

dbartol:
@aeisenberg I've fixed the zombie evaluation issue we discussed in https://github.com/github/vscode-codeql/pull/2291/commits/a151adec06acc359eadc8eb97ddb5bc9bffe55ef.

I just null out `this.runningQuery` when we do a forcible disconnect, and then stop sending progress and log output events when `this.runningQuery` is undefined.

aeisenberg (approved):
Nice...let's 🚢

PR #2295, opened 2023-04-11
Add configurable model filename to data extension editor

koesie10:
This adds a pickable model filename from an existing extension pack to the data extensions editor. This allows the user to edit one of their existing data extensions. This does not yet add the ability to create new extension packs and/or new model files.

This uses the `codeql resolve extensions` command to get the list of available model files. This should be available in all CLI versions which the data extensions editor supports.

https://github.com/github/vscode-codeql/assets/1112623/dda48636-e0dc-46b5-aca7-631a54099e6c

Based on #2293

## Checklist


extensions/ql-vscode/src/data-extensions-editor/extension-packs.ts
```diff
@@ -0,0 +1,92 @@
...
+  // Find the existing model files in the extension pack
+  const additionalPacks = getOnDiskWorkspaceFolders();
+  const extensions = await cliServer.resolveExtensions(
+    extensionPackPath,
+    additionalPacks,
+  );
+
+  const modelFiles = new Set<string>();
+
+  if (extensionPackPath in extensions.data) {
```
aeisenberg, line 63: What happens if this if statement is false? This is a legitimate case if the qlpack does not have any data extensions in it (but it might have some ml-model extensions).

Then the array will be empty and no options will be presented in the quick picker. Is this handled by the follow up PR where you add an option to create a new file?

Maybe we can skip this modal in this case and go right to askign the user to create a new file.

koesie10: Good point, I'll add this in the follow-up PR (#2296) so that we skip this quick pick in case it's empty.

koesie10: I've added this in https://github.com/github/vscode-codeql/pull/2296/commits/59c0610122843951afba8271b8a170bf9c64f6aa.

extensions/ql-vscode/src/cli.ts
```diff
@@ -107,6 +107,20 @@ export type MlModelInfo = {
 /** The expected output of `codeql resolve ml-models`. */
 export type MlModelsInfo = { models: MlModelInfo[] };
 
+export type DataExtensionInfo = {
```
charisk, line 110: Appreciate we use `Info` in other places in the codebase but I think it's good to avoid it - all types carry information. Is this simply a `DataExtension`? Or a `CodeQLDataExtension`?

extensions/ql-vscode/src/cli.ts
```diff
@@ -107,6 +107,20 @@ export type MlModelInfo = {
...
+  predicate: string;
+  file: string;
+  index: number;
+};
+
+/** The expected output of `codeql resolve extensions`. */
+export type ExtensionsInfo = {
```
charisk, line 117: Is this just `Extensions` or `CodeQLExtensions` ?

extensions/ql-vscode/src/data-extensions-editor/extension-packs.ts
```diff
@@ -0,0 +1,99 @@
...
+import { ProgressCallback } from "../progress";
+
+export async function pickExtensionPack(
+  cliServer: CodeQLCliServer,
+  progress: ProgressCallback,
+): Promise<string | undefined> {
+  progress({
+    message: "Resolving extension packs...",
+    step: 1,
+    maxStep: 3,
```
charisk, line 14: There's an implicit knowledge here that this function is called by something that has 3 steps (the `DataExtensionsEditorModule`). That feels a bit strange to me. Should we move the logic out of the module and into here?

extensions/ql-vscode/src/data-extensions-editor/extension-packs.ts
```diff
@@ -0,0 +1,99 @@
+import { relative } from "path";
+import { window } from "vscode";
+import { CodeQLCliServer } from "../cli";
+import { getOnDiskWorkspaceFolders, showAndLogErrorMessage } from "../helpers";
+import { ProgressCallback } from "../progress";
+
+export async function pickExtensionPack(
```
charisk, line 7: Should we be adding some tests for this?

charisk (commented):
I don't know much about the area but the approach sounds sensible to me.

extensions/ql-vscode/src/cli.ts
koesie10: Good point, I'll rename them to match what they are actually returning: `ResolveExtensionsResult` and `DataExtensionResult`. This also matches the names in the CLI.

charisk: The names still don't quite make sense to me but it might be because of my unfamiliarity with the area. Names that come up in my head are `ExtensionResolutionResult` and `ResolvedExtension`. But see what you prefer.

extensions/ql-vscode/src/data-extensions-editor/extension-packs.ts
```diff
@@ -0,0 +1,118 @@
+import { relative } from "path";
+import { window } from "vscode";
+import { CodeQLCliServer } from "../cli";
+import { getOnDiskWorkspaceFolders, showAndLogErrorMessage } from "../helpers";
+import { ProgressCallback } from "../progress";
+
+const maxStep = 3;
```
charisk, line 7: Hmm what do you think about renaming the file to be `extension-pack-picker` or something like that? Then `maxStep` would make sense because it would only be relevant to picking an extension pack model file.

The alternative is ofc to rename the constant to something more specific. I don't know how we see this file going so I don't have an opinion.

extensions/ql-vscode/src/cli.ts
koesie10: I think `ResolveExtensionsResult` and `DataExtensionResult` make it easier to find the definition of these types in the CLI codebase, and changing it to `ExtensionResolutionResult` will make it harder to match those.

extensions/ql-vscode/src/data-extensions-editor/extension-packs.ts
koesie10: The intention of this file is to only contain the extension pack/model file picker, so I've renamed it to `extension-pack-picker.ts`.

charisk (approved):

PR #2296, opened 2023-04-11
Allow creating new model file in existing data extension

koesie10:
This will allow a user to create a new model file in an existing data extension when opening the data extension editor. There is some validation on the name of the model file, which depends on reading in the qlpack.

https://github.com/github/vscode-codeql/assets/1112623/f8904d6c-294d-4f73-baba-9ce5061268a0

Based on #2295

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #2300, opened 2023-04-12
Add creating extension packs when opening the editor

koesie10:
This adds the ability to create a new extension pack when opening the data extension editor. It lets the user select a workspace folder and package name, but does not currently allow selecting the actual directory where it will be saved. For now, this is always in a subfolder of the workspace folder.

https://github.com/github/vscode-codeql/assets/1112623/534588bf-764c-4abb-a025-f3085feffd09

Based on #2296

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
@@ -1,15 +1,25 @@
-import { relative, resolve, sep } from "path";
-import { pathExists, readFile } from "fs-extra";
-import { load as loadYaml } from "js-yaml";
+import { join, relative, resolve, sep } from "path";
+import { outputFile, pathExists, readFile } from "fs-extra";
+import { dump as dumpYaml, load as loadYaml } from "js-yaml";
 import { minimatch } from "minimatch";
 import { CancellationToken, window } from "vscode";
 import { CodeQLCliServer } from "../cli";
-import { getOnDiskWorkspaceFolders, showAndLogErrorMessage } from "../helpers";
+import {
+  getOnDiskWorkspaceFolders,
+  getOnDiskWorkspaceFoldersObjects,
+  showAndLogErrorMessage,
+} from "../helpers";
 import { ProgressCallback } from "../progress";
 import { DatabaseItem } from "../local-databases";
 import { getQlPackPath, QLPACK_FILENAMES } from "../pure/ql";
 
 const maxStep = 3;
+
+const packNamePartRegex = /[a-z0-9](?:[a-z0-9-]*[a-z0-9])?/;
+const packNameRegex = new RegExp(
+  `^(?:(?<scope>${packNamePartRegex.source})/)?(?<name>${packNamePartRegex.source})$`,
+);
+const packNameLength = 128;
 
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
```diff
@@ -153,6 +183,89 @@ async function pickModelFile(
...
+
+        const matches = packNameRegex.exec(value);
+        if (!matches?.groups) {
+          return "Invalid package name: a pack name must contain only lowercase ASCII letters, ASCII digits, and hyphens";
+        }
+
+        const packPath = join(workspaceFolder.path, matches.groups.name);
+        if (await pathExists(packPath)) {
+          return `A pack already exists at ${packPath}`;
+        }
```
aeisenberg, line 228: What happens if this path is in a folder that already contains a qlpack (or a qlpack is in a parent folder)? I think we need to check for this case and throw an error.

It's technically possible to do this, but it's confusing and there are some edge cases where problems occur.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
```diff
@@ -153,6 +183,89 @@ async function pickModelFile(
...
+  const packYamlPath = join(packPath, "codeql-pack.yml");
+
+  await outputFile(
+    packYamlPath,
+    dumpYaml({
+      name,
+      version: "0.0.0",
+      library: true,
+      extensionTargets: {
+        "codeql/java-all": "*",
```
aeisenberg, line 260: What happens if someone tries to create an extension pack targeting something besides java? 

Either make it explicit in one of the dropdowns that this only works for java, or add a new step where users can choose the language.

koesie10: For now, we're only supporting Java in the data extension editor prototype and don't need to support any other languages yet.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
koesie10: Since this is a simple prototype, I don't think we need to handle this right now. This will almost always create a nested extension pack since the workspace folder is probably `codeql-custom-queries-java` which is a qlpack. I've added this to a list of problems we need to solve to make this production-ready.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
aeisenberg: I think that's fine as long as the extension editor clearly states that this only works for Java. (Maybe this already exists...I haven't reviewed that part yet.)

aeisenberg (approved):
Looks good. See my comment for some sort of disclaimer to add to the extension editor.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
koesie10: We don't have that yet, but it will be made clear to the users we're releasing this to and we'll be adding C# support before doing that anyway, so then this will change to `codeql/${databaseItem.language}-all`.

PR #2302, opened 2023-04-12
Change kind input to a dropdown

koesie10:
This changes the kind input from a text field to a dropdown in the data extensions editor. The supported values for each extensible predicate are based on what is currently in-scope for the documentation. Other kinds are not supported.

Screenshot 2023-04-12 at 15 45 09

## Checklist


extensions/ql-vscode/src/data-extensions-editor/predicates.ts
```diff
@@ -0,0 +1,136 @@
...
+
+export type ExtensiblePredicateDefinition = {
+  extensiblePredicate: string;
+  generateMethodDefinition: (method: ExternalApiUsageByType) => any[];
+  readModeledMethod: (row: any[]) => ModeledMethodWithSignature;
+
+  supportedKinds?: string[];
+};
+
+function readRowToMethod(row: any[]): string {
```
aeisenberg, line 21: There are only specific types supported in extension files. There's an opportunity to avoid using `any`.

```suggestion
type Tuple = boolean | number | string;
function readRowToMethod(row: Tuple[]): string {
```

extensions/ql-vscode/src/data-extensions-editor/predicates.ts
```diff
@@ -0,0 +1,136 @@
...
+      },
+    }),
+    supportedKinds: ["remote"],
+  },
+  sink: {
+    extensiblePredicate: "sinkModel",
+    // extensible predicate sinkModel(
+    //   string package, string type, boolean subtypes, string name, string signature, string ext,
+    //   string input, string kind, string provenance
+    // );
```
aeisenberg, line 62: Hmmm...this is the extensible predicate signature for static languages only. So far, we support java and csharp. It's likely that go will use the same signature, but not guaranteed.

However, dynamic languages use a different signature. We have a [document that contains a lot of information about data extensions](https://docs.google.com/document/d/14IYCHX8wWuU-HTvJ2gPSdXQKHKYbWCHQKOgn8oLaa80/edit#heading=h.ulpxg3ohlg6g). And it has a link to specific information around javascript.

https://github.com/github/codeql/blob/main/docs/codeql/codeql-language-guides/customizing-library-models-for-javascript.rst

extensions/ql-vscode/src/data-extensions-editor/predicates.ts
```diff
@@ -0,0 +1,136 @@
...
+    readModeledMethod: (row) => ({
+      signature: readRowToMethod(row),
+      modeledMethod: {
+        type: "sink",
+        input: row[6],
+        output: "",
+        kind: row[7],
+      },
+    }),
+    supportedKinds: ["sql", "xss", "logging"],
```
aeisenberg, line 83: This is a good start. Each language will have its own set of kinds. We will also need to support custom sink kinds. For instance, javascript uses `sql-injection`. `credentials[username]`, `credentials[password]`, `mongodb.sink`, and others.

I think csharp is the same as java, but i haven't done a deep analysis yet.

The ideal situation is that each language exposes the default sink/source kinds it supports through a contextual query and this editor requests that information before displaying. But, that is too much for this initial cut. For now, I don't think it makes sense to hard code this.

aeisenberg (commented):
This is getting complicated. I didn't see that the columns are hard-coded. It's not the case that all languages use the same DSL for defining their sources and sinks. Static and dynamic languages will definitely be using different DSLs. It's likely (but not guaranteed) that all static and all dynamic languages will use the same DSLs. However, I imagine that there's enough weirdness in C/C++ that it would need to use its own variant of a DSL. 

Also, sink/source kinds will be language specific, even if they all share a core set of them. We will also need to support custom sources and sink kinds (though, we might be able to punt on this particular part until later).

aeisenberg:
I'm not sure what the best option here is. Maybe we can just implement two editors: one for static and one for dynamic languages (and just assume that there will be enough commonalities within these groups to have it make sense), or we can have languages expose the structure of the editor that they require (through a contextual query or the like).

extensions/ql-vscode/src/data-extensions-editor/predicates.ts
koesie10: As discussed we're only supporting Java for now, so we fully intend to change this in the future to support other languages.

starcke (approved):
:+1: I think this is good enough for the first version, we can update it afterwards with a contextual query.

PR #2304, opened 2023-04-12
Make types for selection commands more precise

robertbrignull:
This PR aims to make our types for commands that operate on selections more precise and correct. It's possible for a lot of these commands to receive `undefined`s that they weren't expecting, and the behaviour is different for different places where the commands are registered.

I've done a lot of manual testing, logging the arguments of commands and triggering them in as many different ways as I could think of. It turns out that tree views behave differently from the file explorer, and title menus also behave differently from context menus. I've split the types in to the various cases and tried to summarise all my findings in the comments.

(btw, I know the type names are quite long. Any suggestions for better names are welcome.)

I then went and corrected types throughout the application. Thankfully we were actually handling undefined values everywhere, which just shows the types were incorrect and giving a false sense of security. The types now match the behaviour of the code and should avoid future problems.

Side-note, this sort of precise typing wouldn't be possible if we hadn't split our commands so they're only called from one place.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/commands.ts
+++ b/extensions/ql-vscode/src/common/commands.ts
@@ -13,16 +13,51 @@ import type {
 } from "../variant-analysis/shared/variant-analysis";
 
 // A command function matching the signature that VS Code calls when
-// a command on a selection is invoked.
-export type SelectionCommandFunction<Item> = (
-  singleItem: Item,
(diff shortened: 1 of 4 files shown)


charisk (approved):
Thanks for thinking through this, testing the different cases and writing helpful comments!

PR #2306, opened 2023-04-12
Use native VS Code test UI on canary

dbartol:
I decided to take a look at what it would take to move away from the third-party Test Explorer extension for CodeQL tests, and it turned out to be easier than I expected. This PR moves CodeQL testing to use the native VS Code test UI when the canary flag is set. If everything seems to work well after a bit of internal dogfooding, we can remove the old code entirely and enable this for everyone.

## Code changes
- To access the new VS Code testing API, I needed to update our `@types/vscode` dependency. This exposed a name collision in some unrelated quick pick code, which I fixed in a separate commit.

- I was able to reuse much of the existing test-related code. In particular, the test discovery code is entirely unchanged.
  - The code to run a set of tests, including unloading databases beforehand and reloading them afterward, has been factored out into a `TestRunner` service, used by both the legacy and the new test managers.
  - The implementation of the two custom test commands (`acceptOutput` and `showOutputDiffs`) has been moved into a new `TestManagerBase` class, from which both the old and new test manager classes inhierit.
- The CLI server has been updated to use `BaseLogger` instead of `Logger`, since it doesn't need the `show()` function.
- I tweaked the code that runs an async CLI command to await the child process even on failure, to avoid a runtime error about the promise rejection being unhandled.
- I added the `actual` property to `TestCompleted`, since it was already being set by the CLI.
- I implemented the new support in the new `TestManager` service. The model for interacting with the native VS Code test UI is quite a bit simpler than the third-party extension. 
  - We create a `TestController` object and update its list of `TestItem`s as we discover tests. 
  - When we run the tests, the controller passes us a `TestRunRequest` that tells us what to run, and we call the shared `TestRunner` service to handle the CLI invocation and communication.
  - Whenever a test completes, we notify the `TestRun` object. Test results can include `TestMessage` objects, which can have file locations, and can contain diffs between actual and expected text. For compilation failures, we emit the compilation errors as `TestMessage`s. For actual/expected mismatch failure, we create a `TestMessage` containing the result diff.
  - All human-readable output from the CLI invocation is logged via `TestRun.logOutput()`, which sends it to the "Test Results" tab.

## Test changes

- Moved two of the existing test adapter tests into `test-runner.test.ts`, since they really only test the functionality that is now factored out into the `TestRunner` class.
- Added a case to `test-adapter.test.ts` to test the new test controller similar to how we test the legacy one.
- Moved some shared code into a `test-runner-helpers.ts` file.


extensions/ql-vscode/src/test-manager-base.ts
```diff
@@ -0,0 +1,76 @@
...
+      const expectedUri = Uri.file(expectedPath);
+      const actualPath = getActualFile(testId);
+      const options: TextDocumentShowOptions = {
+        preserveFocus: true,
+        preview: true,
+      };
+
+      if (!(await pathExists(expectedPath))) {
+        void showAndLogWarningMessage(
+          `'${basename(expectedPath)}' does not exist. Creating an empty file.`,
```
aeisenberg, line 57: I know you are just moving this around, but it seems odd that we are popping up a warning box here. Maybe a bit annoying. I think better to just log it in the outpuut view without a popup and create the file.

extensions/ql-vscode/src/test-manager.ts
```diff
@@ -0,0 +1,371 @@
...
+import { TestManagerBase } from "./test-manager-base";
+import { App } from "./common/app";
+
+/**
+ * Returns the complete text content of the specified file. If there is an error reading the file,
+ * an error message is added to `testMessages` and this function returns undefined.
+ */
+async function tryReadFileContents(
+  path: string,
+  testMessages: TestMessage[],
```
aeisenberg, line 38: I'd slightly prefer changing this function so that it actually throws the error message and the call site is wrapped in a try-catch that would add to the testMessages array. This is more stylistic, so what you have now is also fine.

extensions/ql-vscode/src/test-manager.ts
```diff
@@ -0,0 +1,371 @@
...
+
+  public async log(message: string, options?: LogOptions): Promise<void> {
+    // "\r\n" because that's what the test terminal wants.
+    const lineEnding = options?.trailingNewline === false ? "" : "\r\n";
+    this.testRun.appendOutput(message + lineEnding);
+  }
+}
+
+/**
+ * Handles test dicovery for a specific workspace folder, and reports back to `TestManager`.
```
aeisenberg, line 78:
```suggestion
 * Handles test discovery for a specific workspace folder, and reports back to `TestManager`.
```

aeisenberg (commented):
Thanks for taking this on.

extensions/ql-vscode/src/databases/ui/db-panel.ts
```diff
@@ -34,11 +34,11 @@ import { DatabasePanelCommands } from "../../common/commands";
 import { App } from "../../common/app";
 
 export interface RemoteDatabaseQuickPickItem extends QuickPickItem {
-  kind: string;
+  remoteDatabaseKind: string;
```
charisk, line 37: Can we avoid changing these please?

charisk:
Would it be possible to extract upgrading `@types/vscode` (and resolving naming conflicts) into a separate PR please? That would make it easier to review and would give us some nicer git history around changes to the db panel code.

aeisenberg:
In a separate PR, we should probably be moving all of the `test-*.ts` files into a `test` directory. We are moving towards that for all of our other ts modules that are similar to each other.

dbartol:
> Would it be possible to extract upgrading `@types/vscode` (and resolving naming conflicts) into a separate PR please? That would make it easier to review and would give us some nicer git history around changes to the db panel code.

#2323

extensions/ql-vscode/src/test-manager.ts
dbartol: I only did it this way to avoid duplicating the code that formats and adds the test message. For a non-exported function, I think it's OK as-is.

extensions/ql-vscode/src/extension.ts
```diff
@@ -879,25 +882,34 @@ async function activateWithInstalledDistribution(
...
-    );
-    ctx.subscriptions.push(testAdapterFactory);
 
-    const testUIService = new TestUIService(app, testHub);
-    ctx.subscriptions.push(testUIService);
+  const testRunner = new TestRunner(dbm, cliServer);
+  ctx.subscriptions.push(testRunner);
+
+  let testManager: TestManagerBase | undefined = undefined;
+  if (isCanary()) {
```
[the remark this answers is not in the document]
dbartol: Hmm. There's no way to "unregister" a `TestController` AFAICT, but maybe just deleting all of its `TestItem`s would be equivalent. I'll have to see if there's a way to unregister the legacy test adapter, but it might not be worth the effort.

extensions/ql-vscode/src/test-manager-base.ts
dbartol: I've just removed the error message altogether. I don't think it's interesting to log.

extensions/ql-vscode/package.json
```diff
@@ -977,6 +977,13 @@
           "when": "viewItem == testWithSource"
         }
       ],
+      "testing/item/context": [
+        {
+          "command": "codeQLTests.acceptOutput",
+          "group": "qltest@1",
+          "when": "controllerId == codeql && testId =~ /^test /"
+        }
```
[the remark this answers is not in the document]
dbartol: Fixed.
I like these command lint warnings, even though I had to tweak the logic in my debugger PR.

extensions/ql-vscode/src/extension.ts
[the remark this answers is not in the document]
dbartol: OK, there is a way to unregister the legacy test stuff. However, handling of the commands we expose is trickier. We share the same command IDs between the two implementations, but they route to slightly different implementations depending on which test UI we're using.

Given how easy the transition to the new UI was, and how annoying our dependency on that third-party extension is, I suspect we'll move this out of canary pretty soon anyway.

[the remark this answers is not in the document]
aeisenberg: If it's going to take significant time and still be imperfect, I don't think it's worth taking on for a transitional feature.

extensions/ql-vscode/src/test-manager.ts
```diff
@@ -125,7 +125,7 @@ export class TestManager extends TestManagerBase {
     // Having this as a parameter with a default value makes passing in a mock easier.
     private readonly testController: TestController = tests.createTestController(
       "codeql",
-      "Fancy CodeQL Tests",
+      "CodeQL Tests",
```
aeisenberg, line 128: 😆

aeisenberg:

```
[GitHub.vscode-codeql]: Menu item references a command `codeQL.debug.quickEvalContextEditor` which is not defined in the 'commands' section.
```

aeisenberg:
```
      {
      },
      {
      },
```

aeisenberg:
Just deleted two comments that were meant for https://github.com/github/vscode-codeql/pull/2307

extensions/ql-vscode/src/test-adapter.ts
```diff
@@ -352,49 +243,40 @@ export class QLTestAdapter extends DisposableObject implements TestAdapter {
...
+        event.failureStage === "RESULT"
+          ? [
+              "",
+              `${state}: ${event.test}`,
+              event.failureDescription || event.diff?.join("\n"),
+              "",
+            ].join("\n")
+          : [
+              "",
+              `${event.failureStage?.toLowerCase()} error: ${event.test}`,
```
koesie10, line 264: This will print `undefined` error if the `failureStage` is empty. Should we printing an empty string or `unknown` instead?

extensions/ql-vscode/src/test-manager.ts
```diff
@@ -0,0 +1,373 @@
...
+  public dispose(): void {
+    this.workspaceFolderHandlers.clear(); // These will be disposed in the `super.dispose()` call.
+    super.dispose();
+  }
+
+  protected getTestPath(node: TestItem): string {
+    if (node.uri === undefined || node.uri.scheme !== "file") {
+      throw new Error("Selected test is not a CodeQL test.");
+    }
+    return node.uri!.fsPath;
```
koesie10, line 158: TypeScript should recognize that `node.uri` is now not undefined:

```suggestion
    return node.uri.fsPath;
```

extensions/ql-vscode/src/test-manager.ts
```diff
@@ -0,0 +1,373 @@
...
+    request: TestRunRequest,
+    token: CancellationToken,
+  ): Promise<void> {
+    const testsToRun = this.computeTestsToRun(request);
+    const testRun = this.testController.createTestRun(request, undefined, true);
+    try {
+      const tests: string[] = [];
+      testsToRun.forEach((t) => {
+        testRun.enqueued(t);
+        tests.push(t.uri!.fsPath);
```
koesie10, line 261: Can we do something different here, like returning if `t.uri === undefined`?

extensions/ql-vscode/src/cli.ts
```diff
@@ -453,9 +454,13 @@ export class CodeQLCliServer implements Disposable {
...
       }
-
-      await childPromise;
     } finally {
+      try {
+        await childPromise;
+      } catch (_e) {
+        // We need to await this to avoid an unhandled rejection, but we want to propagate the
+        // original exception.
+      }
```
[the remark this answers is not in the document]
aeisenberg: Good idea.

extensions/ql-vscode/src/test-manager.ts
dbartol: We actually already had the path we need in the key of the map entry, so I used that instead.

extensions/ql-vscode/src/test-manager.ts
dbartol: Fixed.

extensions/ql-vscode/src/test-adapter.ts
dbartol: Fixed.

extensions/ql-vscode/src/cli.ts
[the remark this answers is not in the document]
dbartol: Done.

aeisenberg (approved):

PR #2307, opened 2023-04-12
Add the CLI version to telemetry events

aeisenberg:
This adds the CLI version to telemetry command-usage events.

Note that the CLI server is created after the telemetry listener is created. The first few telemetry events may have a "not-set" value for the CLI version.

## Checklist


extensions/ql-vscode/src/telemetry.ts
```diff
@@ -157,12 +160,18 @@ export class TelemetryListener extends ConfigListener {
       ? CommandCompletion.Cancelled
       : CommandCompletion.Failed;
 
+    const cliVersion = this._cli
+      ? (await this._cli.getVersion()).toString()
```
koesie10, line 164: Should we be catching the error here and setting the `cliVersion` to something like `errored`? It seems like we could start missing command usages if the CLI server is in a bad state.

koesie10 (commented):
Thanks for adding this, this will definitely be useful!

I'm somewhat concerned about the performance impact this will have on sending commands. If there is no version available, this will spawn a process to get the CLI version, which might delay the telemetry and possibly skip sending the telemetry if VS Code gets closed early.

I also think the CLI version would be useful on other telemetry events (especially errors). I think it would be good if we could use a `addTelemetryProcessor` to add the CLI version tag to every telemetry event we send. The telemetry processor is not `async`, but we could use the cached CLI version here. That would also solve my concern about performance. We should already always be retrieving the version in the CLI server (since it's also used by CLI version constraints and the status bar), so I think this would be a very viable option for adding the CLI version to all telemetry events without the possibility of spawning a process on every telemetry event.

robertbrignull:
I agree with Koen's points. Calculating the CLI version while handling the telemetry event doesn't seem great, both in terms of performance and the chance of errors happening and stopping us sending telemetry. Could we set the CLI version once, and then maybe update it if the CLI changes? I also agree it would be nice to have this on all events, like how we include `isCanary`.

extensions/ql-vscode/src/telemetry.ts
aeisenberg: After the change, this should no longer be relevant.

aeisenberg:
I think I addressed all the comments.

aeisenberg:
Not sure why I'm getting a failing test:

https://github.com/github/vscode-codeql/actions/runs/4693003381/jobs/8320617631#step:9:69

It doesn't seem related to my changes, but it is repeatable in CI.

aeisenberg:
Maybe updating the branch will help.

aeisenberg:

> [GitHub.vscode-codeql]: Menu item references a command `codeQL.debug.quickEvalContextEditor` which is not defined in the 'commands' section.

aeisenberg:
```
      {
      },
      {
      },
```

aeisenberg:
Nope.

robertbrignull:
I'd guess we have some tests that are very close to the timeout value, and they've just tipped over recently 😢

robertbrignull:
We have an internal issue about tests timing out and I've mentioned this there. Either we should look at making the tests faster or raising the timeout value.

koesie10 (approved):
LGTM, thanks for addressing our concerns!

robertbrignull (approved):
LGTM too

aeisenberg:
Here's my analysis:

[The first error](https://github.com/github/vscode-codeql/actions/runs/4693944558/jobs/8321559046?pr=2307#step:9:129) was indeed a timeout error: 

```
    thrown: "Exceeded timeout of 20000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."
```

The tests re-ran and somewhere during the retry, the original query finally completed, but the callback failed since the connection had already been disposed. So, I think increasing the timeout will work, and we should also figure out a way to properly cancel a query in the teardown.

Also, I found another error where `removeAllDatabases` throws. This method is called during teardown and makes a call to the query server to de-register the database. If the query server is already shut down, then an error will be thrown. But if the query server is shut down, then deregistering isn't necessary, so it's ok to ignore. 

I will put a fix in for this and bump the timeouts. Both will be in a separate PR.

PR #2308, opened 2023-04-13
Query history store renaming

charisk:
Reviewed names used in the query history store, models (data/domain) and mappers, and updated them to match conventions agreed. 
- Renamed data types to be referred to as [DTOs](https://en.wikipedia.org/wiki/Data_transfer_object).
- Renamed mapping functions to use the DTO term but also to follow the same naming patterns.
- Renamed files to include `query-history` e.g. `query-history-domain-mapper`.
- Moved ALLOWED_QUERY_HISTORY_VERSIONS to store.
- Introduced an index file for the query history store.
- Tidied up code comments.

This PR looks fairly big but the changes are pretty small and straight forward. Please review commit by commit to see the differences. 

## Checklist
N/A:


extensions/ql-vscode/src/query-history/store/data-mapper.ts
```diff
@@ -54,6 +64,16 @@ function mapLocalQueryDataItemToDomainModel(
 function mapCompletedQueryInfoDataToDomainModel(
   completedQuery: CompletedQueryInfoDto,
 ): CompletedQueryInfo {
+  const sortState =
```
robertbrignull: Not that this mapper looks wrong, but why does renaming the `SortDirectionDto` type mean we now have to provide a mapper? The type itself hasn't changed but just its name. It doesn't seem like we were accidentally using a DTO type where we shouldn't have been, or accidentally referencing a domain type from the DTO type.

Is this some crazy enum magic, where you can cast them to each other if they are defined in separate files but have the same name?

robertbrignull (approved):
LGTM but I have one question where I'm confused about the code diff and commit message.

PR #2310, opened 2023-04-13
Stop pushing QL pack as top level folder to avoid confusing the user

elenatanasoiu:
Came out of https://github.com/github/vscode-codeql/pull/2250#discussion_r1158625615

In the original flow for creating skeleton packs, we were starting out
by choosing a database (e.g. `github/codeql`) and having the extension
create the QL pack for us.

At that point, we were storing the QL pack together with the database in
the extension storage because we weren't interested in committing it to
the repo.

This means we weren't able to see it in the file explorer so in order to
make it visible, we made a decision to push it as a top-level folder in the
workspace.

Hindsight is 20/20.

Let's change this original flow by just creating the folder in the
in the repo instead of the extension storage (which will make it 
visible in the file explorer) and stop pushing it as an extra top level folder 
in the workspace.

This makes it consistent with what we're doing for the skeleton query 
wizard where we start by:
- picking a language
- creating a QL pack in the repo
- downloading a database to the extension storage

For this original flow we:
- still download the database to the extension storage
- we're now creating the QL pack in the repo

Bonus: For both flows we're not pushing the QL pack as a top folder
so it won't [show up twice](https://user-images.githubusercontent.com/42641846/230116960-233ff0fc-901c-4203-97bf-d0b2fd87b79b.png) if you use the skeleton wizard flow, as Shati 
discovered.


extensions/ql-vscode/src/helpers.ts
```diff
@@ -785,3 +785,24 @@ export async function* walkDirectory(
     }
   }
 }
+
+export function getFirstStoragePath() {
```
koesie10, line 789: Now that this has moved into a helpers file, do you think there's a better name for this function? Right now it doesn't really mean anything until you look at the implementation. I think JSDoc documentation would also help here.

extensions/ql-vscode/src/helpers.ts
```diff
@@ -785,3 +785,24 @@ export async function* walkDirectory(
...
+  const workspaceFolders = workspace.workspaceFolders;
```
koesie10, line 790: Can we use `getOnDiskWorkspaceFolders` here instead of `workspace.workspaceFolders`? That will filter out any workspace folders which are not on the disk (e.g. databases).

shati-patel (commented):
Looks good! Though I think there's still some confusion around naming (left comments inline). Let me know if they don't make sense! 🙃 

💭 Rough overview of my understanding of the 3 kinds of "storage":
1. The folders that you have open in your workspace (shown in the sidebar)
2. A folder generated/managed by the extension, which stores databases and general "state" about a particular workspace. There's a different folder per workspace.
   - This is what we typically call "workspace storage".
   - Looks something like `User\workspaceStorage\e261f5bb3c3fc94ac91b18302744f6ad\GitHub.vscode-codeql\...`
3. A folder generated/managed by the extension, which stores the CodeQL CLI executable and various other "global" things.
   - This is what we typically call "global storage".
   - Looks something like `User\globalStorage\github.vscode-codeql\...`

This PR is about adding QL packs to 1, rather than 2, which is why "workspace storage" is a slightly confusing name! 🗳️

extensions/ql-vscode/test/vscode-tests/cli-integration/skeleton-query-wizard.test.ts
```diff
@@ -83,11 +83,12 @@ describe("SkeletonQueryWizard", () => {
...
       {
         name: `codespaces-codeql`,
-        uri: { fsPath: storagePath },
+        uri: { fsPath: storagePath, scheme: "file" },
       },
       {
         name: "/second/folder/path",
-        uri: { fsPath: storagePath },
+        uri: { fsPath: storagePath, scheme: "file" },
+        scheme: "file",
```
shati-patel, line 91: Typo? Or do we need both `scheme` properties? 🤔

shati-patel (commented):
Cool! 😎 Just a note from testing: now that we no longer add `codeql-custom-queries-xxx` as a top-level folder, we get into a clash with the previous "skeleton pack generator" ☠️ 😄 

E.g. When running `Create Query` in the codespaces-codeql repo, it successfully creates `codeql-custom-queries-xxx` as a subfolder of my first workspace folder, and then adds a database. 
After the database gets added, I get the message
```
We've noticed you don't have a CodeQL pack available to analyze this database. Can we set up a query pack for you?
```
which would create another pack at the top level.

Can we turn off this behaviour, or change [`isFolderAlreadyInWorkspace`](https://github.com/github/vscode-codeql/blob/02515d3b3a529bd2a5df668a2a47a35bc42a8797/extensions/ql-vscode/src/helpers.ts#L262-L269) to also check for subfolders?

extensions/ql-vscode/test/vscode-tests/cli-integration/skeleton-query-wizard.test.ts
elenatanasoiu: Definitely a typo! Thanks!

elenatanasoiu:
Oh I see what you mean by subfolders 👍

adityasharad:
Are you planning to make the choice of workspace folder configurable in future, or unconditionally create the pack within the first workspace folder?

extensions/ql-vscode/src/helpers.ts
```diff
@@ -785,3 +785,35 @@ export async function* walkDirectory(
...
+    throw new Error("No workspace folders found");
+  }
+
+  const firstFolderFsPath = workspaceFolders[0];
+
+  // For the vscode-codeql-starter repo, the first folder will be a ql pack
+  // so we need to get the parent folder
+  if (firstFolderFsPath.includes("codeql-custom-queries")) {
+    // return the parent folder
+    return dirname(firstFolderFsPath);
```
aeisenberg, line 814: Is the issue really that the first folder is a qlpack, not that we are part of the `vscode-codeql-starter` repo? If that's the case, would it be better to check if there is a `qlpack.yml` or `codeql-pack.yml` file in the specified directory?

adityasharad:
> > Are you planning to make the choice of workspace folder configurable in future, or unconditionally create the pack within the first workspace folder?
> 
> @adityasharad Not at the moment. The main target is people that are using the Codespace template. Once we get to implementing the new Queries panel in the extension, we can revisit. Do you think it's something we should consider earlier?

In an arbitrary workspace I would find it somewhat confusing as a user to have a new pack quietly written to my first workspace folder; for the Codespaces template we have more control and know exactly what the first folder is so it makes sense. Should not hold up this PR, but I think it's worth considering a follow-up of having the unconditional first-folder insertion only in the template workspace, and a user choice of folder in other workspaces.

elenatanasoiu:
Thanks for your input everyone! This PR is ready for review again.

I have some time to look into being able to configure the workspace folder. Discussed this with Shati, we want to prompt non-codespace users for a workspace folder and then save it as config.

shati-patel (commented):
One more comment from me 😎 

(I haven't got round to testing the "Don't offer to create skeleton pack again" change yet, though the code looks good 💻 )

extensions/ql-vscode/src/helpers.ts
```diff
@@ -791,3 +791,37 @@ export async function* walkDirectory(
...
+  if (!workspaceFolders || workspaceFolders.length === 0) {
+    throw new Error("No workspace folders found");
+  }
+
+  const firstFolderFsPath = workspaceFolders[0];
+
+  // For the vscode-codeql-starter repo, the first folder will be a ql pack
+  // so we need to get the parent folder
+  if (
+    firstFolderFsPath.includes("vscode-codeql-starter/codeql-custom-queries")
```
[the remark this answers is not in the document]
elenatanasoiu: Thanks Shati! You'd think I would've learned my lesson by now!

aeisenberg (approved):
Nice

elenatanasoiu:
Thanks @aeisenberg !

PR #2314, opened 2023-04-13
Add external APIs query in extension

koesie10:
This adds the external API query text to the extension directly to avoid users having to copy the query to their local `codeql` submodule or having to checkout a specific branch.

This is a temporary solution until the queries are stabilized. Once they are, we will upstream these to `github/codeql` and use them like other contextual queries.

Since this is just a temporary solution, this is not the prettiest code and is not intended to be a long-term solution. It does not do proper caching and will create a new temporary directory for every query run. The performance hit of this is acceptable and expected.

## Checklist


extensions/ql-vscode/src/data-extensions-editor/queries/query.ts
```diff
@@ -0,0 +1,6 @@
+export type Query = {
+  mainQuery: string;
+  dependencies?: {
```
starcke, line 3: I wonder if we should just inline `ExternalApi.qll`, so that we just have a single query file that we run instead of having to store multiple files?

starcke (approved):
Generally looks good to me, just a question about whether we can simplify.

extensions/ql-vscode/src/data-extensions-editor/queries/index.ts
```diff
@@ -0,0 +1,6 @@
+import { fetchExternalApisQuery as javaFetchExternalApisQuery } from "./java";
+import { Query } from "./query";
+
+export const fetchExternalApiQueries: Record<string, Query> = {
```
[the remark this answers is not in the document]
koesie10: We do, but the `DatabaseItem.language` is a `string`, so we'd need to cast it to a `QueryLanguage` which might hide problems rather than actually giving us type safety. I've changed it to use a `QueryLanguage` though since that will prevent us from adding unsupported types.

extensions/ql-vscode/src/data-extensions-editor/queries/query.ts
starcke: I dont think it would be too hard to do that split up later, but I dont feel strongly. It also doesnt seem super complicated to have extra dependency files because we also need to have the synthetic pack. So happy to leave it like this.

starcke (approved):
:+1:

PR #2330, opened 2023-04-14
Select latest database when creating skeleton query

elenatanasoiu:
The skeleton query process tries to choose an existing database for you instead of downloading a new one. 

When deciding which database to select, this is the process it goes through:
1. Is there one with the same name as the one we'd be downloading?
2. If that doesn't exist, is there one with the same language? 
3. If that doesn't exist either, it will go ahead and download one for you.

When we were designing how to find an existing database, we weren't sorting them by their creation date. This means that we might be selecting an older one, instead of the latest.

So let's sort our databases by `dateAdded` before attempting to select one.

### DIFF
--- a/extensions/ql-vscode/src/skeleton-query-wizard.ts
+++ b/extensions/ql-vscode/src/skeleton-query-wizard.ts
@@ -261,10 +261,11 @@ export class SkeletonQueryWizard {
       throw new Error("QL Pack storage path is undefined");
     }
 
-    const existingDatabaseItem = await this.findExistingDatabaseItem(
-      this.language,
-      this.databaseManager.databaseItems,
-    );
+    const existingDatabaseItem =
+      await SkeletonQueryWizard.findExistingDatabaseItem(
+        this.language,
+        this.databaseManager.databaseItems,
+      );
 
     if (existingDatabaseItem) {
       // select the found database
@@ -275,7 +276,7 @@ export class SkeletonQueryWizard {
     }
   }
 
-  public async findDatabaseItemByNwo(
+  public static async findDatabaseItemByNwo(
     language: string,
     databaseNwo: string,
     databaseItems: readonly DatabaseItem[],
@@ -287,7 +288,7 @@ export class SkeletonQueryWizard {
     return dbs.pop();
   }
 
-  public async findDatabaseItemByLanguage(
+  public static async findDatabaseItemByLanguage(
     language: string,
     databaseItems: readonly DatabaseItem[],
   ): Promise<DatabaseItem | undefined> {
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -253,51 +265,51 @@ export class SkeletonQueryWizard {
...
+    const dbs = databaseItems.filter((db) => db.language === language);
+
+    return dbs.pop();
   }
 
-  private async findExistingDatabaseItem() {
-    if (this.language === undefined) {
-      throw new Error("Language is undefined");
-    }
+  public async findExistingDatabaseItem(
```
robertbrignull, line 284: These functions are a little confusing as to whether they should be members of the `SkeletonQueryWizard` class or not, because we're passing in arguments that are otherwise available on `this`. It makes the functions slightly harder to implement because you have to know whether there's a `language` argument available and that you should use that instead of `this.language`. We don't want to end up using both from the same method.

We could make these methods `static` so then that removes the chance of accidentally referencing `this`, or move them out of the class so they're just top-level non-exported methods.

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -231,17 +231,29 @@ export class SkeletonQueryWizard {
...
-    const existingDatabaseItem = await this.findExistingDatabaseItem();
+    const existingDatabaseItem = await this.findExistingDatabaseItem(
+      this.language,
+      this.databaseManager.databaseItems,
+    );
 
     if (existingDatabaseItem) {
       await this.databaseManager.setCurrentDatabaseItem(existingDatabaseItem);
     } else {
       await this.downloadDatabase();
```
robertbrignull, line 250: This isn't directly related to this PR but I've noticed that we could "easily" (famous last words) make `downloadDatabase` return the newly-created database item. It seems that `downloadGitHubDatabase` already returns it, so we'd just need to return it from `downloadDatabase`. That would remove the need to call `findExistingDatabaseItem` afterwards and potentially remove a chance of race conditions.

robertbrignull (commented):
This will work as is, but I have a couple of small questions to which I'm keen to hear your thoughts. This is my first time reading this code so you'll have to forgive me if I'm missing some context or misunderstanding things.

extensions/ql-vscode/src/skeleton-query-wizard.ts
robertbrignull: I also just noticed when reviewing https://github.com/github/vscode-codeql/pull/2328 that `downloadDatabase` should theoretically already set the new database as the currently selected item. It calls through to [`databaseArchiveFetcher`](https://github.com/github/vscode-codeql/blob/4170e7f7a7026b3ebcc75b7509dd55f317293a4d/extensions/ql-vscode/src/databaseFetcher.ts#L273) which calls [`setCurrentDatabaseItem`](https://github.com/github/vscode-codeql/blob/4170e7f7a7026b3ebcc75b7509dd55f317293a4d/extensions/ql-vscode/src/databaseFetcher.ts#L326).

Would be good to test if this is working. If it isn't could it be the same single-workspace issue, so once that is fixed this will also work?

extensions/ql-vscode/src/skeleton-query-wizard.ts
elenatanasoiu: I've converted these types of methods into `static` versions.

extensions/ql-vscode/src/skeleton-query-wizard.ts
elenatanasoiu: > If it isn't could it be the same single-workspace issue, so once that is fixed this will also work?

Yes, that's currently what's happening. Because we decided to select the database when we open it, we no longer need to repeat the step here.

robertbrignull (approved):
LGTM

elenatanasoiu:
Thanks @robertbrignull !

PR #2331, opened 2023-04-14
Fix various test flakiness

aeisenberg:
This commit addresses various test flakiness:

1. Bump timeouts for queries tests
2. Add a dispose handler to queryserver-client. This will help us during tests because if there is a test that timesout while a query is running, the query's progress callback won't be invoked. We will still get a timeout error in the first test, but the second test will not get a spurious error.
3. Handle a disposed query server in `deregisterDatabase`. This method will remove the database from the currently running query server. If there is no query server, then there is nothing to remove. So, this error is safe to ignore.
4. Explicitly `end()` a connection `ServerProcess`. I'm not 100% sure if this is necessary, but it seems like it prevents responses from being handled and erroring out.
5. Better handling of ideServer restarts. Previously, if you quickly called `CodeQL: Restart Query Server` twice in a row, you would get an error from the ideServer restart. Restart fails if the server is not already started. So, in this case just call a start.

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -177,7 +177,13 @@ function getCommands(
         cliServer.restartCliServer();
         await Promise.all([
           queryRunner.restartQueryServer(progress, token),
-          ideServer.restart(),
+          async () => {
+            if (ideServer.isRunning()) {
+              await ideServer.restart();
+            } else {
+              await ideServer.start();
+            }
+          },
         ]);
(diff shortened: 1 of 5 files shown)


aeisenberg:
My change to bump the timeout probably conflicts with https://github.com/github/vscode-codeql/pull/2326. I can remove that.

aeisenberg:
Thanks for the review. It's always really hard to know if you've actually solved the issues behind a flaky test. Hopefully, this will help.

PR #2335, opened 2023-04-17
Rename supported to modelled.

starcke:
I thought that using modelled instead of supported makes more sense when we start giving people a tool to model things themselves.

I didnt rename the query because the predicates matches what is in library, but we could consider that in the future.

## Checklist


elenatanasoiu (approved):
Nice! I think `modelled` is clearer. 

I've seen it written as "modeled" in documents. 🇺🇸 Do we want to be consistent with those?

starcke:
> I've seen it written as "modeled" in documents. us Do we want to be consistent with those?
Doh, yes!

elenatanasoiu (approved):
👍

PR #2341, opened 2023-04-17
Prompt non-codespace users for storage path

elenatanasoiu:
Offer non-codespace users the option to configure their storage folder for skeleton packs.

Suggested here: https://github.com/github/vscode-codeql/pull/2310#issuecomment-1507428217

At the moment we're choosing to create our skeleton packs in the first folder in the workspace.

This is fine for the codespace template because we can control the folder structure in that repo.

For users outside of this we'd like to offer them the option to choose where to save their skeleton packs.

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -345,7 +345,7 @@
           "type": "string",
           "default": "",
           "patternErrorMessage": "Please enter a valid folder",
-          "markdownDescription": "The name of the folder where we want to create queries and query packs via the \"CodeQL: Create Query\" command. The folder should exist in the workspace."
+          "markdownDescription": "The name of the folder where we want to create queries and query packs via the \"CodeQL: Create Query\" command. The folder should exist."
         }
       }
     },
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -340,6 +340,12 @@
           "type": "boolean",
           "default": false,
           "description": "Allow database to be downloaded via HTTP. Warning: enabling this option will allow downloading from insecure servers."
+        },
+        "codeQL.createQuery.Folder": {
+          "type": "string",
+          "default": "",
+          "patternErrorMessage": "Please enter a valid folder",
+          "markdownDescription": "The name of the folder where we want to create queries and query packs via the \"CodeQL: Create Query\" command. The folder should exist in the workspace.`)."
         }
       }
     },
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -341,11 +341,11 @@
           "default": false,
           "description": "Allow database to be downloaded via HTTP. Warning: enabling this option will allow downloading from insecure servers."
         },
-        "codeQL.createQuery.Folder": {
+        "codeQL.createQuery.folder": {
           "type": "string",
           "default": "",
           "patternErrorMessage": "Please enter a valid folder",
-          "markdownDescription": "The name of the folder where we want to create queries and query packs via the \"CodeQL: Create Query\" command. The folder should exist in the workspace.`)."
+          "markdownDescription": "The name of the folder where we want to create queries and query packs via the \"CodeQL: Create Query\" command. The folder should exist in the workspace."
         }
       }
     },
--- a/extensions/ql-vscode/src/config.ts
+++ b/extensions/ql-vscode/src/config.ts
@@ -619,3 +619,19 @@ export const ALLOW_HTTP_SETTING = new Setting(
 export function allowHttp(): boolean {
   return ALLOW_HTTP_SETTING.getValue<boolean>() || false;
 }
+
+/**
+ * The name of the folder where we want to create skeleton wizard QL packs.
+ **/
+const SKELETON_WIZARD_FOLDER = new Setting(
+  "folder",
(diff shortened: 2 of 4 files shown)


extensions/ql-vscode/src/config.ts
```diff
@@ -608,3 +608,19 @@ export const CODESPACES_TEMPLATE = new Setting(
 export function isCodespacesTemplate() {
   return !!CODESPACES_TEMPLATE.getValue<boolean>();
 }
+
+/**
+ * The name of the folder where we want to create skeleton wizard QL packs.
+ **/
+const SKELETON_WIZARD_FOLDER = new Setting(
+  "folder",
+  new Setting("skeletonWizard", ROOT_SETTING),
```
shati-patel, line 617: Since this is a semi-public config option (as in, users will be able to see/edit `codeQL.skeletonWizard.folder` in their JSON settings), can we change the name?

From the users' point of view, this is where "CodeQL: Create Query" drops their query, so perhaps something like `createQuery` (instead of `skeletonWizard`) would be more intuitive? 🤔 ☠️

extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -98,6 +103,38 @@ export class SkeletonQueryWizard {
...
+      throw new UserCancellationException("No storage folder entered.");
+    }
+
+    if (!existsSync(storageFolder)) {
+      throw new UserCancellationException(
+        "Invalid folder. Must be a folder that already exists.",
+      );
+    }
+
+    await setSkeletonWizardFolder(storageFolder);
```
shati-patel, line 134: This didn't work for me, and I got the following error:
```
Unable to write to User Settings because codeQL.skeletonWizard.folder is not a registered configuration. (codeQL.createQuery)
```

Adding the config option to package.json might fix this!

shati-patel (commented):
Thanks! 💖 
We should probably add this setting to the package.json file, so we can add some documentation around it too. For example, see [`codeQL.variantAnalysis.controllerRepo`](https://github.com/github/vscode-codeql/blob/9c7979955838ec9d215224041ac7e5cd9d9203f0/extensions/ql-vscode/package.json#L283-L289) 📂

extensions/ql-vscode/src/config.ts
elenatanasoiu: Have renamed. Thanks!

extensions/ql-vscode/package.json
```diff
@@ -340,6 +340,12 @@
           "type": "boolean",
           "default": false,
           "description": "Allow database to be downloaded via HTTP. Warning: enabling this option will allow downloading from insecure servers."
+        },
+        "codeQL.createQuery.Folder": {
```
shati-patel, line 344: I'm still getting the same error locally 😿 
image

Looks like it's just due to this typo! 


```suggestion
        "codeQL.createQuery.folder": {
```

extensions/ql-vscode/package.json
```diff
@@ -340,6 +340,12 @@
...
+          "type": "string",
+          "default": "",
+          "patternErrorMessage": "Please enter a valid folder",
+          "markdownDescription": "The name of the folder where we want to create queries and query packs via the \"CodeQL: Create Query\" command. The folder should exist in the workspace.`)."
```
shati-patel, line 348: Teeny typo ⌨️ 🐈  😃 

```suggestion
          "markdownDescription": "The name of the folder where we want to create queries and query packs via the \"CodeQL: Create Query\" command. The folder should exist in the workspace."
```

shati-patel (commented):
Thanks! ✨ 

Should be good to go once the typo in the config name is fixed!

extensions/ql-vscode/package.json
```diff
@@ -340,6 +340,12 @@
           "type": "boolean",
           "default": false,
           "description": "Allow database to be downloaded via HTTP. Warning: enabling this option will allow downloading from insecure servers."
+        },
+        "codeQL.createQuery.folder": {
+          "type": "string",
+          "default": "",
+          "patternErrorMessage": "Please enter a valid folder",
+          "markdownDescription": "The name of the folder where we want to create queries and query packs via the \"CodeQL: Create Query\" command. The folder should exist in the workspace."
```
shati-patel, line 348: Minor observation:
> The folder should exist in the workspace.

We don't actually enforce that the folder should be in the workspace (it just needs to exist _somewhere_). E.g. 
- I'm in workspace `foo` and I set `codeQL.createQuery.folder` to "foo/my-folder"
- If I change to a different workspace `bar` and run "Create Query", the query still gets created in "foo/my-folder"

It's a bit odd, but probably fine as-is. Users can always manually override `codeQL.createQuery.folder` in their [workspace-specific settings](https://code.visualstudio.com/docs/getstarted/settings#_workspace-settings) 🤷🏽

shati-patel (approved):
Nice, looks good 🚢  

(I've left a small observation inline, but I don't think it's hugely important. Mainly just leaving it here in case it's useful to know in future!)

extensions/ql-vscode/package.json
elenatanasoiu: Right! Ok, I can drop the "in the workspace" part. Thanks Shati! 😍

elenatanasoiu:
Thanks @shati-patel !

PR #2354, opened 2023-04-19
Clear pack cache on create and delete

koesie10:
We were only clearing the pack cache when a pack file was modified. When using a script to create a new pack, the pack file is created at once without a change event firing. This will change the behavior to also clear the pack cache when a pack file is created or deleted.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -707,9 +707,14 @@ async function activateWithInstalledDistribution(
   for (const glob of PACK_GLOBS) {
     const fsWatcher = workspace.createFileSystemWatcher(glob);
     ctx.subscriptions.push(fsWatcher);
-    fsWatcher.onDidChange(async (_uri) => {
+
+    const clearPackCache = async (_uri: Uri) => {
       await qs.clearPackCache();
-    });
+    };
+
+    fsWatcher.onDidCreate(clearPackCache);
+    fsWatcher.onDidChange(clearPackCache);
+    fsWatcher.onDidDelete(clearPackCache);
   }
 
   void extLogger.log("Initializing database manager.");


aeisenberg (approved):
Makes sense.

aeisenberg:
> This looks good to me and it fixes the problem I had with the queryserver getting inconsistent. @aeisenberg do you have any thoughts on whether there is anything we have missed.

I can't think of anything off the top of my head, but missing cache clearing is a common oversight. Any time a structural file changes (qlpack.yml, lock file, workspace file, etc), we need to clear the cache. There is a bit of a race condition since the cache is automatically flushed every 2 seconds. So, it is easy to miss the error if you are debugging.

I wish there were a better way to ensure we clear the cache, but I can't think of any right now.

aeisenberg:
Oh...I didn't even know that was possible. We might want to augment the `clearCache` function so that it takes the qlpack or other file that was changed and sends a change event to the ide server to flush that file.

koesie10:
> I can't think of anything off the top of my head, but missing cache clearing is a common oversight. Any time a structural file changes (qlpack.yml, lock file, workspace file, etc), we need to clear the cache. There is a bit of a race condition since the cache is automatically flushed every 2 seconds. So, it is easy to miss the error if you are debugging.

@aeisenberg It doesn't seem like the cache is flushed every 2 seconds, I can still reproduce this error when I wait for at least 4 seconds between creating the extension pack and running the query again:

https://github.com/github/vscode-codeql/assets/1112623/eaaba464-05f3-4b37-b440-290e1adf1fd9

Is this a bug in the query server?

I'll merge this PR since it fixes the bug we were experiencing and don't really need the language server for our purpose since we only ran into this bug when using extension packs.

aeisenberg:
> It doesn't seem like the cache is flushed every 2 seconds, I can still reproduce this error when I wait for at least 4 seconds between creating the extension pack and running the query again:

Thanks for looking into this. I misspoke. The cache flushing only happens in the CLI server, not the query server. The CLI server will flush its cache before starting a new command if the cache is more than 1 second old. I don't see the same behaviour in the query server (likely it was never implemented, not sure if there's a reason, or if it's an oversight).

@alexet Do you know if there's a reason why we're not flushing the cache in the query server like we are in the CLI server?

PR #2350, opened 2023-04-18
Remove query history title actions

norascheuch:
Removes the title actions of the query history panel, because newest telemetry show that they are not used by users. Additionally, to unify UI behaviour, we remove them so that we don't have any title actions dependent on a specific selection.

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #2356, opened 2023-04-19
Remove determineSelection and simplify selection commands for the query history

robertbrignull:
This PR aims to simplify all our selection commands in the query history manager. The three things we do are:
- Remove the `determineSelection` method.
  - This method was needed for some commands that could be called from the title bar, because there might not be any items selected or even present. But with the merge of https://github.com/github/vscode-codeql/pull/2350 we've removed these actions.
  - I believe this wasn't strictly needed for the other commands, but we included it anyway for consistency and because nobody was quite sure what the rules and the real behaviour were.
  - `singleItem` should now always be defined, and there's some description on [`TreeViewContextMultiSelectionCommandFunction`](https://github.com/github/vscode-codeql/blob/7f3f3380adfa9b56928f128ddfe68d2952f64bc2/extensions/ql-vscode/src/common/commands.ts#L24) about when `multiSelect` is defined or not.
- Make all method parameter types match `TreeViewContextMultiSelectionCommandFunction`
  - Some of the commands defined `singleItem` as possibly-undefined, but this isn't necessary.
- Remove unnecessary checks that `singleItem` is not undefined
  - We should be able to trust our types, and therefore we don't need to test if `singleItem` is undefined or not. If our types are wrong we should change them, but I'm pretty convinced that they're correct.
- Do `multiSelect ||= [singleItem]` in commands where we do operate on multiple items
  - This is working around `multiSelect` being undefined when only a single item is selected. Why VS Code gives us the arguments this way is a mystery to me, but that's what it does.
  - Making sure that `multiSelect` was always populated was something that `determineSelection` did, so now there's a couple of places where we need to do that ourselves. This will go away again with https://github.com/github/vscode-codeql/pull/2287.

If I'm correct, then this PR won't change any behaviour. I've done my best to manually test using the hover and right-click actions in every circumstance I could think of, and the rules follow what I wrote on [`TreeViewContextMultiSelectionCommandFunction`](https://github.com/github/vscode-codeql/blob/7f3f3380adfa9b56928f128ddfe68d2952f64bc2/extensions/ql-vscode/src/common/commands.ts#L24).

This PR is another precursor to https://github.com/github/vscode-codeql/pull/2287 and it'll make that PR a lot simpler. In that PR we'll simplify method types again so they take exactly the arguments they need, and we can remove the need for the `assertSingleQuery` method.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-history/query-history-manager.ts
+++ b/extensions/ql-vscode/src/query-history/query-history-manager.ts
@@ -396,38 +396,34 @@ export class QueryHistoryManager extends DisposableObject {
   }
 
   async handleOpenQuery(
-    singleItem: QueryHistoryInfo | undefined,
+    singleItem: QueryHistoryInfo,
     multiSelect: QueryHistoryInfo[] | undefined,
   ): Promise<void> {
-    const { finalSingleItem, finalMultiSelect } = this.determineSelection(
-      singleItem,
-      multiSelect,
-    );
-    if (!this.assertSingleQuery(finalMultiSelect) || !finalSingleItem) {
+    if (!this.assertSingleQuery(multiSelect)) {
       return;
     }
 
-    if (finalSingleItem.t === "variant-analysis") {
+    if (singleItem.t === "variant-analysis") {
       await this.variantAnalysisManager.openQueryFile(
-        finalSingleItem.variantAnalysis.id,
+        singleItem.variantAnalysis.id,
       );
       return;
     }
 
     let queryPath: string;
-    switch (finalSingleItem.t) {
+    switch (singleItem.t) {
       case "local":
-        queryPath = finalSingleItem.initialInfo.queryPath;
+        queryPath = singleItem.initialInfo.queryPath;
         break;
       default:
-        assertNever(finalSingleItem);
+        assertNever(singleItem);
     }
     const textDocument = await workspace.openTextDocument(Uri.file(queryPath));
     const editor = await window.showTextDocument(textDocument, ViewColumn.One);
 
-    if (finalSingleItem.t === "local") {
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/query-history/query-history-manager.ts
```diff
@@ -627,36 +610,32 @@ export class QueryHistoryManager extends DisposableObject {
...
-    singleItem: QueryHistoryInfo | undefined,
+    singleItem: QueryHistoryInfo,
     multiSelect: QueryHistoryInfo[] | undefined,
   ) {
-    const { finalSingleItem, finalMultiSelect } = this.determineSelection(
-      singleItem,
-      multiSelect,
-    );
-    if (!this.assertSingleQuery(finalMultiSelect) || !finalSingleItem) {
+    if (!this.assertSingleQuery(multiSelect)) {
```
charisk, line 616: Not a problem with your PR but "assert" makes me think an error will be thrown if the expectation isn't met. Perhaps we should rename to `isSingleQuery` or something like that in the future.

charisk (approved):
Changes LGTM! Nice to see this getting tidied up. Thank you for writing a thorough PR description - it answered all my questions 😺

extensions/ql-vscode/src/query-history/query-history-manager.ts
robertbrignull: Yes I agree it's confusingly named. I actually plan to delete it entirely in https://github.com/github/vscode-codeql/pull/2287

robertbrignull:
> There were some tests around `determineSelection` which can now be 🔥

Ooh, I love the chance to delete even more code 🔥 

These tests are now gone!

I guess this is another case of me not running `npm run check-types`, though I thought I'd got quite good at that 😞

PR #2358, opened 2023-04-19
Fiddle with handleCompareWith to make it cleaner

robertbrignull:
The aim of this PR is to make the code for `handleCompareWith` cleaner and clearer.

I've been reading the query history commands a lot recently, and `handleCompareWith` keeps catching my eye. It's unusual because it's the only command we have that actually uses both `singleItem` and `multiSelect`, because it wants to know which of the selected items you actually clicked on so it can use that one as the "from" part of the comparison. Unfortunately the method is also rather confusingly written with lots of redundant checks and generally confusing logic.

This PR should theoretically be completely behaviour preserving. I'm not aiming to change any behaviour, but then the `this.compareWithItem` field in particular is new to me so I might have misunderstood something in its operation.

I haven't split up the commits into separate tiny changes because they're all on top of each other and I wasn't thinking about it as I wrote this. I recommend just reading the final code and seeing what you think of it. I could go back and split it up, but I hope a descriptive list will be just as good. Let me know what you think. The general list of things that are a little wrong in the original code and are changed by this PR are:
- The use of `determineSelection` is unnecessary.
  - This is a change also made in https://github.com/github/vscode-codeql/pull/2356, so see that PR for a full description and justification.
- The code checks that the items are a `LocalQueryItem` and then a `CompletedLocalQueryItem` in multiple places.
  - We check `singleItem` at the start of `handleCompareWith`, but then we check it again in `findOtherQueryToCompare`, and we only check `multiSelect` in `findOtherQueryToCompare`.
  - The error message is different in each place, and also talks as if it's about a single query but doesn't make it clear which of the selected queries it means.
  - This PR makes so we check both `singleItem` and `multiSelect` once at the start of `handleCompareWith` and provide a consistent error message.
- We use `this.compareWithItem` without checking that it is one of the selected queries.
  - This maybe theoretically is fine, since the `treeView.onDidChangeSelection` should have fired and updated `this.compareWithItem`. But maybe it's possible that didn't fire or they happen in the wrong order.
  - I check that `this.compareWithItem` is one of the elements from `multiSelect`.
  - I will write up tomorrow some info on `this.compareWithItem` because I expect most people won't have encountered it before, so I'll try to summarise how I think it works.
- We do a lot of dubious type casting because we don't utilise the type system to enforce things for us
  - I add `isSuccessfulCompletedLocalQueryInfo` and use that at the start to safely cast everything to a `CompletedLocalQueryInfo`. We can then remove a lot of casts and unnecessary error checks.
  - I also use `isSuccessfulCompletedLocalQueryInfo` in `findOtherQueryToCompare` when we're searching the list of all history items for possible comparisons.
- There's a big try-catch and we emit telemetry when really they're user errors
  - I move the try-catch to only be around `findOtherQueryToCompare` and just show the notification without telemetry.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-history/query-history-manager.ts
+++ b/extensions/ql-vscode/src/query-history/query-history-manager.ts
@@ -15,7 +15,6 @@ import {
 import { QueryHistoryConfig } from "../config";
 import {
   showAndLogErrorMessage,
-  showAndLogExceptionWithTelemetry,
   showAndLogInformationMessage,
   showAndLogWarningMessage,
   showBinaryChoiceDialog,
@@ -25,7 +24,7 @@ import { extLogger } from "../common";
 import { URLSearchParams } from "url";
 import { DisposableObject } from "../pure/disposable-object";
 import { ONE_HOUR_IN_MS, TWO_HOURS_IN_MS } from "../pure/time";
-import { asError, assertNever, getErrorMessage } from "../pure/helpers-pure";
+import { assertNever, getErrorMessage } from "../pure/helpers-pure";
 import { CompletedLocalQueryInfo, LocalQueryInfo } from "../query-results";
 import {
   getActionsWorkflowRunUrl,
@@ -54,7 +53,6 @@ import { VariantAnalysisManager } from "../variant-analysis/variant-analysis-man
 import { VariantAnalysisHistoryItem } from "./variant-analysis-history-item";
 import { getTotalResultCount } from "../variant-analysis/shared/variant-analysis";
 import { HistoryTreeDataProvider } from "./history-tree-data-provider";
-import { redactableError } from "../pure/errors";
 import { QueryHistoryDirs } from "./query-history-dirs";
 import { QueryHistoryCommands } from "../common/commands";
 import { App } from "../common/app";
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/query-history/query-history-manager.ts
```diff
@@ -573,40 +571,46 @@ export class QueryHistoryManager extends DisposableObject {
...
+      !multiSelect.every(this.isSuccessfulCompletedLocalQueryInfo)
+    ) {
+      throw new Error(
+        "Please only select local queries that have completed successfully.",
+      );
+    }
 
-      const from = this.compareWithItem || singleItem;
-      const to = await this.findOtherQueryToCompare(from, multiSelect);
+    const fromItem =
```
charisk, line 595: (Optional) I found this a little bit hard to read (because of the conditional). Would it make sense to extract that out or perhaps the whole statement could be extracted as a function?

charisk (approved):
Thanks for tidying this up, the changes LGTM. I only have some minor suggestions that you can ignore.

I've not tested this locally but it looks like there's some code coverage around it.

extensions/ql-vscode/src/query-history/query-history-manager.ts
```diff
@@ -1177,57 +1178,40 @@ export class QueryHistoryManager extends DisposableObject {
...
-      }
-      if (!otherQuery.completedQuery.successful) {
-        throw new Error("Please select a successful query.");
-      }
-      if (otherQuery.initialInfo.databaseInfo.name !== dbName) {
+    if (allItemsSelected.length === 2) {
+      const otherItem =
+        fromItem === allItemsSelected[0]
+          ? allItemsSelected[1]
+          : allItemsSelected[0];
```
elenatanasoiu, line 1084: The previous code had a comment explaining a bit why we're doing this. That meaning is lost a little bit here. Is there something we can do to make it clear again? 

🤔 We could wrap this condition in a method like you've done in other places. Something like `firstUnselectedItem()`?

extensions/ql-vscode/src/query-history/query-history-manager.ts
```diff
@@ -1177,57 +1178,40 @@ export class QueryHistoryManager extends DisposableObject {
...
+      if (otherItem.initialInfo.databaseInfo.name !== dbName) {
         throw new Error("Query databases must be the same.");
       }
-      return otherQuery as CompletedLocalQueryInfo;
+      return otherItem;
     }
 
-    if (multiSelect?.length > 2) {
+    if (allItemsSelected.length > 2) {
```
elenatanasoiu, line 1091: Could we rename this to be `allSelectedItems` ? 

`allItemsSelected` first made me think it's a boolean indicating whether all items have been selected, but it's actually an array.

extensions/ql-vscode/src/query-history/query-history-manager.ts
```diff
@@ -1177,57 +1178,40 @@ export class QueryHistoryManager extends DisposableObject {
...
       .filter(
-        (otherQuery) =>
-          otherQuery !== singleItem &&
-          otherQuery.t === "local" &&
-          otherQuery.completedQuery &&
-          otherQuery.completedQuery.successful &&
-          otherQuery.initialInfo.databaseInfo.name === dbName,
+        (otherItem) =>
+          otherItem !== fromItem &&
+          otherItem.initialInfo.databaseInfo.name === dbName,
```
elenatanasoiu, line 1101: We dig for this database name in a few places. Could we assign `otherItem.initialInfo.databaseInfo.name` to a variable at the top to make this terser, the same way we do for `fromItem`?

e.g.  `otherItemDbName`

extensions/ql-vscode/src/query-history/query-history-manager.ts
robertbrignull: Good point, that name is pretty confusing. I'll rename to `allSelectedItems`.

extensions/ql-vscode/src/query-history/query-history-manager.ts
robertbrignull: We can't really assign to a variable, because we're operating on a different `otherItem` variable each time, even if they use the same name.

What we could do is create a `getDatabaseName` method, which will make things a bit less repetitive.

elenatanasoiu (approved):
Looks good!

PR #2359, opened 2023-04-19
Add configuration option to turn off skeleton pack generation

elenatanasoiu:
We'd like to make auto generation of QL packs the default behaviour in the extension.

We have two flows at the moment that only work in the [codespaces-codeql](https://github.com/github/codespaces-codeql) template:

1. the skeleton wizard flow which triggers when you run the `CodeQL: Create query` command
2. the database panel flow which offers to create a QL pack for you when you download a database from GitHub (see prompt below)

Screenshot 2023-04-19 at 18 25 02

For the second flow, we should offer the user a third option called "No, and never ask me again". 

Screenshot 2023-04-19 at 18 54 15

This will disable auto generation. We are writing this choice to a VS Code setting called `codeQL.createQuery.autogenerateQlPacks` and checking for it before deciding whether to prompt the user again.

https://github.com/github/vscode-codeql/assets/1354439/e9219cbf-a6e8-4fa8-91c4-6e579f0310b2


extensions/ql-vscode/package.json
```diff
@@ -346,6 +346,12 @@
           "default": "",
           "patternErrorMessage": "Please enter a valid folder",
           "markdownDescription": "The name of the folder where we want to create queries and query packs via the \"CodeQL: Create Query\" command. The folder should exist."
+        },
+        "codeQL.autogenerateQlPacks": {
+          "type": "string",
+          "default": "",
+          "patternErrorMessage": "Please choose between 'Yes', 'No' and 'No, and never ask me again'",
```
shati-patel, line 353: Very minor: We don't provide a `pattern` property, so we won't ever see this `patternErrorMessage`! (Same with the above `codeQL.createQuery.folder` setting, though I didn't notice that one at the time 🙈)

(I can't link to the exact section, but for context, search for `patternErrorMessage` on https://code.visualstudio.com/api/references/contribution-points#Configuration-property-schema 🧩 )

extensions/ql-vscode/package.json
```diff
@@ -346,6 +346,12 @@
...
+          "markdownDescription": "The choice of whether to autogenerate QL packs for the current workspace. The choice is 'Yes', 'No' or 'No, and never ask me again'."
```
shati-patel, line 354: Just to check my understanding: is `codeQL.autogenerateQlPacks` ever "Yes" or "No"? 🤔 It looks like `codeQL.autogenerateQlPacks` is either not set, or it's equal to "No, and never ask me again". 

If that's correct, would this make more sense as a binary true/false setting? (e.g. `enableAutogeneratedQlPacks`: `true`/`false`)

extensions/ql-vscode/test/vscode-tests/no-workspace/helpers.test.ts
```diff
@@ -416,7 +417,7 @@ describe("helpers", () => {
     });
   });
 
-  describe("open dialog", () => {
+  describe("showBinaryChoiceDialog", () => {
```
shati-patel, line 420: Nice! Thanks for cleaning up the test structure 🧹

shati-patel (commented):
Looks good overall, thanks! 🍫 🥐 

I've just left some questions inline about the "format" of the config option ⚙

extensions/ql-vscode/package.json
```diff
@@ -346,6 +346,12 @@
           "default": "",
           "patternErrorMessage": "Please enter a valid folder",
           "markdownDescription": "The name of the folder where we want to create queries and query packs via the \"CodeQL: Create Query\" command. The folder should exist."
+        },
+        "codeQL.autogenerateQlPacks": {
```
robertbrignull, line 350: In some places (e.g. the `codeQL.createQuery.folder` setting description) we're using the term "query pack" and in some places "QL pack". Are they the same, or if so can we standardise? If not what's the difference and does this need to be communicated?

extensions/ql-vscode/src/config.ts
```diff
@@ -635,3 +635,16 @@ export function getSkeletonWizardFolder(): string | undefined {
 export async function setSkeletonWizardFolder(folder: string | undefined) {
   await SKELETON_WIZARD_FOLDER.updateValue(folder, ConfigurationTarget.Global);
 }
+
+/**
+ * Option to turn on/off ability to autogenerate QL packs. Values are "Yes" / "No" / "Never ask again"
```
robertbrignull, line 640: Unfortunately this comment of the expected values doesn't match the definition in [`package.json`](https://github.com/github/vscode-codeql/pull/2359/files#diff-be4c5a31505df7de81fa4900e84e66c04d1a6bca9b3da50f8061d2cb2bfbbde1R353)

Could we define a typescript type for it so we aren't dealing in raw strings? Either an enum or a string union would be great.

robertbrignull:
I really like that we're making this configurable. I'm sorry for being nit-picky about the definition of it, but since this is going to become user configuration, we want to be extra sure about it because we want to avoid changing it in the future.

extensions/ql-vscode/package.json
elenatanasoiu: They are the same, yup. The default way to call them is `QL pack` or `CodeQL pack`. However the command to create it is called `Create Query` because we don't want to user to know the details of how we create the query. This was an intentional design/product decision.

elenatanasoiu: I could rename both settings to be something like:
```
codeQL.createQuery.folder -> codeQL.createQuery.qlPackLocation
codeQL.autogenerateQlPacks -> codeQL.createQuery.autogenerateQLPacks
```

robertbrignull: Those setting names look good to be. The name "create query" is fine. I think it makes sense that the command is called that but then some of the settings for controlling it also contain references to "QL packs". But basically the name "query pack" should never appear anywhere.

elenatanasoiu: Makes sense. There are quite a few places where we use "query pack" and I'm a bit weary of renaming them in this PR. I've fixed all the cases related to this flow, so hopefully it keeps it consistent.

extensions/ql-vscode/package.json
elenatanasoiu: Good point. Two of the options are indeed redundant for this setting. If we choose:
- "yes" -> we won't store the config option
- "no" -> we won't store the config option
- "never" -> we store the config option

You could also argue that we only need one option here, but I'd like to keep it this way to mirror the options we present to the user.

extensions/ql-vscode/package.json
elenatanasoiu: Thanks for the docs and the explanations! I've updated this to use Robert's recommended structure. Let me know if it makes sense.

elenatanasoiu:
Thanks for the great explanations @shati-patel & @robertbrignull! 🙇‍♀️ 

I've had to rebase to account for file changes in the extension. 

Bar any CI woes, this is ready for a re-review.

extensions/ql-vscode/src/databases/local-databases.ts
```diff
@@ -742,11 +746,20 @@ export class DatabaseManager extends DisposableObject {
...
       `We've noticed you don't have a CodeQL pack available to analyze this database. Can we set up a query pack for you?`,
     );
 
-    if (!answer) {
+    if (answer === "No") {
+      return;
+    }
+
+    if (answer === "No, and never ask me again") {
+      await setAutogenerateQlPacks("No, and never ask me again");
```
shati-patel, line 762: As above! 

```suggestion
      await setAutogenerateQlPacks("never");
```

elenatanasoiu:
Thanks @shati-patel ❤️ I think it should be correct now.

robertbrignull (commented):
I've had another look through and on the second look I've spotted a few things that could use some more attention. I think this PR is definitely heading in the right direction, but we risk making more bugs for ourselves and a confusing interface for users if we try to go too fast.

extensions/ql-vscode/src/databases/local-databases.ts
```diff
@@ -742,11 +746,20 @@ export class DatabaseManager extends DisposableObject {
       return;
     }
 
-    const answer = await showBinaryChoiceDialog(
+    if (getAutogenerateQlPacks() === "never") {
+      return;
+    }
+
+    const answer = await showNeverAskAgainDialog(
```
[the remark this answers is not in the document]
elenatanasoiu: Actually when the answer is "yes", the intention is to keep asking the user in order to give them the option to turn off the behaviour eventually. I can show you a demo if it makes is easier to visualize.

When you're at the start of your CodeQL journey you'll likely choose "Yes" a few times until you get comfortable with writing your own queries. But once you get comfortable, you'll likely find this behaviour gets in the way and will want to turn it off. That's how the idea of having a "No, and never ask me again" option came about.

extensions/ql-vscode/package.json
```diff
@@ -341,11 +341,26 @@
...
           "patternErrorMessage": "Please enter a valid folder",
-          "markdownDescription": "The name of the folder where we want to create queries and query packs via the \"CodeQL: Create Query\" command. The folder should exist."
+          "markdownDescription": "The name of the folder where we want to create queries and QL packs via the \"CodeQL: Create Query\" command. The folder should exist."
+        },
+        "codeQL.createQuery.autogenerateQlPacks": {
+          "type": "string",
+          "default": "ask",
+          "enum": [
+            "ask",
+            "yes",
```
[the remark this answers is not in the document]
elenatanasoiu: > Sorry, but on second look at this, the opposite of "never" is "always". So the two options should either be "yes" and "no", or "always" and "never".

This comes back to my previous comment:

"Never" means "Never ask me again"
"No" means "Not this time, but keep asking"
"Yes" means "Yes, generate the QL pack and keep asking".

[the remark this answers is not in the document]
elenatanasoiu: Ok I've re-read the options. We don't have a "no" option anymore. I'm gonna have to go through this PR again and see if I can make it clearer. Sorry, it's been sitting around for a bit.

robertbrignull (approved):
One comment that needs updating after we changed the definition of things, but otherwise LGTM from just looking at the code.

I have tried to test this and I can't get the code to trigger, but I think it's because I'm not used to working with the codespace. Have you tested this, especially since we made a lot of modifications in review?

elenatanasoiu:
Here's a recording of the functionality:
https://github.com/github/vscode-codeql/assets/1354439/6ad19134-945a-4c80-9a17-e7155ead9762

PR #2365, opened 2023-04-20
Don't throw error if user cancels out of choosing a language

shati-patel:
Small tweak: When running "Create Query", if the user escapes out of the language quickpick we don't need to throw an error. Instead we can just show a [`UserCancellationException`](https://github.com/github/vscode-codeql/blob/dbdab13d03cd38032bf6e8f427bc4a2d92f7535c/extensions/ql-vscode/src/helpers.ts#L728-L730) notice.

## Checklist

N/A—this is still behind a feature-flag. 

### DIFF
--- a/extensions/ql-vscode/src/skeleton-query-wizard.ts
+++ b/extensions/ql-vscode/src/skeleton-query-wizard.ts
@@ -141,7 +141,7 @@ export class SkeletonQueryWizard {
       maxStep: 3,
     });
 
-    return await askForLanguage(this.cliServer, false);
+    return await askForLanguage(this.cliServer, true);
   }
 
   private async createQlPack() {


extensions/ql-vscode/src/skeleton-query-wizard.ts
```diff
@@ -141,7 +141,7 @@ export class SkeletonQueryWizard {
       maxStep: 3,
     });
 
-    return await askForLanguage(this.cliServer, false);
+    return await askForLanguage(this.cliServer, true);
```
robertbrignull, line 144: That `throwOnEmpty` argument is confusingly named, but that's not the fault of this PR 😅

robertbrignull (approved):
LGTM

I've tried it out and the experience is a lot nicer when you cancel the wizard. Thanks for spotting that and addressing it!

extensions/ql-vscode/src/skeleton-query-wizard.ts
shati-patel: I agree! To me, it's something more like "always throw, but less aggressively" 🙃 

I'll leave it as-is for now though.

PR #2369, opened 2023-04-24
Only attempt to discover QL tests in on-disk workspace folders

dbartol:
This was only a problem for the new test UI, because the third-party Test Explorer extension we used before must have had this filter in its implementation.


koesie10 (approved):
Thanks! Changes LGTM and the error seems to be resolved

PR #2373, opened 2023-04-25
Update wording in the "sorting dropdown" of the variant analysis results view

shati-patel:
"Sorting" dropdown in the variant analysis results view

Renames the following "sorting keys" in the results view:

- "Name" -> "Alphabetically"
- "Results" -> "Number of results"
- "Stars" -> "Popularity"
- "Last updated" -> "Most recent commit"
  - I've also updated the hover text for the "RepoRow" component here. This wasn't necessary for the other renamings.
    hover text for "last updated"

See linked internal issue for more details! 

## Checklist

This is still in beta, so I don't think we need a changelog update. The docs will require some [screenshot updates](https://codeql.github.com/docs/codeql-for-visual-studio-code/running-codeql-queries-at-scale-with-mrva/#exploring-your-results), but we'll save that until the follow-up tasks are also complete.

### DIFF
--- a/docs/test-plan.md
+++ b/docs/test-plan.md
@@ -259,7 +259,7 @@ This requires running a MRVA query and seeing the results view.
 7. Can open query text
 8. Can sort repos
    1. By name
-   2. By results
+   2. By number of results
    3. By popularity
    4. By most recent commit
 9. Can filter repos
--- a/docs/test-plan.md
+++ b/docs/test-plan.md
@@ -261,7 +261,7 @@ This requires running a MRVA query and seeing the results view.
    1. By name
    2. By results
    3. By stars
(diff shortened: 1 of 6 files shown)


robertbrignull (approved):
Changes LGTM

I am still wondering if we should change the internal names on  the`SortKey` enum too. The more I think about it the more I feel we should do, to keep things consistent between internal and external names. What do you think?

shati-patel:
> I am still wondering if we should change the internal names on the `SortKey` enum too. The more I think about it the more I feel we should do, to keep things consistent between internal and external names. What do you think?

Thanks! That sounds like it'll be less confusing long-term 👍🏽 

I'll open a follow-up PR!

PR #2384, opened 2023-04-27
Fix not setting database as selected if it already exists

robertbrignull:
This PR should fix a bug where we aren't able to set a database as the selected database. There are two bugs which are each independently causing this behaviour:
- In `addExistingDatabaseItem` if we find that a database item with that URI exists already then we return immediately, without checking `makeSelected` and therefore without setting the database as selected.
- In `handleSetCurrentDatabase` and `chooseAndSetDatabase` we weren't setting `makeSelected` to true.

This PR hopefully addresses both points. We now check `makeSelected` in both branches in `addExistingDatabaseItem`. We also change the default value of `makeSelected` to true, because we believe this is always the desired behaviour, except in one case in the test runner.

I've structured the commits so that https://github.com/github/vscode-codeql/commit/92f23d18fb94cb3f896c9601f1b5d39a58b318ce where we change the default value should be 100% behaviour preserving. I hope that makes it easier to review, but if not then just ignore the commits and review the PR as a whole.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Fix bug where the "CodeQL: Set Current Database" command didn't always select the database. [#2384](https://github.com/github/vscode-codeql/pull/2384)
+
 ## 1.8.3 - 26 April 2023
 
 - Added ability to filter repositories for a variant analysis to only those that have results [#2343](https://github.com/github/vscode-codeql/pull/2343)
--- a/extensions/ql-vscode/src/databases/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/databases/local-databases-ui.ts
@@ -636,13 +636,7 @@ export class DatabaseUI extends DisposableObject {
               this.queryServer?.cliServer,
             );
           } else {
-            const makeSelected = true;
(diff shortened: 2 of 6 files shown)


shati-patel (approved):
Thanks for fixing this! 🕵🏽‍♂️ I've double-checked the behaviour locally and it looks good 🎉

Since it's a user-visible bug fix, this probably deserves a note in the changelog. (Something like "Fix bug where the _CodeQL: Set Current Database_ command didn't always select the database"? 🔍)

shati-patel:
The test failure looks legitimate too: we now call `resolveDatabaseContents` twice: once in [`createDatabaseItem`](https://github.com/github/vscode-codeql/blob/92f23d18fb94cb3f896c9601f1b5d39a58b318ce/extensions/ql-vscode/src/databases/local-databases.ts#L677), and once in [`setCurrentDatabaseItem`](https://github.com/github/vscode-codeql/blob/92f23d18fb94cb3f896c9601f1b5d39a58b318ce/extensions/ql-vscode/src/databases/local-databases.ts#L954) (as part of [`item.refresh`](https://github.com/github/vscode-codeql/blob/92f23d18fb94cb3f896c9601f1b5d39a58b318ce/extensions/ql-vscode/src/databases/local-databases.ts#L387)).

robertbrignull:
I've also managed to reproduce the bug locally and verified the fix

PR #2392, opened 2023-04-28
Add config options for default sorting/filtering values in variant analysis results view

shati-patel:
Adds new settings for configuring the default display of variant analysis results:

image

This also changes the default display to show repos with results (instead of all), sorted by number of results (instead of by name).

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -329,6 +329,36 @@
           "patternErrorMessage": "Please enter a valid GitHub repository",
           "markdownDescription": "[For internal use only] The name of the GitHub repository in which the GitHub Actions workflow is run when using the \"Run Variant Analysis\" command. The repository should be of the form `<owner>/<repo>`)."
         },
+        "codeQL.variantAnalysis.filterResults": {
+          "type": "string",
+          "default": "withResults",
+          "enum": [
+            "all",
+            "withResults"
+          ],
+          "enumDescriptions": [
+            "Show all repositories in the results view.",
+            "Show only repositories with results in the results view."
+          ],
+          "description": "The filter to apply to the variant analysis results view."
+        },
+        "codeQL.variantAnalysis.sortResults": {
+          "type": "string",
+          "default": "numberOfResults",
+          "enum": [
+            "alphabetically",
+            "popularity",
+            "mostRecentCommit",
+            "numberOfResults"
+          ],
+          "enumDescriptions": [
+            "Sort repositories alphabetically in the results view.",
+            "Sort repositories by popularity in the results view.",
+            "Sort repositories by most recent commit in the results view.",
+            "Sort repositories by number of results in the results view."
+          ],
+          "description": "The default sorting order for repositories in the variant analysis results view."
+        },
         "codeQL.logInsights.joinOrderWarningThreshold": {
           "type": "number",
           "default": 50,
--- a/extensions/ql-vscode/src/pure/interface-types.ts
+++ b/extensions/ql-vscode/src/pure/interface-types.ts
@@ -11,7 +11,10 @@ import {
   VariantAnalysisScannedRepositoryResult,
   VariantAnalysisScannedRepositoryState,
 } from "../variant-analysis/shared/variant-analysis";
-import { RepositoriesFilterSortStateWithIds } from "./variant-analysis-filter-sort";
+import {
+  RepositoriesFilterSortState,
+  RepositoriesFilterSortStateWithIds,
+} from "./variant-analysis-filter-sort";
 import { ErrorLike } from "./errors";
 import { DataFlowPaths } from "../variant-analysis/shared/data-flow-paths";
 import { ExternalApiUsage } from "../data-extensions-editor/external-api-usage";
@@ -405,6 +408,7 @@ export interface ParsedResultSets {
 export interface SetVariantAnalysisMessage {
   t: "setVariantAnalysis";
   variantAnalysis: VariantAnalysis;
+  filterSortState?: RepositoriesFilterSortState;
 }
 
 export type VariantAnalysisState = {
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [UNRELEASED]
 
+- Add settings `codeQL.variantAnalysis.filterResults` and `codeQL.variantAnalysis.sortResults` for configuring how variant analysis results are filtered and sorted in the results view. The default is to show only repositories with results, and to sort by the number of results. [#2392](https://github.com/github/vscode-codeql/pull/2392)
 - On the variant analysis results page, show the count of successful analyses instead of completed analyses, and indicate the reason why analyses were not successful. [#2349](https://github.com/github/vscode-codeql/pull/2349)
 - Fix bug where the "CodeQL: Set Current Database" command didn't always select the database. [#2384](https://github.com/github/vscode-codeql/pull/2384)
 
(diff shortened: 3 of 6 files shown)


extensions/ql-vscode/src/view/variant-analysis/__tests__/VariantAnalysis.spec.tsx
```diff
@@ -46,4 +65,33 @@ describe(VariantAnalysis.name, () => {
...
+
+  it("renders results view with correct filter and sort state", async () => {
+    const variantAnalysis = createMockVariantAnalysis({});
+    render({ variantAnalysis });
+
+    // Without this await, `getByDisplayValue` could not find any selected dropdown option.
+    await new Promise((resolve) => setTimeout(resolve, 0));
+
+    expect(screen.getByDisplayValue("With results")).toBeInTheDocument();
+    expect(screen.getByDisplayValue("Number of results")).toBeInTheDocument();
```
shati-patel, line 77: For some reason, `await new Promise((resolve) => setTimeout(resolve, 0));` was necessary to get these "default" expectations to pass 🤔 

Not sure why, but it works 😅 If anyone knows more about this, let us know!

extensions/ql-vscode/package.json
```diff
@@ -329,6 +329,36 @@
           "patternErrorMessage": "Please enter a valid GitHub repository",
           "markdownDescription": "[For internal use only] The name of the GitHub repository in which the GitHub Actions workflow is run when using the \"Run Variant Analysis\" command. The repository should be of the form `<owner>/<repo>`)."
         },
+        "codeQL.variantAnalysis.filterResults": {
```
koesie10, line 332: Should we call these `defaultFilterResults` and `defaultSortResults` to make it clearer that these options can still be changed?

extensions/ql-vscode/package.json
```diff
@@ -329,6 +329,36 @@
...
+          "default": "withResults",
+          "enum": [
+            "all",
+            "withResults"
+          ],
+          "enumDescriptions": [
+            "Show all repositories in the results view.",
+            "Show only repositories with results in the results view."
+          ],
+          "description": "The filter to apply to the variant analysis results view."
```
koesie10, line 343:
```suggestion
          "description": "The default filter to apply to the variant analysis results view."
```

extensions/ql-vscode/src/view/variant-analysis/__tests__/VariantAnalysis.spec.tsx
koesie10: I'm not sure why this is happening (it could be that some of the VS Code UI toolkit shadow DOM is only rendered on the next tick or something like that), but a slightly more reliable way to solve this is by waiting for the element to start existing:

```ts
await waitFor(() => screen.getByText("With results"));
```

This way we don't depend on the exact tick and it will still fail if it doesn't find the element after some time.

extensions/ql-vscode/src/view/variant-analysis/__tests__/VariantAnalysis.spec.tsx
```diff
@@ -16,6 +18,23 @@ describe(VariantAnalysis.name, () => {
...
+          source: window,
+          origin: window.location.origin,
+          data: msg,
+        }),
+      );
+
+      // The event is dispatched asynchronously, so we need to wait for it to be handled.
+      await new Promise((resolve) => setTimeout(resolve, 0));
+    });
+  };
```
koesie10, line 36: Now that we're using this in at least two tests, do you think it would make sense to move this to a separate file and export the function?

shati-patel: Good idea! I wasn't sure where best to put the shared file. Currently in `src/view/common/post-message.ts` but I'm open to better suggestions 😅

shati-patel:
Thank you for the helpful comments, @koesie10 😍

extensions/ql-vscode/src/view/variant-analysis/__tests__/VariantAnalysis.spec.tsx
shati-patel: This is perfect! Thanks ✨

robertbrignull:
When pairing on this on Friday, we decided to add the sort/filter information to the `setVariantAnalysis` message. This was for ease and because we're always already sending one of those messages at the time we want the sort/filter information.

Over the weekend I've thought more and I think it's still unclear what the best approach is, but I'm leaning more in favour of having a new message type and just sending two separate messages one after the other. That way the naming of the messages is more clear, and we don't have the sort/filter state being optional in the message.

What do you think?

These details of the webview messages are purely internal and there's no backwards compatibility concerns, so we could also do this as a followup PR after this one.

shati-patel:
> When pairing on this on Friday, we decided to add the sort/filter information to the `setVariantAnalysis` message. This was for ease and because we're always already sending one of those messages at the time we want the sort/filter information.
> 
> Over the weekend I've thought more and I think it's still unclear what the best approach is, but I'm leaning more in favour of having a new message type and just sending two separate messages one after the other. That way the naming of the messages is more clear, and we don't have the sort/filter state being optional in the message.
> 
> What do you think?
> 
> These details of the webview messages are purely internal and there's no backwards compatibility concerns, so we could also do this as a followup PR after this one.

Thanks, @robertbrignull! That seems like a reasonable suggestion. Doesn't sounds too big, so I'll change that in this PR 👍🏽

robertbrignull (approved):
LGTM from my point of view. @koesie10 do you want to take another look / have any comments or are you happy?

shati-patel:
Thanks for the reviews! I'll merge/fix conflicts once the 1.8.4 release is complete 🚀

extensions/ql-vscode/src/pure/interface-types.ts
```diff
@@ -407,6 +410,11 @@ export interface SetVariantAnalysisMessage {
   variantAnalysis: VariantAnalysis;
 }
 
+export interface setFilterSortStateMessage {
```
koesie10, line 413:
```suggestion
export interface SetFilterSortStateMessage {
```

extensions/ql-vscode/src/pure/interface-types.ts
```diff
@@ -459,6 +467,7 @@ export interface ShowDataFlowPathsMessage {
 
 export type ToVariantAnalysisMessage =
   | SetVariantAnalysisMessage
+  | setFilterSortStateMessage
```
koesie10, line 470:
```suggestion
  | SetFilterSortStateMessage
```

koesie10 (approved):
LGTM, I just found one minor inconsistency

shati-patel:
> LGTM, I just found one minor inconsistency

Woops! Thanks, @koesie10. Fixed.

shati-patel:
Slight niggle (just discussed with @robertbrignull offline) is that I haven't yet implemented showing "in progress" repos. 

If we only display `With results` by default, then the "in progress" repos are hidden from view when you kick off a variant analysis so it's harder to tell what is being analyzed. If no repos return results, then the view remains empty.

Instead of fixing that in this PR, I'm going to change the default back to `All` here and merge. Then we can think a bit more about the "in progress" filtering in a follow-up.

shati-patel:
Made `All` the default again in https://github.com/github/vscode-codeql/pull/2392/commits/006dd7ea891acc09b25553fdb787d557a8859d25. There are other changes, like the `permissiveFilterSortState` in tests, which are no longer strictly necessary. But I'll keep that in, since I think it makes sense to decouple the default filter sort state from those tests.

robertbrignull (approved):
Yep, this matches what I thought we discussed.

I also tried running it and it works well. I'd be happy to ship this.

PR #2395, opened 2023-05-02
Prefer `#select` result set over other result sets

koesie10:
For `@kind problem` queries, it seems like the BQRS contains four result sets: `edges`, `nodes`, `subpaths`, and `#select`. The user is probably interested in the results of `#select` since that contains the direct result of the query. This changes the extraction of raw (BQRS) results to always prefer `#select` over any other result set. If it can't find a result set named `#select`, it will fall back to the first result set in the BQRS like before.

## Checklist


charisk (approved):
Seems reasonable to me! Thanks for fixing.

PR #2405, opened 2023-05-07
Simplify location URLs if start and end line are the same

Marcono1234:
When the start and end line of a result are the same, currently the extension creates location URLs which specify both line numbers, e.g. `...MyFile.js#L5-L5`. This is appears to be redundant and can be simplified to just `#L5`.

I am not very familiar with this project, so any feedback is appreciated!

## Checklist


aeisenberg:
This looks great. Thank you for your contribution. One minor thing I would like to see is that with this change is that the test is no longer showing any multiline results (arguably this was a limitation of the test before, but there was only a single code path earlier). But, could you add at least one multiline result in one of the test files, like `result-1-github-codeql.md`? You'd also need to update the corresponding test input in `analyses-results.json`. I think this is straight forward, but if not, I can take a look.

aeisenberg:
Failing test is likely a flake.

Marcono1234:
> One minor thing I would like to see is that with this change is that the test is no longer showing any multiline results

I am not completely sure what you mean. If you had written "still shows multiline results", then I would have assumed you mean that something like `#L1-L5` is unaffected, and that actually appears to be the case, see for example:
https://github.com/github/vscode-codeql/blob/788eb2b663d22fce2048319eb8717b9f3922a528/extensions/ql-vscode/test/unit-tests/data/markdown-generation/interpreted-results/path-problem/expected/result-2-meteor-meteor.md?plain=1#L51
https://github.com/github/vscode-codeql/blob/788eb2b663d22fce2048319eb8717b9f3922a528/extensions/ql-vscode/test/unit-tests/data/markdown-generation/interpreted-results/problem/expected/result-1-github-codeql.md?plain=1#L3

So, what did you mean by that?

aeisenberg:
Apologies for that. You are correct. I missed those extra tests. Thanks for pointing them out. This PR looks great.

aeisenberg (approved):

PR #2407, opened 2023-05-08
Use `QueryLanguage` type for representing query languages

koesie10:
This will make both the `askForLanguage` and `findLanguage` functions return a `QueryLanguage` instead of a `string`. This will make it harder to make mistakes when using these functions.

There are also some other changes with regards to `QueryLanguage` such that we never need to use `as QueryLanguage` explicitly anymore, except for the new `isQueryLanguage` function. The only remaining place that I know of where we're using a `string` to represent the `QueryLanguage` is in a database item's language, but this is harder to change and may be relied upon by language authors.

## Checklist


extensions/ql-vscode/src/helpers.ts
```diff
@@ -775,7 +786,16 @@ export async function askForLanguage(
         "Language not found. Language must be specified manually.",
       );
     }
+    return undefined;
   }
+
+  if (!isQueryLanguage(language)) {
+    void showAndLogErrorMessage(
+      `Language '${language}' is not supported. Language must be specified manually.`,
```
charisk, line 794: Should we be showing the list of supported languages instead of saying `Language must be specified manually`?

charisk (commented):
The changes generally look okay to me but I'd rather we get a review from someone like @aeisenberg too since it changes the behaviour around local databases a bit.

extensions/ql-vscode/src/helpers.ts
adityasharad: Does this PR make the extension's validation stricter for local queries, and only allow query runs on languages that are listed in the extension code?

koesie10: It makes validation stricter in some cases, but I believe normal local queries for unsupported languages should still work. What won't work is:
- Creating a skeleton query pack
- Running a MRVA query
- Running a query on multiple databases

adityasharad: That sounds fine. We may want to weaken it in future as we expand to more languages, but for now I think that's an acceptable set.

charisk:
@koesie10 if you update this I can re-review.

charisk (approved):
Sorry about the wait, I was hoping to get more 👀 on it.

PR #2410, opened 2023-05-08
Fix hang/crash when using `--check-databases` option for test run

dbartol:
Trivial fix is to ignore events for tests we didn't explicitly request in the run.

I also discovered that sometimes the CLI emits an empty null-terminated string as an event. We should just ignore those.


extensions/ql-vscode/src/query-testing/test-manager.ts
```diff
@@ -275,68 +275,66 @@ export class TestManager extends TestManagerBase {
...
-              testMessages.push(
-                TestMessage.diff(
-                  "Actual output differs from expected",
-                  expected,
-                  actual,
-                ),
-              );
+        // If the user manually adds the `--check-databases` option, we can get events for the
+        // corresponding database check tests. Just ignore these for now.
+        if (testItem !== undefined) {
```
aeisenberg, line 280: I think this is fine, but maybe can you add a log message for the unexpected `event.test`?

PR #2416, opened 2023-05-11
Resolve queries instead of hardcoding the path

koesie10:
This will change the data extensions editor generator to resolve the queries based on the `modelgenerator` tag. This removes the requirement of having a `ql` folder in the workspace.

I chose to use the path instead of the `id` for now to avoid having to resolve the query 4 times. This also avoids the need to map the language names to the language ID in the tag (i.e. `csharp` -> `cs`).

## Checklist


extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
```diff
@@ -16,44 +16,83 @@ import {
...
   queryRunner: QueryRunner;
   queryStorageDir: string;
-  qlDir: string;
   databaseItem: DatabaseItem;
   progress: ProgressCallback;
   token: CancellationToken;
   onResults: (results: ModeledMethodWithSignature[]) => void | Promise<void>;
 };
 
+export async function resolveQueries(
```
charisk, line 34: This function doesn't need exporting.

charisk (approved):
The code looks good to me but would be good to get 👀 from @starcke on the approach.

extensions/ql-vscode/src/data-extensions-editor/generate-flow-model.ts
```diff
@@ -16,45 +16,84 @@ import {
...
+      postfix: ".qls",
+    })
+  ).path;
+  const suiteYaml = [];
+  for (const qlpack of packsToSearch) {
+    suiteYaml.push({
+      from: qlpack,
+      queries: ".",
+      include: {
+        "tags contain": "modelgenerator",
```
starcke, line 59: I wonder how we should ensure that the tag of these are not changed (and that not too many other queries get into this tag). Good enough for now, but lets put a note in the missing functionality doc: https://docs.google.com/document/d/1OoCEtUcShiQG328sPIoS0wNR2pqIg4Umy79UEDiSqLk/edit#heading=h.9b4gwygshmfg

starcke (approved):
I think this is ok, and not worse than what we were doing before. Later we should establish some agreement with the maintainers of those queries.

PR #2417, opened 2023-05-11
Add support for quick eval count to the query runner

alexet:
This only adds support internally to the query runner, without any UI support. I will add basic support (command palette only) in a separate PR.

I'm not completely happy about the api as we switch between an enum and booleans. The issue is that before running for quick eval we simply have a boolean and afterwards we have an optional position. It should then really become a optional (position and count enabled) but in a bunch of places we need backwards compatibility as we persist it as json. So we end up with an invalid state (no quick eval position, but quickEvalCountOnly  is set), and a difference between the initial enum and the later boolean state. However I couldn't think of much better as I don't want to add yet more boolean arguments to the initial query.

## Checklist

  - No user visible changes in this PR
  -  No user visible changes in this PR
  -  No user visible changes in this PR

### DIFF
--- a/extensions/ql-vscode/src/pure/messages-shared.ts
+++ b/extensions/ql-vscode/src/pure/messages-shared.ts
@@ -68,6 +68,14 @@ export interface CompilationTarget {
  */
 export interface QuickEvalOptions {
   quickEvalPos?: Position;
+  /**
+   * Whether to only count the number of results.
+   *
+   * This is only supported by the new query server
+   * but it isn't worth having a separate type and
+   * it is fine to have an ignored optional field.
+   */
+  countOnly?: boolean;
 }
 
 /**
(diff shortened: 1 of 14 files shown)


extensions/ql-vscode/src/pure/messages-shared.ts
```diff
@@ -68,6 +68,14 @@ export interface CompilationTarget {
  */
 export interface QuickEvalOptions {
   quickEvalPos?: Position;
+  /**
+   * Whether to only count the number of results.
+   *
+   * This is only supported by the new query server
+   * but it isn't worth having a separate type and
+   * it is fine to have an ignored optional field.
```
aeisenberg, line 76: 👍🏼

aeisenberg (approved):
I don't fully understand the context of this change, but it looks generally safe and fine.

PR #2418, opened 2023-05-12
Create a basic tree view for the queries panel

shati-patel:
A _very_ minimal outline of what the queries panel treeview structure will look like, roughly based on the [variant analysis db panel](https://github.com/github/vscode-codeql/tree/main/extensions/ql-vscode/src/databases/ui). See internal linked issue for more info!

@charisk, I've only added a single "child node" so far, since I couldn't figure out how to handle the "parent-child" relationships yet 😅 Happy to try and poke around with that here, or in a follow-up! Also: is there anything here worth testing at the moment?

image

## Checklist

N/A—internal only


extensions/ql-vscode/src/queries-panel/query-node.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * A node in the tree of queries. This will be either a `QueryDirectory` or a `QueryFile`.
+ */
+export abstract class QueryNode {
```
charisk, line 4: What do you think about leaving these classes out for now and not including the query node as part of the tree view item? I'm not convinced we'll need it.

extensions/ql-vscode/src/extension.ts
```diff
@@ -732,6 +733,11 @@ async function activateWithInstalledDistribution(
   );
   ctx.subscriptions.push(databaseUI);
 
+  void extLogger.log("Initializing queries panel.");
```
charisk, line 736: Shall we check whether the relevant feature flag is on before doing the initialization stuff? If we do that there's less of a chance of breaking things while we're in a WIP state, and we avoid logging messages that could confuse the user.

Also, and this can be part of a future PR, we've started using the idea of a "module" (e.g. [`DbModule`](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/databases/db-module.ts)) that deals with initialization etc. and keeps extension.ts changes to a minimum. We could apply the same pattern here.

extensions/ql-vscode/src/queries-panel/query-node.ts
shati-patel: Yes, good point. Removed that in https://github.com/github/vscode-codeql/pull/2418/commits/61d4c9c695f603d43ef0ab73871b0496fbb8a1e4 💥

extensions/ql-vscode/src/extension.ts
shati-patel: Thanks! I had a go at this in https://github.com/github/vscode-codeql/pull/2418/commits/495b8fc960850fbde026640719c81f826cd67659! (based on the DbModule code)
Let me know if that's the right idea 🙇🏽‍♀️

extensions/ql-vscode/src/queries-panel/queries-module.ts
```diff
@@ -0,0 +1,30 @@
...
+import { QueriesPanel } from "./queries-panel";
+
+export class QueriesModule extends DisposableObject {
+  private queriesPanel: QueriesPanel | undefined;
+
+  private constructor(readonly app: App) {
+    super();
+  }
+
+  private async initialize(app: App): Promise<void> {
```
charisk, line 13: I think this and the static `initialize` func don't need to be async/return Promise.

charisk (approved):
LGTM! Just some minor comments.

Thanks for adding the `QueriesModule`!

PR #2421, opened 2023-05-15
Add provenance support to data extensions editor

koesie10:
The data extensions editor was always setting the `provenance` field of MaD to `manual`. This will change the `provenance` to be either `manual` (unchanged, for models which were added by the user), `df-generated` (for models generated by the dataflow generator), and `df-manual` (for models generated and then edited). This makes it easier to trace the origin of a model.

## Checklist


adityasharad:
It is worth looking at the parts of the CodeQL Java libraries that validate the provenance values. As I understand it, there is a fixed list of acceptable provenances, which you would need to update in collaboration with the Java team, otherwise either the analysis could give no results or the consistency checks will fail on the newly-generated models.

aeisenberg:
> you would need to update in collaboration with the Java team

A reasonable alternative would be to use a contextual query to get this from the libraries directly.

koesie10:
> It is worth looking at the parts of the CodeQL Java libraries that validate the provenance values. As I understand it, there is a fixed list of acceptable provenances, which you would need to update in collaboration with the Java team, otherwise either the analysis could give no results or the consistency checks will fail on the newly-generated models.

Good call, I had checked that the provenance is formatted as `origin-verification`, but hadn't checked whether there was a predefined list of `origin` values. It seems [there is](https://github.com/github/codeql/blob/ac1df4de91438904c86e6e3ce5d01f9e6ba2468a/java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImpl.qll#L221-L229), so I've reverted the default back to `manual` rather than `editor-manual`. The `df-manual` and `df-generated` values are already used in the libraries and `df` is defined as a valid origin, so these do not need any changes.

> A reasonable alternative would be to use a contextual query to get this from the libraries directly.

I think for now that's not something we need to do. The editor is now automatically determining which provenance to use and I don't think a contextual query to get all valid ones would help with that.

charisk (approved):

PR #2422, opened 2023-05-15
Allow compare view to work with quick-eval

aeisenberg:
The compare view typically works by matching the results sets of queries. It only allows the results sets of queries with identical names to be compared.

Manually run queries use `#select` as the default result set. However, quick eval queries use a different, generated, name. Therefore, these two kinds of queries cannot be compared.

This commit changes the heuristics so that if there are no identical;y named results sets, the first result set of each query that is prefixed with `#` is used (this is the default result set).

It also makes a slightly nicer error message if there are no comparable results sets.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/compare/compare-view.ts
+++ b/extensions/ql-vscode/src/compare/compare-view.ts
@@ -175,21 +175,40 @@ export class CompareView extends AbstractWebview<
     const commonResultSetNames = fromSchemaNames.filter((name) =>
       toSchemaNames.includes(name),
     );
+
+    // Fall back on the default result set names if there are no common ones.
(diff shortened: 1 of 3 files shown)


charisk (approved):
I don't have a huge amount of context on the area but the code changes look sensible to me.

aeisenberg:
Thanks for the review. This area is not tested very well, but it's also not used very much. I think these changes are low risk.

charisk:
Oh sorry I only just remembered, should we include anything in the CHANGELOG?

aeisenberg:
Right...yes. Forgot that. Thanks for reminding. I'll put something up.

PR #2423, opened 2023-05-16
Simplify types around query panel

robertbrignull:
A couple of small changes and simplifications:
- We copied some method types from the `TreeDataProvider` class but they're more general than we need them to be. We don't need to say that we may return a promise or may not, because we know what we'll return. It still satisfies the type from the inherited class so long as it's more specific and can be cast to the other type.
- In `QueryTreeViewItem` we extend from `TreeItem` and that class already defines a bunch of fields, and it's probably best to avoid accidentally re-defining the field with `public readonly`. Instead we can do things like `this.tooltip = tooltip` to set the field in the constructor.

## Checklist


charisk (approved):
Apologies for missing those during review.

shati-patel (approved):
Thanks! 🧹

PR #2426, opened 2023-05-17
Open query file on click

robertbrignull:
Implements opening a file on click from the queries panel. Also adjusts how we construct `QueryTreeViewItem` a little bit and provides some different dummy data. Instead of passing in all bits separately, just pass in a file path and work out the label and tooltip from that.

I'm thinking that in the future we'll want to construct these from a file path, but if that's not the case just say and I can adjust or remove that change from this PR. I think we will need to know the file path to be able to open the file when clicked.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/queries-panel/query-tree-data-provider.ts
+++ b/extensions/ql-vscode/src/queries-panel/query-tree-data-provider.ts
@@ -17,11 +17,18 @@ export class QueryTreeDataProvider
   private createTree(): QueryTreeViewItem[] {
     // Temporary mock data, just to populate the tree view.
     return [
-      new QueryTreeViewItem("name1", "path1", []),
-      new QueryTreeViewItem("name2", "path2", [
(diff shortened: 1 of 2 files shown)


shati-patel (approved):
Looks good! (Admittedly we paired on it, so happy for other follow-ups) 

Note from testing: The example data doesn't work on Windows since it tries to open/create a file at (e.g.) `\ql\javascript\example.ql` instead of at the relative path.
We've decided this is fine for now, since we'll be replacing the test data with real paths soon.

koesie10 (approved):
LGTM as well. In the future, do we want to track when an item like this is clicked by creating a wrapper around the `vscode.open` command so we can see clicks in telemetry?

robertbrignull:
> do we want to track when an item like this is clicked by creating a wrapper around the `vscode.open` command so we can see clicks in telemetry?

That's a good idea. I suggest we do this as a followup PR, but it would be a good thing to do.

PR #2428, opened 2023-05-17
Add warning when using unsupported CLI version

koesie10:
This adds a warning when the user is using an unsupported version of the CodeQL CLI. The warning is shown once per session, and only if the version is older than the oldest supported version.

Screenshot 2023-05-17 at 16 26 19

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -1737,6 +1737,10 @@ export function shouldDebugCliServer() {
 }
 
 export class CliVersionConstraint {
+  // The oldest version of the CLI that we support. This is used to determine
+  // whether to show a warning about the CLI being too old on startup.
+  public static OLDEST_SUPPORTED_CLI_VERSION = new SemVer("2.7.6");
+
   /**
    * CLI version where building QLX packs for remote queries is supported.
    * (The options were _accepted_ by a few earlier versions, but only from
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -24,7 +24,7 @@ import {
   activate as archiveFilesystemProvider_activate,
   zipArchiveScheme,
 } from "./common/vscode/archive-filesystem-provider";
-import { CodeQLCliServer } from "./codeql-cli/cli";
+import { CliVersionConstraint, CodeQLCliServer } from "./codeql-cli/cli";
 import {
   CliConfigListener,
   DistributionConfigListener,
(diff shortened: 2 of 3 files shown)


extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1737,6 +1737,10 @@ export function shouldDebugCliServer() {
 }
 
 export class CliVersionConstraint {
+  // The oldest version of the CLI that we support. This is used to determine
+  // whether to show a warning about the CLI being too old on startup.
+  public static OLDEST_SUPPORTED_CLI_VERSION = new SemVer("2.7.6");
```
shati-patel, line 1742: Is it possible to get this value automatically from https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/supported_cli_versions.json, instead of having to manually update it? (No problem if that isn't feasible!)

shati-patel (commented):
Looks good! 🎉 One question inline, and could you also add a changelog entry?

extensions/ql-vscode/src/codeql-cli/cli.ts
aeisenberg: We're working on finalizing our deprecation strategy and it's probably going to be the case that we only need to support 1 year's worth of old CLIs. Given that we do a minor version bump every 3 months (coinciding with GHES releases), we only need to be supporting v2.10.x and later.

That being said, I don't think this is something that should be addressed in this PR. It's more of an FYI.

Once this is finalized, then I agree that we should have some processes around deprecations.

extensions/ql-vscode/src/extension.ts
```diff
@@ -408,6 +408,28 @@ export async function activate(
...
+      }
+
+      if (CliVersionConstraint.OLDEST_SUPPORTED_CLI_VERSION.compare(ver) < 0) {
+        return;
+      }
+
+      void showAndLogWarningMessage(
+        `You are using an unsupported version of the CodeQL CLI (${ver}). ` +
+          `The minimum supported version is ${CliVersionConstraint.OLDEST_SUPPORTED_CLI_VERSION}. ` +
+          `Please upgrade to the latest version of the CodeQL CLI.`,
```
aeisenberg, line 429:
```suggestion
          `Please upgrade to a newer version of the CodeQL CLI.`,
```

aeisenberg (commented):
Thanks. This is great.

aeisenberg:
I was mistaken about the supported versions. We need to go back to 2.8.x (ie- 2.7.x can be removed). [See the docs](https://docs.github.com/en/enterprise-server@3.8/admin/all-releases) for supported GHES versions. 

When GHES 3.9 is released (shortly), we can drop support for 2.8.x.

PR #2431, opened 2023-05-19
Add workspace folders to App interface

robertbrignull:
Adds knowledge of workspace folders to the `App` interface. This is going to be used very soon by the queries panel, and adding this to the `App` means that the queries panel itself will depend less on VS Code internals and will be easier to test.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/app.ts
+++ b/extensions/ql-vscode/src/common/app.ts
@@ -4,6 +4,11 @@ import { AppEventEmitter } from "./events";
 import { Logger } from "./logging";
 import { Memento } from "./memento";
 import { AppCommandManager } from "./commands";
+import type {
+  WorkspaceFolder,
+  Event,
+  WorkspaceFoldersChangeEvent,
+} from "vscode";
 
 export interface App {
   createEventEmitter<T>(): AppEventEmitter<T>;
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/common/app.ts
```diff
@@ -4,6 +4,11 @@ import { AppEventEmitter } from "./events";
 import { Logger } from "./logging";
 import { Memento } from "./memento";
 import { AppCommandManager } from "./commands";
+import type {
```
charisk, line 7: Since this is not a massively well known feature of TS, it might be worth putting a comment to explain what is going on. What do you think? I'm happy either way.

charisk (approved):
LGTM! Thanks for the clean PR.

charisk (approved):
LGTM! Thanks for the clean PR.

extensions/ql-vscode/src/common/app.ts
robertbrignull: You're right it's not a hugely well known feature of TypeScript. I didn't know about it until about 6 months ago. But I don't think commenting in the code is a helpful way of teaching people, because they may not see this particular copy and I don't want to include a comment on every single usage. There are 32 other examples already and my hope would be that we add more because it's a cool and useful feature. So I hope that over time the need for explanation will be minimal as everyone will see this feature enough to know what it does.

PR #2433, opened 2023-05-19
Hook queries panel up to real data

robertbrignull:
This PR implements supplying the queries panel with real data from the workspace. We make use of the `Discovery` class, which is already used by the QL Test panel feature and has a very similar set of requirements to us. A lot of inspiration has come directly from the `QLTestDiscovery` class.

Tested locally using the codeql starter workspace:
Screenshot 2023-05-19 at 16 40 08

It successfully handles updating when there's changes to the filesystem, and the automatic remembering of which items are expanded or collapsed seems pretty good.

There are some rough edges that can be improved but I think this is a good starting point. The main thing I'd like to look at in future PRs is single vs multi-root workspaces. In a single-root workspace maybe we hide the top-level item which is just the workspace name.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -135,6 +135,11 @@ export interface SourceInfo {
 }
 
 /**
+ * The expected output of `codeql resolve queries`.
+ */
+export type ResolvedQueries = string[];
+
+/**
  * The expected output of `codeql resolve tests`.
  */
 export type ResolvedTests = string[];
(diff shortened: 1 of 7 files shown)


shati-patel (approved):
Nice! Approving, since it works locally and it's still hidden behind a feature-flag (but also happy if you'd prefer to wait for another review before merging ❤️)

As discussed offline, there are a few specific things we'd like to fix, but we can do these in follow-up PRs 🧹 
1. Hide the extra folders called `.` from the tree view.
2. Remove the spammy `resolve queries` logs from the CodeQL Extension Log (they currently appear whenever any `.ql` file or containing folder changes).
3. Clean up VS Code imports (e.g. decide whether to import more things from `App`.)
   - This should help us write unit tests.

robertbrignull:
Personally I'm happy to merge now since it's fully hidden and we're still developing and iterating.

PR #2434, opened 2023-05-22
Mark AppEventEmitter as disposable

robertbrignull:
Makes `AppEventEmitter` implement `Disposable`. This class is mirroring the VS Code `EventEmitter` class, and that does have a `dispose` method.

This means in places where we use `app.createEventEmitter` we can track the object and dispose of it, which helps mirror the places where we use `EventEmitter` directly and we already do this.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/events.ts
+++ b/extensions/ql-vscode/src/common/events.ts
@@ -4,7 +4,7 @@ export interface AppEvent<T> {
   (listener: (event: T) => void): Disposable;
 }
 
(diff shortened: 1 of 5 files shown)


koesie10 (approved):

robertbrignull:
Thanks for reviewing. I just realised I forgot to also make sure that `DbManager` is disposed of, so I've just pushed that. I think that change is small enough and in keeping with the rest of the PR so it doesn't invalidate the approval.

PR #2436, opened 2023-05-22
Convert getOnDiskWorkspaceFolders to optionally use an App

robertbrignull:
This aims to remove a dependency that `QueryDiscovery` has on VS Code, by converting `getOnDiskWorkspaceFoldersObjects` to optionally take an app and use that instead of accessing `workspace.workspaceFolders` directly. The goal is to remove all dependencies on an actual workspace to make `QueryDiscovery` easier to test.

The reason for the `app` argument being optional is that there are a lot of existing references to `getOnDiskWorkspaceFoldersObjects` and `getOnDiskWorkspaceFolders` and it would be quite a lot of effort to convert all of them at once. Is it acceptable to do it this way and convert them gradually?

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/helpers.ts
+++ b/extensions/ql-vscode/src/helpers.ts
@@ -28,6 +28,7 @@ import { getQlPackPath } from "./pure/ql";
 import { dbSchemeToLanguage } from "./common/query-language";
 import { isCodespacesTemplate } from "./config";
 import { AppCommandManager } from "./common/commands";
+import { App } from "./common/app";
 
 // Shared temporary folder for the extension.
 export const tmpDir = dirSync({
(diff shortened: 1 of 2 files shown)


shati-patel (approved):
I think this makes sense. There might be other places that could benefit from using the `app` version of `getOnDiskWorkspaceFolders` instead of the `vscode` version, so perhaps it's worth a follow-up tech debt issue?

extensions/ql-vscode/src/helpers.ts
```diff
@@ -297,14 +298,18 @@ export async function showNeverAskAgainDialog(
 }
 
 /** Gets all active workspace folders that are on the filesystem. */
-export function getOnDiskWorkspaceFoldersObjects() {
-  const workspaceFolders = workspace.workspaceFolders ?? [];
+export function getOnDiskWorkspaceFoldersObjects(app?: App) {
+  const workspaceFolders =
+    (app !== undefined ? app.workspaceFolders : workspace.workspaceFolders) ??
```
koesie10, line 303:
```suggestion
    (app?.workspaceFolders ?? workspace.workspaceFolders) ??
```

koesie10 (commented):
How exactly are we aiming to test this? If we want to test this in the `no-workspace` tests, we could also use mocking on the `vscode.workspaceFolders` directly, like this:

https://github.com/github/vscode-codeql/blob/cc8f1fd8393551db474255e5cb69ecbec678050b/extensions/ql-vscode/test/vscode-tests/no-workspace/helpers.test.ts#L767

If we want to test this using unit tests instead, I think this won't be enough. The `src/helpers.ts` file imports from `vscode`, so this still has a runtime dependency on `vscode`.

robertbrignull:
I'm going to close this as no longer needed. Instead of writing pure unit tests for the queries panel we'll go with integration tests, so we don't want the extra complexity that this abstraction adds.

PR #2437, opened 2023-05-22
Use app.createEventEmitter in QueryDiscovery

robertbrignull:
Remove a small dependency that `QueryDiscovery` has on VS Code classes. I'm not sure if this one is strictly a blocker for making unit tests, but the method on `App` is already there so it costs us nothing to use it.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/queries-panel/query-discovery.ts
+++ b/extensions/ql-vscode/src/queries-panel/query-discovery.ts
@@ -6,6 +6,7 @@ import { MultiFileSystemWatcher } from "../common/vscode/multi-file-system-watch
 import { App } from "../common/app";
 import { FileTreeDirectory, FileTreeLeaf } from "../common/file-tree-nodes";
 import { getOnDiskWorkspaceFoldersObjects } from "../helpers";
+import { AppEventEmitter } from "../common/events";
 
 /**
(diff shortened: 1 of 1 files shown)


shati-patel (approved):
🚢

extensions/ql-vscode/src/queries-panel/query-discovery.ts
```diff
@@ -36,16 +30,15 @@ interface QueryDiscoveryResults {
 export class QueryDiscovery extends Discovery<QueryDiscoveryResults> {
   private results: QueryDiscoveryResults | undefined;
 
-  private readonly onDidChangeQueriesEmitter = this.push(
-    new EventEmitter<void>(),
-  );
+  private readonly onDidChangeQueriesEmitter;
```
koesie10, line 33: Can we add a type here? I think this is now typed as `any`.

robertbrignull: Ah thanks for pointing this out. For me locally it was working out the correct type. Perhaps it's smart enough to see the single assignment in the constructor.

But I do kinda like things to be explicit and readable to the human eye, so putting a type here sounds good to me regardless.

PR #2438, opened 2023-05-22
Convert multi-file-system-watcher to go through the app to make a watcher

robertbrignull:
The aim here is to remove a dependency that `QueryDiscovery` has on VS Code, by adding `createFileSystemWatcher` to the `App` interface and using it from `MultiFileSystemWatcher`. This will help make `QueryDiscovery` easier to test with unit tests, because it won't rely on a real workspace existing.

This unfortunately necessitated changing `QLTestDiscovery` as well, which dramatically increased the scope of this change. Hopefully this is fine, and theoretically this should have no behavioural change.

I've made the default implementation in `createMockApp` throw an error if called, since I expect this method to get called very infrequently os providing a mock implementation is likely not needed. I was kinda expecting the file system watcher tests to then fail, but they haven't. I'm not sure if that means we're not calling the watchers from the tests. I'm putting the PR up for review now, but I want to have a little bit more of a look into this and make sure everything is ok before the PR is merged.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/app.ts
+++ b/extensions/ql-vscode/src/common/app.ts
@@ -8,6 +8,7 @@ import type {
   WorkspaceFolder,
   Event,
   WorkspaceFoldersChangeEvent,
+  workspace,
 } from "vscode";
 
 export interface App {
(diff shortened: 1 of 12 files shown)


shati-patel (commented):
Nice! ❤️ I think I understand the changes (especially the goal of letting us write unit tests), but I wouldn't mind an additional review from someone who is a bit more comfortable with the `App` 😅 (@charisk possibly?)

robertbrignull:
I'm going to close this as no longer needed. Instead of writing pure unit tests for the queries panel we'll go with integration tests, so we don't want the extra complexity that this abstraction adds.

PR #2440, opened 2023-05-23
Retry results download if connection times out

JarLob:
Even with a good internet connection I sometimes have seen the error "failed to download results". When doing a MRVA on a 1000 top projects this results in that either you admit your run has incomplete results or have to rerun whole 1000 repositories query run, which is time consuming.
I have noticed that the url in the error message is still valid and I'm able to download the zip manually.
The change makes up to 3 retry attempts in a specific cases like timeout.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
@@ -77,6 +77,8 @@ import {
 import { GITHUB_AUTH_PROVIDER_ID } from "../common/vscode/authentication";
 import { FetchError } from "node-fetch";
 
+const maxRetryCount = 3;
+
 export class VariantAnalysisManager
   extends DisposableObject
   implements VariantAnalysisViewManager<VariantAnalysisView>
@@ -626,7 +628,7 @@ export class VariantAnalysisManager
             break;
           } catch (e) {
             if (
-              retry++ < 3 &&
+              retry++ < maxRetryCount &&
               e instanceof FetchError &&
               (e.code === "ETIMEDOUT" || e.code === "ECONNRESET")
             ) {
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
@@ -639,7 +639,7 @@ export class VariantAnalysisManager
               continue;
             }
             void extLogger.log(
-              `Failed to download variable analysis after ${retry} attempts.`,
+              `Failed to download variant analysis after ${retry} attempts.`,
             );
             throw e;
           }
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
@@ -76,6 +76,7 @@ import {
 } from "./repo-states-store";
 import { GITHUB_AUTH_PROVIDER_ID } from "../common/vscode/authentication";
 import { FetchError } from "node-fetch";
+import { extLogger } from "../common";
 
 const maxRetryCount = 3;
 
@@ -632,8 +633,14 @@ export class VariantAnalysisManager
               e instanceof FetchError &&
               (e.code === "ETIMEDOUT" || e.code === "ECONNRESET")
             ) {
+              void extLogger.log(
+                `Timeout while trying to download variant analysis with id: ${variantAnalysis.id}. Retrying...`,
+              );
               continue;
             }
+            void extLogger.log(
+              `Failed to download variable analysis after ${retry} attempts.`,
+            );
             throw e;
           }
         }
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
@@ -75,6 +75,7 @@ import {
   writeRepoStates,
 } from "./repo-states-store";
 import { GITHUB_AUTH_PROVIDER_ID } from "../common/vscode/authentication";
+import { FetchError } from "node-fetch";
 
 export class VariantAnalysisManager
   extends DisposableObject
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -613,12 +614,27 @@ export class VariantAnalysisManager
...
+            await this.variantAnalysisResultsManager.download(
+              variantAnalysis.id,
+              repoTask,
+              this.getVariantAnalysisStorageLocation(variantAnalysis.id),
+              updateRepoStateCallback,
+            );
+            break;
+          } catch (e) {
+            if (
+              retry++ < 3 &&
```
aeisenberg, line 629: Please extract this to a constant and place it at the top of the file.

aeisenberg (commented):
Thanks for your contribution. This looks reasonable to me. One small suggestion.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
JarLob: Done

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -613,12 +616,27 @@ export class VariantAnalysisManager
...
+              updateRepoStateCallback,
+            );
+            break;
+          } catch (e) {
+            if (
+              retry++ < maxRetryCount &&
+              e instanceof FetchError &&
+              (e.code === "ETIMEDOUT" || e.code === "ECONNRESET")
+            ) {
+              continue;
```
aeisenberg, line 635:
```suggestion
              extLogger.log("Timeout while trying to download variant analysis results. Retrying...");
              continue;
```

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -613,12 +616,27 @@ export class VariantAnalysisManager
...
+            }
+            throw e;
```
aeisenberg, line 637:
```suggestion
            extLogger.log(`Failed to download variable analysis after ${retry} attempts.`);
            throw e;
```

aeisenberg (commented):
Apologies. I just thought of this. We should log whenever we do a retry.

My code isn't exactly correct since there's no associated import statement.

aeisenberg (approved):

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -613,12 +617,33 @@ export class VariantAnalysisManager
...
+              e instanceof FetchError &&
+              (e.code === "ETIMEDOUT" || e.code === "ECONNRESET")
+            ) {
+              void extLogger.log(
+                `Timeout while trying to download variant analysis with id: ${variantAnalysis.id}. Retrying...`,
+              );
+              continue;
+            }
+            void extLogger.log(
+              `Failed to download variable analysis after ${retry} attempts.`,
```
JarLob, line 642:
```suggestion
              `Failed to download variant analysis after ${retry} attempts.`,
```

aeisenberg:
Failing tests are flakes. I'll re-run.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -613,12 +617,33 @@ export class VariantAnalysisManager
...
           }
         };
-        await this.variantAnalysisResultsManager.download(
-          variantAnalysis.id,
-          repoTask,
-          this.getVariantAnalysisStorageLocation(variantAnalysis.id),
-          updateRepoStateCallback,
-        );
+        let retry = 0;
+        for (;;) {
```
koesie10, line 621: Could this perhaps contain the `if` condition for the number of retries? That would make it somewhat easier to see that this is using retries rather than being an infinite loop.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -613,12 +617,33 @@ export class VariantAnalysisManager
...
+            );
+            break;
+          } catch (e) {
+            if (
+              retry++ < maxRetryCount &&
+              e instanceof FetchError &&
+              (e.code === "ETIMEDOUT" || e.code === "ECONNRESET")
+            ) {
+              void extLogger.log(
+                `Timeout while trying to download variant analysis with id: ${variantAnalysis.id}. Retrying...`,
```
koesie10, line 637: Can we also log the error message here? You can use the `getErrorMessage` function to convert the error into an error message to put in this log.

koesie10 (commented):
Thanks for adding this! Unfortunately, there is some extra work required to make this work when the connection drops when some parts have already been downloaded. When downloading, we only append to the ZIP file:

https://github.com/github/vscode-codeql/blob/f6b0ae203254cb81b375b76ee7619fb0435f0e03/extensions/ql-vscode/src/variant-analysis/variant-analysis-results-manager.ts#L94

If we restart the download without truncating this file, a corrupt ZIP file will be written. Can you add `await rm(zipFilePath, { force: true });` (with `rm` coming from `fs-extra`) in the `download` method of the `VariantAnalysisResultsManager` just before starting the `fetch`?

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
JarLob: I was unable to avoid `while(true)` because it needs to rethrow if max attempts were reached instead of continuing the loop and ending it on some condition. Of course it is possible to write a "fake" `while (retry < maxRetries)` which on one hand duplicates the maxRetries check and on the other hand won't ever be reached.
If you can come up how to rewrite it I'm open for suggestions.

koesie10: Makes sense, thanks for thinking about it

JarLob:
Good to be merged?

aeisenberg (approved):
Looks good! Thanks for reminding me to take another peak.

PR #2441, opened 2023-05-23
Expand description for `codeQL.cli.executablePath`

RasmusWL:
To match what the code actually does.

## Checklist

- [x] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request. -- **no user visible changes, so has not been updated**
- [x] Issues have been created for any UI or other user-facing changes made by this pull request. -- **no user visible changes, so has no issue has been created**

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -189,7 +189,7 @@
           "scope": "machine-overridable",
           "type": "string",
           "default": "",
-          "markdownDescription": "Path to the CodeQL executable that should be used by the CodeQL extension. The executable is named `codeql` on Linux/Mac and `codeql.exe` on Windows. If empty, the extension will look for a CodeQL executable on your shell PATH, or if CodeQL is not on your PATH, download and manage its own CodeQL executable."
+          "markdownDescription": "Path to the CodeQL executable that should be used by the CodeQL extension. The executable is named `codeql` on Linux/Mac and `codeql.exe` on Windows. If empty, the extension will look for a CodeQL executable on your shell PATH, or if CodeQL is not on your PATH, download and manage its own CodeQL executable (note: if you later introduce CodeQL on your PATH, the extension will prefer a CodeQL executable it has downloaded itself)."
         },
         "codeQL.runningQueries.numberOfThreads": {
           "type": "integer",


extensions/ql-vscode/package.json
```diff
@@ -189,7 +189,7 @@
           "scope": "machine-overridable",
           "type": "string",
           "default": "",
-          "markdownDescription": "Path to the CodeQL executable that should be used by the CodeQL extension. The executable is named `codeql` on Linux/Mac and `codeql.exe` on Windows. If empty, the extension will look for a CodeQL executable on your shell PATH, or if CodeQL is not on your PATH, download and manage its own CodeQL executable."
+          "markdownDescription": "Path to the CodeQL executable that should be used by the CodeQL extension. The executable is named `codeql` on Linux/Mac and `codeql.exe` on Windows. If empty, the extension will look for a CodeQL executable on your shell PATH, or if CodeQL is not on your PATH, download and manage its own CodeQL executable (note: if you later introduce CodeQL on your PATH, the extension will prefer a CodeQL executable it has downloaded itself)."
```
robertbrignull, line 192: Does this only apply if the setting is empty, or also if you set this to an explicit path? I assume the first option otherwise there's no way to stop using the managed executable.

Is there any way to get the extension to stop using the managed executable and switch to the one on the path?

RasmusWL: Only if the VS Code setting is empty.

The [logic is here](https://github.com/github/vscode-codeql/blob/00f4bc1fda591911d2a3c36c739b47af48f7d6d8/extensions/ql-vscode/src/codeql-cli/distribution.ts#L157-L208)

It goes in priority, and picks the first option that is set:

1. `customCodeQlPath` set in VS Code settings
2. `extensionSpecificCodeQlPath` (installation managed by extension)
3. Look through `PATH`

> Is there any way to get the extension to stop using the managed executable and switch to the one on the path?

I guess you can just delete the folders containing the CodeQL CLI downloaded by the extension, but otherwise no.

robertbrignull: Thanks for explaining. This makes me wonder why we use this priority because at first glance it doesn't seem to most friendly, and is hard to override at a later date. Not that I'm blaming you, or expect you to change anything. But it's interesting to highlight what the current behaviour actually is.

robertbrignull (approved):
I don't have any complaints about this PR. Documenting the current behaviour a bit more accurately is definitely good, so I think this PR is for sure beneficial.

It has highlighted that the current behaviour is a bit odd and arguably not very good in certain cases. But that doesn't mean we shouldn't document it.

extensions/ql-vscode/package.json
adityasharad: It is a bit wonky. I remember at the time of developing the extension we wanted to support these three paths, but I'm not convinced the current implementation was exactly the intent 😆 All the same, documenting now sounds great, and we can figure out how to improve it later in a way that minimises breakage.

aeisenberg: > It goes in priority, and picks the first option that is set:
>     1. `customCodeQlPath` set in VS Code settings
>     2. `extensionSpecificCodeQlPath` (installation managed by extension)
>     3. Look through `PATH`

It looks to me that steps 2 and 3 should be reversed.

aeisenberg (approved):

PR #2443, opened 2023-05-24
Remove member variables that could be local variables

robertbrignull:
Just noticed we currently don't use any of these variables outside of the constructor, so they don't technically have to be member variables. We could make them local variables instead and I think this makes things a bit cleaner.

What do you think? We can always change them back to member variables if we need them again in the future.

## Checklist


charisk (approved):
LGTM!

PR #2447, opened 2023-05-25
Add integration tests of QueryDiscovery

robertbrignull:
This PR introduces integration tests of the `QueryDiscovery` class.

Includes one fix for something I discovered while writing the tests. If a workspace folder doesn't include any query files, we probably don't want to create a `FileTreeDirectory` for it. This does raise some questions about what the panel will look like when there's no queries at all, but I think we can solve that and do it separately after this PR.

I've made the tests work with the code is it is now, but honestly after this PR I'm considering removing some of the vscode abstractions. So that basically means reverting https://github.com/github/vscode-codeql/pull/2437 and https://github.com/github/vscode-codeql/pull/2431 because they arguably make the code more confusing and testing harder.

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/minimal-workspace/queries-panel/query-discovery.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/minimal-workspace/queries-panel/query-discovery.test.ts
@@ -12,7 +12,7 @@ import {
 } from "../../../../src/queries-panel/query-discovery";
 import { createMockApp } from "../../../__mocks__/appMock";
 import { mockedObject } from "../../utils/mocking.helpers";
-import { basename, join } from "path";
+import { basename, join, sep } from "path";
 import { sleep } from "../../../../src/pure/time";
 
 describe("QueryDiscovery", () => {
@@ -101,7 +101,11 @@ describe("QueryDiscovery", () => {
     });
 
     it("calls resolveQueries once for each workspace", async () => {
-      const workspaceRoots = ["/workspace1", "/workspace2", "/workspace3"];
+      const workspaceRoots = [
+        `${sep}workspace1`,
+        `${sep}workspace2`,
+        `${sep}workspace3`,
+      ];
       jest.spyOn(workspace, "workspaceFolders", "get").mockReturnValueOnce(
         workspaceRoots.map((root, index) => ({
           uri: Uri.file(root),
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/test/vscode-tests/minimal-workspace/queries-panel/query-discovery.test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+      const results = await (discovery as any).discover();
+      (discovery as any).update(results);
+
+      expect(createFileSystemWatcherSpy).toHaveBeenCalledTimes(2);
+      expect(onDidChangeQueriesSpy).toHaveBeenCalledTimes(1);
+
+      onWatcherDidChangeEvent.fire(workspace.workspaceFolders![0].uri);
+
+      // Wait for refresh to finish
```
robertbrignull, line 183: I'm not fully happy with this. It appears to work, but I'd like to avoid sleeps like this because they can be a source of flakiness.

I was thinking if we could modify the `Discovery` class so you can await the completion of the current refresh. It might be a little fiddly, but it could make these tests and `QLTestDiscovery` tests more robust.

extensions/ql-vscode/test/vscode-tests/minimal-workspace/queries-panel/query-discovery.test.ts
```diff
@@ -0,0 +1,225 @@
...
+      ).toEqual(1);
+      expect(
+        results.queries[0].children[0].children[0].children[0].name,
+      ).toEqual("query2.ql");
+      expect(results.queries[0].children[0].children[1].name).toEqual(
+        "query1.ql",
+      );
+    });
+
+    it("calls resolveQueries once for each workspace", async () => {
```
shati-patel, line 103: Minor:
```suggestion
    it("calls resolveQueries once for each workspace folder", async () => {
```

shati-patel (approved):
Just a small comment, but LGTM! 

> I've made the tests work with the code is it is now, but honestly after this PR I'm considering removing some of the vscode abstractions. So that basically means reverting https://github.com/github/vscode-codeql/pull/2437 and https://github.com/github/vscode-codeql/pull/2431 because they arguably make the code more confusing and testing harder.

Thanks for thinking about this! I don't have a strong opinion on it either way 😅

extensions/ql-vscode/test/vscode-tests/minimal-workspace/queries-panel/query-discovery.test.ts
robertbrignull: I'll open an internal issue for this because I want to address it and not forget about it, but also don't want to delay this PR. I do want to continue working on these tests for the next few days and make them better.

robertbrignull:
> Thanks for thinking about this! I don't have a strong opinion on it either way 😅

I do want to continue working on these tests and make them better and the production code clearer too. So I just want to reassure I don't plan to leave tech debt here if at all possible. 👷🏼

PR #2448, opened 2023-05-25
Add initial implementation of auto-modeling

koesie10:
This adds an initial implementation of auto-modeling using the `/repos/:owner/:name/code-scanning/codeql/auto-model` endpoint. The button is hidden behind the `codeQL.dataExtensions.llmGeneration` setting.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
@@ -68,6 +68,14 @@ export function createAutoModelRequest(
   return request;
 }
 
+/**
+ * For now, we have a simplified model that only models methods as sinks. It does not model methods as neutral,
+ * so we aren't actually able to correctly determine that a method is neutral; it could still be a source or summary.
+ * However, to keep this method simple and give output to the user, we will model any method for which none of its
+ * arguments are modeled as sinks as neutral.
+ *
+ * If there are multiple arguments which are modeled as sinks, we will only model the first one.
+ */
 export function parsePredictedClassifications(
   predicted: Method[],
 ): Record<string, ModeledMethod> {
--- a/extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
@@ -149,5 +149,5 @@ function toMethodClassification(modeledMethod: ModeledMethod): Classification {
 }
 
 function toFullMethodSignature(method: Method): string {
-  return `${method.package}.${method.type}.${method.name}${method.signature}`;
+  return `${method.package}.${method.type}#${method.name}${method.signature}`;
 }
--- a/extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
@@ -48,13 +48,11 @@ export function createAutoModelRequest(
           modeledMethod.type === "none"
             ? undefined
             : toMethodClassification(modeledMethod),
-        usages: externalApiUsage.usages.map((usage) => usage.label),
+        usages: externalApiUsage.usages
+          .slice(0, 10)
+          .map((usage) => usage.label),
         input: `Argument[${argumentIndex}]`,
       };
-
-      if (method.usages.length > 10) {
-        method.usages = method.usages.slice(0, 10);
-      }
 
       if (modeledMethod.type === "none") {
         request.candidates.push(method);
@@ -64,12 +62,8 @@ export function createAutoModelRequest(
     }
   }
 
-  if (request.candidates.length > 100) {
-    request.candidates = request.candidates.slice(0, 100);
-  }
-  if (request.samples.length > 20) {
-    request.samples = request.samples.slice(0, 20);
-  }
+  request.candidates = request.candidates.slice(0, 100);
+  request.samples = request.samples.slice(0, 20);
 
   return request;
 }
--- a/extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
@@ -68,6 +68,61 @@ export function createAutoModelRequest(
   return request;
 }
 
+export function parsePredictedClassifications(
+  predicted: Method[],
+): Record<string, ModeledMethod> {
+  const predictedBySignature: Record<string, Method[]> = {};
+  for (const method of predicted) {
+    if (!method.classification) {
+      continue;
+    }
+
+    const signature = toFullMethodSignature(method);
+
+    if (!(signature in predictedBySignature)) {
+      predictedBySignature[signature] = [];
+    }
+
+    predictedBySignature[signature].push(method);
+  }
+
+  const modeledMethods: Record<string, ModeledMethod> = {};
+
+  for (const signature in predictedBySignature) {
+    const predictedMethods = predictedBySignature[signature];
+
+    const sinks = predictedMethods.filter(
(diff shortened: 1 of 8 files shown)


extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
```diff
@@ -0,0 +1,117 @@
...
+        signature: externalApiUsage.methodParameters,
+        classification:
+          modeledMethod.type === "none"
+            ? undefined
+            : toMethodClassification(modeledMethod),
+        usages: externalApiUsage.usages.map((usage) => usage.label),
+        input: `Argument[${argumentIndex}]`,
+      };
+
+      if (method.usages.length > 10) {
```
charisk, line 55: Do we need the `if` ? I think `slice` deals with a higher "end" number than than the length of the array.

Same comment below.

extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
```diff
@@ -0,0 +1,117 @@
...
+      }
+
+      if (modeledMethod.type === "none") {
+        request.candidates.push(method);
+      } else {
+        request.samples.push(method);
+      }
+    }
+  }
+
```
charisk, line 66: Now that we've created a `Method` for each argument, does it make sense to shuffle the array before slicing, to avoid ending up with the same method multiple times?

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -367,6 +381,40 @@ export class DataExtensionsEditorView extends AbstractWebview<
...
+    externalApiUsages: ExternalApiUsage[],
+    modeledMethods: Record<string, ModeledMethod>,
+  ): Promise<void> {
+    const request = createAutoModelRequest(
+      this.databaseItem.language,
+      externalApiUsages,
+      modeledMethods,
+    );
+
+    const response = await autoModel(this.app.credentials, request);
```
charisk, line 394: As a next step (outside of this PR), we should handle 429s in a nice way - say to the user to try again later or something like that.

charisk (approved):
Thanks for taking this over the line! I've not tested it locally yet but we can tweak separately.

charisk (approved):
Thanks for taking this over the line! I've not tested it locally yet but we can tweak separately.

extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
koesie10: I think we'll lose the sorting by number of usages in that case. I think we still do want that prioritization.

This also made me think of the processing of the `predicted` methods. If we send the same method multiple times, we'll also get multiple results. Right now the editor only supports 1 classification per method, so we would get a "random" classification from the supplied classifications. I'm not sure how to best deal with that.

extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
```diff
@@ -0,0 +1,111 @@
...
+  request.candidates = request.candidates.slice(0, 100);
+  request.samples = request.samples.slice(0, 20);
```
starcke, line 66: I think my intention was 20 candidates and 100 samples.

extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
starcke: I think there are general questions about how to deal with the response, given we classify a single argument.

We return from the LLM either "sink" or "not sink" for a specific argument, so unless the response is sink we cannot actually fill in anything. setting it to neutral would be wrong as that specific argument could be either a source or a summary, and other non checked arguments could still be sinks. I think this is a broader discussion we should record, but dont have to solve now.

I propose we do the following now:
- If there is a single argument of the method that is a sink, then pick the lowest argument index and use that as the result.
- If all arguments are non-sink then model this as neutral.

starcke (commented):
Comments from reviewing `createAutoModelRequest`.

extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
```diff
@@ -68,6 +68,61 @@ export function createAutoModelRequest(
   return request;
 }
 
+export function parsePredictedClassifications(
```
starcke, line 71: Non-blocking: It might be nice putting a comment saying something along the lines of what I wrote here: https://github.com/github/vscode-codeql/pull/2448#discussion_r1206645139

extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
```diff
@@ -68,6 +68,61 @@ export function createAutoModelRequest(
...
+  predicted: Method[],
+): Record<string, ModeledMethod> {
+  const predictedBySignature: Record<string, Method[]> = {};
+  for (const method of predicted) {
+    if (!method.classification) {
+      continue;
```
starcke, line 77: Should we log or warn if some samples come back unclassified?

extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
```diff
@@ -68,6 +68,61 @@ export function createAutoModelRequest(
...
+        input: "",
+        output: "",
+      };
+      continue;
+    }
+
+    // Order the sinks by the input alphabetically. This will ensure that the first argument is always
+    // first in the list of sinks, the second argument is always second, etc.
+    // If we get back "Argument[1]" and "Argument[3]", "Argument[1]" should always be first
+    sinks.sort((a, b) => (a.input ?? "").localeCompare(b.input ?? ""));
```
starcke, line 111: It doesnt matter I think but is `Argument[10]` also going to be sorted properly?

extensions/ql-vscode/test/unit-tests/data-extensions-editor/auto-model.test.ts
```diff
@@ -281,3 +288,96 @@ describe("createAutoModelRequest", () => {
...
+      package: "org.sql2o",
+      type: "Sql2o",
+      name: "Sql2o",
+      signature: "(String,String,String)",
+      usages: ["new Sql2o(...)"],
+      input: "Argument[1]",
+      classification: {
+        type: ClassificationType.Sink,
+        kind: "sql injection sink",
+        explanation: "not a sink",
```
starcke, line 343: It is slightly odd to have explanation `not a sink` when it is actually classified as a sink.

extensions/ql-vscode/test/unit-tests/data-extensions-editor/auto-model.test.ts
```diff
@@ -281,3 +288,96 @@ describe("createAutoModelRequest", () => {
...
+      package: "org.sql2o",
+      type: "Sql2o",
+      name: "Sql2o",
+      signature: "(String,String,String)",
+      usages: ["new Sql2o(...)"],
+      input: "Argument[2]",
+      classification: {
+        type: ClassificationType.Sink,
+        kind: "sql injection sink",
+        explanation: "not a sink",
```
starcke, line 356: As above.

starcke (approved):
Looks good, just some minor comments/questions.

extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
koesie10: No, it currently doesn't sort properly. I'll change it so it properly sorts them.

charisk (approved):
LGTM!

PR #2449, opened 2023-05-25
Add welcome text for the queries panel

robertbrignull:
This gives some default text to display in the queries panel where there aren't any queries in the workspace. Suggestions welcome for what to put as the text, and we can always change the text later on.

Screenshot 2023-05-25 at 12 48 17

This shows whenever `QueryTreeDataProvider` returns no children for the root node, with no extra configuration needed. One complication is it also shows while the CodeQL extension is loading, so ideally the text shouldn't be confusing in this situation.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -1594,6 +1594,10 @@
         "contents": "You have no query history items at the moment.\n\nSelect a database to run a CodeQL query and get your first results."
       },
       {
+        "view": "codeQLQueries",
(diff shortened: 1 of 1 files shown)


robertbrignull:
This might not be visible until https://github.com/github/vscode-codeql/pull/2447 is merged because that PR changes the behaviour to not include workspace folders which don't include any queries, as otherwise it'll always show an entry for the workspace folder, even if that directory has no children.

charisk (approved):

PR #2453, opened 2023-05-26
Add ability to await the discovery refresh promise

robertbrignull:
The intent of this PR is to give you access to a promise for the `Discovery` class refresh. This allows us to make the tests more type safe and less prone to flakiness.

One part of this PR is making `refresh` itself return a promise. This means in the tests we can do `await discovery.refresh()` instead of having to do `await (discovery as any).discover()` and calling the private `discover` method. This means we're no longer circumventing the public interface of the `Discover` class.

Another part of this PR is adding a `getCurrentRefreshPromise` method which means even if we weren't the ones to kick off the refresh directly we can still safely wait for it to be finished. This is useful when we are testing listeners to events that trigger a refresh, such as the listener for files changing on disk. We can now do `await discovery.getCurrentRefreshPromise()` instead of having to do `await sleep(100)` and hope for the best.

The commits are split up into separate changes, if you prefer to review commit-by-commit.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/discovery.ts
+++ b/extensions/ql-vscode/src/common/discovery.ts
@@ -8,7 +8,7 @@ import { getErrorMessage } from "../pure/helpers-pure";
  * same time.
  */
 export abstract class Discovery<T> extends DisposableObject {
-  private retry = false;
+  private restartWhenFinished = false;
   private currentDiscoveryPromise: Promise<void> | undefined;
 
   constructor(private readonly name: string) {
@@ -48,7 +48,7 @@ export abstract class Discovery<T> extends DisposableObject {
 
     if (this.currentDiscoveryPromise !== undefined) {
       // There's already a discovery operation in progress. Tell it to restart when it's done.
-      this.retry = true;
+      this.restartWhenFinished = true;
     } else {
       // No discovery in progress, so start one now.
       this.currentDiscoveryPromise = this.launchDiscovery();
@@ -72,13 +72,13 @@ export abstract class Discovery<T> extends DisposableObject {
       results = undefined;
     }
 
-    if (this.retry) {
+    if (this.restartWhenFinished) {
       // Another refresh request came in while we were still running a previous discovery
       // operation. Since the discovery results we just computed are now stale, we'll launch
       // another discovery operation instead of updating.
-      // Note that by doing this inside of `finally`, we will relaunch discovery even if the
-      // initial discovery operation failed.
-      this.retry = false;
+      // We want to relaunch discovery regardless of if the initial discovery operation
+      // succeeded or failed.
+      this.restartWhenFinished = false;
       await this.launchDiscovery();
     } else {
       this.currentDiscoveryPromise = undefined;
--- a/extensions/ql-vscode/src/common/discovery.ts
+++ b/extensions/ql-vscode/src/common/discovery.ts
@@ -42,7 +42,7 @@ export abstract class Discovery<T> extends DisposableObject {
     } else {
       // No discovery in progress, so start one now.
       this.discoveryInProgress = true;
-      this.launchDiscovery();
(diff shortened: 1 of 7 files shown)


shati-patel (commented):
Thanks! I've read through the commits and I think I understand the idea 😅 As far as I could tell (from local testing), the behaviour is still the same too 🎉 

I'm not super confident on the actual code changes, so I wouldn't mind a second look from a more confident TypeScript-y person 👀 ❤️

extensions/ql-vscode/src/common/discovery.ts
```diff
@@ -36,49 +46,48 @@ export abstract class Discovery<T> extends DisposableObject {
...
+      );
+      results = undefined;
+    }
 
-      .catch((err: unknown) => {
-        void extLogger.log(
-          `${this.name} failed. Reason: ${getErrorMessage(err)}`,
-        );
-      })
+    if (this.retry) {
```
koesie10, line 75: Is this really a retry? I would expect a retry to be something that happens when the discovery fails. Perhaps `restartWhenFinished` would cover this behaviour better?

extensions/ql-vscode/src/common/discovery.ts
```diff
@@ -36,49 +46,48 @@ export abstract class Discovery<T> extends DisposableObject {
...
+      // Another refresh request came in while we were still running a previous discovery
+      // operation. Since the discovery results we just computed are now stale, we'll launch
+      // another discovery operation instead of updating.
+      // Note that by doing this inside of `finally`, we will relaunch discovery even if the
+      // initial discovery operation failed.
+      this.retry = false;
+      await this.launchDiscovery();
+    } else {
+      this.currentDiscoveryPromise = undefined;
```
koesie10, line 84: The responsibility of updating this instance variable seems to be shared between the `refresh` method and this method. Would it perhaps be better to move the responsibility for this to the refresh method by changing the call-site to something like this:

```ts
this.currentDiscoveryPromise = this.launchDiscovery()
  .then(() => {
    this.currentDiscoveryPromise = undefined;
  })
  .catch((err: unknown) => {
    this.currentDiscoveryPromise = undefined;
    throw err;
  });
```

extensions/ql-vscode/src/common/discovery.ts
```diff
@@ -36,49 +46,48 @@ export abstract class Discovery<T> extends DisposableObject {
...
-          // Note that by doing this inside of `finally`, we will relaunch discovery even if the
-          // initial discovery operation failed.
-          this.retry = false;
-          this.launchDiscovery();
-        }
-      });
+      // If the discovery was successful, then update any listeners with the results.
+      if (results !== undefined) {
+        this.update(results);
+      }
```
koesie10, line 89: (May not be valid if implementing the previous suggestion): By doing this after setting `this.currentDiscoveryPromise = undefined;`, it means that waiting on the `refresh` or `getCurrentRefreshPromise` methods doesn't mean that the `update` method has finished. Is this intentional?

extensions/ql-vscode/src/common/discovery.ts
robertbrignull: That makes sense to me. This variable wasn't one I added, but I'm happy to rename while I'm here to make things clearer. I think `restartWhenFinished` is certainly clearer.

extensions/ql-vscode/src/common/discovery.ts
robertbrignull: That's a good point. That wasn't intentional. Thankfully the previous suggestion has already changed this behaviour so that `update` must finish first.

robertbrignull:
I think all comments are addressed now, and the tests are passing again. Should be ready for a quick re-review.

koesie10 (approved):
LGTM, thanks for implementing the suggestions

PR #2454, opened 2023-05-26
Don't log the `resolve queries` CLI command

shati-patel:
To keep the Queries Panel up to date, the CLI runs the `resolve queries` command every time a `.ql` file or surrounding folder changes.

The logging output is very noisy, so we'd like to remove it from the CodeQL Extension Log. I've passed in an optional `silent` flag to the `runCodeQlCliInternal` method, which just skips calling `logger.log`.

I've also added a basic test to check the behaviour with/without the flag.

## Checklist

N/A—internal only 🐕

### DIFF
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -412,10 +412,10 @@ export class CodeQLCliServer implements Disposable {
         const newError =
           stderrBuffers.length === 0
             ? new Error(
-                `${description} failed with args ${argsString}:${EOL}${err}`,
+                `${description} failed with args:${EOL}    ${argsString}${EOL}${err}`,
               )
             : new Error(
-                `${description} failed with args ${argsString}:${EOL}${Buffer.concat(
+                `${description} failed with args:${EOL}    ${argsString}${EOL}${Buffer.concat(
                   stderrBuffers,
                 ).toString("utf8")}`,
               );
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -330,6 +330,7 @@ export class CodeQLCliServer implements Disposable {
     commandArgs: string[],
     description: string,
     onLine?: OnLineCallback,
+    silent?: boolean,
   ): Promise<string> {
     const stderrBuffers: Buffer[] = [];
     if (this.commandInProcess) {
@@ -349,7 +350,12 @@ export class CodeQLCliServer implements Disposable {
       // Compute the full args array
       const args = command.concat(LOGGING_FLAGS).concat(commandArgs);
       const argsString = args.join(" ");
-      void this.logger.log(`${description} using CodeQL CLI: ${argsString}...`);
+      // If we are running silently, we don't want to print anything to the console.
+      if (!silent) {
+        void this.logger.log(
+          `${description} using CodeQL CLI: ${argsString}...`,
+        );
+      }
       try {
         await new Promise<void>((resolve, reject) => {
           // Start listening to stdout
@@ -395,7 +401,9 @@ export class CodeQLCliServer implements Disposable {
         const fullBuffer = Buffer.concat(stdoutBuffers);
         // Make sure we remove the terminator;
         const data = fullBuffer.toString("utf8", 0, fullBuffer.length - 1);
-        void this.logger.log("CLI command succeeded.");
+        if (!silent) {
+          void this.logger.log("CLI command succeeded.");
+        }
         return data;
       } catch (err) {
         // Kill the process if it isn't already dead.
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/test/vscode-tests/cli-integration/run-cli.test.ts
```diff
@@ -23,10 +24,14 @@ describe("Use cli", () => {
...
   let supportedLanguages: string[];
 
+  let logSpy: jest.SpiedFunction<BaseLogger["log"]>;
+
   beforeEach(async () => {
     const extension = await getActivatedExtension();
     cli = extension.cliServer;
     supportedLanguages = await cli.getSupportedLanguages();
+
+    logSpy = jest.spyOn(cli.logger, "log");
```
shati-patel, line 34: I had to make the CodeQLCLIServer's `logger` property `public readonly` for this to work. I think that's okay? But shout if not!

(PS: I initially tried `logSpy = jest.spyOn(extLogger, "log")` but that didn't work.)

extensions/ql-vscode/test/vscode-tests/cli-integration/run-cli.test.ts
```diff
@@ -42,6 +47,23 @@ describe("Use cli", () => {
     expect(result).toEqual(["-J-Xmx4096M", "--off-heap-ram=4096"]);
   });
 
+  describe("silent logging", () => {
+    it("should log command output", async () => {
```
shati-patel, line 51: I could have made these use [`itWithCodeQL`](https://github.com/github/vscode-codeql/blob/256b806cd47b27b430d9ed4fe785cbe2d1fd7b14/extensions/ql-vscode/test/vscode-tests/cli.ts#L31) (which checks whether the CodeQL libraries are checked out) instead of `it`. But I don't actually care whether `resolve queries` finds anything, just whether it was run!

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -412,7 +420,9 @@ export class CodeQLCliServer implements Disposable {
         newError.stack += getErrorStack(err);
         throw newError;
       } finally {
-        void this.logger.log(Buffer.concat(stderrBuffers).toString("utf8"));
+        if (!silent) {
+          void this.logger.log(Buffer.concat(stderrBuffers).toString("utf8"));
+        }
```
shati-patel, line 425: I've only wrapped the `logger.log` calls in the `if` statement. Some of the argument/buffer wrangling is possibly unnecessary if we don't log stuff, but I didn't want to mess with the internals of the command!

robertbrignull (approved):
LGTM

Thanks for adding those comments calling out specific bits. I'm happy with all the decisions you've made, but you're welcome to get another opinion if you'd like from someone more familiar with the `cli.ts` file.

One thing I do wonder is whether to make `resolveTests` silent too. It's behaviour is the same as resolving queries where it runs every time you change any file in the workspace.

extensions/ql-vscode/src/codeql-cli/cli.ts
aeisenberg: Does this mean we are not logging even if there are errors?

Maybe that makes sense since `resolve queries` will be run in the background often and writing errors to the log will be very noisy. However, I wonder if users should be made aware of at least the first error they see?

aeisenberg: And, related, if you do start logging errors, there's no log entry of the `argstring` (I find this _extremely_ useful for debugging). So, the argstring should be emitted if errors are coming out.

And...now that I think about it, stderr in the case show _all_ log messages, so it's probably best to hide it. It would just be nice to ensure that errors are emitted along with the argstring.

extensions/ql-vscode/test/vscode-tests/cli-integration/run-cli.test.ts
aeisenberg: Would something like this work? I know we're trying to avoid `as any` in the code, but I'm not sure what is better...

```suggestion
    logSpy = jest.spyOn((cli as any).logger, "log");
```

aeisenberg (commented):
The only thing I'd like to see in this PR is that the `argsString` is emitted in silent mode when there is an error thrown.

The other parts are just ideas that we can discuss and figure out in future PRs.

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -408,14 +408,16 @@ export class CodeQLCliServer implements Disposable {
       } catch (err) {
         // Kill the process if it isn't already dead.
         this.killProcessIfRunning();
-        // Report the error (if there is a stderr then use that otherwise just report the error cod or nodejs error)
+        // Report the error (if there is a stderr then use that otherwise just report the error code or nodejs error)
         const newError =
           stderrBuffers.length === 0
-            ? new Error(`${description} failed: ${err}`)
+            ? new Error(
+                `${description} failed with args ${argsString}:${EOL}${err}`,
```
aeisenberg, line 415: Thanks. I think this is good. Can I ask for one more thing? When I am debugging, often it is the list of args that I am most interested in, and I copy and paste it so I can run it directly in a terminal. Could you make the text more prominent? Eg- indented and on its own line?

shati-patel: Sure, that sounds like it'll be helpful! How's this?

screenshot of failed `resolveQueries` command

aeisenberg (approved):
Thanks!

PR #2455, opened 2023-05-26
Expand assert-pure.ql to cover common and be a path-problem query

robertbrignull:
Expands the `assert-pure.ql` query in a couple of ways:
- Covers stuff in the `src/common` directory as well as `src/pure`. Stuff in this directory should also be independent of vscode, unless it is in the `src/common/vscode` directory.
- Convert the query to a `path-problem` query. My aim here is to make this query easier to use, by explaining where an import comes from. It can otherwise be a bit hard to understand when the import chain gets more than a couple of files.

The query now has 10 alerts 🎉 😱 

Screenshot 2023-05-26 at 16 44 37

All of these alerts are because of logging, either from importing `extLogger` or `OutputChannelLogger` or `showAndLogErrorMessage`.

The reason there's 6 alerts for `selection-commands.ts` is because there's 6 separate vscode imports that it transitively imports, via the giant `helpers.ts` file. I believe there's not a good way to compress these into a single alert without losing the information of which import it's connected to.

## Checklist

### DIFF
--- a/.github/codeql/queries/assert-pure.ql
+++ b/.github/codeql/queries/assert-pure.ql
@@ -9,8 +9,8 @@
 
 import javascript
 
-class VSCodeImport extends AstNode {
-  VSCodeImport() { this.(Import).getImportedPath().getValue() = "vscode" }
+class VSCodeImport extends ImportDeclaration {
+  VSCodeImport() { this.getImportedPath().getValue() = "vscode" }
 }
 
 class PureFile extends File {
--- a/.github/codeql/queries/assert-pure.ql
+++ b/.github/codeql/queries/assert-pure.ql
@@ -1,21 +1,44 @@
 /**
  * @name Unwanted dependency on vscode API
- * @kind problem
+ * @kind path-problem
  * @problem.severity error
  * @id vscode-codeql/assert-pure
  * @description The modules stored under `pure` and tested in the `pure-tests`
  * are intended to be "pure".
  */
+
 import javascript
 
-class VSCodeImport extends ASTNode {
-  VSCodeImport() {
-    this.(Import).getImportedPath().getValue() = "vscode"
+class VSCodeImport extends AstNode {
+  VSCodeImport() { this.(Import).getImportedPath().getValue() = "vscode" }
+}
+
+class PureFile extends File {
+  PureFile() {
+    (
+      this.getRelativePath().regexpMatch(".*/src/pure/.*") or
+      this.getRelativePath().regexpMatch(".*/src/common/.*")
+    ) and
+    not this.getRelativePath().regexpMatch(".*/src/common/vscode/.*")
   }
+}
+
+Import getANonTypeOnlyImport(Module m) {
+  result = m.getAnImport() and not result.(ImportDeclaration).isTypeOnly()
+}
+
+Module getANonTypeOnlyImportedModule(Module m) {
+  result = getANonTypeOnlyImport(m).getImportedModule()
+}
+
+query predicate edges(AstNode a, AstNode b) {
+  getANonTypeOnlyImport(a) = b or
+  a.(Import).getImportedModule() = b
 }
 
 from Module m, VSCodeImport v
 where
-  m.getFile().getRelativePath().regexpMatch(".*src/pure/.*")  and
-  m.getAnImportedModule*().getAnImport() = v
-select m, "This module is not pure: it has a transitive dependency on the vscode API imported $@", v, "here"
+  m.getFile() instanceof PureFile and
+  getANonTypeOnlyImport(getANonTypeOnlyImportedModule*(m)) = v
+select m, m, v,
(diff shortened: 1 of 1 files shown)


.github/codeql/queries/assert-pure.ql
```diff
@@ -1,21 +1,44 @@
...
 
+Import getANonTypeOnlyImport(Module m) {
+  result = m.getAnImport() and not result.(ImportDeclaration).isTypeOnly()
+}
+
+Module getANonTypeOnlyImportedModule(Module m) {
+  result = getANonTypeOnlyImport(m).getImportedModule()
+}
+
+query predicate edges(AstNode a, AstNode b) {
```
aeisenberg, line 34: Does there need to be a `nodes` query? I'm not 100% sure how this works.

aeisenberg (commented):
This is really neat. Thanks for implementing.

.github/codeql/queries/assert-pure.ql
aeisenberg: I guess it's working based on your screenshot. Have you already fixed the alerts? Otherwise, I'm not sure why code scanning isn't reporting them.

adityasharad: If `nodes` isn't provided it defaults to all possible values of the columns in `edges`, so it's ok unless you need to constraint that set further.

.github/codeql/queries/assert-pure.ql
```diff
@@ -1,21 +1,44 @@
...
  */
+
 import javascript
 
-class VSCodeImport extends ASTNode {
-  VSCodeImport() {
-    this.(Import).getImportedPath().getValue() = "vscode"
+class VSCodeImport extends AstNode {
+  VSCodeImport() { this.(Import).getImportedPath().getValue() = "vscode" }
+}
```
adityasharad, line 14: Since you know the type you need, might as well declare it rather than casting.
```suggestion
class VSCodeImport extends Import {
  VSCodeImport() { this.getImportedPath().getValue() = "vscode" }
}
```

.github/codeql/queries/assert-pure.ql
```diff
@@ -1,21 +1,44 @@
...
+  a.(Import).getImportedModule() = b
+}
+
 from Module m, VSCodeImport v
 where
-  m.getFile().getRelativePath().regexpMatch(".*src/pure/.*")  and
-  m.getAnImportedModule*().getAnImport() = v
-select m, "This module is not pure: it has a transitive dependency on the vscode API imported $@", v, "here"
+  m.getFile() instanceof PureFile and
+  getANonTypeOnlyImport(getANonTypeOnlyImportedModule*(m)) = v
```
adityasharad, line 42: Minor: this looks correct, but it is less confusing if you can easily see the relationship between `edges` and the `select` clause. Right now I think each `getANonTypeOnlyImportedModule` step actually contains two `edges`(importing module -> import -> imported module) and the `getANonTypeOnlyImport` step is one edge (importing module -> import).

I think you could even write this as `edges+(m, v)` and it would work because of the typing of `v`, but verify that to be sure!

adityasharad (commented):
Neat! A couple of passing suggestions.

.github/codeql/queries/assert-pure.ql
robertbrignull: They do show up in code-scanning for me. They won't show up on the PR because the alerts don't appear in the diff, but you can see them on the security tab:
Screenshot 2023-05-30 at 11 32 43

And comparing to main:
Screenshot 2023-05-30 at 11 32 31

.github/codeql/queries/assert-pure.ql
robertbrignull: Changing it to `Import` gets some errors because `Import` is an abstract class.
Screenshot 2023-05-30 at 11 35 57

I assume this was the reason it was like this originally. However, I should be able to change it to `ImportDeclaration` because all of the imports in our codebase are `import` rather than `require`. My understanding is that means we can use `ImportDeclaration` and not miss anything. At least right now it doesn't change the results.

.github/codeql/queries/assert-pure.ql
robertbrignull: Ooh, good point. Thanks for saying this. I agree it's less confusing if it's clear the relation to `edges`, and we can just do `edges+(m, v)` because of the typing of `v`.

.github/codeql/queries/assert-pure.ql
adityasharad: Oof I didn't realise `Import` was abstract.
Changing it to `extends ImportDeclaration` is a good specific solution.
If you want something more generic, try the (relatively new) `instanceof` keyword: `class VSCodeImport extends AstNode instanceof Import`. `instanceof` says we are creating a subtype of `Import` that inherits all its methods, but does not add to its set of values as an abstract class.

adityasharad (approved):

PR #2457, opened 2023-05-30
Retrieve external API usage snippets using SARIF

koesie10:
This implements the retrieval of external API usage snippets using a SARIF file.

The alternative is retrieving the usage manually given a file and a range within this file. However, this is quite hard to do and would create a separate implementation from the one already present in the CodeQL CLI.

## Checklist


extensions/ql-vscode/src/data-extensions-editor/auto-model-usages-query.ts
```diff
@@ -0,0 +1,129 @@
+import { CancellationTokenSource } from "vscode";
```
starcke, line 1: I could use a few more comments inside this file.

extensions/ql-vscode/src/data-extensions-editor/auto-model-usages-query.ts
```diff
@@ -0,0 +1,129 @@
...
+    token: cancellationTokenSource.token,
+  });
+  if (!queryResult) {
+    throw new Error("Query failed");
+  }
+
+  progress({
+    maxStep,
+    step: 1100,
+    message: "Retrieving source locatin prefix",
```
starcke, line 60:
```suggestion
    message: "Retrieving source location prefix",
```

extensions/ql-vscode/src/data-extensions-editor/queries/csharp.ts
```diff
@@ -2,33 +2,52 @@ import { Query } from "./query";
...
- where
-   apiName = api.getApiName() and
-   supported = isSupported(api) and
-   usage = aUsage(api)
- select apiName, supported, usage
+from ExternalApi api, string apiName, Call usage
+where
+ apiName = api.getApiName() and
+ usage = aUsage(api)
+select usage, apiName
```
starcke, line 50: I am ok to do it like this for now. But it _feels_ like we should be able to reuse the previous query as it selects all the same data.

starcke (approved):
Looks pretty good to me. But probably also good to get a second review from Charis as it is touching some existing code.

extensions/ql-vscode/src/data-extensions-editor/auto-model-usages-query.ts
```diff
@@ -0,0 +1,129 @@
...
+    ["--sarif-add-snippets"],
+  );
+
+  progress({
+    maxStep,
+    step: 1400,
+    message: "Parsing results",
+  });
+
+  const snippets: Record<string, string[]> = {};
```
charisk, line 104: What do you think about moving the logic that extracts the snippets into `sarif-utils`? It'll make this function a bit smaller and it'll make it easier to test this aspect too.

extensions/ql-vscode/src/data-extensions-editor/auto-model-usages-query.ts
```diff
@@ -0,0 +1,129 @@
+import { CancellationTokenSource } from "vscode";
+import { join } from "path";
+import { runQuery } from "./external-api-usage-query";
+import { CodeQLCliServer } from "../codeql-cli/cli";
+import { QueryRunner } from "../query-server";
+import { DatabaseItem } from "../databases/local-databases";
+import { interpretResultsSarif } from "../query-results";
+import { ProgressCallback } from "../common/vscode/progress";
+
+type Options = {
```
charisk, line 10: This is okay for what we're doing, but the fact that we have to use `Pick` so much is a smell for me. It means entities aren't well broken down and it hurts readability.

extensions/ql-vscode/src/databases/local-databases/database-item-impl.ts
```diff
@@ -121,7 +121,9 @@ export class DatabaseItemImpl implements DatabaseItem {
   /**
    * Returns information about a database.
    */
-  private async getDbInfo(server: cli.CodeQLCliServer): Promise<cli.DbInfo> {
+  private async getDbInfo(
+    server: Pick<cli.CodeQLCliServer, "resolveDatabase">,
```
charisk, line 125: Why do we have to do this? Can we avoid it? I'm worried it'll hurt readability/maintainability gradually.

extensions/ql-vscode/src/data-extensions-editor/auto-model-usages-query.ts
```diff
@@ -0,0 +1,129 @@
...
+  progress: ProgressCallback;
+};
+
+export async function getAutoModelUsages({
+  cliServer,
+  queryRunner,
+  databaseItem,
+  queryStorageDir,
+  progress,
+}: Options): Promise<Record<string, string[]>> {
```
charisk, line 35: Would it make sense to add type for signature rather than just using `string`?

extensions/ql-vscode/src/data-extensions-editor/external-api-usage-query.ts
```diff
@@ -27,14 +28,17 @@ export type RunQueryOptions = {
...
-export async function runQuery({
-  cliServer,
-  queryRunner,
-  databaseItem,
-  queryStorageDir,
-  progress,
-  token,
-}: RunQueryOptions): Promise<CoreCompletedQuery | undefined> {
+export async function runQuery(
+  queryName: keyof Omit<Query, "dependencies">,
```
charisk, line 32: Similar to other `Pick`/`Omit` comments, can we avoid this change?

extensions/ql-vscode/src/data-extensions-editor/queries/csharp.ts
koesie10: I found a way to use a single query for this, so I'll change it to use a single query for both retrieving the external APIs and for retrieving the usages.

extensions/ql-vscode/src/data-extensions-editor/auto-model-usages-query.ts
koesie10: I don't think it really makes sense to move this into `sarif-utils`. This code is not really reusable since it's very specific to the SARIF file that is being produced by this exact query. Usually, you'd not want to create a map out of it since the message is the alert message.

For just the part `result.locations?.[0]?.physicalLocation?.contextRegion?.snippet?.text`, this is also not reusable. Usually, you would have multiple locations since the path contains multiple nodes.

I can extract the grouping into a separate function, but then we'd be exporting it only for testing. This is probably fine, but what do you think?

extensions/ql-vscode/src/data-extensions-editor/auto-model-usages-query.ts
koesie10: I've removed `Pick` from this definition. That means it's harder to test this part since you now need to know which methods are being called in the function rather than being able to derive that from the type, but it does improve readability.

extensions/ql-vscode/src/data-extensions-editor/external-api-usage-query.ts
koesie10: This is somewhat different in that it was excluding the `dependencies` field from the `Query` type, ensuring you're only able to pick a query name. However, I've changed the query so we only need 1, so this parameter is unnecessary now.

extensions/ql-vscode/src/data-extensions-editor/auto-model-usages-query.ts
charisk: Ah yeah, it's not really re-usable, you're right. Let's leave as is.

extensions/ql-vscode/src/data-extensions-editor/auto-model-usages-query.ts
koesie10: I don't think that will help since TypeScript doesn't strictly check type aliases (i.e. `type Signature = string`). If you pass a `string` where a `Signature` is required, it will allow it without any warning or error.

I have created a type for this return type so it's easier to know what it actually means.

extensions/ql-vscode/src/data-extensions-editor/auto-model-usages-query.ts
charisk: Yeah I appreciate it makes testing harder but I'm wary of using `Pick` and `Omit` as it can get out of hand quickly and really hurt readability. To me it's usually a sign that an interface is too big and needs to be broken down.

it may be that we'll decide to go with `Pick` in the short term anyway, but I'd rather we avoid it if possible as it spreads very easily.

charisk (approved):
LGTM. I'll defer the CodeQL query changes to @starcke.

starcke (approved):
Query looks good to me. It is similar to the approach the CodeML queries have.

extensions/ql-vscode/src/pure/bqrs-cli-types.ts
```diff
@@ -115,7 +115,7 @@ export type BqrsKind =
   | "Entity";
 
 interface BqrsColumn {
-  name: string;
+  name?: string;
```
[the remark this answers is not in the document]
koesie10: This is dependent on the query. In the query we're now using, the BQRS column definitions look like this:

```js
[
  { name: "usage", kind: "Entity" },
  { name: "apiName", kind: "String" },
  { kind: "String" },
  { kind: "String" },
]
```

CodeQL is probably not naming the last two columns because it can't automatically determine a name based on the name of a variable:

```ql
usage, apiName, supported.toString(), "supported"
```

PR #2460, opened 2023-05-31
Move selection-commands.ts to be in the vscode directory

robertbrignull:
Should fix https://github.com/github/vscode-codeql/security/code-scanning/521

The VS Code dependency we have is on `showAndLogErrorMessage` and I can't see a way to remove it without reducing functionality. But we can move it to the vscode directory and I don't think this is a huge problem compared to it being just in the common directory. The places where we'll want to use these methods from are going to already have VS Code dependencies.

## Checklist


shati-patel (approved):
🚢

Issue #2456, opened 2023-05-30
Add `selectedSourceLine` and `selectedSourceColumn` for use in contextual queries


aibaars:
**Is your feature request related to a problem? Please describe.**

For a "printCFG" query I'd like to be able to restrict the view to the Control flow graph of a single function instead of all functions in a file.  Displaying all graphs of all files is too messy and hard to work with.

**Describe the solution you'd like**

In addition to the `selectedSourceFile`, I'd like the selected source line and columns to be available for use in contextual queries.  That way a `printCfg` query can restrict itself to the enclosing function/callable at the position of the cursor.

**Describe alternatives you've considered**
none

see also: https://github.slack.com/archives/CPA9033L4/p1682926461306599?thread_ts=1682664735.271779&cid=CPA9033L4

PR #2461, opened 2023-05-31
View CFG: export selected line and column position

aibaars:
For a "printCFG" query I'd like to be able to restrict the view to the Control flow graph of a single function instead of all functions in a file. Displaying all graphs of all files is too messy and hard to work with.

Therefore this pull request adds `selectedSourceLine`, and `selectedSourceColumn` values to the context of a `printCfg` query in addition to `selectedSourceFile`. That way a printCfg query can restrict itself to the enclosing function/callable at the position of the cursor. 

This PR can be tested with https://github.com/github/codeql/pull/13334

Fixes: #2456

## Checklist


extensions/ql-vscode/src/language-support/contextual/template-provider.ts
```diff
@@ -284,15 +286,17 @@ export class TemplatePrintCfgProvider {
...
-    }
-
     return this.shouldUseCache()
-      ? await this.cache.get(document.uri.toString())
-      : await this.getCfgUri(document.uri.toString());
+      ? await this.cache.get(
+          `${document.uri.toString()}#${line}:${character}`,
+          line,
+          character,
+        )
```
aeisenberg, line 298: So, now we will be caching by line/character/path. Will the cache blow up if someone requests the same CFG but on different caret locations on the same line?

Since this feature is mostly used for debugging CFG implementations, I wonder if it is even useful to use the cache at all? Presumably most users of this graph will run with caching disabled.

aeisenberg (commented):
Thanks for the contribution. Looks broadly reasonable. A few questions inline.

extensions/ql-vscode/src/language-support/contextual/template-provider.ts
```diff
@@ -342,7 +348,9 @@ export class TemplatePrintCfgProvider {
     const queryUri = Uri.file(queries[0]);
 
     const templates: Record<string, string> = {
-      [TEMPLATE_NAME]: zippedArchive.pathWithinSourceArchive,
+      [SELECTED_SOURCE_FILE]: zippedArchive.pathWithinSourceArchive,
+      [SELECTED_SOURCE_LINE]: line.toString(),
+      [SELECTED_SOURCE_COLUMN]: character.toString(),
```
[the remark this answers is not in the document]
aibaars: As far as I know it is fine if a query does not use all the available template parameters, so this should be backwards compatible.

extensions/ql-vscode/src/language-support/contextual/template-provider.ts
aibaars: I don't think the cache is very useful ; I just added the line an column in the cache-key to avoid returning the same CFG each time regardless of where the cursor is. The cache seems to have LRU strategy, so hopefully it should not blow up and evict old results. We can also disable caching together.

aeisenberg (approved):
Thanks for answering my questions. I haven't tried this myself, but it looks safe enough since it's not used externally.

extensions/ql-vscode/src/language-support/contextual/template-provider.ts
aibaars: I think it might still be useful to have the cache. Even though you could click in arbitrary places in a method, as a user of the feature I'd most likely right-click on the method name so the chance of hitting the cache if perhaps not that low.

aibaars:
@aeisenberg could you merge this PR if you are happy with it? I don't have write permissions.

aeisenberg:
Merging with main so that the new, expected checks can run.

PR #2464, opened 2023-05-31
Move the invocation rate limiter class out of helpers.ts

robertbrignull:
Moves the `InvocationRateLimiter` class out of `helpers.ts` and into its own file. Then by having it take the `Memento` interface as an argument instead of the full `ExtensionContext` we can move it to the `common` directory and dramatically reduce the amount of mocking in the tests.

## Checklist


extensions/ql-vscode/test/vscode-tests/no-workspace/codeql-cli/distribution.test.ts
```diff
@@ -286,7 +286,7 @@ describe("Launcher path", () => {
     const manager = new DistributionManager(
       { customCodeQlPath: pathToCmd } as any,
       {} as any,
-      undefined as any,
+      {} as any,
```
robertbrignull, line 289: I'm not hugely happy that we're doing `as any` in the tests, but I don't think the change in this PR makes it any worse than it was already.

extensions/ql-vscode/test/unit-tests/common/invocation-rate-limiter.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    async update(key: string, value: any): Promise<void> {
+      this.map.set(key, value);
+    }
+  }
+
+  class MockGlobalStorage extends MockMemento {
+    public setKeysForSync(_keys: string[]): void {
+      return;
+    }
+  }
```
koesie10, line 61: I see that this is from the previous testing implementation, but do we need this? It seems like just having `MockMemento` is enough.

koesie10 (approved):
Thanks for moving this!

extensions/ql-vscode/test/unit-tests/common/invocation-rate-limiter.test.ts
robertbrignull: Ah, I see why we had this. It's because the real type defined on the `ExtensionContext` is
```typescript
readonly globalState: Memento & {
	/**
	 * Set the keys whose values should be synchronized across devices when synchronizing user-data
	 * like configuration, extensions, and mementos.
	 *
	 * Note that this function defines the whole set of keys whose values are synchronized:
	 *  - calling it with an empty array stops synchronization for this memento
	 *  - calling it with a non-empty array replaces all keys whose values are synchronized
	 *
	 * For any given set of keys this function needs to be called only once but there is no harm in
	 * repeatedly calling it.
	 *
	 * @param keys The set of keys whose values are synced.
	 */
	setKeysForSync(keys: readonly string[]): void;
};
```
So we had to implement `setKeysForSync` too. But now we've declared that `InvocationRateLimiter` only takes a `Memento` instead of `typeof ExtensionContext.globalState`, we don't have to mock this method in the tests anymore.

PR #2467, opened 2023-06-01
Fix neutral definition for CodeQL 2.13.3

koesie10:
In CodeQL 2.13.3, the definition of the neutralModel predicate [has changed](https://github.com/github/codeql/pull/12931) to include the `kind`. This updates the definition of the data extensions editor to match the new definition.

One caveat is that when selecting a `kind` other than `summary`, the method will not be shown as supported. This is because a [`NeutralCallable`](https://github.com/github/codeql/blob/ff78ac98d27c7b9f1adffcf235c56855f8348ad0/java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImpl.qll#L338) only calls into [`neutralSummaryElement`](https://github.com/github/codeql/blob/ff78ac98d27c7b9f1adffcf235c56855f8348ad0/java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImplSpecific.qll#L160). This matches the previous behavior because setting the `kind` to `source` or `sink` only says that the method is either not a source or not a sink, but not both. Only `summary` fully models the method.

## Checklist


starcke (approved):
Nice! I am impressed that the change is so small.

> One caveat is that when selecting a kind other than summary, the method will not be shown as supported. This is because a [NeutralCallable](https://github.com/github/codeql/blob/ff78ac98d27c7b9f1adffcf235c56855f8348ad0/java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImpl.qll#L338) only calls into [neutralSummaryElement](https://github.com/github/codeql/blob/ff78ac98d27c7b9f1adffcf235c56855f8348ad0/java/ql/lib/semmle/code/java/dataflow/internal/FlowSummaryImplSpecific.qll#L160). This matches the previous behavior because setting the kind to source or sink only says that the method is either not a source or not a sink, but not both. Only summary fully models the method.

Good to know, this ties into the whole discussion of whether we should model functions or arguments and how to deal with things that have have multiple classifications (e.g. an argument that is both a sink and a flow summary)

PR #2468, opened 2023-06-01
Remove `ignoreSourceArchive` option from databases

koesie10:
This option was used to ignore source archives for `.testproj` databases. It is only set to `true` or `false` when creating the database and could not be changed, so I don't think we need this option. It can simply be derived from the database URI. This simplifies handling of databases a bit.

## Checklist


norascheuch (approved):
Looks good!

koesie10:
@aeisenberg Do you know if this is actually safe to do? The `ignoreSourceArchive` option is not being set in the extension itself, but could there be users who are changing this value manually in the workspace state?

aeisenberg (approved):
I don't see any problem with this. It looks fine, but I haven't tried this out locally, though.

PR #2471, opened 2023-06-01
Modify QueryDiscovery to resolve queries by language

robertbrignull:
Changes `QueryDiscovery` to pass the `--format=bylanguage` arg to `codeql resolve queries`. This makes it indicate which language each query is for, if that information is known. We're not using this new info yet, but that'll come soon so we want to have the data ready.

The tests proved to be quite fiddly to get right, and creation of mock data is a big clunky. There might be cleaner ways of doing this, and we could look into that if this solution is too awful.

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/minimal-workspace/queries-panel/query-discovery.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/minimal-workspace/queries-panel/query-discovery.test.ts
@@ -5,7 +5,10 @@ import {
   WorkspaceFoldersChangeEvent,
   workspace,
 } from "vscode";
-import { CodeQLCliServer } from "../../../../src/codeql-cli/cli";
+import {
+  CodeQLCliServer,
+  QueryInfoByLanguage,
+} from "../../../../src/codeql-cli/cli";
 import { QueryDiscovery } from "../../../../src/queries-panel/query-discovery";
 import { createMockApp } from "../../../__mocks__/appMock";
 import { mockedObject } from "../../utils/mocking.helpers";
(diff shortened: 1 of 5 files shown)


shati-patel (commented):
[as discussed offline as well]

I've tested this locally and, whilst it does work, it's unbearably slow on my (Windows) machine 😅 E.g. in the `vscode-codeql-starter` workspace, it takes well over a minute to populate/update the queries panel. 

This is an issue with the underlying CLI command to resolve queries by language, so we should think about where/how best to solve this. (I'll link an existing internal issue about the slowness on Windows... 🔗)

extensions/ql-vscode/test/vscode-tests/minimal-workspace/queries-panel/query-discovery.test.ts
```diff
@@ -16,31 +19,63 @@ describe("QueryDiscovery", () => {
...
+    dataFn: (queryDir: string) => {
+      byLanguage?: Record<string, string[]>;
+      noDeclaredLanguage?: string[];
+      multipleDeclaredLanguages?: string[];
+    },
+  ) {
+    return jest.fn().mockImplementation((queryDir: Uri) => {
+      const data = dataFn(queryDir.fsPath);
+      const value: QueryInfoByLanguage = {
+        byLanguage: Object.keys(data.byLanguage || {}).reduce((result, key) => {
```
koesie10, line 32: Could we use `Object.entries` here instead of `Object.keys` to avoid needing to call `data.byLanguage![key]`?

extensions/ql-vscode/test/vscode-tests/minimal-workspace/queries-panel/query-discovery.test.ts
```diff
@@ -16,31 +19,63 @@ describe("QueryDiscovery", () => {
...
+          result[key] = queriesArrayToRecord(data.byLanguage![key]);
+          return result;
+        }, {} as QueryInfoByLanguage["byLanguage"]),
+        noDeclaredLanguage: queriesArrayToRecord(data.noDeclaredLanguage || []),
```
koesie10, line 36: We could move the `|| []` into the declaration of `queriesArrayToRecord` by adding `queries: string[] = []`. I think this would make this a bit shorter.

robertbrignull:
We're no longer going ahead with this PR. We hit some snags using `--format bylanguage` and decided to change the approach we use. Instead of using the CodeQL CLI to get all query information for us, we'll do a lot of it more manually.

PR #2472, opened 2023-06-01
Queries panel: display the query language in the UI

shati-patel:
Updates the `QueryTreeViewItem`, so that it can display the language of a CodeQL query (and display it in the UI). This involved updating the shared `FileTreeNode` class to accept a generic `data` property. 

This will hook in to https://github.com/github/vscode-codeql/pull/2471 (which gets the actual query language). For now, I've hard-coded a placeholder string: 

Queries panel showing a `language` label next to each query

## Checklist

N/A—internal only 🌍 See internal linked issue for more details! 🤫 

### DIFF
--- a/extensions/ql-vscode/src/common/file-tree-nodes.ts
+++ b/extensions/ql-vscode/src/common/file-tree-nodes.ts
@@ -100,7 +100,7 @@ export class FileTreeDirectory<T = undefined> extends FileTreeNode<T> {
  * A single file.
  */
 export class FileTreeLeaf<T = undefined> extends FileTreeNode<T> {
-  constructor(_path: string, _name: string, _data: T) {
+  constructor(_path: string, _name: string, _data?: T) {
     super(_path, _name, _data);
   }
 
--- a/extensions/ql-vscode/src/common/file-tree-nodes.ts
+++ b/extensions/ql-vscode/src/common/file-tree-nodes.ts
@@ -8,7 +8,7 @@ export abstract class FileTreeNode<T = undefined> {
   constructor(
     private _path: string,
     private _name: string,
-    private _data: T | undefined,
+    private _data?: T,
   ) {}
 
   public get path(): string {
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/common/file-tree-nodes.ts
```diff
@@ -91,12 +99,12 @@ export class FileTreeDirectory extends FileTreeNode {
 /**
  * A single file.
  */
-export class FileTreeLeaf extends FileTreeNode {
-  constructor(_path: string, _name: string) {
-    super(_path, _name);
+export class FileTreeLeaf<T = undefined> extends FileTreeNode<T> {
+  constructor(_path: string, _name: string, _data: T) {
```
robertbrignull, line 103:
```suggestion
  constructor(_path: string, _name: string, _data?: T) {
```
If we make the `_data` arg optional here then we can revert the change to `qltest-discovery.ts` and avoid changing that file at all in this PR, but not change behaviour.

robertbrignull (approved):
I have one tiny suggestion which you are free to take or leave. But this PR and approach looks good to me. Admittedly I helped with developing it.

From my point of view you're welcome to get another review, or go ahead with this if you like. I don't think any of the changes here are hard to review, so there's not really any risk.

shati-patel:
> From my point of view you're welcome to get another review, or go ahead with this if you like. I don't think any of the changes here are hard to review, so there's not really any risk.

One more thing I was wondering: is there a useful way to test this change? There are quite a few optional/undefined params, so perhaps it would be helpful to test some different combinations of setting them 🤔

shati-patel:
> > From my point of view you're welcome to get another review, or go ahead with this if you like. I don't think any of the changes here are hard to review, so there's not really any risk.
> 
> One more thing I was wondering: is there a useful way to test this change? There are quite a few optional/undefined params, so perhaps it would be helpful to test some different combinations of setting them 🤔

Update (as discussed offline): there doesn't seem to be much useful to test here, so we'll wait until we've also hooked up the "backend" in https://github.com/github/vscode-codeql/pull/2471 and then test things as a whole.

PR #2474, opened 2023-06-02
Handle 429 from auto-model service

charisk:
Shows a slightly more helpful message when we receive 429s from the auto-model endpoint.

## Checklist
N/A - internal only:

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
@@ -7,6 +7,7 @@ import {
   workspace,
   WorkspaceFolder,
 } from "vscode";
+import { RequestError } from "@octokit/request-error";
 import {
   AbstractWebview,
   WebviewPanelConfig,
@@ -39,7 +40,7 @@ import { createDataExtensionYaml, loadDataExtensionYaml } from "./yaml";
 import { ExternalApiUsage } from "./external-api-usage";
 import { ModeledMethod } from "./modeled-method";
 import { ExtensionPackModelFile } from "./shared/extension-pack";
-import { autoModel } from "./auto-model-api";
+import { autoModel, ModelRequest, ModelResponse } from "./auto-model-api";
 import {
   createAutoModelRequest,
   parsePredictedClassifications,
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -479,4 +483,19 @@ export class DataExtensionsEditorView extends AbstractWebview<
       message: "",
     });
   }
+
+  private async callApi(request: ModelRequest): Promise<ModelResponse | null> {
```
koesie10, line 487: Can we make this name a little less generic? The data extensions editor isn't only for auto-modeling, so it's not clear from the name which API this is calling.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -479,4 +483,19 @@ export class DataExtensionsEditorView extends AbstractWebview<
...
+    try {
+      return await autoModel(this.app.credentials, request);
+    } catch (e) {
+      if (e instanceof RequestError && e.status === 429) {
+        void showAndLogExceptionWithTelemetry(
+          redactableError`Rate limit hit, please try again soon.`,
```
koesie10, line 493: We should probably also report the full error:

```suggestion
          redactableError(e)`Rate limit hit, please try again soon.`,
```

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
charisk: 🤦‍♀️

koesie10 (approved):
LGTM, although the progress doesn't get cleared when I run into a rate limit error. I think you'll need to add `await this.clearProgress();` just before `return;` in the `generateModeledMethodsFromLlm` method.

charisk:
> LGTM, although the progress doesn't get cleared when I run into a rate limit error. I think you'll need to add `await this.clearProgress();` just before `return;` in the `generateModeledMethodsFromLlm` method.

Ah, I was contemplating whether to tidy this up in this PR. I've added `this.clearProgress()` in the catch block to make sure we clear on other failures. too.

koesie10:
> Ah, I was contemplating whether to tidy this up in this PR. I've added `this.clearProgress()` in the catch block to make sure we clear on other failures. too.

Thanks! We might want to move that into a `catch` block inside the `generateModeledMethodsFromLlm` method to also catch errors when the query fails, but this is error is much more likely to happen, so that can happen in a follow-up.

PR #2476, opened 2023-06-02
Code search: use code search api

norascheuch:
Instead of using repo search we want to use the code search api for this feature. Since the search api returns an error if two language parameters are added we made the entry of a language through the UI optional.

Items:
- uses correct api
- implements throttling
- makes 'no language' available

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
+++ b/extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
@@ -9,40 +9,18 @@ import {
 import { Repository } from "./repository";
 import { Progress } from "vscode";
 import { CancellationToken } from "vscode-jsonrpc";
-import { throttling } from "@octokit/plugin-throttling";
 import { Octokit } from "@octokit/rest";
-import { showAndLogWarningMessage } from "../../helpers";
 
 export async function getCodeSearchRepositories(
-  credentials: Credentials,
   query: string,
   progress: Progress<{
     message?: string | undefined;
     increment?: number | undefined;
   }>,
   token: CancellationToken,
+  octokit: Octokit,
 ): Promise<string[]> {
   let nwos: string[] = [];
-  const MyOctokit = Octokit.plugin(throttling);
-  const auth = await credentials.getAccessToken();
-
-  const octokit = new MyOctokit({
-    auth,
-    throttle: {
-      onRateLimit: (retryAfter: number, options: any): boolean => {
-        void showAndLogWarningMessage(
-          `Request quota exhausted for request ${options.method} ${options.url}. Retrying after ${retryAfter} seconds!`,
-        );
-
-        return true;
-      },
-      onSecondaryRateLimit: (_retryAfter: number, options: any): void => {
-        void showAndLogWarningMessage(
-          `SecondaryRateLimit detected for request ${options.method} ${options.url}`,
-        );
-      },
-    },
-  });
 
   for await (const response of octokit.paginate.iterator(
     octokit.rest.search.code,
--- a/extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
+++ b/extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
@@ -7,44 +7,6 @@ import {
   VariantAnalysisSubmissionRequest,
 } from "./variant-analysis";
 import { Repository } from "./repository";
-import { Progress } from "vscode";
-import { CancellationToken } from "vscode-jsonrpc";
-import { Octokit } from "@octokit/rest";
-
-export async function getCodeSearchRepositories(
-  query: string,
-  progress: Progress<{
-    message?: string | undefined;
-    increment?: number | undefined;
-  }>,
-  token: CancellationToken,
-  octokit: Octokit,
-): Promise<string[]> {
-  let nwos: string[] = [];
-
-  for await (const response of octokit.paginate.iterator(
-    octokit.rest.search.code,
-    {
-      q: query,
-      per_page: 100,
-    },
-  )) {
-    nwos.push(...response.data.map((item) => item.repository.full_name));
-    // calculate progress bar: 80% of the progress bar is used for the code search
-    const totalNumberOfRequests = Math.ceil(response.data.total_count / 100);
-    // Since we have a maximum of 1000 responses of the api, we can use a fixed increment whenever the totalNumberOfRequests would be greater than 10
-    const increment =
-      totalNumberOfRequests < 10 ? 80 / totalNumberOfRequests : 8;
-    progress.report({ increment });
-
-    if (token.isCancellationRequested) {
-      nwos = [];
-      break;
-    }
-  }
-
-  return [...new Set(nwos)];
-}
 
 export async function submitVariantAnalysis(
   credentials: Credentials,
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
```diff
@@ -20,18 +22,38 @@ export async function getCodeSearchRepositories(
...
+        return true;
+      },
+      onSecondaryRateLimit: (_retryAfter: number, options: any): void => {
+        void showAndLogWarningMessage(
+          `SecondaryRateLimit detected for request ${options.method} ${options.url}`,
+        );
+      },
+    },
+  });
+
```
norascheuch, line 45: Curiously I never hit the secondary Rate Limit during testing. However, the RateLimit was hit. Even though throttling was used the maximum entries in one list I ever got with one search was 901.

extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
```diff
@@ -20,18 +23,38 @@ export async function getCodeSearchRepositories(
   token: CancellationToken,
 ): Promise<string[]> {
   let nwos: string[] = [];
-  const octokit = await credentials.getOctokit();
+  const MyOctokit = Octokit.plugin(throttling);
+  const auth = await credentials.getAccessToken();
+
+  const octokit = new MyOctokit({
+    auth,
```
robertbrignull, line 30: To copy more closely what we do at https://github.com/github/vscode-codeql/blob/134c4405f9c7a2ca5af3c9eaa0a01a8fb86cdc50/extensions/ql-vscode/src/common/vscode/authentication.ts#L35-L38, this should ideally include the `retry` plugin too. Or is that intentionally missed out? The idea of this plugin is that'll automatically retry requests in cases of a 500 error.

```suggestion
    auth,
    retry,
```

extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
```diff
@@ -20,18 +23,38 @@ export async function getCodeSearchRepositories(
...
+      onRateLimit: (retryAfter: number, options: any): boolean => {
+        void showAndLogWarningMessage(
+          `Request quota exhausted for request ${options.method} ${options.url}. Retrying after ${retryAfter} seconds!`,
+        );
+
+        return true;
+      },
+      onSecondaryRateLimit: (_retryAfter: number, options: any): void => {
+        void showAndLogWarningMessage(
+          `SecondaryRateLimit detected for request ${options.method} ${options.url}`,
```
robertbrignull, line 41:
```suggestion
          `Request quota exceeded for request ${options.method} ${options.url}`.,
```
Just a suggestion to make this error fit with the other one.

extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
robertbrignull: > Even though throttling was used the maximum entries in one list I ever got with one search was 901.

Is this because of throttling, or because of duplicate repos in the search results? If you didn't hit the secondary rate limit then it means you should have results from all 10 pages, but it's just the total number of unique repositories was 901.

robertbrignull (commented):
Some thoughts on wording of things, but I've tried this out and it worked for me.

I don't know too much about this `throttling` plugin or how the github rate limit behaves, so I can't comment too much on that part right now. If you'd like a deeper review of that part just let me know.

I also found the progress reporting a bit confusing, but it seemed to work well enough in practice when I ran some searches.

extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
norascheuch: I think it makes sense to distinguish between the two, so that users know what they are hitting specifically.

extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
norascheuch: That was an oversight, will add it now!

norascheuch: I don't think we need the retry plugin here, as the throttle plugin is taking over retrying?

extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
norascheuch: It seems to be this certain search query. I tested with some other queries and had different results.

norascheuch: I was told the search api returns different results when sending the same request multiple times, this doesn't seem to be true for me right now..

norascheuch:
@robertbrignull  Reviewing this code I noticed that the gh-api-client file where the `getCodeSearchRepositories` method is written in is in the variant-analysis folder. It seems to me that code search belongs more to databases and database lists than variant-analysis per-se. I think it makes more sense to move both `getCodeSearchRepositories` and `getOctokitForSearch` to it's own file inside the `extensions/ql-vscode/src/databases/`-folder.

charisk:
> @robertbrignull Reviewing this code I noticed that the gh-api-client file where the `getCodeSearchRepositories` method is written in is in the variant-analysis folder. It seems to me that code search belongs more to databases and database lists than variant-analysis per-se. I think it makes more sense to move both `getCodeSearchRepositories` and `getOctokitForSearch` to it's own file inside the `extensions/ql-vscode/src/databases/`-folder.

I think this refactoring makes sense - having the API client related code into one file inside `/databases` seems the right thing to do in terms of code structure and it helps with testing. 

Whether that refactoring is done in this PR or a separate one is something that I'll leave to you and @robertbrignull to decide though 😃

extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
robertbrignull: I believe they deal with different concerns. The `retry` plugin will retry in response to 5xx errors. The `throttle` plugin will retry in response to 429 (Too many requests) errors and understands GitHub's specific messages.

But seems that in the latest version of `provideOctokitWithThrottling` the `retry` plugin is included, so that all looks good to me.

extensions/ql-vscode/src/variant-analysis/gh-api/gh-api-client.ts
robertbrignull: I agree users should be able to distinguish and understand them. I was meaning more that we could make the language style more consistent between the two error messages. For example:
- In one we say "quota" and in the other we say "limit". Is there a technical different between these I'm not aware of, and will the user know what they mean?
- In one we use full sentences, and in the other we use CamelCase.

However, this is something that can easily be changed afterwards, if time is tight. So it's ok to defer this to a later PR if needed.

norascheuch: I took this from the documentation and it copies also a script we have somewhere else in the repo. However you're right with the CamelCase, that's just weird. I fixed the CC and adjusted the wording!

robertbrignull (approved):
LGTM, especially given that this is now behind a feature flag.

I did notice a couple of tiny bits we can improve regarding the cancellation token, but we can discuss those outside of this PR. I'll raise issues or discuss them with you.

robertbrignull (approved):
LGTM, especially given that this is now behind a feature flag.

I did notice a couple of tiny bits we can improve regarding the cancellation token, but we can discuss those outside of this PR. I'll raise issues or discuss them with you.

PR #2478, opened 2023-06-05
Protect codeSearch feature for now

norascheuch:
## Checklist


extensions/ql-vscode/src/config.ts
```diff
@@ -718,3 +718,12 @@ const LLM_GENERATION = new Setting("llmGeneration", DATA_EXTENSIONS);
 export function showLlmGeneration(): boolean {
   return !!LLM_GENERATION.getValue<boolean>();
 }
+
+/**
+ * A flag indicating whether to make code search for feeding variant analysis lists accessible.
+ */
+const CODE_SEARCH = new Setting("codeSearch", ROOT_SETTING);
```
charisk, line 725: Would it make sense to make this a variant analysis setting rather than a root level one?

extensions/ql-vscode/package.json
```diff
@@ -967,7 +967,7 @@
         },
         {
           "command": "codeQLVariantAnalysisRepositories.importFromCodeSearch",
-          "when": "view == codeQLVariantAnalysisRepositories && viewItem =~ /canImportCodeSearch/",
+          "when": "view == codeQLVariantAnalysisRepositories && viewItem =~ /canImportCodeSearch/ && config.codeQL.codeSearch",
```
robertbrignull, line 970: This is cool. Not sure how I hadn't realised this but I didn't know you could reference settings directly from `package.json`.

In this case do we need the changes to `config.ts` at all? The `CODE_SEARCH` constant, and `showCodeSearch` function are only there so we can reference the setting from within our own code, but we don't currently need to do that.

extensions/ql-vscode/src/config.ts
robertbrignull: I don't know. On https://github.com/github/vscode-codeql/pull/2476#issuecomment-1576438499 we thought it best to consider it a piece of "databases" functionality rather than variant analysis. So my initial thought was that this setting should also mirror that. Happy to be persuaded otherwise though.

extensions/ql-vscode/package.json
norascheuch: Indeed, my mistake

extensions/ql-vscode/src/config.ts
norascheuch: Removed this part of code because it's not needed

robertbrignull (approved):
Works for me when tested

PR #2480, opened 2023-06-05
Add EnvironmentContext so common code can do localisation

robertbrignull:
Adds `EnvironmentContext` to the `App` which allows access to the `vscode.env.language`, for use with calls to `localeCompare` and similar. This PR shows how much change is needed in order to support passing the env data to where it's needed in `FileTreeNode`.

I've done just this one usage because it will fix https://github.com/github/vscode-codeql/security/code-scanning/518, but obviously in future we could use this in more places. we've got an internal issue to track this and make sure we're doing localisation consistently.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/app.ts
+++ b/extensions/ql-vscode/src/common/app.ts
@@ -23,10 +23,15 @@ export interface App {
   readonly onDidChangeWorkspaceFolders: Event<WorkspaceFoldersChangeEvent>;
   readonly credentials: Credentials;
   readonly commands: AppCommandManager;
+  readonly environment: EnvironmentContext;
 }
 
 export enum AppMode {
   Production = 1,
   Development = 2,
   Test = 3,
+}
+
+export interface EnvironmentContext {
+  language: string;
 }
--- a/extensions/ql-vscode/src/common/vscode/vscode-app.ts
+++ b/extensions/ql-vscode/src/common/vscode/vscode-app.ts
@@ -8,6 +8,7 @@ import { Memento } from "../memento";
 import { VSCodeAppEventEmitter } from "./events";
 import { AppCommandManager, QueryServerCommandManager } from "../commands";
 import { createVSCodeCommandManager } from "./commands";
+import { AppEnvironmentContext } from "./environment-context";
 
 export class ExtensionApp implements App {
   public readonly credentials: VSCodeCredentials;
@@ -71,8 +72,6 @@ export class ExtensionApp implements App {
   }
 
   public get environment(): EnvironmentContext {
-    return {
-      language: vscode.env.language,
-    };
+    return new AppEnvironmentContext();
   }
 }
--- a/extensions/ql-vscode/src/common/vscode/vscode-app.ts
+++ b/extensions/ql-vscode/src/common/vscode/vscode-app.ts
@@ -1,7 +1,7 @@
 import * as vscode from "vscode";
 import { VSCodeCredentials } from "./authentication";
 import { Disposable } from "../../pure/disposable-object";
-import { App, AppMode } from "../app";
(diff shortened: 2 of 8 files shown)


charisk (approved):
LGTM but happy to receive another review since I suggested the `EnvironmentContext` thing.

extensions/ql-vscode/src/common/vscode/vscode-app.ts
```diff
@@ -69,4 +69,10 @@ export class ExtensionApp implements App {
   public createEventEmitter<T>(): AppEventEmitter<T> {
     return new VSCodeAppEventEmitter<T>();
   }
+
+  public get environment(): EnvironmentContext {
+    return {
+      language: vscode.env.language,
```
koesie10, line 75: Should this be an object with a getter? I believe you can change the VS Code language while it's running and this would not reflect that if you're passing `app.environment` to another class.

extensions/ql-vscode/src/common/app.ts
```diff
@@ -23,10 +23,15 @@ export interface App {
   readonly onDidChangeWorkspaceFolders: Event<WorkspaceFoldersChangeEvent>;
   readonly credentials: Credentials;
   readonly commands: AppCommandManager;
+  readonly environment: EnvironmentContext;
```
koesie10, line 26: I'm wondering how we should decide whether something belongs in `environment` or in the `App` itself. Right now it reflects `vscode.env`, but it doesn't seem like there are other properties which we would need in the extension. Would there be any advantages to adding it as `App.language`? That does mean the complete `App` object needs to be passed around, but this seems to already be happening in a lot of places anyway.

I'm definitely in favour of splitting up the `App` object, I'm just not sure when we should split it up, and when we should add it to the `app` directly (like `workspaceFolders`).

extensions/ql-vscode/src/queries-panel/query-discovery.ts
```diff
@@ -124,7 +124,7 @@ export class QueryDiscovery
       return undefined;
     }
 
-    const rootDirectory = new FileTreeDirectory<string>(fullPath, name);
+    const rootDirectory = new FileTreeDirectory<string>(fullPath, name, env);
```
robertbrignull, line 127: Realised something silly I'm doing here. We should be passing in `this.app.environment` here so we're using the `App` that's being passed in where possible.

extensions/ql-vscode/src/common/vscode/vscode-app.ts
robertbrignull: Thanks for pointing that out. I agree using a getting would be good here.

extensions/ql-vscode/src/common/app.ts
charisk: Its a good question and I think it'll generally be difficult to answer. To me it makes sense to try and group things somewhat so that we don't pass the whole App object when we don't need it.  

For `environment` I can imagine other things like timezone information being added to it.

robertbrignull: I've opened https://github.com/github/vscode-codeql/pull/2481 to show what the `Localizer` option would look like.

koesie10: Thanks for trying it out! I think I prefer this version for now because it keeps all the VSCode and testing specific things in the single `App` interface. I'm fine to go ahead with this PR, this is not a block consideration from my view.

koesie10 (approved):

extensions/ql-vscode/src/common/app.ts
charisk: FWIW I also prefer the current implementation but it's great to see the other one in action! Thanks Robert!

PR #2483, opened 2023-06-06
Fix remaining `localeCompare` without locale set

koesie10:
This adds the `locales` parameter to the last two `localeCompare` which were not passing in the locale. This should make sorting more consistent and correct for non-English users.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/auto-model.ts
@@ -195,7 +195,8 @@ export function compareInputOutput(a: string, b: string): number {
 
     // If either one is unknown, it is sorted last
     if (aIndex === -1 && bIndex === -1) {
-      return a.localeCompare(b);
+      // Use en-US because these are well-known strings that are not localized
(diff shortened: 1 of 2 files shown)


robertbrignull:
The changes made here look good. I agree there are cases when we want to always use a consistent ordering instead of the users' locale.

There are a couple of places where we pass `undefined`. Should we update these too?
- https://github.com/github/vscode-codeql/blob/a7a24fc0e9edb462c778fa4eb7a9620364c56180/extensions/ql-vscode/src/run-queries-shared.ts#L568
- https://github.com/github/vscode-codeql/blob/a7a24fc0e9edb462c778fa4eb7a9620364c56180/extensions/ql-vscode/src/pure/variant-analysis-filter-sort.ts#L98

robertbrignull:
Apologies, I'm not thinking straight apparently. I see you commented on the one in `run-queries-shared.ts` in the other issue, and it also has a comment explaining why it doesn't pass a locale. And the other one is addressed in this PR! 🤦🏼

robertbrignull (approved):

PR #2490, opened 2023-06-07
Add manual discovery of queries and query packs

robertbrignull:
A new method for determining CodeQL queries and query packs. Also see the linked internal issue for some more details.

This is quite a large PR, but it's mostly new code and over half of the changes are in the tests, so that hopefully makes it better. This PR is split into commits which deal with each separate area, but I would recommend perhaps not looking at the diff compared to existing code and instead just review the new code as it stands.

There are quite a few differences to how it operated before:
- Uses two discovery classes, one for queries and one for query packs. Each one is watching for file system events and keeping its internal state up to date.
- Takes an approach of trying to avoid scanning the entire directory tree where at all possible. So we listen to events for files changes and keep track of the URIs from those events. The `FilePathDiscovery` base class implements all of this logic, so that the `QueryDiscovery` and `QueryPackDiscovery` classes can focus on their individual aspects.
- The only call to the CodeQL CLI is to determine the dbscheme for a query pack, and this only happens once per query pack or on changes to that pack.
- Modifies the `Discovery` framework slightly by merging the `discover` and `update` methods into one. See https://github.com/github/vscode-codeql/pull/2490#discussion_r1226437598 for a discussion of the behavioural effects this may have. This works better for the `FilePathDiscovery` class because it keeps an internal queue of events of changed files and then when `refresh` is called it works through that list of events. Therefore `discover` is not an idempotent operation with a defined return value, but rather it modifies the internal state to keep up with external changes.

I recommend testing this out manually too to get a feel for how it behaves. It should be able to display queries in any repository. The only limitation currently is it can't display the query language in cases where the query can't be compiled. So it'll work fine in the `vscode-codeql-starter` repo, but won't be able to display the language in other repos. I'm happy to accept this limitation because the main use-case for this is when writing queries, so you'll have an environment where query compilation is possible.

## Checklist

  - No changelog entry because the queries panel is behind a feature flag.


extensions/ql-vscode/src/queries-panel/query-discovery.ts
```diff
@@ -1,147 +1,245 @@
...
-  watchPaths: Uri[];
+interface Query {
+  path: string;
+  language: string | undefined;
 }
 
 /**
  * Discovers all query files contained in the QL packs in a given workspace folder.
  */
 export class QueryDiscovery
```
robertbrignull, line 38: It's occurred to me that there's a lot of duplication between these two discovery classes in the way they have the queue of paths and maintain an array of known files. Very likely a shared base class could be pulled out.

extensions/ql-vscode/src/queries-panel/query-pack-discovery.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  protected async getDataForPath(path: string): Promise<QueryPack> {
+    const language = await this.determinePackLanguage(path);
+    return { path, language };
+  }
+
+  private async determinePackLanguage(
+    path: string,
+  ): Promise<string | undefined> {
+    const packInfo = await this.cliServer.resolveLibraryPath(
```
robertbrignull, line 60: This works when called on query packs inside the QL repo itself, but on other repos (e.g. `github/codeql-action`) I'm getting errors like the following:
```
Resolving library paths failed with args:
    resolve library-path -v --log-to-stderr --format json --query /Users/robertbrignull/github/codeql-action/queries/qlpack.yml --additional-packs /Users/robertbrignull/github/codeql-action
[2023-06-08 16:45:56] This is codeql execute cli-server -v --log-to-stderr
[2023-06-08 16:45:57] [DETAILS] resolve library-path> Resolving query at normalized path /Users/robertbrignull/github/codeql-action/queries/qlpack.yml.
[2023-06-08 16:45:57] [DETAILS] resolve library-path> Found enclosing pack 'codeql-action-custom-queries-javascript' at /Users/robertbrignull/github/codeql-action/queries.
[2023-06-08 16:45:57] [DETAILS] resolve library-path> Adding compilation cache at /Users/robertbrignull/.codeql/compile-cache.
[2023-06-08 16:45:57] [DETAILS] resolve library-path> Resolving library dependencies from /Users/robertbrignull/github/codeql-action/queries/qlpack.yml.
[2023-06-08 16:45:57] [SPAMMY] resolve library-path> [INCOMPATIBILITY] codeql-action-custom-queries-javascript: not 0.0.0 {root: codeql-action-custom-queries-javascript@0.0.0}
[2023-06-08 16:45:57] [SPAMMY] resolve library-path> [DERIVATION] codeql-action-custom-queries-javascript: 0.0.0 {codeql-action-custom-queries-javascript: not 0.0.0 {root: codeql-action-custom-queries-javascript@0.0.0}}
[2023-06-08 16:45:57] [SPAMMY] resolve library-path> [INCOMPATIBILITY] codeql-action-custom-queries-javascript: * [*], codeql-javascript: not * {dependency: codeql-action-custom-queries-javascript@* [*] requires codeql-javascript@*}
[2023-06-08 16:45:57] [SPAMMY] resolve library-path> [DECISION 1] codeql-action-custom-queries-javascript: 0.0.0
[2023-06-08 16:45:57] [SPAMMY] resolve library-path> [DERIVATION] codeql-javascript: * {codeql-action-custom-queries-javascript: * [*], codeql-javascript: not * {dependency: codeql-action-custom-queries-javascript@* [*] requires codeql-javascript@*}}
[2023-06-08 16:45:57] [SPAMMY] resolve library-path> [INCOMPATIBILITY] codeql-javascript: * [*] {pack not found: codeql-javascript}
[2023-06-08 16:45:57] [SPAMMY] resolve library-path> [CONFLICT] codeql-javascript: * [*] {pack not found: codeql-javascript}
[2023-06-08 16:45:57] [SPAMMY] resolve library-path>   [CONSTRAINT] codeql-action-custom-queries-javascript: 0.0.0
[2023-06-08 16:45:57] [SPAMMY] resolve library-path>   [CONSTRAINT] codeql-javascript: *
[2023-06-08 16:45:57] [SPAMMY] resolve library-path>   [SATISFIER] derivation: codeql-javascript: *
[2023-06-08 16:45:57] [SPAMMY] resolve library-path>   [ALMOST_SATISFIER] <none>
[2023-06-08 16:45:57] [SPAMMY] resolve library-path> [INCOMPATIBILITY] codeql-action-custom-queries-javascript: * [*] {conflict}
[2023-06-08 16:45:57] [SPAMMY] resolve library-path> [FAILURE]
[2023-06-08 16:45:58] [SPAMMY] resolve library-path>   Because 'codeql-action-custom-queries-javascript' depends on 'codeql-javascript@*', but pack 'codeql-javascript' was not found, version solving failed.
[2023-06-08 16:45:58] [ERROR] resolve library-path> ERROR: Referenced pack 'codeql-javascript' was not found locally, and no lock file is present for 'codeql-action-custom-queries-javascript'. Run 'codeql pack install' to create the lock file and download the dependencies. (/Users/robertbrignull/github/codeql-action/queries/qlpack.yml:1,1-1)
[2023-06-08 16:45:58] [ERROR] resolve library-path> ERROR: No valid pack solution found:
                              Because 'codeql-action-custom-queries-javascript' depends on 'codeql-javascript@*', but pack 'codeql-javascript' was not found, version solving failed.
                               (/Users/robertbrignull/github/codeql-action/queries/qlpack.yml:1,1-1)
[2023-06-08 16:45:58] Exiting with code 1
```

robertbrignull: Hmm, maybe this is fine, or at least unavoidable. If you open these queries in the editor normally, they don't compile, and if you try to run them on the command line they fail. This is the point of the `vscode-codeql-starter and `codespaces-codeql` repos, to provide an environment where queries can be compiled. In that case it's a shame that we can't show the query language in this case, but it's not the end of the world. So long as it works in all cases where the query could be compiled and run that's ok.

extensions/ql-vscode/src/queries-panel/query-pack-discovery.ts
```diff
@@ -0,0 +1,93 @@
...
+  path: string;
+  language: string | undefined;
+}
+
+/**
+ * Discovers all query packs contained in the workspace.
+ */
+export class QueryPackDiscovery extends FilePathDiscovery<QueryPack> {
+  constructor(app: App, private readonly cliServer: CodeQLCliServer) {
+    super("Query Pack Discovery", "**/{qlpack.yml,codeql-pack.yml}", app);
```
aeisenberg, line 23: Does this capture the scenario when a new pack file is added to the workspace?

extensions/ql-vscode/src/queries-panel/query-discovery.ts
```diff
@@ -1,147 +1,106 @@
...
+      );
+      if (queriesInRoot.length > 0) {
+        const root = new FileTreeDirectory<string>(
+          workspaceFolder.uri.fsPath,
+          workspaceFolder.name,
+          this.app.environment,
+        );
+        for (const query of queriesInRoot) {
+          const dirName = dirname(normalize(relative(root.path, query.path)));
+          const parentDirectory = root.createDirectory(dirName);
```
aeisenberg, line 60: It looks like this creates a hierarchy that is two deep: Workspace root folder -> query. Is there any indication of which query packs each query belongs to?

Will packs with no queries still be shown? Is this what we want?

aeisenberg (commented):
Lots of questions. Not many answers.

robertbrignull:
> Lots of questions. Not many answers.

Thanks for taking a look. Sorry there were a few more unexplained bits in there than I remembered. I'll try to explain my reasoning on these.

extensions/ql-vscode/src/pure/helpers-pure.ts
```diff
@@ -67,3 +68,28 @@ export function asError(e: unknown): Error {
 
   return e instanceof Error ? e : new Error(String(e));
 }
+
+export function isEqualOrHigherPath(parent: string, child: string): boolean {
```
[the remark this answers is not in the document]
robertbrignull: My mind blanked when I was writing this function so I couldn't think of a better name. I think `isChildPath` is a much better name. There's some nuance that it returns true if the paths are equal, but hopefully that can be addressed with documentation comments.

extensions/ql-vscode/src/queries-panel/file-path-discovery.ts
```diff
@@ -0,0 +1,163 @@
...
+interface PathData {
+  path: string;
+}
+
+/**
+ * Discovers all query files contained in the QL packs in a given workspace folder.
+ */
+export abstract class FilePathDiscovery<
+  T extends PathData,
+> extends Discovery<unknown> {
```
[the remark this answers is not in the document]
robertbrignull: Thanks for bringing this up. I should have explained this more.

I think that the `Discovery` superclass could still be useful, but it's not perfect and hence this slight misuse of it. The problem I was running into here is that the `Discovery` class provides the `discover` and `update` methods but I don't really want the `update` method. This new class doesn't operate in quite the same way of building up a big object that describes the whole state, but rather makes small updates keeps track of its own state as it goes.

The piece of behaviour that this supports is that it only calls `update` (and hence only shows any change in the UI) when there hasn't been another refresh requested. I'm not really sure why this behaviour is important, but from my point of view it seems ok to change it to always call `update` or even combine the two methods together so there's only one abstract method to override.

The behavioural effect of merging these methods would be you'd see more updates. So if a refresh is in progress and then another refresh is requested, we'd still update the UI with the first results and then update again with the second set of results. This doesn't seem like bad behaviour to me, and possibly is better because if refreshes happen a lot and discovery takes a long time then you could go a while before the `update` method is called with any new data.

The effect of the code would be one less abstract method on the `Discover` class and we remove the need for the templated type.

Overall these effects both seem good or at least neutral changes, so I'm going to go ahead and make the change. I think it's probably fine, but we could get eyes from `@dbartol` if we're not confident about it.

extensions/ql-vscode/src/pure/helpers-pure.ts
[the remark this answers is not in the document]
robertbrignull: Actually, I've just found the `files.ts` file and the existing `containsPath` method. I think I'll reuse that.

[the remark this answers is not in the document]
robertbrignull: I've opened https://github.com/github/vscode-codeql/pull/2509 which makes some small updates to `containsPath`. I think it's a little buggy, but it's only used from one place so we probably never came across that case.

extensions/ql-vscode/src/queries-panel/query-pack-discovery.ts
```diff
@@ -0,0 +1,93 @@
...
+      packs.length >= 2 &&
+      dirname(packs[0].path) === dirname(packs[1].path)
+    ) {
+      if (basename(packs[0].path) === FALLBACK_QLPACK_FILENAME) {
+        return packs[0];
+      } else {
+        return packs[1];
+      }
+    } else {
+      return packs[0];
```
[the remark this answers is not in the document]
robertbrignull: I'll do a pass and add docs to everything that needs it.

extensions/ql-vscode/src/queries-panel/query-pack-discovery.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    // Sort by descreasing path length to find the pack nearest the query
+    packs.sort((a, b) => b.path.length - a.path.length);
+
+    // If the first two packs are from the same directory then look at the filenames
+    if (
+      packs.length >= 2 &&
+      dirname(packs[0].path) === dirname(packs[1].path)
+    ) {
+      if (basename(packs[0].path) === FALLBACK_QLPACK_FILENAME) {
```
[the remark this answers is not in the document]
robertbrignull: I think it does, though it may not be very clear. The `paths` array should be keeping track of every file with a name from `QLPACK_FILENAMES`, so this should include both kinds of pack names.

If there's both files present in a directory, what does CodeQL do? I've assumed that the one identified by `FALLBACK_QLPACK_FILENAME` (which is "qlpack.yml") wins and we use that file. That's what we're attempting to do `findQueryPack` by filtering to everything in a directory and then if we get two files trying to pick the right one.

Does that sound like the right logic?

extensions/ql-vscode/src/queries-panel/query-pack-discovery.ts
robertbrignull: I believe so. The watcher is installed on every workspace folder and not just those that currently contain query packs. It then triggers whenever a file changes that matches this pattern, as well as whenever a directory changes.

I'll make sure to test this in all the ways I can think of.

extensions/ql-vscode/src/queries-panel/query-discovery.ts
robertbrignull: Yes, the top level entry is the workspace root folder. We don't directly indicate which pack a query comes from, but we display the language of the query next to it and we get that info from the query pack.
Screenshot 2023-06-12 at 16 06 57

> Will packs with no queries still be shown? Is this what we want?

I checked about this and it's not part of the plan to show query packs in the queries panel. So we won't directly show or link to them, and so we also won't show any packs that don't have queries in them. This is obviously something that could be changed in the future if it would be useful for users to have quick access to their query packs, without going back to the regular file explorer tab.

robertbrignull:
@github/code-scanning-secexp, @aeisenberg, this is now ready for review. I've updated the PR description so do please have another read. It's a large PR, which I apologise for, and I appreciate any input even if it's only on part of the PR. At least it's mostly new code and also mostly tests, so I hope that makes it easier to deal with.

I've tried to incorporate all feedback, and I've also rewritten parts to be clearer and better documented. The commits are split up into different areas of interest, though it may also be easier to just review the final form and ignore the diff.

I've tested this manually, and @shati-patel has also tried it out on Windows. I'd appreciate any more testing we can give it, perhaps of edge cases around query pack definitions or adding new packs while the queries panel is active.

extensions/ql-vscode/src/common/file-path-set.ts
```diff
@@ -0,0 +1,42 @@
...
+      // The new path is a child of an existing path, so don't add it.
+      return;
+    } else {
+      // Remove any existing paths that are children of the new path.
+      this.paths = this.paths.filter((p) => !containsPath(path, p));
+      this.paths.push(path);
+    }
+  }
+
+  /** Removes and returns a path from the set, if the set of non-empty. */
```
koesie10, line 38:
```suggestion
  /** Removes and returns a path from the set, if the set is non-empty. */
```

extensions/ql-vscode/src/common/query-language.ts
```diff
@@ -25,7 +25,7 @@ export const PACKS_BY_QUERY_LANGUAGE = {
   [QueryLanguage.Ruby]: ["codeql/ruby-queries"],
 };
 
-export const dbSchemeToLanguage = {
+export const dbSchemeToLanguage: Record<string, string> = {
```
koesie10, line 28: Would it be more accurate to say this instead:

```suggestion
export const dbSchemeToLanguage: Record<string, QueryLanguage> = {
```

(you might also need to update the values to say `QueryLanguage.JavaScript` instead of `javascript`)

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
```diff
@@ -0,0 +1,190 @@
...
+    return this.refresh();
+  }
+
+  private workspaceFoldersChanged(event: WorkspaceFoldersChangeEvent) {
+    event.added.forEach((workspaceFolder) => {
+      this.changedFilePaths.addPath(workspaceFolder.uri.fsPath);
+    });
+    event.removed.forEach((workspaceFolder) => {
+      this.changedFilePaths.addPath(workspaceFolder.uri.fsPath);
+    });
```
koesie10, line 92: Why are we adding paths for workspace folders which were removed? Shouldn't we be removing those instead?

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * Can configure which changes it watches for, which files are considered
+ * relevant, and what extra data to compute for each file.
+ */
+export abstract class FilePathDiscovery<T extends PathData> extends Discovery {
+  /** The set of known paths we are tracking */
+  protected paths: T[] = [];
+  protected readonly onDidChangePathsEmitter: AppEventEmitter<void>;
+
+  private readonly changedFilePaths = new FilePathSet();
```
koesie10, line 35: I would find it helpful to have a comment here explaining how this variable works since you now need to know how to find that information.

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
```diff
@@ -0,0 +1,190 @@
...
+      }
+    }
+
+    if (pathsUpdated) {
+      this.onDidChangePathsEmitter.fire();
+    }
+  }
+
+  private async handledChangedPath(path: string): Promise<boolean> {
+    if (!(await exists(path)) || !this.pathIsInWorkspace(path)) {
```
koesie10, line 130: `exists` seems to be deprecated, I believe we should be using `pathExists` instead.

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
```diff
@@ -0,0 +1,190 @@
...
+      return this.handledRemovedPath(path);
+    }
+    if ((await lstat(path)).isDirectory()) {
+      return await this.handleChangedDirectory(path);
+    }
```
koesie10, line 135: Are we able to combine the `exists` and `lstat` calls? I believe these will make very similar syscalls, and you should be able to derive if a file doesn't exist from the error returned by `lstat`. The example given in [the Node.js docs](https://nodejs.org/api/fs.html#fsexistspath-callback) is:

```js
  if (err) {
    if (err.code === 'ENOENT') {
      console.error('myfile does not exist');
      return;
    }

    throw err;
  }
```

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
```diff
@@ -0,0 +1,190 @@
...
+    return this.handleChangedFile(path);
+  }
+
+  private pathIsInWorkspace(path: string): boolean {
+    return getOnDiskWorkspaceFoldersObjects().some((workspaceFolder) =>
+      containsPath(workspaceFolder.uri.fsPath, path),
+    );
+  }
+
+  private handledRemovedPath(path: string): boolean {
```
koesie10, line 145:
```suggestion
  private handleRemovedPath(path: string): boolean {
```

extensions/ql-vscode/src/queries-panel/query-discovery.ts
```diff
@@ -1,154 +1,116 @@
...
-      const root = await this.discoverQueriesInWorkspace(workspaceFolder);
-      if (root !== undefined) {
-        rootDirectories.push(root);
+  public buildQueryTree(): Array<FileTreeNode<string>> {
+    const roots = [];
+    for (const workspaceFolder of getOnDiskWorkspaceFoldersObjects()) {
+      const queriesInRoot = this.paths.filter((query) =>
+        containsPath(workspaceFolder.uri.fsPath, query.path),
+      );
+      if (queriesInRoot.length > 0) {
```
koesie10, line 64: Could we invert this condition and do a `continue` if the `queriesInRoot.length === 0`? That would reduce the nesting here.

extensions/ql-vscode/src/queries-panel/query-discovery.ts
```diff
@@ -1,154 +1,116 @@
...
+    return newData.language !== existingData.language;
+  }
+
+  private recomputeAllQueryLanguages() {
+    // All we know is that something has changed in the set of known query packs.
+    // We have no choice but to recompute the language for all queries.
+    for (const query of this.paths) {
+      query.language = this.determineQueryLanguage(query.path);
     }
+    this.onDidChangePathsEmitter.fire();
```
koesie10, line 110: It seems strange to be to be re-using this when none of the paths actually changed. Perhaps we should be thinking about composition rather than inheritance?

extensions/ql-vscode/src/queries-panel/query-pack-discovery.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    if (packs.length === 0) {
+      return undefined;
+    }
+
+    // If the first two packs are from the same directory then look at the filenames
+    if (
+      packs.length >= 2 &&
+      dirname(packs[0].path) === dirname(packs[1].path)
+    ) {
```
koesie10, line 55: Can we make this easier to read by splitting this into two if-conditions and returning early? Something like:

```ts
    if (packs.length === 1) {
      return packs[0].language;
    }

    // If the first two packs are from a different directory, then the first one is the nearest
    if (dirname(packs[0].path) !== dirname(packs[1].path)) {
      return packs[0].language;
    }

    // If the first two packs are from the same directory then look at the filenames
    if (basename(packs[0].path) === FALLBACK_QLPACK_FILENAME) {
      return packs[0].language;
    }

    return packs[1].language;
```

extensions/ql-vscode/test/unit-tests/common/file-path-set.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  describe("addPath / popPath", () => {
+    it("should keep all paths when they don't overlap", () => {
+      const v = new FilePathSet();
+      v.addPath("/foo");
+      v.addPath("/bar");
+      v.addPath("/baz");
+      expect(v.popPath()).toBe("/foo");
+      expect(v.popPath()).toBe("/bar");
+      expect(v.popPath()).toBe("/baz");
+      expect(v.popPath()).toBe(undefined);
```
koesie10, line 26: The documentation of `popPath` doesn't specify an order, while the tests do depend on it. Should we be making explicit that paths are popped in insertion-order, or should we change the tests so that they pop all paths and then check the result?

extensions/ql-vscode/test/vscode-tests/minimal-workspace/common/vscode/file-path-discovery.test.ts
```diff
@@ -0,0 +1,460 @@
...
+  let tmpDirRemoveCallback: (() => void) | undefined;
+
+  let workspaceFolder: WorkspaceFolder;
+  let workspacePath: string;
+  let workspaceFoldersSpy: jest.SpyInstance;
+
+  const onDidCreateFile = new EventEmitter<Uri>();
+  const onDidChangeFile = new EventEmitter<Uri>();
+  const onDidDeleteFile = new EventEmitter<Uri>();
+  let createFileSystemWatcherSpy: jest.SpyInstance;
```
koesie10, line 66:
```suggestion
  let createFileSystemWatcherSpy: jest.SpiedFunction<
    typeof workspace.createFileSystemWatcher
  >;
```

extensions/ql-vscode/test/vscode-tests/utils/expect-arrays-equal.ts
```diff
@@ -0,0 +1,6 @@
+export function expectArraysEqual<T>(actual: T[], expected: T[]) {
+  // Check that all of the expected values are present
+  expect(actual).toEqual(expect.arrayContaining(expected));
+  // Check that no extra un-expected values are present
+  expect(expected).toEqual(expect.arrayContaining(actual));
+}
```
koesie10, line 6: Could we make this a matcher instead located in `test/matchers`? That way, we could call it like `expect(actual).toArrayEquals(expected)`.

koesie10 (commented):
I'm wondering if inheritance is the right solution for these discovery classes. Would it make sense to make these concrete classes and have instances of them in the respective classes instead? That would decouple them, probably make them easier to test and make them less dependent on the specific implementation (like calling `onDidChangePathsEmitter` from within a subclass for something where the set of paths didn't actually change).

extensions/ql-vscode/src/common/query-language.ts
robertbrignull: Yep, we can definitely use the `QueryLanguage` enum in more places in this code.

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
robertbrignull: Good point. I've added more documentation to all of the fields in this class.

For this specific field, what I've written is:
```
  /**
   * The set of file paths that may have changed on disk since the last time
   * refresh was run. Whenever a watcher reports some change to a file we add
   * it to this set, and then during the next refresh we will process all
   * file paths from this set and update our internal state to match whatever
   * we find on disk (i.e. the file exists, doesn't exist, computed data has
   * changed).
   */
```

robertbrignull: Now that I've thought about this now, I'm thinking of removing the custom `FilePathSet` class and instead using a regular `Set`.

The reasons for using a set rather than an array is to remove duplicates. If a file path has changed twice since we last ran `refresh` then we don't need to process that file twice. We can just process it once and deal with whatever state we find on disk.

The reason I went with this `FilePathSet` that avoids having child paths in the set, is that for some of the file watcher events it only reports a directory. I think only the case where this happens is when a directory is added or deleted. So if a directory and a subdirectory are deleted then we'd get "duplicate" events because we could get the same effect by just processing the parent directory.

I'm coming around to thinking it's not worth trying to eliminate the child paths from the set. For the case of individual files being added or deleted we won't get any benefit, because they're files and not directories so they don't have any children. And for directories being added or deleted I think it's rarer and doesn't save us that much compared to checking that the directory exists multiple times.

So I'm going to convert this to a regular `Set`. I think the `FilePathDiscovery` tests should then be good enough to show us it's still working.

---

Small aside. I wasn't sure what the behaviour is of a set when you're iterating over it using `for ... of` but then deleting elements and also maybe new elements get added while iterating. Turns out it handles that fine and it includes the new elements.
```
11:23:03 /Users/robertbrignull/github/vscode-codeql git:robertbrignull/resolve-queries-manual
> cat test.js
const xs = new Set();

xs.add(1);
xs.add(2);

for (const x of xs) {
	xs.delete(x);
	if (x === 1) {
		xs.add(3);
	}
	console.log(x);
}

console.log(xs);

11:23:19 /Users/robertbrignull/github/vscode-codeql git:robertbrignull/resolve-queries-manual
> node test.js
1
2
3
Set(0) {}
```

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
robertbrignull: The idea here is just to register that "something has changed at this path" so then in the refresh we'll look at this path and update our internal state to match whatever we find on disk. I'll add a comment as to why we're doing the same thing for both added and removed paths.

I don't think it's worth it to actually update `this.paths` in this method. Then we'd have to worry about calling the event emitter. Adding to `this.changedFilePaths` and calling `refresh` does all that for us.

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
robertbrignull: Yes, I'll switch this to `pathExists`.

Annoyingly since we're already importing the `fs-extra` version of `exists` it doesn't list it as deprecated, but my understanding is that `fs-extra` is a superset of `fs` so the same reasoning should apply why you don't want to use `exists`.

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
robertbrignull: I'm not sure what's best. I agree you could just call `lstat` and then catch the error if the file doesn't exist. I did consider something like this but it doesn't seem great to use exceptions in this way. I'll have more of a think on it.

I'm also now realising there's a lot of "time-of-check-time-of-read" bugs here, where we check the path exists and then do stuff with it a bit later on. I'll have a think about how severe these are too. It's mostly ok if the discovery process fails so long as we can recover from it and try again and don't corrupt any data in the process.

extensions/ql-vscode/test/unit-tests/common/file-path-set.test.ts
robertbrignull: This file has been deleted, so no need to do anything here now.

robertbrignull:
I've responded to some changes and pushed commits. I'll come back to respond to the other comments later. I'm thinking about the suggestion of using composition instead of inheritance. It might work, but we do need to do some customisation of the `FilePathDiscovery` object, and also need to know when to compute extra data such as query pack languages. I'll see if I can make it work and how clean it is.

extensions/ql-vscode/src/queries-panel/query-discovery.ts
robertbrignull: (Using this thread to talk about the idea of composition vs inheritance)

I don't think composition works here, or at least it doesn't make things easier. Or possibly I'm not understanding what you're suggesting. Could you give a little more info about what you had in mind?

The main issue I'm hitting is that we aren't just tracking paths here but we're also tracking some extra data about each path, such as the language of the query pack. We want to avoid re-computing this extra data when not needed, so we want the computation of the data to be tightly integrated so we only do it when needed, so it makes sense for it to be done inside `FilePathDiscovery`.

So essentially I don't think we can remove all customisation from `FilePathDiscovery`. It could have customisation by abstract methods or function closures passed to the constructor, and either works but I think abstract methods is a bit cleaner.

We could essentially split `QueryDiscovery` into two, so there's one class that is the public interface and has `onDidChangeQueries` and `buildQueryTree`, and that class owns a `QueryFilePathDiscovery extends FilePathDiscovery`. Is this what you meant? It's possible but I don't see any great benefits and it makes accessing fields on other classes more clunky.

---

I wonder perhaps if the real problem is the naming of the fields on `FileTreeDiscovery`? It's not just tracking paths but also data about those paths, so then it sounds odd to just say "the paths have changed". I could rename everything from just "path" to "pathData" or another name you think it clearer?

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * Discovers all files matching a given filter contained in the workspace.
+ *
+ * Scans the whole workspace on startup, and then watches for changes to files
+ * to do the minimum work to keep up with changes.
+ *
+ * Can configure which changes it watches for, which files are considered
+ * relevant, and what extra data to compute for each file.
+ */
+export abstract class FilePathDiscovery<T extends PathData> extends Discovery {
```
robertbrignull, line 29: For the record, I did just try removing the inheritance of `Discovery`, but I don't think it was necessarily better.

The idea is that instead of keeping `changedFilePaths` and the `refresh` method, we just process all events immediately. I think this would work, in that we'd be able to process events and keep up with on-disk state just as well.

The problem was it's harder to test because you can't be sure when it's done processing that event. You'd have to introduce a bunch of arbitrary sleeps, which would be a nightmare for speed and flakiness. One benefit that `Discovery` and the `refresh` method gives is that you can easily wait until a refresh is complete.

It also didn't make much difference to line count. The diff was +40 lines and also -40 lines, so completely net-zero.

extensions/ql-vscode/test/vscode-tests/utils/expect-arrays-equal.ts
robertbrignull: Hmm, I've opted instead to delete `expectArraysEqual` and use sets instead. The real problem I was trying to solve is to not specify the exact ordering in the tests, because the ordering of the `paths` array shouldn't matter. We can solve that by using `expect(new Set(...)).toEqual(new Set(...))`.

I tried making the `paths` field from `FilePathDiscovery` just be a set but sadly it's just so difficult to work with. A lot of the operations you rely on for arrays just aren't there, like a `has` method that uses a predicate, or `filter`.

robertbrignull:
@koesie10, I've made all the changes I plan to make in response to the comments. I'm still very happy to discuss more though. I agree these class are not very pretty or ideal, and the use of the `Discovery` base class is also not perfect. I'm just not sure how to make them better in a practical way.

extensions/ql-vscode/src/queries-panel/query-discovery.ts
koesie10: I didn't have a specific idea for the composition, but I think it might be helpful. I agree that we'd just be trading abstract methods for callbacks passed into the constructor, so it might no make sense here.

Perhaps instead we should have a separate event emitter from the `onDidChangePathsEmitter` emitter? We could then do something like this int he constructor:

```ts
    this.push(
      this.onDidChangePathsEmitter.event(() =>
        this.onDidChangeQueriesEmitter.fire(),
      ),
    );
```

The `onDidChangePathsEmitter` could then be made private and only `onDidChangePathsEmitter.event` could be exposed (probably `protected`). This would at least make the separation between the `FilePathDiscovery` and the `QueryDiscovery` clear.

I think changing the naming of the fields would also help by clarifying that it's data about paths, and not just the paths itself. I'm not sure if you'd still need the separation between the two event emitters in that case. However, I do think we are leaking some implementation details of the `FilePathDiscovery` by directly modifying `this.paths`. Could we instead make a method on `FilePathDiscovery` that allows you to update all path data? Something like this:

```ts
  protected async updateAllPathData(
    callback: (pathData: PathData) => Promise<void>,
  ): Promise<void> {
    await Promise.all(this.paths.map(callback));
    this.onDidChangePathsEmitter.fire();
  }
```

Then the subclass doesn't need to know anything about the event emitter and can just call that method. The `onDidChangePathsEmitter` can be private in that case, and only `onDidChangePathsEmitter.event` can be exposed to subclasses.

robertbrignull: I'll update the name to be `pathData` throughout so it's more clear what's what.

I do very much like the idea of making `pathData` private to `FilePathDiscovery` and adding a `updateAllPathData` method to do that operation. I'll implement these two bits.

At this point I'm not sure we need a second event emitter. Unless it makes things noticeably clearer I'd prefer to stick to just one.

I'm still unsure about going fully composable. I've played around with it a bit, and I think there is potential there but it's also a lot of work to get to a point where you see the benefits. I'd prefer to stay with the current approach for now if that's ok. It doesn't block us from changing approach in the future, and we do have test coverage right now. We could make the interfaces cleaner and make things a bit easier to test but it's a lot of work rearranging things to get there.

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
```diff
@@ -0,0 +1,240 @@
...
+  /** The set of known paths and associated data that we are tracking */
+  private pathData: T[] = [];
+
+  /** Event that fires whenever the contents of `pathData` changes */
+  protected readonly onDidChangePathDataEmitter: AppEventEmitter<void>;
```
koesie10, line 39: Could we make this field private and just expose `this.onDidChangePathDataEmitter.event` as a getter? It seems like we're only using the `.event` part outside of this class.

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
```diff
@@ -0,0 +1,240 @@
...
+    this.onDidChangePathDataEmitter = this.push(new EventEmitter<void>());
+    this.push(
+      workspace.onDidChangeWorkspaceFolders(
+        this.workspaceFoldersChanged.bind(this),
+      ),
+    );
+    this.push(this.watcher.onDidChange(this.fileChanged.bind(this)));
+  }
+
+  protected getPathData(): readonly T[] {
```
koesie10, line 74: I think you can make this return type even stricter by also ensuring you can't update any properties of `T`:

```suggestion
  protected getPathData(): ReadonlyArray<Readonly<T>>
```

extensions/ql-vscode/src/pure/files.ts
```diff
@@ -107,3 +107,14 @@ export async function* walkDirectory(
     }
   }
 }
+
+/**
+ * Error thrown from methods from the 'fs' module.
+ */
+export interface IOError {
```
koesie10, line 114: From [the types in the `fs` module](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/2928e1eb0097809e91ec7a77a845329340d03898/types/node/v16/fs.d.ts#L1042), it seems like there's already a type for this: `NodeJS.ErrnoException`. I don't think we should use it since we're only checking for the `code` field, but it's interesting to see that the type also contains other fields.

koesie10 (approved):
I have a few small suggestions, but none are blocking. The code looks good to me and some simple testing of the queries panel I've done also looks really good and is very fast.

extensions/ql-vscode/src/pure/files.ts
robertbrignull: Ah I didn't spot that type existed. I think that's what we're getting in practice though. When I tried logging one of these exceptions I saw all of those fields.

I'm not sure if we should use that type of not. I think in practice the exception will be of that type, but as you say we currently don't really want all of the fields. So in practice it would just mean modelling unneccessary stuff and make the `isErrnoException` method we'd have to write be more complex.

For now I'll leave it as this custom type with just the `code` field, but I'll make a note about the `NodeJS.ErrnoException` type so anyone who reads this code in the future knows it exists.

robertbrignull:
Done another quick test and it's all working for me still. I'll merge once this PR goes green 🎉 🚢 🇮🇹

extensions/ql-vscode/test/vscode-tests/minimal-workspace/queries-panel/query-discovery.test.ts
```diff
@@ -1,203 +1,196 @@
...
+          new FileTreeLeaf(join(workspacePath, "query.ql"), "query.ql", "java"),
+        ]),
+      ]);
+
+      queryPackDiscoverer.getLanguageForQueryFile = () => QueryLanguage.Python;
+      onDidChangeQueryPacks.fire();
+
+      // Wait for the query discovery to recompute the query languages.
+      // This is async but should complete instantly since it's all in-memory.
+      await sleep(100);
```
robertbrignull, line 178: Sadly the language re-computation is now async, but it should complete instantly since it's just calling the mocked `getLanguageForQueryFile` method in a loop. I think this sleep should be pretty safe and not be a risk of flakiness. I'd love if we could avoid this sleep, but I don't think we can right now.

robertbrignull:
Had to fix a bug in the windows tests where the drive letter was lowercased. I think this is an uncontroversial fix and is limited to just the tests.

PR #2504, opened 2023-06-12
Split helpers.ts file

koesie10:
This is the first part in a series that will split the `helpers.ts` file. Each commit splits one or more methods to a separate file or moves it to a more appropriate existing file. This makes them easier to test and decreases the number of things the `helpers.ts` file is doing.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases/qlpack.ts
+++ b/extensions/ql-vscode/src/databases/qlpack.ts
@@ -1,0 +1,130 @@
+import { window } from "vscode";
+import { glob } from "glob";
+import { basename } from "path";
+import { load } from "js-yaml";
+import { readFile } from "fs-extra";
+import { getQlPackPath } from "../pure/ql";
+import { CodeQLCliServer, QlpacksInfo } from "../codeql-cli/cli";
+import { extLogger } from "../common";
+import { getOnDiskWorkspaceFolders } from "../helpers";
+
+export interface QlPacksForLanguage {
+  /** The name of the pack containing the dbscheme. */
+  dbschemePack: string;
+  /** `true` if `dbschemePack` is a library pack. */
+  dbschemePackIsLibraryPack: boolean;
+  /**
+   * The name of the corresponding standard query pack.
+   * Only defined if `dbschemePack` is a library pack.
+   */
+  queryPack?: string;
+}
+
+interface QlPackWithPath {
+  packName: string;
+  packDir: string | undefined;
+}
+
+async function findDbschemePack(
+  packs: QlPackWithPath[],
+  dbschemePath: string,
+): Promise<{ name: string; isLibraryPack: boolean }> {
+  for (const { packDir, packName } of packs) {
+    if (packDir !== undefined) {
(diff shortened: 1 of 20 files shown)


extensions/ql-vscode/src/databases/qlpack.ts
```diff
@@ -0,0 +1,130 @@
...
+  dbschemes.sort();
+  const dbscheme = dbschemes[0];
+
+  if (dbschemes.length > 1) {
+    void window.showErrorMessage(
+      `Found multiple dbschemes in ${datasetFolder} during quick query; arbitrarily choosing the first, ${dbscheme}, to decide what library to use.`,
+    );
+  }
+  return dbscheme;
+}
```
norascheuch, line 130: It seems this is only used for the quick query and tests? Can we move it somewhere more related to that? Even though it works with the db scheme I think it's fair to move it outside of the databases folder, e.g. the local-queries folder.

extensions/ql-vscode/src/pure/files.ts
```diff
@@ -88,3 +88,26 @@ export async function readDirFullPaths(path: string): Promise<string[]> {
...
+  for await (const d of await opendir(dir)) {
+    const entry = join(dir, d.name);
+    seenFiles.add(entry);
+    if (d.isDirectory()) {
+      yield* walkDirectory(entry);
+    } else if (d.isFile()) {
+      yield entry;
+    }
+  }
+}
```
norascheuch, line 113: It looks like only the codeql cli is using this, can we move it into the codeql-cli folder? If not: since we want to phase out the pure folder we should move it into common at least?

norascheuch (commented):
I have two small comments, otherwise looks good!

extensions/ql-vscode/src/databases/qlpack.ts
koesie10: It's also used in the query resolver (in `language-support/contextual/query-resolver.ts`), so I think it makes most sense to keep it in the `databases` folder. I don't think it's related to `language-support` or `local-queries`, but is about the database itself.

extensions/ql-vscode/src/pure/files.ts
koesie10: I think for now it makes the most sense to put it in this file, just so it's related to everything else. I agree that we should move it into the `common` folder if we're going to phase out the `pure` folder, but I don't think splitting up the file handling functions between two files makes this easier to read and understand.

extensions/ql-vscode/src/databases/qlpack.ts
norascheuch: Ok

extensions/ql-vscode/src/pure/files.ts
norascheuch: That makes, sense, best to move the whole file to common!

koesie10: I agree, but I don't think it's part of this PR, I think deprecating/moving files out of the `pure` directory is a separate item.

PR #2520, opened 2023-06-16
Automatically name extension packs

koesie10:
This will change how extension packs are named in the data extensions editor. Before, the user had to pick a workspace folder and a name for the extension pack. Now, the workspace folder will be picked automatically if we can detect it (i.e. it follows the naming structure we expect), or the user will still need to select it. The extension pack name is always auto-generated based on the database name and the database language.

This adds a new `codeQL.dataExtensions.disableAutoNameExtensionPack` setting to disable this behavior while we are still working on changing how the data extensions editor works.

## Checklist


extensions/ql-vscode/src/config.ts
```diff
@@ -714,7 +714,15 @@ export function showQueriesPanel(): boolean {
 
 const DATA_EXTENSIONS = new Setting("dataExtensions", ROOT_SETTING);
 const LLM_GENERATION = new Setting("llmGeneration", DATA_EXTENSIONS);
+const DISABLE_AUTO_NAME_EXTENSION_PACK = new Setting(
```
starcke, line 717: You want it opt-out instead of opt-in? Any specific reason?

starcke: Especially since it ends up creating a directory with a file in.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
```diff
@@ -303,12 +308,145 @@ async function pickNewExtensionPack(
...
+  const packName = autoNameExtensionPack(name, language);
+  if (!packName) {
+    void showAndLogErrorMessage(
+      logger,
+      `Could not automatically name extension pack for database ${name}`,
+    );
+
+    return undefined;
+  }
+
```
starcke, line 340: I could use a few more comments about what is going on.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
```diff
@@ -303,12 +308,145 @@ async function pickNewExtensionPack(
...
+            existingExtensionPackPaths[0]
+          }: ${getErrorMessage(e)}`,
+        },
+      );
+
+      return undefined;
+    }
+
+    return extensionPack;
+  } else if (existingExtensionPackPaths?.length > 1) {
```
starcke, line 361: Do you need `else` here? Could it just be another `if`?

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
```diff
@@ -303,12 +308,145 @@ async function pickNewExtensionPack(
...
+        fullMessage: `Extension pack ${packName} resolves to multiple paths: ${existingExtensionPackPaths.join(
+          ", ",
+        )}`,
+      },
+    );
+
+    return undefined;
+  }
+
+  const matches = packNameRegex.exec(packName);
```
starcke, line 375: Why is this check here? If we autogenerated the name, shouldnt it match the regex?

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
```diff
@@ -303,12 +308,145 @@ async function pickNewExtensionPack(
...
+  if (!matches?.groups) {
+    void showAndLogErrorMessage(
+      logger,
+      `Extension pack ${packName} does not have a valid name`,
+    );
+
+    return undefined;
+  }
+
+  const unscopedName = matches.groups.name;
```
starcke, line 385: If feel odd to us having auto-generated the name, but then pulling it apart here later.

starcke: Maybe we could have a custom type for pack names and generate that?

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
```diff
@@ -303,12 +308,145 @@ async function pickNewExtensionPack(
...
+      `Directory ${packPath} already exists for extension pack ${packName}`,
+    );
+
+    return undefined;
+  }
+
+  return writeExtensionPack(packPath, packName, language);
+}
+
+async function autoPickWorkspaceFolder(
```
starcke, line 400: A bit more comments would be good.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
```diff
@@ -303,12 +308,145 @@ async function pickNewExtensionPack(
...
+  language: string,
+): Promise<WorkspaceFolder | undefined> {
+  const workspaceFolders = getOnDiskWorkspaceFoldersObjects();
+
+  if (workspaceFolders.length === 1) {
+    return workspaceFolders[0];
+  }
+  const starterWorkspaceFolderForLanguage = workspaceFolders.find(
+    (folder) => folder.name === `codeql-custom-queries-${language}`,
```
starcke, line 409: Is it ok to have a pack inside a pack?

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
```diff
@@ -303,12 +308,145 @@ async function pickNewExtensionPack(
...
+  );
+  if (starterWorkspaceFolderForLanguage) {
+    return starterWorkspaceFolderForLanguage;
```
starcke, line 412: Should we check for the presence of `queries.xml` with the right language?

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
starcke, line 400: Ok for now, but it feels like some of this should live outside of the extension editor, especially if considered together with some of the template work?

starcke (commented):
Looks pretty good, just some smaller comments.

It felt a bit to odd to me that we just created the pack file but asked about the model file name. I would _personally_ have expected the opposite. Later we do want the yml file to be based on the library name, so I am fine with it like this if that is easier.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
koesie10: I've removed it and made the `autoNameExtensionPack` a bit more robust and testable so that it should not be possible for the name to not match the regex.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
koesie10: Yes, that works. I agree that it isn't ideal, but I'm not sure where to create the extension packs otherwise. There isn't a workspace folder in the starter workspace that doesn't have a query pack.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
koesie10: I don't think we need to. The extension pack doesn't actually depend on the folder where it's placed, we just need some place to put it. It'll still work if an extension pack for Java is placed in the `codeql-custom-queries-csharp` workspace folder.

extensions/ql-vscode/src/config.ts
koesie10: I think it's a much nicer experience to have the extension pack created automatically. I've kept the previous option in for more advanced users that want to decide where to place their extension pack. Perhaps we can decide whether to make this a choice the first time the user creates an extension pack by letting them select whether to do it automatically or manually. For now, I think having this flow is fine.

koesie10 (commented):
> It felt a bit to odd to me that we just created the pack file but asked about the model file name. I would personally have expected the opposite. Later we do want the yml file to be based on the library name, so I am fine with it like this if that is easier.

They are essentially independent, so I've done only the extension pack name in this PR to keep the sizes of the PRs manageable. The model file name is in https://github.com/github/vscode-codeql/pull/2521. That means that for now the flow is a bit strange, but it helps to keep the changes minimal.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
adityasharad: I can't think of an immediate thing that would break, but this sounds potentially very confusing, since query packs are not usually meant to contain data extensions themselves, or nested packs (except when compiled, in which case they include their library pack dependencies).
I would rather put an empty placeholder folder now into the starter workspace, and then populate your generated extension packs into that folder.

Also where will extension packs go for repos that are not the starter workspace?

extensions/ql-vscode/src/data-extensions-editor/extension-pack-name.ts
```diff
@@ -0,0 +1,93 @@
...
+export function formatPackName(packName: ExtensionPackName): string {
+  return `${packName.scope}/${packName.name}`;
+}
+
+export function autoNameExtensionPack(
+  name: string,
+  language: string,
+): ExtensionPackName | undefined {
+  let packName = `${name}-${language}`;
+  if (!packName.includes("/")) {
```
aeisenberg, line 21: What if `packName` includes 2 or more slashes?

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
```diff
@@ -293,31 +294,172 @@ async function pickNewExtensionPack(
...
+        )} resolves to multiple paths: ${existingExtensionPackPaths.join(
+          ", ",
+        )}`,
+      },
+    );
+
+    return undefined;
+  }
+
+  const packPath = join(workspaceFolder.uri.fsPath, packName.name);
```
aeisenberg, line 382: Does it make sense to put the pack in `github/codeql/extensions`? Or at least have an option to do that? This way, it will automatically be picked up during code scanning.

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
aeisenberg: Possibly, we can add `.github/codeql/extensions` as a workspace folder (possibly empty) and put the extension packs there. This folder is the location where internal extensions are kept. Even though it's not directly useful in the starter repo, this folder is recognized for code scanning runs.

And, I agree with Aditya. I don't see any technical problem of nesting packs in this case, but it is confusing.

extensions/ql-vscode/test/unit-tests/data-extensions-editor/extension-pack-name.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      language: "csharp",
+      expected: "a/b-csharp",
+    },
+    {
+      name: "-a-/b",
+      language: "ruby",
+      expected: "a/b-ruby",
+    },
+    {
+      name: "a/b--d--e-d-",
```
aeisenberg, line 70: What if `name` has multiple slashes? And what about special chars?

extensions/ql-vscode/src/data-extensions-editor/extension-pack-name.ts
koesie10: In that case it will be handled by the combination of splitting it by a `/` and taking everything after the first slash as the name: `sanitizedParts.slice(1).join("-")`. This test covers that functionality: https://github.com/github/vscode-codeql/blob/3323fd4e3bd4f2cba5755563d3a74277ab5fcc3d/extensions/ql-vscode/test/unit-tests/data-extensions-editor/extension-pack-name.test.ts#L39-L43

extensions/ql-vscode/src/data-extensions-editor/extension-pack-picker.ts
koesie10: I've now changed it to put all extension packs in `.github/codeql/extensions`, unless the extension pack already exists. This should remove the nested extension packs and make it clearer where extension packs will go in almost all cases. It will now automatically add a new workspace folder for `.github/codeql/extensions`, unless the user is in a single-root workspace. In that case, it will be added in that workspace folder directly in the `.github/codeql/extensions` directory.

Previously, it would try to automatically determine the workspace folder by checking for the `-<language>` suffix, and otherwise it would just ask the user.

extensions/ql-vscode/src/data-extensions-editor/extensions-workspace-folder.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+function getRootWorkspaceDirectory(): Uri | undefined {
+  // If there is a valid workspace file, just use its directory as the directory for the extensions
+  const workspaceFile = workspace.workspaceFile;
+  if (workspaceFile?.scheme === "file") {
+    return Uri.joinPath(workspaceFile, "..");
+  }
+
+  const workspaceFolders = getOnDiskWorkspaceFoldersObjects();
```
aeisenberg, line 29: Does `getOnDiskWorkspaceFoldersObjects` return only `file:` URIs? If not, you will need to filter out the all the other ones.

extensions/ql-vscode/src/data-extensions-editor/extensions-workspace-folder.ts
```diff
@@ -0,0 +1,143 @@
...
+  }, getAncestors(workspaceFolders[0].uri));
+
+  if (commonRoot.length === 0) {
+    return undefined;
+  }
+
+  // The path closest to the workspace folders is the last element of the common root
+  const commonRootUri = commonRoot[commonRoot.length - 1];
+
+  // If we are at the root of the filesystem, we can't go up any further and there's something
```
aeisenberg, line 56: This will be reasonably common in codeql projets. There's a `quick-query` command that creates a temp folder and adds it to the workspace. Any workspace that has a quick query folder will only have a common root at `/`.

Perhaps you will want to filter folders in the temp directory?

aeisenberg: Also, maybe look for a `.git` folder?

extensions/ql-vscode/src/data-extensions-editor/extensions-workspace-folder.ts
koesie10: Yes, that's exactly what `getOnDiskWorkspaceFoldersObjects` does, so this should only return workspace folders which actually exist on-disk.

extensions/ql-vscode/src/data-extensions-editor/extensions-workspace-folder.ts
koesie10: I've added both filtering out temp directories and looking for the `.git` folder. If it can't find a common ancestor, it will now try to find the `.git` folder closest to one of the workspace folders and pick that.

aeisenberg (approved):
Nice.

Not necessary for this PR, but the heuristic for choosing a location for extension packs is quite complicated. It would be good to see it documented somewhere. (Either in code as comments, or in a separate document.)

PR #2521, opened 2023-06-16
Automatically create different model files per library

koesie10:
This will remove the user input for a model file and will instead create 1 model file per library (JAR/DLL). The model filename will be based on the JAR/DLL name, but will remove the version number and the file extension. It will also normalize the name.

These files will be created automatically, and the editor now also reads in all files contained in an extension pack to read the modeled methods. This could result in duplicates if the user has created a different file to contain the same modeled methods, but this is an edge-case that we're explicitly not handling.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/interface-types.ts
+++ b/extensions/ql-vscode/src/common/interface-types.ts
@@ -532,6 +532,7 @@ export interface OpenExtensionPackMessage {
 
 export interface OpenModelFileMessage {
   t: "openModelFile";
+  library: string;
 }
 
 export interface SaveModeledMethods {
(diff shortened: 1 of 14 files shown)


starcke:
Before digging in I had a few thoughts/questions:
- This seems to make changes in the same place as https://github.com/github/vscode-codeql/pull/2521, so it might make sense to wait for 2521 before I look at the details here.
- The naming based on e.g. jar and dll doesnt quite match what we have currently (in https://github.com/github/codeql/tree/main/java/ql/lib/ext and https://github.com/github/codeql/tree/main/csharp/ql/lib/ext) where the individual file names are packages (and there is also a `.model.` part of the name).

starcke (approved):
I think this looks good. It is of course a bit iffy with the version number, but this is definately good for version 1.

PR #2523, opened 2023-06-19
Display different message when queries are loading vs no queries found

robertbrignull:
Displays a "loading" message while query discovery has not yet finished, and a separate message when no queries have been found. This should help avoid confusing users by saying there are no queries when really the discovery is just in progress.

The approach I've taken is to make the welcome message the "loading" message, and display a custom text node for the "no queries" case. This helps avoid the effect seen in https://github.com/github/vscode-codeql/pull/2523#issuecomment-1597245642 where it flips between the two options before the extension has started initialising.

For the datatypes I opt to create subclasses for `QueryTreeQueryItem` and `QueryTreeTextItem`, where `QueryTreeQueryItem` is identical to the old `QueryTreeViewItem`. We could maybe split this up further into "node" and "leaf" classes, but I haven't done this here.

I also fix a bug where it wouldn't update the tree view in the case that the initial refresh found no queries. This bug just wasn't visible before because it would display the welcome message anyway.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/local-queries/local-queries.ts
+++ b/extensions/ql-vscode/src/local-queries/local-queries.ts
@@ -48,10 +48,7 @@ import { SkeletonQueryWizard } from "../skeleton-query-wizard";
 import { LocalQueryRun } from "./local-query-run";
 import { createMultiSelectionCommand } from "../common/vscode/selection-commands";
 import { findLanguage } from "../codeql-cli/query-language";
-import {
-  QueryTreeQueryItem,
-  QueryTreeViewItem,
-} from "../queries-panel/query-tree-view-item";
+import { QueryTreeViewItem } from "../queries-panel/query-tree-view-item";
 
 interface DatabaseQuickPickItem extends QuickPickItem {
   databaseItem: DatabaseItem;
@@ -133,7 +130,7 @@ export class LocalQueries extends DisposableObject {
   private async runQueryFromQueriesPanel(
     queryTreeViewItem: QueryTreeViewItem,
   ): Promise<void> {
-    if (queryTreeViewItem instanceof QueryTreeQueryItem) {
+    if (queryTreeViewItem.path !== undefined) {
       await this.runQuery(Uri.file(queryTreeViewItem.path));
     }
   }
@@ -141,12 +138,13 @@ export class LocalQueries extends DisposableObject {
   private async runQueriesFromQueriesPanel(
     queryTreeViewItem: QueryTreeViewItem,
   ): Promise<void> {
-    if (queryTreeViewItem instanceof QueryTreeQueryItem) {
-      const uris = queryTreeViewItem.children.map((child) =>
-        Uri.file(child.path),
-      );
-      await this.runQueries(uris);
+    const uris = [];
+    for (const child of queryTreeViewItem.children) {
+      if (child.path !== undefined) {
+        uris.push(Uri.file(child.path));
+      }
     }
+    await this.runQueries(uris);
   }
 
   private async runQuery(uri: Uri | undefined): Promise<void> {
--- a/extensions/ql-vscode/src/local-queries/local-queries.ts
+++ b/extensions/ql-vscode/src/local-queries/local-queries.ts
@@ -48,7 +48,10 @@ import { SkeletonQueryWizard } from "../skeleton-query-wizard";
 import { LocalQueryRun } from "./local-query-run";
 import { createMultiSelectionCommand } from "../common/vscode/selection-commands";
 import { findLanguage } from "../codeql-cli/query-language";
-import type { QueryTreeViewItem } from "../queries-panel/query-tree-view-item";
+import {
+  QueryTreeQueryItem,
+  QueryTreeViewItem,
+} from "../queries-panel/query-tree-view-item";
 
 interface DatabaseQuickPickItem extends QuickPickItem {
   databaseItem: DatabaseItem;
@@ -130,16 +133,20 @@ export class LocalQueries extends DisposableObject {
   private async runQueryFromQueriesPanel(
     queryTreeViewItem: QueryTreeViewItem,
   ): Promise<void> {
-    await this.runQuery(Uri.file(queryTreeViewItem.path));
+    if (queryTreeViewItem instanceof QueryTreeQueryItem) {
+      await this.runQuery(Uri.file(queryTreeViewItem.path));
+    }
   }
 
   private async runQueriesFromQueriesPanel(
     queryTreeViewItem: QueryTreeViewItem,
   ): Promise<void> {
-    const uris = queryTreeViewItem.children.map((child) =>
-      Uri.file(child.path),
-    );
-    await this.runQueries(uris);
+    if (queryTreeViewItem instanceof QueryTreeQueryItem) {
+      const uris = queryTreeViewItem.children.map((child) =>
+        Uri.file(child.path),
+      );
+      await this.runQueries(uris);
+    }
   }
 
   private async runQuery(uri: Uri | undefined): Promise<void> {
--- a/extensions/ql-vscode/src/local-queries/local-queries.ts
+++ b/extensions/ql-vscode/src/local-queries/local-queries.ts
@@ -48,7 +48,7 @@ import { SkeletonQueryWizard } from "../skeleton-query-wizard";
 import { LocalQueryRun } from "./local-query-run";
 import { createMultiSelectionCommand } from "../common/vscode/selection-commands";
 import { findLanguage } from "../codeql-cli/query-language";
-import { QueryTreeViewItem } from "../queries-panel/query-tree-view-item";
+import type { QueryTreeViewItem } from "../queries-panel/query-tree-view-item";
 
 interface DatabaseQuickPickItem extends QuickPickItem {
   databaseItem: DatabaseItem;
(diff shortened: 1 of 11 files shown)


robertbrignull:
Screen recording:

https://github.com/github/vscode-codeql/assets/3749000/56bc0d81-31e2-49e0-a74c-502b2a7a6772



Unfortunately it still displays the welcome message for a bit while the extension is starting, before it calls out tree data provider which can provider the loading message.

I'm considering flipping the behaviours so that we use the welcome message to say that things are in progress, and then a custom message when there were no queries found.

robertbrignull:
I've updated this PR as described earlier:
> I'm considering flipping the behaviours so that we use the welcome message to say that things are in progress, and then a custom message when there were no queries found.

Should be ready for review and testing now.

extensions/ql-vscode/test/vscode-tests/minimal-workspace/queries-panel/query-tree-data-provider.test.ts
```diff
@@ -4,18 +4,35 @@ import {
...
       const dataProvider = new QueryTreeDataProvider({
-        buildQueryTree: () => [],
+        buildQueryTree: () => undefined,
         onDidChangeQueries: jest.fn(),
       });
 
       expect(dataProvider.getChildren()).toEqual([]);
     });
 
+    it("returns a explanatory message when there are no queries", async () => {
```
norascheuch, line 23:
```suggestion
    it("returns an explanatory message when there are no queries", async () => {
```

extensions/ql-vscode/src/queries-panel/query-tree-view-item.ts
```diff
@@ -1,11 +1,17 @@
 import * as vscode from "vscode";
 
-export class QueryTreeViewItem extends vscode.TreeItem {
+export abstract class QueryTreeViewItem extends vscode.TreeItem {
```
norascheuch, line 3: Why do we need the abstract class?

extensions/ql-vscode/src/queries-panel/query-tree-view-item.ts
```diff
@@ -24,3 +30,9 @@ export class QueryTreeViewItem extends vscode.TreeItem {
     }
   }
 }
+
+export class QueryTreeTextItem extends QueryTreeViewItem {
+  constructor(text: string) {
+    super(text);
+  }
+}
```
norascheuch, line 38: Why is this type necessary? We can make the path property of the old `QueryTreeViewItem` optional and reuse that type?

extensions/ql-vscode/src/queries-panel/query-tree-data-provider.ts
```diff
@@ -68,8 +83,10 @@ export class QueryTreeDataProvider
     if (!item) {
       // We're at the root.
       return this.queryTreeItems;
-    } else {
+    } else if (item instanceof QueryTreeQueryItem) {
       return item.children;
+    } else {
+      return [];
```
norascheuch, line 89: We won't need this change if we keep the QueryTreeViewItem-Type as is (and all other of these checks)

norascheuch (commented):
I'm not sure I understand why we need the added complexity of a QueryTreeQueryItem and QueryTreeTextItem Type. 

When making the `path` optional we should be able to reuse the existing type?

extensions/ql-vscode/src/queries-panel/query-tree-view-item.ts
robertbrignull: The reason for making this class abstract, even though it doesn't have any abstract methods, is that there shouldn't be any need to construct an instance of it directly. You should always be using one of the subclasses.

By making it abstract we guide future developers to make the right choice by using one of the subclasses and not accidentally use this class for something it wasn't designed for.

extensions/ql-vscode/src/queries-panel/query-tree-view-item.ts
robertbrignull: The reason for making subclasses is precisely so that the `path` property can remain non-optional.

The idea I'm going for is that as far as is practical it should be impossible to construct a class an in invalid state. If we kept it all as one class with all fields marked as optional, it would be possible to construct an instance where fields were undefined but values are expected by other areas of code. For example, code like https://github.com/github/vscode-codeql/blob/d290b56649527d55695acae73d5df3abb5b7a35c/extensions/ql-vscode/src/local-queries/local-queries.ts#L146 expects `path` to be defined for every node.

Also, if `QueryTreeTextItem` did have more fields or any implementation details, that complexity would not affect `QueryTreeQueryItem`. We keep code separate so each bit implements only what it needs, instead of having one class that tries to cover every case.

robertbrignull: I see that [`DbTreeViewItem`](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/databases/ui/db-tree-view-item.ts) addresses this same concern in a different way.

There, there is one class that covers all cases, so over half of the fields are optional. However, the class constructor is not used outside of that file and there are a large number of exported `createDbTreeViewItemXXX` methods that cover the different cases.

That solves the problem of creating valid instances since stricter typing can be used on the exported methods. It doesn't solve the problem when reading data since then you'll always dealing with a `DbTreeViewItem` so you have to assume data is valid and fields are defined when they should be. However it does avoid the added complexity of having more classes and inheritance.

Overall, I don't mind going for that strategy instead. I think either works and it's not clear that one is strictly better than the other, but being consistent is nice.

extensions/ql-vscode/src/queries-panel/query-tree-view-item.ts
norascheuch: Of course, that makes sense, thanks!

extensions/ql-vscode/src/queries-panel/query-tree-view-item.ts
robertbrignull: I think consistency is important. I've converted it to match the style from the queries panel using exported functions to create `QueryTreeViewItem`s.

extensions/ql-vscode/src/local-queries/local-queries.ts
```diff
@@ -48,7 +48,7 @@ import { SkeletonQueryWizard } from "../skeleton-query-wizard";
 import { LocalQueryRun } from "./local-query-run";
 import { createMultiSelectionCommand } from "../common/vscode/selection-commands";
 import { findLanguage } from "../codeql-cli/query-language";
-import type { QueryTreeViewItem } from "../queries-panel/query-tree-view-item";
+import { QueryTreeViewItem } from "../queries-panel/query-tree-view-item";
```
norascheuch, line 51: Is there a reason you removed the `type` import specification?

extensions/ql-vscode/src/queries-panel/query-pack-discovery.ts
```diff
@@ -37,7 +37,7 @@ export class QueryPackDiscovery extends FilePathDiscovery<QueryPack> {
    */
   public getLanguageForQueryFile(queryPath: string): QueryLanguage | undefined {
     // Find all packs in a higher directory than the query
-    const packs = this.getPathData().filter((queryPack) =>
+    const packs = (this.getPathData() || []).filter((queryPack) =>
```
norascheuch, line 40: It's a bit confusing for me to run a filter over an empty array, but checking `getPathData` first and having another if condition is not really better I suppose?

extensions/ql-vscode/src/queries-panel/query-tree-view-item.ts
```diff
@@ -3,24 +3,43 @@ import * as vscode from "vscode";
...
-        arguments: [vscode.Uri.file(path)],
-      };
-    } else {
-      this.collapsibleState = vscode.TreeItemCollapsibleState.Collapsed;
-      this.contextValue = "queryFolder";
-    }
   }
 }
+
+export function createQueryTreeNodeItem(
```
norascheuch, line 13: I think I would personally prefer using the domain names here, namely QueryTreeFolderItem and QueryTreeFileItem. Right now I have to translate 'node' and 'leaf' in my head to folder and file to have the correct image in my head. However, I don't have a CS background and could imagine it doesn't make sense to call it something other than node and leaf? For me folder and file is just more descriptive.

extensions/ql-vscode/src/queries-panel/query-tree-view-item.ts
```diff
@@ -3,24 +3,43 @@ import * as vscode from "vscode";
...
+  children: QueryTreeViewItem[],
+): QueryTreeViewItem {
+  const item = new QueryTreeViewItem(name, path, children);
+  item.tooltip = path;
+  item.collapsibleState = vscode.TreeItemCollapsibleState.Collapsed;
+  item.contextValue = "queryFolder";
+  return item;
+}
+
+export function createQueryTreeLeafItem(
```
norascheuch, line 25: see comment above re naming

norascheuch (approved):
Some minor comments otherwise looks good. Thanks for introducing these types!

extensions/ql-vscode/src/local-queries/local-queries.ts
robertbrignull: Good spot. Earlier in the lifetime of this PR I had to convert the `type` import because I used `QueryTreeViewItem` in a `instanceof` expression, and that requires importing the type at runtime instead of just at compiletime. But that has been removed now so this can stay as a `type` import.

extensions/ql-vscode/src/queries-panel/query-tree-view-item.ts
robertbrignull: I don't have a strong preference between the names. I was just picking one and my mind goes towards nodes and leaves because we're making `TreeView`/`QueryTree`.

I do now see we're using "queryFolder" and "queryFile" for the context values. I'll change the function names to match those.

PR #2525, opened 2023-06-19
Remove ProgressCallback / CancellationToken arguments where they aren't used

robertbrignull:
The idea here is to remove `ProgressCallback` or `CancellationToken` arguments where they ultimately aren't used. This involves tracing the execution into the query server. I did this by looking in the CodeQL CLI repository in the `EvaluationServer` class, to see if either of those arguments are used in the implementation.

The affected query server commands are:
- `clearCacheInDatabase`
- `registerDatabase`
- `deregisterDatabase`

My understanding is that the state of whether or not a query server command utilises progress reporting is unlikely to change in the foreseeable future. So these assumptions about whether commands use these arguments or not shouldn't go out of date too quickly.

Where they are not used, I've removed them from the `QueryRunner` interface, and then followed the chain of removing unused variables as far as possible.

I've tried testing this by running these three operations, and everything seems to be working just as before. In some cases we intentionally use `withProgress` but then don't make any progress updates to it. In the UI this works and it looks like a progress bar that animates to show activity but doesn't show a particular percentage.

## Checklist


aeisenberg (approved):
Thanks. This makes sense to me and I think is a good cleanup.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -354,16 +353,12 @@ export class DataExtensionsEditorView extends AbstractWebview<
...
-      () =>
-        this.showProgress({
-          step: 3900,
-          maxStep: 4000,
-          message: "Removing temporary database",
-        }),
-      tokenSource.token,
-      database,
-    );
+    await this.showProgress({
```
robertbrignull, line 356: Upon re-reviewing my own PR, I think we should keep this one. I'm not 100% sure if it worked before, but the intention was to do a manual progress update to show that we're removing the database.

PR #2529, opened 2023-06-20
Ensure errors for one path don't stop discovery of other paths

robertbrignull:
Something I've noticed in the `FilePathDiscovery` logic is that one error will abort the whole `discovery` method. It'll then recover the next time that `discover` is called, but this isn't ideal because it might not happen for a while until another path changes.

Instead we should catch any errors and then just carry on. This one an error processing one path won't stop us from processing the other paths and the query panel can still display those other paths.

I don't think it's worth trying to retry individual paths, so I think it's fine for the path to be removed from `this.changedFilePaths`. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
+++ b/extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
@@ -152,9 +152,19 @@ export abstract class FilePathDiscovery<T extends PathData> extends Discovery {
   protected async discover() {
     let pathsUpdated = false;
     for (const path of this.changedFilePaths) {
-      this.changedFilePaths.delete(path);
-      if (await this.handleChangedPath(path)) {
-        pathsUpdated = true;
+      try {
+        this.changedFilePaths.delete(path);
+        if (await this.handleChangedPath(path)) {
+          pathsUpdated = true;
+        }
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
```diff
@@ -152,9 +152,19 @@ export abstract class FilePathDiscovery<T extends PathData> extends Discovery {
...
     for (const path of this.changedFilePaths) {
-      this.changedFilePaths.delete(path);
-      if (await this.handleChangedPath(path)) {
-        pathsUpdated = true;
+      try {
+        this.changedFilePaths.delete(path);
+        if (await this.handleChangedPath(path)) {
+          pathsUpdated = true;
+        }
+      } catch (e) {
```
aeisenberg, line 160: Log the error too?

aeisenberg (approved):
Looks good. Small, non-blocking suggestion.

extensions/ql-vscode/src/common/vscode/file-path-discovery.ts
robertbrignull: Good point. We should include the error message too.

PR #2534, opened 2023-06-21
Queries Panel: run local query from panel

norascheuch:
Adds a 'run'-icon when hovering over a file in the queries panel when a local database is selected. Runs the query when clicked.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -406,6 +406,10 @@
         "title": "CodeQL: Run Query on Selected Database"
       },
       {
+        "command": "codeQL.runLocalQueryFromQueriesPanel",
+        "title": "CodeQL: Run Query on Selected Database"
+      },
+      {
         "command": "codeQL.debugQuery",
         "title": "CodeQL: Debug Query"
       },
@@ -1163,6 +1167,10 @@
         {
           "command": "codeQL.runQuery",
           "when": "resourceLangId == ql && resourceExtname == .ql"
+        },
+        {
+          "command": "codeQL.runLocalQueryFromQueriesPanel",
+          "when": "false"
         },
         {
           "command": "codeQL.runQueryContextEditor",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -1103,7 +1103,7 @@
         {
           "command": "codeQLQueries.runLocalQueryContextInline",
           "group": "inline",
-          "when": "view == codeQLQueries && viewItem == queryFile"
+          "when": "view == codeQLQueries && viewItem == queryFile && codeQL.currentDatabaseItem"
         },
         {
           "command": "codeQLTests.showOutputDifferences",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -1103,7 +1103,7 @@
         {
           "command": "codeQL.runLocalQueryFromQueriesPanel",
           "group": "inline",
-          "when": "view == codeQLQueries && viewItem == queryFile"
+          "when": "view == codeQLQueries && viewItem == queryFile && codeQL.currentDatabaseItem"
         },
         {
           "command": "codeQLTests.showOutputDifferences",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -502,7 +502,7 @@
         "title": "CodeQL: Copy Version Information"
       },
       {
-        "command": "codeQL.runLocalQueryFromQueriesPanel",
+        "command": "codeQLQueries.runLocalQueryFromQueriesPanel",
         "title": "Run local query",
         "icon": "$(run)"
       },
@@ -1101,7 +1101,7 @@
           "when": "viewItem == remoteResultsItem"
         },
         {
-          "command": "codeQL.runLocalQueryFromQueriesPanel",
+          "command": "codeQLQueries.runLocalQueryFromQueriesPanel",
           "group": "inline",
           "when": "view == codeQLQueries && viewItem == queryFile && codeQL.currentDatabaseItem"
         },
@@ -1165,7 +1165,7 @@
           "when": "resourceLangId == ql && resourceExtname == .ql"
         },
         {
-          "command": "codeQL.runLocalQueryFromQueriesPanel",
+          "command": "codeQLQueries.runLocalQueryFromQueriesPanel",
           "when": "false"
         },
         {
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -502,6 +502,11 @@
         "title": "CodeQL: Copy Version Information"
       },
       {
+        "command": "codeQLQueries.runLocalQueryContextInline",
+        "title": "Run local query",
+        "icon": "$(run)"
+      },
+      {
         "command": "codeQLVariantAnalysisRepositories.openConfigFile",
         "title": "Open database configuration file",
         "icon": "$(json)"
@@ -1096,6 +1101,11 @@
           "when": "viewItem == remoteResultsItem"
         },
         {
+          "command": "codeQLQueries.runLocalQueryContextInline",
+          "group": "inline",
+          "when": "view == codeQLQueries && viewItem == queryFile"
+        },
+        {
           "command": "codeQLTests.showOutputDifferences",
           "group": "qltest@1",
           "when": "viewItem == testWithSource"
@@ -1273,6 +1283,10 @@
         {
           "command": "codeQL.openDataExtensionsEditor",
           "when": "config.codeQL.canary && config.codeQL.dataExtensions.editor"
+        },
+        {
+          "command": "codeQLQueries.runLocalQueryContextInline",
+          "when": "false"
         },
         {
           "command": "codeQLVariantAnalysisRepositories.openConfigFile",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -406,10 +406,6 @@
         "title": "CodeQL: Run Query on Selected Database"
       },
       {
-        "command": "codeQLQueries.runLocalQueryFromQueriesPanel",
-        "title": "Run Query on Selected Database"
-      },
-      {
         "command": "codeQL.debugQuery",
         "title": "CodeQL: Debug Query"
       },
@@ -506,7 +502,7 @@
         "title": "CodeQL: Copy Version Information"
       },
       {
-        "command": "codeQLQueries.runLocalQueryContextInline",
+        "command": "codeQL.runLocalQueryFromQueriesPanel",
         "title": "Run local query",
         "icon": "$(run)"
       },
@@ -1105,7 +1101,7 @@
           "when": "viewItem == remoteResultsItem"
         },
         {
-          "command": "codeQLQueries.runLocalQueryContextInline",
+          "command": "codeQL.runLocalQueryFromQueriesPanel",
           "group": "inline",
           "when": "view == codeQLQueries && viewItem == queryFile && codeQL.currentDatabaseItem"
         },
@@ -1169,7 +1165,7 @@
           "when": "resourceLangId == ql && resourceExtname == .ql"
         },
         {
-          "command": "codeQLQueries.runLocalQueryFromQueriesPanel",
+          "command": "codeQL.runLocalQueryFromQueriesPanel",
           "when": "false"
         },
         {
@@ -1291,10 +1287,6 @@
         {
           "command": "codeQL.openDataExtensionsEditor",
           "when": "config.codeQL.canary && config.codeQL.dataExtensions.editor"
-        },
-        {
-          "command": "codeQLQueries.runLocalQueryContextInline",
-          "when": "false"
         },
         {
           "command": "codeQLVariantAnalysisRepositories.openConfigFile",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -1103,7 +1103,7 @@
         {
           "command": "codeQL.runLocalQueryFromQueriesPanel",
           "group": "inline",
-          "when": "view == codeQLQueries && viewItem == queryFile && codeQL.currentDatabaseItem"
+          "when": "view == codeQLQueries && viewItem == queryFile"
         },
         {
           "command": "codeQLTests.showOutputDifferences",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -406,8 +406,8 @@
         "title": "CodeQL: Run Query on Selected Database"
       },
       {
-        "command": "codeQL.runLocalQueryFromQueriesPanel",
-        "title": "CodeQL: Run Query on Selected Database"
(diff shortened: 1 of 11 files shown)


extensions/ql-vscode/src/queries-panel/queries-panel.ts
```diff
@@ -14,4 +20,19 @@
...
+    };
+  }
+
+  private async runLocalQuery(
+    queryTreeViewItem: QueryTreeViewItem,
+  ): Promise<void> {
+    await this.app.commands.execute(
+      "codeQL.runQuery",
+      vscode.Uri.parse(queryTreeViewItem.path),
+    );
```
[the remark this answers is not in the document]
norascheuch: not relevant anymore

extensions/ql-vscode/src/queries-panel/queries-panel.ts
```diff
@@ -14,4 +20,19 @@ export class QueriesPanel extends DisposableObject {
...
+      "codeQLQueries.runLocalQueryContextInline": this.runLocalQuery.bind(this),
+    };
+  }
+
+  private async runLocalQuery(
+    queryTreeViewItem: QueryTreeViewItem,
+  ): Promise<void> {
+    await this.app.queryServerCommands.execute(
+      "codeQLQueries.runLocalQueryFromQueriesPanel",
+      vscode.Uri.parse(queryTreeViewItem.path),
```
robertbrignull, line 35: It would be safer and more correct to use `Uri.file` here:
```suggestion
      vscode.Uri.file(queryTreeViewItem.path),
```

extensions/ql-vscode/src/queries-panel/queries-module.ts
```diff
@@ -6,12 +6,42 @@ import { DisposableObject } from "../common/disposable-object";
...
+  ): QueriesModule {
+    const queriesModule = new QueriesModule(app);
+    app.subscriptions.push(queriesModule);
+
+    queriesModule.initialize(app, cliServer);
+    return queriesModule;
+  }
+
+  public getCommands(): QueriesPanelCommands {
+    if (!isCanary() || !showQueriesPanel()) {
```
robertbrignull, line 30: If we're going to use this check in multiple places, can we pull it out to a function so we don't duplicate the logic or the comment. Maybe into a `shouldInitializeQueriesPanel` method.

Sadly the comment is slightly outdated already because of the merge of https://github.com/github/vscode-codeql/pull/2528, so that's a perfect example of why it's good to avoid duplication.

extensions/ql-vscode/src/queries-panel/queries-module.ts
```diff
@@ -6,12 +6,42 @@ import { DisposableObject } from "../common/disposable-object";
...
+      // Currently, we only want to expose the new panel when we are in development and canary mode
+      // and the developer has enabled the "Show queries panel" flag.
+      return {} as any as QueriesPanelCommands;
```
robertbrignull, line 33: I don't particularly love that we have to do this cast. We shouldn't have to force the type system, so I think it shows that things perhaps aren't quite right at a high-level.

Earlier today we were discussing always registering the command and then only showing the panel if the feature flag is enabled. That would mean something like https://github.com/github/vscode-codeql/commit/42886c69aed49878d66a4e298e0a6829de00cee7, though perhaps I didn't make it clear enough what I meant. What do you think? If you don't think that's better then that's ok, but I'd be interested to hear the reasoning.

Or as another alternative would it be better providing dummy implementations that just log an error? That way if one of these commands does accidentally become callable then the user will get a nicer error message.

extensions/ql-vscode/src/common/app.ts
```diff
@@ -18,6 +18,7 @@ export interface App {
   readonly workspaceState: Memento;
   readonly credentials: Credentials;
   readonly commands: AppCommandManager;
+  readonly queryServerCommands: QueryServerCommandManager;
```
robertbrignull, line 21: Should get a second opinion about exposing the `queryServerCommands` field on the app interface. @koesie10, do you have any thoughts? Is there a reason not to, or did we just not need it until this point?

robertbrignull (commented):
I think it's along the right lines, though I have a couple of small comments and I'd like to get another opinion on things since I suggested some parts.

extensions/ql-vscode/src/queries-panel/queries-panel.ts
shati-patel: Good point! The "Run query" button currently fails for me (on Windows) with:

image


Just tried @robertbrignull's suggestion though, and can confirm that it works 🎉

shati-patel (commented):
Looks good overall 😍 Tested locally and left some comments too 📝

extensions/ql-vscode/src/queries-panel/queries-module.ts
norascheuch: Ah, yes. My thought was to initialize as little as possible. I agree that the cast is a code smell, but to push it all to the panel and initialize that didn't sit well with me either.

extensions/ql-vscode/src/queries-panel/queries-module.ts
norascheuch: 👍 I extracted the check into a method. We can discuss where to push the guard in your other comment.

extensions/ql-vscode/src/queries-panel/queries-panel.ts
norascheuch: 👍

extensions/ql-vscode/src/queries-panel/queries-module.ts
```diff
@@ -6,12 +6,42 @@ import { DisposableObject } from "../common/disposable-object";
...
+    app.subscriptions.push(queriesModule);
+
+    queriesModule.initialize(app, cliServer);
+    return queriesModule;
+  }
+
+  public getCommands(): QueriesPanelCommands {
+    if (!isCanary() || !showQueriesPanel()) {
+      // Currently, we only want to expose the new panel when we are in development and canary mode
+      // and the developer has enabled the "Show queries panel" flag.
```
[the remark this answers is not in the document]
norascheuch: 👍 extracted this out to a method

extensions/ql-vscode/src/common/app.ts
koesie10: I think we just haven't needed it until this point. I don't see a particular reason not to expose it, it's coming from the same file so we aren't importing anything extra.

extensions/ql-vscode/src/common/commands.ts
```diff
@@ -12,6 +12,7 @@ import type {
   VariantAnalysisScannedRepositoryResult,
 } from "../variant-analysis/shared/variant-analysis";
 import type { QLDebugConfiguration } from "../debugger/debug-configuration";
+import { QueryTreeViewItem } from "../queries-panel/query-tree-view-item";
```
koesie10, line 15: This should be a `type` import to ensure this file remains somewhat independent of `vscode` and only needs the types from `vscode`, rather than the complete `vscode` module:

```suggestion
import type { QueryTreeViewItem } from "../queries-panel/query-tree-view-item";
```

extensions/ql-vscode/src/common/commands.ts
```diff
@@ -129,6 +130,7 @@ export type LocalQueryCommands = {
   "codeQL.runQueryOnMultipleDatabasesContextEditor": (
     uri?: Uri,
   ) => Promise<void>;
+  "codeQLQueries.runLocalQueryFromQueriesPanel": (uri?: Uri) => Promise<void>;
```
koesie10, line 133: Is there a reason for the separation of these two commands? They seem pretty related to me and I would expect both `codeQLQueries` commands to be in the same type.

extensions/ql-vscode/src/queries-panel/queries-module.ts
koesie10: We have some other optional commands as well, the `TestUICommands` commands. These are declared as `Partial<TestUICommands>` in the `commands.ts` file so that they aren't required to be set. I think that would work well here as well.

extensions/ql-vscode/src/queries-panel/queries-module.ts
```diff
@@ -6,16 +6,42 @@ import { DisposableObject } from "../common/disposable-object";
...
+    return {
+      ...this.queriesPanel.getCommands(),
+    };
+  }
+
   private initialize(app: App, cliServer: CodeQLCliServer): void {
-    if (!isCanary() || !showQueriesPanel()) {
-      // Currently, we only want to expose the new panel when we are in canary mode
-      // and the user has enabled the "Show queries panel" flag.
+    if (!this.shouldInitializeQueriesPanel) {
```
koesie10, line 44: This is not calling the method, but is checking that the method exists. It should be (also in the other instances):

```suggestion
    if (!this.shouldInitializeQueriesPanel()) {
```

You can also change the `shouldInitializeQueriesPanel` to be a getter, which would be my preferred solution in this case since there are no real side-effects.

extensions/ql-vscode/src/queries-panel/queries-panel.ts
```diff
@@ -14,4 +20,19 @@ export class QueriesPanel extends DisposableObject {
...
+      "codeQLQueries.runLocalQueryContextInline": this.runLocalQuery.bind(this),
+    };
+  }
+
+  private async runLocalQuery(
+    queryTreeViewItem: QueryTreeViewItem,
+  ): Promise<void> {
+    await this.app.queryServerCommands.execute(
+      "codeQLQueries.runLocalQueryFromQueriesPanel",
+      vscode.Uri.file(queryTreeViewItem.path),
```
koesie10, line 35: My preference would be to import `Uri` from `vscode` directly, so that we could call this instead:

```suggestion
     Uri.file(queryTreeViewItem.path),
```

This is how we're doing this in most other files as well.

extensions/ql-vscode/package.json
```diff
@@ -1095,6 +1104,11 @@
           "group": "1_queryHistory@1",
           "when": "viewItem == remoteResultsItem"
         },
+        {
+          "command": "codeQLQueries.runLocalQueryContextInline",
+          "group": "inline",
+          "when": "view == codeQLQueries && viewItem == queryFile && codeQL.currentDatabaseItem"
```
shati-patel, line 1110: Maybe a silly question: do we need `codeQL.currentDatabaseItem`?

If you don't have a current database, the "Run Local Query" command pops up a window and asks you to select one anyway. (I don't think it matters either way, just wondering 🙃)

norascheuch: Hm, no reason given. I just thought this was the requirement.

extensions/ql-vscode/src/queries-panel/queries-module.ts
norascheuch: Thanks for catching this.

extensions/ql-vscode/src/queries-panel/queries-panel.ts
norascheuch: 👍 I used vscode here bc line 18, but adjusted both now

extensions/ql-vscode/src/common/commands.ts
norascheuch: The command here is a new replica of the `runQuery` command that is defined in the local-queries and therefore defined here in `LocalQueryCommands`. It is executed in the queries-panel when a user 'executes' `"codeQLQueries.runLocalQueryContextInline"` through the panel. That's why we have 2 different commands. 

I suppose we could adjust the original implementation in `runQuery` to handle the case when no uri is supplied but a QueryPanelTreeViewItem instead and leave the queries-panel out of it?..

extensions/ql-vscode/package.json
shati-patel: We can probably remove it in that case, to slightly simplify the `when` 🤷🏽

norascheuch: After discussing this and checking the flow for both cases we decided to keep the db check.

extensions/ql-vscode/package.json
```diff
@@ -501,6 +501,11 @@
         "command": "codeQL.copyVersion",
         "title": "CodeQL: Copy Version Information"
       },
+      {
+        "command": "codeQL.runLocalQueryFromQueriesPanel",
```
shati-patel, line 505: We still want this command to be prefixed with `codeQLQueries` instead of `codeQL`, since its a command that is scoped to a specific CodeQL-controlled panel (instead of being generally available in the file editor/command palette). 

(This would match the other commands in `"view/item/context"` too 😎 )

extensions/ql-vscode/test/unit-tests/command-lint.test.ts
```diff
@@ -65,6 +65,12 @@ describe("commands declared in package.json", () => {
     contribContextMenuCmds.add(command);
   });
 
+  menus["view/item/context"].forEach((commandDecl: CmdDecl) => {
+    const { command } = commandDecl;
+    paletteCmds.delete(command);
+    scopedCmds.add(command);
+  });
+
```
shati-patel, line 73: As mentioned above, I think the previous change was correct (deleting this bit, but adding `command.match(/^codeQLQueries\./)` to the "allowed prefixes"). 

We currently add a new prefix for each new panel, so this would be consistent with the others 😺

shati-patel (commented):
> It seems the command lint test was not quite up to date, I adjusted a bit of it to pass the tests sensibly for the current change.

I've left some comments about the command naming, but otherwise this PR looks nice and tidy 🙌🏽

extensions/ql-vscode/package.json
norascheuch: Ah, yes 🤦‍♀️

robertbrignull (approved):
This looks ok to me, though I wonder if in the long run this is introducing coupling that we don't want. I'm happy for it to be merged, depending on what @koesie10 thinks.

Putting the `CodeQLQueries.runLocalQueryFromQueriesPanel` command on the `LocalQueries` class certainly is the least complex solution right now. It avoids some problem like the queries panel not being initialised, though it may be we'll run into that again for the next command we want to add and it'll be harder to solve that time. Long term I wonder if `LocalQueries` should depend on the queries panel, or if it should be more the other way around. In other component we pass in `LocalQueries` as an argument.

I'd also be happy to go for (and maybe have a slight preference for) the [suggestion](https://github.com/github/vscode-codeql/pull/2534#discussion_r1239617305) of using `Partial` to get around the initialisation problem, and thus keeping the command definition in `QueriesPanel`. To make this work we could expose the `LocalQueries.runQuery` method publicly so it can call that. All that method does is wrap the public `compileAndRunQuery` method with a progress bar, so making `runQuery` public too seems ok.

robertbrignull:
I think let's merge this so we unblock this work, but keep in mind the idea of moving the commands to `QueriesPanel` so it makes the dependency go in the other direction which is probably a little cleaner. We can make that future change at any point, so opening a tech-debt issue would be fine.

norascheuch:
Ok, will open the tech dept issue @robertbrignull . I'm wondering how we could make this process more efficient though. All three options went through my head while I was working on this at some point. How do I know when is a good way to discuss these things before the work is finished and would need to be redone?

PR #2535, opened 2023-06-21
Add initial implementation of framework mode

koesie10:
This adds an initial rough implementation of framework mode for the data extensions editor. Framework mode will do the following:
- Load public methods instead of external API calls
- Group by package instead of library
- Remove the usages in the table because we do not have access to usages
- Change "Download and generate" to "Generate"

Short demo:

https://github.com/github/vscode-codeql/assets/1112623/b985942b-a9aa-4c64-b025-c8c14157214c

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
@@ -51,6 +51,7 @@ import {
 import { showLlmGeneration } from "../config";
 import { getAutoModelUsages } from "./auto-model-usages-query";
 import { getOnDiskWorkspaceFolders } from "../common/vscode/workspace-folders";
+import { Mode } from "./shared/mode";
 
 export class DataExtensionsEditorView extends AbstractWebview<
   ToDataExtensionsEditorMessage,
@@ -65,6 +66,7 @@ export class DataExtensionsEditorView extends AbstractWebview<
     private readonly queryStorageDir: string,
     private readonly databaseItem: DatabaseItem,
     private readonly extensionPack: ExtensionPack,
+    private mode: Mode = Mode.Application,
   ) {
     super(ctx);
   }
@@ -139,6 +141,12 @@ export class DataExtensionsEditorView extends AbstractWebview<
         );
 
         break;
+      case "switchMode":
+        this.mode = msg.mode;
+
+        await Promise.all([this.setViewState(), this.loadExternalApiUsages()]);
+
+        break;
       default:
         assertNever(msg);
     }
@@ -160,6 +168,7 @@ export class DataExtensionsEditorView extends AbstractWebview<
       viewState: {
         extensionPack: this.extensionPack,
         showLlmButton: showLlmGeneration(),
+        mode: this.mode,
       },
     });
   }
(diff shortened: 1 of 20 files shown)


extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -313,29 +344,35 @@ export class DataExtensionsEditorView extends AbstractWebview<
...
-      this.cliServer,
-    );
-    if (!database) {
-      await this.clearProgress();
-      void this.app.logger.log("No database chosen");
+    let addedDatabase: DatabaseItem | undefined;
+
+    // In application mode, we need the database of a specific library to generate
+    // the modeled methods. In framework mode, we'll use the current database.
+    if (this.mode === Mode.Application) {
```
starcke, line 351: There are quite a lot of `mode` checks. It is probably ok, but if we start to get even more we might want to consider some kind of mode abstraction, so that we dont have to do an `if` in all the logic.

extensions/ql-vscode/src/data-extensions-editor/yaml.ts
```diff
@@ -116,6 +117,33 @@ export function createDataExtensionYamlsPerLibrary(
   return result;
 }
 
+export function createDataExtensionYamlsForFrameworkMode(
```
starcke, line 120: Something to think about, is that eventually it would be amazing if modelling library X while in application mode for A, would yield the same file as modelling library X in framework mode.

extensions/ql-vscode/src/view/data-extensions-editor/MethodRow.tsx
```diff
@@ -165,11 +168,13 @@ export const MethodRow = ({
           {externalApiUsage.methodParameters}
         </SupportSpan>
       </VSCodeDataGridCell>
-      <VSCodeDataGridCell gridColumn={3}>
-        <UsagesButton onClick={jumpToUsage}>
-          {externalApiUsage.usages.length}
-        </UsagesButton>
-      </VSCodeDataGridCell>
+      {mode === Mode.Application && (
```
starcke, line 171: Hmm... these mode comparisons do go in many places.

extensions/ql-vscode/src/data-extensions-editor/queries/java.ts
```diff
@@ -29,6 +29,100 @@ where
...
+private import semmle.code.java.dataflow.internal.FlowSummaryImpl as FlowSummaryImpl
+private import semmle.code.java.dataflow.internal.ModelExclusions
+
+/** Holds if the given callable is not worth supporting. */
+private predicate isUninteresting(Callable c) {
+  c.getDeclaringType() instanceof TestLibrary or
+  c.(Constructor).isParameterless()
+}
+
+class PublicMethod extends Callable {
```
starcke, line 56: For my understanding: this is basically a copy of `ExternalAPI` with the constructor changed? I cannot completely parse whether that is exactly what we want, but I think at a minimum it is close.

I think moving to the CodeML queries is something we should do, otherwise we probably want to make some changes (e.g. sharing the code for all those methods).

extensions/ql-vscode/src/data-extensions-editor/queries/java.ts
```diff
@@ -29,6 +29,100 @@ where
...
+  PublicMethod() { this.isPublic() and not isUninteresting(this) }
+
+  /**
+   * Gets information about the external API in the form expected by the MaD modeling framework.
```
starcke, line 60: This is not really an external API anymore. If you think it is easier to have the direct copy then I am ok with leaving it.

extensions/ql-vscode/src/data-extensions-editor/queries/csharp.ts
```diff
@@ -27,6 +27,139 @@ where
   supported = isSupported(api) and
   usage = aUsage(api)
 select usage, apiName, supported.toString(), "supported", api.getFile().getBaseName(), "library"
+`,
+  frameworkModeQuery: `/**
```
starcke, line 31: Same comments as the java method below.

starcke (approved):
I love that we already have a rough version of framework mode! I do think that it is a bit more rough than the normal mode, (kinda like the poc we did). I think this can go in now and then next week we can discuss what next steps should be for this mode.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
koesie10: Definitely. I think for now with just two modes this works fine, but this will break down if we have more modes.

extensions/ql-vscode/src/data-extensions-editor/yaml.ts
koesie10: Yes, I think this is something to think about. For now, I don't think we can really detect this reliably because we're depending on two different inputs (database vs JAR name).

extensions/ql-vscode/src/data-extensions-editor/queries/java.ts
koesie10: Yes, this basically changes `not this.fromSource()` to `this.isPublic()` in the constructor. The rest of the methods are very similar. We can work on reducing the duplication in a follow-up.

PR #2536, opened 2023-06-21
Throw error instead of using fail method

robertbrignull:
Should we be using the `fail` method? We seem to be using it in a few tests to catch incorrect setup, so if everything is correct then this code path doesn't run. But when it does run it appears the `fail` method isn't defined. See the following error:
Screenshot 2023-06-21 at 16 03 33

I haven't looked deeply into this yet, so I'm not sure if there's something else causing the above failure and maybe we do still want to use `fail`. But I wonder is there any downside to throwing an exception in any of these cases instead?

## Checklist


koesie10 (approved):
LGTM, this should be almost equivalent to what `fail` would have done if it existed. I think the only difference is in `helpers-pure.test.ts` because it will catch the error, but it should still fail the test since the error message is different.

PR #2538, opened 2023-06-21
Save dirty documents before evaluating queries

### LINKED ISSUE #2390: Quick eval doesn't save file first
**Describe the bug**
When I quick eval, if the file has been edited then it isn't saved and the on-disk version is used. This is quite irritating.

**Version**
CodeQL extension version: 1.8.3 
CodeQL CLI version: 2.12.7 
Platform: darwin x64

Version: 1.78.0-insider (Universal)
Commit: 5b6af0748264d83dc7a95f5f4b1c39196eaae8d9
Date: 2023-04-27T23:57:36.064Z (13 hrs ago)
Electron: 22.4.8
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Darwin x64 22.4.0
Sandboxed: YesVersion: 1.78.0-insider (Universal)
Commit: 5b6af0748264d83dc7a95f5f4b1c39196eaae8d9
Date: 2023-04-27T23:57:36.064Z (13 hrs ago)
Electron: 22.4.8
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Darwin x64 22.4.0
Sandboxed: Yes

**To reproduce**
1. Open a file containing some QL, edit and don't save.
2. Quick eval something in the file
3. The file is not saved, and what is evaluated is based on the version of the file on disk.

**Expected behavior**
The file is saved before the selection is evaluated.

**Additional context**
I looked for relevant settings and the only one I could find was “Code QL › Running Queries: Auto Save”, which is already enabled.

I haven't checked if the same is true for evaluating a query.

dbartol:
Fixes #2390

When I refactored some of the query evaluation code, I broke the existing mechanism for saving dirty documents before evaluation. I still display the prompt to save, but then ignored the result and never saved the documents :facepalm:.

When using the VS Code debugger UI, there's already a setting named `debug.saveBeforeStart` that controls which documents to save before launching a debug session. When launching a query via a built-in VS Code command like "Start Debugging", VS Code saves the documents automatically. I've tried to emulate that behavior for the code paths that don't go through VS Code commands, including the non-debugger "CodeQL: Quick evaluation" and "CodeQL: Run query on selected database". I've also added that behavior to "CodeQL: Debug selection" when using an existing debug session, since it doesn't go through VS Code's debug _launch_ code path.

Depend on the user setting, VS Code will either save all dirty documents in the active tab group, including or excluding untitled documents, or it will not save anything. There's no exposed command that does this saving, so I had to emulate the behavior as best I could. I'm sure it's not _exactly_ equivalent, but it should be close enough that nobody will notice.

I marked our existing CodeQL-specific setting for saving before evaluation as deprecated and ignored, pointing users to the core VS Code setting.

Testing the emulation was tricky, because `TextDocument`'s properties and methods are marked either non-configurable or readonly, so Jest can't spy on them. Instead, see the comments for how I tested using public VS Code APIs and one terrible hack. The Interwebs provided some incantations to allow the spying, but they all involved replacing methods on the prototype of `Object` so they seemed a bit scary. If someone more comfortable with JavaScript and mocking can suggest a better solution, I'm happy to change it.


extensions/ql-vscode/test/vscode-tests/minimal-workspace/debugger/save-before-start.test.ts
```diff
@@ -0,0 +1,137 @@
...
+  // We can't easily mock `TextDocument` because its properties are read-only and/or not
+  // configurable, so we rely on the actual `save()` method and the actual `isDirty` property.
+
+  beforeEach(async () => {
+    await commands.executeCommand("workbench.action.closeAllEditors");
+  });
+
+  it("should not save untitled documents without `includeUntitled`", async () => {
+    const dirtyDoc = await mockDocument(dirtyFile, true, ViewColumn.One);
+    const cleanDoc = await mockDocument(cleanFile, false, ViewColumn.One);
```
aeisenberg, line 72: Should this also include a dirty, untitled doc?

extensions/ql-vscode/test/vscode-tests/minimal-workspace/debugger/save-before-start.test.ts
```diff
@@ -0,0 +1,137 @@
...
+const otherGroupDirtyFile = "debugger/other-dirty.ql";
+const otherGroupCleanFile = "debugger/other-clean.ql";
+
+describe("saveBeforeStart", () => {
+  // We can't easily mock `TextDocument` because its properties are read-only and/or not
+  // configurable, so we rely on the actual `save()` method and the actual `isDirty` property.
+
+  beforeEach(async () => {
+    await commands.executeCommand("workbench.action.closeAllEditors");
+  });
```
aeisenberg, line 68: Maybe do the same in an `afterEach` block?

extensions/ql-vscode/test/vscode-tests/minimal-workspace/debugger/save-before-start.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    const dirtyDoc = await mockDocument(dirtyFile, true, ViewColumn.One);
+    const cleanDoc = await mockDocument(cleanFile, false, ViewColumn.One);
+
+    await saveAllInGroup(false);
+
+    expect(dirtyDoc.isDirty).toBe(false);
+    expect(cleanDoc.isDirty).toBe(false);
+  });
+
+  it("should not save dirty documents in other tab groups", async () => {
```
aeisenberg, line 80: This feature is surprising to me. I would expect all docs in the workspace to be saved when I invoke the debugger. Is this the behaviour of this setting in the vscode?

extensions/ql-vscode/src/config.ts
```diff
@@ -69,6 +84,12 @@ export interface InspectionResult<T> {
   workspaceFolderValue?: T;
 }
 
+const VSCODE_DEBUG_SETTING = new Setting("debug", undefined, "ql");
```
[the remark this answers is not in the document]
dbartol: To be honest, I'm not sure either. How VS Code computes the override behavior between, say, a language-independent workspace setting and a language-specific user setting is up to VS Code, but doesn't change how we implement this feature.

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -491,55 +490,78 @@ async function getSelectedPosition(
   };
 }
 
+type SaveBeforeStartMode =
+  | "nonUntitledEditorsInActiveGroup"
```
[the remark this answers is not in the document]
dbartol: That's not the name I would have chosen, either, but that's what VS Code uses.

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -491,55 +490,78 @@ async function getSelectedPosition(
...
+      break;
 
-      if (chosenItem === cancelItem) {
-        throw new UserCancellationException("Query run cancelled.", true);
+    case "none":
+      break;
+
+    default:
+      // Unexpected value. Fall back to the default behavior.
+      await saveAllInGroup(false);
```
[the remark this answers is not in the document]
dbartol: I'm more worried about VS Code adding a new value, or perhaps a user writing an invalid value in their `settings.json`. So, I do want a runtime check.

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -491,55 +490,78 @@ async function getSelectedPosition(
...
+    if (openDocument.isDirty) {
+      console.warn(`${openDocument.uri.toString()} is dirty.`);
+      if (!openDocument.isUntitled || includeUntitled) {
+        dirtyDocumentUris.set(openDocument.uri.toString(), openDocument);
+      }
+    }
+  }
+  if (dirtyDocumentUris.size > 0) {
+    console.warn(`${window.tabGroups.all.length} tab groups open`);
+    console.warn(`${workspace.textDocuments.length} documents open`);
```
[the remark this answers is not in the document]
dbartol: Oops, printf debugging leftovers. Removed.

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -491,55 +490,78 @@ async function getSelectedPosition(
...
+      // need to know to match it up with a dirty document.
+      if (typeof input === "object") {
+        const uri = (input as any).uri;
+        if (uri instanceof Uri) {
+          const document = dirtyDocumentUris.get(uri.toString());
+          if (document !== undefined) {
+            console.warn(`Saving ${uri.toString()}`);
+            await document.save();
+            // Remove the URI from the dirty list so we don't wind up saving the same file twice
+            // if it's open in multiple editors.
```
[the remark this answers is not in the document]
dbartol: Probably none, but it seemed simple enough to avoid here.

OK, maybe there's one corner case: An untitled document open in two different editors. When we try to save it the first time, the user will be prompted for a file name. If they cancel, it will remain unsaved, so we'd wind up prompting them a second time when we get to the second editor tab for that document.

extensions/ql-vscode/test/vscode-tests/minimal-workspace/debugger/save-before-start.test.ts
dbartol: :facepalm: I was tricked by the name, but the VS Code source code just calls the internal equivalent of `saveAll()`. The actual behavior for other languages saves tabs in other groups as well. I'll fix it.

My bet is that somebody thought that saving only the current tab group was a good idea early on, so that's what the setting was named. Then, when every developer everywhere tried it and hated it, they just made it implicitly mean "save all".

dbartol: I've changed to code to just use `saveAll()`. The handling of untitled documents appears to be slightly different, but close enough that I won't bother emulating the VS Code behavior precisely.

The original test for `saveAllInGroup()` is now irrelevant, so instead I added a debugger test case to make sure the documents get saved at the right points. Since the source files are modified by the test, this required making a temporary copy of the `data` directory to use as the workspace, instead of opening the committed `data` directory directly. I had to move the test queries _into_ the data directory that we use as the workspace folder, too, since we were pulling them directly from source.

extensions/ql-vscode/test/vscode-tests/minimal-workspace/debugger/save-before-start.test.ts
dbartol: This file has been removed.

extensions/ql-vscode/test/vscode-tests/minimal-workspace/debugger/save-before-start.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Returns a promise that resolves after the given number of milliseconds, and returns the string
+ * "timeout".
+ */
+function delay(ms: number): Promise<string> {
+  return new Promise((resolve) => setTimeout(() => resolve("timeout"), ms));
+}
+
+jest.setTimeout(10000); // A little extra time for the save dialog to appear.
```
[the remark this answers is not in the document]
dbartol: This file has been removed.

aeisenberg (approved):
Nice. I'm assuming there's just a mixup in the tests. If so, 👍🏼.

extensions/ql-vscode/test/vscode-tests/cli-integration/debugger/debugger.test.ts
```diff
@@ -150,34 +150,36 @@ describeWithCodeQL()("Debugger", () => {
...
-
-      // Should have saved the dirty document.
-      expect(editor.document.isDirty).toBe(false);
-
-      await controller.expectSucceeded();
-      await controller.expectStopped();
-
-      await editor.edit((editBuilder) => {
-        editBuilder.insert(new Position(0, 0), "/* another comment */");
+  if (process.platform !== "win32") {
```
aeisenberg, line 153: Is this functionality broken on windows, or is it just that the test is failing? Would you be able to add a comment to clarify this?

dbartol: Still struggling to fix a test hang that only repros on CI. Just ignore this until I get CI passing.

dbartol: In what I believe is the third :facepalm: of this PR, I finally determined that the timeouts are _not_ due to the "Save As" dialog popping up on an unexpected dirty document, but rather due to Workspace Trust, which triggers a prompt because I changed the tests to copy their workspace data to a temp folder before opening. I've disabled Workspace Trust when invoking the CLI integration tests.

PR #2539, opened 2023-06-22
Convert RawTable to a function component

robertbrignull:
Converts the `RawTable` from a class component to a function component.

Paired on this with @norascheuch and @shati-patel.

As far as I can tell, all the behaviour still works. The only complicated bit is the scroller. You can trigger this behaviour by using the `codeQLQueryResults.down` command (and similar commands for other directions) to navigate around the raw results table. When the selection goes off the screen it'll scroll to bring it back to the centre.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/results/raw-results-table.tsx
+++ b/extensions/ql-vscode/src/view/results/raw-results-table.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { useCallback, useEffect, useRef, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import {
   className,
   emptyQueryResultsMessage,
@@ -116,15 +116,18 @@ export function RawTable({
     };
   }, [handleNavigationEvent]);
 
-  let dataRows = resultSet.rows;
+  const [dataRows, numTruncatedResults] = useMemo(() => {
+    if (resultSet.rows.length <= RAW_RESULTS_LIMIT) {
+      return [resultSet.rows, 0];
+    }
+    return [
+      resultSet.rows.slice(0, RAW_RESULTS_LIMIT),
+      resultSet.rows.length - RAW_RESULTS_LIMIT,
+    ];
+  }, [resultSet]);
+
   if (dataRows.length === 0) {
     return emptyQueryResultsMessage();
-  }
-
-  let numTruncatedResults = 0;
-  if (dataRows.length > RAW_RESULTS_LIMIT) {
-    numTruncatedResults = dataRows.length - RAW_RESULTS_LIMIT;
-    dataRows = dataRows.slice(0, RAW_RESULTS_LIMIT);
   }
 
   const tableRows = dataRows.map((row: ResultRow, rowIndex: number) => (
--- a/extensions/ql-vscode/src/view/results/raw-results-table.tsx
+++ b/extensions/ql-vscode/src/view/results/raw-results-table.tsx
@@ -41,8 +41,11 @@ export function RawTable({
 }: RawTableProps) {
   const [selectedItem, setSelectedItem] = useState<TableItem | undefined>();
 
-  const scroller = useRef(new ScrollIntoViewHelper());
-  useEffect(() => scroller.current.update());
+  const scroller = useRef<ScrollIntoViewHelper | undefined>(undefined);
+  if (scroller.current === undefined) {
+    scroller.current = new ScrollIntoViewHelper();
+  }
+  useEffect(() => scroller.current?.update());
 
   const setSelection = useCallback((row: number, column: number): void => {
     setSelectedItem({ row, column });
@@ -76,7 +79,7 @@ export function RawTable({
             jumpToLocation(location, databaseUri);
           }
         }
-        scroller.current.scrollIntoViewOnNextUpdate();
+        scroller.current?.scrollIntoViewOnNextUpdate();
         return { row: nextRow, column: nextColumn };
       });
     },
--- a/extensions/ql-vscode/src/view/results/raw-results-table.tsx
+++ b/extensions/ql-vscode/src/view/results/raw-results-table.tsx
@@ -1,4 +1,5 @@
 import * as React from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import {
   className,
   emptyQueryResultsMessage,
@@ -38,22 +39,17 @@ export function RawTable({
   sortState,
   offset,
 }: RawTableProps) {
-  const [selectedItem, setSelectedItem] = React.useState<
-    TableItem | undefined
-  >();
+  const [selectedItem, setSelectedItem] = useState<TableItem | undefined>();
 
-  const scroller = React.useRef(new ScrollIntoViewHelper());
-  React.useEffect(() => scroller.current.update());
+  const scroller = useRef(new ScrollIntoViewHelper());
+  useEffect(() => scroller.current.update());
 
-  const setSelection = React.useCallback(
-    (row: number, column: number): void => {
-      setSelectedItem({ row, column });
-      sendTelemetry("local-results-raw-results-table-selected");
-    },
-    [],
-  );
+  const setSelection = useCallback((row: number, column: number): void => {
+    setSelectedItem({ row, column });
+    sendTelemetry("local-results-raw-results-table-selected");
+  }, []);
 
-  const navigateWithDelta = React.useCallback(
+  const navigateWithDelta = useCallback(
     (rowDelta: number, columnDelta: number): void => {
       setSelectedItem((prevSelectedItem) => {
         const numberOfAlerts = resultSet.rows.length;
@@ -87,7 +83,7 @@ export function RawTable({
     [databaseUri, resultSet, scroller],
   );
 
-  const handleNavigationEvent = React.useCallback(
+  const handleNavigationEvent = useCallback(
     (event: NavigateMsg) => {
       switch (event.direction) {
         case NavigationDirection.up: {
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/view/results/raw-results-table.tsx
```diff
@@ -19,158 +18,157 @@ import { onNavigation } from "./results";
...
+  resultSet,
+  sortState,
+  offset,
+}: RawTableProps) {
+  const [selectedItem, setSelectedItem] = React.useState<
+    TableItem | undefined
+  >();
+
+  const scroller = React.useMemo(() => new ScrollIntoViewHelper(), []);
+  scroller.update();
```
robertbrignull, line 46: I think the behaviour we want for the scroller is we want one scroller per `RawTable` instance, and we don't want to create a new one every time we render. So I'm using `useMemo` to cache and instance but then calling `scroller.update()` directly every time we render. Is this ok? It seems to work. Internally the scroller keeps a `shouldScrollIntoView` field so it'll only actually do anything when it has previously been instructed to do so.

Since the `ScrollIntoViewHelper` class is used from another location I don't want to go too hard on refactoring it right now, but once we've also dealt with `PathTable` we could update the scroller to fit with a functional component better, such as ditching `shouldScrollIntoView` and being triggered using a `useEffect` hook on `selectedItem`.

extensions/ql-vscode/src/view/results/raw-results-table.tsx
```diff
@@ -19,158 +18,157 @@ import { onNavigation } from "./results";
 import { tryGetResolvableLocation } from "../../common/bqrs-utils";
 import { ScrollIntoViewHelper } from "./scroll-into-view-helper";
 import { sendTelemetry } from "../common/telemetry";
+import { assertNever } from "../../common/helpers-pure";
 
-export type RawTableProps = ResultTableProps & {
+export type RawTableProps = {
```
robertbrignull, line 23: I'm not sure why we were requiring `ResultTableProps` as props but then barely accessing any of the members. I've changed this to just require `databaseUri`. The place where we call `RawTable` we do `{...this.props}` and this still works since unknown props are just ignored.

shati-patel (commented):
Thanks! Had another look and tested locally. Seems fine to me! 

Since we paired, and since I don't really understand the `scroller` situation, I'll let someone else review/approve 🙇🏽‍♀️

extensions/ql-vscode/src/view/results/raw-results-table.tsx
koesie10: I think it would be slightly better to call `useEffect` here to make it explicit that we're calling this method on every render. `useEffect` without any dependencies is run on every render as well, but I think it will check if a re-render is necessary after it, so it should be more correct:

```suggestion
  useEffect(() => scroller.update());
```

extensions/ql-vscode/src/view/results/raw-results-table.tsx
```diff
@@ -19,158 +18,157 @@ import { onNavigation } from "./results";
...
+  databaseUri,
+  resultSet,
+  sortState,
+  offset,
+}: RawTableProps) {
+  const [selectedItem, setSelectedItem] = React.useState<
+    TableItem | undefined
+  >();
+
+  const scroller = React.useMemo(() => new ScrollIntoViewHelper(), []);
```
koesie10, line 45: [`useRef`](https://react.dev/reference/react/useRef) is probably better for this since it's made for keeping things such as this. The behavior is probably identical, but `useRef` will guarantee that this value is not cleared and will not cause a re-render.

extensions/ql-vscode/src/view/results/raw-results-table.tsx
```diff
@@ -19,158 +18,157 @@ import { onNavigation } from "./results";
...
+    <RawTableRow
+      key={rowIndex}
+      rowIndex={rowIndex + offset}
+      row={row}
+      databaseUri={databaseUri}
+      selectedColumn={
+        selectedItem?.row === rowIndex ? selectedItem?.column : undefined
+      }
+      onSelected={setSelection}
+      scroller={scroller}
```
koesie10, line 144: (If you take my suggestion of using `useRef` for the scroller, the `RawTableRow` should probably also receive a `ref`. I don't think changing this to `scroller.current` would be correct)

extensions/ql-vscode/src/view/results/raw-results-table.tsx
robertbrignull: Ah, I had forgotten that `useEffect` without any second argument at all (or I guess implicitly passing `undefined`) had that behaviour. I agree that would be better.

I remember always feeling like the empty array and undefined should be the same here, but actually there's complete opposite behaviours. Just got to remember that.

extensions/ql-vscode/src/view/results/raw-results-table.tsx
```diff
@@ -19,158 +18,157 @@ import { onNavigation } from "./results";
...
+  if (dataRows.length === 0) {
+    return emptyQueryResultsMessage();
   }
 
-  componentDidUpdate() {
-    this.scroller.update();
+  let numTruncatedResults = 0;
+  if (dataRows.length > RAW_RESULTS_LIMIT) {
+    numTruncatedResults = dataRows.length - RAW_RESULTS_LIMIT;
+    dataRows = dataRows.slice(0, RAW_RESULTS_LIMIT);
```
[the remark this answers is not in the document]
robertbrignull: Yep, that makes sense. Can definitely remove the `let` here and do the truncation inside a `useMemo`.

extensions/ql-vscode/src/view/results/raw-results-table.tsx
robertbrignull: I don't quite understand the reasoning here. Could you explain a bit more? My thinking is that passing `scroller.current` here should be the thing to do.

My understanding of `useRef` is that when the value changes it does not trigger a redraw. So if the value did change we wouldn't re-run this code. We also don't ever change the value of `scroller` anyway so it doesn't really matter, but let's ignore that.

Let's say we did change the value and it did redraw for some other reason. In that case we want to pass a new value to `RawTableRow` because the `scroller` object itself won't have changed so it won't see that it needs to redraw `RawTableRow` too and not use a cached value. By passing `scroller.current` we would be passing a new object and it would re-render `RawTableRow` correctly.

Or is there another reason that I'm missing?

koesie10: I'm actually not totally sure about this. [The documentation](https://react.dev/reference/react/useRef) mentions "Do not write or read ref.current during rendering." That would imply that we shouldn't read `scroller.current` anywhere, not even when we pass it along. I think one way around this is the following:

1. Pass the `scroller` ref into the `RawTableRow`
2. Make the `RawTableRow` accept `scroller?: RefObject<ScrollIntoViewHelper>;`
3. Instead of using `ref={props.scroller?.current?.ref(isSelected)}`, we should create a wrapper around this so we don't read `props.scroller.current` during rendering. We should be able to do this using `ref={node => props.scroller?.current?.ref(isSelected)(node)}`, but this doesn't work because the functional refs don't interface well with the class-component refs.

I think this would be the ideal situation to avoid the pitfall mentioned on that documentation page, but we can't do that for now. I would suggest just using the current implementation (i.e. setting `scroller={scroller.current}`) and working on fixing this potential bug later.

extensions/ql-vscode/src/view/results/raw-results-table.tsx
```diff
@@ -19,158 +19,155 @@ import { onNavigation } from "./results";
...
+}: RawTableProps) {
+  const [selectedItem, setSelectedItem] = useState<TableItem | undefined>();
 
-  constructor(props: RawTableProps) {
-    super(props);
-    this.setSelection = this.setSelection.bind(this);
-    this.handleNavigationEvent = this.handleNavigationEvent.bind(this);
-    this.state = {};
-  }
+  const scroller = useRef(new ScrollIntoViewHelper());
```
koesie10, line 44: This will recreate the `ScrollIntoViewHelper` on every render, even though it won't be used. [The React documentation](https://react.dev/reference/react/useRef#avoiding-recreating-the-ref-contents) shows an example of how to avoid this, so I think we should follow this:

```suggestion
  const scroller = useRef<ScrollIntoViewHelper | null>(null);
  if (scroller.current === null) {
    scroller.current = new ScrollIntoViewHelper();
  }
```

extensions/ql-vscode/src/view/results/raw-results-table.tsx
robertbrignull: That's interesting. I didn't know you could pass a function to that `ref=` property. I'm not totally sure if that counts as "during rendering" of not.

My understanding of the danger is that it's around changing ref data and because a component is not entirely defined by its props and state. That means we avoid most if not all of the danger because we don't ever change the ref. I think what we have now is probably ok but just ugly. I'm ok with leaving it as it is for now and coming back to fix it later.

Another different approach I thought of could be https://github.com/github/vscode-codeql/commit/f91f7aa9affe2ebe9372526e335bfb666846a320, avoiding using a ref entirely and instead look up the element by id. I'm not 100% sure this is safe either, since I wonder when the DOM is changed and when `useEffect` hooks run.

extensions/ql-vscode/src/view/results/raw-results-table.tsx
robertbrignull: Thanks for pointing this out.

extensions/ql-vscode/src/view/results/raw-results-table.tsx
koesie10: I believe the DOM is already changed when `useEffect` is run. If you want to run something before the DOM is updated, there's another hook [`useLayoutEffect`](https://react.dev/reference/react/useLayoutEffect) that's made specifically for that purpose. I'm happy to go with this solution for now, but removing the ref and making this easier to understand would definitely be my preference.

PR #2541, opened 2023-06-22
Move tests out of pure directory

robertbrignull:
When moving code out of the `/pure` directory and into `/common` and other directories, I forgot about the tests. This PR moves all of the tests out of `/pure` and into other directories to match the files they are testing.

## Checklist


shati-patel (approved):
Good spot! 🪝 

Out of interest, did the "[Unwanted dependency on vscode API](https://github.com/github/vscode-codeql/blob/main/.github/codeql/queries/assert-no-vscode-dependency.ql)" CodeQL query catch these?

robertbrignull:
> Out of interest, did the "[Unwanted dependency on vscode API](https://github.com/github/vscode-codeql/blob/main/.github/codeql/queries/assert-no-vscode-dependency.ql)" CodeQL query catch these test files? Should it?

That wasn't how I noticed about the test files and I don't think we'd expect it to alert on any of the tests files. In the query we only look for files in the "src" directory.

We could expand the query to include the tests, but it's a little less important, and the tests themselves will fail if you include a vscode dependency in the unit tests. I think it's fine as it is and we don't need to adjust the query.

PR #2552, opened 2023-06-26
Queries Panel: run-all queries of a folders

norascheuch:
Adds a run-all icon to folder elements in the new panel to run local queries.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -511,6 +511,11 @@
         "icon": "$(run)"
       },
       {
+        "command": "codeQLQueries.runLocalQueriesFromPanel",
+        "title": "Run local queries",
+        "icon": "$(run-all)"
+      },
+      {
         "command": "codeQL.runLocalQueryFromFileTab",
         "title": "CodeQL: Run local query",
         "icon": "$(run)"
@@ -1140,6 +1145,11 @@
           "when": "view == codeQLQueries && viewItem == queryFile"
         },
         {
+          "command": "codeQLQueries.runLocalQueriesFromPanel",
+          "group": "inline",
+          "when": "view == codeQLQueries && viewItem == queryFolder && codeQL.currentDatabaseItem"
+        },
+        {
           "command": "codeQLTests.showOutputDifferences",
           "group": "qltest@1",
           "when": "viewItem == testWithSource"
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/package.json
```diff
@@ -503,7 +503,13 @@
       },
       {
         "command": "codeQLQueries.runLocalQueryFromQueriesPanel",
-        "title": "Run local query"
+        "title": "Run local query",
+        "icon": "$(run)"
+      },
+      {
+        "command": "codeQLQueries.runLocalQueriesFromPanel",
```
robertbrignull, line 510:
```suggestion
        "command": "codeQLQueries.runLocalQueriesFromQueriesPanel",
```
Should we name this to match https://github.com/github/vscode-codeql/blob/6adb7722d306fb10cbaa40027ab86f25f5989b8b/extensions/ql-vscode/package.json#L505

robertbrignull (commented):
One suggestion on the command name but otherwise looks good to me. I tried it out and it works great.

Also has some conflicts, so would be good to get those sorted and then re-review.

extensions/ql-vscode/package.json
norascheuch: yes, I think in the end I would match Shatis style and remove the Queries Panel words from this since the prefix already specifies that..

norascheuch: I can do it here for both?

robertbrignull: Using the `codeQLQueries.` prefix to mean it's for the queries panel works for me.

I'd prefer to not rename the rest of the existing commands in this PR. If we do that in a separate PR it would avoid holding up this PR and make the history more understandable.

robertbrignull: So, rename the command being added in this PR to `codeQLQueries.runLocalQueries`?
Do you want to do that in this PR, or merge this as it is and rename all three in a followup?

norascheuch: Lets merge this as is and rename in a followup!

shati-patel (approved):
Nice! Happy for you to do the renaming in a follow-up PR ⚡

shati-patel (approved):
Nice! Happy for you to do the renaming in a follow-up PR ⚡

PR #2558, opened 2023-06-28
Exclude methods in tests in framework mode

koesie10:
This excludes methods defined in tests in framework mode, significantly cutting down on the number of methods shown that would need to be modeled.

For C#, this just checks that the file is not a test file, as defined by the QL library.

For Java, this makes a copy of the internal
[`ModelExclusions.qll`](https://github.com/github/codeql/blob/249f9f863db1e94e3c46ca85b49fb0ec32f8ca92/java/ql/lib/semmle/code/java/dataflow/internal/ModelExclusions.qll) file to avoid the use of internal modules. This module will tell us whether a method is "interesting" to model or not. Not all of the checks in this module need to happen for framework mode, but these checks might be useful for telling a user whether a method is interesting to model in application mode.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/queries/java.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/queries/java.ts
@@ -39,9 +39,7 @@ select usage, apiName, supported.toString(), "supported", externalApi.jarContain
 import java
 import AutomodelVsCode
 
-class PublicMethodFromSource extends CallableMethod {
-  PublicMethodFromSource() { this.isPublic() and this.fromSource() }
-}
+class PublicMethodFromSource extends CallableMethod, ModelApi { }
 
 from PublicMethodFromSource publicMethod, string apiName, boolean supported
 where
@@ -81,8 +79,8 @@ class CallableMethod extends Method {
    */
   string getApiName() {
     result =
-      this.getDeclaringType().getPackage() + "." + this.getDeclaringType().nestedName() +
-        "#" + this.getName() + paramsString(this)
+      this.getDeclaringType().getPackage() + "." + this.getDeclaringType().nestedName() + "#" +
+        this.getName() + paramsString(this)
   }
 
   private string getJarName() {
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/data-extensions-editor/queries/java.ts
```diff
@@ -148,6 +146,82 @@ boolean isSupported(CallableMethod method) {
   or
   not method.isSupported() and result = false
 }
+
+// The below is a copy of https://github.com/github/codeql/blob/249f9f863db1e94e3c46ca85b49fb0ec32f8ca92/java/ql/lib/semmle/code/java/dataflow/internal/ModelExclusions.qll
+// to avoid the use of internal modules.
+/** Holds if the given package \`p\` is a test package. */
+pragma[nomagic]
+private predicate isTestPackage(Package p) {
```
starcke, line 154: It is odd that checking whether a file is a test file is public in C# and private in Java. It seems we should be consistent on that matter?

extensions/ql-vscode/src/data-extensions-editor/queries/java.ts
```diff
@@ -148,6 +146,82 @@ boolean isSupported(CallableMethod method) {
...
+  p.getName()
+      .matches([
+          "org.junit%", "junit.%", "org.mockito%", "org.assertj%",
+          "com.github.tomakehurst.wiremock%", "org.hamcrest%", "org.springframework.test.%",
+          "org.springframework.mock.%", "org.springframework.boot.test.%", "reactor.test%",
+          "org.xmlunit%", "org.testcontainers.%", "org.opentest4j%", "org.mockserver%",
+          "org.powermock%", "org.skyscreamer.jsonassert%", "org.rnorth.visibleassertions",
+          "org.openqa.selenium%", "com.gargoylesoftware.htmlunit%", "org.jboss.arquillian.testng%",
+          "org.testng%"
```
starcke, line 163: This seems quite tedious to maintain separately. I am ok with it for now, but we should open an issue to keep track of this. Ideally this is fixed by moving to the CodeML queries.

starcke (approved):
Ok with it for now, but we should probably find a better solution eventually.

extensions/ql-vscode/src/data-extensions-editor/queries/java.ts
koesie10: I had hoped they would both be public, but while we can use the Java `ModelExclusions` file, it didn't seem right to use an internal module. I agree that it's odd that there's a difference between the two languages.

extensions/ql-vscode/src/data-extensions-editor/queries/java.ts
koesie10: This will be part of moving to the CodeML queries, they already [have this implemented for their framework mode](https://github.com/github/codeql/blob/2c99009c1ab6692296460afb38d185ae8d2b10f1/java/ql/src/Telemetry/AutomodelFrameworkModeCharacteristics.qll#L196-L206), so I don't think we need to open a separate issue to keep track of this.

PR #2561, opened 2023-06-28
Enable accessibility linting rules and comment out existing violations

sampart:
cf https://github.com/github/code-stats-collector/pull/390 / https://github.com/github/code-scanning/issues/9347


robertbrignull (approved):
LGTM

I've never seen that syntax for putting eslint directives in TSX react component lines before, so I've also learned something from this 😄

sampart:
I had to experiment a bit with the syntax. [This SO post](https://stackoverflow.com/a/58274180/328817) was helpful.

PR #2563, opened 2023-06-28
Add call classification and supported kind to data extensions editor

koesie10:
This will add a "Test" or "Generated" tag on the method if it's only used in tests or generated code and not in the actual application source code. It also adds retrieving the method "type" (e.g. source or sink) to the queries, but does not show these yet.

Screenshot 2023-07-13 at 16 46 04

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/data-extensions-editor/MethodRow.tsx
+++ b/extensions/ql-vscode/src/view/data-extensions-editor/MethodRow.tsx
@@ -39,13 +39,21 @@ const ViewLink = styled(VSCodeLink)`
   white-space: nowrap;
 `;
 
-const modelTypeOptions = [
+const modelTypeOptions: Array<{ value: ModeledMethodType; label: string }> = [
   { value: "none", label: "Unmodeled" },
   { value: "source", label: "Source" },
   { value: "sink", label: "Sink" },
   { value: "summary", label: "Flow summary" },
   { value: "neutral", label: "Neutral" },
 ];
+
+const typeNames = modelTypeOptions.reduce(
+  (acc, { value, label }) => {
+    acc[value] = label;
+    return acc;
+  },
+  {} as Record<ModeledMethodType, string>,
+);
 
 type Props = {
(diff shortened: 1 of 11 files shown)


extensions/ql-vscode/src/view/data-extensions-editor/MethodRow.tsx
```diff
@@ -197,6 +221,12 @@ function ModelableMethodRow(props: Props) {
           </UsagesButton>
         )}
         <ViewLink onClick={jumpToUsage}>View</ViewLink>
+        {!inSource && (
+          <ClassificationsContainer>
+            {inTest && <VSCodeTag>Test</VSCodeTag>}
```
robertbrignull, line 226: Random thought. Is it possible to put a tooltip / title text on these that explains "This method is only used from test code"?

robertbrignull (approved):
No comments on how the code is written. That all looks great.

My only comment on the styling is whether the classification labels are a bit too large and prominent, and if it makes the view more cluttered. It's the default styling so it's definitely not your fault, but they maybe stand out slightly too much and distract from other elements, and it's unhelpful that they look almost exactly the same as the clickable buttons.

However, I'd be happy to proceed as this is now, with the understanding that we'll be doing a design review once we get to the end of this current implementation work.

koesie10:
> My only comment on the styling is whether the classification labels are a bit too large and prominent, and if it makes the view more cluttered. It's the default styling so it's definitely not your fault, but they maybe stand out slightly too much and distract from other elements, and it's unhelpful that they look almost exactly the same as the clickable buttons.

I agree, the tags are a little bit big. I've made their text sizes smaller so they stand out less, so this should hopefully partly fix that. I've also added the tooltip you suggested. Unfortunately, the webview UI toolkit doesn't have a tooltip component, so it's just a `title` on a `<div>` which shows when you hover over the tags for some time.

robertbrignull (approved):
LGTM

Thanks for those styling tweaks, and putting the classification code in a new component looks good.

PR #2566, opened 2023-06-29
Add library versions to data extensions editor

koesie10:
This adds the versions of libraries to the data extensions editor and hides the filename for Java libraries.

Screenshot 2023-07-14 at 12 03 47

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/bqrs.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/bqrs.ts
@@ -5,6 +5,7 @@ import {
   ExternalApiUsage,
 } from "./external-api-usage";
 import { ModeledMethodType } from "./modeled-method";
+import { parseLibraryFilename } from "./library";
 
 export function decodeBqrsToExternalApiUsages(
   chunk: DecodedBqrsChunk,
@@ -15,7 +16,8 @@ export function decodeBqrsToExternalApiUsages(
     const usage = tuple[0] as Call;
     const signature = tuple[1] as string;
     const supported = (tuple[2] as string) === "true";
-    const library = tuple[4] as string;
+    let library = tuple[4] as string;
+    let libraryVersion: string | undefined = tuple[5] as string;
     const type = tuple[6] as ModeledMethodType;
     const classification = tuple[8] as CallClassification;
 
@@ -37,9 +39,25 @@ export function decodeBqrsToExternalApiUsages(
       methodDeclaration.indexOf("("),
     );
 
+    // For Java, we'll always get back a .jar file, and the library version may be bad because not all library authors
+    // properly specify the version. Therefore, we'll always try to parse the name and version from the library filename
+    // for Java.
+    if (library.endsWith(".jar") || libraryVersion === "") {
+      const { name, version } = parseLibraryFilename(library);
+      library = name;
+      if (version) {
+        libraryVersion = version;
+      }
+    }
+
+    if (libraryVersion === "") {
+      libraryVersion = undefined;
+    }
+
     if (!methodsByApiName.has(signature)) {
       methodsByApiName.set(signature, {
         library,
+        libraryVersion,
         signature,
         packageName,
         typeName,
--- a/extensions/ql-vscode/src/data-extensions-editor/library.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/library.ts
@@ -1,0 +1,55 @@
+import { basename, extname } from "../common/path";
+
+// From the semver package using
+// const { re, t } = require("semver/internal/re");
+// console.log(re[t.LOOSE]);
+// Modifications:
+// - Added version named group which does not capture the v prefix
+// - Removed the ^ and $ anchors
+// - Made the minor and patch versions optional
+// This will match any semver string at the end of a larger string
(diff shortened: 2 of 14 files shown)


extensions/ql-vscode/src/data-extensions-editor/bqrs.ts
```diff
@@ -37,9 +39,25 @@ export function decodeBqrsToExternalApiUsages(
       methodDeclaration.indexOf("("),
     );
 
+    // For Java, we'll always get back a .jar file, and the library version may be bad because not all library authors
+    // properly specify the version. Therefore, we'll always try to parse the name and version from the library filename
+    // for Java.
+    if (library.endsWith(".jar") || libraryVersion === "") {
```
robertbrignull, line 45: I'm not 100% on all the logic going on here, so just correct me if this is all fine and I just don't understand 🙇🏼 

I wonder if it's a good idea to be modifying `library` when `libraryVersion` is unknown. The `parseLibraryFilename` method always at least performs `basename` on the input, so if we couldn't determine the version we'll still perform some operation on the library name.

I wonder should we perhaps always be calling `parseLibraryFilename`? Or should we rename to `parseLibraryVersion` and change the logic so it only returns anything (and therefore we only modify `library` here) if it was able to split the library name into a name and version.

robertbrignull (commented):
Most code looks fine to me, and I tried it out and it appears to work just fine. I have one comment on a piece where I wasn't sure about the logic.

robertbrignull:
This PR is going to have some conflicts in the view code now. If possible, the new design should look like the following, where the dependency version is displayed in the title of each collapsible section.

Screenshot 2023-06-29 at 15 48 12

extensions/ql-vscode/src/data-extensions-editor/bqrs.ts
koesie10: The problem here is that we can get different information from the Java and C# queries. For Java, we get the JAR filename and a (potentially incorrect or missing) version. These are some examples of these tuples:

```
rt.jar, [empty string]
sql2o-1.6.0.jar, [empty string]
spring-boot-3.0.2.jar, [empty string]
mybatis-3.2.4.jar, 3.2.4
persistence-api-1.0.2.jar, 1.0
```

For a lot of libraries, the version is missing. In some libraries, the version is incorrect or missing some parts of the version. Therefore, we always need to parse the filename to check whether there is valid information there.

For C#, the tuples look like this:

```
System.Runtime, 4.2.2.0
Castle.Core, 5.0.0.0
Microsoft.Extensions.Logging.Abstractions, 3.1.9.0
```

That means that in C# we actually have the name of the assembly instead of the filename and have a correct version. Therefore, we shouldn't do any parsing in C#.

I think that means this logic makes sense:
- We shouldn't always parse the filename because that breaks things for C#.
- If the `parseLibraryFilename` function can't find a correct version in the filename, it will just return the filename without its extension. This would be the same as we would be doing if we only returned anything when we could parse the version, so we would be duplicating this logic.

robertbrignull: Thanks for explaining. I agree that considering the data we get from the queries it does make sense to do this processing in these cases.

extensions/ql-vscode/src/data-extensions-editor/library.ts
```diff
@@ -0,0 +1,55 @@
...
+// This will match any semver string at the end of a larger string
+const semverRegex =
+  /[v=\s]*(?<version>([0-9]+)(\.([0-9]+)(?:\.([0-9]+)(?:-?((?:[0-9]+|\d*[a-zA-Z-][a-zA-Z0-9-]*)(?:\.(?:[0-9]+|\d*[a-zA-Z-][a-zA-Z0-9-]*))*))?(?:\+([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?)?)?)/g;
+
+export interface Library {
+  name: string;
+  version?: string;
+}
+
+export function parseLibraryFilename(filename: string): Library {
```
robertbrignull, line 19: The parsing often works, but it did get confused at least once on each of the repos I tried. It seems to mostly be going wrong when there's an extra suffix like a timestamp, `beta`, or `M2`. How hard would it be to cover these cases?

I'm not sure if we can make sure it's always perfect in every case because there will always be a library with a weird-enough version string, but we can certainly reduce the cases where a human looks at it and notices that it's off.

Here are some examples of the data extensions editor on a few different java projects:

- `jhy/jsoup`
    Screenshot 2023-07-14 at 12 25 37

- `google/gson`
    Screenshot 2023-07-14 at 12 26 32

- `apache/maven`
    Screenshot 2023-07-14 at 12 28 08

koesie10: I've made some changes to the regex that makes it more lenient in matching versions. It now works properly for all the examples in the screenshot and didn't regress in matching any of the other versions, so it should be better for matching versions overall.

robertbrignull: Thanks for making those changes. All the examples I just tried worked perfectly.

robertbrignull (approved):

PR #2569, opened 2023-06-29
Convert ResultsApp to a function component

robertbrignull:
Converts `ResultsApp` from a class component to a function component. This was fairly straightforward. I recommend reviewing commit-by-commit and using the feature to ignore whitespace.

The first commit moves functions around so that they aren't referenced before they are defined. This makes the future conversion much cleaner. And it turns out `getSortStates` and `getResultSet` don't actually use any references to the class, so those functions can be moved outside of the component.

The second commit converts to a function component. All functions still part of the component are converted to `useCallback`, though it turns out none of them have any dependencies so the function identities should never change. The props are removed because they are unused, and the state type is unchanged.

## Checklist


extensions/ql-vscode/src/view/results/results.tsx
```diff
@@ -70,239 +71,236 @@ export const onNavigation = new EventHandlerList<NavigateMsg>();
...
-          database: msg.database,
-          interpretation: msg.interpretation,
-          shouldKeepOldResultsWhileRendering:
-            msg.shouldKeepOldResultsWhileRendering,
-          metadata: msg.metadata,
-          queryName: msg.queryName,
-          queryPath: msg.queryPath,
-        });
+export function ResultsApp() {
+  const [state, setState] = React.useState<ResultsViewState>({
```
robertbrignull, line 75: I opted to keep the state as one big object, instead of decomposing into three individual state objects that could be set separately. The reason for this is that in `updateStateWithNewResultsInfo` we set all three fields at once in a few different ways, so it's helpful to set them all together.

The only other place we update the state is in `handleMessage` and there we do only update `isExpectingResultsUpdate`, but thankfully we can do
```typescript
setState((prevState) => ({
  ...prevState,
  isExpectingResultsUpdate: true,
}));
```
so setting just one field and not the others is simple.

---

We could also have taken the option to have three separate state fields. If you call multiple `setState` handlers in a single render / event handler then React will only trigger a single re-render to handle all of the state changes. So performance-wise it would be exactly the same, but I think having them together is maybe slightly easier to code in this case.

I'm open to both options, so I'd be happy to separate them if people think that would be better. It's also worth nothing we aren't locked into this, and we could separate them in the future if that becomes a better option.

norascheuch (approved):
Not very confident in my review but looks good from what I can see? 😬

PR #2571, opened 2023-06-30
Pull out AlertTableHeader to a separate component/file

robertbrignull:
This PR tries to pull out the table header to a separate file to make the `PathTable` component and file a bit smaller. This is a stepping stone to converting `PathTable` to a function component.

I think this is worth it because the props for the header are surprisingly small, and there are quite a few supporting functions that we can also move. So overall I think it removes a moderate amount of complexity from `PathTable`. It's unlikely that the `AlertTableHeader` component could be reused, but reuse wasn't really a driving factor and the main focus is on reducing the size of `alert-table.tsx`.

The new class is named `AlertTableHeader` rather than `PathTableHeader` because I'm assuming that https://github.com/github/vscode-codeql/pull/2570 will also be merged.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/results/alert-table-header.tsx
+++ b/extensions/ql-vscode/src/view/results/alert-table-header.tsx
@@ -42,11 +42,6 @@ export function AlertTableHeader({
     });
   }, [getNextSortState]);
 
-  const clickCallback = useCallback(
-    () => toggleSortStateForColumn(),
-    [toggleSortStateForColumn],
-  );
-
   return (
     <thead>
       <tr>
@@ -54,7 +49,7 @@ export function AlertTableHeader({
         <th
           className={`${sortClass()} vscode-codeql__alert-message-cell`}
           colSpan={3}
-          onClick={clickCallback}
+          onClick={toggleSortStateForColumn}
         >
           Message
         </th>
--- a/extensions/ql-vscode/src/view/results/alert-table-header.tsx
+++ b/extensions/ql-vscode/src/view/results/alert-table-header.tsx
@@ -1,0 +1,76 @@
+import * as React from "react";
+import { useCallback } from "react";
+import { vscode } from "../vscode-api";
+import {
+  InterpretedResultsSortColumn,
+  InterpretedResultsSortState,
+  SortDirection,
+} from "../../common/interface-types";
+import { nextSortDirection } from "./result-table-utils";
+
+export function AlertTableHeader({
+  sortState,
+}: {
+  sortState?: InterpretedResultsSortState;
+}) {
+  const sortClass = useCallback(
+    (column: InterpretedResultsSortColumn): string => {
+      if (sortState !== undefined && sortState.sortBy === column) {
+        return sortState.sortDirection === SortDirection.asc
+          ? "sort-asc"
+          : "sort-desc";
+      } else {
+        return "sort-none";
+      }
+    },
+    [sortState],
+  );
+
+  const getNextSortState = useCallback(
+    (
+      column: InterpretedResultsSortColumn,
+    ): InterpretedResultsSortState | undefined => {
+      const prevDirection =
+        sortState && sortState.sortBy === column
+          ? sortState.sortDirection
+          : undefined;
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/view/results/alert-table-header.tsx
```diff
@@ -0,0 +1,76 @@
...
+} from "../../common/interface-types";
+import { nextSortDirection } from "./result-table-utils";
+
+export function AlertTableHeader({
+  sortState,
+}: {
+  sortState?: InterpretedResultsSortState;
+}) {
+  const sortClass = useCallback(
+    (column: InterpretedResultsSortColumn): string => {
```
koesie10, line 17: Would it make sense to remove the `column` argument from this function, the `getNextSortState` function and the `toggleSortStateForColumn` function? It seems like it's only ever set to `alert-message`, so removing that argument may simplify this

extensions/ql-vscode/src/view/results/alert-table-header.tsx
```diff
@@ -0,0 +1,76 @@
...
+      if (sortState !== undefined && sortState.sortBy === column) {
```
koesie10, line 18: I believe this should be equivalent (according to the types where `column` cannot be `undefined`):

```suggestion
      if (sortState?.sortBy === column) {
```

extensions/ql-vscode/src/view/results/alert-table-header.tsx
```diff
@@ -0,0 +1,76 @@
...
+    (column: InterpretedResultsSortColumn): void => {
+      vscode.postMessage({
+        t: "changeInterpretedSort",
+        sortState: getNextSortState(column),
+      });
+    },
+    [getNextSortState],
+  );
+
+  const clickCallback = useCallback(
```
koesie10, line 55: I think a name like `handleClick` would be better, to match the `onClick` prop.

extensions/ql-vscode/src/view/results/alert-table-header.tsx
```diff
@@ -0,0 +1,76 @@
...
+    },
+    [sortState],
+  );
+
+  const getNextSortState = useCallback(
+    (
+      column: InterpretedResultsSortColumn,
+    ): InterpretedResultsSortState | undefined => {
+      const prevDirection =
+        sortState && sortState.sortBy === column
```
koesie10, line 34:
```suggestion
        sortState?.sortBy === column
```

koesie10 (approved):
The changes itself look good to me and both components seem to have the same behavior. I have some suggestions on how to improve the new component, but I'm also happy with leaving that for the future and just moving the functions around in this PR.

robertbrignull:
> I have some suggestions on how to improve the new component, but I'm also happy with leaving that for the future and just moving the functions around in this PR.

I'm happy to do a bit more cleanup and simplification in this PR. All of your suggestions look good to me, so I'll apply all of those.

extensions/ql-vscode/src/view/results/alert-table-header.tsx
robertbrignull: Yes, I agree we can remove the `column` argument. I considered whether we should refactor https://github.com/github/vscode-codeql/blob/6acfb8d15191c7cb9f7f286bb5d13d600139c1c7/extensions/ql-vscode/src/common/interface-types.ts#L277 to a constant so we can reference that here. I came to the conclusion that it's not worth it because these methods in `alert-table-header.tsx` are still referencing `InterpretedResultsSortState` so it will be a type error if we try to return or do a comparison against an incorrect value.

extensions/ql-vscode/src/view/results/alert-table-header.tsx
robertbrignull: Actually, now that `toggleSortStateForColumn` doesn't take any arguments, I think we can safely delete the `clickCallback`/`handleClick` method. We can instead pass in `toggleSortStateForColumn` directly to the `onClick` prop.

PR #2583, opened 2023-07-05
Show when there are unsaved changes to a model

robertbrignull:
Hook up the "Unsaved" label and "Save" button so it's only shown or enabled when you have made a change to that model.

There are probably a few different ways we could implement this, but I think we need to track the changes in the `DataExtensionsEditor` because that's where we're receiving messages from the extension host. The way that saving works is we send over the state we want to save but then the host send back a `setExternalApiUsages` message. The refresh also works by requesting a `setExternalApiUsages` message. So when we get one of those messages we know there are no unsaved changes, and then any change after that triggers the label to show.

## Checklist


shati-patel (approved):
Nice, seems to work well, from testing locally 🎉 (I don't have any strong opinions about where to implement it, so this looks good!)

PR #2588, opened 2023-07-06
Implement saving only one model at a time

robertbrignull:
Paired with @shati-patel on this.

Hooks up the "Save" button in each collapsible section to save only the changes made to that model. Thankfully this turned out not to be too hard since the implementation of `saveModeledMethods` already (perhaps unintentionally) handles being gives only the methods from one model, because it splits the methods it's given into models and then saves only those files. Therefore it won't overwrite or delete other models that aren't being saved.

We also had to change how `setUnsavedModels` is called. I think what we were doing before was not quite fully correct but we were getting away with it. Now that we can save just one model at a time we need to be more granular.

We've tested this manually but checking that each model is saved individually and none of the other files are edited. Is there any more testing we can do of this?

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/data-extensions-editor/DataExtensionsEditor.tsx
+++ b/extensions/ql-vscode/src/view/data-extensions-editor/DataExtensionsEditor.tsx
@@ -88,7 +88,6 @@ export function DataExtensionsEditor({
             break;
           case "setExternalApiUsages":
             setExternalApiUsages(msg.externalApiUsages);
-            setUnsavedModels(new Set());
             break;
           case "showProgress":
             setProgress(msg);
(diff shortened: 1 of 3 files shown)


charisk (approved):
LGTM! Are there any React/unit tests we can add around this area?

robertbrignull:
> Are there any React/unit tests we can add around this area?

I think the answer is that right now, no not really. It's kinda hard.

There are currently no tests of the `data-extensions-editor` react code, but we could add some fairly easily. However I don't think testing this particular UI change it too useful. Adding tests of other existing functionality would be useful, but that's outside the scope of this PR.

There are also no tests of `DataExtensionsEditorView`, and adding some looks hard. I think we want to move implementation out of this class and into separate classes that'll be more testable, but I'm hesitant to do that in this PR. Would it be ok if I open an another issue to cover moving all of the implementation for loading/saving modelled data to a separate file and testing it?

charisk:
> > Are there any React/unit tests we can add around this area?
> 
> I think the answer is that right now, no not really. It's kinda hard.
> 
> There are currently no tests of the `data-extensions-editor` react code, but we could add some fairly easily. However I don't think testing this particular UI change it too useful. Adding tests of other existing functionality would be useful, but that's outside the scope of this PR.
> 
> There are also no tests of `DataExtensionsEditorView`, and adding some looks hard. I think we want to move implementation out of this class and into separate classes that'll be more testable, but I'm hesitant to do that in this PR. Would it be ok if I open an another issue to cover moving all of the implementation for loading/saving modelled data to a separate file and testing it?

Yeah that makes sense! Happy to improve on test coverage separately.

PR #2591, opened 2023-07-07
Use AI to model just a single section

robertbrignull:
Hooks up the "Model with AI" in each collapsible section to work on just methods from that model. Follows a very similar approach to the "save" button. This thankfully is quite easy since we pass all the data back to the extension host, so we can just send only the data for that model. From then on the extension host sends back some modelled methods and it all works the same as it did before.

The old "Generate using LLM" in the header still works on everything. This button will be removed at a later date once we've finished redesigning the collapsible sections.

Note: the "unsaved" label is a little broken and doesn't update correctly when modelling with AI. It's already broken before this change too so so my rough plan was to go forward with this PR and address it in a separate PR. That is, to make the "unsaved" status is affected correctly when we get new changes from "model from source" or "model with LLM". But just say if you'd prefer to delay this PR and wait for that fix first.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/data-extensions-editor/DataExtensionsEditor.tsx
+++ b/extensions/ql-vscode/src/view/data-extensions-editor/DataExtensionsEditor.tsx
@@ -185,13 +185,27 @@ export function DataExtensionsEditor({
     });
   }, []);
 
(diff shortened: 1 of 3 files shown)


charisk (approved):
LGTM! 🎉

PR #2594, opened 2023-07-10
Include in addModeledMethods whether the methods are unsaved or not

robertbrignull:
This is an attempt to fix the behaviour around showing the "unsaved" label and enabling the save button. Previously it wasn't correctly showing a model as unsaved when generating from the LLM or modelling from source.

The approach in this PR works, but it feels a bit inelegant and brute-forcey. I also don't love that the listener has gained dependencies on a couple of different bits of state, but maybe it's ok as those bits of state don't actually change too often.

Any other suggestions for how to implement this? I tried a few variations of putting names of the affected models in `AddModeledMethodsMessage`, or putting the library/packageName on `ModeledMethod`, but both turned out to be pretty hard to do as we just don't have the info available when generating the modelled methods.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/interface-types.ts
+++ b/extensions/ql-vscode/src/common/interface-types.ts
@@ -509,21 +509,14 @@ export interface ShowProgressMessage {
   message: string;
 }
 
+export interface LoadModeledMethodsMessage {
+  t: "loadModeledMethods";
+  modeledMethods: Record<string, ModeledMethod>;
+}
+
 export interface AddModeledMethodsMessage {
   t: "addModeledMethods";
   modeledMethods: Record<string, ModeledMethod>;
-  /**
-   * Are these modeled methods newly generated in some way and not yet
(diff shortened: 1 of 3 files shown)


robertbrignull:
> Have you considered splitting the messages for loading vs adding new modeled methods?

It's in intriguing idea. It may be worth doing as a refactor to make things easier to understand, so we aren't using one message type for two purposes. However I don't think it helps with the "inelegance" of needing to loop through all of `externalApiUsages` as the message for adding new methods would still need this logic.

I think I'll implement this approach in this PR because it will avoid adding a new `unsaved: boolean` field to the message, so in that way it is strictly cleaner than the current state of this PR.

robertbrignull:
I've split `addModeledMethods` up into `addModeledMethods` and `loadModeledMethods`. This actually means we can remove `overrideNone` too because that was only set to true in the `addModeledMethods` case. I think it works out quite well.

charisk (approved):
Thanks for making the changes, I think that looks great!

PR #2597, opened 2023-07-11
Clear up how we pass around view state

robertbrignull:
I realised we were passing in both `ViewState` and `Mode` to some components. This is a mistake I made earlier and didn't realise the view state was already available.

I also take the opportunity to simplify the types and avoiding passing a possibly-undefined view state around as much as possible. This shouldn't ever be undefined in practice because the first thing we do once the view is loaded is to send a `setDataExtensionEditorViewState` message. The fact that we have to do this is a drawback of the webview model where we have to send initialisation data only after the view is loaded. I think we can treat the brief window where the view state is undefined as a "loading" phase and display a message. This way for the rest of the code we are sure that the variable is defined. In practice this loading message is invisible and I can't spot it when the view is opened.

## Checklist


charisk (approved):
This is great! Thank you for tidying things up and for the nice clear commits.

RubyRusul (commented):
Hi ,,😳👌

PR #2598, opened 2023-07-11
Don't show model from source button in framework mode

robertbrignull:
Builds upon https://github.com/github/vscode-codeql/pull/2597

Avoids showing the "Model from source" button on each collapsible section when in framework mode.

The reason for this is that when in framework mode we're only dealing with a single model file and a single github repository. The view splits up methods by package instead of dependency/model. Using the "Model from source" button affects the whole model and not just the package it was clicked on. Therefore I don't think it makes sense to have this button on each section and we should stick with the single button in the header.

My plan is to review this decision next week along with @asiermartinez, but to avoid being held up right now I'm just making changes how I think things should be. This UI change is very easily reversible if we decide next week to do something different.

## Checklist


charisk (approved):
Makes sense!

asiermartinez:
I'd love to learn more about this. I'm still confused own how we do the modeling (what is the smallest package we want to encourage users to use) because I feel like so far we have been using the word dependency, framework or library as synonyms and everything is starting to look more nuanced now.

PR #2600, opened 2023-07-12
Show more clearly when a method is already modeled

robertbrignull:
This PR does some refactoring of the `MethodRow` component to split it into two cases:
- When the method can be modeled, which is most of the implementation.
- When the method cannot be modeled, because it is already modeled by another source such CodeQL or another extension pack than the one we're currently editing.

We then finally add a simple message saying that a method is already modeled and cannot be edited.

Screenshot 2023-07-12 at 14 59 04

The refactoring is not strictly necessary, and I can remove it or pull it out to two separate PRs if that would be easier. What I like about the refactoring and splitting into two components is that we don't have to define all the various handlers and other computation for the unmodelable case where we don't use any of them.

Note that right now I don't think we can tell between a method modeled by CodeQL or another extension pack. I'd like to make that more clear but I think it would be better as a separate PR. Therefore the message right now has to cover both cases. Suggestions on the wording of the message are very welcome.

Also note, I've made a separate issue for making the checkboxes work and match the designs. So ignore those for now.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/data-extensions-editor/MethodRow.tsx
+++ b/extensions/ql-vscode/src/view/data-extensions-editor/MethodRow.tsx
@@ -60,28 +60,19 @@ type Props = {
 export const MethodRow = (props: Props) => {
   const { externalApiUsage, modeledMethod } = props;
 
-  const jumpToUsage = useCallback(() => {
-    vscode.postMessage({
-      t: "jumpToUsage",
-      // In framework mode, the first and only usage is the definition of the method
-      location: externalApiUsage.usages[0].url,
-    });
-  }, [externalApiUsage]);
-
   const methodCanBeModeled =
     !externalApiUsage.supported ||
     (modeledMethod && modeledMethod?.type !== "none");
 
   if (methodCanBeModeled) {
-    return <ModelableMethodRow {...props} jumpToUsage={jumpToUsage} />;
+    return <ModelableMethodRow {...props} />;
   } else {
-    return <UmmodelableMethodRow {...props} jumpToUsage={jumpToUsage} />;
+    return <UmmodelableMethodRow {...props} />;
   }
 };
 
-function ModelableMethodRow(props: Props & { jumpToUsage: () => void }) {
-  const { externalApiUsage, modeledMethod, mode, onChange, jumpToUsage } =
-    props;
+function ModelableMethodRow(props: Props) {
+  const { externalApiUsage, modeledMethod, mode, onChange } = props;
 
   const argumentsList = useMemo(() => {
     if (externalApiUsage.methodParameters === "()") {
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/view/data-extensions-editor/MethodRow.tsx
```diff
@@ -172,33 +169,8 @@ export const MethodRow = ({
...
-          )}
-          <ViewLink onClick={jumpToUsage}>View</ViewLink>
-        </ApiOrMethodCell>
-        <VSCodeDataGridCell gridColumn={2} />
-        <VSCodeDataGridCell gridColumn={3} />
-        <VSCodeDataGridCell gridColumn={4} />
-        <VSCodeDataGridCell gridColumn={5} />
-      </VSCodeDataGridRow>
-    );
+    return <UmmodelableMethodRow {...props} jumpToUsage={jumpToUsage} />;
```
norascheuch: Typo: `return <UnmodelableMethodRow {...props} jumpToUsage={jumpToUsage} />;`

extensions/ql-vscode/src/view/data-extensions-editor/MethodRow.tsx
```diff
@@ -257,3 +229,39 @@ export const MethodRow = ({
     </VSCodeDataGridRow>
   );
 };
+
+function UmmodelableMethodRow(props: {
```
norascheuch, line 236: Same typo, see above

extensions/ql-vscode/src/view/data-extensions-editor/MethodRow.tsx
```diff
@@ -256,10 +256,9 @@ function UmmodelableMethodRow(props: {
         )}
         <ViewLink onClick={jumpToUsage}>View</ViewLink>
       </ApiOrMethodCell>
-      <VSCodeDataGridCell gridColumn={2} />
-      <VSCodeDataGridCell gridColumn={3} />
-      <VSCodeDataGridCell gridColumn={4} />
-      <VSCodeDataGridCell gridColumn={5} />
+      <VSCodeDataGridCell gridColumn="span 4">
+        Method modeled by CodeQL or another extension pack
```
norascheuch, line 260: We could say  'Method `already` modeled by CodeQL or ~another~ `a different` extension pack' but I have no real preference, just because you asked for suggestions.

norascheuch (approved):
I'm a bit confused by the diff of 'Pull out ModelableMethodRow as a separate component' but I think it looks good, otherwise just 2 minor comments. Looks good!

robertbrignull:
> I'm a bit confused by the diff of 'Pull out ModelableMethodRow as a separate component' but I think it looks good

Yeah it's a little hard to understand because the git diff algorithm realised it's less change to say we split the `MethodRow` function into two and added some more implementation at the top, instead of shortening `MethodRow` by pulling out virtually all of its contents into a new function. In that case maybe easier to just review the final state and ignore the diff.

PR #2605, opened 2023-07-13
Refactor the code for loading/saving modeled methods to disk, and add tests

robertbrignull:
The main aim of this PR is to move some code out of `data-extensions-editor-view.ts` to make that file smaller, and add tests of code where possible. In the end I'm not sure it's adding too much, so if you think this change is not worth it please just say.

The functionality that I've moved to other files is:
- `createDataExtensionYamls`, which just switches on the `Mode` and calls the right function. I think this can sensibly live in `yaml.ts`.
- `saveModeledMethods` and `loadModeledMethods`, which deal with actually reading and writing to files on disk. I think moving these function out of the view is useful, but I can see why we wouldn't want these in `yaml.ts` because we want to keep that file "pure" and not interact with actual files there. I wasn't sure what to name this new file, so suggestions here are very welcome.

The only part I could sensibly write a new test for was `listModelFiles`. I think a CLI test maybe makes sense here, since the complex part is in CodeQL and making sure we're handling its output correctly, so if we mock the CodeQL responses I'm not sure the test is worth writing anymore.

The rest of the loading and saving is pretty simple and the logic in `yaml.ts` is already heavily tested, so I don't think we gain much by adding tests of `saveModeledMethods` and `loadModeledMethods`.

## Checklist


koesie10 (approved):
LGTM, thanks for cleaning this up!

robertbrignull:
I was hitting problems because `codeql resolve extensions` only exists in CodeQL CLI versions v2.10.2 or later, and so it was failing some of the CLI tests. I think we've not noticed this before because we did have any CLI tests using this command, and when testing locally we all had new-enough CLI versions.

I've added a CLI version check when calling the command and when starting up the data extensions editor. Let me know if you think that's ok.

robertbrignull:
The un-awaited functions should be fixed now.

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1274,6 +1274,12 @@ export class CodeQLCliServer implements Disposable {
     suite: string,
     additionalPacks: string[],
   ): Promise<ResolveExtensionsResult> {
+    if (!(await this.cliConstraints.supportsResolveExtensions())) {
+      throw new Error(
+        "Resolving extensions is only supported for CodeQL CLI v2.10.2 or later",
+      );
+    }
+
```
koesie10, line 1282: It seems like we never have a check in `cli.ts` for whether a method is supported which throws an error. We only seem to use it if certain arguments are not supported or if the arguments are different, but in those cases we only log or change the arguments. Could we remove this and instead rely on the consumer to do this check?

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts
```diff
@@ -54,6 +54,16 @@ export class DataExtensionsEditorModule {
...
+        if (
+          !(await this.cliServer.cliConstraints.supportsResolveExtensions())
+        ) {
+          void showAndLogErrorMessage(
+            this.app.logger,
+            "CodeQL CLI version v2.10.2 or later is required to use the data extensions editor.",
+          );
+          return;
+        }
+
```
koesie10, line 66: Could we move this check to be in the same location as the other check (`supportsQlpacksKind`)? Could we then also use `CLI_VERSION_WITH_RESOLVE_EXTENSIONS.format()` instead of hardcoding `v2.10.2`?

extensions/ql-vscode/src/codeql-cli/cli.ts
robertbrignull: I wasn't sure what to do here. I agree there aren't any other examples of a method throwing an error, and all the others fall back to another implementation if the version is too low. I wasn't sure if that was a deliberate design choice, or just coincidence that all other examples of versions checks have a way around them.

But I'm happy to remove this and rely on the caller to check compatibility. If it does fail then the error is pretty clear and we'd know what was wrong.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts
robertbrignull: Yep, sorry I didn't spot that we're already doing a CLI version check here.

We now have two checks next to each other. I've put the one with the stricter version requirement first, it removes the possibility of upgrading to pass the first check and then just hitting the second check.

So actually we were already protected because we already require `v2.12.3`. I'm tempted to leave both checks in, even though the second one is incapable of failing, because it acts as internal documentation that we're using a particular feature. When the CodeQL CLI version advances enough that we can assume `v.2.10.2` then we can remove the check for that.

Do you think this makes sense or does it offer no benefit and we should just remove the weaker version check?

koesie10: That makes sense to me, it makes sense to be explicit about which features (and thus versions) we require.

robertbrignull:
Something is going wrong on windows still. I'll take a look into it.

robertbrignull:
Hmm, still getting semi-random failures:
```
  ● listModelFiles › should ignore model files from other extension packs

    expect(received).toEqual(expected) // deep equality

    - Expected  - 3
    + Received  + 1

    - Set {
    -   "c:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp-5664eE9F5BKpjKIg\\workspace\\extension-pack\\models\\library1.model.yml",
    - }
    + Set {}

      134 |
      135 |     const modelFiles = await listModelFiles(extensionPackPath, cli);
    > 136 |     expect(modelFiles).toEqual(
          |                        ^
      137 |       new Set([join(extensionPackPath, "models", "library1.model.yml")]),
      138 |     );
      139 |   });
```
I'll need to take a look at what is in the response from CodeQL. Maybe the problem is still to do with windows path matching or maybe it's something else.

koesie10:
> I'll need to take a look at what is in the response from CodeQL. Maybe the problem is still to do with windows path matching or maybe it's something else.

In case it helps, there's [a `normalizePath` function](https://github.com/github/vscode-codeql/blob/6cfc7d5cedb6adbd2bf359f617f5e4f328f3c9af/extensions/ql-vscode/src/common/files.ts#L55) which should automatically resolve any differences in Windows paths. This is also used for `pathsEqual` and the `toEqualPath` Jest matcher.

robertbrignull:
> In case it helps, there's [a `normalizePath` function](https://github.com/github/vscode-codeql/blob/6cfc7d5cedb6adbd2bf359f617f5e4f328f3c9af/extensions/ql-vscode/src/common/files.ts#L55) which should automatically resolve any differences in Windows paths. This is also used for `pathsEqual` and the `toEqualPath` Jest matcher.

I agree it's the path matching that seems to be going wrong. I tried using `pathsEqual` but it wasn't quite enough. The data looks like
```
Expected:
c:\Users\RUNNER~1\AppData\Local\Temp\tmp-6008rRF30qAr4H3s\workspace\extension-pack
    

Received:
{
  "models":[],
  "data":{
    "c:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp-6008rRF30qAr4H3s\\workspace\\extension-pack":[
      {
        "predicate":"neutralModel",
        "file":"c:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp-6008rRF30qAr4H3s\\workspace\\extension-pack\\models\\library1.model.yml",
        "index":3
      },
    ]
  }
}
```
It appears there's some extra encoding going on meaning we have double backslashes. I'm not sure if this is problem or not when running the production code or if `readFile` handles these slashes for us. I've added a test of `loadModeledMethods` which should hopefully show if that part is or isn't working on windows.

robertbrignull:
Hmm...
```
-   "c:\\users\\runner~1\\appdata\\local\\temp\\tmp-32602krclotpcknf\\workspace\\extension-pack"
+   "c:\\users\\runneradmin\\appdata\\local\\temp\\tmp-32602krclotpcknf\\workspace\\extension-pack"
```
It now seems the user name and hence the whole path is being changed 😱

robertbrignull:
It seems that `os.tmpdir()` is returning the `runner~1` path, and this differs from `os.homedir()`.
```
    - Expected  - 2
    + Received  + 2

      Array [
    -   "c:\\Users\\runneradmin",
    -   "c:\\Users\\runneradmin\\AppData\\Local\\Temp",
    +   "C:\\Users\\runneradmin",
    +   "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp",
      ]

      170 |
      171 |     it("should have correct env vars (debugging only)", async () => {
    > 172 |       expect([homedir(), tmpdir()]).toEqual([
          |                                     ^
      173 |         "c:\\Users\\runneradmin",
      174 |         "c:\\Users\\runneradmin\\AppData\\Local\\Temp",
      175 |       ]);
```

And this matches the environment variables you get on windows:
```
2023-07-17T14:10:55.0785345Z TEMP                           C:\Users\RUNNER~1\AppData\Local\Temp
2023-07-17T14:10:55.0787364Z TMP                            C:\Users\RUNNER~1\AppData\Local\Temp
2023-07-17T14:10:55.0793705Z USERPROFILE                    C:\Users\runneradmin
```
And a test in another repo has confirmed that these directories are actually the same. If you write a file using one path you can then read it using the other path. They look like symlinks so I don't know why `realpath` isn't converting both into the same canonical directory name.

I'm considering putting in a manual hack in the tests to work around this. I think the `runneradmin` one is the true canonical directory, so if we use that from the start it should be fine.

robertbrignull:
@koesie10, sorry for all the noise on this PR, but I think I've got there at last. I'm still not sure exactly what's going on. I'm running into the limit of my windows knowledge as to why `C:\Users\runneradmin` and `C:\Users\runner~1` are the same.

Is there a better workaround we could do than specifying `dir` to `tmp.dirSync`? According to the docs, it should complain if this directory is outside of the OS temp dir, so this should be safe-ish. I'm not sure if there's anything we can or should do in the production code, rather than making sure that the test setup doesn't hit this case.

I'm moderately sure this isn't a bug that has been introduced by this PR but is instead being exposed because we're writing tests for it. It might only manifest on actions runners because they have this setup of homedir and tempdir.

robertbrignull:
There has been a lot of force pushing, but I think I've successfully avoided changing any of the earlier commits. I've now removed my rough testing commits, so the history should be clean now.

PR #2606, opened 2023-07-13
Add signature and related fields to ModeledMethod

robertbrignull:
I was originally trying to do changes around loading/saving of modeled methods, and it would have been useful to have the signature information directly on `ModeledMethod` because I didn't have any `ExternalApiUsage` objects around. However when I tried doing that I realised this refactor potentially makes a lot of things easier and we can remove three other types that are only there because this information isn't available on `ModeledMethod` itself.

I think adding these fields to `ModeledMethod` shouldn't pose a problem now or going forward. All of this data is already needed to calculate the `signature`, and we always do that to use as the map key.

I recommend reviewing each commit individually.

## Checklist


extensions/ql-vscode/src/data-extensions-editor/external-api-usage.ts
```diff
@@ -25,10 +21,17 @@ export type ExternalApiUsage = {
    * The method parameters, including enclosing parentheses, e.g. `(String, String)`
    */
   methodParameters: string;
+}
+
+export interface ExternalApiUsage extends MethodSignature {
```
robertbrignull, line 26: This could be converted to `type ExternalApiUsage = MethodSignature & {...` if we want to keep that format. We seem to to not be using `interface` in the data extensions code.

robertbrignull:
I checked to see if any `Record<string, ModeledMethod>` types could be converted to just `ModeledMethod[]` and I don't think they can. We do actually always use it to look up methods by their signature.

koesie10 (commented):
This definitely seems to simplify things, but I'm wondering if having the `MethodSignature` being a field on the `ModeledMethod` and `ExternalApiUsage` would make sense? That would make it easier to move between the two types since you don't need to extract 5 fields, but would only need to extract 1 field.

robertbrignull:
> having the `MethodSignature` being a field on the `ModeledMethod` and `ExternalApiUsage` would make sense?

That makes sense to me and I'd be happy to go with that approach. I'll have to think up a better name so we don't end up going `method.signature.signature`.

robertbrignull:
Hmm, upon trying to do this, I'm not sure it's worth it. If it's ok I'd prefer to keep the types as they are now.

Having `MethodSignature` as a field makes [one bit in `MethodRow`](https://github.com/github/vscode-codeql/pull/2606/files#diff-c7ff8716caebaa2d785509a50b64647e5d658325440695381d3ef0734b6c5707R104) easier where we can do `method: externalApiMethod.method` instead of writing out each field. But it makes everywhere else a little bit harder and more verbose and not necessarily easier to read.

koesie10 (approved):
> Hmm, upon trying to do this, I'm not sure it's worth it. If it's ok I'd prefer to keep the types as they are now.

Thanks for trying it! I'm happy with this version as well and the code LGTM.

PR #2621, opened 2023-07-19
Load existing modeled methods before saving, to avoid overwriting data

robertbrignull:
The idea of this PR is to avoid the possibility of losing data when writing data extension YAML files. We read the modeled methods in and then merge the existing contents with the new contents. As you can see from the tests, the behaviour when there are no existing modeled methods is the same, and there are new tests for when there are existing modeled methods.

The feature that's pushing us to implement is to allow saving only some of the modeled methods from a model at a time. This will be necessary to support framework mode where we want to only save methods from a single package.

I've also tested this locally and I believe it's doing the right thing. If I modify or add a new model file while the extensions editor is open, it will not delete the new file and any changes will be merged into the existing content. Framework mode doesn't quite work fully yet so I can't test it there, but implementing framework mode fully is the next thing I will look at.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
@@ -123,6 +123,7 @@ export class DataExtensionsEditorView extends AbstractWebview<
           msg.externalApiUsages,
           msg.modeledMethods,
           this.mode,
+          this.cliServer,
           this.app.logger,
         );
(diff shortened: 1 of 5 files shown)


robertbrignull:
I realise this is a bit of a change to behaviour. I could try to get the framework mode PR up so you can see how this new behaviour will be used.

robertbrignull:
Actually, framework mode, by clicking "switch mode" is working pretty well with this PR and saving individual sections is working as intended.

koesie10 (approved):
These changes make sense to me

PR #2623, opened 2023-07-20
Implement "model dependency" button in application mode

robertbrignull:
This PR adds the implementation for the "model dependency" button. This prompts the user for a repository name, then downloads that repository and opens a new data extensions window in framework mode.

This shares a bunch of implementation with the "model from source" button because they both prompt the user to download a new database. I think both of these will continue to share implementation, so any improvements we make will apply to both.

I have encountered problems when the new databases causes the workspace to transition from single-root to multi-root. I have some ideas for things we can do about this, and these might help with "model from source" as well if that shares the same issue. However I'm thinking about raising those changes as separate PRs.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/interface-types.ts
+++ b/extensions/ql-vscode/src/common/interface-types.ts
@@ -553,6 +553,10 @@ export interface GenerateExternalApiFromLlmMessage {
   modeledMethods: Record<string, ModeledMethod>;
 }
 
+export interface ModelDependencyMessage {
+  t: "modelDependency";
+}
+
 export type ToDataExtensionsEditorMessage =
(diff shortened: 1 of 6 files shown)


shati-patel (approved):
Code looks sensible and we've tested it on my (Windows) machine too 🪟 

Regarding the single-folder workspace bug: this is a more general problem, so I don't think it's worth addressing in this PR. We'll probably have to do more work to get the Data Extensions Editor working in that case (if we even want to support single-folder workspaces full). 

Also: as discussed offline, we weren't sure if/how to test this 🧪 Seems fine to merge as-is, and we can think about that later 🤔

PR #2624, opened 2023-07-21
Delete autoClosingPairs from LanguageConfiguration to appease proposed API

robertbrignull:
This allow the extension to start when using VS Code Insiders version `1.81.0-insider`. We're running afoul of the new [proposed API](https://code.visualstudio.com/api/advanced-topics/using-proposed-api) for [autoClosingPairs](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.languageConfigurationAutoClosingPairs.d.ts).

Once the proposed API has been merged and is supported by all VS Code versions we aim to support, we can remove this line and pass in `autoClosingPairs` in the `LanguageConfiguration` object.

I've tested it and we do still get `autoClosingPairs` behaviour because we're still passing it in [language-configuration.json](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/language-configuration.json). This seems to still work and it accepts the value and provides the desired behaviour.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/language-support/language-support.ts
+++ b/extensions/ql-vscode/src/language-support/language-support.ts
@@ -20,6 +20,7 @@ export function install() {
     decreaseIndentPattern: /^((?!.*?\/\*).*\*\/)?\s*[\}\]].*$/,
     increaseIndentPattern: /^((?!\/\/).)*(\{[^}"'`]*|\([^)"'`]*|\[[^\]"'`]*)$/,
   };
+  delete langConfig.autoClosingPairs;
 
   languages.setLanguageConfiguration("ql", langConfig);
   languages.setLanguageConfiguration("qll", langConfig);


extensions/ql-vscode/src/language-support/language-support.ts
```diff
@@ -20,6 +20,7 @@ export function install() {
     decreaseIndentPattern: /^((?!.*?\/\*).*\*\/)?\s*[\}\]].*$/,
     increaseIndentPattern: /^((?!\/\/).)*(\{[^}"'`]*|\([^)"'`]*|\[[^\]"'`]*)$/,
   };
+  delete langConfig.autoClosingPairs;
```
aeisenberg, line 23: It would be unfortunate if we have to lose this functionality in regular vscode as well. 

According to copilot, this snippet should let us determine what kind of vscode we are running in. Maybe we can use this to conditionally delete the `autoClosingPairs`. 

```js
import * as vscode from 'vscode';

if (vscode.env.appName === 'Visual Studio Code') {
  console.log('Extension is running in Visual Studio Code');
} else if (vscode.env.appName === 'Visual Studio Code - Insiders') {
  console.log('Extension is running in Visual Studio Code Insiders');
} else {
  console.log('Extension is running in an unknown application');
}
```

Though at some point, this needs to be reimplemented using the new API.

aeisenberg: Oh...I missed the fact that we're still getting the auto-close behaviour. Since this file is only used for the `onEnterRules`.

aeisenberg (approved):
Probably fine to go in as-is.

extensions/ql-vscode/src/language-support/language-support.ts
robertbrignull: Yeah the functionality still works, (at least it does when I test it), because `autoClosingPairs` is still in the json file.

Maybe we're skirting the rules by keeping `autoClosingPairs` in the json file, since it means we're using a proposed API feature but in a way where they don't catch it and stop us. Or maybe the json file API is genuinely a different API and that one isn't changing, so it's all fine. That would make sense since it would mean that this proposed API change is bringing the two APIs closer together in terms of features. I haven't interacted with this area before and I don't know enough to say either way.

robertbrignull:
Confirmed again that this starts correctly and the auto-closing pairs functionality works on both normal and insiders versions of VS Code.

PR #2626, opened 2023-07-21
Instead of resetting the database after importing, pass through makeSelected = false

robertbrignull:
For "model from source" and "model dependency" we were importing the new database, setting it as the current database, and then resetting the current database back to what it was before. Instead we can pass through `makeSelected = false` all the way to `openDatabase` and therefore not have to remember and reset the current database.

Needs some testing for edge cases, but it seems to work for me.

## Checklist


shati-patel (approved):
LGTM! 🎈 

> Needs some testing for edge cases, but it seems to work for me.

Anything in particular? I did some testing, but definitely haven't covered every case!

robertbrignull:
> Anything in particular? I did some testing, but definitely haven't covered every case!

Nothing in particular, just a bit more than I did. But if you've done it a couple of times and I did it a bit then I think that's enough to show it's working as well as it was before.

PR #2628, opened 2023-07-21
Merge all our Disposable definitions

robertbrignull:
We had three identical versions of `Disposable` types, all with the intention of avoiding a dependency on VS Code. I propose we can merge them all and just use the one from `disposable-objects.ts`. That file is already in `common` and has a comment at the top saying to avoid dependencies on VS Code.

## Checklist


aeisenberg (approved):

extensions/ql-vscode/src/packages/commands/Disposable.ts
charisk, line 1: I think the intention with `packages/commands` is that it's a stand-alone package that we could potentially publish to npm one day. So I think we need to leave this implementation here.

robertbrignull: Ok, I'll revert that one.

It would be better for the future if this requirement were enforced or documented in the code, so we don't have to rely on remembering it.

charisk: Yep, agreed!

robertbrignull:
PR is updated to just remove the instance from `log-scanner.ts`

charisk (approved):
Thank you for tidying that up!

PR #2631, opened 2023-07-24
Add check for deadcode for CI

robertbrignull:
This PR installs `ts-unused-exports` and uses it to find dead code. This PR doesn't fix any existing violations, so the CI will fail until we fix those in https://github.com/github/vscode-codeql/pull/2632.

The specific thing that this tools helps find is code that is exported but then not imported from anywhere else. This either means the code can be unexported if it's still used within its file, or it can be deleted.

By virtue of `typescript-eslint/no-unused-vars` we already have detection of unexported code (including functions/types/variables) that is not used elsewhere within its file. Using `ts-unused-exports` allows us to go one step further and look for unnecessary exports, and then once unexported our existing eslint rules will find any unused code.

## Checklist


extensions/ql-vscode/find-deadcode-tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
```
robertbrignull, line 1: One problem I had writing this was that our codebase is split into multiple different projects with multiple `tsconfig.json` files that all depend on each other, and all the tools only operate on a single project at a time. Unfortunately just looking at one project you can't tell if a piece of code is unused or not, because it could be used from another project that depends on this one. To work around this I created a new `find-deadcode-tsconfig.json` file that covers the entire repository. I'm not totally sure this means the dead code detection is 100% accurate, and there could be edge cases since the different projects use different rules, but I think it's likely to be good enough.
(Side note: it can't be named `tsconfig.json` because that file already exists in that directory. I could put it in another directory, but I'm not sure which one.)

robertbrignull:
I did try out other tools but I found that `ts-unused-exports` worked the best.
- `import/no-unused-modules` just wasn't working for me. It wasn't detecting any unused exports, and I know that there are lots in the repo but the other methods did detect them. I'm not sure what's going on with this.
- `ts-prune` worked better, but sadly it appears it doesn't understand `.tsx` files, maybe. It was declaring lots of react components to be unused when they clearly were used from other components.

extensions/ql-vscode/package.json
```diff
@@ -1732,6 +1732,7 @@
     "format": "prettier --write **/*.{ts,tsx} && eslint . --ext .ts,.tsx --fix",
     "lint": "eslint . --ext .js,.ts,.tsx --max-warnings=0",
     "lint:markdown": "markdownlint-cli2 \"../../**/*.{md,mdx}\" \"!**/node_modules/**\" \"!**/.vscode-test/**\" \"!**/build/cli/v*/**\"",
+    "find-deadcode": "ts-unused-exports find-deadcode-tsconfig.json --showLineNumber --excludePathsFromReport=$(find src test -type f '(' -name jest.config.ts -or -name 'index.tsx' -or -name 'index.ts' ')' -print0 | tr '\\0' ';')'gulpfile.ts/;src/stories/;test/vscode-tests/jest-runner-installed-extensions.ts'",
```
robertbrignull, line 1735: There are quite a few files that either have to have unused exported items by design (e.g. the `index.tsx` files or anything in `src/stories/`), and perhaps that we just don't care as much about so it's easier to ignore them to avoid jumping through hoops. Feel free to say if you think this list is not correct.

koesie10, line 1735: Would it make sense to use a script to run this so it's also compatible with Windows and a little bit more readable? It seems like [you can use `ts-unused-exports` as a library](https://github.com/pzavolinsky/ts-unused-exports#usage-as-a-library---from-typescript-was-built-with-es5--commonjs).

extensions/ql-vscode/find-deadcode-tsconfig.json
koesie10, line 1: We have some other alternative `tsconfig.json` files as well, but these are named `tsconfig.lint.json` and `tsconfig.spec.json`. Could we call this one `tsconfig.deadcode.json`?

koesie10 (commented):
Thanks for adding this!

extensions/ql-vscode/find-deadcode-tsconfig.json
robertbrignull: Oh interesting! I haven't spotted those but yes we could definitely name it like that.

koesie10 (approved):
LGTM, thanks for converting it to a script!

PR #2633, opened 2023-07-25
Add LLM functionality using auto-model V2

koesie10:
This updates the LLM functionality to use the new request/response format and the query included in codeql when the `codeQL.dataExtensions.llmGenerationV2` config setting is enabled.

The first two commits clean up the query resolver for contextual queries so that it's easier to re-use that functionality for resolving non-contextual queries. The last commit uses this to do the following:
- Resolve the query with tags `automodel <application-mode/framework-mode> candidates` from the CodeQL libraries (either submodule or query pack)
- Run the query
- Convert the results to SARIF
- Encode the SARIF as JSON
- GZIP compress the JSON string
- Base64-encode the GZIPed JSON
- Send the request
- Parse the YAML returned from the API
- Send the modeled methods to the view

## Checklist


charisk:
Any chance you could have a separate PR for the refactoring (first two commits)? It'd be easier for me personally but it would also probably bring feedback from others too since the query resolver/contextual queries isn't really an area we own/touched on a lot as a team.

extensions/ql-vscode/src/common/zlib.ts
```diff
@@ -0,0 +1,43 @@
+import { gzip, gunzip, InputType as ZlibInputType, ZlibOptions } from "zlib";
+
+/**
+ * Promisified version of zlib.gzip
+ * @param buffer Buffer to compress
+ * @param options zlib options
+ */
+export function gzipEncode(
```
charisk, line 8: Can we use [promisify](https://nodejs.org/api/util.html#utilpromisifyoriginal) instead?

extensions/ql-vscode/src/data-extensions-editor/auto-model-codeml-queries.ts
```diff
@@ -0,0 +1,223 @@
...
+import { getOnDiskWorkspaceFolders } from "../common/vscode/workspace-folders";
+import { showAndLogExceptionWithTelemetry, TeeLogger } from "../common/logging";
+import { QueryResultType } from "../query-server/new-messages";
+import { telemetryListener } from "../common/vscode/telemetry";
+import { redactableError } from "../common/errors";
+import { interpretResultsSarif } from "../query-results";
+import { join } from "path";
+import { assertNever } from "../common/helpers-pure";
+
+type Options = {
```
charisk, line 21: I don't quite get why we're calling things "Options" - to me they look like normal method inputs. Also it's hard to know the difference between `Options` and `AutoModelQueryOptions`

extensions/ql-vscode/src/data-extensions-editor/auto-model-codeml-queries.ts
```diff
@@ -0,0 +1,223 @@
...
+      redactableError`Auto-model query ${queryTag} failed: ${
+        completedQuery.message ?? "No message"
+      }`,
+    );
+    return;
+  }
+
+  const interpretedResultsPath = join(
+    queryStorageDir,
+    `interpreted-results-${queryTag.replaceAll(" ", "-")}-${queryRun.id}.sarif`,
```
charisk, line 143: Are we relying on implementation details of the query runner here? Should we be doing something a bit more generic e.g. look for a SARIF file in the dir?

extensions/ql-vscode/src/data-extensions-editor/auto-model-codeml-queries.ts
```diff
@@ -0,0 +1,223 @@
...
+    },
+    sourceInfo,
+    ["--sarif-add-snippets"],
+  );
+
+  return sarif;
+}
+
+export async function runAutoModelQueries(
+  mode: Mode,
```
charisk, line 162: What makes "mode" not an "option" but the rest of the properties "options"? I'm trying to understand why we don't receive all of those as arguments (or as a single argument that is an object with all the properties).

extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/auto-model-codeml-queries.test.ts
```diff
@@ -0,0 +1,188 @@
...
+import { QueryResultType } from "../../../../src/query-server/new-messages";
+import { runAutoModelQueries } from "../../../../src/data-extensions-editor/auto-model-codeml-queries";
+import { Mode } from "../../../../src/data-extensions-editor/shared/mode";
+import { mockedObject } from "../../utils/mocking.helpers";
+import { CodeQLCliServer } from "../../../../src/codeql-cli/cli";
+import { QueryRunner } from "../../../../src/query-server";
+import * as queryResolver from "../../../../src/local-queries/query-resolver";
+import * as standardQueries from "../../../../src/local-queries/standard-queries";
+
+function createMockUri(path = "/a/b/c/foo"): Uri {
```
charisk, line 17: Would be great to move this to a file (like other mocks/stubs) so that it can be re-used easily.

charisk (commented):
Thanks for getting this working and for writing a few tests too!

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -367,48 +380,99 @@ export class DataExtensionsEditorView extends AbstractWebview<
...
+        step: 2500,
+        maxStep,
+        message: "Parsing response",
+      });
+
+      const models = loadYaml(response.models, {
+        filename: "auto-model.yml",
+      });
+
+      const modeledMethods = loadDataExtensionYaml(models);
```
charisk, line 426: I just want to double check: this replaces all modeled methods right? But that's okay because we only have a button that models the whole package and we have 1 YAML per package.

If we want to add a button for a single candidate then we'd need to merge the existing modeled methods and any that we get from automodel.

extensions/ql-vscode/src/data-extensions-editor/auto-model-codeml-queries.ts
koesie10: These are regular method inputs, but with this many inputs it becomes somewhat harder to reason about the order of the arguments. By using an object, they are at least all named so it becomes easier to reason about them.

I've renamed it to `AutoModelQueriesOptions` and moved `mode` into it. I've also moved it closer to the actual function definition so it's easier to distinguish between the two options types.

extensions/ql-vscode/src/data-extensions-editor/auto-model-codeml-queries.ts
koesie10: No, this is the path at which we will store the SARIF. This just creates a unique path to store the SARIF.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
koesie10: It does not replace all modeled methods, it will already merge the modeled methods. Anything returned from the LLM will any already existing modeled methods, but if something is not returned from the LLM, anything the user has already modeled will be unchanged.

extensions/ql-vscode/src/data-extensions-editor/auto-model-codeml-queries.ts
charisk: Ah ofc! Makes sense.

extensions/ql-vscode/src/data-extensions-editor/auto-model-codeml-queries.ts
charisk: Thanks for the changes! The term "options" still seems a bit off to me because it makes me think of optional or variadic arguments, but I don't have a good suggestions atm (only "inputs" comes to mind).

charisk (approved):
Thanks!

PR #2644, opened 2023-07-26
Fix alert by rewriting query history scrubbed to do fewer file operations

robertbrignull:
This PR aims to fix https://github.com/github/vscode-codeql/security/code-scanning/384

We could maybe also get away with just closing the alert as "wont fix". It's not the most important of use-after-check errors. But I think we can make things a little bit better.

I'm not sure there's a way to rewrite the check to use file descriptors as the alert suggests. Although we could pass a file descriptor to `readFile`, I don't think `pathExists` or `stat` support them.

I think a better solution may be to rewrite the function to only make a single access to the file. All we need to do is to attempt to read its contents and then we can determine a lot of the rest from whether that succeeds or fails. We do lose some information so we can't keep 100% of the different error messages, but it's unclear to me if this detailed level of logging is strictly necessary. The tests we have don't explicitly check the error message 🤷🏼

I've also rewritten the code to be three smaller functions instead of one big function. Hopefully that makes the logic a little clearer.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-history/query-history-scrubber.ts
+++ b/extensions/ql-vscode/src/query-history/query-history-scrubber.ts
@@ -6,6 +6,7 @@ import { extLogger } from "../common/logging/vscode";
 import { readDirFullPaths } from "../common/files";
 import { QueryHistoryDirs } from "./query-history-dirs";
 import { QueryHistoryManager } from "./query-history-manager";
+import { getErrorMessage } from "../common/helpers-pure";
 
 const LAST_SCRUB_TIME_KEY = "lastScrubTime";
 
@@ -141,7 +142,7 @@ async function scrubDirectory(
     }
   } catch (err) {
     return {
-      errorMsg: `  Could not delete '${dir}': ${err}`,
+      errorMsg: `  Could not delete '${dir}': ${getErrorMessage(err)}`,
       deleted: false,
     };
   }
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/query-history/query-history-scrubber.ts
```diff
@@ -132,46 +132,53 @@ async function scrubDirectory(
...
+
+async function getTimestamp(
+  timestampFile: string,
+): Promise<number | undefined> {
+  try {
+    const timestampText = await readFile(timestampFile, "utf8");
+    return parseInt(timestampText, 10);
+  } catch (err) {
+    void extLogger.log(
+      `  Could not read timestamp file '${timestampFile}': ${err}`,
```
koesie10, line 180: Should we be using `getErrorMessage` here?

koesie10 (approved):
LGTM. I agree that this should be equivalent and we don't actually need to have 3 different checks before reading the timestamp file.

extensions/ql-vscode/src/query-history/query-history-scrubber.ts
robertbrignull: Good point, we should do. I was being a bit lazy here by just copying a different log call from this file.

robertbrignull:
The tests are all passing, and I've tried forcing the scrubber to run locally and it all worked fine.

PR #2645, opened 2023-07-27
Don't alert for multiple usages of built-in VS Code commands

robertbrignull:
Fixes https://github.com/github/vscode-codeql/security/code-scanning/546

The first commit fixes all the indentation in the query, so I suggest ignoring that or at least reviewing with whitespace ignored.

The second commit updates the query to know about built-in VS Code commands and not to create an alert for those commands.

My reasoning for not alerting on built-in commands is that:
- We don't emit telemetry from them, because we haven't provided the implementation.
- We can't avoid calling them from multiple places, again because we haven't provided the implementation we can't make duplicate commands with different names.

## Checklist

### DIFF
--- a/.github/codeql/queries/unique-command-use.ql
+++ b/.github/codeql/queries/unique-command-use.ql
@@ -15,134 +15,133 @@
  * that should be changed to fix the alert.
  */
 
- import javascript
+import javascript
 
- /**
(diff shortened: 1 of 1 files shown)


koesie10 (approved):
Thanks, I agree that we shouldn't be alerting on built-in commands

PR #2646, opened 2023-07-27
Split codeQL.copyVariantAnalysisRepoList into two commands

robertbrignull:
Fixes https://github.com/github/vscode-codeql/security/code-scanning/461

`codeQL.copyVariantAnalysisRepoList` is used from two places, but they're both internal so it's trivial to split the command into two.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/commands.ts
+++ b/extensions/ql-vscode/src/common/commands.ts
@@ -244,7 +244,11 @@ export type VariantAnalysisCommands = {
     scannedRepo: VariantAnalysisScannedRepository,
     variantAnalysisSummary: VariantAnalysis,
   ) => Promise<void>;
-  "codeQL.copyVariantAnalysisRepoList": (
+  "codeQL.copyVariantAnalysisRepoListQueryHistory": (
+    variantAnalysisId: number,
+    filterSort?: RepositoriesFilterSortStateWithIds,
+  ) => Promise<void>;
+  "codeQL.copyVariantAnalysisRepoListView": (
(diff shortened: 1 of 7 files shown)


shati-patel (approved):
Seems reasonable! Though I'm slightly confused that we now have a lot of command identifiers referencing the same "action". I _think_ it makes sense, since some of them are just the user-facing ones, others are the internal commands? 🤔 

e.g. We already have `codeQLQueryHistory.copyRepoList`, so do we need to define `codeQL.copyVariantAnalysisRepoListQueryHistory` separately?

robertbrignull:
@shati-patel, you raise a very good question. Looking into it more, I agree we don't need to have `codeQLQueryHistory.copyRepoList` immediately delegate to `codeQL.copyVariantAnalysisRepoListQueryHistory`. In fact I don't think we gain anything by having these extra commands and we can delete `codeQL.copyVariantAnalysisRepoListQueryHistory` and `codeQL.copyVariantAnalysisRepoListView`. In both cases we can call `VariantAnalysisManager.copyRepoListToClipboard` directly and there's no reason not to.

The only thing we then have to consider is if we're still logging enough telemetry, so I've added in some telemetry when copying repo lists. This mirrors what we do for exporting results which is defined just next to it.

shati-patel (approved):
Thanks for looking into this in some more detail! Looks good to me locally too ✨

robertbrignull:
> the failing test looks legit though 😅

Yep, I wasn't thinking properly when I wrote that. It should be mocking the right function now 😅

PR #2658, opened 2023-07-31
Remove submodules dependency from data extension editor

norascheuch:
Instead of creating a new temporary folder with dependency files every time we fetch the external api usage, we do this only once now during startup of the editor. This is also the time when we now install pack dependencies and thus remove dependency on submodules.

## Checklist


extensions/ql-vscode/src/data-extensions-editor/external-api-usage-query.ts
```diff
@@ -1,39 +1,35 @@
...
   queryStorageDir: string;
+  queryDir: string;
 
   progress: ProgressCallback;
   token: CancellationToken;
 };
 
 export async function runQuery(
-  queryName: keyof Omit<Query, "dependencies">,
+  mode: string,
```
robertbrignull, line 26: Could this be
```suggestion
  mode: Mode,
```

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts
```diff
@@ -99,13 +110,68 @@ export class DataExtensionsEditorModule {
               return;
             }
 
+            // Create new temporary directory for query files and pack dependencies
```
robertbrignull, line 113: Optional suggestion: could this chunk of code being moved here go into a separate method? It would help to keep the implementation of the `codeQL.openDataExtensionsEditor` command easy to read.

Would it actually make sense to keep this code in `external-api-usage-query.ts`, so then it's co-located with the `runQuery` method which depends on the setup being done by this code.

extensions/ql-vscode/AutomodelVsCode.qll
robertbrignull, line 1: Is this empty file meant to be part of this PR? Could you explain its purpose?

norascheuch: 😬 ups, will remove it, must be a file accidentally created during experiments with tests.

extensions/ql-vscode/src/data-extensions-editor/external-api-usage-query.ts
norascheuch: 👍

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts
norascheuch: Good idea! That also simplifies tests I think.

norascheuch: I made some changes

robertbrignull:
I've tested it out, both in a codespace and locally in a repo without the QL submodule. It worked in both cases! 🎉 🌮

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts
```diff
@@ -99,13 +109,38 @@ export class DataExtensionsEditorModule {
               return;
             }
 
+            if (!isQueryLanguage(db.language)) {
```
robertbrignull, line 112: Could we combine this with the check at https://github.com/github/vscode-codeql/blob/1289ab509c5f55c7757628f904fea68acca71e80/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts#L63 so we aren't checking the language twice and using different error messages.

So the two checks we're performing are:
- Is the language a member of the `QueryLanguage` type?
- Is the language supported by the data extensions editor?

Maybe something like:
```typescript
if (!(SUPPORTED_LANGUAGES.includes(language) && isQueryLanguage(language))) {
  void showAndLogErrorMessage(
    this.app.logger,
    `The data extensions editor is not supported for ${db.language} databases.`,
  );
  return;
}
```

I'd put both of these checks at the start where we current check `SUPPORTED_LANGUAGES`.

extensions/ql-vscode/src/data-extensions-editor/external-api-usage-query.ts
```diff
@@ -78,18 +51,42 @@ export async function runQuery(
...
+    cliServer,
+    queryRunner,
+    databaseItem,
+    queryStorageDir,
+    queryDir,
+    progress,
+    token,
+  }: RunQueryOptions,
+): Promise<CoreCompletedQuery | undefined> {
+  // The below code is temporary to allow for rapid prototyping of the queries. Once the queries are stabilized, we will
```
robertbrignull, line 74: The note about the queries moving to the github/codeql repo makes sense. I'm not 100% sure what "non-pretty" aspect of the code the comment is referring to, and it's a little hard to compare against the code linked in the comment to see what's different.

This comment could maybe be made clearer for new readers, but alternatively it's fine to just leave it if it's too hard. As you say, the code should be deleted soon-ish hopefully.

robertbrignull (approved):
A couple of comments which you could address or leave, but generally I think this PR is ok and it works well for me when testing it.

robertbrignull (approved):

extensions/ql-vscode/src/data-extensions-editor/external-api-usage-query.ts
norascheuch: In this case here I only moved the comment (and code) to a new place, it existed before.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts
norascheuch: Discussed sync and resolved!

PR #2666, opened 2023-08-03
Move ResultTablesHeader to its own component

robertbrignull:
I think we can reduce the size of the `ResultTables` component by pulling out to the bits for the header to a separate component. This header contains the pagination controls as well as the query name and file link.

Screenshot 2023-08-03 at 12 27 10

I've been testing it and as far as I can tell it's still working just as it was before.

## Checklist


extensions/ql-vscode/src/view/results/result-tables.tsx
```diff
@@ -168,10 +164,7 @@ export class ResultTables extends React.Component<
             getResultSets(props.rawResultSets, props.interpretation),
           );
 
-        return {
```
robertbrignull, line 171: This is the only bit where it isn't perfectly replicated on the new component and I'm unsure if it's needed or not.

This is checking if the result set exists and if not setting the selected table and page to default values. I'm not sure how to trigger this state. I also think it should be covered by the `useEffect` hook in the header component which always sets the selected page to the value from the parsed result set.

extensions/ql-vscode/src/view/results/ResultTablesHeader.tsx
```diff
@@ -0,0 +1,141 @@
...
+  selectedTable: string;
+}
+
+export function ResultTablesHeader(props: Props) {
+  const { queryPath, queryName, parsedResultSets, selectedTable } = props;
+
+  const [selectedPage, setSelectedPage] = React.useState(
+    `${parsedResultSets.pageNumber + 1}`,
+  );
+  useEffect(() => {
```
robertbrignull, line 25: Interestingly this effect hook might not strictly be needed. My brain says it should be needed to update `selectedPage`, but when I try removing it everything still works. 🤔 

The case I'd expect to be different is when moving between query results. If I change from viewing the results of one query to a different query, then `parsedResultSets` is updated and we want `selectedPage` to also be reset.

Possibly it's because `ResultTables` is constructing a new copy of this component on every render, so we're always getting the initial calculated value. This could change as we continue to tweak `ResultTables`.

extensions/ql-vscode/src/view/results/ResultTablesHeader.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const onBlurHandler = useCallback(
+    (e: React.FocusEvent<HTMLInputElement, Element>) => {
+      changePage(e.target.value);
+    },
+    [changePage],
+  );
+
+  const onKeyDownHandler = useCallback(
+    (e: React.KeyboardEvent<HTMLInputElement>) => {
+      changePage(e.currentTarget.value);
```
robertbrignull, line 76: Oops, I forgot the logic about checking if it's keycode 13. I'll add that back in.

robertbrignull: I've actually done `e.key === "Enter"` because `e.keyCode` is deprecated now. See https://www.w3.org/TR/uievents-key/#named-key-attribute-values for a list of possible values.

robertbrignull:
A note to any reviewers: I'm not hugely confident that this code is any good 😄 . I'm also not all that attached to it, so if it turns out to be too hard to review or you're also unsure if this PR is an improvement on the current state then I'm happy to just close it. This work is not essential and we can always try it again another time.

charisk (approved):
Sorry it took a while to review - the diff was a bit scary 😬 

I'm not super confident in my review, but I think each code change I've seen makes sense and it's a positive step. I can't see anything wrong. 🚢

extensions/ql-vscode/src/view/results/ResultTablesHeader.tsx
charisk: I think it's worth keeping it - it does seem like the right thing to do.

robertbrignull:
> Sorry it took a while to review - the diff was a bit scary 😬

I'm sorry it was a difficult diff! Next time I do one of these I'll try harder to do each change separately. In particular I should have moved the code to the new file in a separate commit to transforming it. Thank you for persevering. ❤️

Issue #2622, opened 2023-07-19
Variant Analysis Results fails to show results if query message contains placeholder which is not part of source


Marcono1234:
**Describe the bug**
When a query contains a placeholder in the query message (`$@`), and the results of a Variant Analysis run have a value for that placeholder which is not part of the source of a project, then the Variant Analysis Results view fails to show results.

Additionally the following is logged in the extension log:
```
Error when processing SARIF result: TypeError: Cannot read properties of undefined (reading 'startLine')
```

**Version**
Extension version
```
CodeQL extension version: 1.8.8 
CodeQL CLI version: 2.14.0 
Platform: win32 x64
```

VS Code version
```
Version: 1.80.1 (user setup)
Commit: 74f6148eb9ea00507ec113ec51c489d6ffb4b771
Date: 2023-07-12T17:22:07.651Z
Electron: 22.3.14
ElectronBuildId: 21893604
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Windows_NT x64 10.0.19045
```

**To reproduce**
1. Open Quick Query and paste the following query
    ```ql
    /**
     * @kind problem
     */

    import java

    // Pick arbitrary element which is part of source; chose EnumType here because projects
    // likely don't have that many enum types
    // TypeString comes from the JDK and is not part of the source of the project
    from EnumType t, TypeString notPartOfSource
    where t.fromSource()
    select t, "message $@", notPartOfSource, notPartOfSource.getName()
    ```
2. Run Variant Analysis, for example on the top 10 repositories
3. On the Variant Analysis Results view click the triangle to expand the results
  :x: Bug: No results are shown and an error is logged

**Expected behavior**
Results should be shown, but if the placeholder value is not part of the source, there should simply not be a link.

(Also: Maybe in general if an error while processing the SARIF result happens, the complete stack trace should be printed to the log to make debugging easier?)

robertbrignull:
Thank you for reporting this. I've also been able to reproduce the missing results when one of the placeholder values is not from the source archive. I get the error `Error when processing SARIF result: TypeError: Cannot read properties of undefined (reading 'startLine')` when trying to expand the results.

This issue is very closely related to https://github.com/github/vscode-codeql/issues/2629, though perhaps not entirely the same because the results are present but just fail to display. In that issue the results that are not from the source are completely omitted.

We've got an internal issue to track this, but we'll report here when there's any progress on this.

[...]

PR #2672, opened 2023-08-04
Handle when an alert message contains links to files outside of the repository source 

robertbrignull:
Closes https://github.com/github/vscode-codeql/issues/2622

Fixes support for MRVA using queries where a placeholder link goes to outside of the source archive. In this case we have a location without a `region` and this was causing a crash because we were naughtily assuming fields to be defined.

Rendering of alerts now works and the message placeholder is rendered as a regular string instead of a link:
Screenshot 2023-08-04 at 12 28 42

Use the following query to reproduce the bug and observe that it now works:
```ql
/**
 * @kind problem
 */

import java

// Pick arbitrary element which is part of source; chose EnumType here because projects
// likely don't have that many enum types
// TypeString comes from the JDK and is not part of the source of the project
from EnumType t, TypeString notPartOfSource
where t.fromSource()
select t, "message $@", notPartOfSource, notPartOfSource.getName()
```

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/sarif-utils.ts
+++ b/extensions/ql-vscode/src/common/sarif-utils.ts
@@ -2,7 +2,7 @@ import * as Sarif from "sarif";
 import type { HighlightedRegion } from "../variant-analysis/shared/analysis-result";
 import { ResolvableLocationValue } from "../common/bqrs-cli-types";
 
-interface SarifLink {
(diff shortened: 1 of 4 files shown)


charisk (approved):
Thanks for fixing this! I've not tested locally myself but the code changes make sense.

PR #2673, opened 2023-08-04
Remove unused NPM dependencies

robertbrignull:
I ran [depcheck](https://www.npmjs.com/package/depcheck) and then manually went through the results looking at which ones were genuinely unused. As far as I can tell, all of these are unused and can be removed.

After removing each dependency I ran `npm install` and `npm run build && npm run check-types`. I've also tried building and running the extension in development mode and installing from a file.

Of these:
- I'm fairly confident that the `@types` packages are correct. If any of these were used then the `check-types` script would fail.
- It may appear that `stream` is used from `cli.ts` and other files, but that is actually the built-in node.js stream implementation. The `stream` package is for providing the same implementation inside the browser/webviews, and we aren't using it in this context.
- I can't find any references to `minimatch`, `minimist`, `stream-chain`, or `ts-protoc-gen`.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -1809,7 +1809,6 @@
     "@types/fs-extra": "^11.0.1",
     "@types/gulp": "^4.0.9",
     "@types/gulp-replace": "^1.1.0",
-    "@types/gulp-sourcemaps": "0.0.32",
     "@types/jest": "^29.0.2",
     "@types/js-yaml": "^3.12.5",
     "@types/jszip": "^3.1.6",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -1818,7 +1818,6 @@
     "@types/react-dom": "^18.0.11",
     "@types/sarif": "^2.1.2",
     "@types/semver": "^7.2.0",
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/package.json
```diff
@@ -1834,7 +1826,6 @@
     "@types/unzipper": "^0.10.1",
     "@types/vscode": "^1.67.0",
     "@types/webpack": "^5.28.0",
-    "@types/webpack-env": "^1.18.0",
```
koesie10, line 1837: I think this one should actually stay here: in the view code, we can use functions from the Webpack environment, such as `require.resolve`, `module.hot`, etc. I don't think we're currently using any of those, but we do have access to those.

koesie10 (commented):
These look correct to me, but I think `@types/webpack-env` should be in `package.json` (even though it's currently unused).

extensions/ql-vscode/package.json
robertbrignull: I'm happy to leave `@types/webpack-env` in there 👍🏼

koesie10 (approved):
Thanks for cleaning this up!

PR #2675, opened 2023-08-07
Add in-progress UI state for modeling methods

charisk:
Add the ability to mark methods as being "in progress" for modeling. This will be used for the AI modeling flows. The exact UX is still TBD, but some initial mock-ups show a progress ring and drop-downs being replaced with "Thinking...".

I followed the same approach as the one we use to track modified methods which unfortunately has a bit of prop drilling but I don't think it's too bad and we can improve that separately if we like.

Note that this new state and view message is not used yet - it'll be plugged in in a later PR. It can be seen in action using the storybook.

## Checklist
N/A:


starcke (approved):
Looks good to me.

PR #2679, opened 2023-08-08
Implementing jumping to location when clicking on a usage in the details panel

robertbrignull:
The feature being implemented here is that you can click on an entry in the CodeQL model details panel and it'll open that code location in the main editor. This doesn't implement selecting the correct usage when clicking on the "view" button and that will be a different PR.

I've implemented this using a new `codeQLDataExtensions.jumpToUsageLocation` command. I did first try out using `vscode.open` and passing it a URI, but sadly I don't believe we can use that. Instead we can call the existing `showResolvableLocation` method, but making that work involves a few steps:
- We'd like some error handling, and we currently already perform this same error handling in two places. I move the error handling to be inside `showResolvableLocation` so we avoid duplication.
- We need the `DatabaseItem` to be able to resolve the code location to a file that can be displayed in the editor. This means that we need to pass the database item to the `ModelDetailsDataProvider` class along with the `ExternalApiUsage[]` array. I think it makes sense to set these together because you need both to be in sync in order to resolve locations correctly.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts
@@ -147,7 +147,7 @@ export class DataExtensionsEditorModule extends DisposableObject {
               db,
               modelFile,
               Mode.Application,
-              (usages) => this.modelDetailsPanel.setExternalApiUsages(usages),
+              this.modelDetailsPanel.setState.bind(this.modelDetailsPanel),
             );
             await view.openView();
           },
--- a/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts
@@ -156,7 +156,7 @@ export class DataExtensionsEditorModule extends DisposableObject {
           },
         );
       },
-      "codeQLDataExtensions.jumpToUsageLocation": async (
+      "codeQLDataExtensionsEditor.jumpToUsageLocation": async (
         usage: Usage,
         databaseItem: DatabaseItem,
       ) => {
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/data-extensions-editor/model-details/model-details-panel.ts
```diff
@@ -17,7 +18,10 @@ export class ModelDetailsPanel extends DisposableObject {
     this.push(treeView);
   }
 
-  public setExternalApiUsages(externalApiUsages: ExternalApiUsage[]): void {
-    this.dataProvider.setExternalApiUsages(externalApiUsages);
+  public setState(
+    externalApiUsages: ExternalApiUsage[],
+    databaseItem: DatabaseItem,
+  ): void {
+    this.dataProvider.setState(externalApiUsages, databaseItem);
```
charisk, line 25: This surprised me a little bit. I've generally thought of the "DataProvider" classes as the ones managing state and expected the data extensions editor view to use the provider rather than the panel. Having said that, I can't think of a problem with what you've done here 🤔

charisk (approved):
Thank you for splitting into sensible commits - it made reviewing much easier 🎉 

I've got a couple of non-blocking comments.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-module.ts
```diff
@@ -154,6 +156,12 @@ export class DataExtensionsEditorModule extends DisposableObject {
           },
         );
       },
+      "codeQLDataExtensions.jumpToUsageLocation": async (
```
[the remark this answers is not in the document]
robertbrignull: Good point. I was more trying to match the naming scheme of the files for the module, but either way `codeQLDataExtensionsEditor` is a better name.

extensions/ql-vscode/src/data-extensions-editor/model-details/model-details-panel.ts
charisk: Cool! Lets go with that and see if we can identify any pros/cons about either approach. In general it feels like it would be good to standardise a bit to make the codebase a bit easier.

charisk (approved):

PR #2681, opened 2023-08-09
Pass dependency name around when automodeling

charisk:
We will need the dependency name in order to manage automodeling jobs so this is a quick PR to make sure it's available.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/auto-modeler.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/auto-modeler.ts
@@ -29,18 +29,26 @@ export class AutoModeler {
   ) {}
 
   public async startModeling(
+    dependency: string,
     externalApiUsages: ExternalApiUsage[],
     modeledMethods: Record<string, ModeledMethod>,
     mode: Mode,
   ): Promise<void> {
-    await this.modelDependency(externalApiUsages, modeledMethods, mode);
+    await this.modelDependency(
+      dependency,
+      externalApiUsages,
+      modeledMethods,
+      mode,
+    );
   }
 
   private async modelDependency(
+    dependency: string,
     externalApiUsages: ExternalApiUsage[],
     modeledMethods: Record<string, ModeledMethod>,
     mode: Mode,
   ): Promise<void> {
+    void extLogger.log(`Modeling dependency ${dependency}`);
     await withProgress(async (progress) => {
       const maxStep = 3000;
 
--- a/extensions/ql-vscode/src/view/data-extensions-editor/LibraryRow.tsx
+++ b/extensions/ql-vscode/src/view/data-extensions-editor/LibraryRow.tsx
@@ -85,6 +85,7 @@ type Props = {
     modeledMethods: Record<string, ModeledMethod>,
   ) => void;
   onGenerateFromLlmClick: (
+    dependencyName: string,
     externalApiUsages: ExternalApiUsage[],
     modeledMethods: Record<string, ModeledMethod>,
   ) => void;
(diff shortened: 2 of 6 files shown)


extensions/ql-vscode/src/data-extensions-editor/auto-modeler.ts
```diff
@@ -29,18 +29,26 @@ export class AutoModeler {
...
+    );
   }
 
   private async modelDependency(
+    dependency: string,
     externalApiUsages: ExternalApiUsage[],
     modeledMethods: Record<string, ModeledMethod>,
     mode: Mode,
   ): Promise<void> {
+    void extLogger.log(`Modeling dependency ${dependency}`);
```
robertbrignull, line 51: Is this logging intentional or maybe leftover from testing?

I'm just wondering if we need it, since most users won't see it in the logs.

extensions/ql-vscode/src/view/data-extensions-editor/LibraryRow.tsx
```diff
@@ -119,11 +120,11 @@ export const LibraryRow = ({
 
   const handleModelWithAI = useCallback(
     async (e: React.MouseEvent) => {
-      onGenerateFromLlmClick(externalApiUsages, modeledMethods);
+      onGenerateFromLlmClick(title, externalApiUsages, modeledMethods);
```
robertbrignull, line 123: Does this work correctly in framework mode? In that case the `title` is the package name and there is no "dependency".

extensions/ql-vscode/src/data-extensions-editor/auto-modeler.ts
charisk: It was intentional as I thought it might be useful for debugging issues. It's also good to have it right now because otherwise the dependency name isn't used.

extensions/ql-vscode/src/view/data-extensions-editor/LibraryRow.tsx
charisk: Yeah it seems to work.

Perhaps the name I'm using is a bit wrong? What do we call these things if not "dependency"?

robertbrignull: Sorry I wasn't very clear. The variable does have a value, but I was worried it might confuse the LLM so wanted to make sure you were aware the value might not be what you expect.

> Perhaps the name I'm using is a bit wrong? What do we call these things if not "dependency"?

Maybe "source" or something like that? In [`sorting.ts`](extensions/ql-vscode/src/data-extensions-editor/shared/sorting.ts) we call them "groupings". I'm not sure either of those are better though.

Alternatively if there is not sensible value it could contain in framework mode then maybe make it optional?

charisk: Ah, I see, thanks for the clarification. I think it's still okay - we essentially need a unique enough name to manage automodeling jobs so that seems like it would work. I think the value is exactly what we want - it's the name of the package. It's just that the name is confusing, but `source` and `groupings` both sound more confusing to me 😄 is it perhaps a `package`?

charisk: Since we don't have a clear name right away and I've got some changes stacked up on top of this one, I might go ahead with `dependency` for now and circle back to rename later.

robertbrignull: Sounds good to me

robertbrignull (approved):

PR #2682, opened 2023-08-09
Batch automodeling

charisk:
Add batching of automodeling requests and also the ability to stop. See individual commits for the two steps.

Note that there is a bug around progress that I'd like to fix separately as it requires some refactoring of the in progress state. The bug shows up if a user starts modeling a dependency before the previous one has finished. 

The progress and stopping notifications can also do with a bit of improvement, but this is a good enough start in my opinion.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/common/interface-types.ts
+++ b/extensions/ql-vscode/src/common/interface-types.ts
@@ -554,6 +554,11 @@ interface GenerateExternalApiFromLlmMessage {
   modeledMethods: Record<string, ModeledMethod>;
 }
 
+interface StopGeneratingExternalApiFromLlmMessage {
+  t: "stopGeneratingExternalApiFromLlm";
+  dependencyName: string;
+}
+
 interface ModelDependencyMessage {
   t: "modelDependency";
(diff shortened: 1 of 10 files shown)


starcke (approved):
I think this is a good start. I did find a problem where if I start two automodels for different packages then it keep clearing the "Thinking" label and reenable the Model AI button (probably when the other automodel would update the state).

charisk:
> I think this is a good start. I did find a problem where if I start two automodels for different packages then it keep clearing the "Thinking" label and reenable the Model AI button (probably when the other automodel would update the state).

Yep, that's bug I mentioned in the PR description. I need to do a little bit of refactoring to fix that so I wanted to leave it for a separate PR.

starcke:
> Yep, that's the bug I mentioned in the PR description. I need to do a little bit of refactoring to fix that so I wanted to leave it for a separate PR. Thanks for testing! 🎉

Ah doh. I thought it was if you tried to manually model a row while it was running. But re-reading I can see it was the same as I found.

PR #2685, opened 2023-08-09
Fix `getCurrentQuery` to look at active tabs instead of active text editors

shati-patel:
The `getCurrentQuery` method determines the "active" query by checking the VS Code `activeTextEditor`. Due to how VS Code defines text editors, this leads to weird cases where it can consider comment/discussion boxes as the active text editor, instead of the underlying file 😅 E.g. we have a bug in the CodeQL code tour, where we accidentally set this comment as the active editor. 

🔗 See internal linked issue for more details!

image

To avoid this, we can look at the active _editor tab_ instead. This doesn't change the existing behaviour, and fixes the weird comment/discussion box bug. I've tested this in various cases, including running quick eval, quick queries, and just running queries from various places (context menu, file explorer, command palette) 😄 

Note that there are some other places in the code where we use `window.activeTextEditor`. We _could_ change all of those occurrences, but I'm hesitant to overcomplicate things where it's probably not necessary 🙈  

👀 

@dbartol, would you mind giving this a quick look/test? The bug first occured after https://github.com/github/vscode-codeql/commit/0bd0bf1944a42181cb3397f1366134d91dc84536 and the `getCurrentQuery` method is also used in the new debugger [here](https://github.com/github/vscode-codeql/blob/c8ba8d6e1b4d8040d40e02910b73bd562225e8c0/extensions/ql-vscode/src/debugger/debugger-ui.ts#L196). I want to be sure that the "save dirty documents" and debugging stuff all still works as expected too!

## Checklist

I'm not sure this needs a changelog entry, since it doesn't change any behaviour? 🤔 (and the codespace template/code tour is still in beta)


adityasharad:
This likely explains many silent failures I saw while attempting query-running operations in the CodeTour that I could not reliably reproduce or understand!

robertbrignull (approved):
LGTM but I did help to write it

Still likely worth getting a 👍🏼 from @dbartol on the if possible to make sure it will works with the debugging UI

dbartol:
This shouldn't affect query debugging (other than the intended effect, of course). All the "save dirty documents" code just calls `workspace.saveAll()`, without attempting to determine the current editor/tab on its own.

PR #2688, opened 2023-08-10
Implementing expanding/selecting item in details view when clicking "view"

robertbrignull:
This PR fully implements the "view" button when you have the details view enabled. We should highlight the usage in the details view, and also still jump to location in the main editor window.

In order to make it work I had to include the full `Usage` object in the "jumpToUsage" message. Unfortunately I ran into a couple of problems when calling `treeView.reveal` and implementing `getParent`:
- One problem is that those methods rely on object identity equality, and when the usage gets passed into and out of the webview in messages it doesn't stay the same object. It has all the same fields with all the same values, but it is a different javascript object occupying different memory. I worked around this by adding a `resolveCanonicalUsage` method that finds the correct `Usage` object with all the same fields.
- Another problem is that there's no good way to implement `getParent` to go back from the `Usage` to the `ExternalApiUsage` without iterating through every one and checking them.

Both of the above problems are more code-beauty and performance problems than they are correctness problems. I have some ideas for making them more performant which I'll put as comments on this PR, but they have tradeoffs and may not be worth it.

## Checklist


extensions/ql-vscode/src/data-extensions-editor/model-details/model-details-data-provider.ts
```diff
@@ -61,12 +61,45 @@ export class ModelDetailsDataProvider
...
+  ): ModelDetailsTreeViewItem | undefined {
+    if (isExternalApiUsage(item)) {
+      return undefined;
+    } else {
+      return this.externalApiUsages.find((e) => e.usages.includes(item));
+    }
+  }
+
+  public resolveCanonicalUsage(usage: Usage): Usage | undefined {
+    for (const externalApiUsage of this.externalApiUsages) {
```
robertbrignull, line 76: We could make `resolveCanonicalUsage` and `getParent` more performant by adding the `ExternalApiUsage` object into the "jumpToUsage" message as well. This way we would at least only have to search all usages of a single method, instead of searching all methods.

In terms of actual time saved it's probably milliseconds at most, but it may be noticeable on really big databases.

extensions/ql-vscode/src/data-extensions-editor/model-details/model-details-data-provider.ts
```diff
@@ -61,12 +61,45 @@ export class ModelDetailsDataProvider
...
-type ModelDetailsTreeViewItem = ExternalApiUsage | Usage;
+export type ModelDetailsTreeViewItem = ExternalApiUsage | Usage;
 
 function isExternalApiUsage(
   item: ModelDetailsTreeViewItem,
 ): item is ExternalApiUsage {
   return (item as any).usages !== undefined;
 }
+
+function usagesAreEqual(u1: Usage, u2: Usage): boolean {
```
robertbrignull, line 95: We could make this a bit cleaner by giving the `Usage` object some unique "key" field similar to `ExternalApiUsage.signature`. The combination of URI and line/column should be unique, so we could use that.

This has the tradeoff that it makes the `Usage` object larger, so we're just passing around more data all of the time, and this is the only operation that it makes any faster.

extensions/ql-vscode/src/data-extensions-editor/model-details/model-details-panel.ts
```diff
@@ -24,4 +28,11 @@ export class ModelDetailsPanel extends DisposableObject {
   ): void {
     this.dataProvider.setState(externalApiUsages, databaseItem);
   }
+
+  public async revealItem(usage: Usage): Promise<void> {
+    const canonicalUsage = this.dataProvider.resolveCanonicalUsage(usage);
```
robertbrignull, line 33: The public interface of the model details panel and data provider are becoming a little larger and less well defined. I don't know if it's a problem yet but probably worth keeping an eye on and thinking about if we could make things clearer by moving methods around or changing which object is the public interface to the rest of the code.

extensions/ql-vscode/src/data-extensions-editor/model-details/model-details-data-provider.ts
```diff
@@ -61,12 +61,45 @@ export class ModelDetailsDataProvider
       return [];
     }
   }
+
+  getParent(
```
robertbrignull, line 65: I investigated a few different ways we could implement `getParent` without having to search through all items, but none seemed fantastic at the time. I might have another go at it though.

Adding a literal link to the `ExternalApiMethod` object to the `Usage` object seemed a bit ugly and fiddly. Another option would be the add the `ExternalApiMethod.signature` field to the `Usage` object, so then it's only adding a string field and we can easily use that key to look up the `ExternalApiMethod`.

Either way, we could have new types just for the data in this class, so we don't have to add new fields to the types used by the rest of the application.

extensions/ql-vscode/src/data-extensions-editor/model-details/model-details-data-provider.ts
charisk: Agree with the approach you took here - let's see how this performs in practise before adding more things and increasing complexity.

PR #2690, opened 2023-08-10
Code Search: Add integration test

norascheuch:
Adds an integration test for the code search feature.

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/activated-extension/databases/db-panel.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/activated-extension/databases/db-panel.test.ts
@@ -47,10 +47,6 @@ describe("Db panel UI commands", () => {
     );
   });
 
-  beforeAll(async () => {
-    await sleep(5000);
-  });
-
   it("should add new remote db list", async () => {
     // Add db list
     jest.spyOn(window, "showInputBox").mockResolvedValue("my-list-1");
--- a/extensions/ql-vscode/test/vscode-tests/activated-extension/databases/db-panel.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/activated-extension/databases/db-panel.test.ts
@@ -8,15 +8,23 @@ import {
 } from "../../../../src/databases/config/db-config";
 import {
   AddListQuickPickItem,
+  CodeSearchQuickPickItem,
   RemoteDatabaseQuickPickItem,
 } from "../../../../src/databases/ui/db-panel";
 import { DbListKind } from "../../../../src/databases/db-item";
-import { createDbTreeViewItemSystemDefinedList } from "../../../../src/databases/ui/db-tree-view-item";
-import { createRemoteSystemDefinedListDbItem } from "../../../factories/db-item-factories";
+import {
+  createDbTreeViewItemSystemDefinedList,
+  createDbTreeViewItemUserDefinedList,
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/test/vscode-tests/activated-extension/databases/db-panel.test.ts
```diff
@@ -128,15 +133,93 @@ describe("Db panel UI commands", () => {
...
+      remoteLists: [
+        {
+          name: listName,
+          repositories: [],
+        },
+      ],
+    });
+
+    await writeJson(dbConfigFilePath, dbConfigPreparation);
+    await sleep(5000);
```
charisk, line 174: I don't think we should have tests that have `sleep` because they're slow and/or flaky (because they rely on timings). This might be a personal opinion but I'd rather not have a test than have one that is slow/flaky. Happy to hear from others too.

extensions/ql-vscode/test/vscode-tests/activated-extension/databases/db-panel.test.ts
```diff
@@ -1,34 +1,55 @@
...
 import { createVSCodeCommandManager } from "../../../../src/common/vscode/commands";
 import { AllCommands } from "../../../../src/common/commands";
+import { MockGitHubApiServer } from "../../../../src/variant-analysis/gh-api/mocks/mock-gh-api-server";
+import { sleep } from "../../../../src/common/time";
+import { createDbConfig } from "../../../factories/db-config-factories";
 
 jest.setTimeout(60_000);
 
 describe("Db panel UI commands", () => {
+  // This test has some serious problems:
```
charisk, line 34: Thanks for writing a comment to explain things, I should have done that when writing the tests because even I forgot about these limitations.

I just want to call out that this was intentional and the only way I could think of to add some integration tests for the commands without changing the actual implementation (not the test code) somehow. However, since it's turning out hard to add more tests, perhaps we do need to make some changes to the implementation to help us out a bit (not necessarily part of your work).

charisk (commented):
Thanks for persevering with this! Integration tests are hard 😞 

I personally don't think we should go ahead with this but I'm happy to be convinced otherwise..

norascheuch:
I think we need a test for the code search feature. However I agree that what we need to do to make it run smoothly is not great and I'm open to not merge it. But it feels quite suboptimal to have so many parts of the feature untested.

robertbrignull:
I can try to offer an extra opinion here.

I can definitely see the value that these tests bring, because they are very close to how the user interacts with the feature. We can be pretty sure that the feature is working as intended.

However, I think they would benefit from some cleanup first. If we want to merge these we should take a bit more time making sure we're happy with things. The big sleeps and the comment at the start of the tests make me a bit sad, though unless I am being naive I believe we can fix some of them. If we're comfortable with it I think we have time available to invest in improving aspects like this.

norascheuch:
After @robertbrignull and myself paired on this for a time-boxed hour we found that we couldn't mock the used file-watcher. After speaking shortly with @koesie10 about it we found that the mock failed to work because the extension is already initialised before the test, including the mock, is loaded. It is not possible to mock any file-watcher (the used chokidar nor any other vscode-internal filewatcher) retroactively. 

Since we should add a manual test of this feature in our release test plan disregarding of the PR at hand it is better to close it and prevent one more possibly flaky test in our suite.

PR #2703, opened 2023-08-14
Make InProgressMethods immutable so it's safer to use as react state

robertbrignull:
A small modification that makes `setPackagedMethods` return a new object instead of mutating the existing object. The current code is correct, but I worry the chance of the class accidentally being used incorrectly is a bit too high.

The danger of having `InProgressMethods` be mutable is that we use it as react state, and react tracks state changes only by object identity. This means if a developer do something like:
```typescript
setInProgressMethods((oldInProgressMethods) => {
    oldInProgressMethods.setPackagedMethods(packageName, methods);
    return oldInProgressMethods;
);
```
then react would not see that the object was modified and would not trigger a re-render of the component. When the next render happened for whatever other reason, the change in state would be rendered. This leads to bugs and unpredictable visual behaviour.

By making `InProgressMethods` immutable it means you don't need to know that you must call `fromExisting` first to create a new object, and it'll be harder to accidentally mutate an object being used as react state.

It's worth nothing we do also have this problem with other objects and also with arrays. This one stuck out to me because it's a custom object with an interface, so the mutation of lack of it is hidden from the developer. We may also want to look at ways of reducing the chance of this bug in other places. Maybe a CodeQL query?

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/shared/in-progress-methods.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/shared/in-progress-methods.ts
@@ -19,10 +19,7 @@ export class InProgressMethods {
     packageName: string,
     methods: Set<string>,
   ): InProgressMethods {
-    const newMethodMap = new Map<string, Set<string>>();
-    this.methodMap.forEach((value, key) => {
-      newMethodMap.set(key, new Set<string>(value));
-    });
+    const newMethodMap = new Map<string, Set<string>>(this.methodMap);
     newMethodMap.set(packageName, methods);
     return new InProgressMethods(newMethodMap);
   }
--- a/extensions/ql-vscode/src/data-extensions-editor/shared/in-progress-methods.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/shared/in-progress-methods.ts
@@ -1,16 +1,30 @@
 /**
  * A class that keeps track of which methods are in progress for each package.
+ *
+ * This class is immutable and therefore is safe to be used in a react useState hook.
  */
 export class InProgressMethods {
   // A map of in-progress method signatures for each package.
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/data-extensions-editor/shared/in-progress-methods.ts
```diff
@@ -1,16 +1,30 @@
...
+   * Returns a new InProgressMethods instance.
+   */
+  public setPackageMethods(
+    packageName: string,
+    methods: Set<string>,
+  ): InProgressMethods {
+    const newMethodMap = new Map<string, Set<string>>();
+    this.methodMap.forEach((value, key) => {
+      newMethodMap.set(key, new Set<string>(value));
+    });
```
koesie10, line 25: I believe that these two options should also work and be equivalent:

```suggestion
    const newMethodMap = new Map<string, Set<string>>(this.methodMap);
```

```suggestion
    const newMethodMap = new Map<string, Set<string>>(this.methodMap.entries());
```

These options are probably faster because they can be optimized by V8. I don't think [it's allowed](https://tc39.es/ecma262/multipage/keyed-collections.html#sec-map-constructor) that Node.js uses the same map instance if calling `new Map(this.methodMap)`, so this should be fully equivalent.

koesie10 (approved):
LGTM, thanks for making this safe to use in React state

extensions/ql-vscode/src/data-extensions-editor/shared/in-progress-methods.ts
robertbrignull: Thanks for that suggestion. I'd forgotten for this bit that you can pass an iterable to the map constructor.

charisk (approved):
Thanks!

PR #2705, opened 2023-08-14
Allow the details view to handle multiple open data extensions editors

robertbrignull:
When there are multiple data extensions editors open, have the details view track whichever one was active (i.e. focussed by the user) the most recently. I believe this is the behaviour the user would expect when they have multiple editors open.

Note that while there can be at most one active data extensions editor at a time, there also could easily be zero active. If the user clicks over to a different tab, such as when viewing usages of a method, we want the details view to still be correct.

This PR introduces a couple of big changes to the `DataExtensionsEditorView` class:
- Adds a static variable tracking the `WebviewPanel` that has most recently been active.
  - If there's another better way to implement this behaviour I'm very open to it.
- Stores most recent `ExternalApiUsage[]` value, so it can be used to update the details view outside of the `loadExternalApiUsages` method.
  - We don't want to wait to re-run the queries whenever the user changes between tabs.

I've tried manually testing this in all of the configurations of tabs I could think of, and it seems to be working perfectly.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
@@ -3,7 +3,6 @@ import {
   ExtensionContext,
   Uri,
   ViewColumn,
-  WebviewPanel,
   window,
 } from "vscode";
 import {
@@ -49,8 +48,6 @@ export class DataExtensionsEditorView extends AbstractWebview<
   ToDataExtensionsEditorMessage,
   FromDataExtensionsEditorMessage
 > {
-  private static mostRecentlyActivePanel: WebviewPanel | undefined = undefined;
-
   private readonly autoModeler: AutoModeler;
 
   private externalApiUsages: ExternalApiUsage[];
@@ -71,6 +68,15 @@ export class DataExtensionsEditorView extends AbstractWebview<
       databaseItem: DatabaseItem,
     ) => Promise<void>,
     private readonly revealItemInDetailsPanel: (usage: Usage) => Promise<void>,
+    private readonly handleViewBecameActive: (
+      view: DataExtensionsEditorView,
+    ) => void,
+    private readonly handleViewWasDisposed: (
+      view: DataExtensionsEditorView,
+    ) => void,
+    private readonly isMostRecentlyActiveView: (
+      view: DataExtensionsEditorView,
+    ) => boolean,
   ) {
     super(ctx);
 
@@ -100,7 +106,7 @@ export class DataExtensionsEditorView extends AbstractWebview<
 
     panel.onDidChangeViewState(async () => {
       if (panel.active) {
-        await this.onPanelBecameActive();
+        this.handleViewBecameActive(this);
         await this.updateModelDetailsPanelState(
           this.externalApiUsages,
           this.databaseItem,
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -86,15 +91,36 @@ export class DataExtensionsEditorView extends AbstractWebview<
...
+          this.externalApiUsages,
+          this.databaseItem,
+        );
+      }
+    });
+
     await this.waitForPanelLoaded();
   }
 
+  private async isTheMostRecentlyActivePanel(): Promise<boolean> {
```
robertbrignull, line 113: We could split this into two methods, so we call a different method when the view state changes vs when we just want to check if we're the most-recently active. But I also kinda like the simplicity and security of just having one method that checks and updates `DataExtensionsEditorView.mostRecentlyActivePanel`.

koesie10: I wouldn't expect a method called `isTheMostRecentlyActivePanel` to make any updates to the state. I think splitting this into two methods would make sense (one of which can be a getter).

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -86,15 +91,36 @@ export class DataExtensionsEditorView extends AbstractWebview<
...
 
+    panel.onDidChangeViewState(async () => {
+      if (await this.isTheMostRecentlyActivePanel()) {
+        await this.updateModelDetailsPanelState(
+          this.externalApiUsages,
+          this.databaseItem,
+        );
+      }
+    });
+
```
koesie10, line 109: Should we also add something that makes sure that any references to the panel are set to `undefined` when the panel is disposed? If we don't, you might expect methods on `mostRecentlyActivePanel` to work, while they would always fail if the panel is disposed.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -48,8 +49,12 @@ export class DataExtensionsEditorView extends AbstractWebview<
   ToDataExtensionsEditorMessage,
   FromDataExtensionsEditorMessage
 > {
+  private static mostRecentlyActivePanel: WebviewPanel | undefined = undefined;
```
koesie10, line 52: I'm not sure whether a static variable is the best option here. We could also store this on the module as an instance variable, and I think that would behave more predictably. We would need some way to expose this information to the view, but I think we're doing something very similar for the variant analysis views.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
robertbrignull: That is a very good point, that it's not clear from the name that it also updates state. I'll split it into two methods.

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
robertbrignull: Ah, good idea. Yes, storing it on the module should be fine because there's only one module object. I'll move it to there, and then I guess we'll need to pass in a couple of method closures to allow the view to announce that it is now the active view, and check if it is still the most recently active.

robertbrignull: The `DataExtensionsEditorView` is gaining a number of new parameters, but it's for the best here I think. It's also something that should be easy to rewrite in the future if we change our minds.

PR #2710, opened 2023-08-15
v1.8.10

charisk:
Release v1.8.10


shati-patel (approved):
🚀

extensions/ql-vscode/package.json
```diff
@@ -14,7 +14,7 @@
   },
   "engines": {
     "vscode": "^1.67.0",
-    "node": "^16.13.0",
+    "node": "^16.17.1",
```
robertbrignull, line 17: This value seems to be in `package-lock.json` too. Do we need to run `npm install` or something like that, or is it fine with just this change?

charisk: Yep, done now.

PR #2716, opened 2023-08-16
Rename files in src/view/results to match their components

robertbrignull:
We have a bunch of files in `src/view/results` where the filename does not match the name of the component they contain. We may as well rename them all at once.

## Checklist


robertbrignull:
I don't know what's going on with this `Graph` import.

If I import `"./Graph"` then it works locally when I run `npm run build`, and this makes sense to me because that's what the file is called. However my VS Code complains locally and it fails in CI.

If I rename the import to `"./graph"` then VS Code is happy, but `npm run build` fails locally. 😕 

Is there something special about this file? It's not failing on any of the other imports. I was wondering if this file is imported from somewhere else like a special javascript file, because the graph viewer does do some unusual things, but I can't find anything like that.

robertbrignull:
This is the error I get locally in VS Code:
Screenshot 2023-08-16 at 11 51 00

This was fixed by a restart. This maybe makes sense that VS Code was confused by the renamed file and had to have the typescript server be restarted. This [thread](https://github.com/microsoft/TypeScript/issues/43838) helped me work it out. 

Unfortunately that doesn't explain why CI is having problems, because everything in that thread is about needing a restart or prompting typescript to see the new files.

I am wondering if maybe we should change the tsconfig `"include": ["src/**/*.ts"]` to `"include": ["src/**/*.ts(x)"]`, but I'm not confident on that.

robertbrignull:
I worked out the problem is that I didn't actually commit the `graph.tsx => Graph.tsx` rename, because I'm on OSX so its case-insensitive filesystem didn't see the rename. I worked around this by doing the rename in multiple steps by renaming to a completely different intermediate name. I then squashed the commits and git seems to have kept the rename.

I hope this works and doesn't cause problems for anyone else who views this repo on OSX. I had problems when changing branch because it didn't update the graph file and it kept the wrong case. 😟 

We could fully rename the file but I can't think of a good thing to call it. Or just accept we might have some issues when this PR is merged and people pull the changes.

robertbrignull:
> Perhaps `git mv` would have helped with the renaming issues you came across?

Good point. `git mv` probably would have helped. But git knows about the move now so I don't think it'll change how it behaves for other people pulling this code once merged.

PR #2719, opened 2023-08-16
Fix button accessibility

koesie10:
This replaces all `<a href="#" onClick={...}>...</a>` by `<button>` element styled as a link using the `TextButton` component. This improves accessibility. There should be no changes in behavior or UI.

## Checklist


extensions/ql-vscode/src/view/results/ResultTablesHeader.tsx
```diff
@@ -122,17 +128,9 @@ export function ResultTablesHeader(props: Props) {
...
-        {/*
-            eslint-disable-next-line
-            jsx-a11y/anchor-is-valid
-          */}
-        <a
-          href="#"
-          onClick={openQueryHandler}
-          className="vscode-codeql__result-table-location-link"
-        >
+        <OpenQueryLink onClick={openQueryHandler}>
```
robertbrignull, line 131: The styling on this link seems to have changed.

Before:
Screenshot 2023-08-17 at 15 14 15

After:
Screenshot 2023-08-17 at 15 07 28

robertbrignull (commented):
Ah, I didn't know we had `TestButton`. That's cool, and I agree that's the right component to use here.

I think I've tested all the UI elements that have changed, and I found one that was a bit different than it was before, but everything else seems good.

extensions/ql-vscode/src/view/common/TextButton.tsx
```diff
@@ -8,19 +8,29 @@ const StyledButton = styled.button<{ size: Size }>`
...
   onClick: () => void;
+  className?: string;
+  title?: string;
   children: React.ReactNode;
 }) => (
-  <StyledButton size={size} onClick={onClick}>
+  <StyledButton
+    size={size}
+    onClick={onClick}
+    className={className}
```
charisk, line 31: Apologies for the drive-by comment, just a thought: Seems a bit of a shame to be changing our nice, self-contained button and adding this reliance to external CSS classes. Should we create a new component instead? It'd be easier to remove it later on than changing this one again I think.

koesie10: This is not actually used for passing the `className` directly to the component, but is used so we can call `styled(TextButton)` to add to the styles of this component. We should never be passing `className` directly, but being able to augment the styles is really useful: https://github.com/github/vscode-codeql/pull/2719/files#diff-919ab8acd98848b2bd69fe5ef5b98c920c6fdf5db48cf04e9e259d333c0fa6b6R17-R21

charisk: Oh right. That's some confusing syntax!

extensions/ql-vscode/src/view/results/ResultTablesHeader.tsx
koesie10: Thanks for noticing this, I've changed the styling of that component to use styled-components exclusively which fixes the issue.

robertbrignull:
Fixed the merge conflicts with https://github.com/github/vscode-codeql/pull/2726

robertbrignull (approved):
LGTM

PR #2729, opened 2023-08-21
Split out CLI version tests to separate workflow

charisk:
We currently run all CLI integration tests for every supported version of the CodeQL CLI, for windows and linux and for every commit of every PR. This is wasteful and it also adds minutes and chance of hitting flakiness.

We've also not generally seen tests succeed on one version and fail on another (outside of flakes), so pragmatically it doesn't make a huge amount of sense to run these tests so often. There aren't a lot of code changes that are related to different CLI versions.

This PR changes the workflows so that instead of running the whole matrix every time, we:
- Run only against the latest released CLI version for PR workflows
- We have a daily run of the CLI version tests

If we're finding that we're missing legit failures on the new workflow, we can add a process to notify the team when a failure happens. I'm thinking that at the moment it's worth doing this as it may be too noisy (due to flakiness).

You can see the new workflow here: https://github.com/github/vscode-codeql/actions/runs/5923448364

## Checklist
N/A:

### DIFF
--- a/.github/workflows/cli-test.yml
+++ b/.github/workflows/cli-test.yml
@@ -3,7 +3,8 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '0 0 * * *'
-  push:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
     paths:
       - extensions/ql-vscode/src/codeql-cli/**
       - extensions/ql-vscode/src/language-support/**
--- a/.github/workflows/cli-test.yml
+++ b/.github/workflows/cli-test.yml
@@ -3,9 +3,6 @@ on:
   workflow_dispatch:
   schedule:
     - cron: '0 0 * * *'
-  push:
-    branches:
-      - charisk/split-out-cli-tests
 
 jobs:
   find-nightly:
--- a/.github/workflows/cli-test.yml
+++ b/.github/workflows/cli-test.yml
@@ -1,0 +1,103 @@
+name: Run CLI tests
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  find-nightly:
+    name: Find Nightly Release
+    runs-on: ubuntu-latest
+    outputs:
+      url: ${{ steps.get-url.outputs.nightly-url }}
+    steps:
+      - name: Get Nightly Release URL
+        id: get-url
+        env:
+          GITHUB_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+        shell: bash
+        # This workflow step gets an unstable testing version of the CodeQL CLI. It should not be used outside of these tests.
+        run: |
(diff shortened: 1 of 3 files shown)


robertbrignull:
We'll need to remove all the branch protection rules before we can merge this. Which is also a step that'll need to be removed from the CodeQL CLI release process docs.

But fun fact, you can see from "13 successful and 13 expected checks" that this PR reduces the number of CI workflow runs by 50% 😌

robertbrignull (approved):
LGTM

Do we want/need more review on this from the CodeQL team, or are you happy to just go ahead?

charisk:
Let's go ahead - we can always revert if people disagree.

.github/workflows/main.yml
```diff
@@ -198,31 +182,32 @@ jobs:
...
     name: CLI Test
     runs-on: ${{ matrix.os }}
-    needs: [find-nightly, set-matrix]
+    needs: [get-latest-cli-version]
     timeout-minutes: 30
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        version: ${{ fromJson(needs.set-matrix.outputs.cli-versions) }}
+        version: ['${{ needs.get-latest-cli-version.outputs.cli-version }}']
```
robertbrignull, line 207: A thought, that I didn't think of before approving: don't include this in the matrix here. This will make the name of the CI job just `CLI Test (ubuntu-latest)` instead of `CLI Test (ubuntu-latest, v2.14.2)` and it will mean the job name won't keep changing over time as the CLI gets updated. This means we won't have to update the branch protection rules every time the CLI version changes.

charisk: Good shout!

charisk:
Apologies I should have mentioned the plan in the PR description (because even I forgot about it for a minute 🤦 !). I was thinking of the following:
- Get PR reviewed from our team (✅) to make sure we're on the same page
- Check CodeQL experiences folks are happy with this change and create PRs to update the various docs on this
- Remove protection rules
- Merge PR

.github/workflows/cli-test.yml
```diff
@@ -0,0 +1,108 @@
+name: Run CLI tests
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
+  push:
```
aeisenberg, line 6: Should this be `pull_request` instead? Or additionally?

aeisenberg (approved):
It's probably fine as it is, but I just want to be sure. Feel free to merge if the answers to my questions are both _no_.

.github/workflows/cli-test.yml
charisk: 🤦‍♀️ It should be pull_request, like in the main workflow.

charisk:
@robertbrignull let me know if you're still happy with this PR to go ahead

robertbrignull (approved):
I'm still happy with this 👍🏼

PR #2731, opened 2023-08-21
Make automodel and flow queries work without submodule.

starcke:
Resolve automodel and flow queries without submodule.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/auto-model-codeml-queries.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/auto-model-codeml-queries.ts
@@ -3,7 +3,6 @@ import { CoreCompletedQuery, QueryRunner } from "../query-server";
 import { DatabaseItem } from "../databases/local-databases";
 import { ProgressCallback } from "../common/vscode/progress";
 import * as Sarif from "sarif";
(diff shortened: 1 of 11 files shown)


charisk (approved):

aeisenberg:
So, this is creating a lock file in the downloaded library pack? It's generally something we should be avoiding since downloaded packs should be read-only. It looks like you have a cleanup job running that deletes the lock file, so that's a bit better. One concern I have is that you're likely to get different lock files over time as we release new transitive dependencies.

Moving the queries to their own query pack would be a better approach, in my view.

starcke:
> So, this is creating a lock file in the downloaded library pack? It's generally something we should be avoiding since downloaded packs should be read-only. It looks like you have a cleanup job running that deletes the lock file, so that's a bit better. One concern I have is that you're likely to get different lock files over time as we release new transitive dependencies.

It was doing that for the automodel queries (after I integrated them), and that was actually what corrupted my pack directory (as per our discussion on slack). I have now removed that lock creation from the automodel and flow queries. It is now only used by the fetch queries which need it (and here the lock creation happens in a temp dir), but longer term I would prefer to remove the creation of the lock completely as it feels very risky to use.

charisk (approved):
Looks good and works on my machine.

PR #2732, opened 2023-08-21
Split AlertTable into smaller components

robertbrignull:
Refactors the rendering in `AlertTable`. Previous it was multiple nested `foreach` loops adding components to a `rows` array from multiple places, including the rendering of zero results or truncated results. As of this PR it's a series of hierarchical components.

Each small change is done as a separate commit, so theoretically you should be able to follow along with them. Or it may be easier to review the end result on its own if that's what you prefer.

I've tried to test manually and as far as I can tell, everything looks and behaves exactly the same as before 🎉 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/results/AlertTablePathNodeRow.tsx
+++ b/extensions/ql-vscode/src/view/results/AlertTablePathNodeRow.tsx
@@ -4,7 +4,7 @@ import * as Keys from "./result-keys";
 import { SarifLocation } from "./locations/SarifLocation";
 import { selectableZebraStripe } from "./result-table-utils";
 import { ScrollIntoViewHelper } from "./scroll-into-view-helper";
-import { useMemo } from "react";
+import { useCallback, useMemo } from "react";
 
 interface Props {
   step: Sarif.ThreadFlowLocation;
@@ -16,7 +16,7 @@ interface Props {
   sourceLocationPrefix: string;
   updateSelectionCallback: (
     resultKey: Keys.PathNode | Keys.Result | undefined,
-  ) => () => void;
+  ) => void;
   scroller: ScrollIntoViewHelper;
 }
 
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/view/results/AlertTablePathNodeRow.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const pathNodeKey: Keys.PathNode = useMemo(
+    () => ({
+      resultIndex,
+      pathIndex,
+      pathNodeIndex,
+    }),
+    [pathIndex, pathNodeIndex, resultIndex],
+  );
+  const handleSarifLocationClicked = useMemo(
```
charisk, line 44: I think this should use `useCallback`

extensions/ql-vscode/src/view/results/AlertTablePathRow.tsx
```diff
@@ -0,0 +1,78 @@
...
+    selectedItem,
+    toggleExpanded,
+    scroller,
+  } = props;
+
+  const pathKey = useMemo(
+    () => ({ resultIndex, pathIndex }),
+    [pathIndex, resultIndex],
+  );
+  const handleDropdownClick = useMemo(
```
charisk, line 40: I think this should use `useCallback`

extensions/ql-vscode/src/view/results/AlertTableResultRow.tsx
```diff
@@ -0,0 +1,125 @@
...
+    toggleExpanded,
+    scroller,
+  } = props;
+
+  const resultKey: Keys.Result = useMemo(
+    () => ({ resultIndex }),
+    [resultIndex],
+  );
+
+  const handleSarifLocationClicked = useMemo(
```
charisk, line 45: I think this and `handleDropdownClick` should use `useCallback`

charisk (approved):
LGTM in general although I can't promise I've gone through every line and character change since there is a fair bit of code moving around. I think there are a couple of cases where we're using `useMemo` when we should be using `useCallback` (unless I misread the code) but otherwise LGTM!

Thanks for making this better gradually - it's great to see the individual components now.

PR #2735, opened 2023-08-22
Add tests for Model Details View

norascheuch:
Adds tests for
- model details panel set state
- reveal logic
- model details provider getChildren

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-data-provider.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-data-provider.test.ts
@@ -1,13 +1,11 @@
 import { CodeQLCliServer } from "../../../../../src/codeql-cli/cli";
-import {
-  ExternalApiUsage,
-  Usage,
-} from "../../../../../src/data-extensions-editor/external-api-usage";
-import {
-  ModelDetailsDataProvider,
-  ModelDetailsTreeViewItem,
-} from "../../../../../src/data-extensions-editor/model-details/model-details-data-provider";
+import { ExternalApiUsage } from "../../../../../src/data-extensions-editor/external-api-usage";
+import { ModelDetailsDataProvider } from "../../../../../src/data-extensions-editor/model-details/model-details-data-provider";
 import { DatabaseItem } from "../../../../../src/databases/local-databases";
+import {
+  createExternalApiUsage,
+  createUsage,
+} from "../../../../factories/data-extension/external-api-factories";
 import { mockedObject } from "../../../utils/mocking.helpers";
 
 describe("ModelDetailsDataProvider", () => {
@@ -97,26 +95,31 @@ describe("ModelDetailsDataProvider", () => {
   });
 
   describe("getChildren", () => {
+    const supportedExternalApiUsage = createExternalApiUsage({
+      supported: true,
+    });
+
+    const unsupportedExternalApiUsage = createExternalApiUsage({
+      supported: false,
+    });
+
     const externalApiUsages: ExternalApiUsage[] = [
-      mockedObject<ExternalApiUsage>({ supported: true }),
-      mockedObject<ExternalApiUsage>({ supported: false }),
+      supportedExternalApiUsage,
+      unsupportedExternalApiUsage,
     ];
     const dbItem = mockedObject<DatabaseItem>({
       getSourceLocationPrefix: () => "test",
     });
 
+    const usage = createUsage({});
+
     it("should return [] if item is a usage", async () => {
-      const item = { usages: undefined } as unknown as ModelDetailsTreeViewItem;
-      expect(dataProvider.getChildren(item)).toEqual([]);
+      expect(dataProvider.getChildren(usage)).toEqual([]);
     });
 
     it("should return usages if item is external api usage", async () => {
-      const usage = mockedObject<Usage>({});
-      const item = mockedObject<ExternalApiUsage>({
-        usages: [usage],
-      });
-
-      expect(dataProvider.getChildren(item)).toEqual([usage]);
+      const externalApiUsage = createExternalApiUsage({ usages: [usage] });
+      expect(dataProvider.getChildren(externalApiUsage)).toEqual([usage]);
     });
 
     it("should show all externalApiUsages if hideModeledApis is false and item is undefined", async () => {
--- a/extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-data-provider.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-data-provider.test.ts
@@ -6,6 +6,11 @@ import { mockedObject } from "../../../utils/mocking.helpers";
 
 describe("ModelDetailsDataProvider", () => {
   const mockCliServer = mockedObject<CodeQLCliServer>({});
+  let dataProvider: ModelDetailsDataProvider;
+
+  beforeEach(() => {
+    dataProvider = new ModelDetailsDataProvider(mockCliServer);
+  });
 
   describe("setState", () => {
     const hideModeledApis: boolean = false;
@@ -15,7 +20,6 @@ describe("ModelDetailsDataProvider", () => {
     });
 
     it("should not emit onDidChangeTreeData event when state has not changed", async () => {
-      const dataProvider = new ModelDetailsDataProvider(mockCliServer);
       await dataProvider.setState(externalApiUsages, dbItem, hideModeledApis);
 
       const onDidChangeTreeDataListener = jest.fn();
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-data-provider.test.ts
```diff
@@ -89,4 +95,40 @@ describe("ModelDetailsDataProvider", () => {
       expect(onDidChangeTreeDataListener).toHaveBeenCalledTimes(1);
     });
   });
+
+  describe("getChildren", () => {
+    const externalApiUsages: ExternalApiUsage[] = [
+      mockedObject<ExternalApiUsage>({ supported: true }),
```
robertbrignull, line 101: I see that the desire here is to avoid the pain of providing full objects, but `mockedObject` is going to potentially cause some issues here that will be hard to debug for future developers. I think you've already hit this, as evidenced by needing to pass `usages: undefined` in some cases.

In `ModelDetailsDataProvider` we use `isExternalApiUsage` to determine if the object we have is an `ExternalApiUsage` or a `Usage`. Unfortunately this check won't work correctly when given an object created using `mockedObject`, or at least you have to know exactly what you're doing and pass the right field values to make `isExternalApiUsage` behave correctly. While it is possible to have everything work correctly like this, it's quite error prone.

While it's is great for things like `CodeQLCliServer`, I would recommend not using `mockedObject` for types like `ExternalApiUsage` and `Usage`. How much effort would it be to not do this?

extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-panel.test.ts
```diff
@@ -0,0 +1,98 @@
...
+        supportedType: "none",
+        usages: [],
+        signature: "test",
+        packageName: "test",
+        typeName: "test",
+        methodName: "test",
+        methodParameters: "test",
+      },
+    ];
+    const dbItem = mockedObject<DatabaseItem>({
```
robertbrignull, line 29: We define the exact same `dbItem` in each `describe` block. Could we move the definition one level higher, like where we make `mockCliServer`?

extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-panel.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    beforeEach(() => {
+      mockTreeView = {
+        reveal: jest.fn(),
+      } as unknown as TreeView<unknown>;
+      jest.spyOn(window, "createTreeView").mockReturnValue(mockTreeView);
+    });
+
+    it("should reveal the correct item in the tree view", async () => {
+      const externalApiUsages = mockedObject<ExternalApiUsage[]>([
```
robertbrignull, line 73: Again as commented on the other file, I suggest avoiding using `mockedObject` for the `ExternalApiUsage` type. We already define a full `ExternalApiUsage` object earlier in this file, so could we reuse that? Or perhaps add a factory method for making `ExternalApiUsage` objects (see the `test/factories` directory)?

robertbrignull (commented):
I like what the tests are testing. That's exactly what I meant and wanted to see! 🙇🏼 

I think the tests would be better and more robust to changes if we reduced the frequency of `mockedObject` and defined objects more fully. What are the thoughts of others?

extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-data-provider.test.ts
```diff
@@ -89,4 +95,40 @@ describe("ModelDetailsDataProvider", () => {
...
+      mockedObject<ExternalApiUsage>({ supported: false }),
+    ];
+    const dbItem = mockedObject<DatabaseItem>({
+      getSourceLocationPrefix: () => "test",
+    });
+
+    it("should return [] if item is a usage", async () => {
+      const item = { usages: undefined } as unknown as ModelDetailsTreeViewItem;
```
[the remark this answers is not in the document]
norascheuch: 👍

extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-panel.test.ts
norascheuch: 👍

extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-data-provider.test.ts
```diff
@@ -89,4 +93,45 @@ describe("ModelDetailsDataProvider", () => {
...
+      expect(dataProvider.getChildren(usage)).toEqual([]);
+    });
+
+    it("should return usages if item is external api usage", async () => {
+      const externalApiUsage = createExternalApiUsage({ usages: [usage] });
+      expect(dataProvider.getChildren(externalApiUsage)).toEqual([usage]);
+    });
+
+    it("should show all externalApiUsages if hideModeledApis is false and item is undefined", async () => {
+      const hideModeledApis: boolean = false;
```
charisk, line 126: (Minor) you shouldn't need to define the type
```suggestion
      const hideModeledApis = false;
```
Same on L132 and a few places in `model-details-panel.test.ts` - worth doing a search for `const hideModeledApis: boolean =` to find all occurrences.

extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-panel.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  describe("revealItem", () => {
+    let mockTreeView: TreeView<unknown>;
+
+    const hideModeledApis: boolean = false;
+    const usage = createUsage();
+
+    beforeEach(() => {
+      mockTreeView = {
+        reveal: jest.fn(),
+      } as unknown as TreeView<unknown>;
```
charisk, line 44: This is similar to `as any as X` so we should ideally avoid it. Is there a way to get around it easily, or is creating mock `TreeView<>T>` items too painful? I'm happy to take it on as tech debt too.

extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-data-provider.test.ts
```diff
@@ -89,4 +93,45 @@ describe("ModelDetailsDataProvider", () => {
...
+    it("should return [] if item is a usage", async () => {
+      expect(dataProvider.getChildren(usage)).toEqual([]);
+    });
+
+    it("should return usages if item is external api usage", async () => {
+      const externalApiUsage = createExternalApiUsage({ usages: [usage] });
+      expect(dataProvider.getChildren(externalApiUsage)).toEqual([usage]);
+    });
+
+    it("should show all externalApiUsages if hideModeledApis is false and item is undefined", async () => {
```
charisk, line 125: (Minor) The `item is undefined` bit confused me a little bit but I realised what it meant once I got to know the code a bit better. We don't have to change it, but do you think the following description is accurate/better:

```suggestion
    it("should show all externalApiUsages if hideModeledApis is false and looking at the root", async () => {
```

charisk (approved):
LGTM! Just some comments for potential improvements.

extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-data-provider.test.ts
norascheuch: sounds good, will rename!

extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-data-provider.test.ts
norascheuch: 👍

extensions/ql-vscode/test/vscode-tests/no-workspace/data-extensions-editor/model-details/model-details-panel.test.ts
charisk: Lets leave it as is. I imagine creating the mock properly isn't trivial.

robertbrignull: Thanks for talking this through. I think `mockedObject` could be a good solution for this case. I've opened https://github.com/github/vscode-codeql/pull/2768 for this where I've tried to explain my reasoning, so let me know what you think.

PR #2738, opened 2023-08-22
Convert ScrollIntoViewHelper into a React hook

robertbrignull:
This converts `ScrollIntoViewHelper` into a custom react hook. Hopefully this makes it easier to use because you call a single method and internally it tracks when to scroll, instead of needing to manually register when scrolling is needed and when to execute the scroll.

I think we still need to pass a `React.RefObject` around to the actual elements so we can call `getBoundingClientRect`, unless there's a way around this too in which case I'm all for it.

From the manual testing I've done this appears to be behaving exactly as it did before.

## Checklist


charisk (approved):
Nice!

PR #2740, opened 2023-08-23
Store query `@kind` when running a variant analysis

shati-patel:
In a follow-up PR, we'll use this to determine whether the MRVA results view needs to show an option to switch between raw/interpreted results.

See internal linked issue for more details! 

## Checklist

N/A—no user impact yet 🐯 

### DIFF
--- a/extensions/ql-vscode/src/query-history/store/query-history-variant-analysis-domain-mapper.ts
+++ b/extensions/ql-vscode/src/query-history/store/query-history-variant-analysis-domain-mapper.ts
@@ -55,6 +55,7 @@ function mapVariantAnalysisDtoToDto(
       filePath: variantAnalysis.query.filePath,
       language: mapQueryLanguageToDto(variantAnalysis.query.language),
       text: variantAnalysis.query.text,
+      kind: variantAnalysis.query.kind,
     },
     databases: {
       repositories: variantAnalysis.databases.repositories,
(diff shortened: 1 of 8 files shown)


charisk (approved):
Change so far LGTM!

Do we also need to store that information in query history items? 🤔

shati-patel:
> Change so far LGTM!
> 
> Do we also need to store that information in query history items? 🤔

I was hoping to avoid changing the query history items 🤞🏽 But now I'm not so sure... We might need the `kind` in [here](https://github.com/github/vscode-codeql/blob/1fc735639bc71289bd667bf7da08d238661ca3b1/extensions/ql-vscode/src/query-history/store/query-history-variant-analysis-dto.ts#L21-L26) if we're opening variant analysis results from the query history.

charisk:
> > > Change so far LGTM!
> > > Do we also need to store that information in query history items? 🤔
> > 
> > 
> > I was hoping to avoid changing the query history items 🤞🏽 But now I'm not so sure... We might need the `kind` in [here](https://github.com/github/vscode-codeql/blob/1fc735639bc71289bd667bf7da08d238661ca3b1/extensions/ql-vscode/src/query-history/store/query-history-variant-analysis-dto.ts#L21-L26) if we're opening variant analysis results from the query history.
> 
> I guess the `kind` will be optional anyway, so it should actually be fine in terms of backwards compatibility of query history items.

Yep! It would be safe to add it to the existing version.

I think that if we don't persist it then it won't be available to you if you restarted the app for example.

I guess the next question is, do we really need to know the `kind` or can we just look at what result files (BQRS? SARIF? Both?) we have available and decide what to show based on that?

shati-patel:
> I guess the next question is, do we really need to know the `kind` or can we just look at what result files (BQRS? SARIF? Both?) we have available and decide what to show based on that?

Ideally, I think we want to know _before_ we load the results view whether or not to show the alerts/raw toggle. I.e. if we only have raw results, there's no point showing the toggle.

I think if we wait until we have the BQRS/SARIF files, we'll retroactively have to hide/freeze the toggle 🤔

charisk (approved):
LGTM!

PR #2750, opened 2023-08-25
Starcke/fetch queries out of view

starcke:

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
+++ b/extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
@@ -26,12 +26,8 @@ import { runFlowModelQueries } from "./flow-model-queries";
 import { promptImportGithubDatabase } from "../databases/database-fetcher";
 import { App } from "../common/app";
 import { showResolvableLocation } from "../databases/local-databases/locations";
-import { decodeBqrsToExternalApiUsages } from "./bqrs";
 import { redactableError } from "../common/errors";
-import {
-  readQueryResults,
-  runExternalApiQueries,
-} from "./external-api-usage-queries";
+import { runExternalApiQueries } from "./external-api-usage-queries";
 import { ExternalApiUsage, Usage } from "./external-api-usage";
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
```diff
@@ -28,7 +28,10 @@ import { App } from "../common/app";
 import { showResolvableLocation } from "../databases/local-databases/locations";
 import { decodeBqrsToExternalApiUsages } from "./bqrs";
 import { redactableError } from "../common/errors";
-import { readQueryResults, runQuery } from "./external-api-usage-queries";
+import {
+  readQueryResults,
```
charisk: Should this also be renamed? e.g. `readExternalApiQueryResults`

charisk (approved):

extensions/ql-vscode/src/data-extensions-editor/data-extensions-editor-view.ts
starcke: I think it is not used in the view anymore, so I think it makes sense to just be `readQueryResults`. But I am also happy to rename it.

charisk: Oh why is it imported then? 🤔

charisk: Sorry I can't read a diff right 🤦 it looks like we can remove the `export` from that function though.

starcke: I had problems with the test not being able to call it then, but maybe I am missing something?

charisk: Ah, no you're not! So technically the function shouldn't be exported because it's not used by other places in the extension code. And tests _shouldn't_ be testing that directly, but ideally should go through the entry point to the module rather than this function. 

That might be a bit of a faff atm though so I'm happy for us to take it on at tech debt.

PR #2755, opened 2023-08-28
Fix ql code in neutral fetch.

starcke:
This fixes a crash when loading models using the latest QL submodule. This was due to a change in the libraries the java query dependended on.

## Checklist


starcke:
> This looks reasonable. Is there any way to protect against similar changes in the future?

I think we should probably get the queries inside one of the packs (either automodel or std java).

> Is the extension always running against the latest released version of the java automodel queries? 

At the moment it (almost - after my PRs) is. I am thinking that we can actually enforce a specific version if/when needed?

PR #2765, opened 2023-08-30
Extract external api usage name component and reuse it

charisk:
Extracts the `ExternalApiUsageName` component into a standalone component and re-uses that across the model editor and the method modeling view.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/view/model-editor/__tests__/ExternalApiUsageName.spec.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/__tests__/ExternalApiUsageName.spec.tsx
@@ -1,0 +1,18 @@
+import * as React from "react";
+import { render as reactRender, screen } from "@testing-library/react";
+import { ExternalApiUsageName } from "../ExternalApiUsageName";
+import { ExternalApiUsage } from "../../../model-editor/external-api-usage";
+import { createExternalApiUsage } from "../../../../test/factories/data-extension/external-api-factories";
+
+describe(ExternalApiUsageName.name, () => {
+  const render = (props: ExternalApiUsage) =>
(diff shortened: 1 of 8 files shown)


extensions/ql-vscode/src/view/model-editor/__tests__/ExternalApiUsageName.spec.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+describe(ExternalApiUsageName.name, () => {
+  const render = (props: ExternalApiUsage) =>
+    reactRender(<ExternalApiUsageName {...props} />);
+
+  it("renders method name", () => {
+    const apiUsage = createExternalApiUsage();
+    render(apiUsage);
+
+    const name = `${apiUsage.packageName}.${apiUsage.typeName}.${apiUsage.methodName}${apiUsage.methodParameters}`;
```
norascheuch, line 15: I'm wondering if we should duplicate the original implementation here? Is it better to test using a regex for example?

norascheuch (approved):
Looks good! Just one minor, non blocking comment on a test.

extensions/ql-vscode/src/view/model-editor/__tests__/ExternalApiUsageName.spec.tsx
charisk: You're right it feels a bit pointless but unfortunately I can't think of a better way. I can't see how a regex would be good enough (but you're welcome to correct me ofc!). The only other way I can think of is to hardcode the values for package name, type name, method name and method parameters and then simply have a hardcoded string as the expected name.

PR #2768, opened 2023-08-31
Use mockedObject for TreeView

robertbrignull:
This is one possible solution to the conversation from https://github.com/github/vscode-codeql/pull/2735/files#r1303916205.

Effectively this is still using `as unknown as X` but at least that's now contained within `mockedObject` which mocks all other fields and methods and throws a helpful error message when code tries to use them.

I think using `mockedObject` for the `TreeView` type makes sense. It's a hard type to provide a "real" instance of, because:
- It's not a type we own or control.
- It's quite a large type with lots of methods we often don't care about.
- It's an interface that "does stuff" rather than being a data type that just contains static data.

Also importantly, we don't ever do type checking on this type, so fields being mocked doesn't cause problems with this. In the linked PR I objected to using `mockedObject` for the `ExternalApiUsage` and `Usage` types because we have the `isExternalApiUsage` method and therefore the actual fields at runtime matter and not just appeasing the typescript type system.

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/no-workspace/model-editor/methods-usage/methods-usage-panel.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/no-workspace/model-editor/methods-usage/methods-usage-panel.test.ts
@@ -39,9 +39,9 @@ describe("MethodsUsagePanel", () => {
     const usage = createUsage();
 
     beforeEach(() => {
-      mockTreeView = {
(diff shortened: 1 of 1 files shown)


charisk (approved):
Thanks! I think this is an improvement so we should go ahead with it.

I also think that it's not out of the question to provide some functions to mock TreeView objects if we wanted to, but that's something for another day.

PR #2769, opened 2023-08-31
Make the "Create query" location workspace-specific

shati-patel:
If a user picks a folder in which to create custom queries (`createQuery.qlPackLocation`), that should be saved in workspace settings, rather than global settings by default. See internal linked issue for details! 📦 

I've also changed `createQuery.autogenerateQlPacks` to get written to workspace settings, for consistency. There's no harm either way, but I think we should default to workspace settings, and only write to global settings for "higher-level" settings (such as enabling/disabling telemetry) ⚙

## Checklist

N/A—this is still feature-flagged 🎁

### DIFF
--- a/extensions/ql-vscode/src/config.ts
+++ b/extensions/ql-vscode/src/config.ts
@@ -670,7 +670,7 @@ export function getQlPackLocation(): string | undefined {
 }
 
 export async function setQlPackLocation(folder: string | undefined) {
-  await QL_PACK_LOCATION.updateValue(folder, ConfigurationTarget.Global);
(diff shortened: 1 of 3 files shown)


robertbrignull (approved):
LGTM

Us I understand it, if a value has already been written the global settings, then we'll continue to use it. But if we write a new value it'll go to workspace settings. So in that way it's compatible with existing data and that's good.

shati-patel:
> As I understand it, if a value has already been written the global settings, then we'll continue to use it. But if we write a new value it'll go to workspace settings. So in that way it's compatible with existing data and that's good.

That's exactly right! Thanks ✨

PR #2773, opened 2023-09-01
Update jump to usage to include the method

charisk:
Updates the jump to usage message and related functions to include the method. This allows us to surface method information in the method modeling panel when "View" is clicked.

Note that I've not renamed the message - let me know if you think it should be renamed.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
+++ b/extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
@@ -71,6 +71,7 @@ export class MethodsUsageDataProvider
         iconPath: new ThemeIcon("symbol-method"),
       };
     } else {
+      const method = this.getParent(item);
       return {
         label: item.label,
         description: `${this.relativePathWithinDatabase(item.url.uri)} [${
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
```diff
@@ -71,6 +71,7 @@ export class MethodsUsageDataProvider
         iconPath: new ThemeIcon("symbol-method"),
       };
     } else {
+      const method = this.getParent(item);
```
robertbrignull, line 74: This works, but I don't love that we iterate through all external API usages to find the right one. I think this is fine for this PR but then I have a suggestion for avoiding the need to call `getParent` here, and also making `getParent` itself faster. I'll open as a separate PR on top of this one.

robertbrignull (approved):
The code LGTM and I've tried it out and it's working as expected 🎉

extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
charisk: Agreed this is not great long term. I'm guessing that we'll need to start storing the relationship soon so we don't have to do this.

robertbrignull: I've opened https://github.com/github/vscode-codeql/pull/2774 which attempts to add this relationship

charisk: Cool. I'll go ahead and merge this and we can discuss there.

PR #2781, opened 2023-09-04
Emit telemetry for all messages sent from model editor

robertbrignull:
This PR should ensure that we have sufficient telemetry of actions users are taking in the model editor.

Some actions are already covered because we emit telemetry for all commands that get executed. This PR then covers all places where we send a message from the view to the extension host. Theoretically there could also be UI interactions that don't result in a message being sent to the host, but in the case of the model editor I don't believe there are any.

If in the future we do find we've missed something that we would like to have telemetry on, we can always add it.

## Checklist


starcke (approved):
Looks good. Can you link me to the place they will show up?

robertbrignull:
> Can you link me to the place they will show up?

You need to have the following settings enabled:
```
"codeQL.telemetry.enableTelemetry": true,
"codeQL.telemetry.logTelemetry": true,
```

And then look in the "CodeQL Extension Log" logs:
Screenshot 2023-09-05 at 09 48 18

starcke:
> And then look in the "CodeQL Extension Log" logs:

Ah nice! I was actually thinking about the dashboard where we could track it. This is good to know as well.

PR #2786, opened 2023-09-05
Copy more files into the synthetic variant analysis pack

aeisenberg:
Before this change and starting with CLI v2.14.3, if you wanted to run a variant analysis query and the pack it is contained in has at least one query that contains an extensible predicate, this would be an error.

The reason is that v2.14.3 introduced deep validation for data extensions. We are not copying the query that contains an extensible predicate to the synthetic pack we are generating. This means that deep validation will fail because there will be extensions that target the missing extensible predicate.

This change avoids the problem by copying any query files that contain extensible predicates to the synthetic pack. It uses the internal `generate extensible-predicate-metadata` command to discover which query files contain extensible predicates and copies them.

## Checklist


aeisenberg:
@robertbrignull would you be able to look at this?

robertbrignull (approved):
Looks reasonable to me. I don't know as much as some about what effects could happen by including a query file in the pack, but I trust you it's the right thing to do. I did manage to reproduce the error and verify that this branch fixes it.

aeisenberg:
Thanks for looking into this. 

The only (mildly) negative effect is that because we are including more query files in the synthetic query pack, we will need to compile more queries and generating the pack will take slightly longer. However, these queries are ignored since they are not part of the default query suite that is run by the variant analysis runner.

PR #2794, opened 2023-09-08
Improve responsiveness of model editor

koesie10:
This improves the responsiveness of the model editor. It is not necessary to create a lock file when running the external API usage queries because the pack has already installed the dependencies in `setUpPack`. By skipping this step, refreshing becomes almost instant with no changes rather than requiring a 5 second wait to re-install pack dependencies.

This also improves the progress notifications so it feels more responsive and is more accurate as to what the editor is doing.

I tested this both in a the `vscode-codeql-starter` and in an empty workspace, so I believe this should still work even when you do not have the `ql` submodule.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/external-api-usage-queries.ts
+++ b/extensions/ql-vscode/src/model-editor/external-api-usage-queries.ts
@@ -106,8 +106,7 @@ export async function runExternalApiQueries(
         message: update.message,
       }),
     token,
-    // We need to create a lock file, because the query is inside our own pack
-    createLockFile: true,
+    createLockFile: false,
   });
 
   if (!completedQuery) {
--- a/extensions/ql-vscode/src/model-editor/external-api-usage-queries.ts
+++ b/extensions/ql-vscode/src/model-editor/external-api-usage-queries.ts
@@ -92,7 +92,6 @@ export async function runExternalApiQueries(
 
   // Run the actual query
   const completedQuery = await runQuery({
-    cliServer,
     queryRunner,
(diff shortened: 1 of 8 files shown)


extensions/ql-vscode/src/model-editor/external-api-usage-queries.ts
```diff
@@ -106,8 +106,7 @@ export async function runExternalApiQueries(
         message: update.message,
       }),
     token,
-    // We need to create a lock file, because the query is inside our own pack
-    createLockFile: true,
+    createLockFile: false,
```
starcke, line 109: Odd, I thought I tried leaving it out and I got a failure, it must have been something else then (see https://github.com/github/vscode-codeql/pull/2731#issuecomment-1695300839). I had some troubles initially with a corrupted codeql cache.

If this is not needed, then I think we can get rid of the whole argument. It was the only use case I had for it and I was not very happy with having to add it.

starcke (approved):
Great! I was wondering whether that step could be optimized somehow.

extensions/ql-vscode/src/model-editor/external-api-usage-queries.ts
koesie10: I don't think it's necessary in this case. The `createLockFileForStandardQuery` would create a lock file for a standard query, but this is not actually a standard query, it's just a query on the filesystem.

I see that `createLockFile` is always `false` now, so I'll remove the argument.

starcke (approved):
:+1:

PR #2795, opened 2023-09-08
Set loading state when switching between modes

koesie10:
This sets the loading state when switching between application and framework mode. It does this by setting the `methods` to an empty array, which will show the loading state ([`methods.length === 0`](https://github.com/github/vscode-codeql/blob/b0eab8b9fed827a565abb1b4985d3a8a2db24df2/extensions/ql-vscode/src/view/model-editor/ModelEditor.tsx#L291)).

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/model-editor/ModelEditor.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/ModelEditor.tsx
@@ -281,7 +281,6 @@ export function ModelEditor({
       t: "switchMode",
       mode: newMode,
     });
-    setMethods([]);
   }, [viewState?.mode]);
 
   const onHideModeledApis = useCallback(() => {
--- a/extensions/ql-vscode/src/view/model-editor/ModelEditor.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/ModelEditor.tsx
@@ -281,6 +281,7 @@ export function ModelEditor({
       t: "switchMode",
       mode: newMode,
     });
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/view/model-editor/ModelEditor.tsx
```diff
@@ -281,6 +281,7 @@ export function ModelEditor({
       t: "switchMode",
       mode: newMode,
     });
+    setMethods([]);
```
robertbrignull, line 284: In practice this works just fine.

The only thing I'm wondering is if it's ok that the methods in `ModelEditor` and in `ModelEditorView` are getting temporarily out of sync. Another option could be that when handling the `switchMode` message, we set `ModelEditorView.methods = []` and emit a `setMethods` message. Ultimately it doesn't matter much because we then call `loadExternalApiUsages` which updates the methods.

In this case I think this is ok because the methods get recomputed immediately. But it is something to be aware of generally that we need to keep this state in sync.

robertbrignull (approved):

extensions/ql-vscode/src/view/model-editor/ModelEditor.tsx
koesie10: Good point, I think setting the `methods` on the `ModelEditorView` is better, so I'll update the PR to do that instead.

robertbrignull (approved):

PR #2805, opened 2023-09-12
Relax 'resolve ram' test

nickrolfe:
The test currently assumes that the CLI will use an exact 50/50 split between heap and off-heap RAM, but this is not guaranteed. Relax the test so that it doesn't care about the specific numbers.


shati-patel (approved):
Looks good, thanks! 🌷

PR #2815, opened 2023-09-14
Remove unnecessary `activationEvents` from extension's package.json

norascheuch:
From VS Code version 1.74.0 onwards, we no longer need some activations events.

## Checklist


koesie10 (commented):
To do this, we first need to ensure we only support VS Code version 1.74.0 and up. I've placed a link in the internal issue to the process we need to follow to do that.

extensions/ql-vscode/package.json
```diff
@@ -34,27 +34,8 @@
     }
   },
   "activationEvents": [
-    "onLanguage:ql",
     "onLanguage:ql-summary",
```
koesie10, line 37: Can this also be removed? It seems like this is also included in our `contributes.languages`.

norascheuch: Yes indeed! In https://code.visualstudio.com/api/references/activation-events#onLanguage it's stated at some point. Thank you! 
It's 'interesting' that we activate the extension when the evaluator-log.summary is opened (why?..) by defining a 'language' that is connected to this file.

koesie10: It seems like we [do have some language support for summary files](https://github.com/github/vscode-codeql/blob/d0df2966c52e4f25124b59ca1a28717461c3ef14/extensions/ql-vscode/src/log-insights/summary-language-support.ts), so this is probably the easiest way to actually make VS Code use that language.

koesie10 (approved):
LGTM, thanks!

PR #2824, opened 2023-09-15
Respect "telemetry.telemetryLevel" as well as "telemetry.enableTelemetry"

robertbrignull:
This PR tries to make our telemetry better behaved and to fully support the `telemetry.telemetryLevel` setting, as well as the legacy `telemetry.enableTelemetry` setting.

Unfortunately it's a bit hard to work out the intended semantics of the settings. I've been using the following as reference:
- https://code.visualstudio.com/updates/v1_61#_telemetry-settings
- https://code.visualstudio.com/docs/getstarted/telemetry
- https://code.visualstudio.com/api/extension-guides/telemetry

I think we were already partially supporting `telemetry.telemetryLevel` because we're using the `vscode-extension-telemetry` package. This means it'll only send telemetry when the global settings are enabled. I think we weren't correctly supporting `telemetryLevel: "error"` though because we weren't calling `sendTelemetryErrorEvent` for errors and we were instead using `sendTelemetryEvent` for everything.

We also weren't respecting `telemetry.telemetryLevel` in `requestTelemetryPermission` when deciding whether or not to prompt the user to enable telemtry for our extension. If the user has set `telemetryLevel: "off"` we should now no longer prompt the user.

I also clear up a couple of small unrelated things in the `telemetry.ts` file that I noticed while doing this.

## Checklist


robertbrignull:
I've tried to test this by building the extension and installing it locally, but annoying my telemetry events don't seem to be showing up on the other side. I'll have another go at testing it next week.

extensions/ql-vscode/package.json
```diff
@@ -446,7 +446,7 @@
             "type": "boolean",
             "default": false,
             "scope": "application",
-            "markdownDescription": "Specifies whether to send CodeQL usage telemetry. This setting AND the global `#telemetry.enableTelemetry#` setting must be checked for telemetry to be sent to GitHub. For more information, see the [telemetry documentation](https://codeql.github.com/docs/codeql-for-visual-studio-code/about-telemetry-in-codeql-for-visual-studio-code)"
+            "markdownDescription": "Specifies whether to send CodeQL usage telemetry. This setting AND the one of the global telemetry settings (`#telemetry.enableTelemetry#` or `#telemetry.telemetryLevel#`) must be enabled for telemetry to be sent to GitHub. For more information, see the [telemetry documentation](https://codeql.github.com/docs/codeql-for-visual-studio-code/about-telemetry-in-codeql-for-visual-studio-code)"
```
robertbrignull, line 449: In the docs it says
> Tag your custom telemetry setting with `telemetry` and `usesOnlineServices` if you have one.

But I couldn't quite work out what this means. I need to hunt down the general docs for these settings to see how to add tags.

extensions/ql-vscode/src/common/vscode/telemetry.ts
```diff
@@ -297,3 +297,21 @@ export async function initializeTelemetry(
   ctx.subscriptions.push(telemetryListener);
   return telemetryListener;
 }
+
+function isGlobalTelemetryEnabled(): boolean {
```
robertbrignull, line 301: The docs suggest not trying to implement this logic ourselves and instead using `vscode.env.isTelemetryEnabled`. I haven't yet looked into whether we can actually do that and it does what we want, or if we do indeed have to check both config settings.

extensions/ql-vscode/src/common/vscode/telemetry.ts
```diff
@@ -297,3 +297,21 @@ export async function initializeTelemetry(
...
+  // If "enableTelemetry" is set to false, no telemetry will be sent regardless of the value of "telemetryLevel"
```
robertbrignull, line 302: When I have only `telemetry.enableTelemetry` set I get this popup:

Screenshot 2023-09-15 at 16 53 16

That's where I'm getting the logic that if `telemetry.enableTelemetry` is set to `false` then it'll ignore the value of `telemetry.telemetryLevel`.

extensions/ql-vscode/src/common/vscode/telemetry.ts
```diff
@@ -297,3 +297,21 @@ export async function initializeTelemetry(
...
+  const enableTelemetry: boolean | undefined =
+    GLOBAL_ENABLE_TELEMETRY.getValue();
+  if (enableTelemetry === false) {
+    return false;
+  }
+
+  // If a value for "telemetry.telemetryLevel" is provided, then use that
+  const telemetryLevel: string | undefined = GLOBAL_TELEMETRY_LEVEL.getValue();
+  if (telemetryLevel !== undefined) {
+    return telemetryLevel === "error" || telemetryLevel === "on";
```
aeisenberg, line 312: These are the levels that I see:

Window_and_Settings_—_codeql-action

So, I think you want to change `on` to `all`.

extensions/ql-vscode/src/common/vscode/telemetry.ts
aeisenberg: This method is complicated enough that it would be nice to use existing logic for it.

extensions/ql-vscode/package.json
koesie10: I think this means that we need to add a `tags` property to our telemetry configuration options in the `package.json`. In [this documentation](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration) for the configuration contribution points, they give this example:

```json
{
  "contributes": {
    "configuration": {
      "title": "Git",
      "properties": {
        "git.alwaysSignOff": {
          "type": "boolean",
          "scope": "resource",
          "default": false,
          "description": "%config.alwaysSignOff%"
        },
        "git.ignoredRepositories": {
          "type": "array",
          "default": [],
          "scope": "window",
          "description": "%config.ignoredRepositories%"
        },
        "git.autofetch": {
          "type": ["boolean", "string"],
          "enum": [true, false, "all"],
          "scope": "resource",
          "markdownDescription": "%config.autofetch%",
          "default": false,
          "tags": ["usesOnlineServices"]
        }
      }
    }
  }
}
```

The `tags` property doesn't seem to be documented on that page though, so I'm not sure what other tags there are.

extensions/ql-vscode/src/common/vscode/telemetry.ts
robertbrignull: You're right. I was originally reading https://code.visualstudio.com/updates/v1_61#_telemetry-settings but that's a changelog entry and is apparently not outdated. In my local vscode and some other docs I also see what you see.

extensions/ql-vscode/package.json
robertbrignull: > I don't know what that means either. Do you have a link for where that's said?

I read it at https://code.visualstudio.com/api/extension-guides/telemetry#dos-and-donts

> In [this documentation](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration) for the configuration contribution points, they give this example:

Thanks for finding that example. It's a shame there isn't more documentation on this but I think there's enough we can add the `tags` field. I believe it shouldn't cause any harm so we may as well add it.

extensions/ql-vscode/src/common/vscode/telemetry.ts
robertbrignull: I've done some testing of how `isTelemetryEnabled` behaves. The answer is:
- It does indeed take into account both `telemetry.enableTelemetry` and `telemetry.telemetryLevel`
- It only returns true if telemetry is FULLY enabled. So that means `telemetryLevel` is set to `all`. If the level is set to `error` then `isTelemetryEnabled` will return false.

Maybe this is fine for us and the CodeQL extension doesn't need the granularity of only sending error reports and not other telemetry. We certainly aren't losing anything with this PR because we weren't checking the level at all before.

Are people happy with this limitation that if `telemetryLevel` is set to `error` then we won't send any telemetry?

aeisenberg (approved):
This is a lot simpler (though still not quite simple). Nice.

robertbrignull:
Fixing the tests was a little bit involved because:
- We now call `sendTelemetryErrorEvent` as well as `sendTelemetryEvent`
- We now need to mock `env.isTelemetryEnabled` as well as setting config values

robertbrignull:
@aeisenberg, are you still ok with this PR or would you like to take another look?

All the changes since you last approved have been to test files. My plan is I'll merge this on Monday.

robertbrignull:
Also see the update to the docs: https://github.com/github/codeql/pull/14408

aeisenberg:
> @aeisenberg, are you still ok with this PR or would you like to take another look?

Sorry...I missed this. This looks fine to me.

PR #2826, opened 2023-09-18
Avoid use of "as any" in gulpfile.ts/deploy.ts

robertbrignull:
Technically we could satisfy the linter by just removing the explicit typing and let it implicitly be typed as `any`, but that seems like cheating. This PR adds a type for the fields we actually reference and does a quick check that those fields exist.

Maybe not very necessary. If you think this is over the top for this piece of code feel free to say so.

## Checklist

### DIFF
--- a/extensions/ql-vscode/gulpfile.ts/deploy.ts
+++ b/extensions/ql-vscode/gulpfile.ts/deploy.ts
@@ -9,6 +9,7 @@ import {
 } from "fs-extra";
 import { resolve, join } from "path";
 import { isDevBuild } from "./dev";
+import type * as packageJsonType from "../package.json";
 
 export interface DeployedPackage {
   distPath: string;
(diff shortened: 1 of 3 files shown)


koesie10 (commented):
Could we do this instead?

```ts
import type * as packageJsonType from "../package.json";

const packageJson: typeof packageJsonType = JSON.parse(
  await readFile(packageJsonPath, "utf8"),
);
```

This should automatically get the correct type for the `package.json` file. It does seem like that also requires modifying `gulpfile.ts/tsconfig.json` to put the `rootDir` at `..`, but the build still seems to work.

robertbrignull:
Interesting. I like that idea. Getting the typing directly from the file sounds like a good idea. The only extra change I've added is that I do `resolve(__dirname, "../package.json")` inside `deployPackage` instead of having the path be passed in. This way we're not saying that whatever file path we're given will have the typing from `../package.json`. I believe this `deployPackage` function is only called from one place, so we aren't changing any behaviour.

koesie10 (approved):

PR #2827, opened 2023-09-18
Avoid use of "as any" during error handling

robertbrignull:
This PR removes any cases of `as any` in error handling.

There were quite a few cases where we were checking to see if `err.stderr` existed. As far as I can tell, this is useful in one case where `child_process.execFile` throws error with this field and it gives a slightly nicer error message for the user. In all of the other cases I cannot see why that field would be there and I think it's just a case of copy-pasting other code.

I wasn't sure where to put `getChildProcessErrorMessage` so I went with `helpers-pure.ts` alongside `getErrorMessage`. But I'm happy to put it in `cli.ts` or somewhere else if that would be cleaner.

## Checklist


aeisenberg (approved):
Nice.

PR #2833, opened 2023-09-20
Extract model type dropdown to its own component

charisk:
This will allow us to reuse logic for the new method modeling panel.

I've not added any unit tests/stories - we can do that in a separate issue.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/view/model-editor/ModelTypeDropdown.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/ModelTypeDropdown.tsx
@@ -1,0 +1,71 @@
+import * as React from "react";
+import { ChangeEvent, useCallback, useMemo } from "react";
+import { Dropdown } from "../common/Dropdown";
+import {
+  ModeledMethod,
+  ModeledMethodType,
+  Provenance,
+} from "../../model-editor/modeled-method";
+import { Method, getArgumentsList } from "../../model-editor/method";
+
+const options: Array<{ value: ModeledMethodType; label: string }> = [
+  { value: "none", label: "Unmodeled" },
+  { value: "source", label: "Source" },
+  { value: "sink", label: "Sink" },
+  { value: "summary", label: "Flow summary" },
+  { value: "neutral", label: "Neutral" },
+];
+
+type Props = {
+  method: Method;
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/view/model-editor/ModelTypeDropdown.tsx
```diff
@@ -0,0 +1,71 @@
...
+      } else if (modeledMethod?.provenance === "ai-generated") {
+        newProvenance = "ai-manual";
+      }
+
+      const updatedModeledMethod: ModeledMethod = {
+        // If there are no arguments, we will default to "Argument[this]"
+        input: argumentsList.length === 0 ? "Argument[this]" : "Argument[0]",
+        output: "ReturnValue",
+        kind: "value",
+        ...modeledMethod,
```
charisk, line 49: Note that this seems a bit off - we're spreading in the middle of the object creation, which means we could lose the property values we've set above. Seems like we should move the spread operation to L45 instead (or just do assignments explicitly).

For now I wanted to make minimal changes to the logic so I've left it as is.

robertbrignull: I agree, let's not change this in this PR, but I agree it looks odd to do the spread in the middle of the arguments. It would be good to investigate why it is this way, but overall I'd say lets just write it all out explicitly and not use the spread.

robertbrignull (approved):

extensions/ql-vscode/src/view/model-editor/ModelTypeDropdown.tsx
charisk: @koesie10 any thoughts on this?

koesie10: I think we can just remove this spread operator. It seems like all fields are overridden, so this doesn't add any value. We can use an object spread operator on `method` instead, but this can be at the end of this object.

charisk: Thanks all - I'll crate a separate PR for it.

PR #2835, opened 2023-09-20
Extract type for modeled method kind

charisk:
We've been using `ModelMethod["kind"]` instead of an explicit type which is okay but perhaps not as readable, and not very friendly to TypeScript newbies IMO. In this PR I'm adding a specific type. Happy to close the PR if people disagree.

## Checklist
N/A:


extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
```diff
@@ -91,7 +94,7 @@ function ModelableMethodRow(props: MethodRowProps) {
 
       onChange(method, {
         ...modeledMethod,
-        input: target.value as ModeledMethod["input"],
+        input: target.value as ModeledMethodKind,
```
koesie10, line 97: Both this one and the one below are not for `ModeledMethod["kind"]`. Should we also create separate types for those? I'm also fine with changing all three of these (`ModeledMethod["kind"]`, `ModeledMethod["input"]` and `ModeledMethod["output"]`) to use `string` instead since it doesn't really add a lot of value if they are just strings.

charisk: Sorry that was a bit sloppy. I've introduced some more types and remove the `as XXX` since it's not needed. Perhaps in the future we'll be able to turn them in to more "proper" types.

charisk: Sorry I managed to confuse myself. These castings aren't needed, so I've removed them.

robertbrignull (approved):
I'm fine with this, or just referring to it as `string` everywhere.

PR #2837, opened 2023-09-20
Extract model input dropdown to its own component

charisk:
This will allow us to reuse code for the new method modeling panel.

See similar PR: https://github.com/github/vscode-codeql/pull/2833

## Checklist
N/A:
- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #2839, opened 2023-09-20
Extract model output dropdown to its own component

charisk:
This will allow us to reuse code for the new method modeling panel.

See similar PR: https://github.com/github/vscode-codeql/pull/2833

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/view/model-editor/ModelOutputDropdown.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/ModelOutputDropdown.tsx
@@ -1,0 +1,67 @@
+import * as React from "react";
+import { ChangeEvent, useCallback, useMemo } from "react";
+import { Dropdown } from "../common/Dropdown";
+import { ModeledMethod } from "../../model-editor/modeled-method";
+import { Method, getArgumentsList } from "../../model-editor/method";
+
+type Props = {
+  method: Method;
+  modeledMethod: ModeledMethod | undefined;
+  onChange: (method: Method, modeledMethod: ModeledMethod) => void;
+};
+
+export const ModelOutputDropdown = ({
+  method,
+  modeledMethod,
+  onChange,
+}: Props): JSX.Element => {
+  const argumentsList = useMemo(
+    () => getArgumentsList(method.methodParameters),
+    [method.methodParameters],
+  );
+
+  const options = useMemo(
+    () => [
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/view/model-editor/ModelOutputDropdown.tsx
```diff
@@ -0,0 +1,67 @@
...
+  modeledMethod: ModeledMethod | undefined;
+  onChange: (method: Method, modeledMethod: ModeledMethod) => void;
+};
+
+export const ModelOutputDropdown = ({
+  method,
+  modeledMethod,
+  onChange,
+}: Props): JSX.Element => {
+  const argumentsList = useMemo(
```
robertbrignull, line 18: `argumentList` is only used to compute `options`, so we could merge these together and call `geArgumentList` on line 27 and it wouldn't change any behaviour. What do you think?

robertbrignull (approved):
A couple of suggestions to make the code simpler, but also LGTM as it is.

extensions/ql-vscode/src/view/model-editor/ModelOutputDropdown.tsx
```diff
@@ -0,0 +1,67 @@
...
+      { value: "Argument[this]", label: "Argument[this]" },
+      ...argumentsList.map((argument, index) => ({
+        value: `Argument[${index}]`,
+        label: `Argument[${index}]: ${argument}`,
+      })),
+    ],
+    [argumentsList],
+  );
+
+  const enabled = useMemo(
```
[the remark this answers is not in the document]
charisk: I'm following the same pattern as in https://github.com/github/vscode-codeql/pull/2837 so I'd rather keep it consistent. Happy to stick to `useMemo` or drop it, but would rather do the same across the board.

extensions/ql-vscode/src/view/model-editor/ModelOutputDropdown.tsx
charisk: Ah good point. I could do the same in `ModelInputDropdown`.

If we keep things as is we would be consistent with `ModelInputDropdown` and `ModelTypeDropdown` though.

I don't have strong opinions.

charisk:
I'm going to go ahead and merge this as I have another PR lined up but will apply any feedback in a separate PR.

PR #2851, opened 2023-09-22
Hide modeling panel when modeling editor is active

charisk:
We don't want to show the modeling panel if the main editor window is active, since they show similar things.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/model-editor/model-editor-view.ts
+++ b/extensions/ql-vscode/src/model-editor/model-editor-view.ts
@@ -106,6 +106,9 @@ export class ModelEditorView extends AbstractWebview<
           this.databaseItem,
           this.hideModeledMethods,
         );
+        await this.markModelEditorAsActive();
+      } else {
+        await this.updateModelEditorActiveContext();
       }
     });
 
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/model-editor/model-editor-view.ts
```diff
@@ -140,6 +159,17 @@ export class ModelEditorView extends AbstractWebview<
     );
   }
 
+  private isAModelEditorActive(): boolean {
+    return window.tabGroups.all.some((tabGroup) =>
+      tabGroup.tabs.some((tab) => {
+        const viewType: string | undefined = (tab.input as any)?.viewType;
```
koesie10, line 165: To avoid the `as any` here, we could do something like this:

```suggestion
        const viewType =
          tab.input &&
          typeof tab.input === "object" &&
          "viewType" in tab.input &&
          typeof tab.input.viewType === "string"
            ? tab.input.viewType
            : undefined;
```

I'm not sure if that's worth it though since it makes this much less clear. Alternatively, if the types are correct and the `TabInputWebview` is actually a class rather than an object, we should be able to do this (untested);

```suggestion
        const viewType =
          tab.input instanceof TabInputWebview ? tab.input.viewType : undefined;
```

koesie10 (approved):

extensions/ql-vscode/src/model-editor/model-editor-view.ts
charisk: I like the second approach you suggested - `TabInputWebview` seems to be a class (and I imagine it's unlikely that would change).

Since this function is based on `isAModelEditor` open, which uses similar casting, I'll raise a separate PR to make the changes in both methods.

PR #2855, opened 2023-09-25
Don't use 'as any' when checking open view

charisk:
Minor change to remove an `as any` casting. Suggested [here](https://github.com/github/vscode-codeql/pull/2851#discussion_r1335571640).

As part of that we're extracting a `isTabModelEditorView` function that is reused from two places.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/model-editor/model-editor-view.ts
+++ b/extensions/ql-vscode/src/model-editor/model-editor-view.ts
@@ -1,5 +1,6 @@
 import {
   CancellationTokenSource,
+  Tab,
   TabInputWebview,
   Uri,
   ViewColumn,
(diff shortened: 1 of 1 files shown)


koesie10 (commented):
Since these two functions are very similar, could we extract the actual check to a separate function and simplify these? Something like this function:

```ts
export function isTabModelEditorView(tab: Tab): boolean {
  if (!(tab.input instanceof TabInputWebview)) {
    return false;
  }

  // The viewType has a prefix, such as "mainThreadWebview-", but if the
  // suffix matches that should be enough to identify the view.
  return tab.input.viewType.endsWith("model-editor");
}
```

Then the first method can be simplified to:

```ts
  private isAModelEditorOpen(): boolean {
    return window.tabGroups.all.some((tabGroup) =>
      tabGroup.tabs.some((tab) => isTabModelEditorView(tab)),
    );
  }
```

And similarly for the second method.

charisk:
Good idea, I should have tidied that up.

Done now.

koesie10 (approved):
Thanks, LGTM.

PR #2857, opened 2023-09-25
Introduce modeling store and move some state there 

charisk:
Introduces a new `ModelingStore` that will be responsible for managing state related to method modeling. It stores state per database we're modeling. 

In this PR we take the first few steps to get there:
- Introducing the new `ModelingStore` and wiring it up to the module/views.
- Adding the knowledge around active database being modeled to the store.
- Moving the `methods` and `hideModeledMethods` bits of state to the new store.
- Using the three pieces of data (`methods`, `hideModeledMethods` and active db state) and wiring up events/event emitters in order to remove the various callbacks used in the methods usage panel.

There are no guards to enforce that the state doesn't get mutated outside of the store yet. It's something I think we should do, but I didn't want to add more to this PR.

We also don't enforce that a user is only allowed to open 1 model editor per database they're modeling yet but since that is a bit of a corner case, I think that's okay.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/model-editor/modeling-store.ts
+++ b/extensions/ql-vscode/src/model-editor/modeling-store.ts
@@ -1,0 +1,64 @@
+import { App } from "../common/app";
+import { DisposableObject } from "../common/disposable-object";
+import { AppEvent, AppEventEmitter } from "../common/events";
+import { DatabaseItem } from "../databases/local-databases";
+
+interface DbModelingState {
+  // Currently empty but will soon contain information about methods, etc.
+}
+
+export class ModelingStore extends DisposableObject {
+  public readonly onActiveDbChanged: AppEvent<void>;
+  public readonly onDbClosed: AppEvent<string>;
+
+  private state: Map<string, DbModelingState>;
+  private activeDb: string | undefined;
+
+  private readonly onActiveDbChangedEventEmitter: AppEventEmitter<void>;
+  private readonly onDbClosedEventEmitter: AppEventEmitter<string>;
+
+  constructor(app: App) {
+    super();
+
+    // State initialization
+    this.activeDb = undefined;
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -0,0 +1,142 @@
...
+  isActiveDb: boolean;
+}
+
+export class ModelingStore extends DisposableObject {
+  public readonly onActiveDbChanged: AppEvent<void>;
+  public readonly onDbClosed: AppEvent<string>;
+  public readonly onMethodsChanged: AppEvent<MethodsChangedEvent>;
+  public readonly onHideModeledMethodsChanged: AppEvent<HideModeledMethodsChangedEvent>;
+
+  private state: Map<string, DbModelingState>;
```
koesie10, line 31: I believe this can be `readonly` as well, unless we're re-assigning it by creating a new `Map`, which I don't think we'll do.

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -0,0 +1,142 @@
...
+  private readonly onActiveDbChangedEventEmitter: AppEventEmitter<void>;
+  private readonly onDbClosedEventEmitter: AppEventEmitter<string>;
+  private readonly onMethodsChangedEventEmitter: AppEventEmitter<MethodsChangedEvent>;
+  private readonly onHideModeledMethodsChangedEventEmitter: AppEventEmitter<HideModeledMethodsChangedEvent>;
+
+  constructor(app: App) {
+    super();
+
+    // State initialization
+    this.activeDb = undefined;
```
koesie10, line 43: I believe this is the default and basically a no-op.

koesie10 (approved):
LGTM

PR #2858, opened 2023-09-25
Use streaming when creating log symbols file.

dbartol:
Internal users were seeing frequent crashes complaining about trying to create a string that was too long to fit in memory. This was happening when we parse the human-readable log to generate the symbols that map predicates to their human-readable RA. We were simply reading the entire human-readable log into memory at once, and these can be extremely large for complex queries.

The fix was to just use a streaming reader, which required slightly rearranging the code we use to parse the lines coming out of the reader.

I also moved some existing code for splitting a stream at line break boundaries into its own source file, so we could consume it from the new code.

### DIFF
--- a/extensions/ql-vscode/src/common/split-stream.ts
+++ b/extensions/ql-vscode/src/common/split-stream.ts
@@ -1,0 +1,109 @@
+import { Readable } from "stream";
+import { StringDecoder } from "string_decoder";
+
+/**
+ * Buffer to hold state used when splitting a text stream into lines.
+ */
+class SplitBuffer {
+  private readonly decoder = new StringDecoder("utf8");
+  private readonly maxSeparatorLength: number;
+  private buffer = "";
+  private searchIndex = 0;
+
+  constructor(private readonly separators: readonly string[]) {
+    this.maxSeparatorLength = separators
+      .map((s) => s.length)
+      .reduce((a, b) => Math.max(a, b), 0);
+  }
+
+  /**
+   * Append new text data to the buffer.
+   * @param chunk The chunk of data to append.
+   */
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/common/split-stream.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  /**
+   * A version of startsWith that isn't overriden by a broken version of ms-python.
+   *
+   * The definition comes from
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
+   * which is CC0/public domain
+   *
+   * See https://github.com/github/vscode-codeql/issues/802 for more context as to why we need it.
```
aeisenberg, line 42: Looks like the upstream issues have been fixed. Maybe we can remove this workaround. (Not needed for this PR.)

extensions/ql-vscode/src/common/split-stream.ts
aeisenberg, line 1: Is there anything in this file that has changed when you extracted it?

aeisenberg (commented):
Looks reasonable to me. Do we have any tests for `SplitBuffer`?

extensions/ql-vscode/src/common/split-stream.ts
dbartol: No, other than the casing of `LINE_ENDINGS`.

dbartol: Oh, and the bug I discovered when you made me write a unit test:)

aeisenberg (approved):

boldboundless3-spec (commented):
Hi

PR #2792, opened 2023-09-08
fix syntax highlighting after imports with instantiation arguments

ginsbach:
Fixes https://github.com/github/codeql-core/issues/3697

This PR updates the TextMate grammar to allow instantiation arguments in import statements. This will improve syntax highlighting. Specifically, It is meant to fix the highlighting of `pragma` in the example below.

image

Note that I don't know the TextMate grammar world very well, so am not really sure whether what I did here makes any sense.

## Checklist

- [x] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #2859, opened 2023-09-25
Alternate fix for import highlighting with instantiations

dbartol:
This is an alternate fix for https://github.com/github/vscode-codeql/pull/2792. It updates the heuristic for finding the end of an import directive. The new heuristic matches a zero-width string that is preceded by a character that could end the import directive (identifier, `>`) and is not followed by a character that would continue the import directive (`,`, `.`, `<`, `::`).

### DIFF
--- a/extensions/ql-vscode/syntaxes/ql.tmLanguage.yml
+++ b/extensions/ql-vscode/syntaxes/ql.tmLanguage.yml
@@ -170,6 +170,14 @@ repository:
     match: '\]'
     name: punctuation.squarebracket.close.ql
 
+  open-angle:
+    match: '<'
+    name: punctuation.anglebracket.open.ql
+
+  close-angle:
+    match: '>'
+    name: punctuation.anglebracket.close.ql
+
   operator-or-punctuation:
     patterns:
     - include: '#relational-operator'
@@ -186,6 +194,8 @@ repository:
     - include: '#close-brace'
     - include: '#open-bracket'
     - include: '#close-bracket'
+    - include: '#open-angle'
+    - include: '#close-angle'
 
 # Keywords
   dont-care:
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/syntaxes/ql.tmLanguage.yml
```diff
@@ -661,29 +661,38 @@ repository:
...
-    # line breaks between the simple-id and the `.` or `::`. 
-    end: '(?#simple-id) (?!\s*(\.|\:\:))'
-    endCaptures:
-      '0':
-        name: entity.name.type.namespace.ql
+    # TextMate makes it tricky to tell whether an identifier that we encounter is part of the
+    # `import` directive or whether it's the first token of the next module-level declaration.
+    # To find the end of the import directive, we'll look for a zero-width match where the previous
+    # token is either an identifier (other than `import`) or a `>`, and the next token is not a `.`,
+    # `<`, `,`, or `::`. This works for nearly all real-world `import` directives, but it will end the
```
ginsbach, line 685: I think we also need '>' in that list of allowed next tokens. After all, we could import a nested instanitation
```ql
bindingset[this]
signature class TSig;

module M1<TSig T> {
  class type = T;
}

module M2 {
  import M1<M1<int>>
}
```

ginsbach (changes requested):
Thanks a lot for looking into this! Apologies for taking a bit of time to get back to you. I have this one minor suggestion, aside from that it looks good to me.

extensions/ql-vscode/syntaxes/ql.tmLanguage.yml
ginsbach: Actually, it seems there is no discernible difference, and even the compiled grammar stays the same if I add '>'.

ginsbach (approved):
I am approving this issue to signal that I'm happy with the changes, but there is no need to merge - I have already updated [my branch](https://github.com/github/vscode-codeql/pull/2792)

dbartol:
Closing since this was already merged into the real PR

PR #2861, opened 2023-09-26
Move modeled and modified method state to store

charisk:
Move some more state into the new modeling store. In this PR we move the modeled methods and modified methods state.

By moving this state out of the React view, the view gets simplified and uses React state only to store what it has been told by the extension host - it doesn't make any changes to the state directly by itself.

With this move we could also add some test coverage around the more complicated state update functions (e.g. `addModeledMethods`). This will be done in a future PR.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/model-editor/modeling-store.ts
+++ b/extensions/ql-vscode/src/model-editor/modeling-store.ts
@@ -101,7 +101,7 @@ export class ModelingStore extends DisposableObject {
       methods: [],
       hideModeledMethods: INITIAL_HIDE_MODELED_METHODS_VALUE,
       modeledMethods: {},
-      modifiedMethodSignatures: new Set<string>(),
+      modifiedMethodSignatures: new Set(),
     });
   }
 
--- a/extensions/ql-vscode/src/model-editor/modeling-store.ts
+++ b/extensions/ql-vscode/src/model-editor/modeling-store.ts
@@ -3,12 +3,15 @@ import { DisposableObject } from "../common/disposable-object";
 import { AppEvent, AppEventEmitter } from "../common/events";
 import { DatabaseItem } from "../databases/local-databases";
 import { Method } from "./method";
+import { ModeledMethod } from "./modeled-method";
 import { INITIAL_HIDE_MODELED_METHODS_VALUE } from "./shared/hide-modeled-methods";
 
 interface DbModelingState {
   databaseItem: DatabaseItem;
   methods: Method[];
   hideModeledMethods: boolean;
+  modeledMethods: Record<string, ModeledMethod>;
+  modifiedMethodSignatures: Set<string>;
 }
 
 interface MethodsChangedEvent {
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -129,6 +162,100 @@ export class ModelingStore extends DisposableObject {
...
+
+    this.onModifiedMethodsChangedEventEmitter.fire({
+      modifiedMethods: state.modifiedMethodSignatures,
+      dbUri: dbItem.databaseUri.toString(),
+      isActiveDb: dbItem.databaseUri.toString() === this.activeDb,
+    });
+  }
+
+  public addModifiedMethod(dbItem: DatabaseItem, methodSignature: string) {
+    this.addModifiedMethods(dbItem, new Set<string>([methodSignature]));
```
koesie10, line 239: Could we make the argument to `addModifiedMethods` an `Iterable<string>` instead so we don't need to construct a `Set`?

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -129,6 +162,100 @@ export class ModelingStore extends DisposableObject {
...
+  }
+
+  public setModifiedMethods(
+    dbItem: DatabaseItem,
+    methodSignatures: Set<string>,
+  ) {
+    const state = this.getState(dbItem);
+    state.modifiedMethodSignatures = methodSignatures;
+
+    this.onModifiedMethodsChangedEventEmitter.fire({
```
koesie10, line 214: Would it make sense to extract this out of these methods since we are repeating this code three different times?

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -129,6 +162,100 @@ export class ModelingStore extends DisposableObject {
     });
   }
 
+  public addModeledMethods(
+    dbItem: DatabaseItem,
+    methods: Record<string, ModeledMethod>,
+  ) {
+    const state = this.getState(dbItem);
+
+    const newModeledMethods = {
```
koesie10, line 171: Shouldn't this also be stored in the `state`?

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -69,6 +100,8 @@ export class ModelingStore extends DisposableObject {
       databaseItem,
       methods: [],
       hideModeledMethods: INITIAL_HIDE_MODELED_METHODS_VALUE,
+      modeledMethods: {},
+      modifiedMethodSignatures: new Set<string>(),
```
koesie10, line 104: TypeScript should be able to derive the type here:

```suggestion
      modifiedMethodSignatures: new Set(),
```

koesie10 (commented):
Would it perhaps make sense to create a method like this which will do the steps we're repeating in every method (retrieving state, updating state, setting state, firing event)?

```ts
private changeModifiedMethods(
  dbItem: DatabaseItem,
  fn: (state: Set<string>) => void,
) {
  const state = this.getState(dbItem);

  fn(state.modifiedMethodSignatures);

  this.onModifiedMethodsChangedEventEmitter.fire({
    modifiedMethods: state.modifiedMethodSignatures,
    dbUri: dbItem.databaseUri.toString(),
    isActiveDb: dbItem.databaseUri.toString() === this.activeDb,
  });
}
```

This should make it less error prone and shorten the code. For the modeled methods, the `fn` could be something like `fn: (state: Record<string, string>) => Record<string, string>`.

charisk:
> Would it perhaps make sense to create a method like this which will do the steps we're repeating in every method (retrieving state, updating state, setting state, firing event)?

Yeah. I was intending to do some refactoring in the whole file once we're done with the first proper pass but you're right calling it out - might as well do it now. I've made some changes similar to what you suggested but not quite the same, so see what you think.

koesie10 (approved):
Thanks, LGTM

PR #2869, opened 2023-09-26
Add selected method and usage state to modeling store

charisk:
This removes the need for callbacks and sets us up for easily updating the method modeling panel when the selection changes.

I'm not sure about the names that I've used (referring to method and ignoring the usage part) so open to renaming. E.g. we could use `methodUsage` (it just gets a bit long e.g. `onSelectedMethodUsageChanged`).

## Checklist
N/A:


koesie10 (approved):
LGTM, although I do think that `methodUsage` might be a better name for these events since the method does not have to change, only the usage could change.

charisk:
> LGTM, although I do think that `methodUsage` might be a better name for these events since the method does not have to change, only the usage could change.

I think the eventual goal is that method and usage could change. I'm going to leave it for now because I've already built some code on top of it, and we can circle back later.

PR #2870, opened 2023-09-27
Hook method modeling view to modeling store

charisk:
Hooks the method modeling panel to the modeling store so that the method modeling panel deals with updates events from the store, but that it also sends updates when the dropdown values change.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
+++ b/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
@@ -9,6 +9,7 @@ import { App } from "../../common/app";
 import { redactableError } from "../../common/errors";
 import { Method } from "../method";
 import { DisposableObject } from "../../common/disposable-object";
+import { ModelingStore } from "../modeling-store";
 
 export class MethodModelingViewProvider
   extends DisposableObject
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
```diff
@@ -79,4 +101,45 @@ export class MethodModelingViewProvider
...
+        }
+      }
+    });
+
+    this.modelingStore.onModifiedMethodsChanged(async (e) => {
+      if (
+        this.webviewView &&
+        e.isActiveDb &&
+        this.method &&
+        this.method.signature
```
koesie10, line 123: This would be a lot shorter since we could move the complete if-condition to 1 line:

```suggestion
        this.method?.signature
```

However, I'm wondering why we're checking whether the method signature is non-empty. Do we have cases where we would have a method without a signature?

koesie10 (approved):

extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
charisk: No that shouldn't be possible, you're right! I'll remove the signature check.

PR #2873, opened 2023-09-27
Add language context store.

starcke:
This adds an explicit store to contain the Language filter state outside of the databases panel. It is based on `modeling-store.ts`.

There are two points that are somewhat interesting:
- We encapsulated the LanguageState inside the store, so that the users do not need to know how "All" is implemented. This might not be needed or sufficient in the future, but for now it seemed pretty nice to do.
- The `shouldInclude` method takes a string which is asymmetric with with the `setLanguageContext`. This is because we actually check with an arbitrary string (because that is what the database item has). I wonder if there is a better way to do this.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -135,6 +135,7 @@ import { TestManagerBase } from "./query-testing/test-manager-base";
 import { NewQueryRunner, QueryRunner, QueryServerClient } from "./query-server";
 import { QueriesModule } from "./queries-panel/queries-module";
 import { OpenReferencedFileCodeLensProvider } from "./local-queries/open-referenced-file-code-lens-provider";
+import { LanguageContextStore } from "./language-context-store";
 
 /**
  * extension.ts
@@ -774,10 +775,15 @@ async function activateWithInstalledDistribution(
   void dbm.loadPersistedState();
 
   ctx.subscriptions.push(dbm);
+
+  void extLogger.log("Initializing language context.");
+  const languageContext = new LanguageContextStore(app);
+
   void extLogger.log("Initializing database panel.");
   const databaseUI = new DatabaseUI(
     app,
     dbm,
+    languageContext,
     qs,
     getContextStoragePath(ctx),
     ctx.extensionPath,
--- a/extensions/ql-vscode/src/language-context-store.ts
+++ b/extensions/ql-vscode/src/language-context-store.ts
@@ -33,7 +33,10 @@ export class LanguageContextStore extends DisposableObject {
     this.onLanguageContextChangedEmitter.fire();
   }
 
-  // TODO: comment on why string is used here
+  // This method takes a string to allow it to be used in cases
+  // where the language is not always a known one.
+  // The semantics of such an unknown langauge is that it is
+  // only included if the current language context is "All".
   public shouldInclude(language: string): boolean {
     return this.state === "All" || this.state === language;
   }
--- a/extensions/ql-vscode/src/language-context-store.ts
+++ b/extensions/ql-vscode/src/language-context-store.ts
@@ -43,11 +43,7 @@ export class LanguageContextStore extends DisposableObject {
     );
   }
 
-  // This method takes a string to allow it to be used in cases
-  // where the language is not always a known one.
-  // The semantics of such an unknown langauge is that it is
-  // only included if the current language context is "All".
-  public shouldInclude(language: string): boolean {
+  public shouldInclude(language: QueryLanguage | undefined): boolean {
     return this.languageFilter === "All" || this.languageFilter === language;
   }
 }
--- a/extensions/ql-vscode/src/language-context-store.ts
+++ b/extensions/ql-vscode/src/language-context-store.ts
@@ -1,0 +1,40 @@
+import { App } from "./common/app";
+import { DisposableObject } from "./common/disposable-object";
+import { AppEvent, AppEventEmitter } from "./common/events";
+import { QueryLanguage } from "./common/query-language";
+
+type LanguageFilter = QueryLanguage | "All";
+
+export class LanguageContextStore extends DisposableObject {
+  public readonly onLanguageContextChanged: AppEvent<void>;
+  private readonly onLanguageContextChangedEmitter: AppEventEmitter<void>;
+
+  private state: LanguageFilter;
+
+  constructor(app: App) {
+    super();
+    // State initialization
+    this.state = "All";
+
+    // Set up event emitters
+    this.onLanguageContextChangedEmitter = this.push(
+      app.createEventEmitter<void>(),
+    );
+    this.onLanguageContextChanged = this.onLanguageContextChangedEmitter.event;
+  }
+
+  public clearLanguageContext() {
+    this.state = "All";
+    this.onLanguageContextChangedEmitter.fire();
+  }
+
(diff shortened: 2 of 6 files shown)


shati-patel (approved):
This looks good to me, though admittedly we paired, so happy to get another opinion too! 🙃 

(Also worth noting: we discussed adding tests for the language store, but since it doesn't do much on its own, we'll add tests when integrating with the individual components 🧪)

extensions/ql-vscode/src/language-context-store.ts
```diff
@@ -0,0 +1,40 @@
...
+    this.state = "All";
+    this.onLanguageContextChangedEmitter.fire();
+  }
+
+  public setLanguageContext(language: QueryLanguage) {
+    this.state = language;
+    this.onLanguageContextChangedEmitter.fire();
+  }
+
+  // TODO: comment on why string is used here
```
shati-patel, line 36: A leftover "TODO" 👀 Do you want to add a comment here, or shall we remove this? 😄

starcke: I was leaving it to highlight the question, but I did that in the PR message, so maybe I can just remove it :)

starcke:
I removed the TODO (and added a comment). Then I fixed the conflicts.

Lastly I noticed that when the extension was initially started up the "All" value didnt show as selected, because the context value wasnt set yet.
The last commit fixes this (by making the context look for empty string instead) and moves the context manipulation into the state.

extensions/ql-vscode/src/language-context-store.ts
```diff
@@ -0,0 +1,53 @@
...
+      "codeQLDatabases.languageFilter",
+      language,
+    );
+  }
+
+  // This method takes a string to allow it to be used in cases
+  // where the language is not always a known one.
+  // The semantics of such an unknown langauge is that it is
+  // only included if the current language context is "All".
+  public shouldInclude(language: string): boolean {
```
charisk, line 50: Can we use [`isQueryLanguage`](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/common/query-language.ts#L62) here to help us out?

extensions/ql-vscode/src/extension.ts
```diff
@@ -774,10 +775,15 @@ async function activateWithInstalledDistribution(
   void dbm.loadPersistedState();
 
   ctx.subscriptions.push(dbm);
+
+  void extLogger.log("Initializing language context.");
+  const languageContext = new LanguageContextStore(app);
```
charisk, line 780: Are these changes meant to be behind a feature flag or did I imagine that?

extensions/ql-vscode/src/language-context-store.ts
```diff
@@ -0,0 +1,53 @@
...
+import { AppEvent, AppEventEmitter } from "./common/events";
+import { QueryLanguage } from "./common/query-language";
+
+type LanguageFilter = QueryLanguage | "All";
+
+export class LanguageContextStore extends DisposableObject {
+  public readonly onLanguageContextChanged: AppEvent<void>;
+  private readonly onLanguageContextChangedEmitter: AppEventEmitter<void>;
+
+  private state: LanguageFilter;
```
charisk, line 12: (Optional) I'd personally call this `languageFilter` rather than `state` to make code further down easier to read.

extensions/ql-vscode/src/extension.ts
starcke: I thought they were, but I cannot see one before this PR either?

extensions/ql-vscode/src/language-context-store.ts
starcke: Not sure what we would use it for. We could check that the string is a query language, but because the only thing that can be put into the state is a query language or all, then the check inside the second clause of the disjunction should amount to the same thing. But I might be missing something?

charisk: We've made the argument of this function a `string` because the caller could have a `string` rather than a `QueryLanguage` (if I'm understanding this right). But only valid QueryLanguage values should be allowed. So it'd be good for the argument to change to a `QueryLanguage` and the caller to user `isQueryLanguage` to make sure it's sending the right thing.

extensions/ql-vscode/src/extension.ts
shati-patel: The "Languages" dropdown is feature-flagged in the UI here: https://github.com/github/vscode-codeql/blob/6aab4b4090600ce4d6ebd9183feb2cd215b809ac/extensions/ql-vscode/package.json#L1080

So it might make sense to feature-flag this code too 🎏

extensions/ql-vscode/src/language-context-store.ts
starcke: The trouble with having the caller use `isQueryLanguage` is that makes it hard to preserve the semantics (that unknown values should be shown when "All" is selected) without leaking "All" outside the store, or having to expose a method like "IsAllSelected".

One alternative is to pass `QueryLanguage | undefined` (because I actually need that for the queries panel). However that does make it more annoying for the callers, as they then have to convert their `string` to a `QueryLanguge | undefined`. But that is not the end of the world, and we could do that as a helper method similar to `isQueryLanguage`.

The last alternative is to expose `LanguageFilter` outside the store, which is also fine but it seemed nice to encapsulate the "All" implementation.

extensions/ql-vscode/src/extension.ts
starcke: :+1: I have had the canary flag as a "standard" flag, so I didnt even think about it :doh:

How do we normally flag this? If we just omit the creation of the languageContext then we'll have a lot of places where it is suddenly undefined, which seems more risky than having it in the first place?

shati-patel: For the queries panel, we check for the flag before initializing: https://github.com/github/vscode-codeql/blob/2e45dcd7b9a0aa01deeb1605e18d8eaa6515c14e/extensions/ql-vscode/src/queries-panel/queries-module.ts#L29-L33 

But in this case, I think it's fairly harmless to create it without feature-flagging 🤔 I don't really know what's "best" or "safest" 😅

extensions/ql-vscode/src/language-context-store.ts
charisk: `QueryLanguage | undefined` sounds like it makes the most sense to me (and I'm okay with the callers having to put some extra effort in) but I'm also happy with whatever you decide - I've not thought about this deeply, and it's easily changeable if we find that it's causing problems too.

extensions/ql-vscode/src/extension.ts
charisk: Ideally no code that does language filtering stuff should be executed if the flag is off. But I think it's okay to be lenient - it's all about us having enough confidence when releasing rather than sticking to specific rules.

starcke: I dont see a good way of feature-flagging this. But happy to discuss more sync.

extensions/ql-vscode/src/language-context-store.ts
starcke: Pushed a new version that takes in `QueryLanguage | undefined`.

starcke:
Are you both happy with the changes?

charisk (approved):
🚢

shati-patel (approved):

PR #2884, opened 2023-09-28
Avoid double restarts of the query server

aeisenberg:
Previously, if there was an explicit restart of the query server (eg- by changing a configuration setting), then the query server process would be started twice: once by the `close` handler and once by the restart command.

By adding the `removeAllListeners` to the dispose method, we ensure that when the query server shuts down gracefully, there won't be a `close` listener that is going to restart it a second time if there is a different way of restarting it.

## Checklist


charisk (approved):
LGTM

PR #2889, opened 2023-09-29
Apply language context to queries panel

starcke:
This adds language filtering to the queries panel. Adding the filtering was trivial, however there were a few other tweaks I had to make.

1. Previously we made `onDidChangeQueries` the same as `onDidChangePathData`. However if we integrated with that it would require a full rescan of the file system to apply a filter which felt quite slow for me. Instead we create a specific emitter for `onDidChangeQueries` which is triggered by either the filesystem changing or the filter changing.
2. Initially my tests failed when trying to dispose `this.languageContext.onLanguageContextChanged`. This was because the MockApp created a MockEventEmitter which returned `{}` as disposable. However `{}` is actually not disposable (because it lacks the method). So I changed it to return something that is actually disposable.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -790,7 +790,7 @@ async function activateWithInstalledDistribution(
   );
   ctx.subscriptions.push(databaseUI);
 
-  QueriesModule.initialize(app, cliServer);
+  QueriesModule.initialize(app, languageContext, cliServer);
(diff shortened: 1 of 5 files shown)


shati-patel (approved):
This looks good to me! Works locally, and your explanations for `onDidChangeQueries` and the `disposable` make sense.

charisk (approved):

PR #2891, opened 2023-09-29
Dispose tracked objects when panel is disposed

koesie10:
This will change the `AbstractWebview` to dispose its tracked objects (using `this.push`) when the panel is disposed rather than when the view is disposed. This makes `this.push` actually useful in a view. Before, the objects would only get disposed when the extension itself was disposed.

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #2895, opened 2023-10-02
Extract base functionality for WebviewViewProviders into an abstract class

charisk:
This is to help with re-usability but also to make the code more readable since plumbing code etc. get moved out of the concrete class.

This also solves the issue of disposal of event subscriptions at the right time (see https://github.com/github/vscode-codeql/pull/2891). 

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
+++ b/extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
@@ -64,6 +64,8 @@ export abstract class AbstractWebviewViewProvider<
       const disposable = this.disposables.pop()!;
       disposable.dispose();
     }
+
+    this.webviewView = undefined;
   }
 
   protected push<T extends Disposable>(obj: T): T {
@@ -75,6 +77,10 @@ export abstract class AbstractWebviewViewProvider<
 
   protected abstract onMessage(msg: FromMessage): Promise<void>;
 
+  /**
+   * This is called when a view first becomes visible. This may happen when the view is
+   * first loaded or when the user hides and then shows a view again.
+   */
   protected onWebViewLoaded(): void {
     // Do nothing by default.
   }
--- a/extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
+++ b/extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
@@ -47,8 +47,6 @@ export abstract class AbstractWebviewViewProvider<
 
     webviewView.webview.onDidReceiveMessage(async (msg) => this.onMessage(msg));
     webviewView.onDidDispose(() => this.dispose());
-
-    this.onWebViewLoaded();
   }
 
   protected get isShowingView() {
--- a/extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
+++ b/extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
@@ -1,0 +1,81 @@
+import * as vscode from "vscode";
+import { Uri, WebviewViewProvider } from "vscode";
+import { WebviewKind, WebviewMessage, getHtmlForWebview } from "./webview-html";
+import { Disposable } from "../disposable-object";
+import { App } from "../app";
+
+export abstract class AbstractWebviewViewProvider<
+  ToMessage extends WebviewMessage,
+  FromMessage extends WebviewMessage,
+> implements WebviewViewProvider
+{
+  protected webviewView: vscode.WebviewView | undefined = undefined;
+  private disposables: Disposable[] = [];
+
+  constructor(
+    private readonly app: App,
+    private readonly webviewKind: WebviewKind,
+  ) {}
+
+  /**
+   * This is called when a view first becomes visible. This may happen when the view is
+   * first loaded or when the user hides and then shows a view again.
+   */
+  public resolveWebviewView(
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-panel.ts
```diff
@@ -12,7 +12,7 @@
     super();
 
     this.provider = new MethodModelingViewProvider(app, modelingStore);
-    this.push(this.provider);
+    this.provider;
```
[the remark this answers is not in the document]
charisk: Fair 😂

extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+        allowWasmEval: false,
+      },
+    );
+
+    webviewView.webview.html = html;
+
+    this.webviewView = webviewView;
+
+    webviewView.webview.onDidReceiveMessage(async (msg) => this.onMessage(msg));
+    webviewView.onDidDispose(() => this.dispose());
```
koesie10, line 49: Should we also set `this.webviewView = undefined` here?

extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+  ToMessage extends WebviewMessage,
+  FromMessage extends WebviewMessage,
+> implements WebviewViewProvider
+{
+  protected webviewView: vscode.WebviewView | undefined = undefined;
+  private disposables: Disposable[] = [];
+
+  constructor(
+    private readonly app: App,
+    private readonly webviewKind: WebviewKind,
```
koesie10, line 17: Is there a specific reason for choosing to do this differently than the `AbstractWebview`? In there, this gets returned by the `getPanelConfig` method, which also includes other (optional) options. Right now we probably don't need any of those for webview views, but it would be nice to keep it consistent with the `AbstractWebview` and keep the options open (even though it's quite easy to change afterwards).

extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+  protected push<T extends Disposable>(obj: T): T {
+    if (obj !== undefined) {
+      this.disposables.push(obj);
+    }
+    return obj;
+  }
+
+  protected abstract onMessage(msg: FromMessage): Promise<void>;
+
+  protected onWebViewLoaded(): void {
```
koesie10, line 78: Should we clarify in the method description that may be called multiple times? I think it'll only be called once in the `AbstractWebview` because `retainContextWhenHidden` is set, so this may be surprising.

extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
charisk: It just seemed like unnecessary complexity for now but I'm open to changing it.

extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
charisk: Added a comment in the method doc, following the same as `resolveWebviewView`

extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
koesie10: I'm fine with keeping it as-is, but we should reconsider if we ever start exposing other options.

koesie10 (approved):

extensions/ql-vscode/src/common/vscode/abstract-webview-view-provider.ts
```diff
@@ -0,0 +1,87 @@
...
+    return obj;
+  }
+
+  protected abstract onMessage(msg: FromMessage): Promise<void>;
+
+  /**
+   * This is called when a view first becomes visible. This may happen when the view is
+   * first loaded or when the user hides and then shows a view again.
+   */
+  protected onWebViewLoaded(): void {
```
koesie10, line 84: Additional question which I just encountered: This is called when the `viewLoaded` message is received in the `AbstractWebview`, but this is called after the panel is created in this class, without waiting for the webview to actually be loaded. That means you can't do `postMessage` in `onWebViewLoaded` because the React code to receive the messages may not have initialized yet.

There should probably be a requirement to handle the `viewLoaded` message in `onMessage` and call this method. If it's not handled, you will also get `ExhaustivityCheckingError` errors if you use `default: assertNever(msg)`.

koesie10: That also works, but I think we should still have the `onWebViewLoaded` and the requirement to handle `viewLoaded` as well. The view code (in `src/view/webview.tsx`) is shared by the `AbstractWebview` and this class, so we would still be receiving the `viewLoaded` message. If we don't handle the message, it may lead to surprising outcomes if you do `default: assertNever(msg)` because you would still be receiving the `viewLoaded` message.

charisk: Good point! Updated.

PR #2899, opened 2023-10-03
Re-open method modeling panel if closed

charisk:
If a user manually closes the method modeling panel, there is no good way for them to find it again. They'll need to know to trigger the "Focus on CodeQL Method Modeling View" command from the command pallet. With this PR, the panel gets automatically re-opened. More details in the internal linked issue.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-panel.ts
+++ b/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-panel.ts
@@ -35,10 +35,6 @@ export class MethodModelingPanel extends DisposableObject {
   }
 
   public async show(): Promise<void> {
-    if (this.provider.isWebviewViewResolved) {
-      await this.provider.show();
-    } else {
-      await this.app.commands.execute("codeQLMethodModeling.focus");
-    }
+    await this.app.commands.execute("codeQLMethodModeling.focus");
   }
 }
--- a/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-panel.ts
+++ b/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-panel.ts
@@ -10,7 +10,7 @@ export class MethodModelingPanel extends DisposableObject {
   private readonly provider: MethodModelingViewProvider;
 
   constructor(
-    app: App,
+    private readonly app: App,
     modelingStore: ModelingStore,
     editorViewTracker: ModelEditorViewTracker,
   ) {
@@ -25,11 +25,20 @@ export class MethodModelingPanel extends DisposableObject {
       window.registerWebviewViewProvider(
         MethodModelingViewProvider.viewType,
         this.provider,
+        { webviewOptions: { retainContextWhenHidden: true } },
       ),
     );
   }
 
   public async setMethod(method: Method): Promise<void> {
     await this.provider.setMethod(method);
+  }
+
+  public async show(): Promise<void> {
+    if (this.provider.isWebviewViewResolved) {
+      await this.provider.show();
+    } else {
+      await this.app.commands.execute("codeQLMethodModeling.focus");
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-panel.ts
```diff
@@ -25,11 +25,20 @@ export class MethodModelingPanel extends DisposableObject {
...
     );
   }
 
   public async setMethod(method: Method): Promise<void> {
     await this.provider.setMethod(method);
   }
+
+  public async show(): Promise<void> {
+    if (this.provider.isWebviewViewResolved) {
+      await this.provider.show();
```
robertbrignull, line 39: What benefit does calling `WebviewView.show` give over always running the `codeQLMethodModeling.focus` command? It seems they have the same effect so could we make our code simpler by always using the command? Would there be any downsides?

charisk: The only difference is that `WebviewView.show` will not change the focus, but I'm not 100% sure we need that. Perhaps I'll change to `codeQLMethodModeling.focus` and see if we think that's not behaving well once we get to try it out a bit more.

charisk: Done!

robertbrignull: > The only difference is that `WebviewView.show` will not change the focus,

According the the docs, that's only if you pass `true` to it, and we weren't doing that.

But I agree we should pay attention to what we're doing with the focus. We don't want to change the focus unnecessarily if it isn't what the user wants.

charisk: > According the the docs, that's only if you pass true to it, and we weren't doing that.

I think we were? https://github.com/github/vscode-codeql/pull/2899/commits/205134e7481e417974d7ad06aa951d8cb01b2128#diff-61ad437a68ce69419aa5888ad04241912a30629ee7b1f220fcedf8ea5accd16eR61

extensions/ql-vscode/src/model-editor/model-editor-module.ts
```diff
@@ -188,6 +188,7 @@ export class ModelEditorModule extends DisposableObject {
         databaseItem: DatabaseItem,
       ) => {
         this.modelingStore.setSelectedMethod(databaseItem, method, usage);
+        await this.methodModelingPanel.show();
```
robertbrignull, line 191: This is the `codeQLModelEditor.jumpToUsageLocation` command which is called when clicking on items in the method usage panel. It doesn't trigger when clicking the "view" button in the main model editor. I assume we also want it to run then?

charisk: I'm not sure we do tbh. I wanted to get opinions from EPD. This part hasn't exactly been spec'ed out. See internal issue for details.

extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-panel.ts
robertbrignull: Oh sorry. I got confused because there's two methods called `show`

robertbrignull: I've tested it out a bit and something a bit odd is going on when the CodeQL Method Modeling panel is closed (as in, I right-clicked it and selected "hide"). When I click something in the usages panel it shows the modeling panel but doesn't jump to location.

Everything works as expected if the modeling panel is open or collapsed. It's only when it's hidden that it doesn't jump to location correctly.

> We don't want to change the focus unnecessarily if it isn't what the user wants.

I'm sorry, the focus is also a bit unhelpful because it focuses the modeling panel. This means I can't navigate the code which is what I'd expect. What do you think? Is it ok or is the behaviour too unexpected?

charisk: That's a good find! It seems like stealing the focus might be causing the odd behaviour around jumping to location.

I'm starting to think this may be a bad idea so I think I'll part it for now until we've had a better chance to discuss flows.

charisk:
Abandoning this idea for now. See https://github.com/github/vscode-codeql/pull/2899#discussion_r1344211152.

PR #2904, opened 2023-10-04
Convert `yaml.ts` to handle multiple models per method

koesie10:
This changes YAML parsing/creating functions for the model editor to handle multiple models per method. The changes in the actual YAML handling are fairly small because the format itself already supports multiple models per method.

I've introduced a few helper functions to convert between the old and new types. This should only be necessary while we're in the middle of the transition to the new types and can be removed later. For now, we'll just take the first model in the array when converting from the new to the old type. This is a change in the behavior since currently we always take the last model in the array but this behavior is undocumented and unsupported, so it should be fine to change it.

## Checklist


robertbrignull (approved):
LGTM

So the behaviour when creating YAML files is now that all modelings for a method are treated as a unit. So if a method already has modelings `[A, B]` but then we try to write the new modeling `[C]`, it'll remove A and B and just write C, instead of writing `[A, B, C]`. I think that's correct and the tests do cover this.

PR #2905, opened 2023-10-04
Convert `modeled-method-fs.ts` to handle multiple models per method

koesie10:
This will change the input/output types for modeled methods in the `modeled-method-fs.ts` file to take in multiple models per method. This removes the need for conversion functions between this file and `yaml.ts` files. Instead, the conversion functions are done when calling any functions defined in `modeled-method-fs.ts` files.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/modeled-method-fs.ts
+++ b/extensions/ql-vscode/src/model-editor/modeled-method-fs.ts
@@ -89,7 +89,7 @@ export async function loadModeledMethods(
   for (const modeledMethods of Object.values(modeledMethodsByFile)) {
     for (const [key, value] of Object.entries(modeledMethods)) {
       if (!(key in existingModeledMethods)) {
-        existingModeledMethods[key] = value;
+        existingModeledMethods[key] = [];
       }
 
       existingModeledMethods[key].push(...value);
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/model-editor/modeled-method-fs.ts
```diff
@@ -94,7 +88,11 @@ export async function loadModeledMethods(
   );
   for (const modeledMethods of Object.values(modeledMethodsByFile)) {
     for (const [key, value] of Object.entries(modeledMethods)) {
-      existingModeledMethods[key] = value;
+      if (!(key in existingModeledMethods)) {
+        existingModeledMethods[key] = value;
+      }
+
+      existingModeledMethods[key].push(...value);
```
robertbrignull, line 95: This looks like it's going to duplicate the modeled methods.

Not sure if it was intended to be:
```suggestion
      if (!(key in existingModeledMethods)) {
        existingModeledMethods[key] = value;
      } else {
        existingModeledMethods[key].push(...value);
      }
```

or:
```suggestion
      if (!(key in existingModeledMethods)) {
        existingModeledMethods[key] = [];
      }

      existingModeledMethods[key].push(...value);
```

robertbrignull (commented):
One comment but otherwise looks good to me.

This is going exactly the way I intended, by pushing the `convertToLegacyModeledMethods` higher and higher up the call stack, until eventually everything will be converted and we can get rid of those conversion methods. I'm glad it seems to be working out ok.

robertbrignull (approved):

PR #2906, opened 2023-10-04
Convert remaining extension host code to handle multiple models

koesie10:
This converts all remaining extension host code to handle multiple models per method. The only place where we're using the legacy format is in the webview and in the boundary between the webview and the extension host.

## Checklist


extensions/ql-vscode/src/model-editor/modeled-methods-legacy.ts
```diff
@@ -1,23 +1,71 @@
...
+/**
+ * Converts a ModeledMethod[] to a single ModeledMethod for legacy usage. This function should always be used instead
+ * of the trivial conversion to track usages of this conversion.
+ *
+ * This method should only be called inside a `postMessage` call. If it's used anywhere else, consider whether the
+ * boundary is correct: the boundary should as close as possible to the extension host -> webview boundary.
+ *
+ * @param modeledMethods The ModeledMethod[]
+ */
+export function convertToLegacyModeledMethod(modeledMethods: ModeledMethod[]) {
```
robertbrignull, line 69:
```suggestion
export function convertToLegacyModeledMethod(modeledMethods: ModeledMethod[]): ModeledMethod | undefined {
```

This could benefit from an explicit return type because typescript considers array access to always be defined, but this isn't really true.

robertbrignull (commented):
Has some conflicts too that'll need fixing. I can do another review once the PR is updated.

robertbrignull (approved):

PR #2913, opened 2023-10-05
Remove method and modeledMethods from SaveModeledMethods message

robertbrignull:
Removes the `methods` and `modeledMethods` fields from `SaveModeledMethods` in favour of including `methodSignatures`. The extension host can use the `ModelingStore` to look up the methods and modelings using the method signatures.

Involves adding `getMethods` and `getModeledMethods` to `ModelingStore`.

Tested locally and seems to be saving models and updating the "unsaved" labels correctly.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/model-editor-view.ts
+++ b/extensions/ql-vscode/src/model-editor/model-editor-view.ts
@@ -228,47 +228,58 @@ export class ModelEditorView extends AbstractWebview<
 
         break;
       case "saveModeledMethods":
-        await withProgress(
-          async (progress) => {
-            progress({
-              step: 1,
-              maxStep: 500 + externalApiQueriesProgressMaxStep,
-              message: "Writing model files",
-            });
-            await saveModeledMethods(
-              this.extensionPack,
-              this.databaseItem.language,
-              msg.methods,
-              msg.modeledMethods,
-              this.mode,
-              this.cliServer,
-              this.app.logger,
-            );
+        {
+          const methods = this.modelingStore.getMethods(
+            this.databaseItem,
+            msg.methodSignatures,
+          );
+          const modeledMethods = this.modelingStore.getModeledMethods(
+            this.databaseItem,
+            msg.methodSignatures,
+          );
 
-            await Promise.all([
-              this.setViewState(),
-              this.loadMethods((update) =>
-                progress({
-                  ...update,
-                  step: update.step + 500,
-                  maxStep: 500 + externalApiQueriesProgressMaxStep,
-                }),
-              ),
-            ]);
-          },
-          {
-            cancellable: false,
-          },
-        );
+          await withProgress(
+            async (progress) => {
+              progress({
+                step: 1,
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/model-editor/model-editor-view.ts
```diff
@@ -228,47 +228,58 @@ export class ModelEditorView extends AbstractWebview<
...
+              cancellable: false,
+            },
+          );
 
-        void telemetryListener?.sendUIInteraction(
-          "model-editor-save-modeled-methods",
-        );
+          this.modelingStore.removeModifiedMethods(
+            this.databaseItem,
+            Object.keys(modeledMethods),
```
charisk, line 276: Are these really just the method signatures?

charisk (approved):
Looks good!

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -154,6 +154,23 @@ export class ModelingStore extends DisposableObject {
     return this.state.get(this.activeDb);
   }
 
+  /**
+   * Returns the methods for the given database item and method signatures.
+   * If no method signatures are provided, returns all methods.
```
koesie10, line 159: This should probably be more specific. "no method signatures" can also be interpreted as giving an empty array, but this will actually return no methods because an empty array is truthy. I think that's fine, but we should just say that either the `methodSignatures` argument is not provided or the `methodSignatures` argument is `undefined`.

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -182,6 +199,25 @@ export class ModelingStore extends DisposableObject {
     });
   }
 
+  /**
+   * Returns the modeled methods for the given database item and method signatures.
+   * If no method signatures are provided, returns all modeled methods.
```
koesie10, line 204: Same as above

extensions/ql-vscode/src/model-editor/model-editor-view.ts
robertbrignull: I can't remember why I left this as `Object.keys` instead of using `msg.methodSignatures` 🤔 

The only difference would be if we tried to save signatures for which we couldn't find any modeled methods. Maybe it wouldn't matter either way because they wouldn't be displayed in the UI if there aren't any modeled methods. But we should probably err on the side of marking them as not-modified.

robertbrignull: Ah, now I remember. It's because of the "undefined means save everything" behaviour. So `msg.methodSignatures` might be undefined. We could use `methods.map(m => m.signature)` but I think that theoretically it shouldn't matter, so I'll leave it as it is. 🤞🏼

extensions/ql-vscode/src/model-editor/modeling-store.ts
robertbrignull: Good point. I've clarified this comment.

PR #2918, opened 2023-10-06
Add view state to method modeling panel

koesie10:
This adds a view state to the method modeling panel similar to the model editor. This will be used to send the state of the show multiple models feature flag to the webview so this can be used to selectively show/hide components in the method modeling panel.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/method-modeling/MethodModeling.tsx
+++ b/extensions/ql-vscode/src/view/method-modeling/MethodModeling.tsx
@@ -40,6 +40,7 @@ export type MethodModelingProps = {
   modelingStatus: ModelingStatus;
   method: Method;
   modeledMethod: ModeledMethod | undefined;
+  showMultipleModels?: boolean;
   onChange: (modeledMethod: ModeledMethod) => void;
 };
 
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/view/method-modeling/MethodModeling.tsx
```diff
@@ -40,6 +40,7 @@ export type MethodModelingProps = {
   modelingStatus: ModelingStatus;
   method: Method;
   modeledMethod: ModeledMethod | undefined;
+  showMultipleModels?: boolean;
```
koesie10, line 43: I needed to have some way to use the `viewState` in the `MethodModelingView`, so this is just to make use of the `viewState` for now. In a follow-up PR, this will be used.

charisk (approved):
LGTM and happy to get it merged. 

However, it's hard to know what is a "view state" and what isn't so if we're planning to add more things to this we should perhaps have a discussion/make things clearer.

koesie10:
> However, it's hard to know what is a "view state" and what isn't so if we're planning to add more things to this we should perhaps have a discussion/make things clearer.

I agree. For now, we're only planning on adding this, I don't think there's anything else we would be adding here soon.

PR #2921, opened 2023-10-06
Add ability for method modeling panel to render multiple modelings of the same method

koesie10:
This makes it possible for the method modeling panel to render multiple models for the same method. It does not yet add support for updating (i.e. `onChange` is unchanged) or adding/removing new methods. The appearance and behavior when the feature flag is disabled should be unchanged.

The best way to view this locally is using Storybook; there are 3 different scenarios in "Method Modeling/Method Modeling".

Screenshot 2023-10-06 at 16 52 03

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/common/icon/CodiconButton.tsx
+++ b/extensions/ql-vscode/src/view/common/icon/CodiconButton.tsx
@@ -1,0 +1,50 @@
+import * as React from "react";
+import { styled } from "styled-components";
+import classNames from "classnames";
+
+type Size = "x-small" | "small" | "medium" | "large" | "x-large";
+
+const StyledButton = styled.button<{ size: Size }>`
+  background: none;
+  color: var(--vscode-textLink-foreground);
+  border: none;
+  cursor: pointer;
+  font-size: ${(props) => props.size ?? "1em"};
+  padding: 0;
+  vertical-align: text-bottom;
+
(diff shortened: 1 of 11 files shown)


extensions/ql-vscode/src/view/common/icon/CodiconButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  padding: 0;
+  vertical-align: text-bottom;
+
+  &:disabled {
+    color: var(--vscode-disabledForeground);
+    cursor: default;
+  }
+`;
+
+export const CodiconButton = ({
```
robertbrignull, line 19: Is `<VSCodeButton appearance="icon"><Codicon .../></VSCodeButton>` not able to do what we want here? What are the challenges that require a new button component?

I'm not against adding a new component, but I just want to verify that it is needed.

robertbrignull (approved):
LGTM

extensions/ql-vscode/src/view/common/icon/CodiconButton.tsx
koesie10: Thanks, I didn't know that this prop existed. I'll switch to use this instead. Unfortunately it makes testing a bit harder (because they're not native components), but it does mean we'll have less custom components to maintain.

PR #2922, opened 2023-10-06
Add `additionalArgs` option to `launch.json`

dbartol:
This PR adds a new `additionalArgs` property to the QL debug configuration specified in `launch.json`. The property accepts a JSON object and passes all of the properties of that JSON object through to the `runQuery` method of the query server, without validation or interpretation. This makes it easier to prototype new query server features without having to make updates to the VS Code extension. The new property is not declared in `package.json`, to avoid encouraging accidental use.

**Bikeshed Request**: What's a better name for this property? Since it isn't in `package.json`, it won't show up in autocomplete for `launch.json`, so it should be something that internal users can remember.

_Implementation notes:_
Most of the code changes are just plumbing the new property all the way through to the query server.


extensions/ql-vscode/test/vscode-tests/cli-integration/debugger/debugger.test.ts
```diff
@@ -177,4 +177,27 @@ describeWithCodeQL()("Debugger", () => {
...
   });
+
+  it("should pass additionalArgs through to query server", async () => {
+    await withDebugController(appCommands, async (controller) => {
+      await controller.startDebugging(
+        {
+          query: quickEvalQueryPath,
+          additionalArgs: {
+            // Overrides the value passed to the query server
+            queryPath: simpleQueryPath,
```
aeisenberg, line 188: Oh...interesting. This will allow you to override any args set by the extension. Could be dangerous, but then again it's an internal, undocumented feature.

aeisenberg (commented):
Maybe this should be gated by a check on canary mode. I don't feel too strongly about this since it's internal and undocumented anyway.

In terms of naming, something that evokes that it is internal and for running queries might be nicer. Eg- `additionalRunQueryArgsInternal`, or even `internalrunQueryArgs`.

extensions/ql-vscode/test/vscode-tests/cli-integration/debugger/debugger.test.ts
dbartol: Yeah. There are a few sharp edges.

jbj:
> **Bikeshed Request**: What's a better name for this property?

I like `additionalRunQueryArgs`, which is what you've named it internally. That name describes precisely what it does, which is appropriate for such a low-level feature.

> Since it isn't in `package.json`, it won't show up in autocomplete for `launch.json`, so it should be something that internal users can remember.

I'd like to see the property in autocomplete. If we expect more than a handful of internal users to ever use this property, it should be as discoverable as possible. It seems to me that it's a small price to pay if a handful of external users discover it too and find it confusing. This property should only confuse someone who (1) ventures into `launch.json` and (2) doesn't heed the warning that's built into the name and description of the property. We can write a stern warning message in the description, clarifying that this property is passed directly to the `RunQuery` command of the query server without validation.

aeisenberg (approved):

PR #2923, opened 2023-10-09
Update CHANGELOG.md for telemetry changes

robertbrignull:
Update the changelog for https://github.com/github/vscode-codeql/pull/2824, because I forgot to do it in the PR.

Links to the docs which will be updated by https://github.com/github/codeql/pull/14408

## Checklist


shati-patel (approved):
Looks good! 
Worth noting that the docs normally get updated once every 2 weeks, with a CLI release, so we probably won't be able to coordinate this exactly. Do we want to leave out this changelog entry until after the docs have definitely been released? (I don't mind either way, it's a small enough change 😄)

robertbrignull:
> Worth noting that the docs normally get updated once every 2 weeks, with a CLI release, so we probably won't be able to coordinate this exactly. Do we want to leave out this changelog entry until after the docs have definitely been released?

I don't mind too much. I also don't think it's too bad if the docs take an extra week to be updated.

Technically the docs are still correct since we also respect the old setting, but I realise that might be confusing when reading the changelog.

I think I'd err on the side of merging it so that the changelog matches the code change.

PR #2924, opened 2023-10-09
Add ability to add/remove modelings to method modeling panel

koesie10:
This adds the ability to add/remove modelings on the method modeling panel if the feature flag is enabled. This adds a plus/trash icon which are enabled based on the already modeled methods:
- If there is more than 1 modeled method, the remove button is enabled
- If all modeled methods are modeled (i.e. not unmodeled, `type !== 'none'`), then the add button is enabled

I've also added a new story for testing this behaviour since the boundary is at the current story, which means that you weren't able to validate that the new `onChange` behavior works. Now, you can fully experiment with the new story since it also keeps track of state.

Screenshot 2023-10-09 at 15 18 47

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/method-modeling/MultipleModeledMethodsPanel.tsx
+++ b/extensions/ql-vscode/src/view/method-modeling/MultipleModeledMethodsPanel.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { useCallback, useState } from "react";
+import { useCallback, useMemo, useState } from "react";
 import { Method } from "../../model-editor/method";
 import { ModeledMethod } from "../../model-editor/modeled-method";
 import { styled } from "styled-components";
@@ -10,7 +10,7 @@ import { Codicon } from "../common";
 export type MultipleModeledMethodsPanelProps = {
   method: Method;
   modeledMethods: ModeledMethod[];
-  onChange: (modeledMethod: ModeledMethod) => void;
+  onChange: (modeledMethods: ModeledMethod[]) => void;
 };
 
 const Container = styled.div`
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/view/method-modeling/MultipleModeledMethodsPanel.tsx
```diff
@@ -47,19 +54,72 @@ export const MultipleModeledMethodsPanel = ({
...
+      methodName: method.methodName,
+      methodParameters: method.methodParameters,
+    };
+
+    const newModeledMethods = [...modeledMethods, newModeledMethod];
+
+    onChange(newModeledMethods);
+    setSelectedIndex(newModeledMethods.length - 1);
+  }, [onChange, modeledMethods, method]);
+  const handleRemoveClick = useCallback(() => {
```
robertbrignull, line 76:
```suggestion

  const handleRemoveClick = useCallback(() => {
```
Just suggesting adding a newline so it's easier to read.

robertbrignull (approved):
Works fine for me in the story added in this PR. The code also looks basically fine.

Thanks for adding so many tests. I must remember to do that once I get to adding the working buttons into the model editor.

robertbrignull:
> If all modeled methods are modeled (i.e. not unmodeled, `type !== 'none'`), then the add button is enabled

I'm having extra thoughts about this. Do we want this behaviour, or do we want to ignore unmodeled methods and allow adding new modelings?

I worry it could be confusing that the "add" button is disabled because there happens to be an unmodeled method on another page. The unmodeled method won't be visible and there's no indication as to why the button is disabled.

When saving, I believe that any rows with `type === 'none'` will be ignored and discard, so it'll clean itself up when the user tries to save the data.

koesie10:
> I'm having extra thoughts about this. Do we want this behaviour, or do we want to ignore unmodeled methods and allow adding new modelings?

I think it would also make sense to always allow adding new models. I can create a PR to change this behavior, but the reason I implemented it like this is because the first design shows the add button disabled when there is only 1 unmodeled method. I wasn't sure exactly whether this is because there is exactly 1 unmodeled method, or because there is an unmodeled method. Should we change the behavior to disable the add button when you have exactly 1 model and it is unmodeled?

robertbrignull:
> Should we change the behavior to disable the add button when you have exactly 1 model and it is unmodeled?

This seems understandable to me. I'd be happy to go with this behaviour.

PR #2929, opened 2023-10-09
Ensure modified methods updated after changing a method from the modeling panel

charisk:
Small change to make sure we update the modified methods state when the modeled method gets updated. If we don't do that we can end up in a situation where the state/UI is out of date. For example the following bug exists in the current code:
1. Open model editor
2. Click view on some method
3. Change to source (see it being marked as modeled)
4. Click on another method from the method usages panel 
5. Change to flow summary

At this point the second method should be marked as modeled, but it isn't.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
+++ b/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
@@ -100,6 +100,10 @@ export class MethodModelingViewProvider extends AbstractWebviewViewProvider<
           activeState.databaseItem,
           msg.method,
         );
+        this.modelingStore.addModifiedMethod(
+          activeState.databaseItem,
+          msg.method.signature,
+        );
         break;
       }
       case "revealInModelEditor":


extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
```diff
@@ -100,6 +100,10 @@ export class MethodModelingViewProvider extends AbstractWebviewViewProvider<
           activeState.databaseItem,
           msg.method,
         );
+        this.modelingStore.addModifiedMethod(
```
starcke, line 103: It seems we often have to call `addModifiedMethod(s)` after methods like `updateModeledMethod` and `addModeledMethods`. I wonder if we can move the call to update the modified methods inside the store so that the caller does not have to remember to call both functions?

starcke (approved):
Looks good, happy to go with this as a bugfix - but I wonder if we can streamline the modeled and modified method calls?

extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
charisk: Yes I'd definitely like to move towards that, it's a good idea. In general I can see a few simplifications/improvements we could make around the modeling store. But I think of this as v1, with a lot of features currently in flux, so I'd rather wait a bit before tidying things up. Basically let it evolve a bit and then come back and address some tech debt once we have a fuller picture of its use.

PR #2931, opened 2023-10-09
Add "CodeQL: Trim Cache" using new mode parameter of `evaluation/clearCache`

d10c:
The purpose of this change is to add a command to clear the cache
except for `cached` predicates. This is the
same behaviour as `codeql database cleanup --mode=normal`.
In contrast, the existing "CodeQL: Clear Cache" corresponds to `--mode=brutal`.

This change depends on adding a new `mode` parameter to the
`evaluation/clearCache` method of the query server. This is the same
method that the existing "CodeQL: Clear Cache" command targets, which means
that this solution will not fail when used with older query servers
but will have the same behaviour as the existing command.
This somewhat decouples the release process of the VSCode extension and
the CodeQL release.

Testing done so far: manual (VSCode extension debug run). I haven't seen
automated tests for the other command either.

As part of adding the new "mode" parameter to the clearCache method, the
official query server spec will need to be updated to match `new-messages.ts`
here.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/databases/local-databases-ui.ts
@@ -251,7 +251,8 @@ export class DatabaseUI extends DisposableObject {
         this.handleSetDefaultTourDatabase.bind(this),
       "codeQL.upgradeCurrentDatabase":
         this.handleUpgradeCurrentDatabase.bind(this),
-      "codeQL.clearCache": this.handleClearCache.bind(this),
+      "codeQL.clearCache": this.handleClearCache.bind(this, "clear"),
+      "codeQL.trimCache": this.handleClearCache.bind(this, "trim"),
       "codeQLDatabases.chooseDatabaseFolder":
         this.handleChooseDatabaseFolder.bind(this),
       "codeQLDatabases.chooseDatabaseArchive":
@@ -684,7 +685,7 @@ export class DatabaseUI extends DisposableObject {
     );
   }
 
-  private async handleClearCache(): Promise<void> {
+  private async handleClearCache(mode: "trim" | "clear"): Promise<void> {
     return withProgress(
       async (_progress, token) => {
         if (
@@ -694,6 +695,7 @@ export class DatabaseUI extends DisposableObject {
           await this.queryServer.clearCacheInDatabase(
             this.databaseManager.currentDatabaseItem,
             token,
+            mode,
           );
         }
       },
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -684,7 +685,7 @@ export class DatabaseUI extends DisposableObject {
     );
   }
 
-  private async handleClearCache(): Promise<void> {
+  private async handleClearCache(mode: "trim" | "clear"): Promise<void> {
```
aeisenberg, line 688: It would be nice to extract `"trim" | "clear"` into a `type` so it can be reused below.

extensions/ql-vscode/src/query-server/new-messages.ts
```diff
@@ -29,6 +29,35 @@ export interface ClearCacheParams {
...
+export namespace CacheTrimmingMode {
+  /** The entire cache is deleted unconditionally. */
+  export const BRUTAL = 0;
+  /** Only `cached` predicates are kept. */
+  export const NORMAL = 1;
+  /**
+   * Trim the cache down to the configured size, but *within* this limit keep anything that
+   * appears to be even possibly potentially useful in the future.
+   */
+  export const LIGHT = 2;
```
aeisenberg, line 53: Are there plans to support `LIGHT` and `GENTLE` in the future?

extensions/ql-vscode/src/query-server/new-query-runner.ts
```diff
@@ -57,15 +58,20 @@ export class NewQueryRunner extends QueryRunner {
...
     }
 
     const db = dbItem.databaseUri.fsPath;
     const params: ClearCacheParams = {
       dryRun: false,
       db,
+      mode: {
+        trim: CacheTrimmingMode.NORMAL,
+        clear: CacheTrimmingMode.BRUTAL,
+      }[mode],
```
aeisenberg, line 74: Minor: could you extract this to a function? I find this hard to read.

extensions/ql-vscode/src/query-server/new-messages.ts
d10c: Probably not, as they're pretty obscure and there's little practical reason to support them in VSCode. Also, we're renaming "brutal" → "clear" and "normal" → "trim" to be more in line with the VSCode commands, so if you have suggestions on new names for "light" and "gentle", I'm all ears :)

d10c:
Closing this PR in favour of the approach taken in https://github.com/github/vscode-codeql/pull/2928

PR #2936, opened 2023-10-10
Add model validation in method modeling panel

koesie10:
This adds validation of the modeled methods to the method modeling panel. It implements the following two validations:
1. A neutral classification cannot be combined with any other non-neutral classification
2. A classification cannot be duplicated

Multiple neutral classifications will be validated by the second validation rather than the first one.

Screenshot 2023-10-10 at 14 32 01

Screenshot 2023-10-10 at 14 32 12

Testing is only really possible in Storybook right now, without any interactivity.

## Checklist


extensions/ql-vscode/src/model-editor/shared/validation.ts
```diff
@@ -0,0 +1,141 @@
...
+  const seenModeledMethods = new Set<string>();
+  for (const modeledMethod of consideredModeledMethods) {
+    const canonicalModeledMethod = canonicalizeModeledMethod(modeledMethod);
+    const key = JSON.stringify(canonicalModeledMethod);
+
+    if (seenModeledMethods.has(key)) {
+      result.push({
+        title: "Duplicated classification",
+        message:
+          "This method has two identical or conflicting classifications.",
```
robertbrignull, line 107: In the extreme case of more than two duplicate modelings, do we want to avoid adding multiple errors? I'm unsure. I'd say it's a very unlikely case, so perhaps not worth putting in the implementation effort to give a nice user experience in this case.

robertbrignull (commented):
Looks reasonable, with a couple of questions.

We'll likely end up displaying this info in the model editor and method modeling panels. Do we want to perform the validation once in each location, or do it in a shared location and pass the errors into each view?

I think what you have here where we do the validation in the view is likely fine. It will mean we do the validation twice, but the validation should be fast enough and consistent, so doing it twice is safe. At least this way the validation errors will definitely match the data being displayed.

extensions/ql-vscode/src/view/common/Alert.tsx
```diff
@@ -85,11 +85,28 @@ type Props = {
 
   // Inverse the color scheme
   inverse?: boolean;
+
+  /**
+   * Role is used as the ARIA role. "alert" should only be set if the alert requires
+   * the user's immediate attention. "status" should be set if the alert is not
+   * important enough to require the user's immediate attention.
+   *
+   * Can be left out if the alert is not important enough to require the user's
```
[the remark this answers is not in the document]
koesie10: I'll extend these docs a bit to clarify that. Essentially, when no `role` is set, the alert will be treated as normal text by screen readers and the user will not be notified of any changes.

extensions/ql-vscode/src/model-editor/shared/validation.ts
koesie10: I'm not sure either, but I don't think it's worth handling this case for a very rare case.

koesie10:
> We'll likely end up displaying this info in the model editor and method modeling panels. Do we want to perform the validation once in each location, or do it in a shared location and pass the errors into each view?
> 
> I think what you have here where we do the validation in the view is likely fine. It will mean we do the validation twice, but the validation should be fast enough and consistent, so doing it twice is safe. At least this way the validation errors will definitely match the data being displayed.

I think this is a lot easier to manage. If we were to store the validation errors in the modeling store, we would need to run the validation for every method, even if it's not shown in the model editor (for example if the group it's in is collapsed). The validation should always result in the same result, so keeping it in the view shouldn't really make a big difference. If we start experiencing performance problems because of this, we can always consider moving it to the extension host where the user will probably notice it less.

robertbrignull (approved):
👍🏼

PR #2942, opened 2023-10-10
Convert SetModeledMethodsMessage to include multiple modeled methods

robertbrignull:
Converts `SetModeledMethodsMessage` to include multiple modeled methods and pushes that change throughout the model editor. This means it's now possible to view multiple methods, though it's not yet possible to edit them and save the changes.

This can be tested by manually editing a model file to give multiple modelings for the same method.

Screenshot 2023-10-10 at 15 44 08

## Checklist


extensions/ql-vscode/src/view/model-editor/LibraryRow.tsx
```diff
@@ -71,7 +71,7 @@ export type LibraryRowProps = {
   title: string;
   libraryVersion?: string;
   methods: Method[];
-  modeledMethods: Record<string, ModeledMethod>;
+  modeledMethodsMap: Record<string, ModeledMethod[]>;
```
koesie10, line 74: Is there a specific reason for using `modeledMethodsMap` here, while we're using `modeledMethods` everywhere else?

koesie10 (approved):

extensions/ql-vscode/src/view/model-editor/LibraryRow.tsx
robertbrignull: My intention was to change all instances of things with the type `Record<string, ModeledMethod[]>` in the model editor webview to `modeledMethodsMap`, so they're all consistent. That way we aren't calling the same data `modeledMethods` in one component where we call it `modeledMethodsMap` and `modeledMethods` has a different meaning.

Do you think we should just call it `modeledMethods` unless we also happen to need objects of type `ModeledMethod[]` in the same scope?

I didn't spot that my automated renaming had renamed the fields in the type but not when they're used in the component. So we have `modeledMethodsMap: modeledMethods` when we get the value from the props. I'll fix that because I think that's definitely unhelpful, regardless of the decision above.

koesie10: That makes sense, I think the `modeledMethodsMap: modeledMethods` was just confusing to me.

PR #2943, opened 2023-10-10
Fix confusion between modeling store and view states

koesie10:
This fixes three bugs related to the modeling store and view states:
- In the model editor view, when `setModeledMethods` was called, it would do it on the active database, instead of the database that the view was showing. This should not result in any visible bugs since the active database is always the one that is being shown (in theory), but I can imagine that it could cause issues if showing multiple model editors next to each other.
- In the method modeling panel, the "reveal in editor" button would always show the already active model editor. Therefore, if you had multiple open and were still viewing the method of the first one, it would always show the second one.
- In the method modeling panel, the same bug would cause the incorrect modeled methods to be updated.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/modeling-store.ts
+++ b/extensions/ql-vscode/src/model-editor/modeling-store.ts
@@ -347,6 +347,7 @@ export class ModelingStore extends DisposableObject {
     }
 
     return {
+      databaseItem: dbState.databaseItem,
       method: selectedMethod,
       usage: dbState.selectedUsage,
       modeledMethods: dbState.modeledMethods[selectedMethod.signature] ?? [],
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -347,6 +347,7 @@ export class ModelingStore extends DisposableObject {
     }
 
     return {
+      databaseItem: dbState.databaseItem,
```
charisk, line 350: I don't think we're using this but I also don't mind us adding it.

charisk (approved):
LGTM, thanks for tidying this up.

extensions/ql-vscode/src/model-editor/modeling-store.ts
koesie10: Thanks for noticing this, I was intending to use it. Right now, when the webview is loaded it could result in an invalid state because the webview has a different state than the view provider.

PR #2945, opened 2023-10-10
Rename 'open database' button and only show if source available

charisk:
Rename 'open database' button of the model editor to 'open source' and hide it if the source is not available as a workspace folder.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/model-editor-view.ts
+++ b/extensions/ql-vscode/src/model-editor/model-editor-view.ts
@@ -5,6 +5,7 @@ import {
   Uri,
   ViewColumn,
   window,
+  workspace,
 } from "vscode";
 import {
   AbstractWebview,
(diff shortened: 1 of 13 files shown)


extensions/ql-vscode/src/model-editor/model-editor-view.ts
```diff
@@ -353,6 +354,11 @@ export class ModelEditorView extends AbstractWebview<
     const showLlmButton =
       this.databaseItem.language === "java" && this.modelConfig.llmGeneration;
 
+    const sourceArchiveAvailable = workspace.workspaceFolders?.some(
+      (f) =>
+        f.uri.fsPath === this.databaseItem.getSourceArchiveExplorerUri().fsPath,
```
koesie10, line 359: This call [will fail](https://github.com/github/vscode-codeql/blob/e77cf28192d3c23b4cd6d58cf765c268d71bf043/extensions/ql-vscode/src/databases/local-databases/database-item-impl.ts#L164-L166) for test databases, so is there another way we can do this? Otherwise, we should make it possible for `getSourceArchiveExplorerUri` to return `undefined` and update all of its call sites.

It seems like [the database manager uses this](https://github.com/github/vscode-codeql/blob/c93dd7a25b01517ff1f434db779d47c5158dfc29/extensions/ql-vscode/src/databases/local-databases/database-manager.ts#L471-L473) for detecting whether the source is already added to the workspace:

```ts
    return (vscode.workspace.workspaceFolders || []).findIndex((folder) =>
      item.belongsToSourceArchiveExplorerUri(folder.uri),
    );
```

charisk: Good catch. I've updated to use `belongsToSourceArchiveExplorerUri`.

PR #2947, opened 2023-10-11
Convert `SetSelectedMethodMessage` to include multiple modeled methods

koesie10:
This converts the `SetSelectedMethodMessage` to include multiple methods. It also adds a new message to replicate the `SetModeledMethodMessage` with multiple modeled methods so that all state is now pushed through correctly from the modeled method view provider. Unfortunately, the name `SetModeledMethodsMessage` is already taken by the modeled methods map, so I've created a new message `SetMultipleModeledMethodsMessage`. We may want to consider changing these once all messages are properly renamed.

The changes to `canMethodBeModeled` are directly copied from https://github.com/github/vscode-codeql/pull/2942/files#diff-fb5d586d735e8d18fcae2a4225ba24aff4416e9423ea1a4fb1af7bc5d5805da8

## Checklist


extensions/ql-vscode/src/common/interface-types.ts
```diff
@@ -577,6 +577,11 @@ interface SetModeledMethodMessage {
   method: ModeledMethod;
 }
 
+interface SetMultipleModeledMethodsMessage {
```
robertbrignull, line 580: 👍🏼 

Adding a new type makes sense to me. Since we're reusing `SetModeledMethodMessage` in a few places, we'd have to change them all at once.

Once we're fully done with the migration, we can rename this type if we want to.

robertbrignull (approved):
LGTM

extensions/ql-vscode/src/common/interface-types.ts
```diff
@@ -577,6 +577,11 @@ interface SetModeledMethodMessage {
...
+  t: "setMultipleModeledMethods";
+  modeledMethods: ModeledMethod[];
```
robertbrignull, line 582: Did we also discuss including `methodSignature: string;` in this message, so you don't have to work out the signature by looking at the modeled methods?

That would also make it easier to remove all modeled methods for a method signature, instead of having to set it to the array containing one element with `type: "none"`.

I'm happy for you to make this change in this PR or I can do it in a followup, because I'll be using this message in my next PR.

koesie10: It does make sense to add it here. Right now, this message is only used for communication from the extension host to the view, but we'll definitely need the signature for the communication from the view to the extension host. I'll change the message now so that I can also use it in a follow-up PR (after #2924).

robertbrignull (approved):
Thanks for updating and adding `methodSignatures`.

Has conflicts, but otherwise LGTM

PR #2955, opened 2023-10-12
Hash result set name in sorted result set path

kaspersv:
This PR changes the filename used for sorted result sets from using the result set name directly to using a hash. Using a hash is preferable as the result set names are generated from the named results of the final stage. It is therefore partially under user-control and may contain additional special characters that cannot be used in filenames as the RA name character set is extended.  

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/run-queries-shared.ts
+++ b/extensions/ql-vscode/src/run-queries-shared.ts
@@ -24,6 +24,7 @@ import { BaseLogger, showAndLogWarningMessage } from "./common/logging";
 import { extLogger } from "./common/logging/vscode";
 import { generateSummarySymbolsFile } from "./log-insights/summary-parser";
 import { getErrorMessage } from "./common/helpers-pure";
+import { createHash } from "crypto";
 
 /**
  * run-queries.ts
@@ -150,7 +151,12 @@ export class QueryEvaluationInfo extends QueryOutputDir {
     };
   }
   getSortedResultSetPath(resultSetName: string) {
-    return join(this.querySaveDir, `sortedResults-${resultSetName}.bqrs`);
+    const hasher = createHash("sha256");
+    hasher.update(resultSetName);
+    return join(
+      this.querySaveDir,
+      `sortedResults-${hasher.digest("hex")}.bqrs`,
+    );
   }
 
   /**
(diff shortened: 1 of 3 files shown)


kaspersv:
Build failures should be resolved by https://github.com/github/vscode-codeql/pull/2962.

extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -150,7 +151,9 @@ export class QueryEvaluationInfo extends QueryOutputDir {
     };
   }
   getSortedResultSetPath(resultSetName: string) {
-    return join(this.querySaveDir, `sortedResults-${resultSetName}.bqrs`);
+    const hasher = new MurmurHash3.default();
```
koesie10, line 154: Is there a specific reason for choosing to use MurmurHash here? I would prefer to use a hashing algorithm for which we do not need a dependency, so something like SHA256 using [`crypto.createHash`](https://nodejs.org/docs/latest-v18.x/api/crypto.html#cryptocreatehashalgorithm-options) would be just as good IMO. I don't think the performance of this method matters, so using a more expensive hashing algorithm shouldn't really be a problem.

extensions/ql-vscode/package.json
```diff
@@ -2081,6 +2081,7 @@
   "dependencies": {
     "@octokit/plugin-retry": "^6.0.1",
     "@octokit/rest": "^20.0.2",
+    "@types/imurmurhash": "^0.1.2",
```
koesie10, line 2084: If we use MurmurHash, can we also declare a dependency on `imurmurhash`? Right now, the only reason we have access to it because it's included in one of our dev dependencies (`eslint`). It would be unexpected that if eslint removes their use of the package, our code would suddenly break.

koesie10 (commented):
Thanks for implementing this! I have some questions about MurmurHash, but the general implementation looks good.

extensions/ql-vscode/src/run-queries-shared.ts
kaspersv: I saw that `imurmurhash` was already a dependency, but didn't notice it was only a dev dependency. I agree the performance difference shouldn't matter in this case and will update the PR to use SHA256 using `crypto.createHash` instead.

koesie10 (approved):
Thank you for implementing those changes

PR #2964, opened 2023-10-12
Enable/disable the add/remove model buttons at the right times

robertbrignull:
Hooks up the `disabled` attribute for the add/remove model buttons in the model editor.

Uses the same logic as the method modeling panel, and pulls out the implementation to a shared location. Although it turns out the model editor doesn't need the logic for disabling the remove button, because it's never shown in situations when it would be disabled.

Also adds some tests that the buttons are shown / enabled at the right times.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/model-editor/__tests__/MethodRow.spec.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/__tests__/MethodRow.spec.tsx
@@ -251,4 +251,111 @@ describe(MethodRow.name, () => {
     expect(screen.queryByRole("combobox")).not.toBeInTheDocument();
     expect(screen.getByText("Method already modeled")).toBeInTheDocument();
   });
+
+  it("Doesn't show add/remove buttons when multiple methods feature flag is disabled", async () => {
+    render({
+      modeledMethods: [modeledMethod],
+      viewState: {
+        ...viewState,
+        showMultipleModels: false,
+      },
+    });
+
+    expect(screen.queryByLabelText("Add new model")).not.toBeInTheDocument();
+    expect(screen.queryByLabelText("Remove model")).not.toBeInTheDocument();
+  });
+
+  it("Button to add new model is disabled when there are no modeled methods", async () => {
+    render({
+      modeledMethods: [],
+      viewState: {
+        ...viewState,
+        showMultipleModels: true,
+      },
+    });
+
+    const addButton = screen.queryByLabelText("Add new model");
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/view/model-editor/__tests__/MethodRow.spec.tsx
```diff
@@ -251,4 +251,111 @@ describe(MethodRow.name, () => {
     expect(screen.queryByRole("combobox")).not.toBeInTheDocument();
     expect(screen.getByText("Method already modeled")).toBeInTheDocument();
   });
+
+  it("Doesn't show add/remove buttons when multiple methods feature flag is disabled", async () => {
```
koesie10, line 255: Most of our other tests don't start with capitals, so could we also remove those here? I think the intention is that the `it` starts the sentence, so that you'd have "it doesn't show add/remove buttons when multiple methods feature flag is disabled".

extensions/ql-vscode/src/view/model-editor/__tests__/MethodRow.spec.tsx
```diff
@@ -251,4 +251,111 @@ describe(MethodRow.name, () => {
...
+        ...viewState,
+        showMultipleModels: false,
+      },
+    });
+
+    expect(screen.queryByLabelText("Add new model")).not.toBeInTheDocument();
+    expect(screen.queryByLabelText("Remove model")).not.toBeInTheDocument();
+  });
+
+  it("Button to add new model is disabled when there are no modeled methods", async () => {
```
koesie10, line 268: Same here, this should read like a sentence starting with "it"

extensions/ql-vscode/src/view/model-editor/__tests__/MethodRow.spec.tsx
```diff
@@ -251,4 +251,111 @@ describe(MethodRow.name, () => {
...
+    });
+
+    const addButton = screen.queryByLabelText("Add new model");
+    expect(addButton).toBeInTheDocument();
+    expect(addButton).toBeDisabled();
+
+    expect(screen.queryByLabelText("Remove model")).not.toBeInTheDocument();
+  });
+
+  it("Add button is shown and enabed when there is one modeled methods", async () => {
```
koesie10, line 300:
```suggestion
  it("Add button is shown and enabled when there is one modeled method", async () => {
```

koesie10 (approved):

PR #2965, opened 2023-10-12
Fix neutral model validation to consider kind

koesie10:
This fixes a bug where the validation of modeled methods would not consider the kind of the modeled method, and would therefore give an error when there was e.g. a neutral sink and a non-neutral summary.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/shared/validation.ts
+++ b/extensions/ql-vscode/src/model-editor/shared/validation.ts
@@ -75,7 +75,7 @@ function canonicalizeModeledMethod(
         type: "neutral",
         input: "",
         output: "",
-        kind: "",
+        kind: modeledMethod.kind,
         provenance: "manual",
       };
     default:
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/model-editor/shared/validation.ts
```diff
@@ -117,24 +117,40 @@ export function validateModeledMethods(
...
+      .get(neutralModeledMethod.kind)
+      ?.push(neutralModeledMethod);
+  }
+
+  for (const [
+    neutralModeledMethodKind,
+    neutralModeledMethods,
+  ] of neutralModeledMethodsByKind) {
+    const conflictingMethods = consideredModeledMethods.filter(
+      (method) => neutralModeledMethodKind === method.type,
```
robertbrignull, line 140: Comparing `kind` and `type` here doesn't look right. We're also not filtering out the neutral method itself so the `conflictingMethods.length < 1` check below would likely always detect the neutral method.

Should this be
```suggestion
      (method) => method.type !== "neutral" && neutralModeledMethodKind === method.kind,
```

koesie10: I think this is correct. We're trying to match up the neutral's kind with the non-neutral's type. For example, we want to find these two conflicting classifications:

```json5
// Neutral classification
{
  "type": "neutral",
  "kind": "sink",
  // ...
}

// Sink classification
{
  "type": "sink",
  "kind": "code-injection",
  // ...
}
```

Therefore, the condition for a single conflict is something like `neutral.type === 'neutral' && nonNeutral.kind === neutral.type`.

We're explicitly not looking at the non-neutral method's kind since that's `code-injection`, `remote`, or `taint`, etc.

robertbrignull: I see. Thanks for explaining. I hadn't really understood / had forgotten how this worked. I now agree that this makes sense and is the right thing to do.

robertbrignull (approved):
Thanks for updating / explaining. I now agree this implementation looks correct, and the tests also cover all cases.

PR #2968, opened 2023-10-13
Improve immutability of modeling store state

koesie10:
This improves the immutability of the modeling store state by using TypeScript's readonly types to ensure that state can only be modified from within the modeling store or when it's copied. This mostly consists of adding `readonly` to properties and arrays, but this also adds a `DeepReadonly` type to use in `postMessage` arguments to ensure that readonly objects can be passed in. `postMessage` will never modify the objects, so this is safe.

Alternative approaches are also possible such as copying the state when retrieving it, but this should guard against the most common cases where we could be modifying state unintentionally.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/readonly.ts
+++ b/extensions/ql-vscode/src/common/readonly.ts
@@ -1,0 +1,14 @@
+export type DeepReadonly<T> = T extends Array<infer R>
+  ? DeepReadonlyArray<R>
+  : // eslint-disable-next-line @typescript-eslint/ban-types
+  T extends Function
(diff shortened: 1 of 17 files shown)


charisk (approved):
Thanks for working on this! Hopefully it'll help us avoid some bugs.

Some of the signatures become a bit long and repetitive as they're used in various places (`Readonly<Record<string, readonly ModeledMethod[]>>` doesn't quite roll of the tongue/keyboard) so I think there are opportunities to improve readability a bit, but we can address that separately.

PR #2976, opened 2023-10-16
Introduce separate tree item types in the methods usage panel

koesie10:
This creates new tree item types for methods and usages such that these can contain references to their parent and children. This allows us to easily find the parent of a usage and to find the children of a method. This removes an expensive `find` call in `getParent`.

## Checklist


robertbrignull (approved):
Works for me when tested, and the code looks good. I think this is a good approach and very similar to what I tried to implement previously.

I also think renaming `resolveCanonicalUsage` to `resolveUsageTreeViewItem` is a good change. Unfortunately we can't remove this method unless the internal tree view data/type is the exact same as the data outside of the tree view, and that won't be the case now. We still do an expensive lookup in this method but I think this is an ok tradeoff and we call this a lot less frequently than we call `getParent`. It's also something we can look at improving again in the future if it's an issue.

PR #2926, opened 2023-10-09
Only include method signature in JumpToUsageMessage

robertbrignull:
Updates the `JumpToUsageMessage` to not include the full `Method` and `Usage` objects, but instead to identify the method by its signature and let the modeling store provide the full content.

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #2977, opened 2023-10-16
Revert changes to codeQLModelEditor.jumpToMethod to allow jumping to usages other than the first usage

robertbrignull:
In https://github.com/github/vscode-codeql/pull/2926 I was trying to simplify the code as much as possible, but I accidentally changed the `codeQLModelEditor.jumpToMethod` command which does need to be able to jump to usages other than the first usage.

This PR reverts the changes to that command and makes it work again. This does require adding a new `getMethod` method to the modeling store.

I did consider another way of implementing this which was to add a `usageIndex: number` field to the command. That also worked, but I decided it was safer to just revert the changes from the earlier PR and stick more closely to the original state of the code.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/commands.ts
+++ b/extensions/ql-vscode/src/common/commands.ts
@@ -13,7 +13,7 @@ import type {
 import type { QLDebugConfiguration } from "../debugger/debug-configuration";
 import type { QueryTreeViewItem } from "../queries-panel/query-tree-view-item";
 import type { LanguageSelectionTreeViewItem } from "../language-selection-panel/language-selection-data-provider";
-import { Method, Usage } from "../model-editor/method";
+import type { Method, Usage } from "../model-editor/method";
 
 // A command function matching the signature that VS Code calls when
 // a command is invoked from a context menu on a TreeView with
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/common/commands.ts
```diff
@@ -13,6 +13,7 @@ import type {
 import type { QLDebugConfiguration } from "../debugger/debug-configuration";
 import type { QueryTreeViewItem } from "../queries-panel/query-tree-view-item";
 import type { LanguageSelectionTreeViewItem } from "../language-selection-panel/language-selection-data-provider";
+import { Method, Usage } from "../model-editor/method";
```
koesie10, line 16: Can we make these type imports?

```suggestion
import type { Method, Usage } from "../model-editor/method";
```

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -388,18 +401,9 @@ export class ModelingStore extends DisposableObject {
     });
   }
 
-  public setSelectedMethod(dbItem: DatabaseItem, methodSignature: string) {
+  public setSelectedMethod(dbItem: DatabaseItem, method: Method, usage: Usage) {
```
koesie10, line 404: Should we add a comment to this method to specify that the `method` and `usage` need to have been retrieved from the modeling store? Theoretically you could pass in a usage received from the webview which might cause other code relying on object referential identity to break.

koesie10 (approved):

PR #2983, opened 2023-10-16
Handle when the number of modeled methods decreases

robertbrignull:
In the method modeling panel, you may be viewing model 3/3 but then one of the models is deleted. In this case we should now be viewing model 2/2 instead of model 3/2. The "remove model" button in the method modeling panel already handles this, but the list of modeled methods can also be changed by outside sources.

To test this you may need to merge in https://github.com/github/vscode-codeql/pull/2982 so the number of modeled methods can be affected by an external source.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/method-modeling/MultipleModeledMethodsPanel.tsx
+++ b/extensions/ql-vscode/src/view/method-modeling/MultipleModeledMethodsPanel.tsx
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { useCallback, useEffect, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { Method } from "../../model-editor/method";
 import { ModeledMethod } from "../../model-editor/modeled-method";
 import {
@@ -60,7 +60,15 @@ export const MultipleModeledMethodsPanel = ({
 }: MultipleModeledMethodsPanelProps) => {
   const [selectedIndex, setSelectedIndex] = useState<number>(0);
 
+  const selectNewMethod = useRef<number | null>(null);
+
   useEffect(() => {
+    if (selectNewMethod.current === modeledMethods.length - 1) {
+      setSelectedIndex(selectNewMethod.current);
+      selectNewMethod.current = null;
+      return;
+    }
+
     if (selectedIndex >= modeledMethods.length) {
       setSelectedIndex(
         modeledMethods.length > 0 ? modeledMethods.length - 1 : 0,
@@ -97,7 +105,7 @@ export const MultipleModeledMethodsPanel = ({
     const newModeledMethods = [...modeledMethods, newModeledMethod];
 
     onChange(method.signature, newModeledMethods);
-    setSelectedIndex(newModeledMethods.length - 1);
+    selectNewMethod.current = newModeledMethods.length - 1;
   }, [onChange, modeledMethods, method]);
 
   const handleRemoveClick = useCallback(() => {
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/view/method-modeling/MultipleModeledMethodsPanel.tsx
```diff
@@ -60,6 +60,14 @@ export const MultipleModeledMethodsPanel = ({
 }: MultipleModeledMethodsPanelProps) => {
   const [selectedIndex, setSelectedIndex] = useState<number>(0);
 
+  useEffect(() => {
```
robertbrignull, line 63: Hmm, this change actually breaks adding a new model, because in `handleAddClick` we try to set `selectedIndex` but it fails because the number of models hasn't been updated yet. The list of models needs to get sent to the extension host, be updated in the modeling store, and then get sent back.

I'm not really sure how to best solve this. I can think of a couple of non-ideal solutions:
- One option is the `MultipleModeledMethodsPanel` stores the list of modeled methods as state as well as props. Then it can update its internal state right away. I don't think this is a great solution as it introduces extra complexity and places for things to go wrong.
- Another options is the `MultipleModeledMethodsPanel` could have some state to remember that it just tried to add a model, and then when we see the  set of modeled methods change we update `selectedIndex` then. It's a bit clunky though.

robertbrignull: If you want to test this, the user visible effect is that when you click "add a new model" you stay on model 1, so you're now viewing model 1/2. The intended behaviour is that you would be viewing model 2/2, i.e. the new unmodeled method you just added.

robertbrignull:
Thanks for the suggested code change. I've made that change and tested it and it's working correctly and the tests are passing.

koesie10 (commented):
Could you please add a test for this new behavior when adding a modeling? If I comment out all code related to the `selectNewMethod` ref, none of the tests fail. I think the easiest would be to add something like this to an existing "can add modeling" test:

```ts
      rerender(
        <MultipleModeledMethodsPanel
          method={method}
          modeledMethods={
            onChange.mock.calls[onChange.mock.calls.length - 1][1]
          }
          isModelingInProgress={isModelingInProgress}
          onChange={onChange}
        />,
      );

      expect(screen.getByText("2/2")).toBeInTheDocument();
```

This should make it fail when we remove this behavior.

robertbrignull:
> Could you please add a test for this new behavior when adding a modeling? If I comment out all code related to the `selectNewMethod` ref, none of the tests fail.

Sure, I can add an specialised test for this.

I'm surprised none of the tests failed when you commented out the code. They did fail on https://github.com/github/vscode-codeql/pull/2983/commits/96c0feb3e6be9a5bfc6818c31af9cfdda5a1ae6f which was how I realised things were broken.

koesie10:
> I'm surprised none of the tests failed when you commented out the code. They did fail on [96c0feb](https://github.com/github/vscode-codeql/commit/96c0feb3e6be9a5bfc6818c31af9cfdda5a1ae6f) which was how I realised things were broken.

I just tried it again and commented out all code related to `selectNewMethod` (including that condition in the `useEffect` hook), and I don't get any failing tests in `MultipleModeledMethodsPanel.spec.tsx`. That should be equal to the code that was in https://github.com/github/vscode-codeql/commit/96c0feb3e6be9a5bfc6818c31af9cfdda5a1ae6f, so I'm not sure why I'm not getting any failing tests.

robertbrignull:
I've added some tests that the pagination selected is updated correctly when:
- models are added from within the method modeling panel
- models are removed from outside of the method modeling panel

I think this covers the behaviour changes being made in this PR.

koesie10 (approved):
Thanks!

PR #2989, opened 2023-10-17
Keep reference to original items in usage data provider

koesie10:
This changes the usage data provider tree items to keep a reference to the method and usage instead of only including their properties in the tree item. This makes it easier to find the original method and usage when revealing an item in the tree. It also removes the `getParent` call in `getTreeItem`.

The main reason for this fix is to ensure `codeQLModelEditor.jumpToMethod` gets the correct `usage` argument. It received the tree item before, but now we can actually pass the usage that was clicked on.

## Checklist


robertbrignull (approved):
👍🏼

PR #2990, opened 2023-10-17
Use custom grid element instead of VSCodeDataGrid

robertbrignull:
This PR removes the usage of `VSCodeDataGrid` from the model editor and instead applies the `display: grid` property manually.

The issue with `VSCodeDataGrid` is that it's actually each `VSCodeDataGridRow` that has the `display: grid` property. This means each row is an individual grid and getting columns to align across rows is tricky and is impossible is certain situations.

I took all the relevant styling directly from the `VSCodeDataGrid` family of components. If there's extra styling I missed then please say. For the "focused" condition I had to change it from a border to a background colour, because elements with `display: contents` cannot have borders.

One significant change is that we now have to give elements a `gridRow` property as well as `gridColumn`. This does add some extra complexity, but it also opens up the option to have elements that span multiple rows. I plan to use this feature in a followup PR to remove `MultiModelColumn`. My aim is that each dropdown input should be its own grid cell, instead of having all of the "kind" dropdowns into one cell in a flex column.

This screenshot shows the state before this PR (Note how the columns in the header don't match the table body):
Screenshot 2023-10-17 at 15 56 17

This screenshot shows the same view after this PR:
Screenshot 2023-10-17 at 15 52 33

And this screenshot shows how it looks when a row is "focussed" using the "reveal in editor" button (not the same as hovering, which looks identical to how it did before):
Screenshot 2023-10-17 at 16 23 01

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/common/DataGrid.tsx
+++ b/extensions/ql-vscode/src/view/common/DataGrid.tsx
@@ -1,0 +1,79 @@
+import * as React from "react";
+import { styled } from "styled-components";
+
+/*
+ * A drop-in replacement for the VSCodeDataGrid family of components.
+ *
+ * The difference is that the `display: grid` styling is applied to `DataGrid`, whereas
+ * in the VS Code version that styling is applied to `VSCodeDataGridRow`. This gives
+ * column alignment across rows in situation with dynamic contents. It also allows
+ * for cells to span multiple rows and all the other features of data grids.
+ */
+
+const StyledDataGrid = styled.div<{ $gridTemplateColumns: string | number }>`
+  display: grid;
+  grid-template-columns: ${(props) => props.$gridTemplateColumns};
+  padding: calc((var(--design-unit) / 4) * 1px) 0;
+  box-sizing: border-box;
+  width: 100%;
+  background: transparent;
+`;
+
+interface DataGridProps {
+  gridTemplateColumns: string | number;
+  children: React.ReactNode;
+}
+
+export function DataGrid(props: DataGridProps) {
+  const { gridTemplateColumns, children } = props;
+
+  return (
+    <StyledDataGrid
+      aria-label="DataGrid"
+      $gridTemplateColumns={gridTemplateColumns}
+    >
+      {children}
+    </StyledDataGrid>
+  );
+}
+
+export const DataGridRow = styled.div<{ $focused?: boolean }>`
+  display: contents;
+
+  &:hover > * {
+    background-color: var(--list-hover-background);
+  }
+
+  & > * {
+    // Use !important to override the background color set by the hover state
+    background-color: ${(props) =>
+      props.$focused
+        ? "var(--vscode-editor-selectionBackground) !important"
+        : "inherit"};
+  }
+`;
+
+const StyledDataGridCell = styled.div<{
+  $gridRow: string | number;
+  $gridColumn: string | number;
+}>`
+  grid-row: ${(props) => props.$gridRow};
+  grid-column: ${(props) => props.$gridColumn};
+  padding: calc(var(--design-unit) * 1px) calc(var(--design-unit) * 3px);
+`;
+
+interface DataGridCellProps {
+  gridRow: string | number;
+  gridColumn: string | number;
+  children: React.ReactNode;
+}
+
+export function DataGridCell(props: DataGridCellProps) {
+  const { gridRow, gridColumn, children } = props;
+
+  return (
+    <StyledDataGridCell $gridRow={gridRow} $gridColumn={gridColumn}>
+      {children}
+    </StyledDataGridCell>
+  );
+}
--- a/extensions/ql-vscode/src/view/model-editor/HiddenMethodsRow.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/HiddenMethodsRow.tsx
@@ -1,35 +1,41 @@
-import {
-  VSCodeDataGridCell,
-  VSCodeDataGridRow,
-} from "@vscode/webview-ui-toolkit/react";
 import * as React from "react";
 import { styled } from "styled-components";
 import { pluralize } from "../../common/word";
+import { DataGridCell, DataGridRow } from "../common/DataGrid";
+import { ModelEditorViewState } from "../../model-editor/shared/view-state";
 
-const HiddenMethodsCell = styled(VSCodeDataGridCell)`
+const HiddenMethodsText = styled.div`
   text-align: center;
 `;
 
 interface Props {
+  gridRow: number;
   numHiddenMethods: number;
   someMethodsAreVisible: boolean;
+  viewState: ModelEditorViewState;
 }
 
 export function HiddenMethodsRow({
+  gridRow,
   numHiddenMethods,
   someMethodsAreVisible,
+  viewState,
 }: Props) {
   if (numHiddenMethods === 0) {
     return null;
   }
 
+  const gridColumn = viewState.showMultipleModels ? "span 6" : "span 5";
+
   return (
-    <VSCodeDataGridRow>
-      <HiddenMethodsCell gridColumn="span 5">
-        {someMethodsAreVisible && "And "}
-        {pluralize(numHiddenMethods, "method", "methods")} modeled in other
-        CodeQL packs
-      </HiddenMethodsCell>
-    </VSCodeDataGridRow>
+    <DataGridRow>
+      <DataGridCell gridRow={gridRow} gridColumn={gridColumn}>
+        <HiddenMethodsText>
+          {someMethodsAreVisible && "And "}
+          {pluralize(numHiddenMethods, "method", "methods")} modeled in other
+          CodeQL packs
+        </HiddenMethodsText>
+      </DataGridCell>
+    </DataGridRow>
   );
 }
--- a/extensions/ql-vscode/src/view/model-editor/MethodName.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/MethodName.tsx
@@ -4,6 +4,7 @@ import { Method } from "../../model-editor/method";
 
 const Name = styled.span`
   font-family: var(--vscode-editor-font-family);
+  word-break: break-all;
 `;
 
 export const MethodName = (method: Method): JSX.Element => {
--- a/extensions/ql-vscode/src/view/model-editor/ModeledMethodDataGrid.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/ModeledMethodDataGrid.tsx
@@ -1,9 +1,4 @@
 import * as React from "react";
-import {
-  VSCodeDataGrid,
-  VSCodeDataGridCell,
-  VSCodeDataGridRow,
-} from "@vscode/webview-ui-toolkit/react";
 import { MethodRow } from "./MethodRow";
 import { Method, canMethodBeModeled } from "../../model-editor/method";
 import { ModeledMethod } from "../../model-editor/modeled-method";
@@ -12,6 +7,7 @@ import { sortMethods } from "../../model-editor/shared/sorting";
 import { HiddenMethodsRow } from "./HiddenMethodsRow";
 import { ModelEditorViewState } from "../../model-editor/shared/view-state";
 import { ScreenReaderOnly } from "../common/ScreenReaderOnly";
+import { DataGrid, DataGridCell } from "../common/DataGrid";
 
 export const SINGLE_MODEL_GRID_TEMPLATE_COLUMNS =
   "0.5fr 0.125fr 0.125fr 0.125fr 0.125fr";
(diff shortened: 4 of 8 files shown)


extensions/ql-vscode/src/view/model-editor/MethodName.tsx
```diff
@@ -4,6 +4,7 @@ import { Method } from "../../model-editor/method";
 
 const Name = styled.span`
   font-family: var(--vscode-editor-font-family);
+  word-break: break-all;
```
robertbrignull, line 7: I had to add this otherwise it was breaking the grid at low screen widths. I'm not sure why this is needed now and wasn't before.

extensions/ql-vscode/src/view/model-editor/HiddenMethodsRow.tsx
```diff
@@ -1,35 +1,41 @@
...
-  VSCodeDataGridCell,
-  VSCodeDataGridRow,
-} from "@vscode/webview-ui-toolkit/react";
 import * as React from "react";
 import { styled } from "styled-components";
 import { pluralize } from "../../common/word";
+import { DataGridCell, DataGridRow } from "../common/DataGrid";
+import { ModelEditorViewState } from "../../model-editor/shared/view-state";
 
-const HiddenMethodsCell = styled(VSCodeDataGridCell)`
```
robertbrignull, line 9: In a few cases I had to convert styled cells into two nested components. Styling on the cells themselves often seems to be ignored. Possibly this worked before because of shadow dom shenanigans.

robertbrignull:
It's true we could probably fix the header column alignment by adding some invisible fixed-width element into the final cell of the header. But there's so much more that doesn't really work properly when you have lots of little grids pretending to be one big grid.

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,79 @@
...
+  gridTemplateColumns: string | number;
+  children: React.ReactNode;
+}
+
+export function DataGrid(props: DataGridProps) {
+  const { gridTemplateColumns, children } = props;
+
+  return (
+    <StyledDataGrid
+      aria-label="DataGrid"
```
koesie10, line 32: I don't think this label is particularly helpful for screen readers. Could this either be a prop with component-specific text (like "CodeQL modeled methods for $libraryName") or be left out?

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,79 @@
...
+  background: transparent;
+`;
+
+interface DataGridProps {
+  gridTemplateColumns: string | number;
+  children: React.ReactNode;
+}
+
+export function DataGrid(props: DataGridProps) {
+  const { gridTemplateColumns, children } = props;
```
koesie10, line 28:
```suggestion
export function DataGrid({ gridTemplateColumns, children }: DataGridProps) {
```

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,79 @@
...
+ * The difference is that the `display: grid` styling is applied to `DataGrid`, whereas
+ * in the VS Code version that styling is applied to `VSCodeDataGridRow`. This gives
+ * column alignment across rows in situation with dynamic contents. It also allows
+ * for cells to span multiple rows and all the other features of data grids.
+ */
+
+const StyledDataGrid = styled.div<{ $gridTemplateColumns: string | number }>`
+  display: grid;
+  grid-template-columns: ${(props) => props.$gridTemplateColumns};
+  padding: calc((var(--design-unit) / 4) * 1px) 0;
```
koesie10, line 16: The `--design-unit` variable seems to come from FAST foundation, but since this is not a FAST foundation component I'm not sure we should be using it here. Could we simplify this to `1px`?

```suggestion
  padding: 1px 0;
```

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,79 @@
...
+    <StyledDataGrid
+      aria-label="DataGrid"
+      $gridTemplateColumns={gridTemplateColumns}
+    >
+      {children}
+    </StyledDataGrid>
+  );
+}
+
+export const DataGridRow = styled.div<{ $focused?: boolean }>`
```
koesie10, line 40: Is there a particular reason for choosing to prefix the prop name with `$`? Could we simply use `focused` instead?

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,79 @@
...
+    background-color: ${(props) =>
+      props.$focused
+        ? "var(--vscode-editor-selectionBackground) !important"
+        : "inherit"};
+  }
+`;
+
+const StyledDataGridCell = styled.div<{
+  $gridRow: string | number;
+  $gridColumn: string | number;
```
koesie10, line 58: Do we really need these properties? The default seems to be to follow the `grid-template-columns` in DOM definition order, so this should work without defining these manually.

If we want to define these manually later for multiple models per row, I think we should make these optional and only them for those rows (by using something like `span 2` as the value).

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,79 @@
...
+}>`
+  grid-row: ${(props) => props.$gridRow};
+  grid-column: ${(props) => props.$gridColumn};
+  padding: calc(var(--design-unit) * 1px) calc(var(--design-unit) * 3px);
```
koesie10, line 62:
```suggestion
  padding: 4px 12px;
```

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,79 @@
...
+`;
+
+interface DataGridCellProps {
+  gridRow: string | number;
+  gridColumn: string | number;
+  children: React.ReactNode;
+}
+
+export function DataGridCell(props: DataGridCellProps) {
+  const { gridRow, gridColumn, children } = props;
```
koesie10, line 72:
```suggestion
export function DataGridCell({ gridRow, gridColumn, children }: DataGridCellProps) {
```

extensions/ql-vscode/src/view/model-editor/HiddenMethodsRow.tsx
koesie10: Did you try adding a `className` prop to the `DataGridCell` component? `styled-components` works by passing down a separate `className` to the child component, so it should work if it accepts a `className` and passes it down to its rendered component.

extensions/ql-vscode/src/view/model-editor/ModeledMethodDataGrid.tsx
```diff
@@ -72,53 +68,56 @@ export const ModeledMethodDataGrid = ({
...
+            Output
+          </DataGridCell>
+          <DataGridCell gridRow={1} gridColumn={5}>
+            Kind
+          </DataGridCell>
+          {viewState.showMultipleModels && (
+            <DataGridCell gridRow={1} gridColumn={6}>
+              <ScreenReaderOnly>Add or remove models</ScreenReaderOnly>
+            </DataGridCell>
+          )}
```
koesie10, line 93: Can we add header styles to these like we had before? See https://github.com/microsoft/vscode-webview-ui-toolkit/blob/f22b1b0fab54bb41761388cb408be89335d9dc5c/src/data-grid/data-grid-cell.styles.ts#L41-L43

koesie10 (commented):
Thanks for doing this! It seems like the FAST foundation data grid has quite some accessibility features built in (like keyboard navigation), but I don't think we need that for this data grid. Most of the cells are already keyboard navigable through their inputs and focusing on a single row or cell is not necessary. The main thing we're probably missing out on is that it [was possible to navigate through rows by using these keys](https://github.com/microsoft/fast/blob/a7144c63191e4412d6441cce81cd04cc03152510/packages/web-components/fast-foundation/src/data-grid/data-grid.ts#L588-L674):
- Up arrow: Move 1 row up
- Down arrow: Move 1 row down
- Page up: Move 1 "page" up
- Page down: Move 1 "page" down
- Home: Move to the first row
- End: Move to the last row

I don't think it's particularly important to add those now, but it would be nice if we could add those back. We're not using most of the other features of the FAST foundation data grid, so I think this covers the rest of what we are using pretty well.

extensions/ql-vscode/src/view/model-editor/HiddenMethodsRow.tsx
robertbrignull: Thanks for pointing that out. I hadn't realised that's how that worked. That's kinda magic.

Passing the class name prop through works and now I can add another layer of styling onto `DataGridCell`, so we can keep things like `HiddenMethodsCell` the way they are.

extensions/ql-vscode/src/view/common/DataGrid.tsx
robertbrignull: I just copied this from `VSCodeDataGrid`. I agree it's likely not much use, as the screen reader can tell it's a grid already. So either it should contain extra info explaining the contents of the grid, or be left out. For now, I'll just remove it.

extensions/ql-vscode/src/view/common/DataGrid.tsx
koesie10: I see now that these are transient props. Could we ensure that we don't need to use transient props outside the `DataGrid.tsx` file, for example by creating a wrapper component? I think it's fine to use these for props that can only be used from within the file.

robertbrignull: Probably either works, but making it a transient prop could a cleaner. It probably doesn't matter in this case though, so I'm happy with whatever is cleaner.

Sadly, I'm finding that if you omit a prop on a styled component or add extra unknown ones it doesn't cause a build/linting error like it does for regular components. Perhas it would be safer to leave out the `$` prefix where unneeded, so the prop is named how people expect.

For others reading this, see https://github.com/github/vscode-codeql/pull/2992 for more info.

extensions/ql-vscode/src/view/common/DataGrid.tsx
robertbrignull: Good point. I'll remove these references and just use pixel values.

I've taken another look at the styling on the existing VS Code grid and copied the padding styling over more closely.

robertbrignull: Turns out the grid itself shouldn't really have any padding. I was accidentally copying this from the `VSCodeDataGridRow` component, but we can't really add padding to rows anymore because each cell is separate. I'm going to just skip this `1px` padding.

extensions/ql-vscode/src/view/common/DataGrid.tsx
robertbrignull: Yep, seems to work. These can definitely be optional properties, but we still have the option to supply them when needed, which could just be for `span 2` and such and not absolute values.

extensions/ql-vscode/src/view/model-editor/ModeledMethodDataGrid.tsx
robertbrignull: Thanks for pointing this out. I did briefly look for special header styles, but I couldn't find any. I'll add this in.

robertbrignull:
@koesie10, I think I've implemented everything from your review now. I've also done a bit more manual testing and as far as I can tell everything about the grid is working.

I ran into some extra problems with `DataGridRow` because it takes extra properties to be passed on to its children, such as `ref` and `data-testid`. I think I've solved everything but I'm not 100% happy with how understandable or maintainable it'll be going forward because you'll have to manually forward any future props as well. Let me know if you know of any better solutions.

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,111 @@
...
+`;
+
+interface DataGridRowProps {
+  focused?: boolean;
+  children: React.ReactNode;
+  "data-testid"?: string;
+}
+
+export const DataGridRow = forwardRef(
+  (props: DataGridRowProps, ref?: React.Ref<HTMLElement | undefined>) => {
```
koesie10, line 58: You should be able to do something like this here to avoid the `props["data-testid"]`:

```suggestion
  ({ focused, children, "data-testid": testId }: DataGridRowProps, ref?: React.Ref<HTMLElement | undefined>) => {
```

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }
+`;
+
+interface DataGridRowProps {
+  focused?: boolean;
+  children: React.ReactNode;
+  "data-testid"?: string;
+}
+
+export const DataGridRow = forwardRef(
```
koesie10, line 57: Would it make sense to also use `forwardRef` on the other components? It would be nice to be consistent on the usage of `forwardRef` within related components.

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,111 @@
...
+  display: grid;
+  grid-template-columns: ${(props) => props.$gridTemplateColumns};
+  box-sizing: border-box;
+  width: 100%;
+  background: transparent;
+`;
+
+interface DataGridProps {
+  gridTemplateColumns: string | number;
+  children: React.ReactNode;
```
koesie10, line 24:
```suggestion
  children: ReactNode;
```

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,111 @@
...
+    background-color: ${(props) =>
+      props.$focused
+        ? "var(--vscode-editor-selectionBackground) !important"
+        : "inherit"};
+  }
+`;
+
+interface DataGridRowProps {
+  focused?: boolean;
+  children: React.ReactNode;
```
koesie10, line 53:
```suggestion
  children: ReactNode;
```

extensions/ql-vscode/src/view/model-editor/HiddenMethodsRow.tsx
```diff
@@ -1,35 +1,37 @@
...
+const HiddenMethodsText = styled(DataGridCell)`
```
koesie10, line 7: Should these be renamed to `HiddenMethodsCell` again since we're now wrapping the cell again?

```suggestion
const HiddenMethodsCell = styled(DataGridCell)`
```

koesie10 (commented):
Thanks for implementing those changes. I have some additional comments, but the code looks really good already, these are mostly stylistic changes.

extensions/ql-vscode/src/view/common/DataGrid.tsx
```diff
@@ -0,0 +1,111 @@
...
+/*
+ * A drop-in replacement for the VSCodeDataGrid family of components.
+ *
+ * The difference is that the `display: grid` styling is applied to `DataGrid`, whereas
+ * in the VS Code version that styling is applied to `VSCodeDataGridRow`. This gives
+ * column alignment across rows in situation with dynamic contents. It also allows
+ * for cells to span multiple rows and all the other features of data grids.
+ */
+
+const StyledDataGrid = styled.div<{ $gridTemplateColumns: string | number }>`
```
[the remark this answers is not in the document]
robertbrignull: Oops, I agree this isn't useful for the `gridTemplateColumns` property.

On `gridColumn` it lets you write `gridColumn={5}` instead of `gridColumn={"5"}`, but I must have accidentally copied the type to here as well.

robertbrignull:
Thanks for another review. The PR is all updated again.

koesie10 (approved):

PR #2992, opened 2023-10-18
Add $ to columnCount prop used by styled component

robertbrignull:
This stops the following error from appearing during tests:
Screenshot 2023-10-18 at 11 50 08

I'm not totally sure if this has an effect on behaviour in production. Perhaps it was working but just with a warning.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/variant-analysis/RawResultsTable.tsx
+++ b/extensions/ql-vscode/src/view/variant-analysis/RawResultsTable.tsx
@@ -9,7 +9,7 @@ import { RawResultRow } from "./RawResultRow";
 const numOfResultsInContractedMode = 5;
 
 type TableContainerProps = {
-  columnCount: number;
+  $columnCount: number;
 };
 
 const TableContainer = styled.div<TableContainerProps>`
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/view/variant-analysis/RawResultsTable.tsx
```diff
@@ -9,7 +9,7 @@ import { RawResultRow } from "./RawResultRow";
 const numOfResultsInContractedMode = 5;
 
 type TableContainerProps = {
-  columnCount: number;
+  $columnCount: number;
```
robertbrignull, line 12: See https://styled-components.com/docs/basics#attaching-additional-props for docs on props for styled components.

koesie10 (approved):
Thanks, I didn't know about this feature. Apparently, these are [transient props](https://styled-components.com/docs/api#transient-props). I think it's fine to use these when we're only using the prop in the same file, but for components that are used outside of the same file, we should probably not use those. That way we can keep the implementation separate from the exposed interface.

robertbrignull:
Interesting. Thanks for finding that documentation on transient props. I didn't know that's how they worked.

We do have some other examples in the codebase like https://github.com/github/vscode-codeql/blob/96fb0046c5581769004d8e93ff2f758773689c4b/extensions/ql-vscode/src/view/common/TextButton.tsx#L6 which aren't generating this error, but my guess is either because `size` is lowercase, or because `size` is an allowed DOM attribute, however it's not an attribute that does anything on a `button`. I found and inspected this element to see if it would have a `size="x-small"` just sitting there in the DOM, but surprisingly it didn't. 😕 

We could look at also addressing those cases if that's helpful, or just leave them.

robertbrignull:
> We could look at also addressing those cases if that's helpful, or just leave them.

I'll merge this one now since it's making a warning pop up. I'll open an issue to look at the others and see if we do or don't want to do anything there.

PR #2995, opened 2023-10-18
Use one grid cell per model input, and have the method name span multiple rows

robertbrignull:
This follows on from https://github.com/github/vscode-codeql/pull/2990.

This addresses the point of using `MultiModelColumn` to fit multiple inputs for a certain field but different models into a single grid cell. Instead, we use one grid cell per input, and use `gridRow="span N"` to make the method name span multiple rows.

Here's what it looks like:
Screenshot 2023-10-18 at 15 16 39

I've used a narrow window width so things flow onto multiple lines. The grid lines are visible because I've highlighted that element in the developer tools. Sadly the grid lines don't show where cells span multiple rows/columns, but you can see by the contents that it's working.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
@@ -25,12 +25,6 @@ import { Codicon } from "../common";
 import { canAddNewModeledMethod } from "../../model-editor/shared/multiple-modeled-methods";
 import { DataGridCell, DataGridRow } from "../common/DataGrid";
 
-const MultiModelColumn = styled(DataGridCell)`
-  display: flex;
-  flex-direction: column;
-  gap: 0.5em;
-`;
-
 const ApiOrMethodRow = styled.div`
   min-height: calc(var(--input-height) * 1px);
   display: flex;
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
```diff
@@ -199,54 +193,41 @@ const ModelableMethodRow = forwardRef<HTMLElement | undefined, MethodRowProps>(
...
-        {!props.modelingInProgress && (
-          <>
-            <MultiModelColumn>
-              {modeledMethods.map((modeledMethod, index) => (
+        {!props.modelingInProgress &&
+          modeledMethods.map((modeledMethod, index) => (
+            <>
+              <DataGridCell>
                 <ModelTypeDropdown
-                  key={index}
```
robertbrignull, line 207: I was a little concerned about removing these keys. I've done manual tested and I haven't seen anything not updating correctly. Also, the default behaviour if you don't specify a key is to use the index.

robertbrignull: Ah nice. I did try adding a key but it won't even allow it on a `<>`. Good suggestion to switch to `<Fragment>` instead.

koesie10 (approved):

PR #2997, opened 2023-10-18
Show validation results in the model editor

robertbrignull:
Implements displaying validation in the model editor. Each validation result is added as a `span 5` grid cell, so it sits underneath the modelings.

I wasn't sure whether the validation errors should span the full width or sit only underneath the modelings. I went with just under the modelings, but changing this would be very easy to do now or in the future.

For now I've disabled the "click on error to highlight model" functionality. I think it's less useful here than it is for the method modeling panel because all the modelings are visible at once and it's not that the invalid bit is on a different page. If we want to we could add highlighting of individual modelings, but it would add a little complexity.

Screenshot:

Screenshot 2023-10-19 at 13 01 11

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/method-modeling/ModeledMethodAlert.tsx
+++ b/extensions/ql-vscode/src/view/method-modeling/ModeledMethodAlert.tsx
@@ -6,12 +6,12 @@ import { useCallback } from "react";
 
 type Props = {
   error: ModeledMethodValidationError;
-  setSelectedIndex: (index: number) => void;
+  setSelectedIndex?: (index: number) => void;
 };
 
 export const ModeledMethodAlert = ({ error, setSelectedIndex }: Props) => {
   const handleClick = useCallback(() => {
-    setSelectedIndex(error.index);
+    setSelectedIndex?.(error.index);
   }, [error.index, setSelectedIndex]);
 
   return (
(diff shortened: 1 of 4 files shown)


robertbrignull:
I can add some tests, but this is the general idea of how the UI will look and behave.

extensions/ql-vscode/src/view/method-modeling/ModeledMethodAlert.tsx
```diff
@@ -6,12 +6,12 @@ import { useCallback } from "react";
 
 type Props = {
   error: ModeledMethodValidationError;
-  setSelectedIndex: (index: number) => void;
+  setSelectedIndex?: (index: number) => void;
```
koesie10, line 9: It would be nice if we were able to highlight the model on which the error is located, but this can still be added in a follow-up PR if this is something that comes out of a design review.

koesie10 (approved):
LGTM, but adding a few tests and a story showing the validation errors would definitely be useful

extensions/ql-vscode/src/view/method-modeling/ModeledMethodAlert.tsx
robertbrignull: I've opened a separate issue for this so we don't forget it.

PR #2403, opened 2023-05-05
Run integration tests on 1.77.3

aeisenberg:
This avoids a bug on 1.78.0 on linux. See
https://github.com/github/vscode-codeql/issues/2402

This should fix the integration tests.

PR #2998, opened 2023-10-18
Revert "Run integration tests on 1.77.3"

aeisenberg:
Reverts github/vscode-codeql#2403

I don't think this is fixed yet, but let's see.


aeisenberg:
Looks like this PR is no longer needed since tests are already running on `stable`.

PR #3001, opened 2023-10-20
Add higlight on modeled method row when clicking in validation error

koesie10:
This adds back the clickable link in the validation error alert. When clicking on e.g. "Modify or remove the duplicated classification.", the problematic row will be highlighted. This uses a very similar method to the focusing using `revealedMethodRow`.

I think the name `DataGridRow` for this component is still accurate, so I don't think renaming it will make it clearer how this is used.

Screenshot 2023-10-20 at 11 40 35

## Checklist


robertbrignull (approved):
Nice, that's simpler than I was expecting it to be. I'm glad there's no need for passing in `Ref` objects, since everything is happening within one component.

PR #3002, opened 2023-10-20
Enable `react/jsx-key` for fragment shorthands

koesie10:
This will change ESLint to give an error when you do something like this:

```tsx
{x.map((y) => <>{y.name}</>)}
```

In this case, there's no `key` being passed to the top-level element in a map. We should always explicitly pass in a key to make the best use of React. This can be fixed by doing this:

```tsx
{x.map((y) => <Fragment key={y.id}>{y.name}</Fragment>)}
```

We ran into this in #2995: https://github.com/github/vscode-codeql/pull/2995/files#r1363990725

## Checklist


charisk (approved):
Nice, thanks!

PR #3004, opened 2023-10-20
Show stack for redactable error in log

koesie10:
When calling for example `showAndLogExceptionWithTelemetry`, the stack trace would be sent to Application Insights, but there was no way to see the stack trace from within VS Code. This will add the stack trace to the log by returning it from `fullMessageWithStack` and using it in the appropriate places.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/errors.ts
+++ b/extensions/ql-vscode/src/common/errors.ts
@@ -22,6 +22,14 @@ export class RedactableError extends Error {
       .join("");
   }
 
+  public get fullMessageWithStack(): string {
+    if (!this.stack) {
+      return this.fullMessage;
+    }
+
+    return `${this.fullMessage}\n${this.stack}`;
+  }
+
   public get redactedMessage(): string {
     return this.strings
       .map((s, i) => s + (this.hasValue(i) ? this.getRedactedValue(i) : ""))
--- a/extensions/ql-vscode/src/common/errors.ts
+++ b/extensions/ql-vscode/src/common/errors.ts
@@ -1,3 +1,5 @@
+import * as os from "os";
+
 export class RedactableError extends Error {
   constructor(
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/query-server/legacy/run-queries.ts
```diff
@@ -441,7 +441,6 @@ export async function compileAndRunQueryAgainstDatabaseCore(
         const error = result.message
           ? redactableError`${result.message}`
           : redactableError`Failed to run query`;
-        void extLogger.log(error.fullMessage);
```
koesie10, line 444: This was removed because it results in duplicate logs: the `showAndLogExceptionWithTelemetry` will also log it to the `extLogger`.

extensions/ql-vscode/src/common/errors.ts
```diff
@@ -22,6 +22,14 @@ export class RedactableError extends Error {
       .join("");
   }
 
+  public get fullMessageWithStack(): string {
+    if (!this.stack) {
+      return this.fullMessage;
+    }
+
+    return `${this.fullMessage}\n${this.stack}`;
```
charisk, line 30: Should we use `os.EOL` rather than `\n`?

charisk (approved):
Thanks for tidying that up!

extensions/ql-vscode/src/common/errors.ts
koesie10: Unfortunately we can't use that because the `errors.ts` module is also imported from the webview which doesn't have access to `os.EOL`. I'm not sure whether `\n` works in VS Code's log output, but I don't think it's worth the complexity of trying to determine whether we're in the webview and then deciding to use either `os.EOL` or `\n`.

PR #3007, opened 2023-10-23
Deselect current db when the language context changes

shati-patel:
When the language context changes, we should make sure that the current database gets deselected. See internal linked issue for details 🔍 

### Notes for reviewer

- This feels a bit hacky, but seems to work. A few comments/questions about the implementation inline.
- I also haven't been able to write a test for this (since I couldn't figure out to access the relevant properties/events in the tests), so any suggestions on that front also welcome! 🙏🏽 

## Checklist

N/A—still feature-flagged for internal use ⛳

### DIFF
--- a/extensions/ql-vscode/src/databases/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/databases/local-databases-ui.ts
@@ -141,9 +141,7 @@ class DatabaseTreeDataProvider
     return item;
   }
 
-  public async getChildren(
-    element?: DatabaseItem,
-  ): Promise<DatabaseItem[] | null | undefined> {
+  public getChildren(element?: DatabaseItem): ProviderResult<DatabaseItem[]> {
     if (element === undefined) {
       // Filter items by language
       const displayItems = this.databaseManager.databaseItems.filter((item) => {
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -141,7 +141,9 @@ class DatabaseTreeDataProvider
     return item;
   }
 
-  public getChildren(element?: DatabaseItem): ProviderResult<DatabaseItem[]> {
+  public async getChildren(
```
shati-patel, line 144: I had to make `getChildren` async (so that I could use `await this.databaseManager.setCurrentDatabaseItem`). It seems to be fine, but if there's any risk to doing this, let me know 🙈

extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -150,6 +152,19 @@ class DatabaseTreeDataProvider
...
+        !this.languageContext.isSelectedLanguage(
+          tryGetQueryLanguage(this.currentDatabaseItem.language),
+        )
+      ) {
+        this.currentDatabaseItem = undefined;
+        await this.databaseManager.setCurrentDatabaseItem(
+          this.currentDatabaseItem,
+        );
+        this._onDidChangeTreeData.fire(this.currentDatabaseItem);
+      }
```
shati-patel, line 166: This seems to work (from local testing), but I'm wondering if the `getChildren` method is the best place for this logic. Should it maybe be handled in the database manager? 😵

koesie10: I don't think the `getChildren` method is the best place for this. Could we listen to the `onLanguageContextChangedEmitter` event in the database manager and deselect it there? That would make more sense to me.

shati-patel: Makes sense! I'll have a go at that 🍀

shati-patel:
I'm having issues with testing locally at the moment, so I haven't been able to write/debug my test properly 😭 Putting this back up for review for now and I can add tests later. (Or someone else with a working test runner 😅 )

koesie10 (approved):
LGTM, thanks for moving it to the database manager! I don't think we have any existing tests for the database manager, so I'm not sure whether adding tests for this specifically will make sense. There's only a single test where we're currently creating a `DatabaseManager` instance related to actually running queries. I'm not sure if the benefits of this (fairly simple) tests outweigh the costs related to maintaining this test and the construction of the database manager.

PR #3008, opened 2023-10-24
Fix styling of "already modeled" rows when showing multiple methods

robertbrignull:
Addresses two bugs around the `UnmodelableMethodRow` component. I believe we missed these until now because this component is only shown when the "Hide modeled methods" option is deselected.

One bug is it wasn't computing the `gridColumn: span` value correctly, so when multiple models were shown it would break the entire table layout by trying to put two rows worth of elements on the same grid row. Thankfully this visual bug only manifested when the `codeQL.model.showMultipleModels` feature flag was enabled, and there's no visual change when it isn't enabled.

Another bug is that the `MethodClassifications` component was being added at a different point than from the normal method rows. We were adding it near the end of the cell instead of near the start. This PR changes it to match the other cells.

Screenshots when `showMultipleModels` is disabled:
- Before this PR:
    Screenshot 2023-10-24 at 10 16 28
- After this PR:
    Screenshot 2023-10-24 at 10 18 33

Screenshots when `showMultipleModels` is enabled:
- Before this PR:
    Screenshot 2023-10-24 at 10 16 05
- After this PR:
    Screenshot 2023-10-24 at 10 14 33

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
@@ -337,7 +337,9 @@ const UnmodelableMethodRow = forwardRef<
           <MethodClassifications method={method} />
         </ApiOrMethodRow>
       </DataGridCell>
-      <DataGridCell gridColumn="span 4">Method already modeled</DataGridCell>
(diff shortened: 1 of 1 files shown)


norascheuch (approved):
Looks good! 🚢

PR #3010, opened 2023-10-24
Add method modeling flow to test plan

norascheuch:
This PR adds checks for manually testing the new method modeling flow to our release test plan.

## Checklist

### DIFF
--- a/docs/test-plan.md
+++ b/docs/test-plan.md
@@ -152,27 +152,14 @@ Run one of the above MRVAs, but cancel it from within VS Code:
    - Check that the editor loads and shows methods to model.
    - Check that methods are grouped per library (e.g. `rocksdbjni@7.7.3` or `asm@6.0`)
    - Check that the "Open database" link works.
-   - Check that a method can have multiple models:
-      - add a couple of new models for one method
-      - save and check that the modeling file (use the 'open extension pack' button to open it) shows multiple methods
-      - check that a 'duplicated classification' error appears in both model editor and modeling panel when a duplicate modeling occurs
-      - check that a 'conflicting classigication' error appears when a neutral model type is paired with a source kind
-      - check that clicking on the error highlights the correct modeling in both the editor and the modeling panel
-3. Check the Method Usage
-   - Check that the Method Usage Panel opens and jumps to the correct usage when clicking on 'View' in the model editor
-   - Check that the usages are opening when clicking on a usage
-   - Check that the usage icon color turns green when saving a newly modeled method
-   - Check that the usage icon color turns red when saving a newly unmodeld method
-4. Check the Method Modeling
-   - Check that the 'Start modeling' button opens a new model editor
-   - Check that it refreshes the blank state when a model editor is opened/closed
-   - Check that when modeling in the editor the modeling panel updates accordingly
-   - Check that when modeling in the modeling panel the model editor updates accordingly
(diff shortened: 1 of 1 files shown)


norascheuch:
I wonder how many of these checks should be mandatory, what can be optional.
At the moment we're not testing the method modeling panels multiple method behaviour. Should this be mandatory/optional?

docs/test-plan.md
```diff
@@ -152,6 +152,22 @@ Run one of the above MRVAs, but cancel it from within VS Code:
    - Check that the editor loads and shows methods to model.
    - Check that methods are grouped per library (e.g. `rocksdbjni@7.7.3` or `asm@6.0`)
    - Check that the "Open database" link works.
+   - Check that a method can have multiple models:
+      - add a couple of new models for one method
+      - save and check that the modeling file (use the 'open extension pack' button to open it) shows multiple methods
+      - check that a 'duplicated classification' error appears in both model editor and modeling panel when a duplicate modeling occurs
+      - check that a 'conflicting classigication' error appears when a neutral model type is paired with a source kind
```
koesie10, line 159: Neutral models aren't conflicting with source kinds only, they are conflicting with models that have the same `type` as the neutral model's `kind`. What about:

```suggestion
      - check that a 'conflicting classification' error appears when a neutral model type is paired with a model of the same kind
```

docs/test-plan.md
```diff
@@ -152,6 +152,22 @@ Run one of the above MRVAs, but cancel it from within VS Code:
...
+      - check that clicking on the error highlights the correct modeling in both the editor and the modeling panel
+3. Check the Method Usage
+   - Check that the Method Usage Panel opens and jumps to the correct usage when clicking on 'View' in the model editor
+   - Check that the usages are opening when clicking on a usage
```
koesie10, line 163: We had a bug here where the first one would show correctly, but clicking on any other ones would still only show the first one. Would it make sense to change this to something like:

```suggestion
   - Check that the second and third usages show when clicking on them
```

koesie10 (commented):
Thanks for adding this! Multiple models are still feature-flagged right now, but since it is going to be removed soon, it's probably fine to not mention that here.

I'm not sure whether these should be optional or not. It's probably good to test them when we have made any changes to them, but don't need to test them when we've only made changes to unrelated code.

robertbrignull (commented):
I agree with Koen's points. But other than that looks good to me

norascheuch:
I implemented Koens suggestions and also moved most test cases to optional. In the mandatory testing we now only check that both Panels exist, show the first usage (usage panel) and show and update modeling (modeling panel).

docs/test-plan.md
```diff
@@ -189,9 +192,28 @@ Are there any components that are not showing up?
 
 ## Optional Test Cases
 
-These are mostly aimed at MRVA, but some of them are also applicable to non-MRVA queries.
-
-### Selecting repositories to run on
+### Modeling Flow
+
+2. Check that a method can have multiple models:
```
robertbrignull, line 197: The linter has spotted this needs to start with `1`, though I believe it won't actually affect the rendering it'll be nice to be consistent.

robertbrignull (approved):
There's a tiny linter error, but otherwise the changes LGTM

Thanks for moving stuff into "optional". I think that's the right move for these extra tests of the model editor.

PR #3012, opened 2023-10-24
Create new queries in selected folder of queries panel

koesie10:
This will change the behavior of the "Create new query" command to create the new query in the same folder as the first selected item in the queries panel. If no items are selected, the behavior is the same as before.

I've used events to communicate the selection from the queries panel to the local queries module. This is some more code and some extra complexity, but it ensures that we don't have a dependency from the local queries module to the queries panel module. This makes testing easier.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -795,7 +795,11 @@ async function activateWithInstalledDistribution(
   );
   ctx.subscriptions.push(databaseUI);
 
-  QueriesModule.initialize(app, languageContext, cliServer);
+  const queriesModule = QueriesModule.initialize(
+    app,
+    languageContext,
+    cliServer,
+  );
 
   void extLogger.log("Initializing evaluator log viewer.");
   const evalLogViewer = new EvalLogViewer();
@@ -933,6 +937,10 @@ async function activateWithInstalledDistribution(
     languageContext,
   );
   ctx.subscriptions.push(localQueries);
+
+  queriesModule.onDidChangeSelection((event) =>
+    localQueries.setSelectedQueryTreeViewItems(event.selection),
+  );
 
   void extLogger.log("Initializing debugger factory.");
   ctx.subscriptions.push(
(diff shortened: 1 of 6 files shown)


shati-patel (commented):
I haven't taken a really close look at the code yet, but I've been trying to get it to work locally!

At the moment, I can't create a query inside most folders because the `codeql-custom-queries-xxx` folder doesn't exist. E.g. when highlighting the `ql/cpp/ql` folder, I get 
```
Could not create query example file: Error: ENOENT: no such file or directory, scandir 'c:\git-repo\vscode-codeql-starter\ql\cpp\ql\codeql-custom-queries-cpp'
```

Am I missing something about to use this? 😅 Or maybe I'm hitting a Windows bug 🐛 👀

koesie10:
> Am I missing something about to use this? 😅 Or maybe I'm hitting a Windows bug 🐛 👀

I was testing in a different workspace which didn't have the `codeql-custom-queries-*` folders as workspace folders, but it seems like I missed this scenario. I've now removed `isFolderAlreadyInWorkspace` as a check since this should already be handled by `pathExists(join(this.qlPackStoragePath, this.folderName))`. Could you try again?

shati-patel (approved):
Implementation looks good to me! And thank you for fixing the `no such file or directory` errors I had earlier 📂 🎉 

One more thing I'm wondering about (but feel free to just merge this and potentially follow up with James later):
Could we create the query (and the ql pack) _directly_ inside the highlighted folder, instead of nested inside a `codeql-custom-queries-*` folder? E.g. in this video, I'd expect the query to be created directly inside `example/snippets`, since that's what the file explorer does. 

https://github.com/github/vscode-codeql/assets/42641846/bf1f4570-3a96-43d1-92db-e7f31181dfe4

However, it might be awkward to make sure we're in a suitable QL pack, so perhaps this isn't feasible!

koesie10:
Thanks @shati-patel, I agree that that behavior would be more useful. I've created an internal issue to track this, but I'll merge this now since it's already an improvement over the previous behavior.

PR #3014, opened 2023-10-24
Use the same logic for hiding methods in model editor and usages panel

robertbrignull:
Makes sure that the model editor and the usages panel are using the same logic when deciding whether to show or hide a method.

Before this PR, you wouldn't be able to view usages of any methods that were already modeled and the modeling had been saved. We previously thought this bug only applied when you had modelings coming from an external pack, but you can reproduce it within one pack too.

Steps to reproduce:
- Add a modeling to a method
- Save the extension pack
- Click "View" on the modeled method and observe that nothing happens

With this PR, the usages panel and the model editor are fully in sync and show the same methods and clicking "View" always highlights the method in the usages panel.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
+++ b/extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
@@ -9,7 +9,7 @@ import {
   Uri,
 } from "vscode";
 import { DisposableObject } from "../../common/disposable-object";
-import { Method, Usage } from "../method";
+import { Method, Usage, canMethodBeModeled } from "../method";
 import { DatabaseItem } from "../../databases/local-databases";
 import { relative } from "path";
 import { CodeQLCliServer } from "../../codeql-cli/cli";
@@ -146,7 +146,13 @@ export class MethodsUsageDataProvider
   getChildren(item?: MethodsUsageTreeViewItem): MethodsUsageTreeViewItem[] {
     if (item === undefined) {
       if (this.hideModeledMethods) {
-        return this.sortedTreeItems.filter((api) => !api.method.supported);
+        return this.sortedTreeItems.filter((api) =>
+          canMethodBeModeled(
+            api.method,
+            this.modeledMethods[api.method.signature] ?? [],
+            this.modifiedMethodSignatures.has(api.method.signature),
+          ),
+        );
       } else {
         return [...this.sortedTreeItems];
       }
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/view/model-editor/ModeledMethodDataGrid.tsx
```diff
@@ -51,8 +55,7 @@ export const ModeledMethodDataGrid = ({
         modeledMethods,
         methodIsUnsaved,
       );
-
-      if (methodCanBeModeled || !hideModeledMethods) {
+      if (shouldShowMethod(methodCanBeModeled, hideModeledMethods)) {
```
robertbrignull, line 58: The code here and in `MethodsUsageDataProvider` is very similar but not identical, so I didn't think it was sensible to combine `canMethodBeModeled` and `shouldShowMethod` into one shared method. Here we need both values while in the usages panel we only need to know whether to show it or not, and the method modelability it just an intermediate value.

extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
```diff
@@ -146,7 +146,14 @@ export class MethodsUsageDataProvider
   getChildren(item?: MethodsUsageTreeViewItem): MethodsUsageTreeViewItem[] {
     if (item === undefined) {
       if (this.hideModeledMethods) {
-        return this.sortedTreeItems.filter((api) => !api.method.supported);
+        return this.sortedTreeItems.filter((api) => {
```
robertbrignull, line 149: I always forget before opening the initial PR, but I can add some tests that cover this behaviour change.

extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
```diff
@@ -146,7 +146,14 @@ export class MethodsUsageDataProvider
...
+          const methodCanBeModeled = canMethodBeModeled(
+            api.method,
+            this.modeledMethods[api.method.signature] ?? [],
+            this.modifiedMethodSignatures.has(api.method.signature),
+          );
+          return shouldShowMethod(methodCanBeModeled, this.hideModeledMethods);
```
robertbrignull, line 155: I'm being dumb. `this.hideModeledMethods` is always true here, so this is basically a no-op. The real change here is calling `canMethodBeModeled` and we don't need the change to `shouldShowMethod` at all.

charisk (approved):
LGTM!

PR #3017, opened 2023-10-25
Add experimental model editor support for Ruby

hmac:
Make the minimum changes necessary for prototype Ruby support in the model editor.

This consists of:

- Reading/writing modelled methods from/to data extensions in the dynamic languages format
- Special-casing Ruby in a few places where Java/C# was previously assumed.

## Checklist


starcke (approved):
I think this looks great, and works for me as well. We probably want at least someone else from the team to take a look as I was involved in the implementation as well.

extensions/ql-vscode/src/model-editor/model-editor-view.ts
```diff
@@ -376,6 +376,8 @@ export class ModelEditorView extends AbstractWebview<
     const sourceArchiveAvailable =
       this.databaseItem.hasSourceArchiveInExplorer();
 
+    const showModeSwitchButton = this.language !== QueryLanguage.Ruby;
```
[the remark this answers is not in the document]
koesie10: I agree that this is a weird inconsistency. I've pushed a commit that fixes this by adding the `availableModes` to the language definition.

extensions/ql-vscode/src/model-editor/model-editor-module.ts
```diff
@@ -125,17 +129,24 @@ export class ModelEditorModule extends DisposableObject {
...
         void showAndLogErrorMessage(
           this.app.logger,
           `The CodeQL Model Editor is not supported for ${language} databases.`,
         );
         return;
       }
 
+      const definition = getModelsAsDataLanguage(language);
+
+      const initialMode = definition.availableModes?.[0] ?? Mode.Application;
```
robertbrignull, line 148:
```suggestion
      const initialMode = definition.availableModes?.[0] ?? INITIAL_MODE;
```
where `INITIAL_MODE` comes from `mode.ts`. Just so we reuse the constant instead of hard-coding `Mode.Application`.

extensions/ql-vscode/src/model-editor/model-editor-module.ts
```diff
@@ -125,17 +129,24 @@ export class ModelEditorModule extends DisposableObject {
       }
 
       const language = db.language;
-      if (
-        !SUPPORTED_LANGUAGES.includes(language) ||
-        !isQueryLanguage(language)
-      ) {
+
+      // Ruby is only enabled when the config setting is set
+      const isSupportedLanguage =
```
robertbrignull, line 134: If we're making changes anyway, then pulling this little bit out to a function would be good. That way it'll be more reusable if we need to do this in other places.

It could also help to make the `SUPPORTED_LANGUAGES` constant a bit less error-prone. Right now we have a constant which doesn't really do what it sounds like it does because you need to remember to also check Ruby. If we moved it all into a `isSupportedLanguage` function that would be safer.

robertbrignull (approved):
I have a couple of tiny comments, but they're honestly not essential. Otherwise LGTM and I like that the "experimental" bits are all within one of the two `ruby.ts` files so the other languages are basically unaffected.

I haven't really tried out the ruby modeling as I didn't have a database ready and I assume you'll be doing that, but I have briefly tested out java modeling to make sure it's unaffected. Everything looks fine to me.

extensions/ql-vscode/src/model-editor/model-editor-module.ts
koesie10: Thanks for the suggestion, that does seem less error-prone.

PR #3031, opened 2023-10-30
Switch `ModeledMethod` to union of types

koesie10:
This switches the `ModeledMethod` type from 1 type containing all fields to a union of multiple types where each type accurately represents which fields are present for a specific model. This allows more disjointed models to be represented more accurately, such as type models. For type models, having` input`, `output`, etc. doesn't make sense. It instead has fields like `type1`, `type2` and `path`. I don't think adding these properties to every modeled method makes sense, so this is a solution which allows us to represent both models.

This is a very large changeset, but I don't think splitting it up would be helpful (I don't really see how). A lot of the changes are changes to method names or removing unnecessary properties.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/languages/languages.ts
+++ b/extensions/ql-vscode/src/model-editor/languages/languages.ts
@@ -28,7 +28,7 @@ export function getModelsAsDataLanguageModel<
 ): NonNullable<ModelsAsDataLanguagePredicates[T]> {
   const definition = getModelsAsDataLanguage(language).predicates[model];
   if (!definition) {
-    throw new Error(`No models-as-data definition for ${model}`);
+    throw new Error(`No models-as-data predicate for ${model}`);
   }
   return definition;
 }
--- a/extensions/ql-vscode/src/model-editor/languages/languages.ts
+++ b/extensions/ql-vscode/src/model-editor/languages/languages.ts
@@ -1,5 +1,8 @@
 import { QueryLanguage } from "../../common/query-language";
-import { ModelsAsDataLanguage } from "./models-as-data";
+import {
+  ModelsAsDataLanguage,
+  ModelsAsDataLanguagePredicates,
+} from "./models-as-data";
 import { staticLanguage } from "./static";
 
 const languages: Partial<Record<QueryLanguage, ModelsAsDataLanguage>> = {
@@ -13,6 +16,19 @@ export function getModelsAsDataLanguage(
   const definition = languages[language];
   if (!definition) {
     throw new Error(`No models-as-data definition for ${language}`);
+  }
+  return definition;
+}
+
+export function getModelsAsDataLanguageModel<
+  T extends keyof ModelsAsDataLanguagePredicates,
+>(
+  language: QueryLanguage,
+  model: T,
+): NonNullable<ModelsAsDataLanguagePredicates[T]> {
+  const definition = getModelsAsDataLanguage(language).predicates[model];
+  if (!definition) {
+    throw new Error(`No models-as-data definition for ${model}`);
   }
   return definition;
 }
--- a/extensions/ql-vscode/src/model-editor/modeled-method.ts
+++ b/extensions/ql-vscode/src/model-editor/modeled-method.ts
@@ -19,12 +19,87 @@ export type Provenance =
   // Entered by the user in the editor manually
   | "manual";
 
-export interface ModeledMethod extends MethodSignature {
-  readonly type: ModeledMethodType;
+export interface NoneModeledMethod extends MethodSignature {
+  readonly type: "none";
+  // Provenance is always propagated
+  readonly provenance: Provenance;
+}
+
+export interface SourceModeledMethod extends MethodSignature {
+  readonly type: "source";
+  readonly output: string;
+  readonly kind: ModeledMethodKind;
+  readonly provenance: Provenance;
+}
+
+export interface SinkModeledMethod extends MethodSignature {
+  readonly type: "sink";
+  readonly input: string;
+  readonly kind: ModeledMethodKind;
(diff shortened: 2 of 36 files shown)


extensions/ql-vscode/src/model-editor/languages/languages.ts
```diff
@@ -16,3 +19,16 @@ export function getModelsAsDataLanguage(
...
+
+export function getModelsAsDataLanguageModel<
+  T extends keyof ModelsAsDataLanguagePredicates,
+>(
+  language: QueryLanguage,
+  model: T,
+): NonNullable<ModelsAsDataLanguagePredicates[T]> {
+  const definition = getModelsAsDataLanguage(language).predicates[model];
+  if (!definition) {
+    throw new Error(`No models-as-data definition for ${model}`);
```
robertbrignull, line 31:
```suggestion
    throw new Error(`No models-as-data predicate for ${model}`);
```
Should this error message be different from the error message from `getModelsAsDataLanguage` so it's easier to tell what's happened if we don't have the full stack trace.

extensions/ql-vscode/src/model-editor/yaml.ts
```diff
@@ -34,38 +41,92 @@ function createDataProperty(
...
-    sink: [],
-    summary: [],
-    neutral: [],
-  };
+  const methodsByType = {
+    source: [] as SourceModeledMethod[],
+    sink: [] as SinkModeledMethod[],
+    summary: [] as SummaryModeledMethod[],
+    neutral: [] as NeutralModeledMethod[],
+  } satisfies Record<keyof ModelsAsDataLanguagePredicates, ModeledMethod[]>;
```
robertbrignull, line 71: I've never seen the [`satisfies` operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html) before but it looks useful.

extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
```diff
@@ -167,9 +167,6 @@ const ModelableMethodRow = forwardRef<HTMLElement | undefined, MethodRowProps>(
     const handleAddModelClick = useCallback(() => {
       const newModeledMethod: ModeledMethod = {
```
robertbrignull, line 168: Could replace with `createEmptyModeledMethod("none", method)`

extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
```diff
@@ -362,9 +359,6 @@ function modeledMethodsToDisplay(
     return [
       {
```
robertbrignull, line 360: Could replace with `createEmptyModeledMethod("none", method)`

extensions/ql-vscode/src/view/model-editor/ModelKindDropdown.tsx
```diff
@@ -66,19 +67,26 @@ export const ModelKindDropdown = ({
...
+    } else if (kinds.length > 0 && !kinds.includes(value ?? "")) {
       onChangeKind(kinds[0]);
     }
-  }, [modeledMethod?.kind, kinds, onChangeKind]);
+  }, [modeledMethod, value, kinds, onChangeKind]);
 
   return (
     <Dropdown
-      value={modeledMethod?.kind}
+      value={value ?? undefined}
```
robertbrignull, line 89: If `value` had the type `string | false` then this would coerce that to `string | undefined`. However, I believe the type of `value` is already `string | undefined`, so is `?? undefined` unneccesary here?

extensions/ql-vscode/src/model-editor/modeled-method.ts
```diff
@@ -19,12 +19,87 @@ export type Provenance =
   // Entered by the user in the editor manually
   | "manual";
 
-export interface ModeledMethod extends MethodSignature {
-  readonly type: ModeledMethodType;
+export interface NoneModeledMethod extends MethodSignature {
+  readonly type: "none";
+  // Provenance is always propagated
+  readonly provenance: Provenance;
```
robertbrignull, line 25: I don't understand what this means that "provenance is always propagated". Could you explain a bit more? How can we have a provenance on a "none" modeling?

robertbrignull (commented):
I think this looks really good! I like splitting the `ModeledMethod` type up so that each sub-type only contains the fields it needs.

I have a few small comments asking for clarification or small improvements, but the overall idea and changes look great.

extensions/ql-vscode/src/model-editor/yaml.ts
koesie10: It's quite new, but it's perfect here to ensure that all keys are included, and all values are assignable to `ModeledMethod[]`.

extensions/ql-vscode/src/view/model-editor/ModelKindDropdown.tsx
koesie10: Yep, it's unnecessary here, thanks for spotting that. I believe I first had `string | null` here, which was not an acceptable type for the `Dropdown` `value`.

I don't think `string | false` would actually be affected by this operator. `??` is different from `||` and will only change the result when the input is either `null` or `undefined`, not any falsy value. So, this was there to catch the `null` that was here before.

extensions/ql-vscode/src/model-editor/modeled-method.ts
koesie10: This would essentially ensure that if you ran flow generation and then changed the type to unmodeled, and then changed it to something else, it would still show `df-manual`.

I actually think that doesn't make a ton of sense anymore, so I've removed this field. I believe that in a previous iteration it was somewhat harder to do, but there were very few places where we actually used this.

PR #3035, opened 2023-10-31
Fix scroll into view not working when revealing method

koesie10:
This fixes a bug where the method row would not scroll into view when revealing a method. The problem was that the `DataGridRow` component on which the `ref` was set is a `display: contents` element, which does not have a visual representation in the DOM. Therefore, it wasn't possible to scroll the method row into view. This fixes it by moving the ref to the `DataGridCell` component of the first column, which is a normal element.

## Checklist


robertbrignull (approved):
This fixes the bug for me.

PR #3040, opened 2023-11-01
Prevent duplicate query packs when creating a query

koesie10:
This prevents the creation of duplicate query pack names when creating a query in the following ways:
- When you have selected a folder, the query pack name will include the name of the folder. This should prevent duplicate query pack names when creating queries in different folders. - When the folder name includes `codeql` or `queries`, we will not add `codeql-extra-queries-` since that would be redundant.
- After generating the query pack name, we will resolve all qlpacks and check if one with this name already exists. If it does, we will start adding an index to the name until we find a unique name.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/local-queries/qlpack-generator.ts
+++ b/extensions/ql-vscode/src/local-queries/qlpack-generator.ts
@@ -1,12 +1,14 @@
 import { mkdir, writeFile } from "fs-extra";
 import { dump } from "js-yaml";
-import { join } from "path";
+import { dirname, join } from "path";
 import { Uri } from "vscode";
 import { CodeQLCliServer } from "../codeql-cli/cli";
 import { QueryLanguage } from "../common/query-language";
+import { getOnDiskWorkspaceFolders } from "../common/vscode/workspace-folders";
+import { basename } from "../common/path";
 
 export class QlPackGenerator {
-  private readonly qlpackName: string;
+  private qlpackName: string | undefined;
   private readonly qlpackVersion: string;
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/local-queries/qlpack-generator.ts
```diff
@@ -39,6 +43,37 @@ export class QlPackGenerator {
...
 
+  private async determineQlpackName(): Promise<string> {
+    let qlpackBaseName = `getting-started/codeql-extra-queries-${this.queryLanguage}`;
+    if (this.includeFolderNameInQlpackName) {
+      const folderBasename = basename(dirname(this.folderUri.fsPath));
+      if (
+        folderBasename.includes("codeql") ||
+        folderBasename.includes("queries")
+      ) {
+        // If the user has already included "codeql" or "queries" in the folder name, don't include it twice
```
shati-patel, line 54: Nice idea!

shati-patel (approved):
Looks good! 

I found a minor bug where a query got created "one level up" from where it was expected, so the extension can't find/open the query:

https://github.com/github/vscode-codeql/assets/42641846/12c0a7ce-8e3d-436a-8955-271c747fdbab

I.e. in this case, we expect the query to be inside `ql/swift/ql/examples/snippets`, but instead it gets created in `ql/swift/ql/examples` (probably because the latter is the "QL pack root" 🤔 )

---- 

On closer inspection, this is actually a bug I must have missed in a previous PR, not this one... So feel free to merge this and address that separately 🙈

PR #3041, opened 2023-11-01
Add telemetry for method modeling panel

charisk:
Adds some telemetry events for interactions with the modeling panel.

The `method-modeling-set-multiple-modeled-methods` is a bit generic, but we can improve that in the future if we want to.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
+++ b/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
@@ -133,16 +133,26 @@ export class MethodModelingViewProvider extends AbstractWebviewViewProvider<
           this.databaseItem,
           msg.methodSignature,
         );
+
+        void telemetryListener?.sendUIInteraction(
+          "method-modeling-set-multiple-modeled-methods",
+        );
         break;
       }
       case "revealInModelEditor":
         await this.revealInModelEditor(msg.method);
+        void telemetryListener?.sendUIInteraction(
+          "method-modeling-reveal-in-model-editor",
+        );
 
         break;
 
       case "startModeling":
         await this.app.commands.execute(
           "codeQL.openModelEditorFromModelingPanel",
+        );
+        void telemetryListener?.sendUIInteraction(
+          "method-modeling-start-modeling",
         );
         break;
       default:
--- a/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
+++ b/extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
@@ -151,9 +151,6 @@ export class MethodModelingViewProvider extends AbstractWebviewViewProvider<
         await this.app.commands.execute(
           "codeQL.openModelEditorFromModelingPanel",
         );
-        void telemetryListener?.sendUIInteraction(
-          "method-modeling-start-modeling",
-        );
         break;
       default:
         assertNever(msg);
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
```diff
@@ -133,17 +133,27 @@ export class MethodModelingViewProvider extends AbstractWebviewViewProvider<
...
+          "method-modeling-reveal-in-model-editor",
+        );
 
         break;
 
       case "startModeling":
         await this.app.commands.execute(
           "codeQL.openModelEditorFromModelingPanel",
         );
+        void telemetryListener?.sendUIInteraction(
```
robertbrignull, line 154: Do we need to emit ui-interaction telemetry here when the main body of the message handler just triggers the `codeQL.openModelEditorFromModelingPanel` command.

In cases like this where the interaction already triggers a unique command I've generally avoided logged extra telemetry for it.

extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
```diff
@@ -133,17 +133,27 @@ export class MethodModelingViewProvider extends AbstractWebviewViewProvider<
           this.databaseItem,
           msg.methodSignature,
         );
+
+        void telemetryListener?.sendUIInteraction(
```
robertbrignull, line 137: This will get triggered every time anyone changes any aspect of a modeling. Do we want it to trigger that frequently?

It will tell us that people are using the method modeling panel, but I wonder if we have other ways of telling that. Or do we want to get info at this level of detail and frequency?

robertbrignull (commented):
I have a few comments, but they're all very subjective so please do say if you feel otherwise.

There are also maybe some other interactions, but I don't know if we want to emit telemetry for them or not:
- Changing index (i.e. view next/previous modeling)
- Adding a new modeling
- Removing a modeling

Though adding/removing would also trigger the catch-all "setMultipleModeledMethods" event.

extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
charisk: Oh yeah you're right, don't need this!

extensions/ql-vscode/src/model-editor/method-modeling/method-modeling-view-provider.ts
charisk: I've removed this - I was under the impression we did the same in the model editor, but we don't. If we find that we need information around this we can add some telemetry in the future.

charisk:
> I have a few comments, but they're all very subjective so please do say if you feel otherwise.
> 
> There are also maybe some other interactions, but I don't know if we want to emit telemetry for them or not:
> 
> * Changing index (i.e. view next/previous modeling)
> * Adding a new modeling
> * Removing a modeling
> 
> Though adding/removing would also trigger the catch-all "setMultipleModeledMethods" event.

Thanks. I've stopped being lazy and added events for those events too.

extensions/ql-vscode/src/view/method-modeling/MultipleModeledMethodsPanel.tsx
```diff
@@ -123,6 +127,7 @@ export const MultipleModeledMethodsPanel = ({
 
     onChange(method.signature, newModeledMethods);
     setSelectedIndex(newSelectedIndex);
+    sendTelemetry("method-modeling-add-model");
```
robertbrignull, line 130:
```suggestion
    sendTelemetry("method-modeling-remove-model");
```

robertbrignull (approved):
One typo, otherwise LGTM

PR #3043, opened 2023-11-01
Unify model generation query execution

koesie10:
This unifies the two code paths for model generation (flow model generation/generate model) by moving it to the language definition. There are some user-visible changes:
- The notification now shows `CaptureSinkModels.ql` instead of `sink`. Same for the other queries.
- There is no hard error if multiple queries are found for Ruby. This shouldn't be possible in theory since we're resolving a specific path.
- There shouldn't really be any other user-visible changes.

## Checklist


extensions/ql-vscode/src/model-editor/generate.ts
```diff
@@ -0,0 +1,99 @@
...
+  );
+
+  const filteredQueryPaths = filterQueries
+    ? queryPaths.filter(filterQueries)
+    : queryPaths;
+
+  const maxStep = filteredQueryPaths.length * 1000;
+
+  for (let i = 0; i < filteredQueryPaths.length; i++) {
+    const queryPath = filteredQueryPaths[i];
```
starcke, line 53: I dont think this is a problem, but would this just ignore it if one of the queries wasnt found whereas before it would error?

starcke (approved):
Looks good to me. Just a question.

extensions/ql-vscode/src/model-editor/generate.ts
koesie10: Yes, that's another difference that I forgot to mention. It should still error if it doesn't find any queries, so I think this is fine.

PR #3047, opened 2023-11-01
Don't add database source archive folders by default

shati-patel:
No longer add the database source folder to the workspace by default (since this caused bugs in single-folder workspaces). If users want to keep the old behaviour of auto-adding DB source folders, they can set the `addDatabaseSourceToWorkspace` setting to `true`.

See internal linked issue for more details.

## Checklist

  - https://github.com/github/codeql/pull/14668

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -372,18 +372,18 @@
       },
       {
         "type": "object",
-        "title": "Downloading databases",
+        "title": "Adding databases",
         "order": 6,
         "properties": {
-          "codeQL.databaseDownload.allowHttp": {
+          "codeQL.addingDatabases.allowHttp": {
             "type": "boolean",
             "default": false,
             "description": "Allow databases to be downloaded via HTTP. Warning: enabling this option will allow downloading from insecure servers."
           },
-          "codeQL.databaseDownload.addDatabaseSourceToWorkspace": {
+          "codeQL.addingDatabases.addDatabaseSourceToWorkspace": {
             "type": "boolean",
             "default": false,
-            "markdownDescription": "When downloading a database, automatically add the database's source folder as a workspace folder. Warning: enabling this option in a single-folder workspace will cause the workspace to reload as a [multi-root workspace](https://code.visualstudio.com/docs/editor/multi-root-workspaces). This may cause query history and database lists to be reset."
+            "markdownDescription": "When adding a CodeQL database, automatically add the database's source folder as a workspace folder. Warning: enabling this option in a single-folder workspace will cause the workspace to reload as a [multi-root workspace](https://code.visualstudio.com/docs/editor/multi-root-workspaces). This may cause query history and database lists to be reset."
           }
         }
       },
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -378,7 +378,12 @@
           "codeQL.databaseDownload.allowHttp": {
             "type": "boolean",
             "default": false,
-            "description": "Allow database to be downloaded via HTTP. Warning: enabling this option will allow downloading from insecure servers."
+            "description": "Allow databases to be downloaded via HTTP. Warning: enabling this option will allow downloading from insecure servers."
+          },
+          "codeQL.databaseDownload.addDatabaseSourceToWorkspace": {
+            "type": "boolean",
+            "default": false,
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/model-editor/model-editor-view.ts
```diff
@@ -713,7 +713,6 @@ export class ModelEditorView extends AbstractWebview<
       this.cliServer,
       this.databaseItem.language,
       makeSelected,
-      false,
     );
     if (!addedDatabase) {
```
shati-patel, line 717: I assume we no longer need to treat the model editor as a special case, so I've removed the argument here.

extensions/ql-vscode/test/vscode-tests/minimal-workspace/local-queries/local-databases.test.ts
```diff
@@ -765,7 +782,7 @@ describe("local databases", () => {
 
     describe("when codeQL.codespacesTemplate is set to false", () => {
       it("should not create a skeleton QL pack", async () => {
-        jest.spyOn(Setting.prototype, "getValue").mockReturnValue(false);
+        jest.spyOn(config.Setting.prototype, "getValue").mockReturnValue(false);
```
shati-patel, line 785: These other tests could do with some tidying (e.g. by mocking the specific setting, instead of the general `getValue` function 🤔), but I can do that in a separate PR!

extensions/ql-vscode/src/model-editor/model-editor-view.ts
koesie10: I'm not sure about this change: if you are in the model editor and have the setting enabled, we would now suddenly reload when you try to model a dependency. I think this may be unexpected because the user might not know that the database gets added as a normal database.

extensions/ql-vscode/src/config.ts
```diff
@@ -649,6 +649,15 @@ export function allowHttp(): boolean {
   return ALLOW_HTTP_SETTING.getValue<boolean>() || false;
 }
 
+const ADD_DATABASE_SOURCE_TO_WORKSPACE_SETTING = new Setting(
+  "addDatabaseSourceToWorkspace",
+  DATABASE_DOWNLOAD_SETTING,
```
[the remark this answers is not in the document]
shati-patel: Good point! I can rename the overall setting to something like "Adding databases" (instead of "Downloading databases")

extensions/ql-vscode/src/model-editor/model-editor-view.ts
shati-patel: You're right that this might be unexpected. Since it's less explicit that you're adding a database in this case, I'll keep the hard-coded "don't add source archive" setting! 👍🏽

shati-patel:
Docs are in place, and I've had confirmation that this is good to merge!

extensions/ql-vscode/package.json
```diff
@@ -372,13 +372,23 @@
...
+          "codeQL.addingDatabases.allowHttp": {
+            "type": "boolean",
+            "default": false,
+            "description": "Allow databases to be downloaded via HTTP. Warning: enabling this option will allow downloading from insecure servers."
+          },
           "codeQL.databaseDownload.allowHttp": {
+            "type": "boolean",
+            "markdownDeprecationMessage": "**Deprecated**: Please use `#codeQL.addingDatabases.allowHttp#` instead.",
+            "deprecationMessage": "Deprecated: Please use codeQL.addingDatabases.allowHttp instead."
+          },
```
shati-patel, line 387: As mentioned offline, users who still use this old setting name will get a deprecation message:
image

shati-patel: For the record: we will no longer support the old setting name. This isn't the "formal" way of deprecating a setting, but this is a sufficiently niche/insecure setting that it shouldn't have much user impact. 

In the case where a user did have this setting enabled: If they try to download a DB from an `http` location, they'll get the error message "Must use https for downloading a database.", which should point them to check their settings.

robertbrignull: Discussed offline as well, but this should be fine. As said it's not the full proper way of renaming a setting, and we could aim to support both for a transition period. However we expect this setting to be almost, if not completely unused. And as Shati points out, if someone is using this the error should point them towards checking their settings and seeing the deprecation notice. This along with a changelog entry should mean that anyone who is using it sees it and updates their settings.

PR #3048, opened 2023-11-02
Add `type` as modeled method type

koesie10:
This adds support for modeling types. A MaD language can now optionally define a `type` predicate. This allows internally propagating these models. The UI will now simply show a label "type" for type models without any way to edit these.

For completeness, here is the comment for a `typeModel` predicate in Ruby: "Holds if `(type2, path)` should be seen as an instance of `type1`." In this case, that translates to "Holds if `(typeName, methodName, path)` should be seen as an instance of `relatedTypeName`." I'm open to suggestions on a better name for `relatedTypeName`.

## Checklist


hmac:
I've just tried this out, it works great!

One suggestion I have, if it's not too much effort, is to display the `relatedTypeName` in the UI, instead of the disabled select boxes.

Possibly `relatedTypeName` could be `originalTypeName`? But I don't think it's a perfect choice either.

koesie10:
> One suggestion I have, if it's not too much effort, is to display the `relatedTypeName` in the UI, instead of the disabled select boxes.

That's definitely possible. The column labels are confusing since they still show input and output, but a commit I just pushed now shows all values in disabled dropdowns:

Screenshot 2023-11-03 at 15 58 10


> Possibly `relatedTypeName` could be `originalTypeName`? But I don't think it's a perfect choice either.

I'm not sure about `originalTypeName` either since the `originalTypeName` could be the type of the return value or of an argument, and I'm not sure I would call that the `original` type.

starcke (approved):
I think ultimately we want to show type models differently (maybe just use them to give types to method parameters), but for now this is probably good to at least have the data available to view.

PR #3049, opened 2023-11-02
Use transient props for all styled components

koesie10:
This switches all styled components I could find with custom props to using transient props. It's easiest to review this commit-by-commit; each commit changes 1 component.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisRepositoriesStats.tsx
+++ b/extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisRepositoriesStats.tsx
@@ -27,14 +27,14 @@ function getIcon(
     if (variantAnalysisStatus === VariantAnalysisStatus.Canceled) {
       return (
         <>
-          <HorizontalSpace size={2} />
+          <HorizontalSpace $size={2} />
           <ErrorIcon label="Some analyses were stopped" />
         </>
       );
     } else {
       return (
         <>
-          <HorizontalSpace size={2} />
+          <HorizontalSpace $size={2} />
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisRepositoriesStats.tsx
```diff
@@ -68,7 +68,7 @@ export const VariantAnalysisRepositoriesStats = ({
   if (variantAnalysisStatus === VariantAnalysisStatus.Failed) {
     return (
       <>
-        0<HorizontalSpace size={2} />
```
robertbrignull, line 71: I'm not sure the `0` was a bug. Or if we do want to remove it, do we need to also remove some whitespace before the icon?

It might be simpler regardless to leave it for this PR and address as a separate PR. What do you think?

Screenshot before:
Screenshot 2023-11-02 at 16 10 34

Screenshot after:
Screenshot 2023-11-02 at 16 10 48

robertbrignull (commented):
Changes to transient props look good, though possibly with one unintended change.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisRepositoriesStats.tsx
koesie10: I think you're right. This just looked like an accidental typo, but I didn't actually check whether that was the case.

In this case, I would have expected something like `0 <ErrorIcon label="Variant analysis failed" />` rather than a separate element for the spacing.

charisk (approved):
LGTM!

PR #3060, opened 2023-11-07
Enable noUnusedParameters for the view code

robertbrignull:
This PR enables the `noUnusedParameters` option in `src/view/tsconfig.json` and fixes all of the resulting errors.

With this option enabled it finds unused parameters to function components, e.g. it would detect
```typescript
const MyComponent = ({ myUnusedProp }: Props) => {
  return (
    <span>Hello World!</span>
  );
}
```

However it doesn't appear to detect when there is a field defined on the `Props` type that is never referenced anywhere. So if we removed `myUnusedProp` from the function parameters but left the `Props` type definition as
```typescript
interface Props {
  myUnusedProp: string;
}
```
then it would not cause a linting error.

For this PR I've also removed the field from the props types, but long term we should look into if we can catch these automatically as well. There may well be more unused props in our codebase.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/common/DeterminateProgressRing.tsx
+++ b/extensions/ql-vscode/src/view/common/DeterminateProgressRing.tsx
@@ -3,7 +3,6 @@ import { styled } from "styled-components";
 
 type Props = {
   percent: number;
-  label?: string;
 };
 
 const Circle = styled.div`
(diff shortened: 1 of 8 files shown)


koesie10 (approved):
Thanks, LGTM

PR #3061, opened 2023-11-07
Change @typescript-eslint/no-unused-vars "args" option to "after-used"

robertbrignull:
This PR changes the `args` option for the `@typescript-eslint/no-unused-vars` eslint rule to `after-used` instead of `none`. ([docs](https://eslint.org/docs/latest/rules/no-unused-vars#args))

The place where this comes up a lot is when implementing a subclass, or when passing a function as a parameter to something else. As in, places where you are implementing an existing function interface instead of defining it yourself.

Previously we were ignoring all unused arguments so long as they started with a `_` character. After this PR we'll raise a linting error for all such arguments that come after the last argument that is actually used. You can still use the underscore to ignore earlier arguments, because otherwise there isn't really a way to satisfy the rule.

One could argue that we should leave these unused arguments in because they act as documentation to future developers, so they know that that argument exists for them to use if needed. However I'm not sure this justifies the extra cruft from the unused code. If a developer wants to know the full interface of a function they are implementing, then it is advised for them to look directly at the type of the interface they are implementing.

This PR is based on https://github.com/github/vscode-codeql/pull/3060 because that PR also makes the same changes, although only a subset of the changes. I'll leave both PRs open because https://github.com/github/vscode-codeql/pull/3060 is much smaller and also likely to be less controversial. There are two places in this PR where the rule didn't behave 100% perfectly. I'll put comments on those locations in the code.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/modeling-store.ts
+++ b/extensions/ql-vscode/src/model-editor/modeling-store.ts
@@ -211,8 +211,8 @@ export class ModelingStore extends DisposableObject {
         ...methods,
         // Keep all methods that are already modeled in some form in the state
         ...Object.fromEntries(
-          Object.entries(state.modeledMethods).filter(([_, value]) =>
-            value.some((m) => m.type !== "none"),
+          Object.entries(state.modeledMethods).filter((entry) =>
+            entry[1].some((m) => m.type !== "none"),
           ),
         ),
       };
--- a/extensions/ql-vscode/src/model-editor/modeling-store.ts
+++ b/extensions/ql-vscode/src/model-editor/modeling-store.ts
@@ -211,8 +211,8 @@ export class ModelingStore extends DisposableObject {
         ...methods,
         // Keep all methods that are already modeled in some form in the state
         ...Object.fromEntries(
-          Object.entries(state.modeledMethods).filter((entry) =>
-            entry[1].some((m) => m.type !== "none"),
+          Object.entries(state.modeledMethods).filter(([, value]) =>
+            value.some((m) => m.type !== "none"),
           ),
         ),
       };
--- a/extensions/ql-vscode/src/query-testing/test-ui.ts
+++ b/extensions/ql-vscode/src/query-testing/test-ui.ts
@@ -1,23 +1,9 @@
-import {
-  TestHub,
-  TestController,
-  TestAdapter,
-  TestRunStartedEvent,
-  TestRunFinishedEvent,
-  TestEvent,
-  TestSuiteEvent,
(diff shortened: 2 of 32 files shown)


extensions/ql-vscode/src/query-testing/test-ui.ts
```diff
@@ -29,7 +15,7 @@ class QLTestListener extends DisposableObject {
     this.push(adapter.testStates(this.onTestStatesEvent, this));
   }
 
-  private onTestStatesEvent(_e: VSCodeTestEvent): void {
+  private onTestStatesEvent(): void {
```
robertbrignull, line 18: This is an odd case because this is essentially unused code, in that we're deliberately implementing this to do nothing. I think we can safely remove the argument here without losing any information or risking the plumbing breaking down, because we're the interface defined by the `testStates` function, and the `TestRunStartedEvent` type and friends are coming from `vscode-test-adapter-api` dependency.

Essentially we're not actually deleting any code here and it would be possible to recreate this by looking at the types from the `vscode-test-adapter-api` dependency.

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -211,8 +211,8 @@ export class ModelingStore extends DisposableObject {
         ...methods,
         // Keep all methods that are already modeled in some form in the state
         ...Object.fromEntries(
-          Object.entries(state.modeledMethods).filter(([_, value]) =>
```
robertbrignull, line 214: Ideally this shouldn't have generated a linting error. I feel like this array arg where the first element is `_` is morally the same as two parameters where the first is `_`. But the rule unfortunately triggered here so I rewrote this to get around it.

We could instead do
```typescript
const value = entry[1];
```
if you feel that would be clearer.

koesie10: I think the reason that this is giving an error is because the `_` is unnecessary: if you this to `.filter(([, value]) =>`, it should still work.

robertbrignull: Oh interesting, I didn't know you could do that and not specify a name for the first element at all. I'll try that.

Although, there's a chance that syntax is also a little unclear because it's only a single extra comma, and people may not have seen this before.

robertbrignull:
I think I'm going to close this PR. Nobody else appears to feel strongly that this is a helpful change, and one person voiced vague concern that it wasn't a helpful change. If we change our mind in the future this PR will still be here, or it's easy to redo the changes.

PR #3063, opened 2023-11-08
Add instructions for releasing from a branch other than main

robertbrignull:
Updates the instructions for doing an extension release.

The original prompt for this was to add instructions for releasing from a branch other than `main`. This hit some snags because of steps like "lock the main branch", but I think we can change the release process to just avoid all of that.

Major changes are:
- Can now release from a branch other than `main`, so we can do quick bug-fix releases based on the previous release and avoid the risk of releasing other unrelated changes.
  - Release from the release branch instead of merging the release branch into `main` and then releasing from `main`.
  - No locking or unlocking of branches. There are no issues with PRs being merged during the release process, because everything happens on a separate branch.
- The step for checking node modules is removed from the release process and instead added directly to the first responder duties. See internal PR.
- Removed a few places where we explain things git command line arguments. These made the instructions harder to read and should be unnecessary if people follow the instructions. If people do hit issues they can ask for help from other team members and also will learn how to fix the situation themselves for next time.

## Checklist

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -17,8 +17,7 @@
 1. Run the ["Run CLI tests" workflow](https://github.com/github/vscode-codeql/actions/workflows/cli-test.yml) and make sure the tests are green.
     * You can skip this step if you are releasing from `main` and there were no merges since the most recent daily scheduled run of this workflow.
 1. Double-check the `CHANGELOG.md` contains all desired change comments and has the version to be released with date at the top.
-    * If releasing from `main`, go through PRs that have been merged since the previous release and make sure they are properly accounted for.
-    * If doing a minimal bug-fix release, the changelog will only contain the changes you're intending to release.
+    * Go through PRs that have been merged since the previous release and make sure they are properly accounted for.
     * Make sure all changelog entries have links back to their PR(s) if appropriate.
 1. Double-check that the extension `package.json` and `package-lock.json` have the version you intend to release. If you are doing a patch release (as opposed to minor or major version) this should already be correct.
 1. Commit any changes made during steps 4 and 5 with a commit message the same as the branch name (e.g. `v1.3.6`).
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -11,6 +11,9 @@
     * For a regular scheduled release this branch will be based on latest `main`.
     * To do a minimal bug-fix release, base the release branch on the tag from the most recent release and then add only the changes you want to release.
         * Choose this option if you want to release a specific set of changes (e.g. a bug fix) and don't want to incur extra risk by including other changes that have been merged to the `main` branch.
+        ```bash
+        git checkout -b <new_release_branch> <previous_release_tag>
+        ```
 1. Run the ["Run CLI tests" workflow](https://github.com/github/vscode-codeql/actions/workflows/cli-test.yml) and make sure the tests are green.
     * You can skip this step if you are releasing from `main` and there were no merges since the most recent daily scheduled run of this workflow.
 1. Double-check the `CHANGELOG.md` contains all desired change comments and has the version to be released with date at the top.
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -9,6 +9,7 @@
     * Accumulation of many changes, none of which are individually big enough to warrant a minor bump, but which together are. This does not include changes which are purely internal to the extension, such as refactoring, or which are only available behind a feature flag.
 1. Create a release branch named after the new version (e.g. `v1.3.6`):
     * For a regular scheduled release this branch will be based on latest `main`.
+        * Make sure your local copy of `main` is up to date so you are including all changes.    
     * To do a minimal bug-fix release, base the release branch on the tag from the most recent release and then add only the changes you want to release.
         * Choose this option if you want to release a specific set of changes (e.g. a bug fix) and don't want to incur extra risk by including other changes that have been merged to the `main` branch.
 
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -11,9 +11,11 @@
     * For a regular scheduled release this branch will be based on latest `main`.
     * To do a minimal bug-fix release, base the release branch on the tag from the most recent release and then add only the changes you want to release.
         * Choose this option if you want to release a specific set of changes (e.g. a bug fix) and don't want to incur extra risk by including other changes that have been merged to the `main` branch.
+
         ```bash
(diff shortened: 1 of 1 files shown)


docs/releasing.md
```diff
@@ -1,80 +1,52 @@
...
+1. Determine the new version number. We default to increasing the patch version number, but make our own judgement about whether a change is big enough to warrant a minor version bump. Common reasons for a minor bump could include:
+    * Making substantial new features available to all users. This can include lifting a feature flag.
+    * Breakage in compatibility with recent versions of the CLI.
+    * Minimum required version of VS Code is increased.
+    * New telemetry events are added.
+    * Deprecation or removal of commands.
+    * Accumulation of many changes, none of which are individually big enough to warrant a minor bump, but which together are. This does not include changes which are purely internal to the extension, such as refactoring, or which are only available behind a feature flag.
+1. Create a release branch named after the new version (e.g. `v1.3.6`):
+    * For a regular scheduled release this branch will be based on latest `main`.
+    * To do a minimal bug-fix release, base the release branch on the tag from the most recent release and then add only the changes you want to release.
```
charisk, line 12: Maybe have a sample git command here? E.g. `git checkout -b <branch_name> <tag_name>`

docs/releasing.md
```diff
@@ -1,80 +1,52 @@
...
-    * Select "Lock branch"
-    * Click "Save changes"
-1. Ensure that no PRs have been merged since the release PR that you merged. If there were, you might need to unlock `main` temporarily and update the CHANGELOG again.
-1. Build the extension `npm run build` and install it on your VS Code using "Install from VSIX".
+1. Commit any changes made during steps 4 and 5 with a commit message the same as the branch name (e.g. `v1.3.6`).
+1. Open a PR for this release.
+    * The PR diff should contain:
+        * Any missing bits from steps 4 and 5. Most of the time, this will just be updating `CHANGELOG.md` with today's date.
+        * If releasing from a branch other than `main`, this PR will also contain the extension changes being released.
+1. Build the extension using `npm run build` and install it on your VS Code using "Install from VSIX".
```
charisk, line 26: Do we no longer need to lock the branch we're releasing from?

charisk (commented):
Thanks for thinking about this! Great that our release process/tools mostly already work for this. 

You've added instructions about "minimal bug-fix releases" as special cases to the general release steps. Since we almost never do this, I would prefer to avoid that mental load when I'm in the "normal" release process (which is most of the time). Would having a special section for doing "minimal bug-fix releases" be to awkward?

docs/releasing.md
```diff
@@ -1,80 +1,52 @@
...
     * If there is an authentication failure when publishing, be sure to check that the authentication keys haven't expired. See below.
-1. Approve the deployments of the correct Release workflow. This will automatically publish to Open VSX and VS Code Marketplace.
-1. Go to the draft GitHub release in [the releases tab of the repository](https://github.com/github/vscode-codeql/releases), click 'Edit', add some summary description, and publish it.
-1. Confirm the new release is marked as the latest release at <https://github.com/github/vscode-codeql/releases>.
+1. Go to the draft GitHub release in [the releases page](https://github.com/github/vscode-codeql/releases), click 'Edit', add some summary description, and publish it.
+1. Confirm the new release is marked as the latest release.
 1. If documentation changes need to be published, notify documentation team that release has been made.
-1. Review and merge the version bump PR that is automatically created by Actions.
+1. Review and merge the version bump PR that is automatically created by the Release workflow.
```
shati-patel, line 49: You've probably thought about this, so I assume the version bump PRs still work as normal for a "bug fix" release? Or do we risk messing up the changelog when auto-adding [`## [UNRELEASED]`](https://github.com/github/vscode-codeql/pull/3056/files) to the top of the changelog? 🤔

robertbrignull: I admit I'd not thought directly about that part, but I think it will still work out. The reason is that although the code we release is from a branch (which may not include all the changes from `main`), as part of the process we merge that into `main`. Then the version bump happening on the default branch is the correct behaviour still.

This change in release process isn't meant to give us multiple release branches that we could release from independently. You can still only apply changes against the previous release, but the choice is whether to include everything currently on `main` or only a subset.

docs/releasing.md
robertbrignull: Yep, can include a git command here.

docs/releasing.md
robertbrignull: I don't believe so. Please do say if you think I'm wrong.

One branch to consider is the release branch. I don't believe there's a need to lock this branch as it's highly unlikely for another developer to push anything to it while you're doing the release.

I also don't believe it's necessary to lock the `main` branch. It shouldn't be a problem if anything is pushed to main while the release prep is happening. Those changes won't be included in the current release but that's ok, and they'll be included in the next release instead.

Thinking about this did make me realise that there could be conflicts in the changelog file merging the release PR, if any new PRs have been merged. But this solvable, and in this case the new entries should be placed at the top, above the section for the current release, so they will be marked as unreleased and ready to be included in the next release. I'll include a note in the instructions about this.

I'd have to check to make sure but I expect the workflow that updates the changelog after a release should continue to work even if more things have been pushed, so long as people stick to the right format.

robertbrignull:
> You've added instructions about "minimal bug-fix releases" as special cases to the general release steps. Since we almost never do this, I would prefer to avoid that mental load when I'm in the "normal" release process (which is most of the time). Would having a special section for doing "minimal bug-fix releases" be to awkward?

I'm not sure about a special section because the instructions are identical everywhere except when making the new branch, or checkout the changelog.

I can remove the special cases from the changelog section. It can just say "Go through PRs that have been merged since the previous release" and that's correct for both cases. I wanted to try to be as precise as possible, but if that makes it harder to read then we can remove the special cases.

With this, the only mention of the "minimal bug-fix releases" is in the step for creating the release branch and choosing which branch/tag to base it on. Is that ok? I think that's likely the smallest disruption we can get from mentioning this option.

docs/releasing.md
charisk: Ah yeah we're always releasing from a branch now so it makes sense not to have locks! Agree about conflicts but that's manageable.

charisk (approved):
Thanks for updating, this LGTM! 

I think since the branching is only in one place it's as optimal as it can be. We can also test it in the next release and make any improvements we can think of then.

robertbrignull:
Yes, we can always continue to make improvements as we go.

shati-patel (approved):
Looks good to me too 😎

PR #3066, opened 2023-11-13
Add origin metadata to database items

koesie10:
This adds an origin property to the database item to store the source of the database, for example GitHub or an internet URL. This will be used to automatically check for updates to GitHub-downloaded databases in the future.

## Checklist


extensions/ql-vscode/src/databases/config/db-config.ts
```diff
@@ -88,6 +90,7 @@ export interface LocalDatabase {
   name: string;
   dateAdded: number;
   language: string;
+  source: DatabaseSource;
```
robertbrignull, line 93: Is the idea that this doesn't need to be an optional field because the UI currently doesn't allow you to add local databases to this config? This is the config for the new, combined databases panel, right?

extensions/ql-vscode/src/databases/database-fetcher.ts
```diff
@@ -553,10 +571,18 @@ export async function convertGithubNwoToDatabaseUrl(
...
+    );
+    if (!databaseForLanguage) {
+      throw new Error(`No database found for language '${language}'`);
+    }
+
     return {
       databaseUrl: `https://api.github.com/repos/${owner}/${repo}/code-scanning/codeql/databases/${language}`,
       owner,
       name: repo,
+      commitOid: databaseForLanguage.commit_oid,
```
robertbrignull, line 585: It's worth noting that this is the commit oid at the point of checking the database info and it might have changed by the point that we download the database contents. However the download should happen immediately and I'm not sure we can do better anyway since there isn't an API endpoint that returns both the metadata and the contents.

robertbrignull (commented):
A couple of points I just want to clarify, but it seems ok.

adityasharad:
Passing suggestion: if you're concerned about the potential confusion around source or provenance, perhaps "origin"?

extensions/ql-vscode/src/databases/config/db-config.ts
koesie10: Yes, exactly. This type isn't actually used anywhere, but once we start using it, we should have the source for every database that is added. This is the same as the `language` and `dateAdded`, both of which are also optional in the `FullDatabaseOptions`, but required in this type.

extensions/ql-vscode/src/databases/database-fetcher.ts
koesie10: This is something we are aware of. The chance of this race condition happening is unlikely, but if it does, it's also not problematic. The only reason the `commit_oid` will be used is for checking updates to the database. If we run into this race condition, it's probably pretty likely that it's an active repository and we would have new commits coming in shortly after anyways.

koesie10:
> Passing suggestion: if you're concerned about the potential confusion around source or provenance, perhaps "origin"?

Thanks! I think that's a better name for this.

robertbrignull (approved):
LGTM

PR #3071, opened 2023-11-14
Prompt user for database download on startup

koesie10:
This adds a prompt for the user to download a CodeQL database when opening a GitHub repository. This is currently behind the `codeQL.githubDatabase.enable` feature flag, but this setting is exposed in the VS Code settings UI. This is necessary because `updateValue` on a setting doesn't do anything unless it's listed in the `package.json`.

## Checklist

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -421,8 +421,33 @@
       },
       {
         "type": "object",
-        "title": "Log insights",
+        "title": "GitHub Databases",
         "order": 8,
+        "properties": {
+          "codeQL.githubDatabase.enable": {
+            "type": "boolean",
+            "default": false,
+            "markdownDescription": "Enable automatic detection of GitHub databases."
+          },
+          "codeQL.githubDatabase.download": {
+            "type": "string",
+            "default": "ask",
+            "enum": [
+              "ask",
+              "never"
+            ],
+            "enumDescriptions": [
+              "Ask to download a GitHub database when a workspace is opened.",
+              "Never download a GitHub databases when a workspace is opened."
+            ],
+            "description": "Ask to download a GitHub database when a workspace is opened."
+          }
+        }
+      },
+      {
+        "type": "object",
+        "title": "Log insights",
+        "order": 9,
         "properties": {
           "codeQL.logInsights.joinOrderWarningThreshold": {
             "type": "number",
@@ -436,7 +461,7 @@
       {
         "type": "object",
         "title": "Telemetry",
-        "order": 9,
+        "order": 10,
         "properties": {
           "codeQL.telemetry.enableTelemetry": {
             "type": "boolean",
--- a/extensions/ql-vscode/src/databases/github-database-module.ts
+++ b/extensions/ql-vscode/src/databases/github-database-module.ts
@@ -6,7 +6,7 @@ import { asError, getErrorMessage } from "../common/helpers-pure";
 import {
   CodeqlDatabase,
   findGitHubDatabasesForRepository,
-  promptGitHubDatabaseDownload,
+  promptAndDownloadGitHubDatabase,
 } from "./github-database-prompt";
 import {
   GitHubDatabaseConfig,
@@ -124,7 +124,7 @@ export class GithubDatabaseModule extends DisposableObject {
       return;
     }
 
-    void promptGitHubDatabaseDownload(
+    await promptAndDownloadGitHubDatabase(
       octokit,
       githubRepository.owner,
       githubRepository.name,
--- a/extensions/ql-vscode/src/databases/github-database-module.ts
+++ b/extensions/ql-vscode/src/databases/github-database-module.ts
@@ -2,15 +2,47 @@ import { DisposableObject } from "../common/disposable-object";
 import { App } from "../common/app";
 import { findGitHubRepositoryForWorkspace } from "./github-repository-finder";
 import { redactableError } from "../common/errors";
-import { asError } from "../common/helpers-pure";
+import { asError, getErrorMessage } from "../common/helpers-pure";
+import {
+  CodeqlDatabase,
+  findGitHubDatabasesForRepository,
+  promptGitHubDatabaseDownload,
+} from "./github-database-prompt";
+import {
+  GitHubDatabaseConfig,
+  GitHubDatabaseConfigListener,
+  isCanary,
+} from "../config";
+import { AppOctokit } from "../common/octokit";
+import { DatabaseManager } from "./local-databases";
+import { CodeQLCliServer } from "../codeql-cli/cli";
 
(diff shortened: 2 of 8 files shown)


extensions/ql-vscode/src/databases/github-database-prompt.ts
```diff
@@ -0,0 +1,137 @@
...
+          languages,
+        )} CodeQL databases.`;
+
+  const connectMessage =
+    databases.length === 1
+      ? `Connect to GitHub and download the existing database?`
+      : `Connect to GitHub and download any existing databases?`;
+
+  const answer = await showNeverAskAgainDialog(
+    `${databasesMessage} ${connectMessage}`,
```
starcke, line 64: Any specific reason why this string is split in two? It seems that the condition is the same in both cases?

extensions/ql-vscode/src/databases/github-database-module.ts
```diff
@@ -42,8 +82,58 @@ export class GithubDatabaseModule extends DisposableObject {
...
-    void this.app.logger.log(
-      `Found GitHub repository for workspace: '${githubRepository.owner}/${githubRepository.name}'`,
+
+    const hasExistingDatabase = this.databaseManager.databaseItems.some(
+      (db) =>
+        db.origin?.type === "github" &&
+        db.origin.repository ===
+          `${githubRepository.owner}/${githubRepository.name}`,
+    );
+    if (hasExistingDatabase) {
```
starcke, line 92: Is this here where we later want to do an upgrade check?

extensions/ql-vscode/src/databases/github-database-prompt.ts
```diff
@@ -0,0 +1,137 @@
...
+  databaseManager: DatabaseManager,
+  storagePath: string,
+  cliServer: CodeQLCliServer,
+  commandManager: AppCommandManager,
+): Promise<void> {
+  const languages = databases.map((database) => database.language);
+
+  const databasesMessage =
+    databases.length === 1
+      ? `This repository has an origin (GitHub) that has a ${getLanguageDisplayName(
```
starcke, line 51: It feels like we could just `joinLanguages` in both cases because you made it handle single element lists just fine.

extensions/ql-vscode/test/vscode-tests/no-workspace/databases/github-database-prompt.test.ts
```diff
@@ -0,0 +1,267 @@
...
+        config,
+        databaseManager,
+        storagePath,
+        cliServer,
+        commandManager,
+      );
+
+      expect(config.setDownload).toHaveBeenCalledTimes(1);
+      expect(config.setDownload).toHaveBeenCalledWith("never");
+    });
```
starcke, line 180: Do you need a test that if the config says "never" then there is no prompt at all?

starcke (approved):
Looks generally good, just a few comments and questions.

extensions/ql-vscode/package.json
```diff
@@ -421,8 +421,33 @@
...
+          "codeQL.githubDatabase.enable": {
+            "type": "boolean",
+            "default": false,
+            "markdownDescription": "Enable automatic detection of GitHub databases."
+          },
+          "codeQL.githubDatabase.download": {
+            "type": "string",
+            "default": "ask",
+            "enum": [
+              "ask",
```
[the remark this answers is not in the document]
koesie10: I think for now we should just keep these options. James said "My preference for now is that we always show a prompt to get approval from the user whenever there is a database available to download, if that works with the other settings. If we get feedback that this is annoying then we can consider adding a setting to automate downloads in the future." This was mostly in reference to updates, but I think it's even more likely we want to prompt the user for initial downloads.

extensions/ql-vscode/src/databases/github-database-module.ts
```diff
@@ -42,8 +82,58 @@ export class GithubDatabaseModule extends DisposableObject {
...
+      );
+
+      return;
+    }
+
+    if (databases.length === 0) {
+      return;
+    }
+
+    void promptGitHubDatabaseDownload(
```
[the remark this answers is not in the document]
koesie10: I've renamed the function to `promptAndDownloadGitHubDatabase` to make it clearer that this also downloads the database.

extensions/ql-vscode/src/databases/github-database-prompt.ts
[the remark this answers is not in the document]
koesie10: I think "Java CodeQL database" makes more sense when looking at the situation where there are multiple databases:
- "Java and C# CodeQL databases"
- "CodeQL Java and C# databases"

It's a CodeQL database, so I wouldn't expect "Java and C#" to split up that concept.

extensions/ql-vscode/src/databases/github-database-prompt.ts
koesie10: Good point, I've merged these two messages.

extensions/ql-vscode/src/databases/github-database-module.ts
koesie10: Yes, although the upgrade check will happen after retrieving the databases.

extensions/ql-vscode/src/databases/github-database-prompt.ts
koesie10: We could do `joinLanguages` here, but I didn't feel like that makes sense when we know there's just a single language. I think this makes it clearer that there's only a single element. The actually important part is the text around it anyway.

extensions/ql-vscode/test/vscode-tests/no-workspace/databases/github-database-prompt.test.ts
koesie10: That's quite hard to test because it's in the `GitHubDatabaseModule` which starts an `async` task without awaiting it. I also think it's quite a simple check, so I don't feel like implementing a very complex test for testing this makes a ton of sense.

extensions/ql-vscode/src/databases/github-database-module.ts
```diff
@@ -42,8 +82,58 @@ export class GithubDatabaseModule extends DisposableObject {
...
+      );
+    } catch (e) {
+      this.app.telemetry?.sendError(
+        redactableError(
+          asError(e),
+        )`Failed to prompt for GitHub database download`,
+      );
+
+      void this.app.logger.log(
+        `Failed to find GitHub databases for repository: ${getErrorMessage(e)}`,
```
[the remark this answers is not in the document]
starcke: Ah yeah. The nothing found is the `databases.length === 0` down below.

starcke (approved):
:+1:

PR #3072, opened 2023-11-14
Use credentials for database download in non-canary mode

koesie10:
We want users to be able to download databases from private/internal repositories without using canary mode. This will change the prompt to ask for credentials in non-canary mode as well. This means there are now 2 different prompts:

- This repository has an origin (GitHub) that may have one or more CodeQL databases. Connect to GitHub and download any existing databases?
  - Connect
  - No
  - Never
- This repository has an origin (GitHub) that has Ruby and JavaScript CodeQL databases. Download any existing databases from GitHub?
  - Download
  - No
  - Never

When you click on "Connect" in the first notification, we'll prompt for credentials and then for the language (if necessary). In the second one, we'll immediately prompt for the language.

Based on #3071

## Checklist


extensions/ql-vscode/src/databases/github-database-module.ts
```diff
@@ -121,15 +139,29 @@ export class GithubDatabaseModule extends DisposableObject {
     }
 
     if (databases.length === 0) {
+      // If the user didn't have an access token, they have already been prompted,
+      // so we should give feedback.
+      if (!hasAccessToken) {
```
starcke, line 144: I think the behavior is correct, but I found the logic a bit hard to follow. It also seems a bit odd that there is a database prompt (the one to connect) in here instead of in `github-database-prompt.ts`.

I also wonder, what if the repo is public - do we then ask for a non-needed token?

extensions/ql-vscode/src/databases/github-database-prompt.ts
```diff
@@ -30,47 +30,60 @@ export async function findGitHubDatabasesForRepository(
...
   }
 
+  return true;
+}
+
+/**
+ * Download a database from GitHub by asking the user for a language and then
+ * downloading the database for that language.
+ */
+export async function downloadDatabaseFromGitHub(
```
starcke, line 75: Ah now it became a separate function :+1: Should it move to `github-database-module.ts` as it doesnt implement any prompting itself?

starcke (commented):
Mostly looks good, although it would be good to see if the logic could be made a bit simpler. I also wonder about public repos.

extensions/ql-vscode/src/databases/github-database-prompt.ts
koesie10: I'll rename the file to `github-database-download`, then it makes sense to have both of these functions in the same file. This will match with #3079 where I introduce these two functions for updating databases.

extensions/ql-vscode/src/databases/github-database-module.ts
koesie10: The prompt for connecting is defined in this file because it is common to both downloading and updating databases. This should make sense in #3079.

If the repo is public, we do ask for an unnecessary token. I can change this to always make a request first, and then only prompt if we get a 404. However, that will make this logic even harder to follow.

starcke: > However, that will make this logic even harder to follow.

That is probably right, although I am not super sure. I think what makes it complicated
is that we dont want to ask the user to connect and then ask them again to download the DB. If we asked
both questions then it could probably be written something like this:
```
databases = getDatabases(current_credentials)

if (databases.length === 0 && no_credentials) {
  // Ask the user for credentials and try again
  databases = getDatabases(ask_for_credentials)
}

downloadDatabaseFromGitHub(databases)
```

What do you think? How much do we want to prevent the second question?

koesie10: I think I've found a somewhat simpler solution which will make this easier to follow and also skip the prompt for credentials when the repo is public. I'll create a new function `listDatabases` which will:
1. Try retrieving the databases, either with or without credentials depending on what is available
2. If that fails with a 404 and they don't have an access token, ask the user whether they want to connect to GitHub
3. If they do, prompt for credentials and retry the request

This function can then return a result like `{ databases: CodeqlDatabase[], promptedForCredentials: boolean }`. The logic in the module will then be a little bit simpler since the name is `promptedForCredentials` instead of `hasAccessToken`, more closely matching what the intention is.

starcke: That sounds good, thanks!

koesie10: I've just pushed this change, the logic for that is mostly inside `github-database-api.ts` and the `listDatabases` function.

starcke: That looks better and also covers the public repo case. Thanks!

starcke (approved):
:+1:

PR #3086, opened 2023-11-23
Fix tests with `workspaceContains:.git` activation event

koesie10:
It seems like changes to the CLI path setting are not picked up by the extension when it is already activated. This is probably because the `workspace.onDidChangeConfiguration` event is not fired when the update is made programmatically.

This fixes it by manually firing the event so that all listeners (CLI server, query server, IDE server) can pick up the change.

Some other small changes to the tests were necessary to make everything work (mostly with retries), but this should now work properly.

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/cli-integration/databases/database-fetcher.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/cli-integration/databases/database-fetcher.test.ts
@@ -15,7 +15,7 @@ import {
   storagePath,
 } from "../../global.helper";
 import { createMockCommandManager } from "../../../__mocks__/commandsMock";
-import { ensureDir, remove } from "fs-extra";
+import { remove } from "fs-extra";
 
 /**
  * Run various integration tests for databases
@@ -43,7 +43,6 @@ describe("database-fetcher", () => {
   afterEach(async () => {
     await cleanDatabases(databaseManager);
     await remove(storagePath);
-    await ensureDir(storagePath);
   });
 
   describe("importArchiveDatabase", () => {
--- a/extensions/ql-vscode/test/vscode-tests/cli-integration/databases/database-fetcher.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/cli-integration/databases/database-fetcher.test.ts
@@ -15,6 +15,7 @@ import {
   storagePath,
 } from "../../global.helper";
 import { createMockCommandManager } from "../../../__mocks__/commandsMock";
+import { ensureDir, remove } from "fs-extra";
 
 /**
  * Run various integration tests for databases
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/test/vscode-tests/cli-integration/jest.setup.ts
```diff
@@ -31,6 +31,25 @@ beforeAll(async () => {
...
+
+  // Activate the extension
+  const extension = await getActivatedExtension();
+
+  if (process.env.CLI_VERSION && process.env.CLI_VERSION !== "nightly") {
+    const cliVersion = await extension.cliServer.getVersion();
+
+    if (cliVersion.compare(process.env.CLI_VERSION) !== 0) {
+      // This calls the private `updateConfiguration` method in the `ConfigListener`
+      // It seems like the CUSTOM_CODEQL_PATH_SETTING.updateValue() call in
```
charisk, line 43: Apologies for the ignorance, but since this is a problem with  `CUSTOM_CODEQL_PATH_SETTING.updateValue()`, why not trigger an update to the config after that instead of here? 

Also would it make sense to have a public `forceUpdateConfiguration()` (with comments etc. to say that ideally it would only be used by tests) instead of these castings?

extensions/ql-vscode/test/vscode-tests/cli-integration/databases/database-fetcher.test.ts
```diff
@@ -41,6 +42,8 @@ describe("database-fetcher", () => {
 
   afterEach(async () => {
     await cleanDatabases(databaseManager);
+    await remove(storagePath);
+    await ensureDir(storagePath);
```
charisk, line 46: Should this be in `beforeEach`? Create the path at the start of the test, remove it in the end.

charisk (commented):
Thanks for fixing, some questions from me.

extensions/ql-vscode/test/vscode-tests/cli-integration/jest.setup.ts
koesie10: There is a possibility that the extension is not yet activated when we call `updateValue`, so we wouldn't be able to call this method there. By moving it here, we ensure that the extension is activated and the method can be called.

Adding a public `forceUpdateConfiguration` method makes sense, I've added that and added some comments to say that it should only be used in tests.

extensions/ql-vscode/test/vscode-tests/cli-integration/databases/database-fetcher.test.ts
koesie10: It seems like the `ensureDir` call here is not necessary since [the directory is created in the database fetcher](https://github.com/github/vscode-codeql/blob/636f8f1b5f8a12dfb331dd8036392eaf3b47ecb8/extensions/ql-vscode/src/databases/database-fetcher.ts#L359), so I've removed it.

charisk (approved):
Thanks, LGTM!

PR #3107, opened 2023-12-06
Fix loading message for compare view

koesie10:
The loading message for the compare view was never shown because the `comparison` was never `undefined`/`null`. This changes the comparison to be `null` on load to make sure that we explicitly handle this case rather than relying on an empty comparison message. This helps when there is a large number of results and loading them takes a while.

## Checklist


charisk (approved):
LGTM! Do we need to update the CHANGELOG?

koesie10:
> LGTM! Do we need to update the CHANGELOG?

Good idea, I'll add this fix to the CHANGELOG.

PR #3114, opened 2023-12-06
Change compare view default result set name

koesie10:
This changes the default result set that is selected in the compare view to the same as the default result set in the results view. This will select `#select` by default for most queries, rather than the first result set, which could be `edges` or `nodes` for some queries. If `#select` doesn't exist, the behavior is the same as before.

## Checklist


charisk (approved):
Does this need a CHANGELOG entry?

koesie10:
> Does this need a CHANGELOG entry?

I don't think so, the selected result set was pretty random, so I don't think users were dependent on it.

PR #3122, opened 2023-12-11
Run `npm upgrade`

koesie10:
This upgrades all dependencies to their latest patch/minor versions since this is not handled by Dependabot. There were a few files in which I had to fix minor incompatible changes, but I've tested the extension and it still works.

## Checklist


extensions/ql-vscode/src/view/common/DeterminateProgressRing.tsx
```diff
@@ -39,7 +39,7 @@ export const DeterminateProgressRing = ({ percent }: Props) => (
     aria-valuemax={100}
     aria-valuenow={percent}
   >
-    <svg className="progress" viewBox="0 0 16 16">
+    <svg className="progress" viewBox="0 0 16 16" role="presentation">
```
norascheuch, line 42: Just out of curiosity, why a role and why 'presentation'?

extensions/ql-vscode/src/view/results/locations/NonClickableLocation.tsx
```diff
@@ -13,5 +13,6 @@ export function NonClickableLocation({ msg, locationHint }: Props) {
   if (msg === undefined) {
     return null;
   }
+  // eslint-disable-next-line github/a11y-no-title-attribute
```
norascheuch, line 16: Is this something we should fix in a different PR?

norascheuch (approved):
👾

extensions/ql-vscode/src/view/common/DeterminateProgressRing.tsx
koesie10: This fixes the [`github/a11y-svg-has-accessible-name`](https://github.com/github/eslint-plugin-github/blob/main/docs/rules/a11y-svg-has-accessible-name.md) ESLint rule. The `role="presentation"` basically just means that this content shouldn't really be read by screen readers and that they can ignore it. We don't need to add a title to the SVG because it already has `role="progressbar"`, so can already be recognized by screen readers.

extensions/ql-vscode/src/view/results/locations/NonClickableLocation.tsx
koesie10: I don't think this is important enough to fix. This would need design work etc. for a feature that is probably almost never used.

PR #3125, opened 2023-12-12
Avoid creating a multitoken when finding definitions

aeisenberg:
This will avoid showing a query-server popup on hovers. When someone does CTRL+hover on source code in a database archive, the definitions provider is triggered. The first time this is run, there will be a query that is invoked in order to find definitions. This can be a long process.

Previously, this query brought up a popup window that could be cancelled. However, this is confusing users since the query is automatically cancelled when the mouse hovers away from the element.

The downside of this PR is that even when find definitions is explicitly invoked, (using F12), there will still be no hover.

I think this is an improvement, but I am happy to discuss if others disagree.

## Checklist

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Avoid showing a popup when hovering over source elements in database source files. [#3125](https://github.com/github/vscode-codeql/pull/3125)
+
 ## 1.11.0 - 13 December 2023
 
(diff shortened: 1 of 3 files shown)


robertbrignull (approved):

robertbrignull:
> I think this is an improvement, but I am happy to discuss if others disagree.

I don't have strong opinions on what the behaviour should. I'm very happy to trust your judgement as you use this particular feature a lot more and are closer to the users. I think the code changes look fine.

aeisenberg:
Thanks. I'll add a changelog note and then merge. I'm assuming the two failing jobs are flakes.

PR #3127, opened 2023-12-12
Provide option to point ReleasesApiConsumer at nightly builds repo

robertbrignull:
Adds in the `codeQL.cli.channel` config option, as a way to point the `ReleasesApiConsumer` class at `dsp-testing/codeql-cli-nightlies` instead of at `github/codeql-cli-binaries`. This will be useful for the CodeQL team to test nightly builds of the CodeQL CLI.

When changing the cli channel, the intended behaviour is that it will change the installed distribution to whatever is the latest release of that channel. From reading the code I believe it should already behave this way. I can't find any checks of the semver of the currently installed extension vs the new extension, so I believe it relies on the releases always going forwards and never doing something like re-releasing an old version. Therefore there's no need for us to add any tracking of where the currently-installed extension came from.

I haven't tested this yet because I'm not in a position to do that today, but I will test it tomorrow when I'm more able.

When writing this I also spotted some ways we can make the `ReleasesApiConsumer` cleaner and fit in more with the style of the rest of the extension, but I'll save those changes for a separate PR.

## Checklist


koesie10 (commented):
Unfortunately, this doesn't quite work. There is a check that [all releases satisfy the semver version range `2.x`](https://github.com/github/vscode-codeql/blob/66c0714606b92c625f35ef7b60450a13e661c611/extensions/ql-vscode/src/codeql-cli/distribution.ts#L604), and nightly CLI releases are tagged like `codeql-bundle-20231212`, which are neither semver nor do they satisfy the semver range.

If we remove this semver check for nightly builds, then I think this should work. The update check seems to be based on the release ID, and doesn't even check that a release is newer when updating.

robertbrignull:
> There is a check that [all releases satisfy the semver version range `2.x`](https://github.com/github/vscode-codeql/blob/66c0714606b92c625f35ef7b60450a13e661c611/extensions/ql-vscode/src/codeql-cli/distribution.ts#L604), and nightly CLI releases are tagged like `codeql-bundle-20231212`, which are neither semver nor do they satisfy the semver range.

Good point. I agree I think we want to just remove that check when using nightly builds. We'll just take the latest nightly release regardless of its name.

> The update check seems to be based on the release ID, and doesn't even check that a release is newer when updating.

That's how I understood the code too. It uses the release ID to tell if we're on the exact same release already and therefore don't need to update. If the release IDs do not match then it always upgrades regardless of semver of the current and new release.

robertbrignull:
I've worked around the nightly releases not having semver tags, but it wasn't pretty 😢 

There are probably better ways of doing this. I'm also not hugely happy with the `ReleasesApiConsumer` class as a whole, but I'm resisting the urge to refacting it in this PR. I plan to do a second PR after this to make `ReleasesApiConsumer` a bit more modern and move it to its own file.

If this code works is it acceptable?

extensions/ql-vscode/src/codeql-cli/distribution.ts
```diff
@@ -587,10 +631,9 @@ export class ReleasesApiConsumer {
     });
     // Tag names must all be parsable to semvers due to the previous filtering step.
     const latestRelease = compatibleReleases.sort((a, b) => {
-      const versionComparison = semver.compare(
-        semver.parse(b.tag_name)!,
-        semver.parse(a.tag_name)!,
-      );
+      const versionComparison = orderBySemver
+        ? semver.compare(semver.parse(b.tag_name)!, semver.parse(a.tag_name)!)
+        : b.id - a.id;
```
robertbrignull, line 636: I wasn't sure about whether to order based on `id` or `createdAt`, but went with `id` because it's a number and having to parse date strings (which might be invalid) so we can compare them is a pain.

I also wondered about passing in a comparison function instead this `orderBySemver` boolean, but thought maybe that could wait for followup refactoring.

F4ST3R215:
what...? I'm done no..!!!

extensions/ql-vscode/src/codeql-cli/distribution.ts
```diff
@@ -475,19 +494,41 @@ class ExtensionSpecificDistributionManager {
...
+    } else if (this.config.channel === "nightly") {
+      return NIGHTLY_DISTRIBUTION_REPOSITORY_NAME;
+    } else {
+      return DEFAULT_DISTRIBUTION_REPOSITORY_NAME;
+    }
+  }
+
+  private usingNightlyReleases(): boolean {
+    return (
+      !this.config.ownerName &&
```
robertbrignull, line 526: I find it odd that you can set a custom owner name and repo name separately, and it'll maybe use a custom org name with a default value for repo name. I'm not sure why someone would want that, but that's what we have. Therefore we have to check if either is set, and we can't just combine our values together into one NWO.

extensions/ql-vscode/src/codeql-cli/distribution.ts
```diff
@@ -475,19 +494,41 @@ class ExtensionSpecificDistributionManager {
...
     return new ReleasesApiConsumer(
-      ownerName,
-      repositoryName,
+      this.distributionOwnerName(),
+      this.distributionRepositoryName(),
       this.config.personalAccessToken,
     );
   }
 
+  private distributionOwnerName(): string {
```
koesie10, line 504: Very optional: Since these methods don't have any side effects, it might be a bit nicer to make them getters so you don't need to call into them:

```suggestion
  private get distributionOwnerName(): string {
```

The same could be done for `distributionRepositoryName` and `usingNightlyReleases`.

extensions/ql-vscode/src/codeql-cli/distribution.ts
```diff
@@ -475,19 +494,41 @@ class ExtensionSpecificDistributionManager {
...
+    if (this.config.repositoryName) {
+      return this.config.repositoryName;
+    } else if (this.config.channel === "nightly") {
+      return NIGHTLY_DISTRIBUTION_REPOSITORY_NAME;
+    } else {
+      return DEFAULT_DISTRIBUTION_REPOSITORY_NAME;
+    }
+  }
+
+  private usingNightlyReleases(): boolean {
```
koesie10, line 524: Could this perhaps be based on matching the distribution owner and repository name to `dsp-testing/codeql-cli-nightlies`? That might be a bit easier and just changes the behavior if you manually set it to this repository before this change. That would have given an error before, and would now start working.

extensions/ql-vscode/src/codeql-cli/distribution.ts
```diff
@@ -560,7 +601,8 @@ export class ReleasesApiConsumer {
   }
 
   public async getLatestRelease(
-    versionRange: semver.Range,
+    versionRange: semver.Range | undefined,
+    orderBySemver = true,
```
koesie10, line 605: Since you added this parameter here, there seem to be some call sites that are unintentionally setting `orderBySemver` instead of `includePrerelease` in the tests.

extensions/ql-vscode/src/config.ts
```diff
@@ -111,13 +112,16 @@ const DISTRIBUTION_CHANGE_SETTINGS = [
   PERSONAL_ACCESS_TOKEN_SETTING,
 ];
 
+export type CLIChannel = "released" | "nightly";
```
koesie10, line 115: Perhaps call it `stable` instead of `released`? Another alternative would be to just make `channel` optional and make it possible to be `undefined`.

koesie10 (commented):
General approach looks good to me, but I agree that we could refactor this code to be easier to understand and perhaps split up this file into a few separate files containing the different classes.

extensions/ql-vscode/src/codeql-cli/distribution.ts
robertbrignull: Sounds good to me. I'll change them all to use `get`

extensions/ql-vscode/src/codeql-cli/distribution.ts
robertbrignull: That's a good idea. I'll change this is to check if the value matches the default value.

extensions/ql-vscode/src/codeql-cli/distribution.ts
robertbrignull: Oops. Sorry I thought I got them all.

I wish I knew why "find referencs" often just ignores references in tests.

extensions/ql-vscode/src/config.ts
robertbrignull: I'm happy to change to "stable". To be honest I think that's a better name. I think I went with "released" because that terminology had been used in the internal issue, but now I look again I can't really see it.

I'd prefer to keep this enum with two values rather than being potentially undefined, just so it's clearer what the non-nightly option is.

robertbrignull:
I've pushed changes in response to @koesie10's comments and also added some tests to cover the new behaviour and hopefully fixed the existing tests.

extensions/ql-vscode/test/vscode-tests/no-workspace/codeql-cli/distribution.test.ts
```diff
@@ -98,15 +107,26 @@ describe("Releases API consumer", () => {
       }
     }
 
-    it("picked release has version with the highest precedence", async () => {
+    it("picked release is non-prerelease with with the highest semver", async () => {
```
koesie10, line 110:
```suggestion
    it("picked release is non-prerelease with the highest semver", async () => {
```

koesie10 (approved):
LGTM, thanks!

PR #3129, opened 2023-12-12
Create issue when CLI tests fail on the default branch

robertbrignull:
We'd like to know when the daily CLI tests fail on the default branch. This PR updates the workflow to create an issue when the workflow fails on the default branch.

To avoid creating lots of issues, it looks to see if an issue already exists with the `cli-test-failure` label, and if so it'll add a new comment to that issue. My idea is that we can use that issue to discuss and document whatever is causing the tests to fail, and then we'll close it when the tests are fixed. When the tests next break, they'll open a new issue and we can have a separate discussion of this new failure.

I'm taking inspiration from how it works on another github repo. More details in the internal issue.

## Checklist


.github/workflows/cli-test.yml
```diff
@@ -109,3 +109,40 @@ jobs:
...
+          # Try to find an existing open issue if there is one
+          ISSUE="$(gh issue list --repo "$GITHUB_REPOSITORY" --label "cli-test-failure" --state "open" --limit 1 --json number -q '.[0].number')"
+
+          if [[ -n "$ISSUE" ]]; then
+            echo "Found open issue number $ISSUE ($GITHUB_SERVER_URL/$GITHUB_REPOSITORY/issues/$ISSUE)"
+          else
+            echo "Did not find an open tracking issue. Creating one."
+
+            ISSUE_BODY="issue-body.md"
+            printf "CLI tests have failed on the default branch.\n\n@github/code-scanning-secexp-reviewers" > "$ISSUE_BODY"
```
shati-patel, line 135: Maybe this isn't relevant, but I was wondering who the issue author would be. Is it `github-actions` bot?

If so, I'm not sure that we'll get the team notification for `@github/code-scanning-secexp-reviewers` 🤔 (Since the "mentioner" would need to be part of the GitHub org to mention the team) 

We can try it out though and see what it looks like!

robertbrignull: Interesting point. I believe yes it'll be the `github-actions` bot. I wasn't aware that might block it from notifying us.

I suppose maybe the team mention might be unnecessary if anyone who wants these notifications subscribes to the label.

robertbrignull:
If there are no objections to the general approach I can do a test in a separate repo so I can iterate more easily and make sure it works.

shati-patel:
> If there are no objections to the general approach I can do a test in a separate repo so I can iterate more easily and make sure it works.

The general approach certainly looks fine to me!

robertbrignull:
I've now tested this issue in a separate repo and confirmed it makes issues correctly. E.g. https://github.com/dsp-testing/robertbrignull-ci-failure-issue/issues/2

Notifications could still be a problem, if it suppresses notifications coming from the actions bot. I suggest maybe we just try it out? If we're finding notifications to be an issue then we could try using a separate bot account for this, or look into other ways around it.

.github/workflows/cli-test.yml
```diff
@@ -109,3 +109,43 @@ jobs:
...
+            ISSUE="$(gh issue create --repo "$GITHUB_REPOSITORY" --label "cli-test-failure" --title "CLI test failure" --body-file "$ISSUE_BODY")"
+            # `gh issue create` returns the full issue URL, not just the number.
+            echo "Created issue with URL $ISSUE"
+          fi
+
+          COMMENT_FILE="comment.md"
+          RUN_URL=$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID
+          printf 'CLI test [%s](%s) failed on ref `%s`' "$GITHUB_RUN_ID" "$RUN_URL" "$GITHUB_REF" > "$COMMENT_FILE"
+
+          # `gh issue create` returns an issue URL, and `gh issue list | cut -f 1` returns an issue number.
```
koesie10, line 149: We don't seem to be using `cut -f 1`. Should this comment be updated for using `-q` in `gh issue list`?

koesie10 (approved):
LGTM

robertbrignull:
FYI, I'll wait to merge this until January, just to avoid it potentially breaking over the break.

PR #3141, opened 2023-12-14
Add a consistency check for the model editor

koesie10:
This adds an internal consistency check for the model editor. This catches cases where the model editor thinks something is supported, but the model editor queries return that it it not supported.

This would for example have caught https://github.com/github/codeql/pull/15089 without needing to manually check the query.

Screenshot 2023-12-14 at 12 09 40

## Checklist


charisk (approved):
LGTM!

I wouldn't even hide it behind a flag since it's only logging to the extension log. Those logs could be useful when debugging a user problem.

koesie10:
> I wouldn't even hide it behind a flag since it's only logging to the extension log. Those logs could be useful when debugging a user problem.

Good point, I've removed the feature flag. I ran into some incorrect checks for the Ruby model editor, so I've excluded `type` models from being considered as supported. I've also updated the message to reference the model editor since it could now be shown to users if there is a bug in the queries or the model editor.

charisk (approved):
🚢

PR #3150, opened 2023-12-18
Switch to `yauzl` in the archive filesystem provider

koesie10:
This switches the archive filesystem provider to `yauzl` instead of `unzipper` in preparation for removing `unzipper`. There should be no changes in behavior (except when Node > 18.15.0, in which case `unzipper` might actually read corrupted ZIP files).

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/unzip.ts
+++ b/extensions/ql-vscode/src/common/unzip.ts
@@ -1,0 +1,84 @@
+import { Entry as ZipEntry, open, Options as ZipOptions, ZipFile } from "yauzl";
+import { Readable } from "stream";
+
+// We can't use promisify because it picks up the wrong overload.
+export function openZip(
+  path: string,
+  options: ZipOptions = {},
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/common/unzip.ts
```diff
@@ -0,0 +1,84 @@
+import { Entry as ZipEntry, open, Options as ZipOptions, ZipFile } from "yauzl";
+import { Readable } from "stream";
+
```
charisk, line 3: What do you think about adding some tests for these functions? Using actual zip files we have in a test data dir.

charisk (approved):
Code changes look sensible to me.

extensions/ql-vscode/src/common/unzip.ts
koesie10: Good idea, I'll add those tests in a follow-up PR so I can continue working with these functions in other follow-up PRs.

PR #3156, opened 2023-12-19
 CodeQL model editor: support saving single/selected models

shati-patel:
Adds the ability to click specific rows in the model editor, and just save those instead of the entire library/model. See internal linked issue for more details. (Paired with @robertbrignull 🍐)

Probably easiest to review commit-by-commit, since there's quite a bit of boilerplate.

### 📹 Demo

https://github.com/github/vscode-codeql/assets/42641846/77c90ed6-6d2c-4bf6-bf2a-2db0a2143e42

### 📝 Notes

I'm opening the PR now to get some general feedback on the approach, since we can change the styling more easily. I already have a few concerns/questions:
- The "focused" property (from clicking "Review in editor") also temporarily highlights a row (until you click away). I've changed the "focused" background colour to an orange-y thing, but it might still be confusing:

  https://github.com/github/vscode-codeql/assets/42641846/b6a89583-c0e2-4cee-a501-42fc1157cf96

- I wonder if changing the background colour is obvious enough as a way to select a row. Perhaps an icon or text is more intuitive? 🤔 
- When you click anywhere in the method row (including on "View" or any of the dropdowns), it also selects/deselects the method row. Perhaps we can make just the method _name_ clickable? 🤔 (How would that work?)

## Checklist

  - ~A docs update will probably be helpful. I'll hold off until we confirm the exact behaviour!~ Not needed for now

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [UNRELEASED]
 
+- In the CodeQL model editor, you can now select individual method rows and save changes to only the selected rows, instead of having to save the entire library model. [#3156](https://github.com/github/vscode-codeql/pull/3156)
 - If you run a query without having selected a database, we show a more intuitive prompt to help you select a database. [#3214](https://github.com/github/vscode-codeql/pull/3214)
 - The UI for browsing and running CodeQL tests has moved to use VS Code's built-in test UI. This makes the CodeQL test UI more consistent with the test UIs for other languages.
   This change means that this extension no longer depends on the "Test Explorer UI" and "Test Adapter Converter" extensions. You can uninstall those two extensions if they are
--- a/extensions/ql-vscode/src/view/model-editor/LibraryRow.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/LibraryRow.tsx
@@ -71,11 +71,13 @@ export type LibraryRowProps = {
   methods: Method[];
   modeledMethodsMap: Record<string, ModeledMethod[]>;
   modifiedSignatures: Set<string>;
+  selectedSignatures: Set<string>;
   inProgressMethods: Set<string>;
   viewState: ModelEditorViewState;
   hideModeledMethods: boolean;
   revealedMethodSignature: string | null;
   onChange: (methodSignature: string, modeledMethods: ModeledMethod[]) => void;
+  onMethodClick: (methodSignature: string) => void;
   onSaveModelClick: (methodSignatures: string[]) => void;
   onGenerateFromLlmClick: (
     dependencyName: string,
@@ -92,11 +94,13 @@ export const LibraryRow = ({
   methods,
   modeledMethodsMap,
   modifiedSignatures,
+  selectedSignatures,
   inProgressMethods,
   viewState,
   hideModeledMethods,
   revealedMethodSignature,
   onChange,
+  onMethodClick,
   onSaveModelClick,
   onGenerateFromLlmClick,
   onStopGenerateFromLlmClick,
@@ -228,11 +232,13 @@ export const LibraryRow = ({
             methods={methods}
             modeledMethodsMap={modeledMethodsMap}
             modifiedSignatures={modifiedSignatures}
+            selectedSignatures={selectedSignatures}
             inProgressMethods={inProgressMethods}
             viewState={viewState}
             hideModeledMethods={hideModeledMethods}
             revealedMethodSignature={revealedMethodSignature}
             onChange={onChange}
+            onMethodClick={onMethodClick}
           />
           <SectionDivider />
           <ButtonsContainer>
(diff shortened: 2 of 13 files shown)


extensions/ql-vscode/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [UNRELEASED]
 
+- In the CodeQL model editor, you can now click individual method rows to save them, instead of having to save the entire library model. [#3156](https://github.com/github/vscode-codeql/pull/3156)
```
robertbrignull, line 5:
```suggestion
- In the CodeQL model editor, you can now select individual method rows and save changes to only the selected rows, instead of having to save the entire library model. [#3156](https://github.com/github/vscode-codeql/pull/3156)
```
The current wording makes it sound like it's the act of clicking itself that saves the method row, rather than selecting the row so you can save all selected rows.

Feel free to not take my suggestion though. Any wording is fine but I just wanted to point out the potential confusion.

robertbrignull (commented):
The code changes all look fine to me. When trying it out I did find it a little confusing to use, so I wonder if we should spend more time thinking about UX and get input from Asier before merging 🤔. I'm definitely feeling underqualified to say yes this is good to ship.

> I wonder if changing the background colour is obvious enough as a way to select a row. Perhaps an icon or text is more intuitive? 🤔

I agree I'm not sure if it's obvious enough right now. But maybe it'll become more obvious if other bits of the UI mentioning selections are changing when you click on rows.

> When you click anywhere in the method row (including on "View" or any of the dropdowns), it also selects/deselects the method row. Perhaps we can make just the method name clickable? 🤔 (How would that work?)

I also found this annoying. Particularly when using the dropdowns to edit a modeling it counts as "clicking on the row" and makes it selected or unselected. We can fix this particular behaviour with the dropdowns by calling `event.preventDefault()` in the on-click handlers of the dropdowns.

> The "focused" property (from clicking "Review in editor") also temporarily highlights a row (until you click away). I've changed the "focused" background colour to an orange-y thing, but it might still be confusing:

Another data point is, because both "focussed" and "selected" work by changing the background colour, you can't tell if both are active. So after focussing a row it's unclear what would be saved at that moment.

---

Another thought I had was when using I wanted a way to "deselect all". Would it be helpful to add a button for this, maybe in the top-bar or in each section?
Also what do you think about deselecting everything after saving?

shati-patel:
> The code changes all look fine to me. When trying it out I did find it a little confusing to use, so I wonder if we should spend more time thinking about UX and get input from Asier before merging 🤔. I'm definitely feeling underqualified to say yes this is good to ship.

Thanks for the review! And yes, don't worry, I don't intend to merge this until the UI is sorted out 😅 I've asked internally for design improvements 🎨 

> I also found this annoying. Particularly when using the dropdowns to edit a modeling it counts as "clicking on the row" and makes it selected or unselected. We can fix this particular behaviour with the dropdowns by calling `event.preventDefault()` in the on-click handlers of the dropdowns.

I haven't been able to do this yet 🤔 I'll poke around a bit, but might need some more help 🙇🏽‍♀️ 

> Another thought I had was when using I wanted a way to "deselect all". Would it be helpful to add a button for this, maybe in the top-bar or in each section?
> Also what do you think about deselecting everything after saving?

Both seem reasonable! I'll have a go at this too 🙌🏽

shati-patel:
- Fixed the bug where clicking on a button or dropdown (de)selected a row 
- Added a button to deselect all rows:

  https://github.com/github/vscode-codeql/assets/42641846/313a5a4a-717f-47c5-a237-cf684c13fe6f

shati-patel:
Having discussed with `@jf205`, this is good to ship from a product point of view (and we can improve on the design over time). See internal linked issue. 

Could I get a final review? 🙏🏽

extensions/ql-vscode/src/view/model-editor/LibraryRow.tsx
```diff
@@ -228,16 +232,18 @@ export const LibraryRow = ({
...
             inProgressMethods={inProgressMethods}
             viewState={viewState}
             hideModeledMethods={hideModeledMethods}
             revealedMethodSignature={revealedMethodSignature}
             onChange={onChange}
+            onMethodClick={onMethodClick}
           />
           <SectionDivider />
           <ButtonsContainer>
             <VSCodeButton onClick={handleSave} disabled={!hasUnsavedChanges}>
```
robertbrignull, line 245: I spotted one odd piece of UI behaviour. If you have unsaved changes then the button will be enabled, but it could be that you haven't selected methods with unsaved changes, so pressing the button does nothing. It could be better if the button only became enabled when there were unsaved changes in the selected methods.

However I suggest we can leave this as a followup, since I believe we're expecting to do a design pass on this at some point anyway.

robertbrignull (approved):
One observation that we can leave for a followup. Otherwise everything looks great.

extensions/ql-vscode/src/view/model-editor/LibraryRow.tsx
shati-patel: Thanks for the review! That's a good point 🦅 👀 

But I also agree that it can be a follow-up change instead, along with any other design/user feedback!

myronyuktaras:
**Еволюція мозку серця душі**_[]()document.pdf

_тарас миронюк @bruce

myronyuktaras:
**Еволюція мозку серця душі_комерційна ідея_**[]()![document.pdf](https://github.com/github/vscode-codeql/files/14156688/document.pd 
# - [ ] - Тарас Миронюк @shati-patel ~~~~

PR #3158, opened 2023-12-20
Do not reregister watchers on every file change for qltests

koesie10:
During the qltest discovery, we were recreating the watchers for qltests on every file change. This was causing the watchers to be recreated on each change, while there were no functional changes to the watchers themselves.

This commit moves the creation of the watchers to the constructor of the QLTestDiscovery class, and removes the creation of the watchers from the discover() method. The behavior should be the same.

## Checklist


robertbrignull (approved):
I also can't see any reason to recreate the watchers each time. As far as I can tell there isn't anything that could change between calls that would mean we need to remake them. The `this.workspaceFolder` is readonly and that's the only input.

cc @dbartol just in case he can see something / remembers something.

However, I think go ahead merging this when you're happy, and we can easily merge more changes to this area if we discover this has changed behaviour.

PR #3163, opened 2023-12-21
Do not use the CLI server to determine query pack language

koesie10:
This removes the use of the CLI `codeql resolve library-path` command to detect the language of a query pack. Instead, it uses the `qlpack.yml` file to determine the language. This is slightly less correct since it only works for `codeql/${language}-all` dependencies, but it is much faster and more reliable. It also doesn't result in the CLI server restarting for invalid query packs (such as in the `github/codeql` repository or in any workspaces containing it). Query pack discovery on `github/codeql` now takes less than 3 seconds rather than more than 11 seconds, and it also doesn't slow down other extension operations using the CLI server anymore.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/qlpack-language.ts
+++ b/extensions/ql-vscode/src/common/qlpack-language.ts
@@ -1,7 +1,5 @@
-import { load } from "js-yaml";
-import { readFile } from "fs-extra";
-import { QlPackFile } from "../packaging/qlpack-file";
 import { QueryLanguage } from "./query-language";
+import { loadQlpackFile } from "../packaging/qlpack-file-loader";
 
 /**
  * @param qlpackPath The path to the `qlpack.yml` or `codeql-pack.yml` file.
@@ -11,11 +9,9 @@ import { QueryLanguage } from "./query-language";
 export async function getQlPackLanguage(
   qlpackPath: string,
 ): Promise<QueryLanguage | undefined> {
-  const qlPack = load(await readFile(qlpackPath, "utf8")) as
-    | QlPackFile
-    | undefined;
+  const qlPack = await loadQlpackFile(qlpackPath);
   const dependencies = qlPack?.dependencies;
-  if (!dependencies || typeof dependencies !== "object") {
+  if (!dependencies) {
     return;
   }
 
--- a/extensions/ql-vscode/src/common/qlpack-language.ts
+++ b/extensions/ql-vscode/src/common/qlpack-language.ts
@@ -1,0 +1,30 @@
+import { load } from "js-yaml";
+import { readFile } from "fs-extra";
+import { QlPackFile } from "../packaging/qlpack-file";
+import { QueryLanguage } from "./query-language";
+
+/**
+ * @param qlpackPath The path to the `qlpack.yml` or `codeql-pack.yml` file.
+ * @return the language of the given qlpack file, or undefined if the file is
+ * not a valid qlpack file or does not contain exactly one language.
+ */
+export async function getQlPackLanguage(
+  qlpackPath: string,
+): Promise<QueryLanguage | undefined> {
+  const qlPack = load(await readFile(qlpackPath, "utf8")) as
+    | QlPackFile
+    | undefined;
+  const dependencies = qlPack?.dependencies;
+  if (!dependencies || typeof dependencies !== "object") {
+    return;
+  }
+
+  const matchingLanguages = Object.values(QueryLanguage).filter(
+    (language) => `codeql/${language}-all` in dependencies,
+  );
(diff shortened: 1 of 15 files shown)


aeisenberg (commented):
IIUC, this will only work if the language pack dependency is a direct dependency of the query pack. This will not find the language pack if it is a transitive dependencies.

I think the approach you have here is fine, but if the language is not found maybe it should fall back to the old approach.

koesie10:
> IIUC, this will only work if the language pack dependency is a direct dependency of the query pack. This will not find the language pack if it is a transitive dependencies.
> 
> I think the approach you have here is fine, but if the language is not found maybe it should fall back to the old approach.

That is correct, but I think it's actually fine to not support transitive dependencies for this class. The language here is only used for the queries panel, and I'd guess in most cases that users have the `codeql/${language}-all` as a direct dependency of their custom query packs. I'd like to avoid calling into the CLI server for this since it's quite a bit slower and if the query pack is invalid, the CLI server will crash and need to restart, which is really slow. Unless it's common for users to have the language packs as transitive dependencies, I think this approach should be good enough for this purpose.

extensions/ql-vscode/src/common/qlpack-language.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @param qlpackPath The path to the `qlpack.yml` or `codeql-pack.yml` file.
+ * @return the language of the given qlpack file, or undefined if the file is
+ * not a valid qlpack file or does not contain exactly one language.
+ */
+export async function getQlPackLanguage(
+  qlpackPath: string,
+): Promise<QueryLanguage | undefined> {
+  const qlPack = load(await readFile(qlpackPath, "utf8")) as
+    | QlPackFile
+    | undefined;
```
robertbrignull, line 16: I believe this function as it's currently written is safe, but also this type isn't entirely correct. As far as I can tell the `load` function can return anything, and if it can't parse the file contents as yaml then it returns a string of the entire file contents. It would be safer to use `unknown` and then have a type checking function to determine if the object is a qlpack object.

The danger of the current code is the manual checking of the `dependencies` object and not checking any of the other fields. If someone in the future does try to use on the other fields it could cause an error.

What I'm wondering about is whether it's worth the extra complexity to make it safer. I'm tempted to say yes, we should do everything we can to make the types as accurate as possible and avoid any unexpected behaviour. What do you think?

extensions/ql-vscode/src/local-queries/skeleton-query-wizard.ts
```diff
@@ -253,24 +252,12 @@ export class SkeletonQueryWizard {
       return undefined;
     }
 
-    const qlPack = load(await readFile(qlPackPath, "utf8")) as
```
robertbrignull, line 256: Ah, I didn't realise we already had this code for manually parsing a query pack. However I think we should still hold ourselves to high standards even though it's pre-existing code, because we're now using it in a new location that will get a lot more usage and run on all query packs in the workspace.

extensions/ql-vscode/src/queries-panel/query-pack-discovery.ts
```diff
@@ -71,32 +67,13 @@ export class QueryPackDiscovery extends FilePathDiscovery<QueryPack> {
...
-        true,
-      );
-    } catch (err) {
-      void extLogger.log(
-        `Query pack discovery failed to determine language for query pack: ${path}${EOL}Reason: ${getErrorMessage(
-          err,
-        )}`,
-      );
+      language = await getQlPackLanguage(path);
+    } catch (e) {
```
robertbrignull, line 73: We should probably log this exception. It won't be as noisy as it was before, but we should still highlight if this code does fail.

I think the only reasons for an error here are the file not existing (which shouldn't happen unless the file was just deleted) or there's a programming bug in `getQlPackLanguage`.

Side note: I'm surprised this isn't giving a warning for an unused variable.

robertbrignull (commented):
I don't have a problem with the idea of this PR, assuming @aeisenberg is happy. It is definitely faster doing this checking ourselves. Hopefully the restriction of not being able to check transitive dependencies will be ok in practice.

I do have a couple of small points about the `getQlPackLanguage` method where we can make it as safe as possible.

extensions/ql-vscode/src/common/qlpack-language.ts
koesie10: Yes, I agree. The easiest way to do this is just to generate a schema from the `QlpackFile` type and use that to validate the returned YAML, so I'll add that.

robertbrignull (approved):
Looks good from my point of view

PR #3172, opened 2023-12-21
Remove legacy messages

koesie10:
This removes all remaining usages of legacy messages. The first commit changes the most: It removes the `result` on a `QueryHistoryItem`, which was populated based on information that is already present in `CompletedQueryInfo` anyway. Old history items which only have the legacy result populated have not been created for at least 30 days now since the legacy query runner hasn't been used for quite a while now. Therefore, we can safely remove it.

## Checklist


charisk:
What is the significance of the 30 days? Is that the default query history item expiry time? I believe that's configurable so someone may have changed it. What would happen in that case (I've not looked at the code changes yet)?

koesie10:
> What is the significance of the 30 days? Is that the default query history item expiry time? I believe that's configurable so someone may have changed it. What would happen in that case (I've not looked at the code changes yet)?

Yes, exactly. However, I think this should actually be fine for all query history items created in the extension in at least the last year. When [we enabled the new query server by default](https://github.com/github/vscode-codeql/pull/1744), we [were already setting `successful` and `message`](https://github.com/github/vscode-codeql/blob/94ef752c0bee016e278a3505c23ad0c1f0addfe5/extensions/ql-vscode/src/legacy-query-server/run-queries.ts#L421-L423). So, even for those items it should work correctly. Only for items which are even older (created in extension versions before v1.7.1, https://github.com/github/vscode-codeql/pull/1498), would users now not be able to see the status or view results. However, the extension doesn't crash since [there's a default for these](https://github.com/github/vscode-codeql/pull/3172/files#diff-a1d5c2e9741c20af647a3b5e5403529f4737acc10eadd6b713d7ffe47a85e041R60-R61). They would also still be able to retrieve the BQRS and SARIF files.

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -159,6 +159,24 @@ export type ResolvedQueries = string[];
  */
 export type ResolvedTests = string[];
 
+/**
+ * A compilation message for a test message (either an error or a warning)
+ */
+export interface CompilationMessage {
```
charisk, line 165: Should this go in `new-messages.ts`?

Also should we rename `new-messages.ts` (and I guess `new-query-runner.ts`)? Happy for that to be a follow up.

charisk (approved):
Thanks! It's great to see this cleaned up. I can't think of anything wrong with these changes.

extensions/ql-vscode/src/codeql-cli/cli.ts
koesie10: Not really, since this is not a message that is produced by the query server. This message is only produced for tests, and never as part of a query run. Therefore, I don't think it makes sense to move it to `new-messages.ts`.

I can rename `new-messages.ts` in a follow-up PR. `new-query-runner.ts` has already been renamed as part of https://github.com/github/vscode-codeql/pull/3171

PR #3174, opened 2023-12-22
Switch view to ESBuild and JSX runtime

koesie10:
This switches the view to ESBuild and to [the JSX runtime](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html). This reduces the number of different tools we're using and switches to a more modern React pipeline. There are essentially no changes in behavior, and only minor changes to the code for things that ESBuild doesn't support (like glob imports).

Most of the changes in the code are removing the `import * as React from 'react'` statement which isn't required anymore with the new JSX runtime. This is just to satisfy the linter, the code would work with this line.

It's easiest to review this commit-by-commit.

## Checklist


robertbrignull (approved):
Thanks for doing this. It's exciting to remove webpack and simplify things to one build system.

I've had a look through the changes and they look ok, though I'm not too familiar with this area. I've also tried building and running the code locally and the views appear to be working exactly as before.

PR #3183, opened 2024-01-02
Remove ownerName and repositoryName from DistributionConfig

robertbrignull:
As far as I can tell, these fields do not work. They aren't linked up to read values from the config, and setting `"codeQL.cli.repositoryName": "foo"` in your config does nothing.

This reason for this is because the only way these values are accessed is through the `DistributionConfigListener` class, which implements the `DistributionConfig` interface. That class does not define these fields and therefore they get their default value of `undefined`. The reason we didn't notice this earlier is because the field is allowed to be `undefined`. If you change the type to `string | null` then it produces an error.

Looking back in the history, I believe the intention was to fully remove these in https://github.com/github/vscode-codeql/pull/174 (see the comments on the linked internal issues) but for whatever reason that PR didn't remove all references and instead only marked them as maybe undefined. I'm a little confused by that, but I think that removing these final references does not change behaviour and is in the spirit of that earlier change.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/codeql-cli/distribution.ts
+++ b/extensions/ql-vscode/src/codeql-cli/distribution.ts
@@ -512,9 +512,7 @@ class ExtensionSpecificDistributionManager {
   }
 
   private get distributionOwnerName(): string {
-    if (this.config.ownerName) {
-      return this.config.ownerName;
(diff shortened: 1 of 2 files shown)


koesie10 (approved):
Thanks, LGTM!

PR #3185, opened 2024-01-03
Add GitHub Actions workflow for updating Node version

koesie10:
This adds a GitHub Actions workflow that runs on a schedule to update the Node version based on the latest released VS Code version. The PR will only be created when there is a change.

## Checklist

### DIFF
--- a/.github/workflows/update-node-version.yml
+++ b/.github/workflows/update-node-version.yml
@@ -1,0 +1,58 @@
+name: Update Node version
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '15 12 * * *' # At 12:15 PM UTC every day
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  create-pr:
+    name: Create PR
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: extensions/ql-vscode/.nvmrc
+          cache: 'npm'
+          cache-dependency-path: extensions/ql-vscode/package-lock.json
+      - name: Install dependencies
+        working-directory: extensions/ql-vscode
+        run: |
+          npm ci
+        shell: bash
+      - name: Get current Node version
+        working-directory: extensions/ql-vscode
+        id: get-current-node-version
+        run: |
+          echo "version=$(cat .nvmrc)" >> $GITHUB_OUTPUT
+        shell: bash
+      - name: Update Node version
+        working-directory: extensions/ql-vscode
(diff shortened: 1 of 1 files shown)


.github/workflows/update-node-version.yml
```diff
@@ -0,0 +1,58 @@
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: extensions/ql-vscode/.nvmrc
+          cache: 'npm'
+          cache-dependency-path: extensions/ql-vscode/package-lock.json
+      - name: Install dependencies
+        working-directory: extensions/ql-vscode
+        run: |
+          npm ci
+        shell: bash
```
robertbrignull, line 27: As far as I know, specifying `shell: bash` is unnecessary. But it doesn't hurt to be explicit and I have no problem with it.

It's unnecessary because `bash` is already [the default shell for `ubuntu-latest`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell). There's an edge case where if `bash` isn't available then it'll use `sh`, but it will always be available on hosted runnners. Also as far as I know, everything in this workflow behaves the same on `bash` and `sh`.

robertbrignull (approved):
Looks ok to me as a workflow for calling `scripts/update-node-version.ts` and notifying us that the version has changed.

Have you tested it (either in this repo or another) to make sure it works?

I had to remind myself what `.nvmrc` actually controls. As far as I understand, it basically only controls the node version used during CI. In production and during local development it'll use the version of nodejs and the vscode extension you have installed locally. Therefore we do indeed want `.nvmrc` to track the latest version used by vs code.

.github/workflows/update-node-version.yml
koesie10: I mostly copied this from our other workflows where we're also specifying `shell: bash`. I'm not sure why, but it seems like we should try to be consistent in either using it or not using it.

koesie10:
> Have you tested it (either in this repo or another) to make sure it works?

I've tested running this workflow using a `push` trigger with no changes, which works correctly: https://github.com/github/vscode-codeql/actions/runs/7396235203. I haven't tested this when there are changes, but this workflow is very similar to the [`bump-cli.yml` workflow](https://github.com/github/vscode-codeql/blob/main/.github/workflows/bump-cli.yml), which has been tested and the `.github/actions/create-pr` action should just work.

.github/workflows/update-node-version.yml
robertbrignull: Thanks for explaining. Sounds good to me.

robertbrignull:
Thanks for confirming the testing. No objections from me to merging this.

PR #3188, opened 2024-01-03
Move ReleasesApiConsumer to a separate file and do simple refactors

robertbrignull:
This PR aims to do some simple refactors of the `ReleasesApiConsumer` class and associated code. There is more refactoring I want to do but I'm breaking it up into multiple PRs.

Each change is made in a separate commit, and a lot of it is just moving code around, so I recommend looking at each commit individually.

The changes we make are:
- Move various classes out of `distribution.ts` and into their own files
- Use newer methods of defining fields instead of having to prefix with underscores
- Add / adjust comments on classes and methods
- Move the tests of `ReleasesApiConsumer` out of the no-workspace tests and into unit-tests so they'll be faster
- Merge owner name and repo name into a single repo NWO field, because after https://github.com/github/vscode-codeql/pull/3183 we no longer need to be able to specify them individually anywhere

This PR deliberately does not attempt to change the parameters of the `getLatestRelease` method. I think that change will be a little more complicated so I wanted to save that for another PR.

## Checklist


koesie10 (approved):
These changes look good to me. However, perhaps all files related to the distribution could be moved into a `codeql-cli/distribution` directory? It seems like files like `github-api-error.ts` don't belong together with the CLI code directly. If you're already doing this in a follow-up PR, that's fine as well.

robertbrignull:
> However, perhaps all files related to the distribution could be moved into a `codeql-cli/distribution` directory?

That's a good idea. I'll move them to that directory in this PR to keep all file file-moving in one PR.

robertbrignull:
I've moved the files into `codeql-cli/distribution` but I haven't moved `distribution.ts` yet as it would create slightly ugly imports like `distribution/distribution`. We could solve this by a `index.ts` file, or just continue the work of decomposing `distribution.ts` into smaller and better-named files. We can also just move it into the `distribution` directory and that's fine.

Since it's so easy to move files around I'm not too bothered about getting it perfect right away, as we can always tweak things. So I won't hold up this PR any more and will merge it now.

PR #3202, opened 2024-01-05
Fix missing `await` in `prepareModelEditorQueries`

koesie10:
This fixes a missing `await` in `prepareModelEditorQueries` that caused the function to potentially return before the query files were written.

This was discovered by looking at the error in https://github.com/github/vscode-codeql/pull/3197.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/model-editor-queries.ts
+++ b/extensions/ql-vscode/src/model-editor/model-editor-queries.ts
@@ -55,10 +55,12 @@ export async function prepareModelEditorQueries(
     return false;
   }
   // Create the query file.
-  Object.values(Mode).map(async (mode) => {
-    const queryFile = join(queryDir, queryNameFromMode(mode));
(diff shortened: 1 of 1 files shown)


charisk (approved):
Nice one!

robertbrignull:
I was wondering why the linting rule didn't catch this, but it's because we were creating an `Promise[]` object and not using `Promise.all`. That does seem hard for the linter to catch and also hopefully unlikely to happen too often.

koesie10:
> I was wondering why the linting rule didn't catch this, but it's because we were creating an `Promise[]` object and not using `Promise.all`. That does seem hard for the linter to catch and also hopefully unlikely to happen too often.

The reason I found this at all is because it is flagged by the linter now ([TypeScript ESLint v6.17.0](https://github.com/typescript-eslint/typescript-eslint/releases/tag/v6.17.0)), so hopefully it should now not be introduced again.

PR #3203, opened 2024-01-05
Enforce consistent type imports

koesie10:
This PR enables two ESLint rules that together enforce a consistent style for type imports:

- [`@typescript-eslint/consistent-type-imports`](https://typescript-eslint.io/rules/consistent-type-imports/): This enforces that when an import is only used as a type, it is imported using `import type`
- [`import/consistent-type-specifier-style`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/consistent-type-specifier-style.md): This enforces the styling of the type import to be `import type { A, B }` rather than `import { type A, type B }`.

Together, these ensure that we're always using a consistent type import for imports which are only used as types. This will help us when checking for cycles in our dependency tree since cycles on type-only imports [do not have a runtime effect](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md).

I've created this PR by enabling one rule at a time and then running `npm run lint -- --fix`. I've made no manual changes to any files other than `.eslintrc.js`. You should be able to validate this by starting from the same base commit and running `npm run lint -- --fix` after enabling both rules. There should be no changes compared to this PR.

## Checklist


charisk (approved):
Seems like a good change to me. Thanks for noting that there are no manual changes outside of editing the lint rules - it made it much easier to review.

Do we want to make sure the rest of the team is onboard with this change?

robertbrignull (approved):
LGTM too. Especially because the rule can be auto-fixed.

PR #3214, opened 2024-01-09
Improve user experience when no database is selected

shati-patel:
Previously, if you ran a query without having a DB selected, the extension would pop up a file explorer window prompting you to select a database archive from your file system. This isn't the most common way to select a database so it was a fairly confusing pop-up. See internal issue for more info 🔍 

### Changes

New flow (inspired by designs from the internal issue, though I updated the wording slightly):
- Choose between an existing or new DB:
  image
- If you choose "existing", we list the DBs from your DB panel
  image
- If you choose "new", we offer the different options for importing a DB:
  image

In all cases, we select that DB and run the query against it 🎉 (Note: this also works in the same way for "[CodeQL: Debug Query](https://github.com/github/vscode-codeql/pull/2291)") 

### Notes

- Suggestions on wording and code structure very welcome as always! 👀 
- I haven't done anything particularly clever to decide which DBs to display. I just show all of them, without filtering/sorting by language for example. That might be a nice follow-up task, but this seems like a good improvement already 😅 

## Checklist

  - No docs update needed

### DIFF
--- a/extensions/ql-vscode/src/databases/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/databases/local-databases-ui.ts
@@ -814,10 +814,10 @@ export class DatabaseUI extends DisposableObject {
    * notification if it tries to perform any long-running operations.
    */
   private async getDatabaseItemInternal(
-    progress: ProgressContext | undefined,
+    progressContext: ProgressContext | undefined,
   ): Promise<DatabaseItem | undefined> {
     if (this.databaseManager.currentDatabaseItem === undefined) {
-      progress?.progress({
+      progressContext?.progress({
         maxStep: 2,
         step: 1,
         message: "Choosing database",
--- a/extensions/ql-vscode/src/databases/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/databases/local-databases-ui.ts
@@ -5,6 +5,7 @@ import type {
   ProviderResult,
   TreeDataProvider,
   CancellationToken,
+  QuickPickItem,
 } from "vscode";
 import {
   EventEmitter,
@@ -28,7 +29,11 @@ import type {
   ProgressCallback,
   ProgressContext,
 } from "../common/vscode/progress";
-import { withInheritedProgress, withProgress } from "../common/vscode/progress";
+import {
+  UserCancellationException,
+  withInheritedProgress,
+  withProgress,
+} from "../common/vscode/progress";
 import {
   isLikelyDatabaseRoot,
   isLikelyDbLanguageFolder,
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -797,12 +817,119 @@ export class DatabaseUI extends DisposableObject {
     progress: ProgressContext | undefined,
   ): Promise<DatabaseItem | undefined> {
     if (this.databaseManager.currentDatabaseItem === undefined) {
-      await this.chooseAndSetDatabase(false, progress);
+      progress?.progress({
```
norascheuch, line 820: Is it possible to remove one of the `progress` here? Similar to how it's done in https://github.com/github/vscode-codeql/blob/koesie10/code-server-docker-compose/extensions/ql-vscode/src/databases/database-fetcher.ts#L143

norascheuch (approved):
I think the flow looks amazing, just one minor question on the code. Thanks for making these changes! 💐

charisk:
Drive by comment, shall we add some tests? Perhaps as a separate issue/PR.

extensions/ql-vscode/src/databases/local-databases-ui.ts
robertbrignull: It's because `progress` here is a `ProgressContext` instead of a regular `ProgressCallback`. I'm not totally sure why this code uses `ProgressContext`, but it doesn't seem to be something added by this PR.

You could do something like change the name of `progress` to `progressContext` (or whatever you want) and then do
```typescript
const { progress, token } = progressContext;
```
and then you can use `progress` and `token` as normal.

Or you could go further and basically remove the `ProgressContext` type. As far as I can tell, we only use it from this file, and we only ever use the `progress` field from it and never the `token` field. So arguably we could clean up this code and change all references to `ProgressContext` to just use `ProgressCallback` instead.

If that didn't make any sense I'm happy to explain more, or pair on it. We could also save it for a separate PR after this one.

shati-patel: Thanks! For now, I've just renamed the first `progress` to `progressContext` to avoid some confusion. Agreed that the other changes would be nice, but slightly out of scope for this PR 😅

shati-patel:
> Drive by comment, shall we add some tests? Perhaps as a separate issue/PR.

Yes, I'm keen to add tests as a follow-up PR! 🧪  If I don't get round to it in the next few days, I'll open a new issue too.

PR #3232, opened 2024-01-11
Switch to built-in VS Code test UI unconditionally

### LINKED ISSUE #1099: Migrate to new vscode testing APIs
Currently, this extension uses the [UI test explorer](https://github.com/hbenl/vscode-test-explorer) as a ui for testing. However, this extension has recently become deprecated due to the release of [native APIs for testing](https://code.visualstudio.com/api/extension-guides/testing).

We should migrate to this new API. It will simplify our development since this extension would no longer have any dependencies on a third party extension.

Here is a [migration guide](https://code.visualstudio.com/api/extension-guides/testing). 

Note that when we change this, we can also change the devcontainer files on the vscode-codeql starter repo to avoid installing the test UI extension by default.

dbartol:
Now that we've been using the new test UI for several months in canary builds, I've enabled that new UI unconditionally. This removes our dependency on the Test Explorer UI" and "Test Adapter Converter" extensions and related packages.

The initial commit just removes the old implementation from `test-adapter.ts`, but leaves the shared base class in `test-manager-base.ts` intact. A later commit merges all of that code into `test-manager.ts` now that we don't have two different test managers to worry about.

I've also updated both the repo readme and the extension readme to remove mentions of those two external extensions.

Fixes #1099


extensions/ql-vscode/src/extension.ts
```diff
@@ -977,27 +971,8 @@ async function activateWithInstalledDistribution(
   const testRunner = new TestRunner(dbm, cliServer);
   ctx.subscriptions.push(testRunner);
 
-  let testManager: TestManagerBase | undefined = undefined;
-  if (isCanary()) {
```
aeisenberg, line 981: So previously, only our canary users were using this feature? And with this change, all users will be using it. It's probably safe since most users of the testing feature are internal and therefore using canary.

aeisenberg (approved):

extensions/ql-vscode/src/extension.ts
dbartol: Yes.

PR #3233, opened 2024-01-11
Use CLI to bundle packs for MRVA

dbartol:
This PR updates the VS Code extension to use `codeql pack bundle --mrva` to create the compiled pack to submit to MRVA. If the current CLI does not support `codeql pack bundle --mrva`, we fall back to the old manual pack creation path, where we manually copy the appropriate files and adjust the dependencies.

A commit-by-commit review is suggested. The first couple commits have the major changes. The remainder are either trivial fixes or minor test logging improvements. These test logging improvements were useful for debugging CI failures, so I left them in for future use.

Key changes:

#### Version testing
- The extension can now consume the JSON output of `codeql version` to see what features the current CLI version supports. We use this to detect `--mrva` support for `codeql pack bundle`. This is the preferred feature detection mechanism going forward.
- 
#### `generateQueryPack()`
`generateQueryPack()` for MRVA now has three main cases:
1. The selected query is in a pack, and the CLI supports `--mrva`. In this case, we just invoke the CLI to do the bundling.
2. The selected query is in a pack, and the CLI does not support `--mrva`. In this case, we do what we did before: copy the necessary parts of the pack to a temporary directory, try to fix up the dependencies as best we can, and then have the CLI bundle that temporary pack.
3. The selected query is not in a pack. Regardless of CLI support, we just create a temporary pack with that query and have the CLI bundle that temporary pack.

#### Test improvements
In the test code, I extended the Jest `expect` object to add `.toExistInPack()`. This tests whether a particular file path exists in the QL pack, and gives a better failure message if it does not.

#### 8.3 path expansion
Before we invoke `codeql pack bundle`, we now expand any 8.3 path components in the directory path on Windows. The `pack bundle` command ignores any files outside of the pack root, but current versions of the CLI wind up expanding 8.3 paths for the paths of the individual files, but do not expand the path of the pack root itself. This causes the CLI to think that, for example, `C:\SomeLongName\foo.ql` is not inside of `C:\SomeLo~1` and leave the file out of the bundled pack. 

I couldn't find any standard NodeJS file operation that will expand 8.3 paths, so I had to write one myself. My implementation traverses each 8.3 component in the path and enumerates the children of the parent directory, looking for one whose dev+inode matches the dev+inode for the short path.

By default, modern versions of Windows don't even generate 8.3 paths anymore. However:
1. The Actions runner images have opted in to 8.3 path generation.
2. The username under which the job runs on the worker is long (`runneradmin`).
3. The temp directory is under the `C:\Users\runneradmin` directory.
4. The `TEMP` environment variable uses the 8.3 (`C:\Users\runner~1`) version of the path.

This lead to failures that only reproduced in CI. Because local dev machines don't enable 8.3 path generation, I didn't add any specific tests for this. We're basically just testing it indirectly when we run the MRVA tests in CI.

### DIFF
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -34,6 +34,7 @@ import { QueryLanguage } from "../common/query-language";
 import { LINE_ENDINGS, splitStreamAtSeparators } from "../common/split-stream";
 import type { Position } from "../query-server/messages";
 import { LOGGING_FLAGS } from "./cli-command";
+import type { CliFeatures, VersionAndFeatures } from "./cli-version";
 
 /**
  * The version of the SARIF format that we are using.
@@ -193,7 +194,9 @@ type OnLineCallback = (
   line: string,
 ) => Promise<string | undefined> | string | undefined;
 
-type VersionChangedListener = (newVersion: SemVer | undefined) => void;
+type VersionChangedListener = (
+  newVersionAndFeatures: VersionAndFeatures | undefined,
+) => void;
 
 /**
  * This class manages a cli server started by `codeql execute cli-server` to
@@ -211,8 +214,8 @@ export class CodeQLCliServer implements Disposable {
   /**  A buffer with a single null byte. */
   nullBuffer: Buffer;
 
-  /** Version of current cli, lazily computed by the `getVersion()` method */
-  private _version: SemVer | undefined;
+  /** Version of current cli and its supported features, lazily computed by the `getVersion()` method */
+  private _versionAndFeatures: VersionAndFeatures | undefined;
 
   private _versionChangedListeners: VersionChangedListener[] = [];
 
@@ -288,7 +291,7 @@ export class CodeQLCliServer implements Disposable {
     const callback = (): void => {
       try {
         this.killProcessIfRunning();
-        this._version = undefined;
+        this._versionAndFeatures = undefined;
         this._supportedLanguages = undefined;
       } finally {
         this.runNext();
@@ -1417,6 +1420,27 @@ export class CodeQLCliServer implements Disposable {
     );
   }
 
+  public async packCreate(
+    dir: string,
+    workspaceFolders: string[],
+    outputPath: string,
+    moreOptions: string[],
+  ): Promise<void> {
+    const args = [
+      "--output",
+      outputPath,
+      dir,
+      ...moreOptions,
+      ...this.getAdditionalPacksArg(workspaceFolders),
+    ];
+
+    return this.runJsonCodeQlCliCommandWithAuthentication(
+      ["pack", "create"],
+      args,
+      "Creating pack",
+    );
+  }
+
   async packBundle(
     dir: string,
     workspaceFolders: string[],
@@ -1481,27 +1505,35 @@ export class CodeQLCliServer implements Disposable {
     );
   }
 
-  public async getVersion() {
-    if (!this._version) {
+  public async getVersion(): Promise<SemVer> {
+    return (await this.getVersionAndFeatures()).version;
+  }
+
+  public async getFeatures(): Promise<CliFeatures> {
+    return (await this.getVersionAndFeatures()).features;
+  }
+
+  public async getVersionAndFeatures(): Promise<VersionAndFeatures> {
+    if (!this._versionAndFeatures) {
       try {
-        const newVersion = await this.refreshVersion();
-        this._version = newVersion;
+        const newVersionAndFeatures = await this.refreshVersion();
+        this._versionAndFeatures = newVersionAndFeatures;
         this._versionChangedListeners.forEach((listener) =>
-          listener(newVersion),
+          listener(newVersionAndFeatures),
         );
 
         // this._version is only undefined upon config change, so we reset CLI-based context key only when necessary.
         await this.app.commands.execute(
           "setContext",
           "codeql.supportsQuickEvalCount",
-          newVersion.compare(
+          newVersionAndFeatures.version.compare(
             CliVersionConstraint.CLI_VERSION_WITH_QUICK_EVAL_COUNT,
           ) >= 0,
         );
         await this.app.commands.execute(
           "setContext",
           "codeql.supportsTrimCache",
-          newVersion.compare(
+          newVersionAndFeatures.version.compare(
             CliVersionConstraint.CLI_VERSION_WITH_TRIM_CACHE,
           ) >= 0,
         );
@@ -1512,23 +1544,23 @@ export class CodeQLCliServer implements Disposable {
         throw e;
       }
     }
-    return this._version;
+    return this._versionAndFeatures;
   }
 
   public addVersionChangedListener(listener: VersionChangedListener) {
-    if (this._version) {
-      listener(this._version);
+    if (this._versionAndFeatures) {
+      listener(this._versionAndFeatures);
     }
     this._versionChangedListeners.push(listener);
   }
 
-  private async refreshVersion() {
+  private async refreshVersion(): Promise<VersionAndFeatures> {
     const distribution = await this.distributionProvider.getDistribution();
     switch (distribution.kind) {
       case FindDistributionResultKind.CompatibleDistribution:
       // eslint-disable-next-line no-fallthrough -- Intentional fallthrough
       case FindDistributionResultKind.IncompatibleDistribution:
-        return distribution.version;
+        return distribution.versionAndFeatures;
 
       default:
         // We should not get here because if no distributions are available, then
@@ -1744,5 +1776,9 @@ export class CliVersionConstraint {
     return this.isVersionAtLeast(
       CliVersionConstraint.CLI_VERSION_WITH_EXTENSIBLE_PREDICATE_METADATA,
     );
+  }
+
+  async supportsMrvaPackCreate(): Promise<boolean> {
+    return (await this.cli.getFeatures()).mrvaPackCreate === true;
   }
 }
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -1420,27 +1420,6 @@ export class CodeQLCliServer implements Disposable {
     );
   }
 
-  public async packCreate(
-    dir: string,
-    workspaceFolders: string[],
-    outputPath: string,
-    moreOptions: string[],
-  ): Promise<void> {
-    const args = [
-      "--output",
-      outputPath,
-      dir,
-      ...moreOptions,
-      ...this.getAdditionalPacksArg(workspaceFolders),
-    ];
-
-    return this.runJsonCodeQlCliCommandWithAuthentication(
-      ["pack", "create"],
-      args,
-      "Creating pack",
-    );
-  }
-
   async packBundle(
     dir: string,
     workspaceFolders: string[],
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -1420,16 +1420,28 @@ export class CodeQLCliServer implements Disposable {
     );
   }
 
+  /**
+   * Compile a CodeQL pack and bundle it into a single file.
+   *
+   * @param sourcePackDir The directory of the input CodeQL pack.
+   * @param workspaceFolders The workspace folders to search for additional packs.
+   * @param outputBundleFile The path to the output bundle file.
+   * @param outputPackDir The directory to contain the unbundled output pack.
+   * @param moreOptions Additional options to be passed to `codeql pack bundle`.
+   */
   async packBundle(
-    dir: string,
+    sourcePackDir: string,
     workspaceFolders: string[],
-    outputPath: string,
+    outputBundleFile: string,
+    outputPackDir: string,
     moreOptions: string[],
   ): Promise<void> {
     const args = [
       "-o",
-      outputPath,
-      dir,
+      outputBundleFile,
+      sourcePackDir,
+      "--pack-path",
+      outputPackDir,
       ...moreOptions,
       ...this.getAdditionalPacksArg(workspaceFolders),
     ];
@@ -1492,7 +1504,7 @@ export class CodeQLCliServer implements Disposable {
     return (await this.getVersionAndFeatures()).features;
   }
 
-  public async getVersionAndFeatures(): Promise<VersionAndFeatures> {
+  private async getVersionAndFeatures(): Promise<VersionAndFeatures> {
     if (!this._versionAndFeatures) {
       try {
         const newVersionAndFeatures = await this.refreshVersion();
--- a/extensions/ql-vscode/src/common/short-paths.ts
+++ b/extensions/ql-vscode/src/common/short-paths.ts
@@ -1,11 +1,7 @@
 import { platform } from "os";
 import { basename, dirname, join, normalize, resolve } from "path";
 import { lstat, readdir } from "fs/promises";
-import { extLogger } from "./logging/vscode";
-
-async function log(message: string): Promise<void> {
-  await extLogger.log(message);
-}
+import type { BaseLogger } from "./logging";
 
 /**
  * Expand a single short path component
@@ -16,8 +12,9 @@ async function log(message: string): Promise<void> {
 async function expandShortPathComponent(
   dir: string,
   shortBase: string,
+  logger: BaseLogger,
 ): Promise<string> {
-  await log(`Expanding short path component: ${shortBase}`);
+  void logger.log(`Expanding short path component: ${shortBase}`);
 
   const fullPath = join(dir, shortBase);
 
@@ -25,36 +22,36 @@ async function expandShortPathComponent(
   const stats = await lstat(fullPath, { bigint: true });
   if (stats.dev === BigInt(0) || stats.ino === BigInt(0)) {
     // No inode info, so we won't be able to find this in the directory listing.
-    await log(`No inode info available. Skipping.`);
+    void logger.log(`No inode info available. Skipping.`);
     return shortBase;
   }
-  await log(`dev/inode: ${stats.dev}/${stats.ino}`);
+  void logger.log(`dev/inode: ${stats.dev}/${stats.ino}`);
 
   try {
     // Enumerate the children of the parent directory, and try to find one with the same dev/inode.
     const children = await readdir(dir);
     for (const child of children) {
-      await log(`considering child: ${child}`);
+      void logger.log(`considering child: ${child}`);
       try {
         const childStats = await lstat(join(dir, child), { bigint: true });
-        await log(`child dev/inode: ${childStats.dev}/${childStats.ino}`);
+        void logger.log(`child dev/inode: ${childStats.dev}/${childStats.ino}`);
         if (childStats.dev === stats.dev && childStats.ino === stats.ino) {
           // Found a match.
-          await log(`Found a match: ${child}`);
+          void logger.log(`Found a match: ${child}`);
           return child;
         }
       } catch (e) {
         // Can't read stats for the child, so skip it.
-        await log(`Error reading stats for child: ${e}`);
+        void logger.log(`Error reading stats for child: ${e}`);
       }
     }
   } catch (e) {
     // Can't read the directory, so we won't be able to find this in the directory listing.
-    await log(`Error reading directory: ${e}`);
+    void logger.log(`Error reading directory: ${e}`);
     return shortBase;
   }
 
-  await log(`No match found. Returning original.`);
+  void logger.log(`No match found. Returning original.`);
   return shortBase;
 }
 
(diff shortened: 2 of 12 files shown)


extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
```diff
@@ -20,10 +20,50 @@ import { ExtensionApp } from "../../../../src/common/vscode/vscode-app";
...
 import { load } from "js-yaml";
 import type { ExtensionPackMetadata } from "../../../../src/model-editor/extension-pack-metadata";
 import type { QlPackLockFile } from "../../../../src/packaging/qlpack-lock-file";
+import { expect } from "@jest/globals";
+import type { ExpectationResult } from "expect";
+
+/**
+ * Custom Jest matcher to check if a file exists in a query pack.
+ */
+function toExistInPack(
```
charisk, line 34: Shall we move this to `/test/matchers`  (and maybe call it toExistInCodeQLPack`)?

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
```diff
@@ -20,10 +20,50 @@ import { ExtensionApp } from "../../../../src/common/vscode/vscode-app";
...
+  actual: unknown,
+  packFS: QueryPackFS,
+): ExpectationResult {
+  if (typeof actual !== "string") {
+    throw new TypeError("Expected actual value to be a string.");
+  }
+
+  const pass = packFS.fileExists(actual);
+  const files = packFS.allFiles();
+  const filesString = files.length > 0 ? files.join("\n") : "<none>";
```
charisk, line 45: Should we use `os.EOL` instead of `\n`? Similar comment for the message.

extensions/ql-vscode/src/common/short-paths.ts
```diff
@@ -0,0 +1,111 @@
...
+    return join(dir, shortBase);
+  }
+
+  // This component looks like it has a short name, so try to expand it.
+  const longBase = await expandShortPathComponent(dir, shortBase);
+  return join(dir, longBase);
+}
+
+/**
+ * Expands a path that potentially contains 8.3 short names (e.g. "C:\PROGRA~1" instead of "C:\Program Files").
```
robertbrignull, line 88: I don't think 8.3 short names is a concept that will be familiar to many people. I hadn't heard of it, though thankfully once I'd recognised that it's a windows thing a google search did point me in the right direction.

Should we include more explanation or link to something like https://en.wikipedia.org/wiki/8.3_filename?

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
```diff
@@ -242,19 +279,36 @@ function isFileSystemRoot(dir: string): boolean {
...
+  compiledPackDir: string;
+  bundleFile: string;
+}
+
+async function createRemoteQueriesTempDirectory(): Promise<RemoteQueryTempDir> {
+  const shortRemoteQueryDir = await dir({
     dir: tmpDir.name,
     unsafeCleanup: true,
   });
+  // Expand 8.3 filenames here to work around a CLI bug where `codeql pack bundle` produces an empty
```
robertbrignull, line 294: How big of a problem do you expect this to be in practice? My research says they're a feature from pre-windows 95 and now only supported for compatibility with old systems. We don't consider these 8.3 filenames anywhere else, but should we? How likely are we to encounter them with users?

I think it's impressive that we're handling them and I don't want to discount the work you've done here, but the code in `short-paths.ts` is quite complex so I just want to be sure that the maintenance burden is going to be worth it compared to how many users it will unblock. Do you have any more insights or data on this?

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1417,6 +1420,27 @@ export class CodeQLCliServer implements Disposable {
     );
   }
 
+  public async packCreate(
```
charisk, line 1423: I don't see any usages of this. Is that intentional?

extensions/ql-vscode/src/common/short-paths.ts
```diff
@@ -0,0 +1,111 @@
...
+ * @param shortPath The path to expand.
+ * @returns A normalized, absolute path, with any short components expanded.
+ */
+export async function expandShortPaths(shortPath: string): Promise<string> {
```
robertbrignull, line 92: Can we add some tests of this please? It's quite complex and esoteric, and since this code only runs on windows we're not going to be testing it in our day-to-day running (except for poor Shati who will hit all the bugs). So this would greatly benefit from some unit tests that run on windows.

extensions/ql-vscode/src/extension.ts
```diff
@@ -431,13 +431,16 @@ export async function activate(
...
+        CliVersionConstraint.OLDEST_SUPPORTED_CLI_VERSION.compare(ver.version) <
+        0
+      ) {
         return;
       }
 
       void showAndLogWarningMessage(
         extLogger,
-        `You are using an unsupported version of the CodeQL CLI (${ver}). ` +
+        `You are using an unsupported version of the CodeQL CLI (${ver.version}). ` +
```
robertbrignull, line 443: How confident are you you've found all of these places where we insert the version into a string? I haven't exhaustively gone through trying to find them, so I just thought I'd  ask how you feel.

If we miss any it'll lead to a classic `[Object object]` which wouldn't look too good.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
```diff
@@ -58,84 +60,119 @@ interface GeneratedQueryPack {
...
-      originalPackRoot,
-      queryFile,
-      queryPackDir,
-      packRelativePath,
-    );
+  const language: QueryLanguage | undefined = mustSynthesizePack
+    ? await askForLanguage(cliServer) // open popup to ask for language if not already hardcoded
+    : await findLanguage(cliServer, Uri.file(queryFile));
+  if (!language) {
+    throw new UserCancellationException("Could not determine language.");
```
robertbrignull, line 82:
```suggestion
    throw new UserCancellationException("Could not determine language");
```

I think the normal convention is to not use punctuation in error messages so they can more easily be composed together and stuff.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
```diff
@@ -58,84 +60,119 @@ interface GeneratedQueryPack {
...
+  }
 
-    language = await findLanguage(cliServer, Uri.file(targetQueryFileName));
-  } else {
-    // open popup to ask for language if not already hardcoded
-    language = await askForLanguage(cliServer);
+  let queryPackDir: string;
+  let precompilationOpts: string[];
```
robertbrignull, line 86: Can we move this definition of `precompilationOpts` down to line 141 just before the `if-else` where it is defined? Might make it easier for future readers.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
```diff
@@ -58,84 +60,119 @@ interface GeneratedQueryPack {
...
 
-  await cliServer.packInstall(queryPackDir, {
-    workspaceFolders,
-  });
+    // Clear the CLI cache so that the most recent qlpack lock file is used.
+    await cliServer.clearCache();
+  }
 
   // Clear the CLI cache so that the most recent qlpack lock file is used.
   await cliServer.clearCache();
```
robertbrignull, line 140: Can we delete this `clearCache` call? We might have already just done it on line 136, or on line 103. So I think we'll have always already cleared after the last CLI command.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
```diff
@@ -58,84 +60,119 @@ interface GeneratedQueryPack {
...
+      join(queryPackDir, packRelativePath),
+      // We need to specify the extension packs as dependencies so that they are included in the MRVA pack.
+      // The version range doesn't matter, since they'll always be found by source lookup.
+      ...extensionPacks.map((p) => `--extension-pack=${p}@*`),
+    ];
+  } else {
+    if (await cliServer.cliConstraints.usesGlobalCompilationCache()) {
+      precompilationOpts = ["--qlx"];
+    } else {
+      const ccache = join(originalPackRoot, ".cache");
```
robertbrignull, line 154:
```suggestion
      const cache = join(originalPackRoot, ".cache");
```
Typo, or is this intentional?

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
```diff
@@ -20,10 +20,50 @@ import { ExtensionApp } from "../../../../src/common/vscode/vscode-app";
...
+/**
+ * Custom Jest matcher to check if a file exists in a query pack.
+ */
+function toExistInPack(
+  this: jest.MatcherContext,
+  actual: unknown,
+  packFS: QueryPackFS,
+): ExpectationResult {
+  if (typeof actual !== "string") {
+    throw new TypeError("Expected actual value to be a string.");
```
robertbrignull, line 40:
```suggestion
    throw new TypeError(
      `Expected actual value to be a string. Found: ${typeof actual}`,
    );
```
So it'll be easier to debug if this fails.

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
```diff
@@ -20,10 +20,50 @@ import { ExtensionApp } from "../../../../src/common/vscode/vscode-app";
...
+  }
+
+  const pass = packFS.fileExists(actual);
+  const files = packFS.allFiles();
```
robertbrignull, line 44: Technically we don't have to construct `files` or `filesString` except in the "fail" case. They could be moved into that branch.

extensions/ql-vscode/test/vscode-tests/cli.ts
```diff
@@ -6,7 +6,10 @@ import { workspace } from "vscode";
  */
 function hasCodeQL() {
   const folders = workspace.workspaceFolders;
-  return !!folders?.some((folder) => folder.uri.path.endsWith("/codeql"));
+  return !!folders?.some(
+    (folder) =>
+      folder.uri.path.endsWith("/codeql") || folder.uri.path.endsWith("/ql"),
```
robertbrignull, line 11:
```suggestion
      folder.uri.path.endsWith(`${os.sep}codeql`) || folder.uri.path.endsWith(`${os.sep}ql`),
```
I realise the original code is also using "/codeql", but since we're looking at it, should this be using `os.sep` or is forward slash always the right thing to use?

charisk (commented):
This is generally looking good to me, although there's quite a lot going on so I could have easily missed something.

I have a few comments but the main functionality seems okay to me. I've also tested (with the currently released version of the CLI) and things seem to be working fine.

robertbrignull (commented):
I've taken a look through. Generally looks good though I have quite a few comments because this is quite a large change to a few different areas.

I haven't gone through the short paths code in high details because it's an area I'm really not familiar with.

I've also done a manual test and it's correctly building a query pack and able to run MRVA, at least with CodeQL CLI 2.16.0. Have you done manual testing using old versions of the CodeQL CLI?

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
charisk: This is the existing naming, I guess because it stands for compilation cache?

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
robertbrignull: Sorry I didn't read the PR description well enough. I see you've explained that you hit this on the actions runners and why it's happening. In that case I see you're point that it's only really a case for CI and not for user machines. I still think some unit tests would be helpful, unless they're really hard to write. But I see the reason why this is implemented.

extensions/ql-vscode/src/common/short-paths.ts
```diff
@@ -0,0 +1,111 @@
+import { platform } from "os";
+import { basename, dirname, join, normalize, resolve } from "path";
+import { lstat, readdir } from "fs/promises";
+import { extLogger } from "./logging/vscode";
+
+async function log(message: string): Promise<void> {
+  await extLogger.log(message);
```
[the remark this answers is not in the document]
koesie10: If we're not removing it, I'd suggest using the `BaseLogger` interface instead of using `extLogger` directly. All of the code in this file is independent of VS Code and could be tested using unit tests, but adding the `extLogger` import adds a dependency on the `vscode` module, which requires any tests to be located in the `vscode-tests` directory.

extensions/ql-vscode/test/vscode-tests/cli.ts
koesie10: I think the `uri` is normalized to always contain `/`. Only `fsPath` contains the platform-specific path separators.

However, could we simply use `dirname` here since that seems to be what we're checking against? Something like this (I kept it short to avoid introducing extra variables, but it might be a good idea to do explicit `===` checks instead of `includes`):

```suggestion
      ["codeql", "ql"].includes(dirname(folder.uri.fsPath)),
```

extensions/ql-vscode/src/codeql-cli/cli.ts
dbartol: No, I think I must have added it before I realized we actually needed to call `pack bundle` instead. Removed.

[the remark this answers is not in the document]
dbartol: Removed.

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1417,6 +1420,27 @@ export class CodeQLCliServer implements Disposable {
...
+  ): Promise<void> {
+    const args = [
+      "--output",
+      outputPath,
+      dir,
+      ...moreOptions,
+      ...this.getAdditionalPacksArg(workspaceFolders),
+    ];
+
+    return this.runJsonCodeQlCliCommandWithAuthentication(
```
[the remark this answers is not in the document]
dbartol: I believe there are some scenarios where `codeql pack bundle` can attempt to download packs from the registry, but I'm not sure which cases those are TBH. The CLI will only prompt for auth info if it actually does try to contact the registry, so this shouldn't affect the case where we're just accessing local packs.

extensions/ql-vscode/src/common/short-paths.ts
```diff
@@ -0,0 +1,111 @@
+import { platform } from "os";
```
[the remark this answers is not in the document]
dbartol: Which repo should we track the tech debt in?
Any unit test we write would require manual machine configuration to run on a typical developer machine, since 8.3 name generation has been off by default for the last decade or so. It would still run in CI, though, since the Actions runners seem to have that feature enabled.

extensions/ql-vscode/src/common/short-paths.ts
```diff
@@ -0,0 +1,111 @@
...
+import { basename, dirname, join, normalize, resolve } from "path";
+import { lstat, readdir } from "fs/promises";
+import { extLogger } from "./logging/vscode";
+
+async function log(message: string): Promise<void> {
```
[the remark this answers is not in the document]
dbartol: This code never gets executed on non-CI machines, so I found it very convenient to have one place where I could replace the `extlogger` call with `console.log` and have messages show up in the CI log directly. I can add a comment to this effect.

Is there a better way to see the extension log from a CI run? If so, I'll put that in a comment at the top of this file for anyone unlucky enough to have to debug it.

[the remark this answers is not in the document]
dbartol: Nevermind, I've moved the code over to accept a `BaseLogger` object, so now we could just pass a console logger at the call site if we temporarily want console debugging. No need for the private `log()` function anymore.

extensions/ql-vscode/src/common/short-paths.ts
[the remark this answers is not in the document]
dbartol: Fixed.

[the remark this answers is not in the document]
dbartol: This code is only executed on machines with a particular legacy feature enabled, and even then only if we actually see an 8.3 filename. Outside of our CI machines, which have this enabled, I expect that fewer than 1% of customers will ever get a single line logged from this code. On the other hand, if something does go wrong in this code, we're unlikely to notice without a log message, and it will be difficult to reproduce locally. So, I'd prefer to leave them in.

I did convert everything over to use `BaseLogger`, though, which solved one of Robert's comments and will make it easier to write a unit test in the future if we decide to.

extensions/ql-vscode/src/common/short-paths.ts
dbartol: Added a link.

Now I guess I'll crawl back in my cave and finish that painting of a mastodon I've been working on.

extensions/ql-vscode/src/common/short-paths.ts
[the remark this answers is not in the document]
dbartol: Good idea. Done.

dbartol: See https://github.com/github/vscode-codeql/pull/3233/files#r1456142526

extensions/ql-vscode/src/extension.ts
dbartol: I feel pretty confident. The only place the `VersionAndFeatures` gets exposed outside of `CodeQLCliServer` is via the version changed listener event (here). I made `CodeQLCliServer.getVersionAndFeatures` private, so anyone else will either call `getVersion` or `getFeatures`.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
dbartol: Fixed, although this was the same message we had before.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
dbartol: I think that should work. Fixed.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
dbartol: I suspect it was a typo. Fixed.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
```diff
@@ -58,84 +60,119 @@ interface GeneratedQueryPack {
...
   void extLogger.log(
     `Compiling and bundling query pack from ${queryPackDir} to ${bundlePath}. (This may take a while.)`,
   );
-  await cliServer.packBundle(
-    queryPackDir,
-    workspaceFolders,
-    bundlePath,
-    precompilationOpts,
-  );
+  await cliServer.packBundle(queryPackDir, workspaceFolders, bundlePath, [
```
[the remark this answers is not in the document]
dbartol: Easy enough to do now, since this is the only call site. See doc comment on `packBundle`

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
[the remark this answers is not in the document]
dbartol: https://github.com/github/codeql-core/issues/4089

extensions/ql-vscode/test/vscode-tests/cli.ts
dbartol: I've made the change. I assume we're happy to assume these are file system paths?

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
dbartol: Done

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
dbartol: Done

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
dbartol: Done

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
dbartol: Done

dbartol:
Thanks for all the reviews. I believe I've addressed all the feedback.

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
charisk: Thanks! Just a couple `\n` left on the message (L32).

extensions/ql-vscode/src/common/short-paths.ts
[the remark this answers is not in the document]
charisk: I'll raise an issue to cover this

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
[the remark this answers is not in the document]
charisk: Nice one! Thanks for the link. Reminder that that's a private repo though 😬

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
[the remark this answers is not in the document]
charisk: Thank you, that's much more readable to me!

robertbrignull (approved):
This looks good to me, and I've given it another quick manual test.

Probably want to get an approval from @charisk too, but overall looks like this could be good to go and it was a very quick and nice turnaround.

charisk (approved):
Thanks for addressing the various comments and for getting this over the line! 🎉

PR #3242, opened 2024-01-16
Open suggest box with Ctrl + Space

koesie10:
This adds a keyboard shortcut to open the suggest box when it's closed. This matches the behavior in VS Code itself. Unfortunately, as far as I could find we can't retrieve the keybindings that are set by the user, so we can't dynamically change the shortcut based on the user's settings.

To test this, you can use Storybook. Inside the story, you can already press Escape to close the suggest box. If you now press Ctrl + Space, it should re-open the suggest box.

## Checklist


koesie10:
> Slightly unrelated, but on the subject of keybindings: I expect Tab to select the autocomplete option from the suggestion box, similar to autocomplete in VS Code. Would it make sense to add that? 👀

That sounds like a good idea, I'll open a separate PR to add that.

PR #3245, opened 2024-01-16
Add prototype implementation for codeQL.runVariantAnalysisPublishedPack

robertbrignull:
Implements the `codeQL.runVariantAnalysisPublishedPack` command. This can likely still be considered a prototype implementation, but it seems to work quite well. Of course right now it fails saying you can't run multiple queries, but that will be fixed in the future.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
@@ -215,42 +215,65 @@ export class VariantAnalysisManager
   }
 
   private async runVariantAnalysisFromPublishedPack(): Promise<void> {
-    const language = await askForLanguage(this.cliServer);
+    return withProgress(async (progress, token) => {
+      progress({
+        maxStep: 8,
+        step: 0,
+        message: "Determining query language",
+      });
 
-    const packName = `codeql/${language}-queries`;
-    const packDownloadResult = await this.cliServer.packDownload([packName]);
-    const downloadedPack = packDownloadResult.packs[0];
+      const language = await askForLanguage(this.cliServer);
 
-    const packDir = join(
-      packDownloadResult.packDir,
-      downloadedPack.name,
-      downloadedPack.version,
-    );
+      progress({
+        maxStep: 8,
+        step: 1,
+        message: "Downloading query pack",
+      });
 
-    const suitePath = join(
-      packDir,
-      "codeql-suites",
-      `${language}-code-scanning.qls`,
-    );
-    const resolvedQueries = await this.cliServer.resolveQueries(suitePath);
+      const packName = `codeql/${language}-queries`;
+      const packDownloadResult = await this.cliServer.packDownload([packName]);
+      const downloadedPack = packDownloadResult.packs[0];
 
-    const problemQueries =
-      await this.filterToOnlyProblemQueries(resolvedQueries);
-
-    if (problemQueries.length === 0) {
-      void this.app.logger.showErrorMessage(
-        `Unable to trigger variant analysis. No problem queries found in published query pack: ${packName}.`,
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -214,7 +215,51 @@ export class VariantAnalysisManager
...
 
   private async runVariantAnalysisFromPublishedPack(): Promise<void> {
-    throw new Error("Command not yet implemented");
+    const language = await askForLanguage(this.cliServer);
+
+    const packName = `codeql/${language}-queries`;
+    const packDownloadResult = await this.cliServer.packDownload([packName]);
+    const downloadedPack = packDownloadResult.packs[0];
+
+    const packDir = `${packDownloadResult.packDir}/${downloadedPack.name}/${downloadedPack.version}`;
```
koesie10, line 224: Should we use `join` here to handle different path separators?

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -214,7 +215,51 @@ export class VariantAnalysisManager
...
+
+    const suitePath = `${packDir}/codeql-suites/${language}-code-scanning.qls`;
```
koesie10, line 226: Same here, should this use `join`?

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -214,7 +215,51 @@ export class VariantAnalysisManager
...
+    const resolvedQueries = await this.cliServer.resolveQueries(suitePath);
+
+    const problemQueries =
+      await this.filterToOnlyProblemQueries(resolvedQueries);
+
+    if (problemQueries.length === 0) {
+      void this.app.logger.showErrorMessage(
+        `Unable to trigger variant analysis. No problem queries found in published query pack: ${packName}.`,
+      );
```
koesie10, line 235: Should we also return here?

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -214,7 +215,51 @@ export class VariantAnalysisManager
...
+    }
+
+    return withProgress((progress, token) =>
```
koesie10, line 238: I haven't tried running this yet, but would it be helpful to start the progress earlier since quite some work is being done is this method already (downloading the pack, resolving the queries and then resolving query metadata for all queries)?

koesie10 (approved):
LGTM to me as well, I just have some non-blocking improvements that I think could be made.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
robertbrignull: Yes, definitely. Apologies I was a bit sloppy here and forgot about `join`.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
robertbrignull: Yes, this is also a silly mistake. Thanks for pointing it out.

robertbrignull:
I've address the review comments (thank you @koesie10, all of your points were very helpful) and tested it again locally. Everything is working and it hasn't broken existing variant analysis flows as far as I can tell, so I'll go ahead and merge so we can keep moving.

robertbrignull:
> Do you think it's worth adding an integration test to cover the new functionality? Can be a separate PR/Issue.

Yes, I think this would be useful. I'll make an issue so we do this soon.

PR #3248, opened 2024-01-16
Add test for "promptForDatabase"

shati-patel:
Follow-up to https://github.com/github/vscode-codeql/pull/3214 to add a basic test of the new "Select database" prompt. 

I plan to open a new issue for more comprehensive tests and to improve the feature generally, so this is just a starting point and to get advice on whether this is a sensible way to test... 

(Example follow-up work: we could probably filter any existing DBs by language, check that we actually _have_ existing DBs, test those different cases...) 

## Checklist

N/A - test change only

### DIFF
--- a/extensions/ql-vscode/src/databases/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/databases/local-databases-ui.ts
@@ -827,7 +827,7 @@ export class DatabaseUI extends DisposableObject {
     return this.databaseManager.currentDatabaseItem;
   }
 
-  public async promptForDatabase(): Promise<void> {
+  private async promptForDatabase(): Promise<void> {
     const quickPickItems: DatabaseSelectionQuickPickItem[] = [
       {
         label: "$(database) Existing database",
@@ -862,7 +862,7 @@ export class DatabaseUI extends DisposableObject {
     }
   }
 
-  public async selectExistingDatabase() {
+  private async selectExistingDatabase() {
     const dbItems: DatabaseQuickPickItem[] =
       this.databaseManager.databaseItems.map((dbItem) => ({
         label: dbItem.name,
@@ -884,7 +884,7 @@ export class DatabaseUI extends DisposableObject {
     );
   }
 
-  public async importNewDatabase() {
+  private async importNewDatabase() {
     const importOptions: DatabaseImportQuickPickItems[] = [
       {
         label: "$(github) GitHub",
--- a/extensions/ql-vscode/src/databases/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/databases/local-databases-ui.ts
@@ -243,7 +243,7 @@ export interface DatabaseQuickPickItem extends QuickPickItem {
   databaseItem: DatabaseItem;
 }
 
-interface DatabaseImportQuickPickItems extends QuickPickItem {
+export interface DatabaseImportQuickPickItems extends QuickPickItem {
   importType: "URL" | "github" | "archive" | "folder";
 }
 
--- a/extensions/ql-vscode/src/databases/local-databases-ui.ts
+++ b/extensions/ql-vscode/src/databases/local-databases-ui.ts
@@ -235,7 +235,7 @@ async function chooseDatabaseDir(byFolder: boolean): Promise<Uri | undefined> {
   return getFirst(chosen);
 }
 
-interface DatabaseSelectionQuickPickItem extends QuickPickItem {
+export interface DatabaseSelectionQuickPickItem extends QuickPickItem {
   databaseKind: "new" | "existing";
 }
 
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -827,7 +827,7 @@ export class DatabaseUI extends DisposableObject {
     return this.databaseManager.currentDatabaseItem;
   }
 
-  private async promptForDatabase(): Promise<void> {
+  public async promptForDatabase(): Promise<void> {
```
shati-patel, line 830: I'm not particularly happy about making all these methods public... Is this okay, or is there a better way?

extensions/ql-vscode/test/vscode-tests/no-workspace/databases/local-databases-ui.test.ts
```diff
@@ -7,14 +7,57 @@ import {
...
+    } as any,
+    {
+      onLanguageContextChanged: () => {
+        /**/
+      },
+    } as any,
+    {} as any,
+    storageDir,
+    storageDir,
+  );
```
shati-patel, line 60: I moved this code up from the "orphaned databases" test, since I needed access to the same `databaseUI` and example data and didn't want to duplicate the code. I don't think this is a problem, but happy to change approach otherwise!

extensions/ql-vscode/src/databases/local-databases-ui.ts
charisk: A way to do that would be to test the public function that is using this one (e.g. `downloadDatabaseFromGitHub`). That's the public interface of the module, and what callers will use, so you would be testing the same flows as the callers. IMO that would be the best way to go than making 3 members public. Would that make the test too difficult/complex?

Another approach is to move `promptForDatabase` in its own file/module, showing that in itself it has a public interface that it can be used and tested independently.  

Having said all that, I want to call out that there are often cases where we have to make things public for testing and I think that's acceptable and better than not having tests. Ideally we'd add a code comment to tell people that a member is only exported for testing and should not be used in non test code.

shati-patel: > A way to do that would be to test the public function that is using this one (e.g. `downloadDatabaseFromGitHub`). That's the public interface of the module, and what callers will use, so you would be testing the same flows as the callers. IMO that would be the best way to go than making 3 members public. Would that make the test too difficult/complex?

`promptForDatabase` is only called from `getDatabaseItemInternal` (which is also private). I could try testing the public wrapper `getDatabaseItem` though!

charisk: Ah sorry, it looks like there's a `promptForDatabases` (plural) and I was looking at that instead 🤦‍♀️ 

Thanks, it would be good to try against `getDatabaseItem`. But as I mentioned, if it's too hard it may not be worth the trouble.

extensions/ql-vscode/test/vscode-tests/no-workspace/databases/local-databases-ui.test.ts
```diff
@@ -121,6 +120,106 @@ describe("local-databases-ui", () => {
...
+          ),
+        );
+
+      const selectExistingDatabaseSpy = jest
+        .spyOn(databaseUI as any, "selectExistingDatabase")
+        .mockResolvedValue(undefined);
+
+      const importNewDatabaseSpy = jest
+        .spyOn(databaseUI as any, "importNewDatabase")
+        .mockResolvedValue(undefined);
```
shati-patel, line 195: Slightly tweaked the tests so didn't have to make all the methods public 🤫 This meant I had to use `as any` here to spy on these methods, but that's probably better?

charisk: Hmm so we're mocking out methods on the object we're testing? That seems a bit odd. Doesn't it means that we're not actually testing them?

shati-patel: Yeah, I haven't added tests for `selectExistingDatabase` and `importNewDatabase` in this PR, just for the overall `promptForDatabase` dropdown 👀  

Happy to have a go now if that makes more sense? Though I was also planning to open a follow-up issue to improve the feature + tests generally 💅🏽

charisk: It's a little bit odd to me that we're testing a function that calls other functions but those other functions are being mocked. You'd normally mock out dependencies etc. but have real implementations of the component that is being tested.

But it's a step forward so I'm happy for us to do some further improvements later on.

shati-patel:
Thanks for the patient reviewing! Finally got round to looking at tests in a bit more detail 🔎

I haven't tested all the possible options, and I'm still mocking some methods (e.g. I didn't have access to a real `DatabaseManager`, so I mocked `setCurrentDatabaseItem`).
But I _think_ this is now more sensible as a test? 😅

charisk (approved):
Thanks, LGTM!

PR #3258, opened 2024-01-18
Revert "Bump actions/dependency-review-action from 3 to 4"

norascheuch:
Reverts github/vscode-codeql#3252

It looks like our workflow runner uses ubuntu-latest, which has node18 https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -13,4 +13,4 @@ jobs:
       - name: 'Checkout Repository'
         uses: actions/checkout@v4
       - name: 'Dependency Review'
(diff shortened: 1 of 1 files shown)


koesie10 (commented):
I don't think this is necessary: actions that use Node 20 are fine, for example `actions/checkout@v4` also uses Node 20. I think Actions also has Node 20 installed, and it will use the version that is specified by the action.

norascheuch:
Yes, someone from the dependency-review team just confirmed this for me as well.

PR #3260, opened 2024-01-22
Use object for pack generation details

charisk:
The `generateQueryPack` function in `run-remote-query.ts` currently has logic to calculate a few things like the QL pack root, paths to the query files, the language, etc. In order to add multi-query support, a few of these calculations need to be moved further up the stack and then passed in to `generateQueryPack` as arguments. 

In this PR, we're introducing a new type (`QlPackDetails`) in the `/variant-analysis` domain to hold such information so that a single object can be passed in to `generateQueryPack` and using that to include the `queryFile`. In follow up PRs, we'll also need to move the language, and other related information. We'll also transition from a singular `queryFile` to multiple ones.

I'm not 100% convinced on the `QlPackDetails` name, so open to suggestions! We can also rename this once it has been fleshed out a bit more.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
+++ b/extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
@@ -39,6 +39,7 @@ import { tryGetQueryMetadata } from "../codeql-cli/query-metadata";
 import { askForLanguage, findLanguage } from "../codeql-cli/query-language";
 import type { QlPackFile } from "../packaging/qlpack-file";
 import { expandShortPaths } from "../common/short-paths";
+import type { QlPackDetails } from "./ql-pack-details";
 
 /**
  * Well-known names for the query pack used by the server.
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
```diff
@@ -59,9 +60,11 @@ interface GeneratedQueryPack {
  */
 async function generateQueryPack(
   cliServer: CodeQLCliServer,
-  queryFile: string,
+  qlPackDetails: QlPackDetails,
```
robertbrignull, line 63: This function handles when the queries are in an existing pack as well as when they are not. In that case would the `QlPackDetails` be undefined, or would we still pass this object but with some of the fields within it empty? It's description says "details about the original QL pack" which might not be incorrect in the "no existing pack" case.

robertbrignull (approved):
LGTM

One question which maybe doesn't have to be answered right now, and could get answered on its own as we flesh out this object. But could be good to think about now anyway so we avoid creating incorrect documentation.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
charisk: I'm imagining `QlPackDetails` to still be defined (with some empty fields), so ideally I would prefer a different name for it, but I haven't been able to come up with a better name. Perhaps one will more naturally come up as we make the changes (or once we're done with them).

robertbrignull: A better name would definitely be nice, but I also can't think of anything right now. Happy to go ahead and address it later when we have a better idea of things.

PR #3261, opened 2024-01-22
Stop variant analysis from published pack if no language selected

charisk:
Tiny change to stop processing a variant analysis using a published back if the user doesn't select a language.

## Checklist
N/A - the functionality being changed is behind an unveiled feature flag:

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
@@ -223,6 +223,9 @@ export class VariantAnalysisManager
       });
 
       const language = await askForLanguage(this.cliServer);
+      if (!language) {
+        return;
+      }
 
       progress({
         maxStep: 8,


extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -223,6 +223,9 @@ export class VariantAnalysisManager
       });
 
       const language = await askForLanguage(this.cliServer);
+      if (!language) {
```
robertbrignull, line 226: Does this have any behavioural effect right now? We're using the default value of `throwOnEmpty = true` so it should throw a `UserCancellationException` if the user doesn't select a language? Is that how it works?

Regardless there are still some code paths where it still returns undefined, and I agree we should abort in this case.

Aborting without logging anything should be fine. As far as I can tell we'll always have already logged the reason when returning `undefined` from `askForLanguage`.

robertbrignull (approved):
LGTM with one comment just to clarify my own understanding of things

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
charisk: > Does this have any behavioural effect right now? We're using the default value of `throwOnEmpty = true` so it should throw a `UserCancellationException` if the user doesn't select a language? Is that how it works?

Ah yeah it looks like there is no behaviour change. That's how I'm reading the code too. 

So this will mainly make the types a bit nicer.

PR #3262, opened 2024-01-22
Push QL pack details creation to commands

charisk:
Pushes creating `QlPackDetails` even further up the stack - where details will be decided, which is nearer the commands.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
+++ b/extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
@@ -332,24 +332,14 @@ interface PreparedRemoteQuery {
 export async function prepareRemoteQueryRun(
   cliServer: CodeQLCliServer,
   credentials: Credentials,
-  uris: Uri[],
+  qlPackDetails: QlPackDetails,
   progress: ProgressCallback,
   token: CancellationToken,
   dbManager: DbManager,
 ): Promise<PreparedRemoteQuery> {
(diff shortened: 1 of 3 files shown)


robertbrignull (approved):
Ok by me to merge if this is the order you want to do it, though alternatives are also fine with me.

The general idea of defining `QlPackDetails` earlier and passing it through makes sense to me.

What's a little odd is that it requires adding TODO comments and walking back some of the changes we made for supporting multiple queries. I expect the very next change will be to re-revert these changes when we add multi-query support into `QlPackDetails`.

Would it be easier and require fewer steps to convert `QlPackDetails` to support multiple queries first or in this PR as well?

charisk:
> Ok by me to merge if this is the order you want to do it, though alternatives are also fine with me.
> 
> The general idea of defining `QlPackDetails` earlier and passing it through makes sense to me.
> 
> What's a little odd is that it requires adding TODO comments and walking back some of the changes we made for supporting multiple queries. I expect the very next change will be to re-revert these changes when we add multi-query support into `QlPackDetails`.
> 
> Would it be easier and require fewer steps to convert `QlPackDetails` to support multiple queries first or in this PR as well?

Agree, either approach is fine. It's a bit odd either way - we either use `queryFiles` and only allow a single file later on, or we have some TODOs 🤷 I'm going to go ahead with this for now and build on it.

PR #3263, opened 2024-01-22
Fix MRVA integration test failure

dbartol:
Newer CLI versions preserve the metadata for extensible predicates in the compiled pack, so they no longer preserve queries that contain extensible predicates when creating a MRVA pack.

Fixes https://github.com/github/semmle-code/issues/48686

### DIFF
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -1737,6 +1737,9 @@ export class CliVersionConstraint {
    */
   public static CLI_VERSION_WITH_TRIM_CACHE = new SemVer("2.15.1");
 
+  public static CLI_VERSION_WITH_EXTENSIBLE_PREDICATE_METADATA_IN_QLX =
+    new SemVer("2.16.1");
+
   constructor(private readonly cli: CodeQLCliServer) {
     /**/
   }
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1737,6 +1737,9 @@ export class CliVersionConstraint {
    */
   public static CLI_VERSION_WITH_TRIM_CACHE = new SemVer("2.15.1");
 
+  public static CLI_VERSION_WITH_EXTENSIBLE_PREDICATE_METADATA_IN_QLX =
+    new SemVer("2.16.1");
```
cklin, line 1741: This constant is a bit misleading:
- Extensible predicate metadata is stored in a created qlpack (in `.packinfo` files), not in QLX
- The CLI version that starts generating extensible predicate metadata is 2.16.0, not 2.16.1

Perhaps the constant name we want is something like `CLI_VERSION_WITHOUT_MRVA_EXTENSIBLE_PREDICATE_HACK`?

dbartol: I've renamed the constant and the function to better reflect the details, plus added a few comments.

cklin (approved):

aeisenberg (approved):
Any problem merging this now so we can fix our integration tests?

PR #3273, opened 2024-01-24
Use Uri to convert paths to platform specific fs paths

robertbrignull:
Aims to fix the CLI tests on windows. We believe they were broken in https://github.com/github/vscode-codeql/pull/3262 by the change to `getFile`. This PR adds back in the use of `Uri.file(...).fsPath` to convert the paths supplied by the test code (which always use forward slashes) to a platform-dependent file path.

Successful CLI test run: https://github.com/github/vscode-codeql/actions/runs/7639479147

There may well be other ways of achieving this, though I can't immediately find one that's clearly better. Stackoverflow is suggesting `somepath.split(path.sep).join(locale.sep)` though I'm not sure I like that more than using `Uri`. Since it's just for the CLI tests I don't think we need to be too worried about being perfect, so if this works it should be fine.

## Checklist


robertbrignull:
I'll merge this once CI is green.

Side note, we could also have fixed this by converting the tests to supply paths in the platform-dependent format. I decided not to because that would change more code and here I'm mainly just trying to fix the tests by reverting the change that broke them.

PR #3274, opened 2024-01-24
Add support for running multiple queries in MRVA

charisk:
Updates query pack creation and relevant MRVA commands to allow having multiple queries.

I'm still doing a bit of manual testing but worth starting to review the code changes.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/ql-pack-details.ts
+++ b/extensions/ql-vscode/src/variant-analysis/ql-pack-details.ts
@@ -5,14 +5,14 @@ import type { QueryLanguage } from "../common/query-language";
  * a variant analysis.
  */
 export interface QlPackDetails {
-  // The full paths of the query files.
+  // The absolute paths of the query files.
   queryFiles: string[];
 
-  // The path to the QL pack that is used for triggering a variant analysis.
+  // The absolute path to the QL pack that is used for triggering a variant analysis.
   // If there is no query pack, this is the same as the directory of the query files.
   qlPackRootPath: string;
 
-  // The path to the QL pack file (a qlpack.yml or codeql-pack.yml) or undefined if
+  // The absolute path to the QL pack file (a qlpack.yml or codeql-pack.yml) or undefined if
   // it doesn't exist.
   qlPackFilePath: string | undefined;
 
--- a/extensions/ql-vscode/src/variant-analysis/ql-pack-details.ts
+++ b/extensions/ql-vscode/src/variant-analysis/ql-pack-details.ts
@@ -5,7 +5,8 @@ import type { QueryLanguage } from "../common/query-language";
  * a variant analysis.
  */
 export interface QlPackDetails {
-  queryFile: string;
+  // The full paths of the query files.
+  queryFiles: string[];
 
   // The path to the QL pack that is used for triggering a variant analysis.
   // If there is no query pack, this is the same as the directory of the query files.
--- a/extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
+++ b/extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
@@ -1,6 +1,6 @@
 import type { CancellationToken } from "vscode";
 import { Uri, window } from "vscode";
-import { relative, join, sep, basename } from "path";
+import { join, sep, basename, relative } from "path";
 import { dump, load } from "js-yaml";
 import { copy, writeFile, readFile, mkdirp } from "fs-extra";
 import type { DirectoryResult } from "tmp-promise";
@@ -34,7 +34,6 @@ import {
   QLPACK_FILENAMES,
   QLPACK_LOCK_FILENAMES,
 } from "../common/ql";
-import { tryGetQueryMetadata } from "../codeql-cli/query-metadata";
 import type { QlPackFile } from "../packaging/qlpack-file";
 import { expandShortPaths } from "../common/short-paths";
 import type { QlPackDetails } from "./ql-pack-details";
@@ -56,10 +55,7 @@ async function generateQueryPack(
   qlPackDetails: QlPackDetails,
   tmpDir: RemoteQueryTempDir,
 ): Promise<string> {
-  const queryFile = qlPackDetails.queryFile;
-
   const originalPackRoot = qlPackDetails.qlPackRootPath;
-  const packRelativePath = relative(originalPackRoot, queryFile);
   const workspaceFolders = getOnDiskWorkspaceFolders();
   const extensionPacks = await getExtensionPacksToInject(
     cliServer,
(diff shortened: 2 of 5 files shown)


extensions/ql-vscode/src/variant-analysis/ql-pack-details.ts
```diff
@@ -5,7 +5,8 @@ import type { QueryLanguage } from "../common/query-language";
  * a variant analysis.
  */
 export interface QlPackDetails {
-  queryFile: string;
+  // The full paths of the query files.
```
robertbrignull, line 8: Does "full path" mean "absolute path"?

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -312,22 +307,35 @@ export class VariantAnalysisManager
...
 
-  private async runVariantAnalysisCommand(uri: Uri): Promise<void> {
-    // Build up details to pass to the functions that run the variant analysis.
-    const qlPackRootPath = await findPackRoot(uri.fsPath);
+  private async runVariantAnalysisCommand(queryFiles: Uri[]): Promise<void> {
+    if (queryFiles.length === 0) {
+      throw new Error("Please select a .ql file to run as a variant analysis");
+    }
+
+    const qlPackRootPath = await findPackRoot(queryFiles[0].fsPath);
```
robertbrignull, line 315: `findPackRoot` needs a little attention.

It works fine for the case where the queries are in a pack, but for the case where there is no pack it currently returns the immediate directory containing the query. For multiple queries not in a pack and not in the same directory this will make it complain they don't belong to the same query pack, when instead we could make it work by placing the pack root higher up the directory tree.

I suggest we change the output to `string | undefined` and then have a separate way of computing the ideal pack root in the case of multiple queries? That or `findPackRoot` takes an array of query paths and does this calculation and validation itself.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -374,7 +380,12 @@ export class VariantAnalysisManager
       this.dbManager,
     );
 
-    const queryName = getQueryName(queryMetadata, queryFile);
+    const firstQueryFile = qlPackDetails.queryFiles[0];
```
robertbrignull, line 383: Getting the metadata and query text of the first query is a bit of a hack to fit in with the existing data format. Do we want to mark it as such with a comment? We don't plan to be addressing this any time soon so it could sit like this for a while and we'll forget it's like this.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -383,20 +394,22 @@ export class VariantAnalysisManager
...
+    const queryText = await readFile(firstQueryFile, "utf8");
 
-    // TODO: Once we have basic support multiple queries, and qlPackDetails has
-    // more than 1 queryFile, we should set this to have a proper value
-    // (e.g. { language: variantAnalysisLanguage })
-    const queries: VariantAnalysisQueries | undefined = undefined;
+    const queries: VariantAnalysisQueries | undefined =
+      qlPackDetails.queryFiles.length === 1
+        ? undefined
+        : {
```
robertbrignull, line 402: I'm not suggesting we do anything more in this PR, but I think now would be a good time to think more about this `queryFiles` field. I'll write up some of my thoughts in an issue.

robertbrignull (commented):
Overall looks fine and in the right direction, and works for me when I tested it on single queries. I only have a couple of small questions.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
```diff
@@ -191,12 +192,16 @@ async function createNewQueryPack(
...
-  packRelativePath: string,
 ) {
-  const toCopy = await cliServer.packPacklist(originalPackRoot, false);
+  const originalPackRoot = qlPackDetails.qlPackRootPath;
+  const queryFiles = qlPackDetails.queryFiles;
+
+  const toCopy = await cliServer.packPacklist(
+    qlPackDetails.qlPackRootPath,
+    false,
+  );
```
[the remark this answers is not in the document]
charisk: I can't think of anything we could do except catch in code reviews etc. One thing i want to call out is that I generally tried to stick to original naming to help with the code review. I'd be happy to do another pass (maybe in a separate PR) and make things a bit more consistent. WDYT?

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
charisk: Ah yes, I was planning to leave a comment but forgot!

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
charisk: Do you think it's worth changing the underlying code (run-remote-query.ts) so that it doesn't expect `qlPackRootPath` to always be set, instead of the change here? I've been tempted to change it but I thought that would cause too much churn and divert me from my task. But it does feel like it's the more correct thing to do. 

If we decide to go with that, I think it should be a separate PR though.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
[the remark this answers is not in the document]
robertbrignull: I think another pass would be good. I appreciate wanting to keep the reviewing easy, but it does leave the code in a bit of a middle-ground. I'm happy for it to be part of this PR if you are, or do it as a followup PR.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
robertbrignull: As in make `QlPackDetails.qlPackRootPath` optional? That makes sense to me, and I was a little surprised when I read that interface to see that the field was always defined even if the query pack didn't exist.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -374,7 +380,15 @@ export class VariantAnalysisManager
       this.dbManager,
     );
 
-    const queryName = getQueryName(queryMetadata, queryFile);
+    // For now we ge the metadata for the first query in the pack.
```
robertbrignull, line 383:
```suggestion
    // For now we get the metadata for the first query in the pack.
```

robertbrignull (approved):
I think this is good now (except one typo) and I think we could merge now and do more things in separate PRs, if that's what you prefer.

extensions/ql-vscode/src/variant-analysis/run-remote-query.ts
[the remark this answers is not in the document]
charisk: I started doing this on my PR and I really think it'll mess the diff up, so for the benefit of future readers, I'll have a separate PR.

PR #3275, opened 2024-01-25
Start using eslint-plugin-deprecation to find deprecated code

norascheuch:
Adds the plugin and resolves existing issues. In two cases the deprecated methods couldn't be easily replaced. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/.storybook/vscode-theme-addon/manager.tsx
+++ b/extensions/ql-vscode/.storybook/vscode-theme-addon/manager.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
-import { addons, types } from "@storybook/manager-api";
+import { addons } from "@storybook/manager-api";
+import { Addon_TypesEnum } from "@storybook/types";
 import { ThemeSelector } from "./ThemeSelector";
 
 const ADDON_ID = "vscode-theme-addon";
@@ -7,8 +8,7 @@ const ADDON_ID = "vscode-theme-addon";
 addons.register(ADDON_ID, () => {
   addons.add(ADDON_ID, {
     title: "VSCode Themes",
-    // eslint-disable-next-line deprecation/deprecation
-    type: types.TOOL,
+    type: Addon_TypesEnum.TOOL,
     match: ({ viewMode }) => !!(viewMode && viewMode.match(/^(story|docs)$/)),
     render: () => <ThemeSelector />,
   });
--- a/extensions/ql-vscode/.storybook/vscode-theme-addon/manager.tsx
+++ b/extensions/ql-vscode/.storybook/vscode-theme-addon/manager.tsx
@@ -7,6 +7,7 @@ const ADDON_ID = "vscode-theme-addon";
 addons.register(ADDON_ID, () => {
   addons.add(ADDON_ID, {
     title: "VSCode Themes",
+    // eslint-disable-next-line deprecation/deprecation
     type: types.TOOL,
     match: ({ viewMode }) => !!(viewMode && viewMode.match(/^(story|docs)$/)),
     render: () => <ThemeSelector />,
--- a/extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
+++ b/extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
@@ -125,10 +125,13 @@ Example.args = {
 };
 
 faker.seed(42);
-const uniqueStore = {};
+const uniqueNumbers = [...Array(1000).keys()].sort(() => Math.random() - 0.5);
 
 const manyScannedRepos = Array.from({ length: 1000 }, (_, i) => {
   const mockedScannedRepo = createMockScannedRepo();
+  const randomInt = uniqueNumbers.pop();
+  const uniqueId =
+    randomInt === undefined ? Math.floor(Math.random() * 100000) : randomInt;
 
   return {
     ...mockedScannedRepo,
@@ -136,13 +139,8 @@ const manyScannedRepos = Array.from({ length: 1000 }, (_, i) => {
     resultCount: faker.number.int({ min: 0, max: 1000 }),
     repository: {
       ...mockedScannedRepo.repository,
-      // We need to ensure the ID is unique for React keys
-      id: faker.helpers.unique(faker.number.int, [], {
-        store: uniqueStore,
-      }),
-      fullName: `octodemo/${faker.helpers.unique(faker.word.sample, [], {
-        store: uniqueStore,
-      })}`,
+      id: uniqueId,
+      fullName: `octodemo/${uniqueId}`,
     },
   };
 });
--- a/extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
+++ b/extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
@@ -129,7 +129,7 @@ faker.seed(42);
 
 const manyScannedRepos = Array.from({ length: 1000 }, (_, i) => {
   const mockedScannedRepo = createMockScannedRepo();
-  const nanoid = customAlphabet("0123456789");
+  const nanoid = customAlphabet("123456789");
   return {
     ...mockedScannedRepo,
     analysisStatus: VariantAnalysisRepoStatus.Succeeded,
--- a/extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
+++ b/extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
@@ -1,6 +1,7 @@
 import type { Meta, StoryFn } from "@storybook/react";
 
 import { faker } from "@faker-js/faker";
+import { customAlphabet } from "nanoid";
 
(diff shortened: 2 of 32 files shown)


extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
```diff
@@ -125,24 +125,22 @@ Example.args = {
...
 
 faker.seed(42);
-const uniqueStore = {};
+const uniqueNumbers = [...Array(1000).keys()].sort(() => Math.random() - 0.5);
 
 const manyScannedRepos = Array.from({ length: 1000 }, (_, i) => {
   const mockedScannedRepo = createMockScannedRepo();
+  const randomInt = uniqueNumbers.pop();
+  const uniqueId =
+    randomInt === undefined ? Math.random() * 8000 + 1001 : randomInt;
```
norascheuch, line 134: Since array.pop() returns a number OR undefined I needed to add an alternative for the undefined case. This is a best effort for a random number bigger than 1000 without using a library to create it. This case should never happen since the uniqueNumbers array is never undefined.

extensions/ql-vscode/.storybook/vscode-theme-addon/manager.tsx
```diff
@@ -1,13 +1,14 @@
...
+import { Addon_TypesEnum } from "@storybook/types";
 import { ThemeSelector } from "./ThemeSelector";
 
 const ADDON_ID = "vscode-theme-addon";
 
 addons.register(ADDON_ID, () => {
   addons.add(ADDON_ID, {
     title: "VSCode Themes",
-    type: types.TOOL,
+    type: Addon_TypesEnum.TOOL,
```
norascheuch, line 11: There is no clear guidance how to replace `types`, however `Addon_TypesEnum` from storybook/types is used in the source for `types`, so I think this should work.

extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
```diff
@@ -125,24 +125,22 @@ Example.args = {
...
       ...mockedScannedRepo.repository,
-      // We need to ensure the ID is unique for React keys
-      id: faker.helpers.unique(faker.number.int, [], {
-        store: uniqueStore,
-      }),
-      fullName: `octodemo/${faker.helpers.unique(faker.word.sample, [], {
-        store: uniqueStore,
-      })}`,
+      id: uniqueId,
+      fullName: `octodemo/${uniqueId}`,
```
charisk, line 143: Any idea why we're generating our own unique id? Can't faker do this for us? E.g. 
```
`octodemo/${faker.string.uuid}`
```

I'm guessing it doesn't have to be a number, since it's in a string.

norascheuch: Creating a unique number is what they deprecated, there are some discussions and the gist of it is that they ask people to create their own packages for this functionality.

charisk: But do we need a number in this case?

norascheuch: No, but I suppose the string wouldn't be unique either. I don't know what the likelihood is for creating two similar strings randomly.  We could use `https://fakerjs.dev/api/string.html#alpha` but we would have to check for uniqueness.

norascheuch: I think I should add the uniqueness comment from Koen back in?!

extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
robertbrignull: I'm not a huge fan of having code that's not intended to execute, and that branch isn't just throwing an error. I have a few potential solutions to make things a little cleaner:

### Throw an error instead

I'd be happy if we threw an error if `pop` ever returns `undefined`. As you say it shouldn't ever happen, so if it does then the code needs to be updated rather than using a fallback strategy.

### Iterate over `uniqueNumbers`

Instead of doing `Array.from({ length: 1000 }, (_, i) =>` to initiate the loop, do `uniqueNumbers.map(repoId =>` so that we remove the need to call `pop`.

### Just use `Math.random()` and don't precompute a random list

The repo ids don't have to go from 0-999. They just need to be unique, and even then it's only the storybook so not too critical. If we use big enough numbers, e.g. `Math.floor(Math.random() * 1000000000000)` (AFAIK the max safe int in javascript is `9007199254740991`), then the chance of getting duplicates in our 1000 repo ids becomes very low.

extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
robertbrignull: What happens if the keys aren't unique? Is the failure mode that it throws an error, or it silently does wrong stuff?

That will help us answer the question of how careful do we need to be. If React safely detects this case and tells us, this seems fine. The manual solution the user can take is to reload the page and get a new set of random numbers, and this is so easy to do that maybe it's not worth complicating the code.

Overall I don't mind. I've also left a few different suggestions on the other comment thread.

extensions/ql-vscode/.storybook/vscode-theme-addon/manager.tsx
robertbrignull: I also can't see any documentation on intended migration strategy. I tracked down https://github.com/storybookjs/storybook/pull/24834 where this change was made and there's nothing there either.

I agree I think using `Addon_TypesEnum` seems fine.

extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
charisk: > No, but I suppose the string wouldn't be unique either. 

Ah I didn't realise uniqueness was completely removed. In that case, then perhaps we should use `nanoid` like we do elsewhere (e.g. [here](https://github.com/github/vscode-codeql/blob/2f66be3fccb1ed0077acf20d0e790585e907e65e/extensions/ql-vscode/src/run-queries-shared.ts#L506))?

Agree about leaving the uniqueness comment though.

extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
norascheuch: I think I prefer the error, as you say it should never happen and if it does we should look at it!

extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
norascheuch: Ah nice, we already use a library for unique ids! That sounds like a good solution!

robertbrignull: `nanoid` is also fine by me. 👍🏼 

For the record, as far as I know this is also just a random number generator. There's no guarantee of uniqueness, but the randomness is enough that the change of conflict is negligible. But that's fine for this context.

extensions/ql-vscode/src/stories/variant-analysis/VariantAnalysisAnalyzedRepos.stories.tsx
```diff
@@ -125,22 +126,19 @@ Example.args = {
...
   return {
     ...mockedScannedRepo,
     analysisStatus: VariantAnalysisRepoStatus.Succeeded,
     resultCount: faker.number.int({ min: 0, max: 1000 }),
     repository: {
       ...mockedScannedRepo.repository,
-      id: uniqueId,
-      fullName: `octodemo/${uniqueId}`,
+      // We need to ensure the ID is unique for React keys
+      id: parseInt(nanoid()),
```
charisk, line 140: Is `parseInt` a left over from before? `nanoid()` gives you back a string, not a number.

norascheuch: Yes, that's why I have to convert it to a number for the id. Nanoid has no method that returns a number, however by giving it the custom alphabet we can parse the string back to a number.

charisk: Ok I'm now following a bit more what we're trying to do 😂 I didn't realise this is an id for a repo (which is expected to be a number). Although I'm still unclear why we aren't just using `Math.random()` 🤔 

I think what you have is fine and I think it'll work because there are enough characters to avoid collisions. This is also storybook code so it's not terrible if things break 😬 

Do we want to skip `0` since you could end up with a number of `0123`?  `parseInt` still works but it doesn't skip the `0`, it just assumes the number is octal rather than decimal so can be a bit confusing.

charisk (approved):
LGTM, thanks for cleaning up!

PR #3285, opened 2024-01-25
Store multiple queries on variant analysis object

robertbrignull:
The idea here is to update the `VariantAnalysis` to store info for multiple queries. When showing the results view or otherwise displaying anything about the variant analysis, we just use data for the first query. The intention of this PR is that everything should work exactly as before when there's only one query, but it's not expected to do anything sensible yet when there are multiple queries.

Also note that this is just changing the internal type, and this PR isn't changing the query history types or any other data that is stored or sent outside of the extension.

This PR did end up a little larger than I expected it to. I wasn't expecting this change to the type to affect so many places. Wherever I've had to put in a hack of using the first query, I've added a TODO comment. My hope is that we can go through and remove these as we convert the rest of the extension to handle multiple queries.

## Checklist


robertbrignull:
I've pulled out part of this to https://github.com/github/vscode-codeql/pull/3287

robertbrignull:
I'll close this PR since it's become a bit stale, and if we want to reproduce it then making the changes again isn't hard.

PR #3286, opened 2024-01-29
Add basic multi-query support to MRVA results view header

charisk:
Adds a very minimal solution to showing query information for a multi-query MRVA in the results view. This is something we'll expand on in the future.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
@@ -218,14 +218,15 @@ export class VariantAnalysisView
   }
 
   private getTitle(variantAnalysis: VariantAnalysis | undefined): string {
-    if (variantAnalysis) {
-      if (variantAnalysis.queries) {
-        return `Variant Analysis using multiple queries - Results`;
-      }
+    if (!variantAnalysis) {
+      return `Variant Analysis ${this.variantAnalysisId} - Results`;
+    }
 
+    if (variantAnalysis.queries) {
+      return `Variant Analysis using multiple queries - Results`;
+    } else {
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
```diff
@@ -218,9 +218,14 @@ export class VariantAnalysisView
...
-      ? `${variantAnalysis.query.name} - Variant Analysis Results`
-      : `Variant Analysis ${this.variantAnalysisId} - Results`;
+    if (variantAnalysis) {
+      if (variantAnalysis.queries) {
+        return `Variant Analysis using multiple queries - Results`;
+      }
+
+      return `${variantAnalysis.query.name} - Variant Analysis Results`;
+    }
+    return `Variant Analysis ${this.variantAnalysisId} - Results`;
```
robertbrignull, line 228: I find this set of if-statements a little hard to read. Possibly because all of the return statements are at different indentation levels and the whitespace/newline implies that the second two return statements are grouped together more than they should be.

Feel free to ignore this suggestion or do something else, but I suggest the following:
```suggestion
    if (!variantAnalysis) {
      return `Variant Analysis ${this.variantAnalysisId} - Results`;
    }

    if (variantAnalysis.queries) {
      return `Variant Analysis using multiple queries - Results`;
    } else {
      return `${variantAnalysis.query.name} - Variant Analysis Results`;
    }
```

Get the "no variant analysis" case out of the way first, and then give equal weight to the single-query and multi-query cases.

robertbrignull (approved):
Works for me. One small optional comment.

As said earlier, I think merging this now is a good idea so we improve the UI a little bit, and it doesn't block us doing more later.

PR #3287, opened 2024-01-29
Move language and pack to top level of variant analysis object

robertbrignull:
A slimmed down version of https://github.com/github/vscode-codeql/pull/3285 that doesn't modify the `query` object. The idea being to split up that change and make it easier if we do want to do it or something similar to it in the future.

My reasoning on `pack` and `language` is that they're really properties of the variant analysis rather than properties of the query. This becomes more true when you consider multiple queries, but I think it's still true now.

I'm deliberately not modifying the query history format in this PR. We could do, but since it's a lossless transformation in both directions we can just leave it. Or we could save the query history update for when we add in multiple query support and batch those changes together.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/markdown-generation.ts
+++ b/extensions/ql-vscode/src/variant-analysis/markdown-generation.ts
@@ -42,7 +42,7 @@ interface VariantAnalysisMarkdown {
  * Generates markdown files with variant analysis results.
  */
 export async function generateVariantAnalysisMarkdown(
-  variantAnalysis: Pick<VariantAnalysis, "query">,
+  variantAnalysis: Pick<VariantAnalysis, "language" | "query">,
   results: AsyncIterable<
     [VariantAnalysisScannedRepository, VariantAnalysisScannedRepositoryResult]
   >,
@@ -77,7 +77,7 @@ export async function generateVariantAnalysisMarkdown(
       for (const interpretedResult of result.interpretedResults) {
         const individualResult = generateMarkdownForInterpretedResult(
           interpretedResult,
-          variantAnalysis.query.language,
+          variantAnalysis.language,
         );
         resultsFileContent.push(...individualResult);
       }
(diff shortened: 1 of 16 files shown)


extensions/ql-vscode/src/variant-analysis/markdown-generation.ts
```diff
@@ -42,7 +42,7 @@ interface VariantAnalysisMarkdown {
  * Generates markdown files with variant analysis results.
  */
 export async function generateVariantAnalysisMarkdown(
-  variantAnalysis: Pick<VariantAnalysis, "query">,
+  variantAnalysis: Pick<VariantAnalysis, "language" | "query">,
```
charisk, line 45: Not a problem necessarily introduced in this PR, but I would prefer to _not_ use `Pick` if we're going to be getting more than 1 item, as it gradually becomes silly (e.g. https://github.com/github/vscode-codeql/pull/3287/files#diff-ba6fe1916e2a1bce3900880797ea49539e39f9f76c431c2742a4ca5926643b26R50)

charisk (approved):
Makes sense to pull them up a bit! Thanks.

extensions/ql-vscode/src/variant-analysis/markdown-generation.ts
robertbrignull: For this PR I'll just keep it simple and add the extra field. I don't feel this PR crosses any new silliness-thresholds that weren't already crossed.

I'm not really sure why we're using `Pick` here (or the place you linked). I'm not sure what benefit it gives us in terms of ease of writing or maintainability, over taking the whole object. Perhaps it makes craft data in the tests easier, but we have `createMockX` methods for that generally.

robertbrignull:
I dug into the CI failure and uncovered a existing bug in the tests that we just didn't notice until now. I've opened https://github.com/github/vscode-codeql/pull/3290 to separate the test fix from this change.

robertbrignull:
I found another case of us mixing internal and DTO types in the tests. I've opened https://github.com/github/vscode-codeql/pull/3293 for this one.

robertbrignull:
At last the tests are green! 🎉

PR #3289, opened 2024-01-29
Improve handling of unknown QL pack roots for multi-query MRVAs

charisk:
Add some logic to handle calculating QL pack roots for MRVA better.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/common/files.ts
+++ b/extensions/ql-vscode/src/common/files.ts
@@ -1,5 +1,5 @@
 import { pathExists, stat, readdir, opendir } from "fs-extra";
-import { isAbsolute, join, relative, resolve } from "path";
+import { isAbsolute, join, relative, resolve, sep } from "path";
 import { tmpdir as osTmpdir } from "os";
 
 /**
@@ -131,4 +131,31 @@ export function isIOError(e: any): e is IOError {
 // This function is a wrapper around `os.tmpdir()` to make it easier to mock in tests.
 export function tmpdir(): string {
   return osTmpdir();
+}
+
+/**
+ * Finds the common parent directory of an arbitrary number of paths. If the paths are absolute,
+ * the result is also absolute. If the paths are relative, the result is relative.
+ * @param paths The array of paths.
+ * @returns The common parent directory of the paths.
+ */
+export function findCommonParentDir(...paths: string[]): string {
+  // Split each path into its components
+  const pathParts = paths.map((path) => path.split(sep));
+
+  let commonDir = "";
+  // Iterate over the components of the first path and checks if the same
+  // component exists at the same position in all the other paths. If it does,
+  // add the component to the common directory. If it doesn't, stop the
+  // iteration and returns the common directory found so far.
+  for (let i = 0; i < pathParts[0].length; i++) {
+    const part = pathParts[0][i];
+    if (pathParts.every((parts) => parts[i] === part)) {
+      commonDir = `${commonDir}${part}${sep}`;
+    } else {
+      break;
+    }
+  }
+
+  return commonDir;
 }
--- a/extensions/ql-vscode/src/common/files.ts
+++ b/extensions/ql-vscode/src/common/files.ts
@@ -1,6 +1,6 @@
 import { pathExists, stat, readdir, opendir } from "fs-extra";
-import { isAbsolute, join, relative, resolve, sep } from "path";
-import { tmpdir as osTmpdir } from "os";
+import { isAbsolute, join, parse, relative, resolve, sep } from "path";
+import { homedir, tmpdir as osTmpdir } from "os";
 
 /**
  * Recursively finds all .ql files in this set of Uris.
@@ -134,24 +134,27 @@ export function tmpdir(): string {
 }
 
 /**
- * Finds the common parent directory of an arbitrary number of paths. If the paths are absolute,
- * the result is also absolute. If the paths are relative, the result is relative.
+ * Finds the common parent directory of an arbitrary number of paths. The result
+ * will be an absolute path.
  * @param paths The array of paths.
  * @returns The common parent directory of the paths.
  */
 export function findCommonParentDir(...paths: string[]): string {
-  // Split each path into its components
-  const pathParts = paths.map((path) => path.split(sep));
+  const normalizedPaths = paths.map((path) => normalizePath(path));
 
-  let commonDir = "";
-  // Iterate over the components of the first path and checks if the same
+  // Split each path into its components
+  const pathParts = normalizedPaths.map((path) => path.split(sep));
+
+  // Start with the root directory
+  let commonDir = parse(homedir()).root;
+
+  // Iterate over the components of the first path and check if the same
   // component exists at the same position in all the other paths. If it does,
   // add the component to the common directory. If it doesn't, stop the
   // iteration and returns the common directory found so far.
-  for (let i = 0; i < pathParts[0].length; i++) {
-    const part = pathParts[0][i];
+  for (const [i, part] of pathParts[0].entries()) {
     if (pathParts.every((parts) => parts[i] === part)) {
-      commonDir = `${commonDir}${part}${sep}`;
+      commonDir = join(commonDir, part);
     } else {
       break;
     }
--- a/extensions/ql-vscode/src/common/files.ts
+++ b/extensions/ql-vscode/src/common/files.ts
@@ -1,6 +1,6 @@
 import { pathExists, stat, readdir, opendir } from "fs-extra";
-import { isAbsolute, join, parse, relative, resolve, sep } from "path";
-import { homedir, tmpdir as osTmpdir } from "os";
+import { isAbsolute, join, relative, resolve, sep } from "path";
+import { tmpdir as osTmpdir, platform } from "os";
 
 /**
  * Recursively finds all .ql files in this set of Uris.
@@ -134,19 +134,24 @@ export function tmpdir(): string {
 }
 
 /**
- * Finds the common parent directory of an arbitrary number of paths. The result
+ * Finds the common parent directory of an arbitrary number of absolute paths. The result
  * will be an absolute path.
  * @param paths The array of paths.
  * @returns The common parent directory of the paths.
  */
 export function findCommonParentDir(...paths: string[]): string {
+  if (paths.length === 0) {
+    throw new Error("At least one path must be provided");
+  }
+  if (paths.some((path) => !isAbsolute(path))) {
+    throw new Error("All paths must be absolute");
+  }
+
   const normalizedPaths = paths.map((path) => normalizePath(path));
 
-  // Split each path into its components
-  const pathParts = normalizedPaths.map((path) => path.split(sep));
+  const pathParts = normalizedPaths.map((path) => getPathParts(path));
 
-  // Start with the root directory
-  let commonDir = parse(homedir()).root;
+  const commonParts = [];
 
   // Iterate over the components of the first path and check if the same
   // component exists at the same position in all the other paths. If it does,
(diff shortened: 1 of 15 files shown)


extensions/ql-vscode/src/common/files.ts
```diff
@@ -132,3 +132,30 @@ export function isIOError(e: any): e is IOError {
 export function tmpdir(): string {
   return osTmpdir();
 }
+
+/**
+ * Finds the common parent directory of an arbitrary number of paths. If the paths are absolute,
```
robertbrignull, line 137: Are the input paths expected to be files or directories, or potentially either or a mixture?

I ask this because arguably `findCommonParentDir("foo/") === "foo/")` but `findCommonParentDir("foo/query.ql") === "foo/"`.

We may have to check which paths are files and call `dirname` on them.

extensions/ql-vscode/src/common/files.ts
```diff
@@ -132,3 +132,30 @@ export function isIOError(e: any): e is IOError {
...
+  // Split each path into its components
+  const pathParts = paths.map((path) => path.split(sep));
+
+  let commonDir = "";
+  // Iterate over the components of the first path and checks if the same
+  // component exists at the same position in all the other paths. If it does,
+  // add the component to the common directory. If it doesn't, stop the
+  // iteration and returns the common directory found so far.
+  for (let i = 0; i < pathParts[0].length; i++) {
+    const part = pathParts[0][i];
```
robertbrignull, line 152: Could be one line if you prefer. Completely optional.
```suggestion
  for (const [i, part] of pathParts[0].entries()) {
```

extensions/ql-vscode/src/variant-analysis/ql.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  // Calculate the pack root for each query file
+  const packRoots: Array<string | undefined> = [];
+  for (const queryFile of queryFiles) {
+    const packRoot = await findPackRoot(queryFile);
+    packRoots.push(packRoot);
+  }
+
+  if (queryFiles.length === 1) {
```
robertbrignull, line 34: Do we need this special case for one query? I think the logic below still works when there's a single file. Since there's not much below this I wonder if it's simpler to reduce the number of special cases, rather returning early and reducing the amount of code executed.

extensions/ql-vscode/src/variant-analysis/ql.ts
```diff
@@ -0,0 +1,87 @@
...
+  } else {
+    // All in the same pack, return that pack's root
+    return uniquePackRoots[0];
+  }
+}
+
+/**
+ * For queries that are not in a pack, a potential pack root is the
+ * common parent dir of all the queries. However, we only want to
+ * return this if all the queries are in the same workspace folder.
```
robertbrignull, line 59: Can you explain in the code why we want them to be in the same workspace folder

extensions/ql-vscode/src/variant-analysis/ql.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+function findQlPackRootForQueriesWithNoPack(
+  queryFiles: string[],
+  workspaceFolders: string[],
+): string {
+  const queryFileWorkspaceFolders = queryFiles.map((queryFile) =>
+    workspaceFolders.find((workspaceFolder) =>
+      queryFile.startsWith(workspaceFolder),
```
robertbrignull, line 67: I suggest using `containsPath` for this. That helps avoid a number of edge cases.

```suggestion
      containsPath(workspaceFolder, queryFile),
```

Also, this only really works if we're using absolute paths at this point. Is that always the case?

extensions/ql-vscode/src/variant-analysis/ql.ts
```diff
@@ -0,0 +1,87 @@
...
+  if (queryFileWorkspaceFolders.includes(undefined)) {
+    throw Error(
+      "Queries that are not part of a pack need to be part of the workspace",
+    );
+  }
+
+  // Check if query files are not in the same workspace folder.
+  if (new Set(queryFileWorkspaceFolders).size > 1) {
+    throw Error(
+      "Queries that are not part of a pack need to be in the same workspace folder",
```
robertbrignull, line 81: Suggestion for a more readable error message

```suggestion
      "All queries need to be in the same workspace folder",
```

extensions/ql-vscode/src/variant-analysis/ql.ts
```diff
@@ -0,0 +1,87 @@
...
+    return uniquePackRoots[0];
+  }
+}
+
+/**
+ * For queries that are not in a pack, a potential pack root is the
+ * common parent dir of all the queries. However, we only want to
+ * return this if all the queries are in the same workspace folder.
+ */
+function findQlPackRootForQueriesWithNoPack(
```
robertbrignull, line 61: The current implementation is to compute all workspace folders first and do checks and then call `findCommonParentDir` at the end.

An alternative could be to call `findCommonParentDir` first and then check if the resulting path in within a workspace dir. e.g.
```typescript
const commonParentDir = findCommonParentDir(...queryFiles);

for (const workspaceFolder of workspaceFolders) {
  if (containsPath(workspaceFolder, commonParentDir)) {
    return commonParentDir;
  }
}
throw Error(
  "All queries must be within the workspace and within the same workspace root",
);
```

extensions/ql-vscode/src/variant-analysis/ql.ts
```diff
@@ -0,0 +1,87 @@
+import { dirname } from "path";
```
robertbrignull, line 1: Is there a more descriptive name we can use for this file? Unfortunately I can't think of any better examples, but I find just "ql" a bit unclear. It also isn't helped that there's `extensions/ql-vscode/src/common/ql.ts` too which this could get confused with.

extensions/ql-vscode/test/unit-tests/common/files.test.ts
```diff
@@ -417,3 +459,41 @@ describe("walkDirectory", () => {
     expect(files.sort()).toEqual([file1, file2, file3, file4, file5, file6]);
   });
 });
+
+describe("findCommonParentDir", () => {
```
robertbrignull, line 463: The docs on `findCommonParentDir` says it works with absolute and relative paths, but the unit tests only cover relative paths.

Could also do with more unit tests in general, if that's ok. There are so many edge cases with paths 😭

robertbrignull (commented):
I like the behaviour this is aiming for, but I have quite a few comments about the specific implementation.

Apologies for being quite exacting about `findCommonParentDir`. Since we're writing a shared helper file I'd like if we make sure it's as correct as we can make it. I recognise we might not hit these edge-cases right now, but it'll help avoid hard-to-detect bugs if someone else starts using this function in the future.

extensions/ql-vscode/src/variant-analysis/ql.ts
charisk: `ql.ts` made sense as it's QL-related code for MRVA. I couldn't think of a better name. We could prefix it if you think it helps? e.g. `variant-analysis-ql.ts`?

BTW I can imagine other functions being moved here (and out of run-remote-query.ts).

extensions/ql-vscode/src/variant-analysis/ql.ts
charisk: I think there is a slight difference when there is a single query. If the query is not in the workspace, we currently allow MRVA to run. So I wanted to keep that behaviour. For multiple queries it gets more complicated, because if they're not in a pack we want to find a common dir for them so they need to be in the workspace to avoid reaching places in the filesystem where we shouldn't.

extensions/ql-vscode/src/variant-analysis/ql.ts
charisk: It seems wrong to allow queries to be in different workspace folders 🤔 [This](https://github.com/github/vscode-codeql/pull/3289/files#diff-95b4bdfe0ae067210693ab52616487856ee8aa4c3c22f171aea5daebd43e76f7R79) is a the bit where we force that.

I guess we don't strictly need that restriction though, so I can remove it.

extensions/ql-vscode/src/variant-analysis/ql.ts
charisk: Yeah they're absolute paths that we get from the original `Uri`s.

extensions/ql-vscode/src/variant-analysis/ql.ts
robertbrignull: > If the query is not in the workspace, we currently allow MRVA to run. So I wanted to keep that behaviour.

Why do we want this behaviour? To me it seems like an oversight and unlikely that anyone would be relying on this to run MRVA using query files outside of their workspace.

Would it not be better to have consistent behaviour when there's one query vs multiple queries?

charisk: I could have opened a workspace and dragged a query file from somewhere else into VS Code. So I can imagine that happening, but definitely not often.

I agree it would be good to have the same behaviour but perhaps if we want to change it we should have a specific PR for that, with a CHANGELOG note. WDYT?

extensions/ql-vscode/src/common/files.ts
```diff
@@ -132,3 +132,30 @@ export function isIOError(e: any): e is IOError {
...
   return osTmpdir();
 }
+
+/**
+ * Finds the common parent directory of an arbitrary number of paths. If the paths are absolute,
+ * the result is also absolute. If the paths are relative, the result is relative.
+ * @param paths The array of paths.
+ * @returns The common parent directory of the paths.
+ */
+export function findCommonParentDir(...paths: string[]): string {
```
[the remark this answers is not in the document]
charisk: Thanks for the suggestion. I've updated to use `normalizePath`.

extensions/ql-vscode/src/variant-analysis/ql.ts
charisk: Taken the suggestion!

extensions/ql-vscode/src/variant-analysis/ql.ts
robertbrignull: Sorry that wasn't hugely clear. What I was asking for is to document the intent of the restriction in the code. I think a future developer could work through the code and find the place where we enforce the restriction, but what's useful to document as a comment is why we do this. What benefits does it give and what is it protecting us from?

To be honest, I'm 95% sure it's not a security issue to allow the common parent directory to be outside of the workspace. We don't access or write to any files in that actual parent directory. We construct a new query pack and copy the queries into it using the relative path from the shared parent directory.

I also don't think it will block anybody if we impose this restriction, and I think it's an understandable restriction that people would accept. So I'm not against us applying this restriction.

It's also something we can easily change in the future. It should just be documented so that if we do revisit this decision in the future then we can be more confident we know what we're doing.

extensions/ql-vscode/test/unit-tests/common/files.test.ts
charisk: I've added a few but please let me know if you can think of any more test cases.

extensions/ql-vscode/src/variant-analysis/ql.ts
charisk: Ah gotcha. Done!

extensions/ql-vscode/test/unit-tests/common/files.test.ts
```diff
@@ -417,3 +459,125 @@ describe("walkDirectory", () => {
...
+    ];
+
+    const commonDir = findCommonParentDir(...paths);
+
+    expect(commonDir).toEqualPath(join("/foo", "bar"));
+  });
+
+  it("should return empty path if paths have no common parent", () => {
+    const paths = [
+      join("/foo", "bar", "baz"),
```
robertbrignull, line 498: I think the reason for the test failures on windows is because `join('/foo', 'bar', 'baz')` returns `\foo\bar\baz` without a drive letter, but then we're expecting the common parent dir to be `D:\`.

Would it work to use `join(root, 'foo', 'bar', 'baz')` instead? This will also apply to every other place we use `/` as a prefix to the path.

charisk: Taking https://github.com/github/vscode-codeql/pull/3289#discussion_r1471396830 has made the tests go green 🎉

robertbrignull (approved):
Code looks good to me. If we want to make any more tweaks I think they can be separate PRs. Thankfully everything is tested well so we should be covered if we do change the methods.

I've also briefly tried it out in a workspace with no query packs and it appeared to work just fine.

PR #3290, opened 2024-01-29
Don't accidentally mix DTO and non-DTO types in query-history-store.test.ts

robertbrignull:
I hit some errors in https://github.com/github/vscode-codeql/actions/runs/7695850568/job/20969605471?pr=3287 and this made me realise that `query-history-store.test.ts` is being a bit naughty when writing test data. We're writing the query history file but the data being written is the internal query history types instead of the query history DTO types.

This PR introduces a `writeRawQueryHistory` method which accepts the correct `QueryHistoryDto` type and writes it to disk. Hopefully this will help avoid people making the same mistake in the future by discouraging direct use of `writeFileSync` and writing incorrect JSON.

When I merge this branch into the `robertbrignull/move_language_pack` branch then the `no-workspace` tests pass locally (where they don't before) so that convinces me this change is correct.

## Checklist

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #3293, opened 2024-01-30
Don't compare internal and dto types in variant-analysis-history.test.ts

robertbrignull:
In `variant-analysis-history.test.ts` we are reading the raw query history DTO types from a file and then trying to compare them to the internal types from the query history manager. On `main` this happens to work because the types are the same, but the whole point of having these two types is that they don't have to be the same.

The "workspace-query-history.json" file is where the query history manager also loads its data from. However in the tests we never actually want it as this raw data and instead just want to make sure it's done the loading correctly.

Similar to https://github.com/github/vscode-codeql/pull/3290, I found this because of a test failure on https://github.com/github/vscode-codeql/actions/runs/7699342606/job/20983096500?pr=3287 where the internal types are changing but the DTO types are not.

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/no-workspace/query-history/variant-analysis-history.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/no-workspace/query-history/variant-analysis-history.test.ts
@@ -25,6 +25,8 @@ import { mockedObject } from "../../utils/mocking.helpers";
 import { createMockQueryHistoryDirs } from "../../../factories/query-history/query-history-dirs";
 import { createMockApp } from "../../../__mocks__/appMock";
 import { LanguageContextStore } from "../../../../src/language-context-store";
+import { mapQueryHistoryVariantAnalysisToDomainModel } from "../../../../src/query-history/store/query-history-variant-analysis-dto-mapper";
+import type { VariantAnalysisHistoryItem } from "../../../../src/query-history/variant-analysis-history-item";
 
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/test/vscode-tests/no-workspace/query-history/variant-analysis-history.test.ts
```diff
@@ -72,9 +74,9 @@ describe("Variant Analyses and QueryHistoryManager", () => {
 
     disposables = new DisposableBucket();
 
-    rawQueryHistory = readJSONSync(
+    queryHistory = readJSONSync(
       join(STORAGE_DIR, "workspace-query-history.json"),
-    ).queries;
+    ).queries.map(mapQueryHistoryVariantAnalysisToDomainModel);
```
robertbrignull, line 79: One thing I'm not totally clear on is what exactly we're testing. So is it appropriate to use `mapQueryHistoryVariantAnalysisToDomainModel` to generate the expected data, or are we basically just calling the same function twice and comparing the result?

I also tried writing out the expected data manually and explicitly, instead of loading it from this file. The issues I ran into were:
- we duplicate a lot of data (but that's kinda the point)
- the data contains some `PLACEHOLDER`s in the file paths, because we don't know where the file will be written so we can't hard-code the path to the QL files

charisk (approved):
Seems reasonable!

PR #3301, opened 2024-01-31
Fix Ruby method parsing for methods containing brackets

koesie10:
This fixes parsing of access paths for Ruby methods that contain square brackets, such as `Liquid::Context#[]`. The previous implementation would incorrectly stop capturing at the `]` character, resulting in a method name of `[` for an access path of `Method[[]].ReturnValue`. This fixes it by switching to the shared access path parsing code, which correctly handles square brackets.

## Checklist


shati-patel (approved):
Thanks! Is it worth adding a test for this?

PR #3302, opened 2024-01-31
Stop allowing running MRVA with query outside of workspace

charisk:
We currently allow users to run a MRVA with a query that is not in their workspace. However, this can become problematic, specially when adding multi-query support. 

In this change we introduce a restriction where users cannot run a MRVA if the query is not in their workspace.

## Checklist
TODO:

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -1,7 +1,7 @@
 # CodeQL for Visual Studio Code: Changelog
 
 ## [UNRELEASED]
-- Stop allowing running MRVA with query outside of workspace. [#3302](https://github.com/github/vscode-codeql/pull/3302)
+- Stop allowing running MRVA with a query outside of the workspace. [#3302](https://github.com/github/vscode-codeql/pull/3302)
 
 ## 1.12.1 - 31 January 2024
 
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CodeQL for Visual Studio Code: Changelog
 
 ## [UNRELEASED]
+- Stop allowing running MRVA with query outside of workspace. [#3302](https://github.com/github/vscode-codeql/pull/3302)
 
 ## 1.12.1 - 31 January 2024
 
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/test/unit-tests/common/files.test.ts
```diff
@@ -580,4 +572,23 @@ describe("findCommonParentDir", () => {
 
     expect(commonDir).toEqualPath(rootDir);
   });
+
+  it("should return the parent dir of a single file", async () => {
+    const dataDir = join(__dirname, "../../data");
```
robertbrignull, line 577: Hmm, this works, but there is a non-zero chance it'll cause confusing errors for us in the future. The `extensions/ql-vscode/test/data` directory and `extensions/ql-vscode/test/data/query.ql` file aren't really related to this test, so someone could easily move them and not realise it'll affect these tests.

I suppose maybe it's ok because if those files are moved then these tests will immediately start failing by calling `lstatSync` on a path that doesn't exist. So maybe it's fine, but I wanted to mention it.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # CodeQL for Visual Studio Code: Changelog
 
 ## [UNRELEASED]
+- Stop allowing running MRVA with query outside of workspace. [#3302](https://github.com/github/vscode-codeql/pull/3302)
```
robertbrignull, line 4:
```suggestion
- Stop allowing running MRVA with a query outside of the workspace. [#3302](https://github.com/github/vscode-codeql/pull/3302)
```
or 
```suggestion
- Stop allowing running MRVA with queries outside of the workspace. [#3302](https://github.com/github/vscode-codeql/pull/3302)
```
?

robertbrignull (approved):
LGTM assuming we have agreement that we want this behaviour change

extensions/ql-vscode/test/unit-tests/common/files.test.ts
charisk: There are many places where we do this so this wouldn't be the only test that would be affected. It's not ideal, but that's how we've done things so far.

PR #3304, opened 2024-01-31
Send access path suggestions to the model editor view

shati-patel:
If you have the relevant feature flag enabled, the model editor will now generate and load access path suggestions on start-up. 

I haven't yet added logic to the webview to actually _use_ those suggestions, but they will eventually be used to display autocomplete suggestions! 

See internal linked issue for more info 🔍 

## Checklist

N/A 

- [ ] [CHANGELOG.md](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) has been updated to incorporate all user visible changes made by this pull request.
- [ ] Issues have been created for any UI or other user-facing changes made by this pull request.
- [ ] _[Maintainers only]_ If this pull request makes user-facing changes that require documentation changes, open a corresponding docs pull request in the [github/codeql](https://github.com/github/codeql/tree/main/docs/codeql/codeql-for-visual-studio-code) repo and add the `ready-for-doc-review` label there.

PR #3305, opened 2024-02-01
CodeQL model editor: Show access path suggestions in the webview

shati-patel:
Reads the access path suggestions (from #3304 and previous PRs) in the webview and shows them in a "suggest box":

image

This is only enabled for Ruby, so the public model editor (i.e. for Java & C#) is unaffected. See internal linked issue for more details 🔗

❓  Should this have a view test? The `SuggestBox` component is already tested, so I wasn't sure if it made sense to test e.g. the `ModelInputSuggestBox` separately too.

## Checklist

N/A—still feature-flagged for internal use 🦞

### DIFF
--- a/extensions/ql-vscode/src/view/common/SuggestBox/SuggestBox.tsx
+++ b/extensions/ql-vscode/src/view/common/SuggestBox/SuggestBox.tsx
@@ -23,8 +23,7 @@ import type { Diagnostic } from "./diagnostics";
 import { useOpenKey } from "./useOpenKey";
 
 const Input = styled(VSCodeTextField)<{ $error: boolean }>`
-  width: 430px;
-
+  width: 100%;
   font-family: var(--vscode-editor-font-family);
 
   ${(props) =>
@@ -36,7 +35,6 @@ const Input = styled(VSCodeTextField)<{ $error: boolean }>`
 `;
 
 const Container = styled.div`
-  width: 430px;
   display: flex;
   flex-direction: column;
   border-radius: 3px;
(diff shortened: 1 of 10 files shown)


extensions/ql-vscode/src/view/common/SuggestBox/SuggestBox.tsx
```diff
@@ -36,7 +35,6 @@ const Input = styled(VSCodeTextField)<{ $error: boolean }>`
 `;
 
 const Container = styled.div`
-  width: 430px;
   display: flex;
```
shati-patel, line 38: This was a styling fix to stop the model editor looking like this 😅

image

extensions/ql-vscode/src/view/model-editor/ModelInputSuggestBox.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  return (
+    <SuggestBox<AccessPathOption, AccessPathDiagnostic>
+      value={value}
+      onChange={setValue}
+      options={suggestions}
+      parseValueToTokens={(value) =>
+        parseAccessPathTokens(value).map((t) => t.text)
+      }
```
koesie10, line 78: Can we move this to a separate function/constant defined outside of the component so React doesn't think this property changes on every render?

koesie10 (commented):
Would we already expect the icon and details to be visible in the suggestions as well? It currently seems to be missing (probably because we're not passing `getIcon` and `getDetails`).

Screenshot 2024-02-01 at 14 02 42

Also, it seems like for some type models, we're not able to edit anything, while for others, we are. This might be unrelated to your changes.

Screenshot 2024-02-01 at 14 01 23

I think we can also use the suggest box for the type path (i.e. the text field shown in the input field for type models), but this would need to use the `output` suggestions (since the `ReturnValue` is valid for type models). This can also be done in a follow-up PR.

koesie10 (approved):
LGTM, thanks!

PR #3306, opened 2024-02-01
Remove bundled Ruby queries

koesie10:
This removes the bundled Ruby queries since these queries are in a CodeQL release now, so we should never find that the queries don't exist.

## Checklist


shati-patel (approved):
Thanks for the clean-up!

charisk (approved):
Nice! I like that we're removing the queries but still leaving support for them, as we'll very likely need it in the future.

PR #3315, opened 2024-02-05
Execute a distribution check when changing the CLI channel

koesie10:
When you change the CLI channel setting, there is no distribution check, so the user doesn't get prompted to update their CLI to for example the latest nightly. This adds the setting as a setting that forces a distribution check so the user will get the prompt after changing the setting.

## Checklist


robertbrignull (approved):
:bow:

PR #3317, opened 2024-02-06
Fix bug in SARIF comparison

koesie10:
The SARIF comparison code was comparing the index of the artifact location, which is not useful for comparison and may differ between runs of very similar queries. This adds a function to convert a SARIF result to a canonical form, which removes the index from the artifact location.

## Checklist


robertbrignull (approved):
Looks reasonable to me.

If we find we're wanting to do more of these modifications then I wonder if it's worth investing in a real `deepCopy` function so the code is simpler. For just this I think what's in this PR is good.

PR #3318, opened 2024-02-06
Make query constraints for access paths query configurable

koesie10:
This makes the query constraints for the access paths query configurable since these can differ between the languages. It also updates the query constraint for the Ruby query from the `access-path-suggestions` tag to `access-paths`, since that is [how the query was merged into CodeQL `main`](https://github.com/github/codeql/pull/15503/files#diff-891818fe6372315cfd5d569d6ec1dc4cdbd9614fab0306a14f8d98afeb72e2fdR6).

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/languages/models-as-data.ts
+++ b/extensions/ql-vscode/src/model-editor/languages/models-as-data.ts
@@ -44,6 +44,7 @@ type ModelsAsDataLanguageModelGeneration = {
 };
 
 type ModelsAsDataLanguageAccessPathSuggestions = {
+  queryConstraints: (mode: Mode) => QueryConstraints;
   parseResults: (
     // The results of a single predicate of the query.
     bqrs: DecodedBqrsChunk,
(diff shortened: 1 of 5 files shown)


shati-patel (approved):
Makes sense! 🚢 

I guess this just means that we temporarily can't use the access paths query in the extension? (Since the "old" version used a different tag, and this new version doesn't have the correct output format for now)

koesie10:
> I guess this just means that we temporarily can't use the access paths query in the extension? (Since the "old" version used a different tag, and this new version doesn't have the correct output format for now)

Yes, but you'd have to have manually added the "old" query to your workspace, so the access path queries were probably not working for the majority of people anyway. You should still be able to use the old query by changing the `access-path-suggestions` tag to `access-paths` and not updating the CodeQL submodule to `main`. After https://github.com/github/vscode-codeql/pull/3320 is merged, you'd then need to remove the old query and update the submodule (or copy the new query).

PR #3320, opened 2024-02-06
Parse new access paths query BQRS output

koesie10:
This parses the new access paths query BQRS output introduced in https://github.com/github/codeql/pull/15503. Somewhat dependent on https://github.com/github/vscode-codeql/pull/3318 for resolving the correct query.

The only difference in the output format is the fourth column (`details`), which was a string before, but has now been renamed to `node` and is an `Entity`. The same information is available in `node.label`, so that's the only change this makes (in addition to validating this new format).

To test this, you can use the `github/codeql` repository on `main` or update a submodule in another workspace to `main`.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/languages/ruby/suggestions.ts
+++ b/extensions/ql-vscode/src/model-editor/languages/ruby/suggestions.ts
@@ -1,9 +1,38 @@
 import type { BaseLogger } from "../../../common/logging";
-import type { DecodedBqrsChunk } from "../../../common/bqrs-cli-types";
+import type {
+  BqrsCellValue,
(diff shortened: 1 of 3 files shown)


shati-patel:
Looks good from the code, but could you merge in the latest changes from `main`? Then we can test everything together with the new query constraints! 🎉

shati-patel (approved):
Nice! Can confirm it works with codeql on `main` now ✨ 

Btw, not related to this PR, just a general observation—the suggestions are generally still pretty laggy for me. Potentially just a side effect of having a slower machine, but we might want to check if there's any ways to improve that! 🤔

PR #3322, opened 2024-02-07
Fix not being able to open command palette in suggest box

koesie10:
This fixes not being able to open the command palette using Cmd/Ctrl+Shift+P when in a suggest box. Documentation about the `returnFocus` prop can be found here: [`returnFocus`](https://floating-ui.com/docs/floatingfocusmanager#returnfocus).

The problem seems to be that if you press Cmd/Ctrl+Shift+P, focus moves to the `body` element. Floating UI then tries to focus the suggest box again, which removes the focus from the command palette. By adding this prop and setting it to `false` (the default is `true`), Floating UI doesn't try to regain focus. This results in a somewhat worse experience when returning from the command palette, but is still preferable to not being able to open the command palette at all.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/view/common/SuggestBox/SuggestBox.tsx
+++ b/extensions/ql-vscode/src/view/common/SuggestBox/SuggestBox.tsx
@@ -243,6 +243,7 @@ export const SuggestBox = <
               context={context}
               initialFocus={-1}
               visuallyHiddenDismiss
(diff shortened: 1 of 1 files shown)


charisk (approved):
Good catch! I think the change makes sense.

Shall we add a code comment so people don't have to go digging into git history?

Issue #3219, opened 2024-01-10
Creating skeleton query pack logs error


Marcono1234:
**Describe the bug**
When creating a skeleton query pack using the extension, the following error is logged to the "CodeQL Extension Log":
> Could not create skeleton QL pack: Parsing output of Adding and installing java pack dependency. failed: Unexpected token 'P', "Package in"... is not valid JSON

However, it seems adding the pack was actually successful, but the extension erroneously expected JSON output from the CodeQL CLI when the output seems to be plaintext (?).

**Version**
Extension:
```
CodeQL extension version: 1.11.0 
CodeQL CLI version: 2.15.5 
Platform: win32 x64
```

VS Code:
```
Version: 1.85.1 (user setup)
Commit: 0ee08df0cf4527e40edc9aa28f4b5bd38bbff2b2
Date: 2023-12-13T09:49:37.021Z
Electron: 25.9.7
ElectronBuildId: 25551756
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Windows_NT x64 10.0.19045
```

**To reproduce**
1. Open a new empty folder with VS Code
2. In the CodeQL tab under "Queries" click "Create one to get started", or alternatively run the command "> CodeQL: Create Query"
3. Select "Java" as language
4. Open the "CodeQL Extension Log" and scroll to the end
:x: Bug: You see the error mentioned above being logged

**Expected behavior**
No error should be logged

**Additional context**
In general the query pack logging output seems to be quite verbose. Is that intended?
Probably not a big problem though because I assume users are normally not expected to look at "CodeQL Extension Log" except for troubleshooting.

shati-patel:
Thank you for the issue @Marcono1234! You're right that we're incorrectly expecting JSON output in the [`packAdd`](https://github.com/github/vscode-codeql/blob/83148f6d576e40d493921b4a371a0714ec471d53/extensions/ql-vscode/src/codeql-cli/cli.ts#L1382) command.

We'll take a proper look when time permits 👀

PR #3323, opened 2024-02-07
Don't try to parse output of `codeql pack add` as JSON

shati-patel:
Fixes https://github.com/github/vscode-codeql/issues/3219.

The `packAdd` command is only used in the "Create query" flow, where it was causing some confusing/useless log messages. This is because we were incorrectly expecting JSON output.

I've tested this by running "Create query" and checking that it still works, but _doesn't_ display the following error message 
> Could not create skeleton QL pack: Parsing output of Adding and installing java pack dependency. failed: Unexpected token 'P', "Package in"... is not valid JSON

## Checklist

  - I don't think this a changelog update it necessary? It just removes an errant log message 🤔  


charisk (approved):
Makes sense!

PR #3325, opened 2024-02-07
Delete src/additional-typings.d.ts

robertbrignull:
I'm looking at this file because I'm trying to remove uses of `any`. In this case it's not clear whether to change it to `unknown` or delete the entire file.

The comment in this file says it is needed to make compilation succeed, but if I delete it then compilation still works. Therefore hopefully that means this file is no longer needed. It was introduced back in https://github.com/github/vscode-codeql/pull/1111 and hasn't been touched since, and we've changed the build system quite a bit since then.

I've tried using the graph viewer to see if it still works and unfortunately I can't get it to work on `main` or this branch. In both cases I get a blank screen 🤷🏼 

Screenshot 2024-02-07 at 13 01 04

@aeisenberg, @hvitved, @nickrolfe, I believe you were the originators of the graph viewer. What do you think?

## Checklist


shati-patel:
> I've tried using the graph viewer to see if it still works and unfortunately I can't get it to work on `main` or this branch. In both cases I get a blank screen 🤷🏼

I can confirm it works on `main` and on this branch for me 🎉 
image

It looks like your graph is trying to use the wrong query 🤔 (`printAst` instead of `printCfg`)?

aeisenberg:
I think this is safe since this is only about making `tsc` happy. I'm not sure why it is working now, but wasn't before. Possibilities:

1. Have we updated d3 to a newer version so the typings are more flexible?
2. Have we changed how we compile things so that the missing typings are no longer an error?

Either way, if it's compiling, then it's safe to remove.

aeisenberg (approved):

robertbrignull:
Thanks both for investigating and confirming. I agree if tsc is happy then it should be fine. I just wanted to check.

PR #3326, opened 2024-02-07
Add endpoint type to method definition

koesie10:
This adds an `endpointType` property to method definitions since we already have different endpoints (for example [methods](https://github.com/github/codeql/blob/ebefcb8d35cc71ed7a17de03c84004843dc808ab/ruby/ql/src/utils/modeleditor/ModelEditor.qll#L58) and [modules](https://github.com/github/codeql/blob/ebefcb8d35cc71ed7a17de03c84004843dc808ab/ruby/ql/src/utils/modeleditor/ModelEditor.qll#L222) in Ruby) and it would be useful to make a distinction between these (it's not possible to model a module as a source/sink/summary, but it should be possible to model as a type).

In tests, I've tried to use `satisfies` in all places where this wasn't already checked to ensure we get type errors instead of failing tests for future changes to these types.

## Checklist


extensions/ql-vscode/src/model-editor/languages/ruby/index.ts
```diff
@@ -37,11 +38,11 @@ export const ruby: ModelsAsDataLanguage = {
         );
         return {
           type: "source",
-          input: "",
```
shati-patel, line 40: (Minor, just so I understand)
Is removing these input/output fields related to the `endpointType`s? Or is this general clean up?

extensions/ql-vscode/src/stories/model-editor/LibraryRow.stories.tsx
```diff
@@ -22,6 +22,7 @@ LibraryRow.args = {
       library: "sql2o",
       libraryVersion: "1.6.0",
       signature: "org.sql2o.Connection#createQuery(String)",
+      endpointType: EndpointType.Method,
```
shati-patel, line 25: I don't know what the `EndpointType.Class` should "look" like. Is it worth having an example of that one somewhere? (e.g. in stories or in tests)

shati-patel (approved):
A couple of minor comments, mainly due to my limited understanding of how these models work 😅 LGTM otherwise though 👍🏽

extensions/ql-vscode/src/model-editor/languages/ruby/index.ts
koesie10: This is just general clean up, but I believe it had to do with `input` and `output` not being in some of the test cases after adding `satisfies`, so it might have been necessary.

extensions/ql-vscode/src/stories/model-editor/LibraryRow.stories.tsx
koesie10: I think that will come naturally as the endpoint types start getting used, so I don't think it's necessary for now.

PR #3329, opened 2024-02-07
Use type 'string | Buffer' for process stdout/stderr

robertbrignull:
The intention here is to avoid use of `any` when logging process stdout/stderr. This data comes to us in chunks which we want to call `toString()` on. We can do that with type safety if we introduce a `asString` method (very open to other names).

The only issues are the `null` and `undefined` values. In this I've gone with returning human-readable values since we're printing this to a log, but we could return the empty string instead. In practice I expect these values to never actually happen.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -19,6 +19,7 @@ import type { CliConfig } from "../config";
 import type { DistributionProvider } from "./distribution";
 import { FindDistributionResultKind } from "./distribution";
 import {
+  asString,
   assertNever,
   getErrorMessage,
   getErrorStack,
@@ -1605,8 +1606,8 @@ export function spawnServer(
   command: string[],
   commandArgs: string[],
   logger: Logger,
-  stderrListener: (data: any) => void,
-  stdoutListener?: (data: any) => void,
+  stderrListener: (data: unknown) => void,
+  stdoutListener?: (data: unknown) => void,
   progressReporter?: ProgressReporter,
 ): ChildProcessWithoutNullStreams {
   // Enable verbose logging.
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1605,8 +1606,8 @@ export function spawnServer(
   command: string[],
   commandArgs: string[],
   logger: Logger,
-  stderrListener: (data: any) => void,
-  stdoutListener?: (data: any) => void,
+  stderrListener: (data: unknown) => void,
+  stdoutListener?: (data: unknown) => void,
```
koesie10, line 1610: I think we can change this to more strict types since [the documentation mentions that this can only be a string or `Buffer`](https://nodejs.org/api/stream.html#event-data), unless the stream is operating in object mode. A child process's stdout or stderr will never operate in object mode, so we should be able to change this to:

```suggestion
  stderrListener: (data: string | Buffer) => void,
  stdoutListener?: (data: string | Buffer) => void,
```

I don't get any TypeScript errors locally when doing this, so it seems like narrowing the types here should work without giving any problems. I don't think we need the `asString` function in that case.

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1626,7 +1627,7 @@ export function spawnServer(
     );
   }
 
-  let lastStdout: any = undefined;
+  let lastStdout: unknown = undefined;
```
koesie10, line 1630:
```suggestion
  let lastStdout: string | Buffer = undefined;
```

extensions/ql-vscode/src/codeql-cli/cli.ts
robertbrignull: Thanks for linking to the docs. That makes sense to me. It's a shame they don't have separate type information for non-object mode streams, and therefore all streams have to default to `any`.

I'll take a look at using `string | Buffer` and get rid of the `asString` method.

koesie10 (approved):
Thanks, LGTM!

PR #3330, opened 2024-02-08
Remove use of any and make typing clearer in debugger code

robertbrignull:
This aims to remove the use of `any` from `debug-configuration.ts` and `debug-protocol.ts`. The first commit does this by replacing instances of `any` with `unknown`, and this appears to work fine.

The second commit tries to improve type clarity by reusing the `NotArray` type. I'm not 100% sure but I think this is what `makeArray` is trying to achieve with `T extends Exclude<any, any[]>`. Using a named type there and reusing code instead of having two hacks for the same thing will hopefully make it easier to understand in the future.

As part of this I had to augment `NotArray` to handle other primitive types instead of just `object`, but I don't think this will have any negative effect on `selection-commands.ts`.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/helpers-pure.ts
+++ b/extensions/ql-vscode/src/common/helpers-pure.ts
@@ -7,6 +7,18 @@
 
 import { RedactableError } from "./errors";
 
+// Matches any type that is not an array. This is useful to help avoid
+// nested arrays, or for cases like createSingleSelectionCommand to avoid T
+// accidentally getting instantiated as DatabaseItem[] instead of DatabaseItem.
+export type NotArray =
+  | string
+  | bigint
+  | number
+  | boolean
+  | (object & {
+      length?: never;
+    });
+
 /**
  * This error is used to indicate a runtime failure of an exhaustivity check enforced at compile time.
  */
--- a/extensions/ql-vscode/src/common/vscode/selection-commands.ts
+++ b/extensions/ql-vscode/src/common/vscode/selection-commands.ts
@@ -3,13 +3,9 @@ import type {
   TreeViewContextMultiSelectionCommandFunction,
   TreeViewContextSingleSelectionCommandFunction,
 } from "../commands";
+import type { NotArray } from "../helpers-pure";
 import type { NotificationLogger } from "../logging";
 import { showAndLogErrorMessage } from "../logging";
-
-// A hack to match types that are not an array, which is useful to help avoid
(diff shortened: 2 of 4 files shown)


extensions/ql-vscode/src/common/helpers-pure.ts
```diff
@@ -7,6 +7,18 @@
 
 import { RedactableError } from "./errors";
 
+// Matches any type that is not an array. This is useful to help avoid
+// nested arrays, or for cases like createSingleSelectionCommand to avoid T
+// accidentally getting instantiated as DatabaseItem[] instead of DatabaseItem.
+export type NotArray =
```
charisk, line 13: How about this insted?
```ts
type NonArray<T> = T extends any[] ? never : T;
```

robertbrignull: Unfortunately I don't think this works, unless I've made a mistake in my testing.

Note I had to make a couple of modifications:
- We don't want to use `any`, so I changed that to `T extends unknown[]`
- Having `NotArray` have a type argument itself is a little awkward. Using it could force us to do something like `foo<T extends NotArray<S>, S>` which is weird because `foo` now has an extra type arg it doesn't really need. Instead I went with `NotArray<T = unknown` to provide a default so it can be used just as `NotArray`.

Sadly it doesn't appear to be working. I can do `const x: NotArray = [1, 2, 3];` and it doesn't cause any errors, while it does cause an error with the `NotArray` definition in this PR.

robertbrignull: Hmm, I now see that `NotArray<T = unknown>` basically means the ternary compiles down to `T = unknown`, so that isn't very helpful. That explains why it wasn't causing an error in my test.

Unfortunately I still don't see how it make it work as intended with this approach.

@charisk, do you have any explicit issues with the current approach in this PR that should be changed?

charisk: Ah okay, not a great suggestion then. I was just not super keen on having to define every type that isn't an array, but it's not bad.

charisk (approved):

extensions/ql-vscode/src/common/helpers-pure.ts
robertbrignull: At least it's not a very long list, and it isn't dependent on anything we do in our code so it shouldn't ever have to change.

PR #3341, opened 2024-02-12
Use unknown instead of any in type functions

robertbrignull:
When checking if an object is a certain type, we should use `unknown` instead of `any`. The only difference this makes to the implementation is we need to add some extra `obj !== null` and `typeof obj === "object"` checks, but this makes it safer anyway since before if some of these were passed a `null` value they would throw an error when trying to access fields on it.

## Checklist


koesie10 (approved):
Thanks!

PR #3344, opened 2024-02-12
Use correct types where possible instead of any

robertbrignull:
This is all the cases I could find where we are using `any` but there exists some other type we can use instead that doesn't involve changing anything else. If everything still compiles then this should be safe to do so.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases/database-fetcher.ts
+++ b/extensions/ql-vscode/src/databases/database-fetcher.ts
@@ -590,7 +590,7 @@ export async function convertGithubNwoToDatabaseUrl(
       repo,
     });
 
-    const languages = response.data.map((db: any) => db.language);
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/databases/database-fetcher.ts
```diff
@@ -590,7 +590,7 @@ export async function convertGithubNwoToDatabaseUrl(
       repo,
     });
 
-    const languages = response.data.map((db: any) => db.language);
+    const languages = response.data.map((db) => db.language);
```
robertbrignull, line 593: In this case the correct type is inferred, so we're not just getting an implicit `any` type here.

shati-patel (approved):
Looks good, apart from the `import [type] ...` linter errors!

PR #3349, opened 2024-02-13
Add new type parameters to functions where possible

robertbrignull:
These are a few functions that currently use `any` but could instead have a type parameter.

I think each of these cases increases our type safety a little bit. I can't see any downsides, unless I've missed an edge case.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/vscode/commands.ts
+++ b/extensions/ql-vscode/src/common/vscode/commands.ts
@@ -40,13 +40,13 @@ export function createVSCodeCommandManager<
  * @param logger The logger to use for error reporting.
  * @param telemetry The telemetry listener to use for error reporting.
  */
-export function registerCommandWithErrorHandling(
+export function registerCommandWithErrorHandling<S extends unknown[], T>(
   commandId: string,
-  task: (...args: any[]) => Promise<any>,
+  task: (...args: S) => Promise<T>,
   logger: NotificationLogger = extLogger,
   telemetry: AppTelemetry | undefined = telemetryListener,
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/common/vscode/commands.ts
```diff
@@ -40,13 +40,13 @@ export function createVSCodeCommandManager<
  * @param logger The logger to use for error reporting.
  * @param telemetry The telemetry listener to use for error reporting.
  */
-export function registerCommandWithErrorHandling(
+export function registerCommandWithErrorHandling<S extends unknown[], T>(
```
robertbrignull, line 43: It's annoying that `registerCommandWithErrorHandling` gained two args, but thankfully in all the cases we use it the types are inferred so we don't need to do anything. In reality in both uses the type could be `() => void`, so I wouldn't be against just hardcoding that here for simplicity, and we can revisit later if this function needs to handle more complex functions.

charisk: I'm happy either way

charisk (approved):

extensions/ql-vscode/src/common/vscode/commands.ts
robertbrignull: I've had another go at this and condensed the two type args into one. I don't know why I didn't do this before. Maybe it didn't work before and some other change made it possible, or maybe I just didn't do it properly.

I think this is a better solution overall and makes things clearer.

charisk (approved):

PR #3350, opened 2024-02-13
Add unit tests for "short-paths" functions

shati-patel:
Follow-up to https://github.com/github/vscode-codeql/pull/3233/files#r1455336296, where we suggested adding some unit tests. 

The functionality is only relevant on Windows, so I've skipped most of the tests on ubuntu 🪟 I've added a couple of tests so far, based on examples from the [8.3 filename Wikipedia page](https://en.wikipedia.org/wiki/8.3_filename#VFAT_and_computer-generated_8.3_filenames), but I'm open to more suggestions! E.g. I couldn't work out how to test the [stuff around `inode`s and symlinks](https://github.com/github/vscode-codeql/blob/d7e9606bfab6108156276245d14ced7b3dbd868e/extensions/ql-vscode/src/common/short-paths.ts#L54-L84) 😵

## Checklist

N/A—only affects tests and test data

### DIFF
--- a/extensions/ql-vscode/test/unit-tests/common/short-paths.test.ts
+++ b/extensions/ql-vscode/test/unit-tests/common/short-paths.test.ts
@@ -1,0 +1,96 @@
+import { platform } from "os";
+import type { BaseLogger } from "../../../src/common/logging";
+import { expandShortPaths } from "../../../src/common/short-paths";
+import { join } from "path";
+
+describe("expandShortPaths", () => {
+  let logger: BaseLogger;
+
+  beforeEach(() => {
+    logger = {
+      log: jest.fn(),
+    };
+  });
+
+  describe("on POSIX", () => {
+    if (platform() === "win32") {
+      console.log(`Skipping test on Windows`);
+      return;
+    }
+
+    it("should return the same path for non-Windows platforms", async () => {
+      const path = "/home/user/some~path";
+      const result = await expandShortPaths(path, logger);
+
+      expect(logger.log).not.toHaveBeenCalled();
+      expect(result).toBe(path);
+    });
+  });
+
+  describe("on Windows", () => {
+    if (platform() !== "win32") {
+      console.log(`Skipping test on non-Windows`);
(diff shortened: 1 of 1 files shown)


shati-patel:
I'm merging this for now, since these are fairly uncontroversial unit tests 😄 I'm still happy for follow-up comments from @dbartol or anyone else of course!

dbartol:
@shati-patel The reason that there were no unit tests in the original PR is that testing short path expansion is trickier than one would expect.

First, short path generation can be enabled or disabled per file system volume. Some file systems, including the new ReFS file system used by the [Dev Drive feature](https://learn.microsoft.com/en-us/windows/dev-drive/) does not support 8.3 filename generation at all. If short path generation is disabled, none of the test cases that expand an actual short path will find a short path to expand.

Second, the expansion of a given short path depends on the state of the file system. The path component `PROGRA~1` could expand to `Program Files`, `Program Files (x86)`, or `Prog radishes play in odd thyme signatures`. It all depends on which order the files with the long path names were created in. The first would typically be `PROGRA~1`, the second `PROGRA~2`, etc. So, any test that tries to expand a particular existing short path and expects to get a particular long path is making an unwarranted assumption about the underlying file system.

Your test cases that verify that we don't try to expand any already-long paths are definitely fine. The test cases that try to expand, say, `C:\PROGRA~1` to `C:\Program Files` could fail the `C:` drive volume has long path generation disabled, or if `C:\PROGRA~1` is the short name for some other file in the root directory, or if `Program Files` is on a different drive for that particular machine.

In practice I expect that these tests will still pass for most Windows machines, because AFAICT:
- It's difficult to install Windows such that the `Program Files` directory is not at `C:\Program Files`
- It's (currently) difficult to install Windows with a ReFS system drive
- Default installations of Windows still seem to have short path name generation enabled on the system drive by default.

All that said, I think we should leave your tests as-is unless/until someone find a machine on which they don't work. We probably never will, unless some future version of Windows disables 8.3 generation by default.

PR #3352, opened 2024-02-13
Add support for Python in the model editor

koesie10:
This adds support for parsing and constructing Python access paths in the model editor. It does not yet add support for type models since I'm not quite sure about the format for those yet. It also does not add support for `Method[foo]` tokens since I'm not sure how these are used (they are different from `Member[foo]`).

To test this, you can use the CodeQL branch [`koesie10/model-editor-python`](https://github.com/github/codeql/tree/koesie10/model-editor-python) with for example a `psf/requests` database.

## Checklist


extensions/ql-vscode/src/model-editor/languages/python/index.ts
```diff
@@ -0,0 +1,207 @@
...
+    `${typeName}#${methodName}`,
+  endpointTypeForEndpoint: (method) => pythonEndpointType(method),
+  predicates: {
+    source: {
+      extensiblePredicate: sharedExtensiblePredicates.source,
+      supportedKinds: sharedKinds.source,
+      supportedEndpointTypes: [EndpointType.Method, EndpointType.Function],
+      // extensible predicate sourceModel(
+      //   string type, string path, string kind
+      // );
```
shati-patel, line 26: Not necessarily related to this PR (since the other languages do the same), but I was wondering what all these commented-out extensible predicates are for 🤔 Are we waiting on a new feature before these can be used?

shati-patel (commented):
I don't feel qualified to say much about the modeling and access path parsing themselves, but I was just trying to take a look more generally 😅 

Should the model editor work fully with this change? 👀 I can open the editor, but as soon as I try to expand a row, I get a "No models-as-data definition for python" error and the webview clears (same thing happens for `psf/requests`): 

https://github.com/github/vscode-codeql/assets/42641846/9254fc74-0430-4d3d-bb8c-3be49d0800e4

If this is unexpected, I might need to check my setup (or potentially see if it's just a slow Windows-machine thing 🤔 )

extensions/ql-vscode/src/model-editor/languages/python/index.ts
koesie10: These are basically just reminders for what the extensible predicates look like since this is very useful to know for knowing what `row[0]`, `row[1]`, etc. refer to. In this case, this is referring to [this definition](https://github.com/github/codeql/blob/fb2d36ddac8b719f7679eb458383d61706507a01/python/ql/lib/semmle/python/frameworks/data/internal/ApiGraphModelsExtensions.qll#L11).

koesie10:
> Should the model editor work fully with this change? 👀 I can open the editor, but as soon as I try to expand a row, I get a "No models-as-data definition for python" error and the webview clears (same thing happens for `psf/requests`):

Can you try checking out the branch first, and then running `npm run build`? It seems like there is a bug in the watch process which causes some files to not be recompiled for the view (fixed by https://github.com/github/vscode-codeql/pull/3355), so some changes do not show up in the webview until you re-run `npm run build`.

shati-patel (approved):
> Can you try checking out the branch first, and then running `npm run build`? It seems like there is a bug in the watch process which causes some files to not be recompiled for the view (fixed by #3355), so some changes do not show up in the webview until you re-run `npm run build`.

That works, thanks! I've been able to interact with the model editor as expected now, so approving from the UI point of view at least 😄 

image

PR #3353, opened 2024-02-13
Enable no-explicit-any throughout production code

robertbrignull:
This PR contains:
- the final changes to enable `@typescript-eslint/no-explicit-any` on all production code (i.e. not tests or build code)
- fixes for any uses of `any` where I couldn't find a good solution so I've marked then as ignored

## Checklist


extensions/ql-vscode/src/packages/commands/CommandManager.ts
```diff
@@ -10,6 +10,7 @@ import type { Disposable } from "./Disposable";
 /**
  * A command function is a completely untyped command.
  */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
robertbrignull, line 13: I couldn't find a good way around this so ended up just ignoring the eslint error. Admittedly I didn't try super hard, but I didn't want to end up with a solution that was unnecessarily complex either.

extensions/ql-vscode/src/view/common/SuggestBox/useEffectEvent.ts
```diff
@@ -9,6 +9,7 @@ import { useCallback, useInsertionEffect, useRef } from "react";
  *
  * @param callback The callback to call when the event is triggered.
  */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
robertbrignull, line 12: We could rewrite this function to avoid `any` but I'm hesitant because I don't fully understand what it's doing and it's copied verbatim from another source. We could alternatively rewrite this.

extensions/ql-vscode/src/view/results/ResultsApp.tsx
```diff
@@ -173,7 +173,7 @@ export function ResultsApp() {
               },
               selectedTable: tableName,
             },
-            origResultsPaths: undefined as any, // FIXME: Not used for interpreted, refactor so this is not needed
+            origResultsPaths: undefined as unknown as ResultsPaths, // FIXME: Not used for interpreted, refactor so this is not needed
```
robertbrignull, line 176: This whole situation in `ResultsApp.tsx` is a bit unfortunate. There are a few arguments where we pass `undefined` or other "fake" values. I briefly looked at trying to do the refactoring that this comment asks for but I found this area of code very confusing. For now I think doing `as unknown as ResultsPaths` gets the same effect and isn't any worse than `as any`. 🤷🏼

extensions/ql-vscode/gulpfile.ts/.eslintrc.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  overrides: [
+    {
+      files: ["*"],
+      rules: {
+        "@typescript-eslint/no-explicit-any": "off",
+      },
+    },
+  ],
+};
```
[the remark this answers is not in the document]
robertbrignull: Oh good point, you can have multiple overrides for different sets of file paths in the same `.eslintrc.js` file. Somehow I missed that 🙈 . I'll add them there instead of in separate files.

koesie10 (approved):

PR #3354, opened 2024-02-13
Add a CLI test for MRVA with multiple queries

robertbrignull:
This PR adds a simple test for running MRVA with two queries. This was actually quite easy because we can re-use the `doVariantAnalysisTest` method with only very small changes.

There are more edge cases of course, but I think they're mostly covered by the no-workspace integration tests. In the CLI tests which are even slower to run, I hope it's ok to just run a single MRVA with two queries and that shows that the process works.

There is already a test for running queries from a published pack, so we don't need to add anything more for that.

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
@@ -343,6 +343,22 @@ describe("Variant Analysis Manager", () => {
       });
     });
 
+    it("should run multiple queries that are part of the same pack", async () => {
+      await doVariantAnalysisTest({
+        queryPaths: [
+          "data-qlpack-multiple-queries/query1.ql",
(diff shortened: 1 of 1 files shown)


charisk:
Generally looking good. Let's make sure the tests pass in the various versions of the CLI that we expect them too. Would be good to run the [CLI tests workflow](https://github.com/github/vscode-codeql/actions/workflows/cli-test.yml) against your branch.

robertbrignull:
> Let's make sure the tests pass in the various versions of the CLI that we expect them too. Would be good to run the [CLI tests workflow](https://github.com/github/vscode-codeql/actions/workflows/cli-test.yml) against your branch.

Good point the test should only run on supported CLI versions.

I'm a bit confused why it's currently failing on the latest CLI version as it passed locally. I'll take a look.

robertbrignull:
Run of all CLI tests: https://github.com/github/vscode-codeql/actions/runs/7900642613

robertbrignull:
The CLI tests are all green. Should be ready for re-review now.

charisk (approved):
🚢

PR #3355, opened 2024-02-14
Recompile view when shared files change

koesie10:
When you change a file not in the view directory, but which is imported by the view directory (for example `src/model-editor/languages/languages.ts`, the view would not be recompiled while these do have an impact on the compiled output. This is very surprising behavior and not what we would expect: I expect a change to any file to immediately be visible in the view as well, without having to re-run `npm run build` or `npm run watch`. This fixes it by watching for all changes in the `src` directory.

## Checklist


charisk (approved):
Thanks!

PR #3362, opened 2024-02-14
Replace deprecated telemetry setting in the tests

shati-patel:
In https://github.com/github/vscode-codeql/pull/2824, we added support for the `telemetry.telemetryLevel` setting.
However, our [tests](https://github.com/github/vscode-codeql/blob/0f4a7788a70e1a9ff552cfbb1589982ce62bc29b/extensions/ql-vscode/test/vscode-tests/no-workspace/telemetry.test.ts#L76) still tested the deprecated `telemetry.enableTelemetry` setting, so I've now replaced those checks with `telemetry.telemetryLevel`. I _could_ have kept tests for the deprecated setting as well, but it didn't seem worthwhile. 

I've just done a basic replacement of `telemetry.enableTelemetry`: `true`/`false` with `telemetry.telemetryLevel`: `"all"`/`"off"`. The telemetry levels can be a bit more fine-grained than that, but I didn't want to add more complexity to the tests in this PR 😄 
 
(More comprehensive testing could be a follow-up PR, but I don't think it's super important, since the decision to send telemetry is handled by VS Code libraries instead of by us.) 

## Checklist

N/A—test changes only 🧪

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/no-workspace/telemetry.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/no-workspace/telemetry.test.ts
@@ -18,7 +18,7 @@ jest.setTimeout(10000);
 
 describe("telemetry reporting", () => {
   let originalTelemetryExtension: boolean | undefined;
-  let originalTelemetryGlobal: boolean | undefined;
+  let originalTelemetryGlobal: string | undefined;
   let isCanary: string;
   let ctx: ExtensionContext;
   let telemetryListener: ExtensionTelemetryListener;
@@ -48,7 +48,7 @@ describe("telemetry reporting", () => {
     try {
       // in case a previous test has accidentally activated this extension,
       // need to disable it first.
-      // Accidentaly activation may happen asynchronously due to activationEvents
+      // Accidental activation may happen asynchronously due to activationEvents
       // specified in the package.json.
       globalTelemetryListener?.dispose();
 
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/test/vscode-tests/no-workspace/telemetry.test.ts
```diff
@@ -574,6 +573,16 @@ describe("telemetry reporting", () => {
...
   }
 
+  async function setTelemetryLevel(section: string, value: string | undefined) {
+    await workspace
+      .getConfiguration(section)
+      .update("telemetryLevel", value, ConfigurationTarget.Global);
+
+    // Need to wait some time since the onDidChangeConfiguration listeners fire
+    // asynchronously. Must ensure they to complete in order to have a successful test.
+    await wait(100);
```
charisk, line 583: An observation, but not introduced in this PR so doesn't need to be address now: We seem to call `enableTelemetry` and `setTelemetry` sequentially. They're both updating the config and then wait for 100ms. I wonder if we could merge the calls into one and only wait once. It would shave off some waiting timing.

charisk (approved):
Thanks.

extensions/ql-vscode/test/vscode-tests/no-workspace/telemetry.test.ts
shati-patel: Good point! 👀 I'll leave things as-is for now, since there'll likely be more changes to the tests anyway, if we manage to upgrade the telemetry package 🤞🏽

Issue #1459, opened 2022-08-10
The extension buffers progress messages from the CLI instead of showing them as they happen


aschackmull:
**Describe the bug**
When the extension runs CodeQL CLI commands and those commands attempt to report progress through stderr then the messages are buffered and only shown after the CLI command has finished. This means that slow CLI commands cannot report progress when run through the VSCode extension.

**Version**
CodeQL extension version: 1.6.9 
CodeQL CLI version: 2.10.2 
Platform: darwin x64

**To reproduce**
Run any query. Then you'll see the `CLI command succeeded.` message before any `[PROGRESS]` messages from that CLI command. E.g. something like this:
```
Generating JSON log summary using CodeQL CLI: generate log-summary -v --log-to-stderr --format=predicates /Users/schack/Library/Application Support/Code/User/globalStorage/github.vscode-codeql/queries/quickquery.ql-d6L9VW0HE6rwNMfIW1ZHx/evaluator-log.jsonl /Users/schack/Library/Application Support/Code/User/globalStorage/github.vscode-codeql/queries/quickquery.ql-d6L9VW0HE6rwNMfIW1ZHx/evaluator-log.summary.jsonl...
CLI command succeeded.
[2022-08-10 11:07:09] [PROGRESS] generate log-summary> Beginning to generate summary for query log located at /Users/schack/Library/Application Support/Code/User/globalStorage/github.vscode-codeql/queries/quickquery.ql-d6L9VW0HE6rwNMfIW1ZHx/evaluator-log.jsonl
[2022-08-10 11:07:09] [PROGRESS] generate log-summary> Continuing to generate log summary: have currently processed about 99% of input file
[2022-08-10 11:07:09] [PROGRESS] generate log-summary> Finished generating log summary
[2022-08-10 11:07:09] [PROGRESS] generate log-summary> Beginning to generate summary for query log located at /Users/schack/Library/Application Support/Code/User/globalStorage/github.vscode-codeql/queries/quickquery.ql-d6L9VW0HE6rwNMfIW1ZHx/evaluator-log.jsonl
[2022-08-10 11:07:09] [PROGRESS] generate log-summary> Continuing to generate log summary: have currently processed about 99% of input file
[2022-08-10 11:07:09] [PROGRESS] generate log-summary> Finished generating log summary
```

**Expected behavior**
I expect the progress messages from the CLI command to be output while it is running and not be delayed until after it has finished.

**Additional context**
[...]

PR #3369, opened 2024-02-16
Stream CLI server command stderr instead of buffering

koesie10:
This streams the CLI server command stderr instead of buffering it and only printing it when the CLI command has completed. This helps users to know that something is actually happening for long-running commands. The full stderr will still be completed in an error when the command fails.

I think an easy way to test this is running a variant analysis. While the query is compiling, output should be streamed to the extension log output.

Closes https://github.com/github/vscode-codeql/issues/1459

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -362,6 +362,8 @@ export class CodeQLCliServer implements Disposable {
     silent?: boolean,
   ): Promise<string> {
     const stderrBuffers: Buffer[] = [];
+    // The array of fragments of stderr of this line. To be used for logging.
+    let currentLineStderrBuffers: Buffer[] = [];
     if (this.commandInProcess) {
       throw new Error("runCodeQlCliInternal called while cli was running");
     }
@@ -419,6 +421,32 @@ export class CodeQLCliServer implements Disposable {
           // Listen to stderr
           process.stderr.addListener("data", (newData: Buffer) => {
             stderrBuffers.push(newData);
+
+            if (!silent) {
+              currentLineStderrBuffers.push(newData);
+
+              // Print the stderr to the logger as it comes in. We need to ensure that
+              // we don't split messages on the same line, so we buffer the stderr and
+              // split it on EOLs.
+              let currentLineBuffer = Buffer.concat(currentLineStderrBuffers);
+              const eolBuffer = Buffer.from(EOL);
+              if (currentLineBuffer.includes(eolBuffer)) {
+                while (currentLineBuffer.includes(eolBuffer)) {
+                  const line = currentLineBuffer.subarray(
+                    0,
+                    currentLineBuffer.indexOf(eolBuffer),
+                  );
+
+                  void this.logger.log(line.toString("utf-8"));
+
+                  currentLineBuffer = currentLineBuffer.subarray(
+                    currentLineBuffer.indexOf(eolBuffer) + eolBuffer.length,
+                  );
+                }
+
+                currentLineStderrBuffers = [currentLineBuffer];
+              }
+            }
           });
           // Listen for process exit.
           process.addListener("close", (code) =>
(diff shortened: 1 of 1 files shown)


koesie10:
> What's the reasoning behind using different logic for stdout and stderr? Shall we always stream?

The stdout usually doesn't contain output that users need to deal with: it contains the output that we consume from within the extension, for example a JSON object with resolved data. The stderr only contains output that is useful for users and is not parsed by us (except in case the command fails, in which case we sometimes try to extract the error message). So, we don't need to stream stdout because it isn't useful information for users (and usually also isn't streamed from the CLI side, only printed at the end of a command).

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -362,6 +362,8 @@ export class CodeQLCliServer implements Disposable {
     silent?: boolean,
   ): Promise<string> {
     const stderrBuffers: Buffer[] = [];
+    // The array of fragments of stderr of this line. To be used for logging.
+    let currentLineStderrBuffers: Buffer[] = [];
```
robertbrignull, line 366: Would it be simpler to store a single `currentLineBuffer: Buffer` instead of an array of buffers?

Currently, whenever we want to print the current line we have to concatenate all the buffers together, but this could be avoided if we stored the concatenated buffer. And I don't think it would make for extra work when adding a new data fragment. The current process is:
- add new fragment to `currentLineStderrBuffers` array
- Concatenate `currentLineStderrBuffers` array into `currentLineBuffer` and check if it contains an `EOL`
- After handling any newlines, set `currentLineStderrBuffers = [currentLineBuffer]`. Maintaining the original individual fragments isn't important and after handling any newlines that data is lost.

I think we could simplify so whenever any new fragment comes in we do `currentLineBuffer = Buffer.concat([currentLineBuffer, newData])` and then proceed as before, updating `currentLineBuffer` whenever we find a `EOL` and print a line.

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -419,6 +421,32 @@ export class CodeQLCliServer implements Disposable {
...
+
+                  void this.logger.log(line.toString("utf-8"));
+
+                  currentLineBuffer = currentLineBuffer.subarray(
+                    currentLineBuffer.indexOf(eolBuffer) + eolBuffer.length,
+                  );
+                }
+
+                currentLineStderrBuffers = [currentLineBuffer];
+              }
```
robertbrignull, line 448: We check whether the buffer contains `eolBuffer` or calculate the index quite a lot here. Once first of all, then once as the loop condition, and then twice within the loop. I recognise it's not a very expensive operation of the buffer is in-memory and likely to be short, but I wonder if it's worth avoiding anyway.

I normally find it confusing the set a variable in the loop condition, but this could be a case where it works quite well. Then we calculate it only once per loop.

If you also go with my earlier suggestion of converting the array of buffers into a single buffer, the code could look like:
```suggestion
              let eolIndex;
              while (
                (eolIndex = currentLineStderrBuffer.indexOf(eolBuffer)) !== -1
              ) {
                const line = currentLineStderrBuffer.subarray(0, eolIndex);
                void this.logger.log(line.toString("utf-8"));

                currentLineStderrBuffer = currentLineStderrBuffer.subarray(
                  eolIndex + eolBuffer.length,
                );
              }
```

robertbrignull (approved):
I've tried it out and comparing it to `main` you can definitely tell that the output is snappier and more responsive than it was before.

I have a couple of small potential improvements for the code but they're not essential if others disagree, and the code works fine as it is.

I'm happy if you want to wait for a review from CodeQL Experiences or go ahead with just reviews from us.

extensions/ql-vscode/src/codeql-cli/cli.ts
koesie10: Thanks, I think that makes sense. I did find that because we're doing `subarray` here, we also need to copy the buffer to prevent memory leaks (because a subarray keeps a reference to the original buffer), but this was also present in my original code.

robertbrignull (approved):
Thanks for updating. The new changes LGTM too.

PR #3374, opened 2024-02-19
Add sentToLLM to ModelingStore

robertbrignull:
This PR adds a new `sentToLLMMethods: Set<string>` field to `ModelingStore` along with all the associated events and plumbing.

Note: very open to suggestions for the naming of fields/methods. Naming is hard 😢 

The intention here is to very simply track which methods have been sent to the LLM in the current modeling session (i.e. it'll be reset if the model editor is closed and reopened). This state can be used along with the set of modeled methods to identify the set of positive and negative predictions from the LLM.

Methods are added to `sentToLLMMethods` whenever we successfully receive a response from the LLM. We add all methods from that batch, regardless of the modelings included in the LLM response. Currently there's no way to remove methods from `sentToLLMMethods` because we want the state to persist for the remainder of the session, but there's no reason a way of removing methods from this set couldn't be added.

UI changes for positive and negative LLM predictions will come in future PRs. Here we're just laying the groundwork.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/modeling-store.ts
+++ b/extensions/ql-vscode/src/model-editor/modeling-store.ts
@@ -14,7 +14,7 @@ interface InternalDbModelingState {
   modeledMethods: Record<string, ModeledMethod[]>;
   modifiedMethodSignatures: Set<string>;
   inProgressMethods: Set<string>;
-  sentToLLMMethods: Set<string>;
+  processedByAutoModelMethods: Set<string>;
   selectedMethod: Method | undefined;
   selectedUsage: Usage | undefined;
 }
@@ -27,7 +27,7 @@ interface DbModelingState {
   readonly modeledMethods: Readonly<Record<string, readonly ModeledMethod[]>>;
   readonly modifiedMethodSignatures: ReadonlySet<string>;
   readonly inProgressMethods: ReadonlySet<string>;
-  readonly sentToLLMMethods: ReadonlySet<string>;
+  readonly processedByAutoModelMethods: ReadonlySet<string>;
   readonly selectedMethod: Method | undefined;
   readonly selectedUsage: Usage | undefined;
 }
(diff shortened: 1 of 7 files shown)


extensions/ql-vscode/src/view/model-editor/ModelEditor.tsx
```diff
@@ -148,6 +148,10 @@ export function ModelEditor({
             setInProgressMethods(new Set(msg.methods));
             break;
           }
+          case "setSentToLLMMethods": {
+            // TODO: set state
```
robertbrignull, line 152: We'll need to add this as React state and pass it down through the UI elements to where it's needed, but this can be done in a future PR.

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -336,6 +342,15 @@ export class ModelingStore extends DisposableObject {
     });
   }
 
+  public addSentToLLMMethods(dbItem: DatabaseItem, sentToLLMMethods: string[]) {
+    this.changeSentToLLMMethods(dbItem, (state) => {
```
robertbrignull, line 346: I decided to follow the existing pattern of having a `changeX` method, even though it's only called from here and we don't have a `removeSentToLLMMethods` method yet. Hopefully that keeps the code more consistent and easier to understand, but let me know if you disagree.

koesie10 (commented):
Changes LGTM, but it seems like there's a test that expects a new method to be mocked.

For naming: Perhaps `processedByAutoModel` would be better? That hides the implementation detail of the auto modeler using an LLM. An alternative might be `modeledByAutoModel`, but that might be unclear (since "modeled" could also mean that there's at least 1 model for it).

charisk:
+1 for `processedByAutoModel` or `processedByLlm` (we've been using the term LLM elsewhere, and automodel could include non-LLM model generation too)

robertbrignull:
I'll change to `processedByAutoModel`. I like that a lot better.

I originally had it called "processed" because I think that word is clearer, but it's also longer so I changed to "sent" for being shorter. But since you both suggested it I'm very happy to switch back to "processed".

I wasn't sure whether to use the term "LLM" but indeed we do use it in a few other places. I think "AutoModel" is a better term to use here and I agree it's good that it could include non-LLM options too.

koesie10 (approved):
Thanks, LGTM

PR #3376, opened 2024-02-20
Don't send methods to AutoModel more than once

robertbrignull:
Once a method has been processed by AutoModel once, don't try to process it again. Pretty simple now we have `processedByAutoModelMethods` available.

The only problem I've noticed with this is that it can make the UI confusing when there's nothing to model. You'll press to "model with AI" button and it'll appear to do nothing. I'm going to open a separate PR to follow this one to try to address that.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/model-editor-view.ts
+++ b/extensions/ql-vscode/src/model-editor/model-editor-view.ts
@@ -656,7 +656,10 @@ export class ModelEditorView extends AbstractWebview<
       methodSignatures,
     );
     const processedByAutoModelMethods =
-      this.modelingStore.getProcessedByAutoModelMethods(this.databaseItem);
+      this.modelingStore.getProcessedByAutoModelMethods(
+        this.databaseItem,
+        methodSignatures,
+      );
     const mode = this.modelingStore.getMode(this.databaseItem);
     await this.autoModeler.startModeling(
       packageName,
--- a/extensions/ql-vscode/src/model-editor/model-editor-view.ts
+++ b/extensions/ql-vscode/src/model-editor/model-editor-view.ts
@@ -655,11 +655,14 @@ export class ModelEditorView extends AbstractWebview<
       this.databaseItem,
       methodSignatures,
     );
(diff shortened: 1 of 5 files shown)


koesie10 (approved):
LGTM

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -344,6 +344,20 @@ export class ModelingStore extends DisposableObject {
...
+    dbItem: DatabaseItem,
+    methodSignatures?: string[],
+  ): Set<string> {
+    const processedByAutoModelMethods =
+      this.getState(dbItem).processedByAutoModelMethods;
+    if (!methodSignatures) {
+      return processedByAutoModelMethods;
+    }
+    return new Set(
+      Array.from(processedByAutoModelMethods).filter(methodSignatures.includes),
```
robertbrignull, line 357: Unfortunately using `filter(methodSignatures.includes)` is causing an error. I thought it would be ok since it's a local variable 😢 

Converting to an anonymous function gets everything working.

extensions/ql-vscode/src/model-editor/model-editor-view.ts
```diff
@@ -655,11 +655,14 @@ export class ModelEditorView extends AbstractWebview<
       this.databaseItem,
       methodSignatures,
     );
+    const processedByAutoModelMethods =
+      this.modelingStore.getProcessedByAutoModelMethods(this.databaseItem);
```
robertbrignull, line 659: Oops, I implemented the `methodSignatures` argument for this method but then forgot to use it here.

I've added that in now and done another manual test that everything is working ok.

extensions/ql-vscode/src/model-editor/modeling-store.ts
koesie10: It might be calling `Array.includes` with the wrong `this` parameter, so `methodSignatures.includes.bind(this)` will probably work. I do think an anonymous function is clearer here.

PR #3384, opened 2024-02-21
Moved logic to get standard pack for variant analysis

charisk:
Moved the logic that downloads the standard Code Scanning pack and prepares for a variant analysis, into a new module so that it can be re-used for model evaluation.

I've removed the tests in `variant-analysis-manager.test.ts` because it no longer seemed worth keeping it around considering the main bit that was being tested was moved to `code-scanning-pack.test.ts`.

I'm not too happy with the name of the new module/function so please shout if you have any suggestions.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
+++ b/extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
@@ -1,0 +1,77 @@
+import { join } from "path";
+import type { App } from "../common/app";
+import type { QueryLanguage } from "../common/query-language";
+import type { CodeQLCliServer } from "../codeql-cli/cli";
+import type { QlPackDetails } from "./ql-pack-details";
+import { getQlPackFilePath } from "../common/ql";
+
+export async function getCodeScanningPack(
+  app: App,
+  cliServer: CodeQLCliServer,
+  language: QueryLanguage,
+): Promise<QlPackDetails> {
+  // Get pack
+  void app.logger.log(`Downloading pack for language: ${language}`);
+  const packName = `codeql/${language}-queries`;
+  const packDownloadResult = await cliServer.packDownload([packName]);
+  const downloadedPack = packDownloadResult.packs[0];
+
+  const packDir = join(
+    packDownloadResult.packDir,
+    downloadedPack.name,
+    downloadedPack.version,
+  );
+
+  // Resolve queries
+  void app.logger.log(`Resolving queries for pack: ${packName}`);
+  const suitePath = join(
+    packDir,
+    "codeql-suites",
+    `${language}-code-scanning.qls`,
+  );
+  const resolvedQueries = await cliServer.resolveQueries(suitePath);
+
+  const problemQueries = await filterToOnlyProblemQueries(
+    app,
+    cliServer,
+    resolvedQueries,
+  );
+
+  if (problemQueries.length === 0) {
+    throw Error(
+      `No problem queries found in published query pack: ${packName}.`,
+    );
+  }
+
+  // Return pack details
+  const qlPackFilePath = await getQlPackFilePath(packDir);
+
+  const qlPackDetails: QlPackDetails = {
+    queryFiles: problemQueries,
+    qlPackRootPath: packDir,
+    qlPackFilePath,
+    language,
+  };
+
+  return qlPackDetails;
+}
+
+async function filterToOnlyProblemQueries(
+  app: App,
+  cliServer: CodeQLCliServer,
+  queries: string[],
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/code-scanning-pack.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    const pack = await getCodeScanningPack(app, cli, QueryLanguage.Javascript);
+    // Should include queries. Just check that at least one known query exists.
+    // It doesn't particularly matter which query we check for.
+    expect(
+      pack.queryFiles.find((q) => q.includes("PostMessageStar.ql")),
+    ).toBeDefined();
+    // Should not include non-problem queries.
+    expect(
+      pack.queryFiles.find((q) => q.includes("LinesOfCode.ql")),
+    ).not.toBeDefined();
```
koesie10, line 28: I think this is slightly clearer:

```suggestion
    expect(
      pack.queryFiles.some((q) => q.includes("PostMessageStar.ql")),
    ).toBeTruthy();
    // Should not include non-problem queries.
    expect(
      pack.queryFiles.some((q) => q.includes("LinesOfCode.ql")),
    ).toBeFalsy();
```

extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
```diff
@@ -0,0 +1,77 @@
+import { join } from "path";
+import type { App } from "../common/app";
+import type { QueryLanguage } from "../common/query-language";
+import type { CodeQLCliServer } from "../codeql-cli/cli";
+import type { QlPackDetails } from "./ql-pack-details";
+import { getQlPackFilePath } from "../common/ql";
+
+export async function getCodeScanningPack(
```
koesie10, line 8: Perhaps something like `resolveCodeScanningQueryPack` would be a slightly better name?

extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
```diff
@@ -0,0 +1,77 @@
...
+  app: App,
```
koesie10, line 9: Since only the logger is used in here, could this instead be:

```suggestion
  logger: BaseLogger,
```

extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
```diff
@@ -0,0 +1,77 @@
...
+      queryMetadata.kind === "problem" ||
+      queryMetadata.kind === "path-problem"
+    ) {
+      problemQueries.push(query);
+    } else {
+      void app.logger.log(`Skipping non-problem query ${query}`);
+    }
+  }
+  return problemQueries;
+}
```
koesie10, line 77: Probably not relevant for this refactoring PR, but I think this can be done a lot faster by changing how we resolve queries. Instead of calling `resolve queries` with the standard code scanning suite, I think we can create a custom file that applies some extra conditions to the code scanning suite. For example, a `.qls` file like this would probably result in the same result:

```yaml
- import: codeql-suites/javascript-code-scanning.qls
  from: codeql/javascript-queries
- exclude:
    kind:
    - diagnostic
- exclude:
    kind:
    - metric
```

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -229,54 +230,12 @@ export class VariantAnalysisManager
...
-      }
-
-      const qlPackFilePath = await getQlPackFilePath(packDir);
-
       // Build up details to pass to the functions that run the variant analysis.
-      const qlPackDetails: QlPackDetails = {
-        queryFiles: problemQueries,
-        qlPackRootPath: packDir,
-        qlPackFilePath,
+      const qlPackDetails = await getCodeScanningPack(
```
koesie10, line 234: Shouldn't we still call `progress` before executing this? Something like:

```suggestion
      progress({
        maxStep: 8,
        step: 1,
        message: "Downloading and resolving queries",
      });

      const qlPackDetails = await getCodeScanningPack(
```

koesie10 (approved):

extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
charisk: Thanks for the suggestion! I've created an issue to cover this.

koesie10 (approved):
Thanks!

PR #3385, opened 2024-02-21
Add model evaluation run to store and view

charisk:
This PR introduces the `ModelEvalutionRun` domain entity and wires up model evaluation run information to the modeling store and the model editor view. Some notes:
- The `ModelEvaluationRun` interface is pretty minimal at the moment. I expect it to grow while we build more features (e.g. it might need to contain model packs).
- The model evaluation run information is not yet persisted - it's only in memory (in the `ModelingStore`).
- A new `ModelEvaluator` class has been added to contain logic around model evaluation in order to keep changes in `model-editor-view` minimal. The implementation is pretty minimal for now and will be fleshed out in the near future.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/model-editor/model-evaluation-run.ts
+++ b/extensions/ql-vscode/src/model-editor/model-evaluation-run.ts
@@ -1,0 +1,20 @@
+export interface ModelEvaluationRun {
+  status: ModelEvaluationStatus;
+  variantAnalysisId: number | undefined;
+}
+
+type ModelEvaluationStatus =
+  | "preparing"
+  | "inProgress"
+  | "succeeded"
+  | "failed"
+  | "canceled";
+
+export function evaluationRunIsRunning(
+  evaluationRun: ModelEvaluationRun,
+): boolean {
+  return (
+    evaluationRun.status === "preparing" ||
+    evaluationRun.status === "inProgress"
+  );
+}
--- a/extensions/ql-vscode/src/model-editor/model-evaluation-run.ts
+++ b/extensions/ql-vscode/src/model-editor/model-evaluation-run.ts
@@ -1,20 +1,4 @@
 export interface ModelEvaluationRun {
-  status: ModelEvaluationStatus;
+  isPreparing: boolean;
   variantAnalysisId: number | undefined;
-}
-
-type ModelEvaluationStatus =
-  | "preparing"
-  | "inProgress"
-  | "succeeded"
-  | "failed"
-  | "canceled";
-
-export function evaluationRunIsRunning(
-  evaluationRun: ModelEvaluationRun,
-): boolean {
-  return (
-    evaluationRun.status === "preparing" ||
-    evaluationRun.status === "inProgress"
-  );
 }
--- a/extensions/ql-vscode/src/model-editor/model-evaluator.ts
+++ b/extensions/ql-vscode/src/model-editor/model-evaluator.ts
@@ -1,0 +1,46 @@
+import type { ModelingStore } from "./modeling-store";
+import type { DatabaseItem } from "../databases/local-databases";
+import type { ModelEvaluationRun } from "./model-evaluation-run";
+import { DisposableObject } from "../common/disposable-object";
+import { sleep } from "../common/time";
+
+export class ModelEvaluator extends DisposableObject {
+  public constructor(
+    private readonly modelingStore: ModelingStore,
+    private readonly dbItem: DatabaseItem,
+  ) {
+    super();
+  }
+
+  public async startEvaluation() {
+    // Update store with evaluation run status
+    const evaluationRun: ModelEvaluationRun = {
+      status: "preparing",
+      variantAnalysisId: undefined,
+    };
+    this.modelingStore.updateModelEvaluationRun(this.dbItem, evaluationRun);
+
+    // For now, just wait 5 seconds and then update the store with the succeeded status.
+    // In the future, this will be replaced with the actual evaluation process.
+    void sleep(5000).then(() => {
+      const completedEvaluationRun: ModelEvaluationRun = {
+        status: "succeeded",
+        variantAnalysisId: undefined,
+      };
+      this.modelingStore.updateModelEvaluationRun(
+        this.dbItem,
+        completedEvaluationRun,
+      );
+    });
+  }
+
+  public async stopEvaluation() {
+    // For now just update the store with the canceled status.
+    // This will be fleshed out in the near future.
+    const evaluationRun: ModelEvaluationRun = {
+      status: "canceled",
+      variantAnalysisId: undefined,
+    };
+    this.modelingStore.updateModelEvaluationRun(this.dbItem, evaluationRun);
+  }
+}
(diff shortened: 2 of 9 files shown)


extensions/ql-vscode/src/model-editor/model-evaluation-run.ts
```diff
@@ -0,0 +1,20 @@
+export interface ModelEvaluationRun {
+  status: ModelEvaluationStatus;
+  variantAnalysisId: number | undefined;
+}
+
+type ModelEvaluationStatus =
```
robertbrignull, line 6: I'm wondering are we duplicating too much data from the variant analysis status?

Whenever the variant analysis run changes status from in-progress to some other state, we have to see that event and then update the state here. I worry it could lead to bugs if the two states get out of sync.

An alternative could be essentially `ModelEvaluationStatus = "preparing" | "delegateToVariantAnalysisStatus"` (but with better names).

Is this something you considered? Apologies I haven't been following the earlier PRs or issues closely so I likely missed any discussion there.

charisk: This is a valid point. I did consider it, but because `VariantAnalysisStatus` is an enum it's quite awkward - you can't just extend it. And since the variant analysis statuses don't really change (since they come from the back-end), it felt safe enough to re-define them. If we want to go down that route, I would suggest we change to use strings rather than enums.

The way the type is created right now also comes with more flexibility about what states we want to expose for model evaluation runs.

robertbrignull: What you've said makes perfect sense, but it wasn't quite what I was suggesting. Apologies I didn't make myself clear enough.

It being an enum does indeed make it hard to extend, and I don't have any objection to redefining the type for passing around transient values internally. My point was rather should we be storing that type anywhere for any length of time (even in memory, let alone to disk).

What I was trying to describe wasn't to "extend" the `VariantAnalysisStatus` type, but rather that one of the values in `ModelEvaluationStatus` means "I don't know. Go check the variant analysis status instead". By doing this it means we don't have to worry about keeping the value up to date whenever the variant analysis status changes, and instead we go calculate it when needed.

Therefore the logic for getting the model evaluation status would look like:
```
getModelEvalationStatus(): InternalModelEvaluationStatus {
  if (modelEvaluationRun.status === "preparing") {
    // It's preparing still and the variant analysis hasn't been triggered yet
    return "preparing";
  } else {
    // Ask the variant analysis manger for the status and process / return that
    const variantAnalysis = variantAnalysisManager.getVariantAnalysis(modelEvaluationRun.variantAnalysisId);
    if (variantAnalysis.status === "in_progress") {
      return "in_progress";
    } else if (variantAnalysis.status === "canceled") {
      return "canceled";
    } else if (...) {
      etc...
    }
  }
}
```

charisk: Oh I see! That's interesting - it's not something I considered. Do you mind if we discuss this separately (we have an issue to track work around domain entities)? Seems like something that would be good to discuss with @shati-patel when she's back, and ideally I wouldn't like it to block this PR.

extensions/ql-vscode/src/model-editor/model-evaluator.ts
```diff
@@ -0,0 +1,73 @@
...
+    this.push(
+      this.modelingEvents.onModelEvaluationRunChanged(async (event) => {
+        if (
+          event.evaluationRun &&
+          event.dbUri === this.dbItem.databaseUri.toString()
+        ) {
+          const run: ModelEvaluationRunState = {
+            isPreparing: event.evaluationRun.isPreparing,
+
+            // TODO: Get variant analysis from id.
```
charisk, line 65: I will do this in the next PR. It needs the `VariantAnalysisManager` wired in and it would make the PR bigger than it is so I left it out for now.

extensions/ql-vscode/src/model-editor/shared/model-evaluation-run-state.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function modelEvaluationRunIsRunning(
+  run: ModelEvaluationRunState,
+): boolean {
+  return (
+    run.isPreparing ||
+    !!(
+      run.variantAnalysis &&
+      run.variantAnalysis.status !== VariantAnalysisStatus.InProgress
```
shati-patel, line 16: Is this correct? I'd expect that we want 

```suggestion
      run.variantAnalysis.status === VariantAnalysisStatus.InProgress
```

extensions/ql-vscode/src/model-editor/model-evaluator.ts
```diff
@@ -0,0 +1,73 @@
...
+  public async startEvaluation() {
+    // Update store with evaluation run status
+    const evaluationRun: ModelEvaluationRun = {
+      isPreparing: true,
+      variantAnalysisId: undefined,
+    };
+    this.modelingStore.updateModelEvaluationRun(this.dbItem, evaluationRun);
+
+    // For now, just wait 5 seconds and then update the store with the succeeded status.
+    // In the future, this will be replaced with the actual evaluation process.
```
shati-patel, line 32: We can probably remove the comment about "succeeded" status, since we now just use the `isPreparing` boolean

extensions/ql-vscode/src/model-editor/model-evaluator.ts
```diff
@@ -0,0 +1,73 @@
...
+      this.modelingStore.updateModelEvaluationRun(
+        this.dbItem,
+        completedEvaluationRun,
+      );
+    });
+  }
+
+  public async stopEvaluation() {
+    // For now just update the store with the canceled status.
+    // This will be fleshed out in the near future.
```
shati-patel, line 47: As above, we don't really have a "canceled" status 🙅🏽

extensions/ql-vscode/src/view/model-editor/ModelEditor.tsx
```diff
@@ -87,20 +89,21 @@ const ModelEvaluation = ({
...
   onStopEvaluation: () => void;
-  evaluationInProgress: boolean;
+  evaluationRun: ModelEvaluationRunState | undefined;
 }) => {
   if (!viewState.showEvaluationUi) {
     return null;
   }
 
-  if (!evaluationInProgress) {
+  console.log("***** check evaluationRun", evaluationRun);
```
shati-patel, line 105: I assume these `console.log`s shouldn't be merged in 😄

extensions/ql-vscode/src/view/model-editor/ModelEditor.tsx
```diff
@@ -214,6 +219,10 @@ export function ModelEditor({
           case "setAccessPathSuggestions":
             setAccessPathSuggestions(msg.accessPathSuggestions);
             break;
+          case "setModelEvaluationRun":
+            console.log("**** setModelEvaluationRun", msg.run);
```
shati-patel, line 223: another `console.log` 📝

shati-patel (commented):
Looks good ✨ Just a few minor comments!

extensions/ql-vscode/src/view/model-editor/ModelEditor.tsx
charisk: Busted!

shati-patel (approved):
Good stuff 💪🏽 😀

PR #3395, opened 2024-02-22
Set modeled methods and modified methods together

robertbrignull:
This is a small-ish PR that aims to fix the flickering when modeling a method by setting both the modeled methods and modified method signatures at the same time.

We just augment the existing `SetModeledMethodsMessage` and `ModeledMethodsChangedEvent` types so they can also carry information about the modified method signatures. We could change the names of these types, but I haven't done that in this PR yet.

I considered making the new field optional, and in fact that's what I did in my initial attempt at implementing this, but I can't see any big downsides to just always sending it. It makes the types simpler and less likely that you'll miss an update by forgetting to set or consume that field.

This turned out to be fairly simple to do, and other than the potential naming of these types I can't see any big problems with it.

## Checklist


extensions/ql-vscode/src/model-editor/modeling-events.ts
```diff
@@ -25,6 +25,7 @@ interface ModeChangedEvent {
 
 interface ModeledMethodsChangedEvent {
   readonly modeledMethods: Readonly<Record<string, ModeledMethod[]>>;
+  readonly modifiedMethodSignatures: ReadonlySet<string>;
```
koesie10, line 28: If we add this to this event, we should remove `ModifiedMethodsChangedEvent`. It now seems like when modified methods are added, we fire this event, while if they are removed we fire `ModifiedMethodsChangedEvent`. This is really confusing and means you need to listen to two different events for receiving the state of 1 set.

We should then also find some other usages of the `onModifiedMethodsChanged` event which need to be updated because not all changes to this state are now sent out via that event.

koesie10 (commented):
I think it makes sense to merge these 2 messages to fix this problem, but we should merge the 2 events as well to ensure we don't need to listen to two different events for receiving the state of a single set. I think we now have a bug where the method modeling panel does not properly send the `setMethodModified` message when the method is modified.

robertbrignull:
@koesie10, thanks for finding the bug. I've now merged the two events together, so there isn't a dedicated "set modified methods" event anymore. Let me know if you this is what you meant.

As far as I can tell this has fixed the bug with the method modeling panel, but please also take a look at this and see if you can still reproduce it.

extensions/ql-vscode/test/__mocks__/model-editor/modelingEventsMock.ts
```diff
@@ -7,8 +7,8 @@ export function createMockModelingEvents({
   onMethodsChanged = jest.fn(),
   onHideModeledMethodsChanged = jest.fn(),
   onModeChanged = jest.fn(),
-  onModeledMethodsChanged = jest.fn(),
-  onModifiedMethodsChanged = jest.fn(),
+  onModeledAndModifiedMethodsChanged:
+    onModeledAndModifiedMethodsChanged = jest.fn(),
```
koesie10, line 11:
```suggestion
  onModeledAndModifiedMethodsChanged = jest.fn(),
```

koesie10 (approved):
Thanks, LGTM!

PR #3396, opened 2024-02-22
Merge model editor webview messages and modeling store events

robertbrignull:
This PR follows on from https://github.com/github/vscode-codeql/pull/3395 and tries to show what things would look like if we just merged everything.

This isn't necessarily what I'm suggesting we do, but I just wanted to see what it would look like. The answer is it reduces code a lot and I don't think it makes things any harder to work with or reason about.

I'd be interested to hear thoughts.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/modeling-events.ts
+++ b/extensions/ql-vscode/src/model-editor/modeling-events.ts
@@ -6,11 +6,15 @@ import type { Method, Usage } from "./method";
 import type { ModeledMethod } from "./modeled-method";
 import type { Mode } from "./shared/mode";
 
-interface MethodsChangedEvent {
-  readonly methods: readonly Method[];
+interface ModelingStateChangedEvent {
   readonly dbUri: string;
   readonly databaseItem: DatabaseItem;
   readonly isActiveDb: boolean;
+  readonly methods?: readonly Method[];
+  readonly modeledMethods?: Readonly<Record<string, ModeledMethod[]>>;
+  readonly modifiedMethodSignatures?: ReadonlySet<string>;
+  readonly inProgressMethodSignatures?: ReadonlySet<string>;
+  readonly processedByAutoModelMethodSignatures?: ReadonlySet<string>;
 }
 
 interface HideModeledMethodsChangedEvent {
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/model-editor/modeling-events.ts
```diff
@@ -6,11 +6,15 @@ import type { Method, Usage } from "./method";
...
-  readonly methods: readonly Method[];
+interface ModelingStateChangedEvent {
   readonly dbUri: string;
   readonly databaseItem: DatabaseItem;
   readonly isActiveDb: boolean;
+  readonly methods?: readonly Method[];
+  readonly modeledMethods?: Readonly<Record<string, ModeledMethod[]>>;
+  readonly modifiedMethodSignatures?: ReadonlySet<string>;
+  readonly inProgressMethodSignatures?: ReadonlySet<string>;
+  readonly processedByAutoModelMethodSignatures?: ReadonlySet<string>;
```
koesie10, line 17: I think it's quite confusing that all fields are optional here, which means that the main change is that you only need to subscribe to 1 event rather than 5 different events. However, you still need to make all kinds of checks and would theoretically need to handle all combinations of fields that can be set (since none are mutually exclusive). This can still result in sending multiple webview messages when states are modified sequentially.

koesie10 (commented):
I think this makes it harder to reason about the event that gets sent out and makes it much harder to handle the event (you now need to handle 1 to 5 fields of the event being set).

extensions/ql-vscode/src/model-editor/modeling-events.ts
charisk: What if: 
- the fields weren't optional and we always set the state
- the state is combined into a single object rather than the 5 different pieces we have

Would that cause too many unnecessary re-renders? 🤔 Are the state pieces generally related and end up changing together often? I think this is true for `modeledMethods` and `methodSignatures` at least, but unsure about the rest. 

I think `useMemo` relies on object equality so we can't just rely on that to avoid re-renders - we'd need to make sure we don't change objects, which is generally a pain IMO.

robertbrignull:
I'm going to close this PR since I think it's unlikely we'll pursue this particular approach. I'm happy to discuss this area more, but perhaps we should do it on an issue or on slack rather than here.

PR #3398, opened 2024-02-22
Move all sorting to the modeling store, and then all views preserve the sorting

robertbrignull:
This PR aims to make sorting of methods easier and prevent them from changing sort order and flying around when you make manual changes. Sorting of methods is moved to the modeling store so it happens in one place only, and then it is the job of each view to preserve the ordering of methods it is given.

This gives us a lot more control over what actions trigger a change in method sorting. As of this PR the only actions that should cause methods to change order are:
- the initial loading of methods
- saving modeled methods
- new AI predictions

It also helps reduce some of the code bloat caused by https://github.com/github/vscode-codeql/pull/3382 because we no longer need to pass as much information to all the different views.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
+++ b/extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
@@ -73,9 +73,7 @@ export class MethodsUsageDataProvider
       this.modifiedMethodSignatures !== modifiedMethodSignatures
     ) {
       this.methods = methods;
-      this.sortedTreeItems = createTreeItems(
-        sortMethodsInGroups(methods, mode),
-      );
+      this.sortedTreeItems = createTreeItems(createGroups(methods, mode));
       this.databaseItem = databaseItem;
       this.sourceLocationPrefix =
         await this.databaseItem.getSourceLocationPrefix(this.cliServer);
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
```diff
@@ -253,27 +246,9 @@ function urlValueResolvablesAreEqual(
...
 }
 
-function sortMethodsInGroups(
-  methods: readonly Method[],
-  modeledMethods: Readonly<Record<string, readonly ModeledMethod[]>>,
-  mode: Mode,
-  modifiedMethodSignatures: ReadonlySet<string>,
-  processedByAutoModelMethods: ReadonlySet<string>,
-): Method[] {
+function sortMethodsInGroups(methods: readonly Method[], mode: Mode): Method[] {
```
koesie10, line 249: Perhaps name this something like `createGroups` or `groupMethods` since we're not actually sorting the methods?

koesie10 (approved):
LGTM, but there seem to be a few failing tests dependent on the sorting

extensions/ql-vscode/src/model-editor/methods-usage/methods-usage-data-provider.ts
robertbrignull: We can't really call it `groupMethods` because that's also the name of the method in `sorting.ts`. I've renamed it to `createGroups` and we can always tweak again in the future if we want to.

PR #3400, opened 2024-02-23
Extract saving of model extension YAML file

koesie10:
This refactors the saving of YAML files for model extension files. Instead of directly constructing a string from the modeled methods, we will now create the structure of the YAML as a JS object. Then, we consume that object to create a YAML string. This would also allow us to e.g. dump the YAML directly, but the primary objective is to re-use this new function in a follow-up PR.

The YAML functions itself are pretty well-tested in [`yaml.test.ts`](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/test/unit-tests/model-editor/yaml.test.ts), so I'm pretty confident that this results in the same output in all instances.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/model-extension-file.schema.json
+++ b/extensions/ql-vscode/src/model-editor/model-extension-file.schema.json
@@ -8,35 +8,38 @@
         "extensions": {
           "type": "array",
           "items": {
-            "type": "object",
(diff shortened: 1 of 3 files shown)


charisk (approved):
Nice! I've been wanting us to do this since we first added the model editor 😬

PR #3401, opened 2024-02-23
Generate separate file for generated type models in Ruby by default

koesie10:
This will generate a separate file `models/ruby.model.generated.yml` when running the model editor for Ruby and type models are hidden (the default). This file is ignored by the model editor, but stores all type models generated by the query. This is part of hiding type models by default.

Screenshot 2024-02-23 at 15 55 38

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/generate.ts
+++ b/extensions/ql-vscode/src/model-editor/generate.ts
@@ -5,19 +5,15 @@ import type { QueryRunner } from "../query-server";
 import type { CodeQLCliServer } from "../codeql-cli/cli";
 import type { ProgressCallback } from "../common/vscode/progress";
 import { getOnDiskWorkspaceFolders } from "../common/vscode/workspace-folders";
-import type { ModeledMethod } from "./modeled-method";
 import { runQuery } from "../local-queries/run-query";
 import type { QueryConstraints } from "../local-queries";
 import { resolveQueries } from "../local-queries";
 import type { DecodedBqrs } from "../common/bqrs-cli-types";
+
 type GenerateQueriesOptions = {
   queryConstraints: QueryConstraints;
   filterQueries?: (queryPath: string) => boolean;
-  parseResults: (
-    queryPath: string,
-    results: DecodedBqrs,
-  ) => ModeledMethod[] | Promise<ModeledMethod[]>;
-  onResults: (results: ModeledMethod[]) => void | Promise<void>;
+  onResults: (queryPath: string, results: DecodedBqrs) => void | Promise<void>;
 
   cliServer: CodeQLCliServer;
   queryRunner: QueryRunner;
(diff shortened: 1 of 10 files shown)


koesie10:
I'll change this to use the `codeQL.model.showTypeModels` setting instead of canary mode once https://github.com/github/vscode-codeql/pull/3399 is merged.

extensions/ql-vscode/src/model-editor/model-editor-view.ts
```diff
@@ -731,7 +748,25 @@ export class ModelEditorView extends AbstractWebview<
...
+          return;
         }
+
+        progress({
+          step: 4000,
+          maxStep: 4000,
+          message: "Saving generated models",
+        });
+
+        const fileContents = `# This file was automatically generated based from ${this.databaseItem.name}. Manual changes will not persist.\n\n${modelExtensionFileToYaml(extensionFile)}`;
```
shati-patel, line 760: Minor suggestion:

```suggestion
        const fileContents = `# This file was automatically generated from ${this.databaseItem.name}. Manual changes will not persist.\n\n${modelExtensionFileToYaml(extensionFile)}`;
```

Or maybe change `from` to `on` or something like that instead?

extensions/ql-vscode/src/model-editor/generate.ts
```diff
@@ -5,19 +5,15 @@ import type { QueryRunner } from "../query-server";
...
 import { resolveQueries } from "../local-queries";
 import type { DecodedBqrs } from "../common/bqrs-cli-types";
+
 type GenerateQueriesOptions = {
   queryConstraints: QueryConstraints;
   filterQueries?: (queryPath: string) => boolean;
-  parseResults: (
-    queryPath: string,
-    results: DecodedBqrs,
-  ) => ModeledMethod[] | Promise<ModeledMethod[]>;
```
shati-patel, line 19: Why are we getting rid of the `parseResults` option? 🤔

shati-patel (commented):
Looks mostly good, just a few questions! 📝 

----------- 

I also had a general comment from testing: When I run `Generate` from the model editor, there's a log entry saying
```
CLI command succeeded.
No predicate found for typeVariableModel
``` 

I don't think it's necessarily related to your changes, but I also couldn't see exactly where it's coming from 😅 (This is on the nightly CLI version, and a recent `main` codeql submodule checkout—but I tried a few different versions too 🤔)

extensions/ql-vscode/src/model-editor/generate.ts
koesie10: `parseResults` was very specific to producing `ModeledMethod[]`, while that's not something we want to do for this type model generation (we can just put the rows as-is into a YAML file). The only thing that `parseResults` was used for was `await onResults(await parseResults(queryPath, bqrs))`, which can very easily be replaced by `await onResults(queryPath, bqrs)` with very minor changes on the consumer side.

koesie10:
> I don't think this "typeVariableModel" is necessarily related to your changes, but I also couldn't see exactly where it's coming from 😅 (This is on the nightly CLI version, and a recent `main` codeql submodule checkout—but I tried a few different versions too 🤔)

I believe this has been there since the beginning. The model editor doesn't support `typeVariableModel`, but it's a predicate that's produced from the generation query. We can ignore this log message.

shati-patel (approved):
Thanks for the explanations ✨ Looks good!

PR #3405, opened 2024-02-26
Add support for 'canceling' status for variant analysis

charisk:
Introduces a new 'canceling' status for variant analyses which is used to show whether we're in the middle of stoping the analysis.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
@@ -34,6 +34,7 @@ import {
   isVariantAnalysisComplete,
   parseVariantAnalysisQueryLanguage,
   VariantAnalysisScannedRepositoryDownloadStatus,
+  VariantAnalysisStatus,
 } from "./shared/variant-analysis";
 import { getErrorMessage } from "../common/helpers-pure";
 import { VariantAnalysisView } from "./variant-analysis-view";
@@ -611,8 +612,20 @@ export class VariantAnalysisManager
       return;
     }
 
-    if (!this.variantAnalyses.has(variantAnalysis.id)) {
+    const originalVariantAnalysis = this.variantAnalyses.get(
+      variantAnalysis.id,
+    );
+
+    if (!originalVariantAnalysis) {
       return;
+    }
+
+    // Maintain the canceling status if we are still canceling.
+    if (
+      originalVariantAnalysis.status === VariantAnalysisStatus.Canceling &&
+      variantAnalysis.status === VariantAnalysisStatus.InProgress
+    ) {
+      variantAnalysis.status = VariantAnalysisStatus.Canceling;
     }
 
     await this.setVariantAnalysis(variantAnalysis);
@@ -838,6 +851,11 @@ export class VariantAnalysisManager
       throw new Error(`No variant analysis with id: ${variantAnalysisId}`);
     }
 
+    await this.onVariantAnalysisUpdated({
+      ...variantAnalysis,
+      status: VariantAnalysisStatus.Canceling,
+    });
+
     if (!variantAnalysis.actionsWorkflowRunId) {
       throw new Error(
         `No workflow run id for variant analysis with id: ${variantAnalysis.id}`,
@@ -848,7 +866,17 @@ export class VariantAnalysisManager
       this.app.logger,
       "Cancelling variant analysis. This may take a while.",
     );
-    await cancelVariantAnalysis(this.app.credentials, variantAnalysis);
+    try {
+      await cancelVariantAnalysis(this.app.credentials, variantAnalysis);
+    } catch (e) {
+      if (variantAnalysis.status === VariantAnalysisStatus.Canceling) {
+        await this.onVariantAnalysisUpdated({
+          ...variantAnalysis,
+          status: VariantAnalysisStatus.InProgress,
+        });
+      }
+      throw e;
+    }
   }
 
   public async openVariantAnalysisLogs(variantAnalysisId: number) {
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
@@ -146,6 +146,7 @@ export class VariantAnalysisManager
       new VariantAnalysisMonitor(
         app,
         this.shouldCancelMonitorVariantAnalysis.bind(this),
+        this.getVariantAnalysisStatus.bind(this),
       ),
     );
     this.variantAnalysisMonitor.onVariantAnalysisChange(
@@ -605,6 +606,19 @@ export class VariantAnalysisManager
     return !this.variantAnalyses.has(variantAnalysisId);
   }
 
+  private getVariantAnalysisStatus(
+    variantAnalysisId: number,
+  ): VariantAnalysisStatus {
+    const variantAnalysis = this.variantAnalyses.get(variantAnalysisId);
+    if (!variantAnalysis) {
+      throw new Error(
+        `No variant analysis found with id: ${variantAnalysisId}.`,
+      );
+    }
+
+    return variantAnalysis.status;
+  }
+
   public async onVariantAnalysisUpdated(
     variantAnalysis: VariantAnalysis | undefined,
   ): Promise<void> {
@@ -618,14 +632,6 @@ export class VariantAnalysisManager
 
     if (!originalVariantAnalysis) {
       return;
-    }
-
-    // Maintain the canceling status if we are still canceling.
-    if (
-      originalVariantAnalysis.status === VariantAnalysisStatus.Canceling &&
-      variantAnalysis.status === VariantAnalysisStatus.InProgress
-    ) {
-      variantAnalysis.status = VariantAnalysisStatus.Canceling;
     }
 
     await this.setVariantAnalysis(variantAnalysis);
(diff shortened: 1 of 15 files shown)


extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -848,7 +866,17 @@ export class VariantAnalysisManager
...
     );
-    await cancelVariantAnalysis(this.app.credentials, variantAnalysis);
+    try {
+      await cancelVariantAnalysis(this.app.credentials, variantAnalysis);
+    } catch (e) {
+      if (variantAnalysis.status === VariantAnalysisStatus.Canceling) {
+        await this.onVariantAnalysisUpdated({
+          ...variantAnalysis,
+          status: VariantAnalysisStatus.InProgress,
+        });
```
shati-patel, line 876: Why are we setting the status as `InProgress` here? I expected `Canceling`, but I might be getting confused with the logic 🙈

shati-patel (commented):
I had one question inline, but looks good otherwise! 🖼️ 

Is it worth adding a storybook example for the "canceling" status? 📖

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
charisk: We're in a code branch where we failed to cancel the variant analysis (for whatever reason), so it seemed that the right thing to do is to put it back to in progress so the user can try and cancel again.

shati-patel: Ah I see, thanks!

shati-patel (approved):
🚢

shati-patel:
Oh wait, a CHANGELOG entry would indeed be helpful!

charisk:
> Oh wait, a CHANGELOG entry would indeed be helpful!

Cool, I'll update. BTW I won't merge until we hear back from Product.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -848,7 +866,17 @@ export class VariantAnalysisManager
       this.app.logger,
       "Cancelling variant analysis. This may take a while.",
     );
-    await cancelVariantAnalysis(this.app.credentials, variantAnalysis);
+    try {
+      await cancelVariantAnalysis(this.app.credentials, variantAnalysis);
+    } catch (e) {
+      if (variantAnalysis.status === VariantAnalysisStatus.Canceling) {
+        await this.onVariantAnalysisUpdated({
```
robertbrignull, line 873: Shall we add a log message here saying we were unable to cancel the variant analysis?

extensions/ql-vscode/src/view/variant-analysis/__tests__/VariantAnalysisActions.spec.tsx
```diff
@@ -45,6 +45,24 @@ describe(VariantAnalysisActions.name, () => {
     expect(onStopQueryClick).toHaveBeenCalledTimes(1);
   });
 
+  it("renders the stopping query disabled button when canceling", async () => {
```
robertbrignull, line 48: Do we need more tests?

For example:
- the status is set to cancel when the button is clicked
- the status is set back to in-progress if the request to cancel fails
- the monitoring job retains the canceling status

robertbrignull (commented):
I like the approach, but I've seen a few things in the code that don't look quite right to me. Although I haven't tried manually running the code to verify things. Have you manually tested this?

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
charisk: Isn't throwing the exception enough here?

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -611,10 +612,22 @@ export class VariantAnalysisManager
...
-    if (!this.variantAnalyses.has(variantAnalysis.id)) {
+    const originalVariantAnalysis = this.variantAnalyses.get(
+      variantAnalysis.id,
+    );
+
+    if (!originalVariantAnalysis) {
       return;
     }
 
+    // Maintain the canceling status if we are still canceling.
```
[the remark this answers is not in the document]
charisk: I think you're right. I did a last minute refactor here and didn't think through all the cases I had in mind on Friday. 

I'm also having some trouble with testing manually atm - I seem to be testing against older versions of the code somehow. I'm either being forgetful with builds or the watcher/build is a bit off. 

I'll move some logic to `mapUpdatedVariantAnalysis`.

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
robertbrignull: Maybe that's good enough. If the user sees an uncaught exception hopefully that clues them into the idea that something went wrong with canceling the analysis.

charisk:
@shati-patel @robertbrignull this is now ready to review again. While testing I noticed a small bug around the "stats" area so I fixed that (see https://github.com/github/vscode-codeql/pull/3405/commits/f0ea6409927f7e338e5abc397be6bdc90a8f1bd7).

I've also done the following manual tests:
- Run a normal variant analysis
- Run a variant analysis and press "Stop query"
- Change the gh-actions-api-client.ts to always throw an error so I can simulate error when canceling, run a variant analysis and press "Stop query".

extensions/ql-vscode/src/variant-analysis/variant-analysis-mapper.ts
```diff
@@ -49,6 +50,7 @@ export function mapUpdatedVariantAnalysis(
     VariantAnalysis,
     "language" | "query" | "queries" | "databases" | "executionStartTime"
   >,
+  currentStatus: VariantAnalysisStatus | undefined,
```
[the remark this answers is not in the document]
robertbrignull: Maybe we should rewrite the exposed interface of this file. Right now we have
```
mapVariantAnalysis(VariantAnalysisSubmission, ApiVariantAnalysis): VariantAnalysis
mapUpdatedVariantAnalysis(Pick<...>, ApiVariantAnalysis): VariantAnalysis
```
where `mapVariantAnalysis` depends on `mapUpdatedVariantAnalysis`. We could rewrite `mapUpdatedVariantAnalysis` to be
```
mapUpdatedVariantAnalysis(VariantAnalysis, ApiVariantAnalysis): VariantAnalysis
```
and then have a third un-exported method that can have `Pick` or other ugly things for its args, but it'll be an internal function and not exposed outside the file.

extensions/ql-vscode/src/variant-analysis/variant-analysis-monitor.ts
```diff
@@ -121,6 +125,7 @@ export class VariantAnalysisMonitor extends DisposableObject {
 
       variantAnalysis = mapUpdatedVariantAnalysis(
         variantAnalysis,
+        this.getVariantAnalysisStatus(variantAnalysis.id),
```
robertbrignull, line 128: Could you explain why we call `getVariantAnalysisStatus(variantAnalysis.id)` to fetch the status from the variant analysis manager instead of using the `variantAnalysis.status` value we already have?

extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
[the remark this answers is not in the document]
robertbrignull: Oops, I missed originally that we pass `onVariantAnalysisUpdated` into `variantAnalysisMonitor.onVariantAnalysisChange`, so it was called in other places. But the point still stands about when we want to set that status back when cancelation fails, so the other changes you made are still the right thing.

extensions/ql-vscode/src/variant-analysis/variant-analysis-mapper.ts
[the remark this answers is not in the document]
charisk: Yeah that's the sort of thing I had in mind (although I'm still not into using Pick and would rather avoid it). Let's play around that in a separate PR.

extensions/ql-vscode/src/variant-analysis/variant-analysis-monitor.ts
charisk: Because the `variantAnalysis.status` we already have is one we got from the API so it won't necessarily have the "true" status (which is "canceling"). I should probably at least add a code comment to explain this.

robertbrignull: Hmm, I agree this is necessary for now, but I think it's not ideal. I'd like to look at changing it in a followup tech-debt issue.

In the code here, `variantAnalysisSummary` comes from the API, but the `variantAnalysis` variable we have here is of the internal type and has gone through mapping / processing. The problem is that it's a variable we keep within this `_monitorVariantAnalysis` function so it doesn't see any changes that come from outside the monitor. Previously this was fine because the monitor was the only thing that changed the variant analysis state and then it just fed data back to the manager. But now changes can come from other places and the monitor doesn't see these changes and tries to overwrite them.

Adding in `getVariantAnalysisStatus` works right now, but only because the `status` field is the only bit that's changing. It could cause problems if any other fields change in other places.

We could make it more general and fetch the whole variant analysis from the manager. We could also rewrite the monitor so it can track any changes by listening to the `onVariantAnalysisStatusUpdated` event from the manager.

charisk: Yeah agree it's an area that could do with improvement. IMO it goes hand in hand with https://github.com/github/vscode-codeql/pull/3405#discussion_r1503890621.

charisk: I've added a comment. I would like to make some incremental improvements around this and the mapping functions in separate PRs, if that's okay.

shati-patel (approved):
Poked it again, and things still work for me! Thanks for fixing the bug in the "Stats" header too ⚡ 

I'll leave @robertbrignull to re-review + approve, since he had more comments than I did 😛

extensions/ql-vscode/src/variant-analysis/variant-analysis-monitor.ts
```diff
@@ -119,8 +123,13 @@ export class VariantAnalysisMonitor extends DisposableObject {
         continue;
       }
 
+      // Get the current status of the variant analysis as known by the rest
+      // of the app, because it may have changed by the user, but the API
```
robertbrignull, line 127: Very optional suggestion which I think makes the comment slightly more correct, but I leave the decision up to you if you want to accept this or something else or do nothing:
```suggestion
      // of the app, because it may have been changed by the user and this code
```

PR #3414, opened 2024-02-28
Add check for id property when running variant analysis

koesie10:
This adds a check for variant analyses to ensure that the `@id` metadata is set when running a `problem` or `path-problem` query. This will ensure that this error is caught early.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
@@ -327,6 +327,17 @@ export class VariantAnalysisManager
       );
     }
 
+    // It's not possible to interpret a BQRS file to SARIF without an id property.
+    if (
+      queryMetadata?.kind &&
+      ["problem", "path-problem"].includes(queryMetadata.kind) &&
+      !queryMetadata.id
+    ) {
+      throw new UserCancellationException(
+        `${firstQueryFile} does not have the required @id property for a ${queryMetadata.kind} query.`,
+      );
+    }
+
     const {
       actionBranch,
       base64Pack,
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
@@ -309,21 +309,6 @@ export class VariantAnalysisManager
       message: "Getting credentials",
     });
 
-    const {
-      actionBranch,
-      base64Pack,
-      repoSelection,
-      controllerRepo,
-      queryStartTime,
-    } = await prepareRemoteQueryRun(
-      this.cliServer,
-      this.app.credentials,
-      qlPackDetails,
-      progress,
-      token,
-      this.dbManager,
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/variant-analysis/variant-analysis-manager.ts
```diff
@@ -342,6 +327,32 @@ export class VariantAnalysisManager
       );
     }
 
+    // It's not possible to interpret a BQRS file to SARIF without an id property.
+    if (
+      queryMetadata?.kind &&
+      ["problem", "path-problem"].includes(queryMetadata.kind) &&
```
robertbrignull, line 333: Should we also include `alert` and `path-alert` in this list?

koesie10: Yes, good point. I've changed this to use a function instead and also used this function in other places where I could find `path-problem`.

extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
```diff
@@ -64,10 +65,7 @@ async function filterToOnlyProblemQueries(
   const problemQueries: string[] = [];
   for (const query of queries) {
     const queryMetadata = await cliServer.resolveMetadata(query);
-    if (
-      queryMetadata.kind === "problem" ||
-      queryMetadata.kind === "path-problem"
-    ) {
+    if (isSarifResultsQueryKind(queryMetadata.kind)) {
```
robertbrignull, line 68: This shouldn't change behaviour since all queries in the published packs should be using the new query kinds. Also it should be safe anyway and more correct than it was before.

extensions/ql-vscode/src/view/variant-analysis/RepositoriesSearchSortRow.tsx
```diff
@@ -47,10 +48,7 @@ const RepositoriesResultFormatColumn = styled(RepositoriesResultFormat)`
 function showResultFormatColumn(
   variantAnalysisQueryKind: string | undefined,
 ): boolean {
-  return (
-    variantAnalysisQueryKind === "problem" ||
-    variantAnalysisQueryKind === "path-problem"
-  );
+  return isSarifResultsQueryKind(variantAnalysisQueryKind);
```
robertbrignull, line 51: It's possible that this could cause a tiny UI inconsistency until https://github.com/github/codeql-variant-analysis-action/blob/4a8a27800b99bca0d1900c940648b70243804e2f/src/codeql.ts#L351 is also updated. This is because we decide to show the column based on the query kind instead of whether there are SARIF results available, because we can't check every repo and we need to show the column before any repos have been analysed.

However, I think it's fine to merge this as it's highly unlikely to affect anybody and it's fine so long as we also update the variant analysis action soon.

robertbrignull (approved):
LGTM

This can technically change behaviour in a couple of places, but it's very minor and I think it's fine and it's only making us more correct and consistent overall.

PR #3415, opened 2024-02-28
Update variant analysis monitor to get the whole variant analysis object

charisk:
Currently the variant analysis monitor gets the up to date status of the variant analysis from the manager. In order to avoid potentially operating on outdated information, we change that to get the whole variant analysis object instead.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-monitor.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-monitor.ts
@@ -5,7 +5,6 @@ import { RequestError } from "@octokit/request-error";
 import type {
   VariantAnalysis,
   VariantAnalysisScannedRepository,
-  VariantAnalysisStatus,
 } from "./shared/variant-analysis";
 import {
   isFinalVariantAnalysisStatus,
@@ -18,6 +17,7 @@ import { sleep } from "../common/time";
 import { getErrorMessage } from "../common/helpers-pure";
 import type { App } from "../common/app";
 import { showAndLogWarningMessage } from "../common/logging";
+import type { QueryLanguage } from "../common/query-language";
 
 export class VariantAnalysisMonitor extends DisposableObject {
   // With a sleep of 5 seconds, the maximum number of attempts takes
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/variant-analysis/variant-analysis-monitor.ts
```diff
@@ -60,20 +60,28 @@ export class VariantAnalysisMonitor extends DisposableObject {
...
   }
 
   private async _monitorVariantAnalysis(
-    variantAnalysis: VariantAnalysis,
+    variantAnalysisId: number,
+    controllerRepoId: number,
+    executionStartTime: number,
+    queryName: string,
+    language: QueryLanguage,
   ): Promise<void> {
```
shati-patel, line 81: Maybe a silly question: why has this changed to take all these bits of information instead of just `variantAnalysis`?

shati-patel (approved):
Looks good, though I just had a little question 🔍

extensions/ql-vscode/src/variant-analysis/variant-analysis-monitor.ts
charisk: I wanted to avoid passing a variant analysis object (and calling it `variantAnalysis`) as I think it'd make the code confusing further down (when we get the "actual" variant analysis from the manager).

So it was an intentional change for readability rather than a need. This is subjective ofc, so we if we don't think it makes things more readable then we should pass a `variantAnalysis` (or maybe call it something else?).

I think variant analyses generally have some information that doesn't change (the contoller repo id, the query info, etc.) so we could think about introducing a new. type to cover that. I think it could be used in other places too (e.g. the mapping functions). But anyway, that's a different rabbit hole 😂

shati-patel: Ah okay, that makes sense! I've definitely been getting confused about all the `variantAnalysis` vars floating around (and when they're being updated) 😵 
So I'm happy with this just accepting the "unchanging" properties!

PR #3421, opened 2024-02-29
Add logic to stop a model evaluation run

charisk:
Updated the `ModelEvaluator` logic to deal with stopping of an evaluation run. As part of this I extracted the logic that creates and cancels a variant analysis into a private `runVariantAnalysis` function.

The code isn't super straight forward so I've added a bunch of comments.

I've also started adding some tests for the `ModelEvaluator` but only focused on the stopping aspect for now. 

I've done the following manual tests:
- Stop evaluation as soon as it starts
- Stop evaluation during query pack building (the query pack building continues and you can see output in the extension log but the notification goes away)
- Stop evaluation during variant analysis 
- Stop variant analysis from the variant analysis results view page
- Start/stop a few times in the same session

## Checklist


shati-patel (approved):
Thanks! In terms of sensible manual testing, things work generally as expected for me too. 

(I haven't been able to test "Stop variant analysis from the variant analysis results view page" since we don't auto-open the view any more)

-----------

In terms of ✨ chaos ✨ testing: If I hammer the Evaluate/Stop Evaluation button enough, I can get into a place where everything lags, and then multiple evaluation runs happen and things get a bit confusing. It seems like we kick off multiple MRVA runs, but can only cancel the latest one. It's hard to see what's actually going on since we don't display the results anywhere, so I'm not too worried for now! 🙃

PR #3433, opened 2024-03-01
Import testproj databases into workspace storage

aeisenberg:
Commit-by-commit review is recommended.

If a folder that ends with `.testproj` is encountered, assume it is a
database created by a codeql test. When the user wants to import this
database, copy it into workspace storage. The database can be
re-imported, which first removes the old version before importing it
again.

Ask user if they want to re-import outdated testproj dbs

Before running a query now, do the following:

1. Check if the selected database is imported from a testproj
2. If so, check the last modified time of the `codeql-datase.yml` file
   of the imported database with that of its origin.
3. If the origin database has a file that is newer, assume that the
   database has been recreated since the last time it was imported.
4. If newer, then ask the user if they want to re-import before running
   the query.

Also, this change appends the `(test)` label to all test databases in
the database list.

To verify this change, do the following:

1. In the development workspace directory, create a codeql test:
    ```
    echo "import javascript
    select 1" | tee  test.ql
    echo "dependencies:
        codeql/javascript-all: \"*\"
    extractor: javascript
    tests: test"
    echo hucairz | tee qlpack.yml
    ```
2. Start the vscode dev workbench
3. In the vscode editor, go to the test tab (looks like a beaker)
4. Run the test. It should fail
5. Right click on the `test.testproj` directory and select "(Re-)Import test database"
6. Notice that the database is imported
7. Run the query if you like and you should see `1` as the result.
9. Change test to `select 2`
10. Re-run the test
11. Re-run the query
12. This time, you should get a popup asking if you want to re-import
13. Select `yes`
14. See the results are now `2`.

## Checklist


extensions/ql-vscode/package.json
```diff
@@ -2068,7 +2081,8 @@
     "ts-loader": "^9.4.2",
     "ts-node": "^10.7.0",
     "ts-unused-exports": "^10.0.0",
-    "typescript": "^5.0.2"
+    "typescript": "^5.0.2",
+    "unzipper": "^0.10.14"
```
koesie10, line 2085: I haven't looked at the full PR yet, but is there a reason we need to be using `unzipper` here? We do already have [`yauzl`](https://github.com/thejoshwolfe/yauzl) as a dependency and some helpers functions in [`unzip.ts`](https://github.com/github/vscode-codeql/blob/96b7722f260b44c75b8f8da309fe0b7e8fa30b85/extensions/ql-vscode/src/common/unzip.ts) and [`unzip-concurrently.ts`](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/src/common/unzip-concurrently.ts).

aeisenberg: Thanks. I wasn't aware of those dependencies. I can switch.

aeisenberg:
I'm not sure what's happening with the `EPERM` errors when renaming the directory on windows. It must be that the OS is hanging on to an open file handle somewhere, somehow, but I don't know where this would be happening. I _could_ try to add a 1 second wait before renaming in case there's some sort of cleanup that needs to finish, but that feels very unsatisfactory.

adityasharad:
Two possibilities come to mind:
- If you've just created the copy, then Defender might be scanning the newly-created files.
- Does an existing CodeQL process have a lock on the DB?

aeisenberg:
The next thing to try would be to add a few second wait on windows.

aeisenberg:
OK...that seems to have worked.

extensions/ql-vscode/src/databases/database-fetcher.ts
```diff
@@ -294,16 +296,18 @@ export async function importArchiveDatabase(
   cli: CodeQLCliServer,
 ): Promise<DatabaseItem | undefined> {
   try {
-    const item = await databaseArchiveFetcher(
+    const origin: DatabaseOrigin = {
+      type: databaseUrl.endsWith(".testproj") ? "testproj" : "archive",
+      // TODO validate that archive origins can use a file path, not a URI
```
koesie10, line 301: I don't think we currently use the origin for archives anywhere, so it should be fine to change this to an `fsPath`. However, if we do start using this in the future, we do need to handle both URIs and file paths for archives.

extensions/ql-vscode/src/databases/database-fetcher.ts
```diff
@@ -366,7 +370,11 @@ async function databaseArchiveFetcher(
   const unzipPath = await getStorageFolder(storagePath, databaseUrl);
 
   if (isFile(databaseUrl)) {
-    await readAndUnzip(databaseUrl, unzipPath, cli, progress);
+    if (origin.type === "testproj") {
```
koesie10, line 373: Would it make sense to change this condition to checking whether the `databaseUrl` is a directory on the filesystem? That seems like it should be more generic.

extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -744,6 +746,54 @@ export class DatabaseUI extends DisposableObject {
     );
   }
 
+  private async handleImportTestDatabase(uri: Uri): Promise<void> {
+    return withProgress(
+      async (progress) => {
+        try {
+          // Assume user has selected an archive if the file has a .zip extension
```
koesie10, line 753: This comment seems like it is outdated, I think this comment can be removed:

```suggestion
```

extensions/ql-vscode/src/databases/local-databases/database-manager.ts
```diff
@@ -159,6 +163,93 @@ export class DatabaseManager extends DisposableObject {
...
+
+    // Compare timestmps of the codeql-database.yml files of the original and the
+    // imported databases.
+    const originDbYml = join(dbItem.origin.path, "codeql-database.yml");
+    const importedDbYml = join(
+      dbItem.databaseUri.fsPath,
+      "codeql-database.yml",
+    );
+
+    // TODO add error handling if one does not exist.
```
koesie10, line 224: Should this still be added?

extensions/ql-vscode/src/query-server/query-runner.ts
```diff
@@ -63,9 +63,22 @@ export interface CoreQueryRun {
 export type CoreCompletedQuery = CoreQueryResults &
   Omit<CoreQueryRun, "evaluate">;
 
+type OnQueryRunStargingListener = (dbPath: Uri) => Promise<void>;
```
koesie10, line 66:
```suggestion
type OnQueryRunStartingListener = (dbPath: Uri) => Promise<void>;
```

extensions/ql-vscode/test/vscode-tests/cli-integration/utils.ts
```diff
@@ -7,3 +7,7 @@ import { join } from "path";
 export function getDataFolderFilePath(path: string): string {
   return join(workspace.workspaceFolders![0].uri.fsPath, path);
 }
+
+export async function wait(ms = 1000) {
```
koesie10, line 11: Could you use the `sleep` function from `src/common/time.ts`?

koesie10 (commented):
Everything seems to be working well for me, I just have some minor comments.

extensions/ql-vscode/src/databases/database-fetcher.ts
aeisenberg: If in the future we do want to use URIs, I'd suggest adding a new field `uri` to hold this value. We've had bugs in the past where we assigned URI strings to `path` properties.

extensions/ql-vscode/src/databases/database-fetcher.ts
aeisenberg: OK. I'll change it to make it more consistent with `readAndUnzip` and `fetchAndUnzip`.

extensions/ql-vscode/src/databases/database-fetcher.ts
```diff
@@ -294,16 +296,17 @@ export async function importArchiveDatabase(
...
       storagePath,
       undefined,
-      {
-        type: "archive",
-        path: databaseUrl,
-      },
+      origin,
       progress,
       cli,
     );
```
koesie10, line 312: I can't leave a comment on the exact line, but there's a message "Database unzipped and imported successfully.", which is somewhat confusing if you're importing a testproj database since nothing is being unzipped. Could this message be changed when you're using a testproj database?

koesie10 (approved):
LGTM. However, there are some confusing logs when you re-import a database before running a query. I see the following log lines:

```
Deleting database from filesystem.
Deleted '/Users/koesie10/Library/Application Support/Code/User/workspaceStorage/c3466bf33a6faffab7f413267a6ffe3f/GitHub.vscode-codeql/modeleditor-1'
```

However, there's nothing to indicate that the database has been imported again, so it'd be good to add that as well.

aeisenberg:
Thanks for such a detailed review, Koen!

PR #3437, opened 2024-03-05
Avoid using a @treemap query in testplan

robertbrignull:
Replaces the query used in the test plan from `@kind treemap` to `@kind table`. The intention is to test a generic non-problem query and not specifically `treemap` which is a kind that is rarely used and we don't necessarily have full support for. It's more important to test a `table` query since those are more likely to be run by users.

## Checklist


shati-patel (approved):
Makes sense 👍🏽

PR #3438, opened 2024-03-05
Fix `isQuickEvalCount` logic

RasmusWL:
The UI would often show `Quick evaluation counts` even for normal (non counts) quick evaluations.

Since `false !== undefined` evaluates to true, we would assign the
wrong value to `isQuickEvalCount` when `selectedQuery.quickEval?.quickEvalCount` is false

## Checklist

**No user facing changes**

- [ x Issues have been created for any UI or other user-facing changes made by this pull request.

### DIFF
--- a/extensions/ql-vscode/src/run-queries-shared.ts
+++ b/extensions/ql-vscode/src/run-queries-shared.ts
@@ -495,8 +495,7 @@ export async function createInitialQueryInfo(
   outputDir: QueryOutputDir,
 ): Promise<InitialQueryInfo> {
   const isQuickEval = selectedQuery.quickEval !== undefined;
-  const isQuickEvalCount =
-    selectedQuery.quickEval?.quickEvalCount !== undefined;
+  const isQuickEvalCount = Boolean(selectedQuery.quickEval?.quickEvalCount);
   return {
     queryPath: selectedQuery.queryPath,
     isQuickEval,


extensions/ql-vscode/src/run-queries-shared.ts
```diff
@@ -495,8 +495,7 @@ export async function createInitialQueryInfo(
   outputDir: QueryOutputDir,
 ): Promise<InitialQueryInfo> {
   const isQuickEval = selectedQuery.quickEval !== undefined;
-  const isQuickEvalCount =
-    selectedQuery.quickEval?.quickEvalCount !== undefined;
+  const isQuickEvalCount = Boolean(selectedQuery.quickEval?.quickEvalCount);
```
robertbrignull, line 498: Alternatively we could simplify further and just pass through `selectedQuery.quickEval?.quickEvalCount` without any manipulation.

The `isQuickEvalCount` field of `InitialQueryInfo` already allows undefined and has a comment saying "Missing is false for backwards compatibility", and therefore all the code consuming this value already has to handle it being `undefined`.

i.e.
```suggestion
  const isQuickEvalCount = selectedQuery.quickEval?.quickEvalCount;
```
or remove this variable entirely and later on do
```typescript
isQuickEvalCount: selectedQuery.quickEval?.quickEvalCount,
```

robertbrignull (approved):
I agree this looks like the right fix, and if this works for you then it seems fine. I have one suggestion for how to make it even simpler, but what you've written is also perfectly fine.

extensions/ql-vscode/src/run-queries-shared.ts
RasmusWL: ah yes, that seems like a fine solution :+1:

robertbrignull (approved):

PR #3439, opened 2024-03-05
Refactor events to fix the method modeling panel getting stuck showing "start modeling"

robertbrignull:
This PR fixes the method modeling panel getting stuck showing "start modeling" if you open the model editor before you open the QL sidebar panel.

The problem was that when the webview loaded we weren't setting the `language` field correctly in the webview. There needs to be a call to `setViewState()` after determining the language.

I recommend reviewing commit by commit.

The actual fix is made in https://github.com/github/vscode-codeql/commit/d4413424c19121b2bd5b81539d8619f9480cc814 which is the third commit. The two commits before that are refactoring methods so we can reuse code. For this fix I've not worried about being inefficient with sending webview messages. For example since `setDatabaseItem` calls `setViewState` we don't have to call it before in this case, but I felt the complexity of having branches and relying on methods always calling other methods was not worth the minimal cost of calling `setViewState` twice when loading initial data.

The commits after the fix could be moved to a separate PR if preferable. I noticed another potential bug which was that we had both the `setMethod` and the `setSelectedMethod` webview messages which both set the method in the view but `setMethod` doesn't contain all the other fields. This could mean if we used `setMethod` we could end up with state like `isModified` or `modeledMethods` being outdated and incorrect. Luckily we avoided the bug because every time we did call `setMethod` (except when setting the method to `undefined`) we also called `setSelectedMethod`, so essentially `setMethod` is unnecessary. Indeed we can remove it and replace it with `setSelectedMethod` and a new `setNoMethodSelected` message and also simplify class interfaces by changing where we listen to modeling store events from.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/interface-types.ts
+++ b/extensions/ql-vscode/src/common/interface-types.ts
@@ -693,11 +693,6 @@ interface SetMethodModelingPanelViewStateMessage {
   viewState: MethodModelingPanelViewState;
 }
 
-interface SetMethodMessage {
-  t: "setMethod";
-  method: Method | undefined;
-}
-
 interface SetMethodModifiedMessage {
   t: "setMethodModified";
   isModified: boolean;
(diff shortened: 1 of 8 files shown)


norascheuch (approved):
The code changes look good! I haven't tested it locally though.

robertbrignull:
I've been playing around with it trying to test it manually and I haven't found any bugs or anything else wrong.

PR #3440, opened 2024-03-05
Push decision to use credentials or not down to where the creds are used

robertbrignull:
This fixes a potential bug I spotted where we might unexpectedly prompt a user for credentials to download a CodeQL database.

The code for downloading databases from GitHub is in `promptImportGithubDatabase` and `downloadGitHubDatabase` and both of those took an optional `Credentials` parameter. It is then up to the caller to decide whether to use credentials or not. In this PR we push the decision further down so we only decide in `downloadGitHubDatabase` and therefore use the same logic for all cases.

The current logic around using credentials is:
- In `handleChooseDatabaseGithub` in `local-databases-ui.ts` (called when manually importing a database by command palette or UI), we were using credentials only when the canary flag was set.
- In `createSkeletonQuery` in `local-queries.ts`, we were using credentials only when the canary flag was set.
- In `promptImportDatabase` in `model-editor-view.ts` (called when modeling a dependency or generating modeled methods), we were always using credentials.

This seems like a bug to me as the rest of the model editor doesn't require canary mode, and in all other instances of downloading a database it doesn't prompt for credentials when not in canary mode.

## Checklist


koesie10 (approved):
LGTM. I agree that we shouldn't be using the credentials in the model editor in non-canary mode, especially since this would only result in a credentials prompt when using flow generation.

robertbrignull:
I realised that this is a user-visible change so it should probably have a changelog entry. I hope that's ok. If you disagree we can editor or remove it again.

PR #3441, opened 2024-03-05
Add "model alerts indicator" component

shati-patel:
Adds "alert indicators" in the model editor to show how many results a certain model has generated: 

image

- We only show these indicators when there's an evaluation run available
- Eventually these indicators will link to a separate model alerts view with more details
- We don't currently have any way of knowing which models are responsible for which alerts (see "alert provenance" work), so we just show a random number for now 🙃 

## Checklist

N/A, this is feature-flagged for internal use/testing 🎏 

### DIFF
--- a/extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
@@ -49,7 +49,7 @@ const ApiOrMethodRow = styled.div`
   gap: 0.5em;
 `;
 
-const ButtonsContainer = styled.div`
+const ModelButtonsContainer = styled.div`
   min-height: calc(var(--input-height) * 1px);
   display: flex;
   flex-direction: row;
@@ -361,7 +361,7 @@ const ModelableMethodRow = forwardRef<HTMLElement | undefined, MethodRowProps>(
                     />
                   </DataGridCell>
                   <DataGridCell>
-                    <ButtonsContainer>
+                    <ModelButtonsContainer>
                       <ModelAlertsIndicator
                         viewState={viewState}
                         modeledMethod={modeledMethod}
@@ -391,7 +391,7 @@ const ModelableMethodRow = forwardRef<HTMLElement | undefined, MethodRowProps>(
                           <Codicon name="trash" />
                         </CodiconRow>
                       )}
-                    </ButtonsContainer>
+                    </ModelButtonsContainer>
                   </DataGridCell>
                 </DataGridRow>
               );
--- a/extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
+++ b/extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
@@ -38,6 +38,8 @@ import type { AccessPathOption } from "../../model-editor/suggestions";
 import { ModelInputSuggestBox } from "./ModelInputSuggestBox";
 import { ModelOutputSuggestBox } from "./ModelOutputSuggestBox";
 import { getModelsAsDataLanguage } from "../../model-editor/languages";
+import { ModelAlertsIndicator } from "./ModelAlertsIndicator";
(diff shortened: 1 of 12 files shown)


extensions/ql-vscode/src/view/model-editor/ModelAlertsIndicator.tsx
```diff
@@ -0,0 +1,55 @@
...
+  align-items: center;
+`;
+
+const ModelAlertsButton = styled.button`
+  color: var(--vscode-editor-foreground);
+  background-color: var(--vscode-input-background);
+  border: none;
+  border-radius: 40%;
+  cursor: pointer;
+`;
```
shati-patel, line 19: I copied this styling from the [`UsagesButton`](https://github.com/github/vscode-codeql/blob/1eba41affb9ebb4aaafe48109d72ccc37aac11d1/extensions/ql-vscode/src/view/model-editor/MethodRow.tsx#L52-L58), though I was wondering if we could use the [`VSCodeBadge`](https://github.com/microsoft/vscode-webview-ui-toolkit/blob/main/src/badge/README.md) in both places? 🤔

robertbrignull:
As an alternative suggestion for the columns, we go with having the alert count and the add/remove modeling icon in one column using a flex-box row. I think it should all work ok and then when the alert count badge isn't there we won't get an extra gap. I my mind, merging those two columns should be ok because they're both constant-width items and both set to `max-content` in the `gridTemplateColumns`.

I think you originally had a go at this and perhaps it didn't work or got a bit confusing, but I'm happy to help out if you want to try it again.

shati-patel:
> I think you originally had a go at this and perhaps it didn't work or got a bit confusing, but I'm happy to help out if you want to try it again.

Thanks! Will have another go (and potentially pester you for some more assistance 😄)

extensions/ql-vscode/src/view/model-editor/ModelAlertsIndicator.tsx
robertbrignull: If `VSCodeBadge` looks and works the same I think switching both places to use it makes sense. Probably a second PR if that's ok.

I think I wrote the other example and I probably just didn't know the `VSCodeBadge` component existed or forgot to look for it. I don't remember any reason I didn't use it.

robertbrignull (approved):
LGTM but I did pair on writing it a bit. Maybe best to wait for @charisk to have a look too next week.

extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
```diff
@@ -47,6 +49,14 @@ const ApiOrMethodRow = styled.div`
   gap: 0.5em;
 `;
 
+const ButtonsContainer = styled.div`
```
charisk, line 52: Should this be a bit more specific e.g. `ModelButtonsContainer`? We have other buttons (e.g. Usages button)

extensions/ql-vscode/src/view/model-editor/ModelAlertsIndicator.tsx
charisk: Are we planning to switch to a `VSCodeBadge` in a separate PR?

charisk (approved):
Looking good! Just some non-blocking coments/questions.

extensions/ql-vscode/src/view/model-editor/MethodRow.tsx
shati-patel: Good point, done!

extensions/ql-vscode/src/view/model-editor/ModelAlertsIndicator.tsx
shati-patel: Yep, I'll do that in a follow-up PR 📛

PR #3444, opened 2024-03-07
Add tests of MethodModelingViewProvider.viewLoaded

robertbrignull:
This PR is trying to add tests that would have caught the bug fixed in https://github.com/github/vscode-codeql/pull/3439

To do this we need to initialize a `MethodModelingViewProvider` and send it a "viewLoaded" message. Then we can test that it does the right thing in various `ModelingStore` states. I believe the only thing the test can verify is what messages are sent to the webview.

I've focussed on what it does when the view is first loaded and haven't tried to add tests for all the other behaviours of `MethodModelingViewProvider`. We could add more tests later if we think that would be useful.

## Checklist

### DIFF
--- a/extensions/ql-vscode/test/vscode-tests/no-workspace/model-editor/method-modeling/method-modeling-view-provider.test.ts
+++ b/extensions/ql-vscode/test/vscode-tests/no-workspace/model-editor/method-modeling/method-modeling-view-provider.test.ts
@@ -1,0 +1,212 @@
+import type {
+  Uri,
+  Webview,
+  WebviewView,
+  WebviewViewResolveContext,
+} from "vscode";
+import { CancellationTokenSource, EventEmitter } from "vscode";
+import type { ModelConfigListener } from "../../../../../src/config";
+import { MethodModelingViewProvider } from "../../../../../src/model-editor/method-modeling/method-modeling-view-provider";
+import { createMockApp } from "../../../../__mocks__/appMock";
+import { createMockModelingEvents } from "../../../../__mocks__/model-editor/modelingEventsMock";
+import { createMockModelingStore } from "../../../../__mocks__/model-editor/modelingStoreMock";
+import { mockedObject } from "../../../../mocked-object";
+import type { FromMethodModelingMessage } from "../../../../../src/common/interface-types";
+import { DisposableObject } from "../../../../../src/common/disposable-object";
+import type { ModelingEvents } from "../../../../../src/model-editor/modeling-events";
+import type {
+  DbModelingState,
+  SelectedMethodDetails,
+} from "../../../../../src/model-editor/modeling-store";
+import { mockDatabaseItem } from "../../../utils/mocking.helpers";
+import {
+  createMethod,
+  createUsage,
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/test/vscode-tests/no-workspace/model-editor/method-modeling/method-modeling-view-provider.test.ts
```diff
@@ -0,0 +1,212 @@
...
+  createUsage,
+} from "../../../../factories/model-editor/method-factories";
+
+describe("method modeling view provider", () => {
+  // Modeling store
+  let getStateForActiveDb: jest.Mock<DbModelingState | undefined, []>;
+  let getSelectedMethodDetails: jest.Mock<
+    SelectedMethodDetails | undefined,
+    []
+  >;
```
koesie10, line 33: Would it make sense to use the function definitions here rather than manually specifying the arguments and return types?

```suggestion
  let getStateForActiveDb: jest.MockedFunction<
    ModelingStore["getStateForActiveDb"]
  >;
  let getSelectedMethodDetails: jest.MockedFunction<
    ModelingStore["getSelectedMethodDetails"]
  >;
```

extensions/ql-vscode/test/vscode-tests/no-workspace/model-editor/method-modeling/method-modeling-view-provider.test.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    const webviewView = mockedObject<WebviewView>({
+      webview,
+      onDidDispose: jest.fn(),
+    });
+
+    viewProvider.resolveWebviewView(
+      webviewView,
+      mockedObject<WebviewViewResolveContext>({}),
+      new CancellationTokenSource().token,
```
koesie10, line 94: It seems like it should be safe to remove these parameters from the `resolveWebviewView` method implementation so you don't need to pass these.

koesie10 (approved):
LGTM

extensions/ql-vscode/test/vscode-tests/no-workspace/model-editor/method-modeling/method-modeling-view-provider.test.ts
robertbrignull: `resolveWebviewView` is implementing a method from the `WebviewViewProvider` provider so that's why it had those unused arguments. They're there for documentation purposes, though personally I agree I think it's fine to not define them on our definition if we're not using them.

Previously it was just a matter of documentation preference with no other concerns, but now that we're calling this method ourselves in the tests it's making our testing harder as we have to mock more objects. So I think that's enough justification to remove them now.

extensions/ql-vscode/test/vscode-tests/no-workspace/model-editor/method-modeling/method-modeling-view-provider.test.ts
robertbrignull: Nice. I wasn't aware of `jest.MockedFunction` but that is a lot cleaner.

PR #3454, opened 2024-03-08
Make the "Run variant analysis against published pack" command cancellable

koesie10:
This makes the "Run variant analysis against published pack" command cancellable by starting a separate CLI process for downloading packs and bundling the query pack since these are both long-running commands.

Unfortunately, I wasn't able to fix an issue where the tests would timeout on ubuntu-latest because Jest didn't exit. Therefore, I've added the `forceExit` flag to the CLI integration tests.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/model-editor/model-evaluator.ts
+++ b/extensions/ql-vscode/src/model-editor/model-evaluator.ts
@@ -18,6 +18,7 @@ import type { CancellationToken } from "vscode";
 import { CancellationTokenSource } from "vscode";
 import type { QlPackDetails } from "../variant-analysis/ql-pack-details";
 import type { App } from "../common/app";
+import { MultiCancellationToken } from "../common/vscode/multi-cancellation-token";
 import { ModelAlertsView } from "./model-alerts/model-alerts-view";
 
 export class ModelEvaluator extends DisposableObject {
@@ -68,11 +69,11 @@ export class ModelEvaluator extends DisposableObject {
 
     // Submit variant analysis and monitor progress
     return withProgress(
-      (progress) =>
+      (progress, token) =>
         this.runVariantAnalysis(
           qlPack,
           progress,
-          this.cancellationSource.token,
+          new MultiCancellationToken(token, this.cancellationSource.token),
         ),
       {
         title: "Run model evaluation",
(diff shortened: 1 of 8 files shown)


robertbrignull (approved):
Looks ok to me. I haven't spotted anything wrong, and I've tried it out very quickly and it seems to be canceling the process correctly.

extensions/ql-vscode/src/model-editor/model-evaluator.ts
```diff
@@ -75,7 +76,7 @@ export class ModelEvaluator extends DisposableObject {
         ),
       {
         title: "Run model evaluation",
-        cancellable: false,
+        cancellable: true,
```
charisk, line 79: I'm not sure we want this to be cancelable because we have a button on the view to stop the run. @shati-patel?

robertbrignull: Ah yes. Apologies I spotted this in the code but didn't realise it was being changed in this PR.

We can make it cancellable via the progress bar (which would be nice because that's how users normally expect to be able to cancel things) if we use `MultiCancellationToken` on line 75 above.

When calling `this.runVariantAnalysis`, use `new MultiCancellationToken(token, this.cancellationSource.token)` (and add `token` as an arg to the `withProgress` callback).

robertbrignull: I've opened https://github.com/github/vscode-codeql/pull/3464 which would hopefully have caught this, and also helped fine one other alert in our codebase.

koesie10: Thanks! I'll change this to use the `MultiCancellationToken` since it probably makes sense to have a cancel button on a notification that we're showing. If we only want the cancel button in the model editor view, we should probably also only show progress in the model editor view.

shati-patel: 👀 Would it be okay to leave this as non-cancellable for now? 🙏🏽 

I agree that it makes sense for it to be cancellable in future, but we need to do a bit more testing to see that the interactions between the UI buttons and the pop-up work correctly!

koesie10: Yes, sure, I've set `cancellable: false` for now and removed the `MultiCancellationToken` creation.

shati-patel: Thank you! ✨

PR #3455, opened 2024-03-08
Fix incomplete removal of databases

koesie10:
When downloading a database from GitHub, the folder structure looks like this within the `${context.storageUri}/cpp` directory:

Screenshot 2024-03-08 at 13 46 44

When removing this database, we were only deleting the shown `cpp` directory, and not the top-level `cpp` directory because we were deleting the directory that contained the database, rather than the directory that was created during the download.

This fixes it by storing the directory that we create during download on the database item so we can remove that directory when removing the database.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases/database-fetcher.ts
+++ b/extensions/ql-vscode/src/databases/database-fetcher.ts
@@ -400,6 +400,7 @@ async function databaseArchiveFetcher(
       nameOverride,
       {
         addSourceArchiveFolder,
+        extensionManagedLocation: unzipPath,
       },
     );
     return item;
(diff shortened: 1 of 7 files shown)


robertbrignull (approved):
The code change look ok and I assume you've tested it locally that it fixes the bug and works ok. So I'm happy to approve on those grounds.

I'm surprised there are no test changes. Are there no tests of deleting databases?

koesie10:
> I'm surprised there are no test changes. Are there no tests of deleting databases?

Thanks for noticing that. There do seem to be some tests for deleting databases, so I'll add a new one that tests the extension managed location.

robertbrignull (approved):

PR #3456, opened 2024-03-08
Make database storage paths more unique

koesie10:
This tries to pick more unique database storage paths. For GitHub databases, we will now try to use the actual repository name instead of the last part of the URL, which is always the language (e.g. `java` or `cpp`). For example, this is the old and new storage location of the C++ database of `google/brotli`:
- Old: `${storageUri}/cpp/cpp`
- New: `${storageUri}/google-brotli/cpp`

This makes it less likely to run into the 100 databases limit that we have if more than 100 databases have the same storage path. It also makes it easier to find which folder in the storage corresponds to which database.

Separately, this also adds a fallback for the 100 databases limit where we will now try to append a nanoid instead of a counter. So, when all folders from `java-1` to `java-100` exist, we would error out with "Could not find a unique name for downloaded database." before, which wasn't helpful for users (and because of https://github.com/github/vscode-codeql/pull/3455 they couldn't get out of this situation by deleting databases, even if they knew that would help resolve the problem). Now, we will generate a nanoid and append it to the base name, so we would try to use e.g. `java-odkDB1k61hq7PDJnv0qGY` instead. This should essentially remove the databases limit.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/databases/database-fetcher.ts
+++ b/extensions/ql-vscode/src/databases/database-fetcher.ts
@@ -15,6 +15,7 @@ import {
 } from "fs-extra";
 import { basename, join } from "path";
 import type { Octokit } from "@octokit/rest";
+import { nanoid } from "nanoid";
 
 import type { DatabaseManager, DatabaseItem } from "./local-databases";
 import { tmpDir } from "../tmp-dir";
@@ -365,7 +366,11 @@ async function databaseArchiveFetcher(
     throw new Error("No storage path specified.");
   }
   await ensureDir(storagePath);
-  const unzipPath = await getStorageFolder(storagePath, databaseUrl);
+  const unzipPath = await getStorageFolder(
+    storagePath,
+    databaseUrl,
+    nameOverride,
+  );
 
   if (isFile(databaseUrl)) {
     await readAndUnzip(databaseUrl, unzipPath, cli, progress);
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/databases/database-fetcher.ts
```diff
@@ -429,6 +459,11 @@ async function getStorageFolder(storagePath: string, urlStr: string) {
     counter++;
     folderName = join(realpath, `${lastName}-${counter}`);
     if (counter > 100) {
```
robertbrignull, line 461: Is there any need to keep the behaviour of calling databases `java-{1-100}` before we switch to another method? This is only for generating new databases, so it doesn't need to be compatible with existing databases, does it?

Why not use `java-${nanoid()}` all the time?

koesie10: The reason for not doing that is that this name may be shown to the user (as mentioned in [the comment](https://github.com/github/vscode-codeql/pull/3456/files#diff-c165a3c0ec497920fff4a40c198416d0c49572366578137ea82e19620767c256R441-R443): "we need to generate a folder name for the unzipped archive, this needs to be human readable since we may use this name as the initial name for the database"). `java-100` is still more readable than `java-odkDB1k61hq7PDJnv0qGY` and would probably make it easier for users to recognize which database they have just added.

extensions/ql-vscode/src/databases/database-fetcher.ts
```diff
@@ -408,16 +413,41 @@ async function databaseArchiveFetcher(
...
+async function getStorageFolder(
+  storagePath: string,
+  urlStr: string,
+  nameOverrride?: string,
+) {
+  let lastName: string;
+
+  if (nameOverrride) {
+    // Lowercase everything
+    let name = nameOverrride.toLowerCase();
```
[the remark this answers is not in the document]
koesie10: I did find some code that looked similar, but wasn't completely identical. I'll change all of those to use the same code for creating these filenames by introducing a new helper function.

extensions/ql-vscode/src/databases/database-fetcher.ts
robertbrignull: Thanks for explaining. It seems we use the filename only if there wasn't a name override (which comes when downloading from github) or and the database itself doesn't define a name (I don't know how common this is, but I expect any database built from a git repo will have a name).

robertbrignull: Another alternative thought I had was what's the relative performance of checking a single file exists vs listing all files in the directory? If we just list all files and then do our checks in memory I would expect it to work fine up to thousands of databases.

extensions/ql-vscode/src/databases/database-fetcher.ts
```diff
@@ -429,6 +445,11 @@ async function getStorageFolder(storagePath: string, urlStr: string) {
     counter++;
     folderName = join(realpath, `${lastName}-${counter}`);
     if (counter > 100) {
+      // If there are more than 100 similarly named databases,
+      // give up on using a counter and use a random string instead.
+      folderName = join(realpath, `${lastName}-${nanoid()}`);
+    }
+    if (counter > 200) {
       throw new Error("Could not find a unique name for downloaded database.");
```
robertbrignull, line 453: Although this error shouldn't happen anymore if we use `nanoid`, shall we still update this error message to indicate that the problem is to do with having too many databases?

Maybe just:
```suggestion
      throw new Error("Could not find a unique name for downloaded database. Please remove some databases and try again.");
```
?

extensions/ql-vscode/src/databases/database-fetcher.ts
koesie10: Yes, that would make sense. I'll change it to `readdir` once and change the maximum counter value to 10,000.

extensions/ql-vscode/src/databases/database-fetcher.ts
```diff
@@ -408,31 +415,54 @@ async function databaseArchiveFetcher(
...
-    folderName = join(realpath, `${lastName}-${counter}`);
-    if (counter > 100) {
-      throw new Error("Could not find a unique name for downloaded database.");
+    folderName = `${lastName}-${counter}`;
+    if (counter > 10_000) {
+      // If there are more than 10,000 similarly named databases,
+      // give up on using a counter and use a random string instead.
+      folderName = `${lastName}-${nanoid()}`;
+    }
+    if (counter > 20_000) {
```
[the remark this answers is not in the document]
koesie10: I've set it to be 5 nanoid tries for now which should ensure that we never run into the error condition.

robertbrignull (approved):
LGTM 👍🏼

PR #3457, opened 2024-03-08
Use suite file instead of manually filtering queries

koesie10:
When resolving the Code Scanning pack, we were manually filtering out non-problem queries by calling `codeql resolve metadata` on every resolved query. This switches it to using a custom suite file (`.qls`) which can do the same but without requiring many calls to `codeql resolve metadata`. There should be no changes in behavior.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
+++ b/extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
@@ -29,10 +29,6 @@ export async function resolveCodeScanningQueryPack(
   // Resolve queries
   void logger.log(`Resolving queries for pack: ${packName}`);
 
-  const suiteFile = await file({
-    postfix: ".qls",
-  });
-  const suitePath = suiteFile.path;
   const suiteYaml: SuiteInstruction[] = [
     {
       import: `codeql-suites/${language}-code-scanning.qls`,
@@ -45,11 +41,20 @@ export async function resolveCodeScanningQueryPack(
       },
     },
   ];
-  await outputFile(suitePath, dump(suiteYaml), "utf8");
 
-  const resolvedQueries = await cliServer.resolveQueries(suitePath);
+  let resolvedQueries: string[];
+  const suiteFile = await file({
+    postfix: ".qls",
+  });
+  const suitePath = suiteFile.path;
 
-  await suiteFile.cleanup();
+  try {
+    await outputFile(suitePath, dump(suiteYaml), "utf8");
+
+    resolvedQueries = await cliServer.resolveQueries(suitePath);
+  } finally {
+    await suiteFile.cleanup();
+  }
 
   if (resolvedQueries.length === 0) {
     throw Error(
--- a/extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
+++ b/extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
@@ -1,10 +1,13 @@
 import { join } from "path";
+import { outputFile } from "fs-extra";
+import { dump } from "js-yaml";
+import { file } from "tmp-promise";
 import type { BaseLogger } from "../common/logging";
 import type { QueryLanguage } from "../common/query-language";
 import type { CodeQLCliServer } from "../codeql-cli/cli";
 import type { QlPackDetails } from "./ql-pack-details";
 import { getQlPackFilePath } from "../common/ql";
-import { isSarifResultsQueryKind } from "../common/query-metadata";
+import type { SuiteInstruction } from "../packaging/suite-instruction";
 
 export async function resolveCodeScanningQueryPack(
   logger: BaseLogger,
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
```diff
@@ -25,20 +28,30 @@ export async function resolveCodeScanningQueryPack(
...
+  await outputFile(suitePath, dump(suiteYaml), "utf8");
+
   const resolvedQueries = await cliServer.resolveQueries(suitePath);
 
-  const problemQueries = await filterToOnlyProblemQueries(
-    logger,
-    cliServer,
-    resolvedQueries,
-  );
+  await suiteFile.cleanup();
```
robertbrignull, line 52: To avoid leaving this file around in the case of an error, could this be in a `finally` block?

extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
```diff
@@ -25,20 +28,30 @@ export async function resolveCodeScanningQueryPack(
...
+  const suiteYaml: SuiteInstruction[] = [
+    {
+      import: `codeql-suites/${language}-code-scanning.qls`,
+      from: `${downloadedPack.name}@${downloadedPack.version}`,
+    },
+    // Exclude any non-problem queries
+    {
+      exclude: {
+        kind: ["diagnostic", "metric"],
+      },
```
robertbrignull, line 45: Would
```
include: {
  kind: SARIF_RESULTS_QUERY_KINDS,
}
```
have the same effect? Or maybe include both filters?

The original intention here was that we limit to only queries that can produce SARIF because we don't want to have to deal with displaying raw results from multiple queries right now. That'll come later once we move multi-query MRVA out of the prototype phase.

robertbrignull (commented):
Thanks for looking into this. I wasn't originally sure if the idea was to use this constructed suite file for running the queries too, but just using it temporarily to obtain the list of queries makes perfect sense. Then the code from here afterwards is unchanged.

extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
koesie10: I tested this and unfortunately this doesn't have any effect. After including an existing suite, you can only override it by using `exclude`, not by using `include`. `include` would simply add all queries that match that `kind`, but wouldn't exclude any of the existing already included queries.

robertbrignull: It's buried in the documentation at the end of [Reusability Examples](https://docs.github.com/en/code-security/codeql-cli/using-the-advanced-functionality-of-the-codeql-cli/creating-codeql-query-suites#reusability-examples) but I believe we can follow this example:
```yaml
- import: codeql-suites/cpp-security-and-quality.qls
  from: codeql/cpp-queries
- exclude: {}
- include:
    precision:
      - very-high
      - high
```

> Notice the empty `exclude` instruction. This is required to ensure that the subsequent `include` instruction is able to filter queries from the imported suite.

I tried it out locally with the following suite file and it seems to behave as intended.

robertbrignull: I don't believe it will cause it to include queries that weren't in the original code scanning suite. That would be the difference between `import` and `apply`.

I tried to test this. The following suite contains 75 queries:
```yaml
- import: codeql-suites/java-code-scanning.qls
  from: codeql/java-queries@0.8.9
```

If I exclude `diagnostic` and `metric` queries then it decreases to 62 queries:
```yaml
- import: codeql-suites/java-code-scanning.qls
  from: codeql/java-queries@0.8.9
- exclude:
    kind:
      - diagnostic
      - metric
```

If I instead exclude all queries and then only import `problem` and `path-problem` queries then we get the exact same 62 queries:
```yaml
- import: codeql-suites/java-code-scanning.qls
  from: codeql/java-queries@0.8.9
- exclude: {}
- include:
    kind:
      - problem
      - path-problem
```

And to show the difference vs using `apply`, if I change `import` to `apply` in the last case then it sky rockets to 453 queries:
```yaml
- apply: codeql-suites/java-code-scanning.qls
  from: codeql/java-queries@0.8.9
- exclude: {}
- include:
    kind:
      - problem
      - path-problem
```

---

However, this likely doesn't matter and right now we could choose either option. As shown they give the exact same result because we know that the code scanning suites will only contain problem queries and diagnostic/metric queries.

My (possibly incorrect) understanding of the future for this feature is that eventually (no timeline yet) we'll implement viewing raw results and we'll likely also want to expand it to run on custom suites as well as the code scanning suites. For the code scanning suites we'll still want to exclude the metric/diagnostic queries as they're likely not what the user wants. For custom suites if we haven't implemented raw results yet then we'll want to limit it to problem queries, but it's also likely that implementing raw results will come first so it won't matter by that point anyway.

In short, I think including the kinds we want is still a bit cleaner than excluding the ones we don't, but for the code scanning suites they're equivalent and should remain that way for the foreseeable future. If you like the current approach and want to keep it as it is that's fine.

robertbrignull (approved):

extensions/ql-vscode/src/variant-analysis/code-scanning-pack.ts
koesie10: Thanks, I didn't realize that when specifying an empty `exclude`, the other instructions from the imported file still apply. I agree that only including the query kinds we support makes more sense, so I'll change this to use the empty `exclude` followed by an `include` with `SARIF_RESULTS_QUERY_KINDS`.

robertbrignull (approved):

robertbrignull:
I also did a quick manual test and everything ran correctly and it ran the correct 62 queries (for Java).

PR #3461, opened 2024-03-12
Only allow one open model alerts view

shati-patel:
There should only ever be one model alerts view open, so this PR adds some logic to check that we don't already have one! 👀 

The second commit is a related tweak that makes sure we can't open the model alerts view until the variant analysis has kicked off. (This is similar to the MRVA results view, which only opens once the actual workflow run has been triggered, since there's nothing to show otherwise.)

## Checklist

N/A—feature-flagged for internal use/testing 🏴 

### DIFF
--- a/extensions/ql-vscode/src/model-editor/model-evaluation-run.ts
+++ b/extensions/ql-vscode/src/model-editor/model-evaluation-run.ts
@@ -1,5 +1,4 @@
 export interface ModelEvaluationRun {
   isPreparing: boolean;
   variantAnalysisId: number | undefined;
-  isModelAlertsViewOpen?: boolean;
 }
--- a/extensions/ql-vscode/src/model-editor/model-evaluation-run.ts
+++ b/extensions/ql-vscode/src/model-editor/model-evaluation-run.ts
@@ -1,4 +1,5 @@
 export interface ModelEvaluationRun {
   isPreparing: boolean;
   variantAnalysisId: number | undefined;
+  isModelAlertsViewOpen?: boolean;
 }
--- a/extensions/ql-vscode/src/model-editor/model-evaluator.ts
+++ b/extensions/ql-vscode/src/model-editor/model-evaluator.ts
@@ -107,8 +107,27 @@ export class ModelEvaluator extends DisposableObject {
   }
 
   public async openModelAlertsView() {
-    const view = new ModelAlertsView(this.app);
-    await view.showView();
+    const evaluationRun = this.modelingStore.getModelEvaluationRun(this.dbItem);
+    if (evaluationRun === undefined) {
+      return;
+    } else if (this.modelingStore.isModelAlertsViewOpen(evaluationRun)) {
+      this.modelingEvents.fireFocusModelAlertsViewEvent(
+        this.dbItem.databaseUri.toString(),
+      );
+      return;
+    } else {
+      this.modelingStore.updateModelEvaluationRun(this.dbItem, {
+        ...evaluationRun,
+        isModelAlertsViewOpen: true,
+      });
+      const view = new ModelAlertsView(
+        this.app,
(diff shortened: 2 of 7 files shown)


extensions/ql-vscode/src/model-editor/model-evaluation-run.ts
```diff
@@ -1,4 +1,5 @@
 export interface ModelEvaluationRun {
   isPreparing: boolean;
   variantAnalysisId: number | undefined;
+  isModelAlertsViewOpen?: boolean;
```
charisk, line 4: Shall we make this a non-optional boolean to avoid having to deal with `undefined`?

I'm in two minds about whether the `ModelEvaluationRun` domain entity is the right place to hold information around whether a view is open or not. It feels a bit wrong because it's not a property of the evaluation run, just some information around the user's modeling setup. So it could be information we have in the `InternalDbModelingState` instead. Any thoughts?

extensions/ql-vscode/src/model-editor/model-evaluator.ts
```diff
@@ -107,8 +107,27 @@ export class ModelEvaluator extends DisposableObject {
   }
 
   public async openModelAlertsView() {
-    const view = new ModelAlertsView(this.app);
-    await view.showView();
+    const evaluationRun = this.modelingStore.getModelEvaluationRun(this.dbItem);
+    if (evaluationRun === undefined) {
+      return;
+    } else if (this.modelingStore.isModelAlertsViewOpen(evaluationRun)) {
```
charisk, line 113: Do we need to go to the store for this or can we just do

```suggestion
    } else if (evaluationRun.isModelAlertsViewOpen) { 
```

charisk (commented):
Generally looking good, just a question around where the information will be stored

extensions/ql-vscode/src/model-editor/model-evaluation-run.ts
shati-patel: I think I agree that `ModelEvaluationRun` feels wrong 🤔 I hadn't thought of `InternalDbModelingState`, but will try adding it there instead!

shati-patel:
Pushed a new commit to add `isModelAlertsViewOpen` to `(Internal)DbModelingState` instead. It works as expected (from testing manually), but I'm not 100% sure I've done things correctly about when to use public/private/internal... 🤔 😅

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -20,6 +20,7 @@ interface InternalDbModelingState {
   selectedMethod: Method | undefined;
   selectedUsage: Usage | undefined;
   modelEvaluationRun: ModelEvaluationRun | undefined;
+  isModelAlertsViewOpen: boolean | undefined;
```
charisk, line 23: Shall we drop the possibility of this being undefined?

extensions/ql-vscode/src/model-editor/modeling-store.ts
```diff
@@ -499,7 +502,12 @@ export class ModelingStore extends DisposableObject {
     );
   }
 
-  public isModelAlertsViewOpen(evaluationRun: ModelEvaluationRun): boolean {
-    return evaluationRun.isModelAlertsViewOpen ?? false;
+  public isModelAlertsViewOpen(dbItem: DatabaseItem): boolean {
+    return this.getState(dbItem).isModelAlertsViewOpen ?? false;
+  }
+
+  public updateIsModelAlertsViewOpen(dbItem: DatabaseItem, isOpen: boolean) {
```
charisk, line 509: Can we follow the pattern we use elsewhere for update methods?

charisk (approved):
LGTM!

PR #3464, opened 2024-03-13
Create token-not-used.ql to catch cases where we aren't using the progress bar token correctly

robertbrignull:
This PR adds a CodeQL query to catch cases where we use `withProgress` with `cancellable: true` but then don't actually use the token argument. This will lead to cases where the user clicks the "cancel" button and the progress bar goes away but the background process continues and the user cannot control it.

This was inspired by https://github.com/github/vscode-codeql/pull/3454#discussion_r1523033368 which was only caught because we manually spotted it.

This should hopefully also detect one other existing alert at https://github.com/github/vscode-codeql/blob/3005dacf4ed480aa76e541b7d3697f5a34571faf/extensions/ql-vscode/src/databases/local-databases-ui.ts#L950.

## Checklist

### DIFF
--- a/.github/codeql/queries/token-not-used.ql
+++ b/.github/codeql/queries/token-not-used.ql
@@ -1,5 +1,5 @@
 /**
- * @name Don't ignore the token for a cancelable progress bar
+ * @name Don't ignore the token for a cancellable progress bar
  * @kind problem
  * @problem.severity warning
  * @id vscode-codeql/token-not-used
--- a/.github/codeql/queries/token-not-used.ql
+++ b/.github/codeql/queries/token-not-used.ql
@@ -10,32 +10,8 @@
  */
 
 import javascript
+import ProgressBar
 
-class NewTokenSource extends CallExpr {
-  NewTokenSource() {
-    this.getCalleeName() = "withProgress" or this.getCalleeName() = "withInheritedProgress"
-  }
-
-  Function getCallback() {
-    this.getCalleeName() = "withProgress" and result = this.getArgument(0)
-    or
-    this.getCalleeName() = "withInheritedProgress" and result = this.getArgument(1)
-  }
-
-  ObjectExpr getOptions() {
-    this.getCalleeName() = "withProgress" and result = this.getArgument(1)
-    or
-    this.getCalleeName() = "withInheritedProgress" and result = this.getArgument(2)
-  }
-
-  predicate usesToken() { this.getCallback().getNumParameter() >= 2 }
-
-  predicate isCancellable() {
-    this.getOptions().getPropertyByName("cancellable").getInit().(BooleanLiteral).getBoolValue() =
-      true
-  }
-}
-
-from NewTokenSource t
+from ProgressBar t
 where t.isCancellable() and not t.usesToken()
 select t, "This progress bar is cancelable but the token is not used"
--- a/.github/codeql/queries/token-not-used.ql
+++ b/.github/codeql/queries/token-not-used.ql
@@ -1,0 +1,41 @@
+/**
+ * @name Don't ignore the token for a cancelable progress bar
+ * @kind problem
+ * @problem.severity warning
+ * @id vscode-codeql/token-not-used
+ * @description If we call `withProgress` with `cancellable: true` but then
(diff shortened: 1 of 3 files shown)


.github/codeql/queries/token-not-used.ql
```diff
@@ -0,0 +1,41 @@
...
+
+class NewTokenSource extends CallExpr {
+  NewTokenSource() {
+    this.getCalleeName() = "withProgress" or this.getCalleeName() = "withInheritedProgress"
+  }
+
+  Function getCallback() {
+    this.getCalleeName() = "withProgress" and result = this.getArgument(0)
+    or
+    this.getCalleeName() = "withInheritedProgress" and result = this.getArgument(1)
```
koesie10, line 22: Can we factor this out into a private predicate `isInherited` such that we're not repeating these names 6 times in this file?

.github/codeql/queries/token-not-used.ql
```diff
@@ -0,0 +1,41 @@
...
+
+  predicate isCancellable() {
+    this.getOptions().getPropertyByName("cancellable").getInit().(BooleanLiteral).getBoolValue() =
+      true
+  }
+}
+
+from NewTokenSource t
+where t.isCancellable() and not t.usesToken()
+select t, "This progress bar is cancelable but the token is not used"
```
koesie10, line 41: I think it might be nice to make `cancellable` an entity here so it's clickable. Something like this:

```suggestion
select t, "This progress bar is $@ but the token is not used", t.getCancellableProperty(),
  t.getCancellableProperty().getName()
```

I'll leave a comment on `isCancellable` for how I think `getCancellableProperty` can be implemented.

.github/codeql/queries/token-not-used.ql
```diff
@@ -0,0 +1,41 @@
...
+    or
+    this.getCalleeName() = "withInheritedProgress" and result = this.getArgument(2)
+  }
+
+  predicate usesToken() { this.getCallback().getNumParameter() >= 2 }
+
+  predicate isCancellable() {
+    this.getOptions().getPropertyByName("cancellable").getInit().(BooleanLiteral).getBoolValue() =
+      true
+  }
```
koesie10, line 36: See my comment below:

```suggestion
  Property getCancellableProperty() { result = this.getOptions().getPropertyByName("cancellable") }

  predicate isCancellable() {
    this.getCancellableProperty().getInit().(BooleanLiteral).getBoolValue() = true
  }
```

koesie10 (approved):
Thanks, LGTM, I just have a few minor suggestions.

Do you think it would also make sense to create a query for the other scenario where we're using the `token`, but the progress bar is not cancellable (even though this is less of a UX problem)?

robertbrignull:
> Do you think it would also make sense to create a query for the other scenario where we're using the `token`, but the progress bar is not cancellable (even though this is less of a UX problem)?

Ooh good idea. Yes, we could have a query for the opposite case, and it would be basically the same so we could share a lot of the machinery.

.github/codeql/queries/token-not-used.ql
robertbrignull: I think I see what you mean. Sadly that would only reduce it to 4 times because annoyingly I believe you can't reference member predicates in the class predicate thing (I can't remember the real name).

We could make it an abstract class with two subclasses and that way we avoid all duplication. I end using abstract classes in a bunch of my queries so I hope I'm not overusing them, but they are quite useful.

I'll make this change and push it so you can see what I mean.

.github/codeql/queries/token-not-used.ql
robertbrignull: I've ended up only putting this link into the query for when the progress bar is cancellable because then there's always something to link to. For the other query there usually won't be because it's rare to explicitly do `cancellable: false`.

robertbrignull:
The "token is being used even though progress bar is not cancellable" query gets 10 results and as far as I can tell they're all true positives. Admittedly this is less important than the other case because it's just dead code instead of a bad user experience. I think it's arguably still worth changing if it makes our code simpler.

Check out https://github.com/github/vscode-codeql/security/code-scanning?query=is%3Aopen+pr%3A3464 for the list of alerts found by these queries.

PR #3490, opened 2024-03-19
Fix ModelAlerts view stories

charisk:
The `ModelAlerts` view won't return any HTML if the `variantAnalysis` isn't set, so I've updated the component to be able to receive that, along with repo states and repo results, through props. We follow this pattern elsewhere.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/stories/model-alerts/ModelAlerts.stories.tsx
+++ b/extensions/ql-vscode/src/stories/model-alerts/ModelAlerts.stories.tsx
@@ -1,6 +1,7 @@
 import type { Meta, StoryFn } from "@storybook/react";
 
 import { ModelAlerts as ModelAlertsComponent } from "../../view/model-alerts/ModelAlerts";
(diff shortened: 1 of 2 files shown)


norascheuch (commented):
Was this noticeable in the extension itself? Your changes are in the general ModelAlerts view, did we have a bug?

charisk:
> Was this noticeable in the extension itself? Your changes are in the general ModelAlerts view, did we have a bug?

No there's no bug (also reminder that this view is for a prototype, not a released thing, so it probably has lots of "bugs" 😬 ).

It's just that the component returns [empty HTML if variant analysis isn't set](https://github.com/github/vscode-codeql/pull/3490/files#diff-90d193d10610be695e61ca64cc16a88b1a928bdcf10ae7e8685079a3a0127d10R101) so the story is kind of pointless without it.

The changes to the view make it possible to make it "storyable".

norascheuch (approved):
🚢

PR #3503, opened 2024-03-21
Wire up processing model evaluation run results and showing alerts

charisk:
Wires up the model evaluator to trigger loading of variant analysis repository results, process them and update the model alerts view.

In the absence of alert provenance information, we fake process the analysis results for now, and return mock results.

## Checklist
N/A:

### DIFF
--- a/extensions/ql-vscode/src/common/interface-types.ts
+++ b/extensions/ql-vscode/src/common/interface-types.ts
@@ -732,6 +732,11 @@ interface SetModelAlertsViewStateMessage {
   viewState: ModelAlertsViewState;
 }
 
+interface SetReposResultsMessage {
+  t: "setReposResults";
+  reposResults: VariantAnalysisScannedRepositoryResult[];
+}
+
 interface OpenModelPackMessage {
   t: "openModelPack";
   path: string;
@@ -749,7 +754,8 @@ interface StopEvaluationRunMessage {
 export type ToModelAlertsMessage =
   | SetModelAlertsViewStateMessage
   | SetVariantAnalysisMessage
-  | SetRepoResultsMessage;
+  | SetRepoResultsMessage
+  | SetReposResultsMessage;
 
 export type FromModelAlertsMessage =
   | CommonFromViewMessages
--- a/extensions/ql-vscode/src/model-editor/model-evaluator.ts
+++ b/extensions/ql-vscode/src/model-editor/model-evaluator.ts
@@ -268,16 +268,15 @@ export class ModelEvaluator extends DisposableObject {
 
     this.push(
       this.variantAnalysisManager.onRepoStatesUpdated(async (e) => {
-        if (e.variantAnalysisId === variantAnalysisId) {
-          if (
-            e.repoState.downloadStatus ===
+        if (
+          e.variantAnalysisId === variantAnalysisId &&
+          e.repoState.downloadStatus ===
             VariantAnalysisScannedRepositoryDownloadStatus.Succeeded
-          ) {
-            await this.readAnalysisResults(
-              variantAnalysisId,
-              e.repoState.repositoryId,
-            );
-          }
+        ) {
+          await this.readAnalysisResults(
+            variantAnalysisId,
+            e.repoState.repositoryId,
+          );
         }
       }),
     );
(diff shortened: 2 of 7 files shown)


extensions/ql-vscode/src/common/interface-types.ts
```diff
@@ -749,7 +754,8 @@ interface StopEvaluationRunMessage {
 export type ToModelAlertsMessage =
   | SetModelAlertsViewStateMessage
   | SetVariantAnalysisMessage
-  | SetRepoResultsMessage;
+  | SetRepoResultsMessage
+  | SetReposResultsMessage;
```
koesie10, line 758: I think the difference between the `SetRepoResultsMessage` and `SetReposResultsMessage` is not clear from the name. The data seems to be exactly the same, so it's quite confusing (at least to me) when we have both messages and they have different functions. Perhaps a name like `ReplaceRepoResultsMessage` would be better (based on its function in the view code)? The same might be useful for the functions on the model alerts view, something like `updateRepoResult` and `replaceRepoResults`?

extensions/ql-vscode/src/model-editor/model-evaluator.ts
```diff
@@ -260,6 +266,22 @@ export class ModelEvaluator extends DisposableObject {
       ),
     );
 
+    this.push(
+      this.variantAnalysisManager.onRepoStatesUpdated(async (e) => {
+        if (e.variantAnalysisId === variantAnalysisId) {
+          if (
+            e.repoState.downloadStatus ===
+            VariantAnalysisScannedRepositoryDownloadStatus.Succeeded
+          ) {
```
koesie10, line 275: Could this be a single `if`-statement with an `&&` to reduce nesting?

extensions/ql-vscode/src/model-editor/model-evaluator.ts
```diff
@@ -268,4 +290,37 @@ export class ModelEvaluator extends DisposableObject {
...
+    if (!repository) {
+      void this.app.logger.log(
+        `Could not find reposiotry with id ${repositoryId} in scanned repos`,
+      );
+      throw new Error(
+        "There was an error when trying to retrieve repository information",
+      );
+    }
+
+    await this.variantAnalysisManager.loadResults(
```
koesie10, line 321: Perhaps add a comment here that this will trigger the `onRepoResultsLoaded` event handler? Otherwise it seems like the result is unused.

extensions/ql-vscode/src/model-editor/model-evaluator.ts
```diff
@@ -268,4 +290,37 @@ export class ModelEvaluator extends DisposableObject {
...
+        "There was an error when trying to retrieve variant analysis information",
+      );
+    }
+
+    const repository = variantAnalysis.scannedRepos?.find(
+      (r) => r.repository.id === repositoryId,
+    );
+    if (!repository) {
+      void this.app.logger.log(
+        `Could not find reposiotry with id ${repositoryId} in scanned repos`,
```
koesie10, line 314:
```suggestion
        `Could not find repository with id ${repositoryId} in scanned repos`,
```

extensions/ql-vscode/src/view/model-alerts/ModelAlerts.tsx
```diff
@@ -81,6 +96,18 @@ export function ModelAlerts({
...
   }, []);
 
+  const modelAlerts = useMemo(() => {
+    if (!repoResults) {
+      return [];
+    }
+
+    const alerts = repoResults
+      .filter((r) => r.interpretedResults !== undefined)
+      .flatMap((a) => a.interpretedResults!);
```
koesie10, line 106: Can we avoid the `!` here by doing something like this?

```suggestion
    const alerts = repoResults.flatMap((a) => a.interpretedResults ?? []);
```

extensions/ql-vscode/src/common/interface-types.ts
charisk: Agree it's quite confusing. I think that `setReposResultsMessage` is the correct name for what I'm adding, but that the other messages (and related view functions) should be "addRepoResults" or something like that. Essentially `set` should be used when we're actually setting the whole thing, otherwise we should use some other term. 

I was hoping to tidy this up in a future PR, what do you think?

extensions/ql-vscode/src/model-editor/model-evaluator.ts
charisk: Good idea! Done, let me know if it's not clear.

extensions/ql-vscode/src/common/interface-types.ts
koesie10: That makes sense to me, I don't think this needs to be blocking for this PR.

koesie10 (approved):
LGTM

PR #3515, opened 2024-03-26
Use file handle in TeeLogger

koesie10:
This switches the `TeeLogger` from using the `appendFile` function to manually opening and closing a file handle and calling `appendFile` on the file handle. This results in a more efficient implementation as the file handle is only opened once and closed once, instead of being opened and closed for every log message. This should result in less "too many open files" errors.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/disposable-object.ts
+++ b/extensions/ql-vscode/src/common/disposable-object.ts
@@ -88,7 +88,3 @@ export class DisposableObject implements Disposable {
     }
   }
 }
-
-export function isDisposable(obj: unknown): obj is Disposable {
-  return obj !== undefined && typeof (obj as Disposable).dispose === "function";
-}
--- a/extensions/ql-vscode/src/common/disposable-object.ts
+++ b/extensions/ql-vscode/src/common/disposable-object.ts
@@ -88,3 +88,7 @@ export class DisposableObject implements Disposable {
     }
   }
 }
+
+export function isDisposable(obj: unknown): obj is Disposable {
+  return obj !== undefined && typeof (obj as Disposable).dispose === "function";
+}
--- a/extensions/ql-vscode/src/common/logging/tee-logger.ts
+++ b/extensions/ql-vscode/src/common/logging/tee-logger.ts
@@ -58,6 +58,13 @@ export class TeeLogger implements Logger, Disposable {
       } catch (e) {
         // Write an error message to the primary log, and stop trying to write to the side log.
         this.error = true;
+        try {
+          await this.fileHandle?.close();
+        } catch (e) {
+          void this.logger.log(
+            `Failed to close file handle: ${getErrorMessage(e)}`,
+          );
+        }
         this.fileHandle = undefined;
         const errorMessage = getErrorMessage(e);
         await this.logger.log(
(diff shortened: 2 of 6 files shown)


extensions/ql-vscode/src/common/disposable-object.ts
```diff
@@ -88,3 +88,7 @@ export class DisposableObject implements Disposable {
     }
   }
 }
+
+export function isDisposable(obj: unknown): obj is Disposable {
+  return obj !== undefined && typeof (obj as Disposable).dispose === "function";
```
robertbrignull, line 93: I believe this will throw an error if `obj` is `null`.

You could add a `obj !== null` check, but you could write it without using `as` at all like:
```typescript
return (
  obj !== null &&
  typeof obj === "object" &&
  "dispose" in obj &&
  typeof obj.dispose === "function"
);
```

extensions/ql-vscode/src/local-queries/local-query-run.ts
```diff
@@ -61,6 +62,10 @@ export class LocalQueryRun {
     private readonly localQueries: LocalQueries,
     private readonly queryInfo: LocalQueryInfo,
     private readonly dbItem: DatabaseItem,
+    /**
+     * The logger is only available while the query is running and will be disposed of when the
+     * query completes.
+     */
     public readonly logger: Logger, // Public so that other clients, like the debug adapter, know where to send log output
```
robertbrignull, line 69: I believe that we always pass a logger that is disposable. Is it worth putting that in the type here instead of using `isDisposable` in every place?

If we don't want to define a new `DispoableLogger` type we could instead do `logger: Logger & Disposable` here.

extensions/ql-vscode/src/common/logging/tee-logger.ts
```diff
@@ -50,6 +58,7 @@ export class TeeLogger implements Logger {
       } catch (e) {
         // Write an error message to the primary log, and stop trying to write to the side log.
         this.error = true;
+        this.fileHandle = undefined;
```
robertbrignull, line 61: Are you sure we shouldn't call `await this.fileHandle?.close()` here? I'm wondering if the exception could have been from something that didn't come from the file not being open correctly. In that case we'll leak the file handle and open a new one next time we try to log.

extensions/ql-vscode/src/local-queries/local-query-run.ts
```diff
@@ -92,6 +97,10 @@ export class LocalQueryRun {
     // Note we must update the query history view after showing results as the
     // display and sorting might depend on the number of results
     await this.queryHistoryManager.refreshTreeView();
+
+    if (isDisposable(this.logger)) {
```
robertbrignull, line 101: I think the disposal of the logger is currently being done in all the right places. I was wondering whether it would be clearer and safer to make `LocalQueryRun` implement `Disposable`. I decided that I don't think it matters much. It would make some things more consistent but we'd hit a blocker in `debugger-ui.ts` and ultimately I'm not sure it brings any extra safety.

So overall I don't think anything needs to change. I just wanted to post my thoughts on it 🤷🏼

extensions/ql-vscode/src/local-queries/local-query-run.ts
koesie10: Good idea, I've used `Logger & Disposable` here and removed the `isDisposable` function.

extensions/ql-vscode/src/common/logging/tee-logger.ts
koesie10: Good point, we should close it here. There isn't a risk that we open a new file handle to the same file next time we try to log since `this.error` is never reset, but we won't close the file handle in `dispose` either (since `this.fileHandle` is now `undefined`), so we should close it here.

robertbrignull (approved):
LGTM

PR #3517, opened 2024-03-26
Mark progress bars as cancellable where it appears we are respecting the token

robertbrignull:
This PR should fix all instances of https://github.com/github/vscode-codeql/security/code-scanning?query=is%3Aopen+branch%3Amain+rule%3Avscode-codeql%2Fprogress-not-cancellable

These are cases where the progress bar is not marked as cancellable, but we are using the token as if it were cancellable. Theoretically everything should just work if we say the progress bar is cancellable, and this gives the user more power to cancel long-running operations.

## Checklist


charisk (approved):
Code changes LGTM. 

Is it worth testing each of these actions to make sure they are truly cancellable (that the token passed is actually used appropriately)? Otherwise we could be giving people a cancel button that is broken.

extensions/ql-vscode/src/extension.ts
```diff
@@ -203,6 +203,7 @@ function getCommands(
       },
       {
         title: "Restarting Query Server",
+        cancellable: true,
```
koesie10, line 206: I'm not sure if making these cancellable makes sense. It doesn't look like this will actually cancel the restart, but will instead only cancel some tasks that are run after the restart. This is probably not what the user would expect.

A quick search of those handlers only turns up [`reregisterDatabases`](https://github.com/github/vscode-codeql/blob/31cfaebb598afb144b833beba85fc7ddc9d81dea/extensions/ql-vscode/src/databases/local-databases/database-manager.ts#L421), which actually doesn't use the token at all.

extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -673,6 +673,7 @@ export class DatabaseUI extends DisposableObject {
       },
       {
         title: "Clearing cache",
+        cancellable: true,
```
koesie10, line 676: I'm not sure if these should be cancellable. These seem to be short-running commands, and the execution of the command itself is not actually cancellable in the CodeQL CLI. The only part that is cancellable is getting an exclusive lock on the database, but that probably isn't clear to the user.

extensions/ql-vscode/src/query-server/query-server-client.ts
```diff
@@ -125,6 +125,7 @@ export class QueryServerClient extends DisposableObject {
           this.restartQueryServerInternal(progress, token),
         {
           title: "Restarting CodeQL query server due to unexpected termination",
+          cancellable: true,
```
koesie10, line 128: I don't think this should be cancellable either for the same reason as the other query server restart.

koesie10 (commented):
I'm not sure most of these should be cancellable. It would probably be better to remove the token instead.

robertbrignull:
Thanks for looking deeper into what the cancellation token actually does in these commands. I was originally hoping that because we pass it to the query server that it'll be used accordingly. I've now also dug into the code for the query server and I agree those three cases aren't actually as cancellable as they appear. Since we expect restarting the query server and trimming the cache to be pretty fast I agree we don't need to make them cancellable. It also isn't any change in behaviour from before this PR and is just cleaning up the code so it should be safe.

robertbrignull:
I've updated the PR to only make the model editor one cancellable and for the query server ones leave them as not cancellable and clean up the code instead.

I decided to remove the `token` argument from the query runner `onStart` since all of the uses of it passed a callback that didn't actually use the token anyway.

I've also quickly tested each of these cases manually just to make sure they work still.

koesie10 (approved):
LGTM, thanks!

PR #3518, opened 2024-03-27
Refactor: extract function for canonicalising modeled methods

shati-patel:
[🍐 paired with Charis]

We don't currently (I think) have an easy way to check whether two modeled methods are the same. I've shuffled some stuff around, so that you can now compare modeled methods by checking that their "key" is the same, i.e. `createModeledMethodKey(method1) === createModeledMethodKey(method2)`.

(Not used yet, but I plan to use this for jumping between corresponding models in the model editor and the model alerts view)

## Checklist

N/A: no user-facing changes

### DIFF
--- a/extensions/ql-vscode/src/model-editor/modeled-method.ts
+++ b/extensions/ql-vscode/src/model-editor/modeled-method.ts
@@ -1,3 +1,4 @@
+import { assertNever } from "../common/helpers-pure";
 import type { MethodSignature } from "./method";
 import type { ModelingStatus } from "./shared/modeling-status";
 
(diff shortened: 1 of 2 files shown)


charisk (approved):
LGTM, but since we paired, I'd be keen to get reviews from others too.

koesie10 (approved):
LGTM, but it would be nice to add some tests. This method was implicitly tested by `validation.test.ts` before, but now that it has been moved around and is in a different file, it might be good to add some tests for this method specifically.

shati-patel:
Good idea about unit tests! I'll do that in a follow-up PR 👀

PR #3527, opened 2024-03-28
Fix the qlpack json schema

aeisenberg:
This schema was incorrectly tagging `defaultSuite` properties that aren't arrays as errors.

I think there are more, similar errors in this schema, but I only fixed this one so that I can get `github/semmle-code` working with test discovery.

## Checklist


extensions/ql-vscode/src/packaging/suite-instruction.ts
```diff
@@ -5,8 +5,8 @@ export interface SuiteInstruction {
   qlpack?: string;
   query?: string;
   queries?: string;
-  include?: Record<string, string[]>;
-  exclude?: Record<string, string[]>;
+  include?: Record<string, string[]> | Record<string, string>;
+  exclude?: Record<string, string[]> | Record<string, string>;
```
koesie10, line 9: Should this be something like this because it's possible to mix `string` and `string[]` in the same `include`/`exclude` property?

```suggestion
  include?: Record<string, string | string[]>;
  exclude?: Record<string, string | string[]>;
```

koesie10 (commented):
Thanks for looking into this! I think we should be more lenient on the `include`/`exclude` properties, but the other changes look good to me.

extensions/ql-vscode/src/packaging/suite-instruction.ts
aeisenberg: Yes. Thanks. That's a better option. I'll make that change.

PR #3530, opened 2024-04-03
Add helper for substituting config variables

koesie10:
This adds a helper for substituting config variables, such as the following:
- `.github/codeql/extensions/${name}-${language}` becomes `.github/codeql/extensions/vscode-codeql-java`
- `${owner}/${name}-${language}` becomes `github/vscode-codeql-java`
- `${workspaceFolder}${pathSeparator}.github/workflows/codeql-analysis.yml` becomes `/home/your-username/your-project/.github/workflows/codeql-analysis.yml`

This is based on [the implementation in VS Code](https://github.com/microsoft/vscode/blob/edfd5b8ba54d50f3f5c2ebee877af088803def88/src/vs/base/common/labels.ts#L316C1-L400) that is used for [the `window.title` setting](https://code.visualstudio.com/docs/editor/variables-reference#_is-variable-substitution-supported-in-user-and-workspace-settings). Some of the unit tests are also based on [the VS Code unit tests](https://github.com/microsoft/vscode/blob/edfd5b8ba54d50f3f5c2ebee877af088803def88/src/vs/base/test/common/labels.test.ts#L111-L145).

This helper function will be used in follow-up PRs that introduce new settings for model packs.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/common/config-template.ts
+++ b/extensions/ql-vscode/src/common/config-template.ts
@@ -1,0 +1,57 @@
+// Based on https://github.com/microsoft/vscode/blob/edfd5b8ba54d50f3f5c2ebee877af088803def88/src/vs/base/common/labels.ts#L316C1-L400
+
+/**
+ * Helper to insert values for specific template variables into the string. E.g. "this ${is} a ${template}" can be
+ * passed to this function together with an object that maps "is" and "template" to strings to have them replaced.
+ *
+ * @param template string to which template is applied
+ * @param values the values of the templates to use
+ */
+export function substituteConfigVariables(
+  template: string,
+  values: {
+    [key: string]: string | undefined | null;
+  },
+): string {
+  const segments: string[] = [];
+
+  let inVariable = false;
+  let currentValue = "";
+  for (const char of template) {
+    // Beginning of variable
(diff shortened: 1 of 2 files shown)


robertbrignull (approved):
Implementation LGTM. Thanks for including the link to the vscode repo.

Side question: as well as the new model packs uses, do you plan to replace the interpolation in `HistoryItemLabelProvider` with this implementation?

koesie10:
> Side question: as well as the new model packs uses, do you plan to replace the interpolation in `HistoryItemLabelProvider` with this implementation?

I don't think we can without breaking existing settings since that uses a format like `%q on %d - %s %r [%t]` which is incompatible with this "standard" VS Code format. We might be able to deprecate the old format and switch to this new format, but I don't think that's something we're looking into right now.

charisk:
> > Side question: as well as the new model packs uses, do you plan to replace the interpolation in `HistoryItemLabelProvider` with this implementation?
> 
> I don't think we can without breaking existing settings since that uses a format like `%q on %d - %s %r [%t]` which is incompatible with this "standard" VS Code format. We might be able to deprecate the old format and switch to this new format, but I don't think that's something we're looking into right now.

Might be worth an issue to track this as potential tech/product debt.

PR #3542, opened 2024-04-05
Remove mocking of modeling events

charisk:
There isn't a good reason to mock modeling events and it was making testing a bit awkward, so I've removed mocking and started using real `MockModelingEvents` instances.

## Checklist
N/A:


norascheuch (approved):
Nice! ✨

extensions/ql-vscode/test/__mocks__/appMock.ts
```diff
@@ -54,19 +54,21 @@ export function createMockApp({
 
 class MockAppEventEmitter<T> implements AppEventEmitter<T> {
   public event: AppEvent<T>;
+  private listeners: Array<(event: T) => void> = [];
 
   constructor() {
-    this.event = () => {
+    this.event = (listener) => {
+      this.listeners.push(listener);
       return new MockAppEvent();
```
koesie10, line 62: Should we also try to ensure that the `MockAppEvent.dispose` method correctly removes the listener? It could be quite confusing to receive events when you have already disposed the event listener. I think we can also remove the `fire` method from the `MockAppEvent` since that's not used or part of the interface.

We might even consider removing the `MockAppEvent` class and simply doing this:

```suggestion
      return {
        dispose: () => {
          this.listeners = this.listeners.filter((l) => l !== listener);
        },
      };
```

charisk: That's a good shout, let me tidy that up.

koesie10 (approved):

PR #3552, opened 2024-04-11
Remove `codeQL.model.showTypeModels` config

koesie10:
This removes the `codeQL.model.showTypeModels` setting. We will now always show type models in the model editor. This also means that the "auto-generation mode" needs to change to `models` to ensure that the type models are actually shown to the user. This is done for Ruby, which is currently the only language to show type models.

This does not need a changelog entry because the Ruby model editor is only available in canary mode.

## Checklist


charisk (approved):
I can't claim I'm following 100% but the code changes LGTM.

PR #3569, opened 2024-04-16
Detect when the github-enterprise.uri setting is set

robertbrignull:
This adds code that reads the `github-enterprise.uri` setting and uses that to decide if we're on github.com, a GHES instance, or GHEC-DR (which as far as I know is the public name so I'll try to refer to it that way in this repo).

The logic for parsing the URI is taken from https://github.com/microsoft/vscode-pull-request-github/blob/a116c509bf98f066f302f25f1716d884df3322db/src/github/utils.ts#L1275 so we keep vaguely consistent with that extension. I'm not totally sure about the logic of ignoring an invalid URI but that's how they've coded it (or actually their code might throw an error).

Then we also add another setting to act as a feature flag to allow using MRVA on GHEC-DR. Right now this is the only thing that these settings control, and they're only checked when trying to trigger a new MRVA. I think that only checking when triggering a new MRVA is probably the best compromise, rather than trying to hide all existence of MRVA in the extension. I think it fits nicely with our decision around requiring a different workspace for each GitHub environment.

Eventually we'll remove the feature flag so MRVA is always enabled on GHEC-DR, though we'll have to keep some of this code to check for other values of the `github-enterprise.uri` setting like GHES URIs.

I do think this turned out more complicated that I expected it to be. I'm not sure if that's because I've made it more complicated than it needs to be or if this is just how it has to be.

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/config.ts
+++ b/extensions/ql-vscode/src/config.ts
@@ -4,7 +4,7 @@ import type {
   ConfigurationScope,
   Event,
 } from "vscode";
-import { ConfigurationTarget, EventEmitter, workspace } from "vscode";
+import { ConfigurationTarget, EventEmitter, workspace, Uri } from "vscode";
 import type { DistributionManager } from "./codeql-cli/distribution";
 import { extLogger } from "./common/logging/vscode";
 import { ONE_DAY_IN_MS } from "./common/time";
@@ -14,6 +14,7 @@ import {
   SortKey,
 } from "./variant-analysis/shared/variant-analysis-filter-sort";
 import { substituteConfigVariables } from "./common/config-template";
+import { getErrorMessage } from "./common/helpers-pure";
 
 export const ALL_SETTINGS: Setting[] = [];
 
@@ -68,6 +69,52 @@ export const VSCODE_SAVE_BEFORE_START_SETTING = new Setting(
   "saveBeforeStart",
   VSCODE_DEBUG_SETTING,
 );
+
+const VSCODE_GITHUB_ENTERPRISE_SETTING = new Setting(
+  "github-enterprise",
+  undefined,
+);
+export const VSCODE_GITHUB_ENTERPRISE_URI_SETTING = new Setting(
+  "uri",
+  VSCODE_GITHUB_ENTERPRISE_SETTING,
+);
+
+/**
+ * Get the value of the `github-enterprise.uri` setting, parsed as a URI.
+ * If the value is not set or cannot be parsed, return `undefined`.
+ */
+export function getEnterpriseUri(): Uri | undefined {
+  const config = VSCODE_GITHUB_ENTERPRISE_URI_SETTING.getValue<string>();
+  if (config) {
+    try {
+      let uri = Uri.parse(config, true);
+      if (uri.scheme === "http") {
+        uri = uri.with({ scheme: "https" });
+      }
+      return uri;
+    } catch (e) {
+      void extLogger.log(
+        `Failed to parse the GitHub Enterprise URI: ${getErrorMessage(e)}`,
+      );
+    }
+  }
+  return undefined;
+}
+
+/**
+ * Is the GitHub Enterprise URI set?
+ */
+export function hasEnterpriseUri(): boolean {
+  return getEnterpriseUri() !== undefined;
+}
+
+/**
+ * Is the GitHub Enterprise URI set to something that looks like GHEC-DR?
+ */
+export function hasGhecDrUri(): boolean {
+  const uri = getEnterpriseUri();
+  return uri !== undefined && uri.authority.match(/.*\.ghe\.com/) !== null;
+}
 
 const ROOT_SETTING = new Setting("codeQL");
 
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/test/vscode-tests/minimal-workspace/variant-analysis/ghec-dr.test.ts
```diff
@@ -0,0 +1,44 @@
+import { ConfigurationTarget } from "vscode";
```
robertbrignull, line 1: I added these tests to `minimal-workspace` because that's where `config.test.ts` is located. I'm not sure if these config tests need a workspace, but as far as I know there's not a huge performance difference between `no-workspace` and `minimal-workspace` so it likely doesn't matter too much.

extensions/ql-vscode/src/config.ts
```diff
@@ -576,6 +623,11 @@ export function getVariantAnalysisDefaultResultsSort(): SortKey {
  */
 const ACTION_BRANCH = new Setting("actionBranch", VARIANT_ANALYSIS_SETTING);
 
+export const VARIANT_ANALYSIS_ENABLE_GHEC_DR = new Setting(
+  "enable_ghec_dr",
```
koesie10, line 627: I think it would be better for this name to be consistent with other settings:

```suggestion
  "enableGhecDr",
```

extensions/ql-vscode/src/config.ts
```diff
@@ -69,6 +70,52 @@ export const VSCODE_SAVE_BEFORE_START_SETTING = new Setting(
...
+export function hasEnterpriseUri(): boolean {
+  return getEnterpriseUri() !== undefined;
+}
+
+/**
+ * Is the GitHub Enterprise URI set to something that looks like GHEC-DR?
+ */
+export function hasGhecDrUri(): boolean {
+  const uri = getEnterpriseUri();
+  return uri !== undefined && uri.authority.match(/.*\.ghe\.com/) !== null;
```
koesie10, line 116: Since we're only matching on authority here, could this work as well?

```suggestion
  return uri !== undefined && uri.authority.toLowerCase().endsWith(".ghe.com");
```

extensions/ql-vscode/src/variant-analysis/ghec-dr.ts
```diff
@@ -0,0 +1,18 @@
...
+  VARIANT_ANALYSIS_ENABLE_GHEC_DR,
+  hasEnterpriseUri,
+  hasGhecDrUri,
+} from "../config";
+
+/**
+ * Determines whether MRVA should be enabled or not for the current GitHub host.
+ * This is based on the `github-enterprise.uri` setting.
+ */
+export function isVariantAnalysisEnabledForGitHubHost(): boolean {
```
koesie10, line 11: Would it make sense to make this part of the `VariantAnalysisConfig`?

extensions/ql-vscode/test/vscode-tests/minimal-workspace/variant-analysis/ghec-dr.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  it("returns false when GHEC-DR URI is set but variant analysis feature flag is not set", async () => {
+    await VSCODE_GITHUB_ENTERPRISE_URI_SETTING.updateValue(
+      "https://github.example.com",
+      ConfigurationTarget.Global,
+    );
+    expect(isVariantAnalysisEnabledForGitHubHost()).toBe(false);
+  });
+
+  it("returns true when GHEC-DR URI is set and variant analysis feature flag is set", async () => {
```
koesie10, line 33: Should we also have a test for a GHES-DC URI but without the feature flag set?

koesie10 (commented):
Code looks good to me, I only have a few minor suggestions

extensions/ql-vscode/src/config.ts
```diff
@@ -69,6 +70,52 @@
...
```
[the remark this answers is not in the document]
robertbrignull: This is correct. We should have put a `$` to ensure it matches the end of the string. But I think I'll go with @koesie10's suggestion of using `endsWith` anyway so that'll also avoid this error.

extensions/ql-vscode/src/variant-analysis/ghec-dr.ts
robertbrignull: I think right now it wouldn't make any difference because that config is only used to tell `DbTreeDataProvider` when to update, and I believe that the GHEC-DR setting wouldn't affect that view.
Since `VariantAnalysisConfig` only contains a couple of fields and by no means all of the variant analysis config fields, I'm tempted to also not add this one right now. 🤷🏼

extensions/ql-vscode/test/vscode-tests/minimal-workspace/variant-analysis/ghec-dr.test.ts
robertbrignull: I think that's covered by the "returns false when GHEC-DR URI is set but variant analysis feature flag is not set" test just above this one. Am I understanding correctly?

koesie10: Ah, I should have left a comment on that test. I was looking at the contents of the test rather than the names, but the test above this one uses `https://github.example.com` rather than a GHEC-DR URI.

robertbrignull: Ah good spot, the URI was incorrect in that test. I'll fix that. I'll also add another test to cover the case of a GHES URI and feature flag not being set, so ever case is covered.

PR #3610, opened 2024-05-14
Handle exit and error events for query server child processes

angelapwen:
Previously, we were using the [`close` event handler](https://nodejs.org/api/child_process.html#event-close) for the query server child process(es). This did not always properly handle the case where the child query server process crashed, because `close` is only emitted after all stdio streams of the child process are closed (see back-linked issue).

This PR handles the `exit` and `error` processes instead: `exit` is emitted even if child process stdio streams are still open. Note that the documentation indicates that `exit` may still fire after `error`, so this change includes some simple logic to make sure that we only restart the query server once if both fire. 

## Checklist

### DIFF
--- a/extensions/ql-vscode/src/query-server/query-server-client.ts
+++ b/extensions/ql-vscode/src/query-server/query-server-client.ts
@@ -240,8 +240,32 @@ export class QueryServerClient extends DisposableObject {
     this.nextCallback = 0;
     this.nextProgress = 0;
     this.progressCallbacks = {};
-    child.on("close", () => {
-      this.restartQueryServerOnFailure();
+
+    // 'exit' may or may not fire after 'error' event, so we want to guard against restarting the
+    // query server twice if both events fire.
+    let wasExitOrErrorHandled = false;
+    child.on("error", (err: Error) => {
+      if (!wasExitOrErrorHandled) {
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/src/query-server/query-server-client.ts
```diff
@@ -240,8 +240,32 @@ export class QueryServerClient extends DisposableObject {
...
+    // query server twice if both events fire.
+    let wasExitOrErrorHandled = false;
+    child.on("error", (err: Error) => {
+      if (!wasExitOrErrorHandled) {
+        void this.logger.log(`Query server terminated with error: ${err}.`);
+        this.restartQueryServerOnFailure();
+        wasExitOrErrorHandled = true;
+      }
+    });
+    child.on("exit", (code: number, signal: string) => {
```
aeisenberg, line 254: Just making sure: can `close` ever fire without `exit` also firing?

angelapwen: Yes, but only in the `error` case:

> The 'close' event will always emit after ['exit'](https://nodejs.org/api/child_process.html#event-exit) was already emitted, or ['error'](https://nodejs.org/api/child_process.html#event-error) if the child failed to spawn.

aeisenberg (approved):
Nice!

PR #3643, opened 2024-06-18
Re-enable windows 8.3 paths in tests

robertbrignull:
This PR reverts https://github.com/github/vscode-codeql/pull/3637 and also includes configuration that was removed from the windows actions runner images in https://github.com/actions/runner-images/commit/5ecfd27e8fd0a1b4e8c88072772f9bbab9c82b83

This PR supersedes https://github.com/github/vscode-codeql/pull/3638.

Sadly we have to keep the support for 8.3 filenames in the extension as users may still encounter them, as we believe the feature is still enabled by default on Windows 11 for the `C:` drive, although the actions team determined it was not enabled for the Azure Marketplace Windows 2019 image.

@dbartol

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -144,6 +144,19 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
     steps:
+      # Enable 8.3 filename creation. This is not required to run the extension but it is required for the unit tests to pass.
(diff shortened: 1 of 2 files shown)


aeisenberg (approved):
LGTM. Thanks for digging into this and fixing the tests.

PR #3653, opened 2024-07-02
Use rich type columns and endpoint kinds for Python model editor

koesie10:
This makes changes to the Python model editor parsing code to incorporate the query changes from https://github.com/github/codeql/pull/15655. These changes consist of:
- Adding support for reading rich type columns (as implemented by https://github.com/github/codeql/pull/16490): This makes it possible to replace the columns `["requests", "Member[adapters].Member[HTTPAdapter].Instance.Member[add_headers]"]` by `["requests.HTTPAdapter", "Member[add_headers]"]`.
- Write rich type columns instead of complex paths. We still retain support for reading non-rich type columns.
- Use the `kind` column that has been added in https://github.com/github/codeql/pull/15655 (the last column returned in the query). This makes it possible to properly handle class methods, which we would otherwise need to detect by the name of the method argument rather than just receiving this information from the query.

For more information about the format of Python MaD, see [the documentation](https://github.com/github/codeql/blob/main/docs/codeql/codeql-language-guides/customizing-library-models-for-python.rst).

### DIFF
--- a/extensions/ql-vscode/src/model-editor/bqrs.ts
+++ b/extensions/ql-vscode/src/model-editor/bqrs.ts
@@ -33,6 +33,7 @@ export function decodeBqrsToMethods(
     let libraryVersion: string | undefined;
     let type: ModeledMethodType;
     let classification: CallClassification;
+    let endpointKindColumn: string | BqrsEntityValue | undefined;
     let endpointType: EndpointType | undefined = undefined;
 
     if (mode === Mode.Application) {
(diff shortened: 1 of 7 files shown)


robertbrignull (approved):
I've tried running this and it worked fine for me, but it's the first time I've used the model editor for python so I'm not very confident I tested it thoroughly.

Looking through the code I can't see anything wrong, but there's a lot going on so it's hard to tell.

LGTM, unless you want to get someone from the python team to also try it out

PR #3654, opened 2024-07-03
Enable type models for Python in the model editor

koesie10:
This enables type models for Python in the model editor. There is no query for generating type models or for complex access path suggestions, so a simple text box will be shown to enter the path and related type.

- Based on https://github.com/github/vscode-codeql/pull/3653


robertbrignull (approved):
I've tried running this and it worked fine for me. I'm not very familiar with the code but nothing looks wrong to me.

LGTM unless you want to get someone from the python team to try it out as well

norascheuch (approved):

PR #3658, opened 2024-07-05
Allow compatible CLIs to generate summary symbols file

nickrolfe:
This should be much faster than generating it in the extension.

No change-note since this should be a seamless change.

### DIFF
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -1211,10 +1211,15 @@ export class CodeQLCliServer implements Disposable {
     outputPath: string,
     endSummaryPath: string,
   ): Promise<string> {
+    const supportsGenerateSummarySymbolMap =
+      await this.cliConstraints.supportsGenerateSummarySymbolMap();
     const subcommandArgs = [
       "--format=text",
       `--end-summary=${endSummaryPath}`,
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1211,13 +1211,19 @@ export class CodeQLCliServer implements Disposable {
...
-      inputPath,
-      outputPath,
-    ];
+    ]
+      .concat(
+        supportsGenerateSummarySymbolMap
+          ? ["--summary-symbol-map", "--minify-output"]
+          : [],
+      )
+      .concat([inputPath, outputPath]);
```
aeisenberg, line 1226: Is this easier to read?

```suggestion
    ...(supportsGenerateSummarySymbolMap
          ? ["--summary-symbol-map", "--minify-output"] 
          : []),
    inputPath, 
    outputPath
  ]
```

aeisenberg (approved):
LGTM (you can take or leave my suggestion)

extensions/ql-vscode/src/codeql-cli/cli.ts
nickrolfe: Thanks! I suspected there was a better way of doing it, but I just wasn't aware of the `...` operator.

Issue #3659, opened 2024-07-10
Feature request: Display length of shortest path in UI


p-:
**Is your feature request related to a problem? Please describe.**

When using MRVA with a query (custom or default) that yields a lot of results it is often hard to figure out which of the results are the most promising ones. Only after clicking on "Show paths" one learns how long the paths are and likely starts looking at results with a lower path count.

![Image](https://github.com/user-attachments/assets/73500d6a-08f7-4192-bcf0-272d938d0096)

See also: [Feature request: allow to limit results by path length in UI](https://github.com/github/vscode-codeql/issues/3660)

**Describe the solution you'd like**

It would be nice if the shortest paths for all the results would be displayed in the overview. (For MRVA and maybe also for local queries).

I've hacked this together in my [experimental branch](https://github.com/p-/vscode-codeql/tree/p--max-path-steps) and it's already helpful (but not so beautiful 😉 ) - see commit https://github.com/p-/vscode-codeql/commit/20617d2ab821ad71bb766fa73fec016c5287ff63.

![Image](https://github.com/user-attachments/assets/4c9c43f6-3689-459a-85bc-6ff48a8d0918)

**Describe alternatives you've considered**

An alternative or a nice addition would be if they results were ordered with regards to their shortest path length (shortest length on top).

**Additional context**

See also: [Feature request: allow to limit results by path length in UI](https://github.com/github/vscode-codeql/issues/3660)

jf205:
Hi @p-! Thanks for creating this issue. This change looks useful and is a nice simple UI change. The copy looks fine to me. We would be happy for you to contribute this change to the extension as a pull request and someone on the team will review it. Please let us know when the PR is ready.

p-:
Hi @jf205 
Thanks!

I've created a PR: https://github.com/github/vscode-codeql/pull/3671

p-:
I've prototyped this for the local queries as well (using a badge due to size constraints).
![Image](https://github.com/user-attachments/assets/19b7a913-4f8a-4910-8489-b6119949f78a)


@jf205 Should I add the change for local queries to the same PR? (it's only for MRVA at the moment)

/cc @owen-mc

owen-mc:
[...]

PR #3671, opened 2024-07-24
Display length of shortest path in MRVA UI

p-:
Hey 👋 

This PR implements my feature request from #3659.
I'm totally open for changes of course (e.g. visualize it differently etc. - e.g. use a callout displaying only the number with a tooltip for more information)

Note:
* Due to the type definitions I assumed that the codeflows are available in the instances where this UI are displayed. (if not more checks would be warranted)
* I've added a `data-testid` to assert the textcontent of the label on one line.

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Update variant analysis view to display the length of the shortest path for path queries. #3659
+
 ## 1.13.1 - 29 May 2024
 
 - Fix a bug when re-importing test databases that erroneously showed old source code. [#3616](https://github.com/github/vscode-codeql/pull/3616)
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-- Update variant analysis view to display the length of the shortest path for path queries. #3659
+- Update variant analysis view to display the length of the shortest path for path queries. [#3671](https://github.com/github/vscode-codeql/pull/3671)
 
 ## 1.13.1 - 29 May 2024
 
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Update variant analysis view to display the length of the shortest path for path queries. #3659
```
charisk, line 5: Minor suggestion to match the style of the rest of the changelog:

```suggestion
- Update variant analysis view to display the length of the shortest path for path queries. [#3659](https://github.com/github/vscode-codeql/pull/3659)
```

extensions/ql-vscode/src/view/common/CodePaths/CodePaths.tsx
```diff
@@ -40,6 +52,9 @@ export const CodePaths = ({
   return (
     <>
       <ShowPathsLink onClick={onShowPathsClick}>Show paths</ShowPathsLink>
+      <Label data-testid="shortest-path-length">
```
charisk, line 55: Minor suggestion. I don't think we've generally used test ids in the codebase so it would be good to avoid it here. In the test, can we just look for a `span` with the matching text instead?

charisk (approved):
LGTM in general. I've left a couple of small suggestions. 

Thank you for contributing!

extensions/ql-vscode/CHANGELOG.md
p-: Thanks! I've now also changed the ID to refer to the PR as the other entries in the changelog.
(I originally linked to the issue)

extensions/ql-vscode/src/view/common/CodePaths/CodePaths.tsx
p-: 👍 
Somehow I thought I couldn't make it work without the `test-id` due to the whitespaces 🙈 . - But it works now 😁

charisk (approved):
🚢

PR #3676, opened 2024-07-31
Remove feature flag for Python model editor

koesie10:
This removes the feature flag `codeQL.model.enablePython` config setting and enables the Python model editor for everyone.

### DIFF
--- a/extensions/ql-vscode/src/model-editor/supported-languages.ts
+++ b/extensions/ql-vscode/src/model-editor/supported-languages.ts
@@ -9,19 +9,15 @@ export const SUPPORTED_LANGUAGES: QueryLanguage[] = [
   QueryLanguage.Java,
   QueryLanguage.CSharp,
   QueryLanguage.Ruby,
+  QueryLanguage.Python,
 ];
 
 export function isSupportedLanguage(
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/model-editor/supported-languages.ts
```diff
@@ -9,20 +9,16 @@ export const SUPPORTED_LANGUAGES: QueryLanguage[] = [
   QueryLanguage.Java,
   QueryLanguage.CSharp,
   QueryLanguage.Ruby,
+  QueryLanguage.Python,
 ];
 
 export function isSupportedLanguage(
   language: QueryLanguage,
-  modelConfig: ModelConfig,
+  _modelConfig: ModelConfig,
```
koesie10, line 17: I've left this parameter here for now since it's likely we'll be adding new languages at some point in the future with a feature flag.

charisk (approved):
LGTM!

extensions/ql-vscode/src/model-editor/supported-languages.ts
melorie88manuel: melorie

PR #3681, opened 2024-08-02
Remove support for CodeQL CLI versions older than v2.15.5

charisk:
Removes support for CodeQL CLI versions under v2.15.5.


koesie10 (approved):
LGTM, thanks!

PR #3683, opened 2024-08-06
Fix docker command

dbartol:
Discovered in https://github.com/github/vscode-codeql/actions/runs/10268779662/job/28412615133?pr=3682
Root cause was a breaking change announced [here](https://github.blog/changelog/2024-04-10-github-hosted-runner-images-deprecation-notice-docker-compose-v1/).


charisk (approved):
Thanks!

PR #3700, opened 2024-08-30
Upgrade Octokit packages

koesie10:
This upgrades all Octokit packages to their latest version for which the main change is the switch to ESM. Therefore, we need to transform some more packages in the Jest configuration, but there is only 1 change in the code we need to make.


angelapwen (approved):
Looks good to me! Thanks for improving the readability, that helps a lot 😅

PR #3706, opened 2024-09-05
Upgrade node-fetch to v3

koesie10:
This upgrades the node-fetch package to v3. The [main breaking change](https://github.com/node-fetch/node-fetch/blob/8b3320d2a7c07bce4afc6b2bf6c3bbddda85b01f/docs/v3-UPGRADE-GUIDE.md) is a switch to ESM, so I've added a few packages to the Jest config. This will hopefully stop https://github.com/github/vscode-codeql/pull/3702 from failing when it's rebased on top of this PR.


charisk (approved):
Thanks!

PR #3717, opened 2024-09-13
Upgrade path-to-regexp to 6.3.0

koesie10:
This should close https://github.com/github/vscode-codeql/security/dependabot/69.


angelapwen (approved):
Thanks for resolving that!!

PR #3719, opened 2024-09-16
Remove automodel-related functionality

koesie10:
This removes all automodel/LLM-related functionality. There is quite a lot of code that we can remove, so this PR is quite large. Hopefully it still shouldn't be too hard to review since it's mostly removals. The only "additions" that are shown is moving code outside of an if-statement and some very small other changes.


charisk (approved):
Thanks for cleaning up!

PR #3726, opened 2024-09-23
Avoid uploading eslint sarif for dependabot PR

aeisenberg:
Dependabot does not have `security-events: write` permission.

See https://github.com/github/codeql-action/pull/2485 for the original instance of this problem.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -103,7 +103,7 @@ jobs:
           npm run lint-ci
 
       - name: Upload ESLint results to Code Scanning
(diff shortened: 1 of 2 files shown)


charisk (approved):

aeisenberg:
Hmmm...unsure why the CI jobs are failing. Looks like it is failing on `main` as well. https://github.com/github/vscode-codeql/actions/runs/10994181318/job/30522243194#step:7:95

aeisenberg:
Bumping the timeout worked.

PR #3728, opened 2024-09-25
Remove support for CodeQL CLI versions older than v2.16.6

norascheuch:
Remove support for CodeQL CLI versions older than v2.16.6


extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1909,38 +1909,12 @@ function shouldDebugCliServer() {
...
-
-  public static CLI_VERSION_WITHOUT_MRVA_EXTENSIBLE_PREDICATE_HACK = new SemVer(
-    "2.16.1",
-  );
-
-  /**
-   * CLI version where there is support for multiple queries on the pack create command.
-   */
-  public static CLI_VERSION_WITH_MULTI_QUERY_PACK_CREATE = new SemVer("2.16.1");
+  public static OLDEST_SUPPORTED_CLI_VERSION = new SemVer("2.16.6");
```
norascheuch, line 1912: I'm a bit confused about the version number syntax. We use version numbers with `v` in the `supported_cli_versions.json`, but here we have historically used no `v`. In the code I removed we used `SemVer("2.16.1")`, so without the `v`. Which one is correct?

robertbrignull: https://www.npmjs.com/package/semver says "A leading "=" or "v" character is stripped off and ignored." so either is fine. Personally I'd go with no "v" here.

norascheuch: Thanks!

robertbrignull: To explain more...

From a behaviour point of view I don't think it makes any difference. From a code readability point of view it makes sense to remove unnecessary boilerplate. From a consistency point of view I suppose you might want them to match the json file, but I think any human who reads them would easily correlate the numbers and not get confused.

robertbrignull (approved):

extensions/ql-vscode/test/vscode-tests/cli-integration/variant-analysis/variant-analysis-manager.test.ts
```diff
@@ -362,7 +354,7 @@ describe("Variant Analysis Manager", () => {
         qlPackRootPath,
         qlPackFilePath,
         expectedPackName: "codeql/java-queries",
-        filesThatExist: [queryToRun, ...extraQueries],
+        filesThatExist: [queryToRun, ...["Telemetry/ExtractorInformation.ql"]],
```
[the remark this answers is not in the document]
norascheuch: 🤦‍♀️ thank's for spotting that!

mammadovaziz30-art (commented):
G

PR #3762, opened 2024-10-16
Store state of CodeQL distribution on filesystem instead of in `globalState`

koesie10:
The `globalState` is shared between all VS Code instances, regardless of whether they are local or remote (e.g. Codespaces, SSH). Since the distribution (i.e. the CodeQL CLI) is stored on the filesystem, this does not always match the `globalState` (e.g. when switching between local and Codespaces).

This fixes that problem by also storing the distribution state on the local filesystem in a `distribution.json` file. This file contains the same information as the `globalState` did, but is local to each filesystem. For example, this is what it contains on my machine:

```json
{
    "folderIndex": 244,
    "release": {
        "assets": [
            {
                "id": 196506420,
                "name": "codeql-linux64.zip",
                "size": 725141406
            },
            {
                "id": 196506419,
                "name": "codeql-linux64.zip.checksum.txt",
                "size": 85
            },
            {
                "id": 196506418,
                "name": "codeql-osx64.zip",
                "size": 944929533
            },
            {
                "id": 196506422,
                "name": "codeql-osx64.zip.checksum.txt",
                "size": 83
            },
            {
                "id": 196506421,
                "name": "codeql-win64.zip",
                "size": 345891346
            },
            {
                "id": 196506424,
                "name": "codeql-win64.zip.checksum.txt",
                "size": 83
            },
            {
                "id": 196506416,
                "name": "codeql.zip",
                "size": 1745315855
            }
        ],
        "createdAt": "2024-10-04T20:51:14Z",
        "id": 178107771,
        "name": "v2.19.1"
    }
}
```

The migration to this format is quite trivial: if the `distribution.json` file doesn't exist, take the values from `globalState` and create it. We can then always use the `distribution.json` file.

In addition, this adds a filesystem-based lock when updating the distribution. This ensures that there are no race conditions between multiple VS Code instances when they all try to update the distribution.


robertbrignull (commented):
I want to do a little bit of manual testing but as an approach it looks reasonable to me, and I like that we're using a library for the lock file instead of writing our own handler for that.

robertbrignull (approved):
👍🏼

PR #3763, opened 2024-10-16
Clean up old distributions

koesie10:
This adds a background process for cleaning up old distributions (i.e. old extension-managed CodeQL versions). These could be left behind as a result of a bug (which is fixed by https://github.com/github/vscode-codeql/pull/3762).

On start-up (after extension activation), it will asynchronously list all directories in the `globalStorage` directory. It will then select all directories which match the pattern `distribution<n>` where `<n>` is a number. Then, it will sequentially remove all folders which have a lower folder index (i.e. the `<n>`) than the currently installed CodeQL CLI. It will wait 10 seconds between each directory to avoid overloading the system. The `remove` call itself should also be asynchronous, so it should not block the extension thread.

I've tested this locally by copying my current distribution a couple of times to some other distribution folders and it does remove the correct directories. I've also added some tests for this.


extensions/ql-vscode/src/codeql-cli/distribution/cleaner.ts
```diff
@@ -0,0 +1,113 @@
...
+      return;
+    }
+
+    void this.logger.log(
+      `Cleaning up ${cleanableDirectories.length} old versions of the CLI.`,
+    );
+
+    for (const cleanableDirectory of cleanableDirectories) {
+      // Wait 60 seconds between each cleanup to avoid overloading the system (even though the remove call should be async)
+      await sleep(10_000);
```
aeisenberg, line 92: This is 10 seconds.

aeisenberg (commented):
This looks generally good. My only question is how does this behave if we have multiple workspaces open each trying to delete the same set of distributions? I'm guessing one of them will throw an error, but continue deleting.

Could you make sure that an error message is not sent to the log if a distribution has already been deleted when the cleaner tries to clean it?

aeisenberg:
We also have a similar process to remove orphaned databases.

koesie10:
> This looks generally good. My only question is how does this behave if we have multiple workspaces open each trying to delete the same set of distributions? I'm guessing one of them will throw an error, but continue deleting.
> 
> Could you make sure that an error message is not sent to the log if a distribution has already been deleted when the cleaner tries to clean it?

Thanks, that's a good point. I've changed the process to be less likely to all clean-up the same directory by shuffling the array, but I've also added a check for `ENOENT` which should ensure that when the directory doesn't exist, we don't log the error.

aeisenberg (approved):
Nice!

robertbrignull (approved):

PR #3771, opened 2024-10-22
Implement streaming for compare view

koesie10:
This implements streaming for the compare view to avoid a crash when the total JSON.stringified size of the comparison is larger than 1GB (i.e. the Node.js string limit). It does this as follows:
- All results are chunked into 1,000 results per message. This means we'll send at most 1,000 `to` **and** at most 1,000 `from` results in 1 message (i.e. 2,000 actual results).
- We'll send a "setup" message to tell the webview that we're starting a streaming result. This also contains a chunk.
- We'll then send as many "add results" messages as necessary to send all results.
- We'll send a "complete" message to tell the webview that we're done.

The 1,000 results per message is just a safe estimate where we should never run into the string limit. However, I can't really test this.

### DIFF
--- a/extensions/ql-vscode/src/compare/compare-view.ts
+++ b/extensions/ql-vscode/src/compare/compare-view.ts
@@ -183,13 +183,92 @@ export class CompareView extends AbstractWebview<
         message = getErrorMessage(e);
       }
 
+      await this.streamResults(result, currentResultSetDisplayName, message);
+    }
+  }
+
+  private async streamResults(
+    result: QueryCompareResult | undefined,
+    currentResultSetName: string,
+    message: string | undefined,
+  ) {
+    // Since there is a string limit of 1GB in Node.js, the comparison is send as a JSON.stringified string to the webview
+    // and some comparisons may be larger than that, we sometimes need to stream results. This uses a heuristic of 2,000 results
+    // to determine if we should stream results.
+
+    if (!this.shouldStreamResults(result)) {
       await this.postMessage({
         t: "setComparisons",
         result,
-        currentResultSetName: currentResultSetDisplayName,
+        currentResultSetName,
         message,
       });
+      return;
     }
+
+    // Streaming itself is implemented like this:
+    // - 1 setup message which contains the first 1,000 results
+    // - n "add results" messages which contain 1,000 results each
+    // - 1 complete message which just tells the webview that we're done
+
+    await this.postMessage({
+      t: "streamingComparisonSetup",
+      result: this.chunkResults(result, 0, 1000),
+      currentResultSetName,
+      message,
+    });
+
+    const { from, to } = result;
+
+    const maxResults = Math.max(from.length, to.length);
+    for (let i = 1000; i < maxResults; i += 1000) {
+      const chunk = this.chunkResults(result, i, i + 1000);
+
+      await this.postMessage({
+        t: "streamingComparisonAddResults",
+        result: chunk,
+      });
+    }
+
+    await this.postMessage({
+      t: "streamingComparisonComplete",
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/compare/compare-view.ts
```diff
@@ -183,13 +184,97 @@ export class CompareView extends AbstractWebview<
...
+      };
+    }
+
+    if (result.kind === "interpreted") {
+      return {
+        ...result,
+        from: result.from.slice(start, end),
+        to: result.to.slice(start, end),
+      };
+    }
```
aeisenberg, line 275: It looks like the bodies of these if statements are the same. Should you combine them?

extensions/ql-vscode/src/view/compare/Compare.tsx
```diff
@@ -53,6 +61,87 @@ export function Compare(_: Record<string, never>): React.JSX.Element {
...
+              break;
+            }
+
+            let result: QueryCompareResult;
+            switch (prev.result.kind) {
+              case "raw":
+                if (msg.result.kind !== "raw") {
+                  throw new Error(
+                    "Streaming comparison: expected raw results, got interpreted results",
+                  );
```
aeisenberg, line 90: Any danger of getting here if you switch in the middle of rendering a very large comparison result?

extensions/ql-vscode/src/view/compare/Compare.tsx
```diff
@@ -53,6 +61,87 @@ export function Compare(_: Record<string, never>): React.JSX.Element {
...
+                  throw new Error(
+                    "Streaming comparison: expected interpreted results, got raw results",
+                  );
+                }
+
+                result = {
+                  ...prev.result,
+                  from: [...prev.result.from, ...msg.result.from],
+                  to: [...prev.result.to, ...msg.result.to],
+                };
```
aeisenberg, line 110: It looks like this assignment is the same as the one above. Can you merge the `case`s?

aeisenberg (commented):
A handful of simplifications, and a question.

Looks good, but I have not ran this myself.

extensions/ql-vscode/src/compare/compare-view.ts
koesie10: Unfortunately that's not possible in this case because TypeScript can't determine that automatically from these overlapping types: `Returned expression type {kind: "raw" | "interpreted", from: Row[] | Result[], to: Row[] | Result[]} is not assignable to type QueryCompareResult `.

extensions/ql-vscode/src/view/compare/Compare.tsx
koesie10: This is split out for the same reason as above: we get a TypeScript error if we don't have this.

extensions/ql-vscode/src/view/compare/Compare.tsx
koesie10: There shouldn't be: https://github.com/github/vscode-codeql/pull/3771/commits/510a269dde893d3fd7eb4dafdccecf5532071d97 adds a check for that race condition by adding an `id` to every message so that if a different setup message is received, it will ignore all future messages from the previous stream.

extensions/ql-vscode/src/compare/compare-view.ts
aeisenberg: Unfortunate.

aeisenberg (approved):
Given the limitations of the type system, I think this is fine. I'm sure there's some type wizardry we can do to make this cleaner, but it's really not worth the effort.

PR #3772, opened 2024-10-23
Only compare source and sink in SARIF comparison

koesie10:
This changes the SARIF comparison to only include the source and the sink of a path as an input into the comparison. This will result in a more useful comparison by filtering out results that are the same except for their path.

The results of a comparison on this branch:
Screenshot 2024-10-23 at 13 11 02

The same comparison on `main`:

Screenshot 2024-10-23 at 13 15 06

### DIFF
--- a/extensions/ql-vscode/src/compare/sarif-diff.ts
+++ b/extensions/ql-vscode/src/compare/sarif-diff.ts
@@ -85,7 +85,7 @@ function toCanonicalResult(result: Result): Result {
 /**
  * Compare the alerts of two queries. Use deep equality to determine if
  * results have been added or removed across two invocations of a query.
- * It first canonicalizes the results by to ensure that when small changes
+ * It first canonicalizes the results to ensure that when small changes
  * to the query are made, the results are still considered the same. This
  * includes the removal of all paths except for the source and sink.
  *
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/compare/sarif-diff.ts
```diff
@@ -79,11 +85,9 @@ function toCanonicalResult(result: Result): Result {
 /**
  * Compare the alerts of two queries. Use deep equality to determine if
  * results have been added or removed across two invocations of a query.
- *
- * Assumptions:
- *
- * 1. Queries have the same sort order
- * 2. Results are not changed or re-ordered, they are only added or removed
+ * It first canonicalizes the results by to ensure that when small changes
```
aeisenberg, line 88:
```suggestion
 * It first canonicalizes the results to ensure that when small changes
```

aeisenberg (approved):
Nice. Typo in a comment, and good to merge after that.

charisk:
Did we need to update the CHANGELOG for this at all?

koesie10:
> Did we need to update the CHANGELOG for this at all?

Good call, this will update the changelog: https://github.com/github/vscode-codeql/pull/3780

PR #3774, opened 2024-10-23
Upgrade @faker-js/faker to v9

koesie10:
This just upgrades Faker to v9 since this version was released more than a month ago and we haven't received a Dependabot update for it. [The upgrade guide](https://fakerjs.dev/guide/upgrading.html) doesn't seem to contain anything that is relevant for us, so it's a simple bump of the dependency.

### DIFF
--- a/extensions/ql-vscode/package-lock.json
+++ b/extensions/ql-vscode/package-lock.json
@@ -51,7 +51,7 @@
         "@babel/preset-env": "^7.24.4",
         "@babel/preset-react": "^7.24.1",
         "@babel/preset-typescript": "^7.21.4",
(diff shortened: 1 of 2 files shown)


charisk (approved):
Any idea why dependabot didn't kick in?

koesie10:
> Any idea why dependabot didn't kick in?

Dependabot only opens 5 PRs by default, so it might not be able to keep up with the dependencies that are updated. We could increase the [`open-pull-requests-limit`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit), but that will result in a lot more PRs being opened.

PR #3781, opened 2024-10-29
Fix dependabot vulnerability

aeisenberg:
https://github.com/github/vscode-codeql/security/dependabot/73

Not able to be fixed automatically.

### DIFF
--- a/extensions/ql-vscode/package-lock.json
+++ b/extensions/ql-vscode/package-lock.json
@@ -21,6 +21,7 @@
         "ajv": "^8.11.0",
         "child-process-promise": "^2.2.1",
         "chokidar": "^3.6.0",
+        "cookie": ">=0.7.0",
         "d3": "^7.9.0",
         "d3-graphviz": "^5.0.2",
         "fs-extra": "^11.1.1",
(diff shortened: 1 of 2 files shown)


robertbrignull (approved):

koesie10 (changes requested):
I don't think this actually fixes the vulnerability since both `express` and `@bundled-es-modules/cookie` still depend on a different (vulnerable) version of `cookie`:

- https://github.com/github/vscode-codeql/blob/e303fecf8e3cf4144921fa8123ef2a9e541d7026/extensions/ql-vscode/package-lock.json#L2231-L2238
- https://github.com/github/vscode-codeql/blob/e303fecf8e3cf4144921fa8123ef2a9e541d7026/extensions/ql-vscode/package-lock.json#L12432-L12440

I think we will need to wait for upstream to update before this vulnerability can be fixed. This will just include a different (unused) version of `cookie`, in addition to the vulnerable versions. We are tracking this vulnerability already (see internal linked issue).

aeisenberg:
Thanks. I'll close this since it's not a path worth going down.

PR #3797, opened 2024-11-07
Allow import all database subfolders by selecting a folder

reitowo:
Add a palette command that allow user to select a folder and import all database subfolders.

This helps a lot when dealing with micro services that has dozens of databases and save user's time from repeatly choosing folders one by one.

image

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -841,7 +841,7 @@
       },
       {
         "command": "codeQL.chooseDatabaseFoldersParent",
-        "title": "CodeQL: Choose Parent Folder and import all databases directly contained in it"
+        "title": "CodeQL: Import All Databases Directly Contained in a Parent Folder"
       },
       {
         "command": "codeQL.chooseDatabaseArchive",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -841,7 +841,7 @@
       },
       {
         "command": "codeQL.chooseDatabaseFoldersParent",
-        "title": "CodeQL: Choose Folder to import all databases contained in it"
+        "title": "CodeQL: Choose Parent Folder and import all databases directly contained in it"
       },
       {
         "command": "codeQL.chooseDatabaseArchive",
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -840,7 +840,7 @@
         "title": "CodeQL: Choose Database from Folder"
       },
       {
-        "command": "codeQL.chooseMultipleDatabaseFolder",
+        "command": "codeQL.chooseDatabaseFoldersParent",
         "title": "CodeQL: Choose Folder to import all databases contained in it"
       },
       {
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -840,6 +840,10 @@
         "title": "CodeQL: Choose Database from Folder"
       },
       {
+        "command": "codeQL.chooseMultipleDatabaseFolder",
+        "title": "CodeQL: Choose Folder contains all Database Folders to import"
+      },
+      {
         "command": "codeQL.chooseDatabaseArchive",
         "title": "CodeQL: Choose Database from Archive"
       },
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -841,7 +841,7 @@
       },
       {
         "command": "codeQL.chooseMultipleDatabaseFolder",
-        "title": "CodeQL: Choose Folder contains all Database Folders to import"
+        "title": "CodeQL: Choose Folder to import all databases contained in it"
       },
       {
         "command": "codeQL.chooseDatabaseArchive",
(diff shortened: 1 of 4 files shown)


extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -962,27 +978,67 @@ export class DatabaseUI extends DisposableObject {
...
+          } catch (e) {
+            failures.push(entry[0]);
+          }
+        }
+      }
+
+      if (failures.length) {
+        void showAndLogErrorMessage(
+          this.app.logger,
+          `Failed to import ${failures.length} database(s) (${failures.join(
```
aeisenberg, line 1014: What gets printed here? Is this the full path to the database? If so, this would be too long to show in the popup.

Can you use the `fullMessage` option to contain all of the database path names, and just the number of failures in the `message` argument?

aeisenberg (commented):
Thanks for your contribution. I've given it an initial review, but I haven't tried this out locally yet.

extensions/ql-vscode/src/databases/local-databases-ui.ts
reitowo: Sounds good to me! Didn't know there's that argument.

reitowo:
I seperated this logic out of original import folder function, for better structure. Ad also optimized log output.

extensions/ql-vscode/package.json
```diff
@@ -840,8 +840,8 @@
         "title": "CodeQL: Choose Database from Folder"
       },
       {
-        "command": "codeQL.chooseMultipleDatabaseFolder",
-        "title": "CodeQL: Choose Folder contains all Database Folders to import"
+        "command": "codeQL.chooseDatabaseFoldersParent",
+        "title": "CodeQL: Choose Folder to import all databases contained in it"
```
aeisenberg, line 844: I think this is better:

```suggestion
        "title": "CodeQL: Choose Parent Folder and import all databases directly contained in it"
```

extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -978,68 +977,86 @@ export class DatabaseUI extends DisposableObject {
...
+        }
       }
+    }
 
-      return databases;
+    if (failures.length) {
+      void showAndLogErrorMessage(
+        this.app.logger,
+        `Failed to import ${failures.length} database(s), successfully imported ${databases.length} database(s).`,
+        { fullMessage: `Failed folders to import:\n${failures.join("\n")}` },
```
aeisenberg, line 1044: Include the original message in the `fullMessage`. Indentation makes it easier to read.

```suggestion
        { fullMessage: `Failed to import ${failures.length} database(s), successfully imported ${databases.length} database(s). Failed folders to import:\n  ${failures.join("\n  ")}` },
```

extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -978,68 +977,86 @@ export class DatabaseUI extends DisposableObject {
...
+    const entries = await workspace.fs.readDirectory(uri);
+
+    for (const [index, entry] of entries.entries()) {
+      progress({
+        step: index + 1,
+        maxStep: entries.length,
+        message: `Importing ${entry[0]}`,
+      });
+
+      if (entry[1] === FileType.Directory) {
```
aeisenberg, line 1027: So, this will ignore databases that are zips and it will also treat `testproj` databases as normal databases.

Probably the best approach would be to extract [lines 970 to 1001](https://github.com/github/vscode-codeql/pull/3797/files/1196902adc2b85b567cccc901600475d2cab956d..ce0b25feae469d7efd4f6aeef477f4dc6fb1213a#diff-24cce3c26831674245c5877c9683c8c8250c8595c4fa0d1eeece5c9b50bc5c8cR987-R1001) above into a separate function and call it hear and above.

reitowo: Good suggestion.

extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -986,6 +997,104 @@ export class DatabaseUI extends DisposableObject {
...
+          entry[1] === FileType.Directory,
+          subProgress,
+        );
+        if (database) {
+          databases.push(database);
+        } else {
+          failures.push(entry[0]);
+        }
+      } catch {
+        failures.push(entry[0]);
```
aeisenberg, line 1070: The error message should be captured here and printed alongside the failures of each database. Something like this:

```suggestion
      } catch (e) {
        failures.push(`${entry[0]}: ${toErrorMessag(e)}`);
```

(I typed this by hand so the formatting is probably incorrect.)

aeisenberg (commented):
I tried this out and it seems to work fine. I have some more suggestions around the error messages.

extensions/ql-vscode/src/databases/local-databases-ui.ts
```diff
@@ -359,6 +363,12 @@ export class DatabaseUI extends DisposableObject {
     );
   }
 
+  private async handleChooseDatabaseFoldersParentFromPalette(): Promise<void> {
+    return withProgress(async (progress) => {
+      await this.chooseDatabasesParentFolder(progress);
+    });
+  }
```
reitowo, line 370: Updated according to your suggestions with minor refine.

I also removed the title of this progress bar, because otherwise the whole message will be too long and doesn't display the actual step message of each database.

Currently will be:
image

Before:
image

aeisenberg (approved):

reitowo:
Thanks!

aeisenberg:
Thanks for your contribution!

PR #3799, opened 2024-11-08
Use standard configuration variable format for history item labels

koesie10:
This changes the `codeQL.queryHistory.format` to use a format that [is also used by VS Code](https://code.visualstudio.com/docs/editor/variables-reference), rather than using our own format. We will support the "legacy" format to ensure users that have configured this setting do not need to make any changes.

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -302,8 +302,8 @@
         "properties": {
           "codeQL.queryHistory.format": {
             "type": "string",
-            "default": "%q on %d - %s %r [%t]",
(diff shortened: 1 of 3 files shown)


robertbrignull (approved):
Code-wise LGTM. I can appreciate that the new part is basically a single call to `substituteConfigVariables` and the rest is the old legacy implementation.

It's a shame we do have to keep the legacy implementation, but of course we do have to continue supporting the old format (for potentially quite a long time if not forever).

I haven't tried running it, but I assume you have tested it and the automated tests look good.

PR #3800, opened 2024-11-11
Fix and re-enable variant analysis submission integration tests

koesie10:
This fixes the variant analysis submission integration tests by using the mock GitHub API server included in the extension to avoid issues with overriding the `fetch` function from outside of the extension code.


robertbrignull (approved):
🎉

PR #3804, opened 2024-11-13
PoC: move evaluator external log processing here

esbena:
This is a PoC for moving the external implementations we have for evaluator log processing into this repository.
This does not yet modify the runtime of this repository, but it does add immutable.js to the package.json file.

Of particular hackathon interest is the local testing feature: https://github.com/github/vscode-codeql/blob/99b43dd063bd78d4f7199871c886b7934c8517f1/extensions/ql-vscode/src/log-insights/core/cli.ts#L24

And the TypeScript versions of those: https://github.com/github/vscode-codeql/blob/99b43dd063bd78d4f7199871c886b7934c8517f1/extensions/ql-vscode/src/log-insights/core/log-processors/badness-metrics.ts#L23-L31


esbena:
(the test failures have nothing to do with my changes)

asgerf (approved):

PR #3805, opened 2024-11-13
compare-perf: Add support for sorting options

tausbn:
Adds a dropdown with (at present) two options: sorting by delta and sorting by absolute delta.


asgerf (approved):
Looks great! 👍

PR #3809, opened 2024-11-14
More work on RA pretty printing and overall presentation

asgerf:
- Styles the RA, highlighting the last part of an identifier and keywords
- Makes abbreviations ("...") clickable to show the whole abbreviation
- Also abbreviates predicates in the top-level table, not just in RA steps
- Applies a gray background to table headers. Helps distinguish the different uses of bold.

Commit by commit review recommended as a file was renamed from `.ts` to `.tsx` which messes up the combined diff.


tausbn (approved):
I didn't scrutinise the RA pretty-printing that much (I figure if it looks good, that's good enough), but the rest looks good to me. 👍

PR #3815, opened 2024-11-14
compare-perf: Add positive/negative 

tausbn:
Sadly there is no Unicode symbol for "green triangle down".

Here's what it looks like in action:
image


asgerf:
Perhaps this would look better Codicons. See https://microsoft.github.io/vscode-codicons/dist/codicon.html and `WarningIcon.tsx` for reference.

asgerf (approved):
LGTM! Feel free to merge once conflict is resolved

PR #3817, opened 2024-11-14
PoC: compare-performance of external logs

esbena:
Completes the command `codeQL.compare-performance.downloadExternalLogs` story.

If `esbena/auto/esbena/more-downloads/230f947/1731498635271` is entered as the experiment in the prompt for the `CodeQL: Download External Logs for Performance Comparison` command, then the remote logs will be downloaded, processed, and used in the comparison view.

Also turns the jsonl parser into a streaming parser with better logging

### DIFF
--- a/extensions/ql-vscode/src/compare-performance/remote-logs.ts
+++ b/extensions/ql-vscode/src/compare-performance/remote-logs.ts
@@ -10,7 +10,7 @@ import {
   remove,
   writeFileSync,
 } from "fs-extra";
-import path, { basename, dirname, join } from "path";
+import { basename, dirname, join, relative } from "path";
 import { Uri, window, workspace } from "vscode";
 import type { CodeQLCliServer } from "../codeql-cli/cli";
 import type { App } from "../common/app";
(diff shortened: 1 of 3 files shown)


extensions/ql-vscode/src/compare-performance/remote-logs.ts
```diff
@@ -0,0 +1,568 @@
...
+  private async getPotentialTargetInfos(
+    experimentName: string,
+  ): Promise<Array<MinimalDownloadsType["targets"]["string"]>> {
+    const dir = mkdtempSync(
+      path.join(tmpDir.name, "codeql-compare-performance"),
+    );
+    const tasksDir = join(dir, "tasks");
+    await ensureDir(tasksDir);
+    // XXX hardcoded path
+    const dca = "/Users/esbena/Documents/codeql-dca/dca";
```
asgerf, line 454: Seems like this won't work for anyone else. Is there an easy way around this?

esbena: good catch. I forgot about that hardcoding! Let me see what can be done.

PR #3829, opened 2024-11-18
Faster streaming 'jsonl' parser

asgerf:
Replaces the streaming JSONL parser with a faster one that doesn't call `readline`. The streaming JSONL parser on our branch was slower than the original sync version (but doesn't run out of memory), and it seems `readline` is a bottleneck:

Running the benchmark on a 21 MB logfile:
- `readJsonlReferenceImpl`: 172.4 ms (original non-streaming version)
- `readJsonlFile`: 283.3 ms (streaming version based on `readline`)
- `readJsonlFile2`: 151.3 ms (🚀  new version without `readline`)
- `justReadline`: 187.5 ms (consumes the file with `readline` and nothing else)

Note: targeting the hackathon branch. I think it's best to polish that up and try to merge it into `main`.

### DIFF
--- a/extensions/ql-vscode/src/common/jsonl-reader.ts
+++ b/extensions/ql-vscode/src/common/jsonl-reader.ts
@@ -1,4 +1,4 @@
-import { statSync } from "fs";
+import { stat } from "fs/promises";
 import { createReadStream } from "fs-extra";
 import { createInterface } from "readline";
 
@@ -33,7 +33,7 @@ export async function readJsonlFile<T>(
   }
 
   void logger?.log(
-    `Parsing ${path} (${statSync(path).size / 1024 / 1024} MB)...`,
+    `Parsing ${path} (${(await stat(path)).size / 1024 / 1024} MB)...`,
   );
   const fileStream = createReadStream(path, "utf8");
   const rl = createInterface({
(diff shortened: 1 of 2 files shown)


esbena:
nice! I would expect the non-streaming version to be faster for small files like the one you benchmarked on, and would have advocated that as the default for small files. 
For completeness, can you try on some larger files like https://github.com/codeql-dca-runners/codeql-dca-worker_javascript/actions/runs/11840886420/artifacts/2188663427 (amphtml on https://github.com/github/codeql-dca-main/issues/24803)

esbena:
(on second thought, that will be a lot of work, so don't worry about it)

asgerf:
I ran it anyway:
- `readJsonlReferenceImpl`: out of memory
- `readJsonlFile`: 6439.4 ms
- `readJsonlFile2`: 3538.4 ms
- `justReadline`: 3664.3 ms

esbena:
Excellent. That is indeed convincing!

esbena:
My two guesses:
- `readline` reads smaller chunks than a plain `readStream` does by default
- `readline` [has many string checking operations per chunk](https://github.com/nodejs/node/blob/64c647e3421e7756d68b4b12ee45e64d101b2956/lib/internal/readline/interface.js#L575-L620), the `readStream` implementation only cares for `\n`

asgerf:
Although nobody formally 'Approved' the PR I'm gonna merge to the hackathon branch so I can start preparing the PR against main.

PR #3830, opened 2024-11-18
Publish extension using Azure credentials

koesie10:
This will change the publishing of the extension to the VS Code Marketplace to use [a user-assigned managed identity](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-manage-user-assigned-managed-identities). We are [using a federated credential](https://learn.microsoft.com/en-us/azure/developer/github/connect-from-azure-identity) to allow logging in to this managed identity from the `publish-vscode-marketplace` environment. The secrets have already been set in the environment and I've tested them when retrieving the user ID.

The `@vscode/vsce` CLI will automatically retrieve the credentials when we pass the `--azure-credential` flag and run it within the context of the `azure/cli` Action.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -124,8 +124,9 @@ jobs:
     needs: build
     environment: publish-vscode-marketplace
     runs-on: ubuntu-latest
-    env:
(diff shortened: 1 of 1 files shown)


charisk (approved):
LGTM!

PR #3831, opened 2024-11-19
Add streaming 'jsonl' parser

asgerf:
Replaces the jsonl parser with a streaming version that is at least as fast as the sync version. Also see [original PR](https://github.com/github/vscode-codeql/pull/3829) against the hackathon branch.

Thanks to @esbena for the initial streaming version used for the hackaton. That version used the `readline` library to split lines, which turned out to be a bottleneck, however, making the streaming parser slower than the original sync version. So I wrote one that doesn't use `readline`.

Running the benchmark on a 21 MB logfile:
- `readJsonlReferenceImpl`: 172.4 ms (original non-streaming version)
- `readJsonlFile`: 283.3 ms (streaming version based on `readline`)
- `readJsonlFile2`: 151.3 ms (new version without `readline`)
- `justReadline`: 187.5 ms (consumes the file with `readline` and nothing else)

On a 520 MB logfile:
- `readJsonlReferenceImpl`: out of memory
- `readJsonlFile`: 6439.4 ms
- `readJsonlFile2`: 3538.4 ms
- `justReadline`: 3664.3 ms

I've added the benchmark script although the project doesn't seem to have much infrastructure for benchmark scripts. At the moment you'll have to run it with something like `ts-node` and there's no tests to ensure the benchmark script keeps working (but it will be checked for compilation errors). I'm on the fence about whether it should be committed.

### DIFF
--- a/extensions/ql-vscode/src/common/jsonl-reader.ts
+++ b/extensions/ql-vscode/src/common/jsonl-reader.ts
@@ -15,6 +15,8 @@ export async function readJsonlFile<T>(
   handler: (value: T) => Promise<void>,
   logger?: { log: (message: string) => void },
 ): Promise<void> {
+  // Stream the data as large evaluator logs won't fit in memory.
+  // Also avoid using 'readline' as it is slower than our manual line splitting.
   void logger?.log(
     `Parsing ${path} (${(await stat(path)).size / 1024 / 1024} MB)...`,
   );
--- a/extensions/ql-vscode/src/common/jsonl-reader.ts
+++ b/extensions/ql-vscode/src/common/jsonl-reader.ts
@@ -41,16 +41,15 @@ export async function readJsonlFile<T>(
       }
     });
     stream.on("end", async () => {
-      if (buffer.trim().length > 0) {
-        try {
+      try {
+        if (buffer.trim().length > 0) {
           await handler(JSON.parse(buffer));
-        } catch (e) {
-          reject(e);
-          return;
         }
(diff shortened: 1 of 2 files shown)


esbena (approved):
LGTM, but maybe the owning team wants a say as well..

extensions/ql-vscode/src/common/jsonl-reader.ts
```diff
@@ -1,26 +1,56 @@
...
+        }
+      }
+    });
+    stream.on("end", async () => {
+      if (buffer.trim().length > 0) {
+        try {
+          await handler(JSON.parse(buffer));
+        } catch (e) {
+          reject(e);
+          return;
```
aeisenberg, line 48: Minor: If you move the `logger.log` statement and the `resolve()` call into the try block, you won't need a `return` here.

aeisenberg (commented):
Looks great. Some minor comments.

Also, can you paste the results of running the benchmark somewhere?

aeisenberg:
I think it's fine to check the benchmark script in.

extensions/ql-vscode/test/benchmarks/jsonl-reader.bench.ts
aeisenberg, line 1: Can you add a file comment, or a README describing what the benchmark does and that it's not being run on a regular basis? Also, it would be nice to include the results that you have in the PR description.

aeisenberg (approved):
Nice.

PR #3833, opened 2024-11-20
Avoid resetting scroll position after clicking a node in the graph viewer

asgerf:
When clicking a node in the graph viewer, it would reset the scroll position, similar to what happens when clicking a link with `href="#"` ([such as this link](#)). This PR fixes that.

reset-scroll


aeisenberg (approved):
Nice. I haven't tried this out, but it looks reasonable.

PR #3841, opened 2024-11-25
Bump cross-spawn version

koesie10:
I just ran `npm upgrade cross-spawn` to upgrade to a non-vulnerable version of cross-spawn.

See https://github.com/github/vscode-codeql/security/dependabot/76


angelapwen (approved):
:shipit:

PR #3845, opened 2024-11-27
Fix Spurious Compilation Errors

### LINKED ISSUE #3220: Creating skeleton query pack shows spurious CodeQL compilation errors
**Describe the bug**
When creating a skeleton query pack using the extension it creates an example query (`example.ql`) and opens it automatically.
However, it directly reports compilation errors there about unresolved modules. These errors seem to be spurious though, compiling the query works fine.

This is quite problematic because creating the example query is probably the first thing new users do, and it can be quite confusing if they directly run into (spurious) errors then.

Whether this error occurs seems to depend on the state of `%userprofile%/.codeql` (on Windows; possibly similar on Linux).

**Version**
```
CodeQL extension version: 1.11.0 
CodeQL CLI version: 2.15.5 
Platform: win32 x64
```

VS Code:
```
Version: 1.85.1 (user setup)
Commit: 0ee08df0cf4527e40edc9aa28f4b5bd38bbff2b2
Date: 2023-12-13T09:49:37.021Z
Electron: 25.9.7
ElectronBuildId: 25551756
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Windows_NT x64 10.0.19045
```

**To reproduce**
1. Close VS Code
2. Delete `%userprofile%/.codeql` (on Windows)
(to be safe verify first that it does not contain anything valuable)
3. Open an empty folder with VS Code
4. In the CodeQL tab under "Queries" click "Create one to get started", or alternatively run the command "> CodeQL: Create Query"
5. Select "Java" as language
6. Wait until the extension automatically opens the `example.ql` file
:x: Bug: The extension shows multiple compilation errors
Compilation errors screenshot
7. Try to run the query
✔️ Compilation succeeds, as expected

**Expected behavior**
The extension should not show spurious compilation errors

**Additional context**
Besides that I really like this functionality of creating a skeleton pack; this seems to make it really easy to get started (though I had been using the extension before, so maybe a real new user might experience this differently).

kaeluka:
Closes #3220

issue: https://github.com/github/codeql-core/issues/4654

This notifies the language server of newly downloaded packs once a pack is added to the current ql pack, or a pack's dependencies are installed.

This is the user facing behaviour I observed:

 - ✅ creating a query pack and example.ql file using the skeleton wizard:
   - works as expected
 - ✅ creating a query pack and ql file manually, then installing the ql pack's dependencies using the vs code command palette:
   - works: this removes the spurious compilation errors, even in open files.
   - :warning: the price of supporting that is that we restart the language server each time a pack is added or a pack's dependencies are installed. There might be cheaper solutions, but I don't know them. If the reviewers don't know of a cheaper solution: I think this overhead is worth it because pack add/pack install are rare operations and the user would likely have to manually restart or reload vs code instead to make it work.

## Dear Reviewer
 - This is a PoC that works for me.
 - I had to inject the language server reference into the cli server.
   - I'm not sure if this would be considered a design violation. I'd appreciate some feedback.
   - If that's a problem, I could refactor this to break the dependency using dependency injection. Meaning: pass in a lambda function that does the notification.
 - Maybe, the same effect could also be achieved by different means (like watching the `~/.codeql/packages` subdirectories, ...). Chances are, there's room for improvement.

### DIFF
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -8,7 +8,10 @@ import type { Log } from "sarif";
 import { SemVer } from "semver";
 import type { Readable } from "stream";
 import tk from "tree-kill";
-import type { CancellationToken, Disposable, Uri } from "vscode";
+import type { CancellationToken, Disposable } from "vscode";
+import { Uri } from "vscode";
+
+import { existsSync } from "fs";
 
 import type {
   BqrsInfo,
@@ -37,6 +40,11 @@ import { LOGGING_FLAGS } from "./cli-command";
 import type { CliFeatures, VersionAndFeatures } from "./cli-version";
 import { ExitCodeError, getCliError } from "./cli-errors";
 import { UserCancellationException } from "../common/vscode/progress";
+import type { LanguageClient } from "vscode-languageclient/node";
+import {
+  DidChangeWatchedFilesNotification,
+  FileChangeType,
+} from "vscode-languageclient/node";
 
 /**
  * The version of the SARIF format that we are using.
@@ -277,6 +285,7 @@ export class CodeQLCliServer implements Disposable {
 
   constructor(
     private readonly app: App,
+    private readonly languageClient: LanguageClient,
     private distributionProvider: DistributionProvider,
     private cliConfig: CliConfig,
     public readonly logger: Logger,
@@ -1584,11 +1593,13 @@ export class CodeQLCliServer implements Disposable {
   async packAdd(dir: string, queryLanguage: QueryLanguage) {
     const args = ["--dir", dir];
     args.push(`codeql/${queryLanguage}-all`);
-    return this.runCodeQlCliCommand(
+    const ret = await this.runCodeQlCliCommand(
       ["pack", "add"],
       args,
       `Adding and installing ${queryLanguage} pack dependency.`,
     );
+    await this.notifyPackChanged(dir);
+    return ret;
   }
 
   /**
@@ -1628,11 +1639,13 @@ export class CodeQLCliServer implements Disposable {
         ...this.getAdditionalPacksArg(workspaceFolders),
       );
     }
-    return this.runJsonCodeQlCliCommandWithAuthentication(
+    const ret = await this.runJsonCodeQlCliCommandWithAuthentication(
       ["pack", "install"],
       args,
       "Installing pack dependencies",
     );
+    await this.notifyPackChanged(dir);
+    return ret;
   }
 
   /**
@@ -1748,6 +1761,29 @@ export class CodeQLCliServer implements Disposable {
       listener(this._versionAndFeatures);
     }
     this._versionChangedListeners.push(listener);
+  }
+
+  private async notifyPackChanged(packDir: string) {
+    const packFilePath = join(packDir, "codeql-pack.yml");
+    if (!existsSync(packFilePath)) {
+      throw new Error(`Pack file ${packFilePath} does not exist`);
+    }
+    await this.languageClient.sendNotification(
+      DidChangeWatchedFilesNotification.type,
+      {
+        changes: [
+          {
+            type: FileChangeType.Changed,
+            uri: Uri.file(packFilePath).toString(),
+          },
+        ],
+      },
+    );
+
+    // restarting the language client has the effect of removing compilation
+    // errors in open ql/qll files that are caused by the pack not having been
+    // installed previously:
+    await this.languageClient.restart();
   }
 
   private async refreshVersion(): Promise<VersionAndFeatures> {
--- a/extensions/ql-vscode/src/codeql-cli/cli.ts
+++ b/extensions/ql-vscode/src/codeql-cli/cli.ts
@@ -8,10 +8,7 @@ import type { Log } from "sarif";
 import { SemVer } from "semver";
 import type { Readable } from "stream";
 import tk from "tree-kill";
-import type { CancellationToken, Disposable } from "vscode";
-import { Uri } from "vscode";
-
-import { existsSync } from "fs";
+import type { CancellationToken, Disposable, Uri } from "vscode";
 
 import type {
   BqrsInfo,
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/src/extension.ts
```diff
@@ -792,12 +792,16 @@ async function activateWithInstalledDistribution(
   const languageSelectionPanel = new LanguageSelectionPanel(languageContext);
   ctx.subscriptions.push(languageSelectionPanel);
 
+  void extLogger.log("Initializing CodeQL language server.");
+  const languageClient = createLanguageClient(qlConfigurationListener);
```
kaeluka, line 796: this changes the initialization order of things because we now inject the language client into sub-components. It seems to work fine for me locally.

kaeluka:
Koen pointed out this could be generalized: https://github.com/github/codeql-core/issues/4654#issuecomment-2506073240 -> draft mode

extensions/ql-vscode/src/extension.ts
```diff
@@ -748,9 +748,13 @@ async function activateWithInstalledDistribution(
     );
   ctx.subscriptions.push(qlConfigurationListener);
 
+  void extLogger.log("Initializing CodeQL language server.");
+  const languageClient = createLanguageClient(qlConfigurationListener);
+
```
kaeluka, line 753: moving this up a couple hundred lines is perhaps the most risky change in this PR — this changes the initialization order of things because we now inject a reference to the language client into thes cli server. It seems to work fine for me locally.

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1750,6 +1763,29 @@ export class CodeQLCliServer implements Disposable {
...
+            uri: Uri.file(packFilePath).toString(),
+          },
+        ],
+      },
+    );
+
+    // restarting the language client has the effect of removing compilation
+    // errors in open ql/qll files that are caused by the pack not having been
+    // installed previously:
+    await this.languageClient.restart();
```
kaeluka, line 1786: Restarting the client is a relatively big hammer to solve this problem. I suspect there might be a cheaper way to accomplish the same thing. But considering how unusual the pack add/pack install operations are and that they take a significant amount of time in the first place — it's probably not the end of the world 🤷

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1750,6 +1763,29 @@ export class CodeQLCliServer implements Disposable {
     this._versionChangedListeners.push(listener);
   }
 
+  private async notifyPackChanged(packDir: string) {
+    const packFilePath = join(packDir, "codeql-pack.yml");
```
koesie10, line 1767: I believe pack files can also be located at `qlpack.yml` for legacy pack files. You should be able to use the `getQlPackFilePath` function to get the correct pack file.

koesie10 (commented):
I've left a few small comments, but I'll leave a full review of this approach to someone that is more familiar with the language server/CLI interactions.

kaeluka:
Thank you for your comments <3 I'll work on them next week.

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1750,6 +1763,29 @@ export class CodeQLCliServer implements Disposable {
...
+      DidChangeWatchedFilesNotification.type,
+      {
+        changes: [
+          {
+            type: FileChangeType.Changed,
+            uri: Uri.file(packFilePath).toString(),
+          },
+        ],
+      },
+    );
```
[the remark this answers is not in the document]
kaeluka: Oh, good one! Chances are, this can actually go. I'll try it out next week

extensions/ql-vscode/src/codeql-cli/cli.ts
aeisenberg: The time to restart the language server is much smaller than the time to run most `pack` commands. I think this is fine.

extensions/ql-vscode/src/codeql-cli/cli.ts
kaeluka: Thanks for pointing that out (now I know!). But addressing your other comment "Do you still need this if you are also restarting the language client?" meant this logic was removed altogether :)

extensions/ql-vscode/src/codeql-cli/cli.ts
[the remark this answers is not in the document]
kaeluka: It works! Thx! 😄

extensions/ql-vscode/src/codeql-cli/cli.ts
```diff
@@ -1750,6 +1763,29 @@ export class CodeQLCliServer implements Disposable {
     this._versionChangedListeners.push(listener);
   }
 
+  private async notifyPackChanged(packDir: string) {
+    const packFilePath = join(packDir, "codeql-pack.yml");
+    if (!existsSync(packFilePath)) {
+      throw new Error(`Pack file ${packFilePath} does not exist`);
```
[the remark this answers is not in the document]
kaeluka: OK, I agree with the reasoning! But due to Koen's insight that we may not need the explicit file change notification (we don't!), the condition in question could be removed altogether 😄

extensions/ql-vscode/src/codeql-cli/cli.ts
kaeluka: Thanks for confirming :) 👍

kaeluka:
Thanks @koesie10 and @aeisenberg for your reviews! They've increased my confidence, as well as made the PR shorter and simpler :) This commit is the result: 17542d1041470d8ee02066428b6f3bde5defb6ce

koesie10 (approved):
Thanks, LGTM! I've also tested this using the reproduction steps in the public and internal issues and it doesn't show any compilation errors once the pack has been installed.

Can you please add "Closes https://github.com/github/vscode-codeql/issues/3220" to the issue description since this PR fixes that issue?

kaeluka:
> I've also tested this using the reproduction steps in the public and internal issues and it doesn't show any compilation errors once the pack has been installed.

Thanks for taking the time to do that, I appreciate it!

> Can you please add "Closes #3220" to the issue description since this PR fixes that issue?

done!

Issue #3672, opened 2024-07-25
Less obtrusive way to ask users to collect telemetry


isidorn:
Hi VS Code PM Here 👋 

1. Install CodeQl
3. See telemetry dialog 🐛 

Here are our telemetry collection guidelines https://code.visualstudio.com/api/extension-guides/telemetry

Based on that I strongly suggest that you do not show this dialog, but instead just respect the respect the `isTelemetryEnabled` and `onDidChangeTelemetryEnabled` from the vscode API. That will make a much smoother user experience, and the user just has to disable telemetry once in vscode and all the extensions will respect that. Thanks

![Image](https://github.com/user-attachments/assets/cf331495-0dec-41a0-8f4e-9776f2f72056)

fyi @TylerLeonhardt

jf205:
Hi @isidorn 

Thanks for creating this issue and for the recommendation. I'll discuss this change with the team.

jf205:
Discussed with the team: we agree that we should respect the `isTelemetryEnabled` and `onDidChangeTelemetryEnabled` from the vscode API. Currently unsure what size effort this would be and therefore also unsure when we'll be able to pick it up.

isidorn:
Thanks for discussing. I do not think this should be a large investment.

stonish:
@isidorn We discussed this a little further as a team, we're onboard with making this change but it's not something that we think we can tackle right now. Even though the investment shouldn't be too large, we're working towards some tight deadlines for shipping new features at Universe and can't take on anything else unless absolutely necessary.

I'm going to set this up for us to revisit it after Universe when I expect we'll have more capacity to take this on. If you think this is a mistake, please let me know and/or re-add this issue to our team board to be re-triaged

isidorn:
Post Universe makes sense. Thank you!

PR #3853, opened 2024-12-05
Remove extension telemetry setting and use VS Code setting instead

koesie10:
This removes the `codeQL.telemetry.enableTelemetry` setting and uses [the VS Code settings](https://code.visualstudio.com/docs/getstarted/telemetry) instead. This is done automatically in the `vscode-extension-telemetry` package, so we don't need to check the VS Code settings.

Since this changes the telemetry when you currently have the extension telemetry disabled, but the VS Code telemetry enabled, this also adds a non-dismissable notification when starting the extension (if VS Code telemetry is enabled and `codeQL.telemetry.enableTelemetry` is `false`).

The E2E tests are expected to fail since it installs the released extension rather than the extension from the repo.

- Closes https://github.com/github/vscode-codeql/issues/3672

### DIFF
--- a/extensions/ql-vscode/src/common/vscode/telemetry.ts
+++ b/extensions/ql-vscode/src/common/vscode/telemetry.ts
@@ -1,6 +1,7 @@
 import type { Extension, ExtensionContext } from "vscode";
+import { ConfigurationTarget, env, Uri, window } from "vscode";
 import TelemetryReporter from "vscode-extension-telemetry";
-import { LOG_TELEMETRY, isCanary } from "../../config";
+import { ENABLE_TELEMETRY, isCanary, LOG_TELEMETRY } from "../../config";
 import type { TelemetryClient } from "applicationinsights";
 import { extLogger } from "../logging/vscode";
 import { UserCancellationException } from "./progress";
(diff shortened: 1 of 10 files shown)


extensions/ql-vscode/src/common/vscode/telemetry.ts
```diff
@@ -271,21 +158,45 @@ export class ExtensionTelemetryListener
...
+    isCloseAffordance: false,
+  };
+  const codeqlTelemetryItem = {
+    title: "More Information about CodeQL Telemetry",
+    isCloseAffordance: false,
+  };
+  let chosenItem;
+
+  do {
+    chosenItem = await window.showInformationMessage(
```
robertbrignull, line 176: It's slightly annoying you get a "cancel" button but if you click it the dialog just comes back. Sadly I can't see a way to get rid of it and there's nothing sensible we can really do to "cancel" this update, so I agree I think we just leave it like this. Users will only see it once at least.

Screenshot 2024-12-05 at 16 41 03

robertbrignull (approved):
Code LGTM and I tried it out and the popup worked as expected (one non-blocking comment)

PR #3859, opened 2024-12-06
Upgrade code-server version for E2E tests

koesie10:
This upgrades the version of `code-server` and `alpine` as used in the E2E tests. The E2E tests are currently expected to fail until we do a release due to the changes in https://github.com/github/vscode-codeql/pull/3853.

### DIFF
--- a/extensions/ql-vscode/test/e2e/README.md
+++ b/extensions/ql-vscode/test/e2e/README.md
@@ -9,12 +9,12 @@ Setup
 - install playwright if you haven't yet (`npx playwright install`)
 - go to the e2e test folder on your terminal
 - make sure docker is running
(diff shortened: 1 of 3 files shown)


charisk (approved):
Looks good, but what prompted the upgrade?

koesie10:
> Looks good, but what prompted the upgrade?

Nothing in particular, but I was updating the E2E tests as part of https://github.com/github/vscode-codeql/pull/3853 and noticed we were on a very old version of code-server (which means we'd also be on a very old version of VS Code). It would probably be a good idea to add a Dependabot config for updating the `Dockerfile`, so I'll do that in a follow-up PR.

PR #3862, opened 2024-12-09
Add cache of query schemas to results view

alexet:
The pagination this does is slow as it scan to compute the index of each page. By caching we mostly only do the scanning once and then later we can change page quickly. Unfortunately we must clear the cache on sorting though as that may reuse an existing bqrs.

For very large result sets (my test had 20,000,000 tuples) this takes the time to change page from about 10s to seemingly instant to me (I didn't precisely measure this, I just looked at a clock to get 10s) making looking through sorted data for certain results much easier.

I also removed a call to bqrs info with a page size of zero. Actually passing the pageSize of zero to the cli would result in an error, but as 0 is falsy so we treat it the same as the case without a pagesize. It seems clearer to just not provide the argument.

I am not sure why the E2E tests are failing though so I may need help debugging that.


aeisenberg:
I merged into main. The E2E tests should be fixed now.

aeisenberg:
Hmmm....maybe not.

alexet:
I merged in main in another PR and got the same error: https://github.com/github/vscode-codeql/actions/runs/12282822206/job/34274881774?pr=3867

aeisenberg:
We need to do a new extension release in order to get the tests passing again. See https://github.com/github/vscode-codeql/pull/3869.

aeisenberg (approved):
Looks good (after updating the changelog).

PR #3869, opened 2024-12-11
Revert "Bump @babel/plugin-transform-modules-commonjs from 7.24.7 to 7.26.3 in /extensions/ql-vscode"

aeisenberg:
Reverts github/vscode-codeql#3857

Let's see if we can get our e2e tests passing after this.

### DIFF
--- a/extensions/ql-vscode/package-lock.json
+++ b/extensions/ql-vscode/package-lock.json
@@ -46,7 +46,7 @@
       },
       "devDependencies": {
         "@babel/core": "^7.24.6",
-        "@babel/plugin-transform-modules-commonjs": "^7.26.3",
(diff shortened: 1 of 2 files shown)


angelapwen:
Looks like the test is still failing now 😢

aeisenberg:
With the same error as before.

charisk:
I believe the failures are expected (see note in https://github.com/github/vscode-codeql/pull/3853)

aeisenberg:
Ah...thanks. I missed that. Does this mean we need to create an extension release in order for the tests to start passing?

charisk:
> Ah...thanks. I missed that. Does this mean we need to create an extension release in order for the tests to start passing?

That's my understanding.

PR #3882, opened 2024-12-20
Fix publishing to VS Code Marketplace

koesie10:
This may fix publishing to the VS Code Marketplace, which is currently broken because `npx` isn't included in the `azure/cli` Docker image. This is based on the publishing steps of a Microsoft VS Code extension: https://github.com/microsoft/vscode-azureapicenter/blob/13175ef0638b56fa1e605bd4750ec77b1ba5b72d/.github/workflows/publish.yml#L57-L71

I don't think we can test this without creating another release.


aeisenberg (approved):

aeisenberg:
I'll try to republish later today.

PR #3903, opened 2025-01-19
Fix regex in CodeQL TextMate grammar that was silently failing

slevithan:
The TM grammar for CodeQL includes an invalid regex. Specifically, in Oniguruma, lookahead is invalid when used within lookbehind. This leads to the regex failing silently in `vscode-textmate`.

Additionally, this makes the CodeQL grammar fail entirely when used with [Shiki](https://shiki.style/)'s JS engine (see CodeQL listed as unsupported in Shiki's JS engine [compatibility report](https://shiki.style/references/engine-js-compat)) because Shiki's JS engine doesn't fail *silently* for invalid Oniguruma regexes.

Because the lookahead appears at the end of the lookbehind, moving it outside of the lookbehind does not change the behavior of the regex at all (if the previous version had worked in the first place, like it would in some other regex flavors).

In this PR, I updated `extensions/ql-vscode/syntaxes/ql.tmLanguage.yml` and then ran the `syntaxes/updateSyntax` script to generate the JSON version of the grammar. I've also updated the changelog.


slevithan:
@aeisenberg any concerns with landing this?

aeisenberg:
Looks fine to me. Thanks for your contribution!

aeisenberg (approved):

PR #3923, opened 2025-02-11
ComparePerformance: fix crash in RTJO mode

d10c:
With `--dynamic-join-order-mode=all`, both `first` and `second` are
undefined, leading to a crash. Until RTJO mode gets proper tuple
counting, this change prevents a crash when unfurling the steps of a
predicate.


asgerf (commented):
LGTM and I wrote the affected code; though someone with merge rights will need to approve

PR #3927, opened 2025-02-13
Add permissions block and actions analysis

aeisenberg:
This change does two things. If it is complicated to review I will split up.

First, this ensures that all workflows have minimal permissions blocks. Second, this adds actions analysis.

A little bit of extra security hardening.

### DIFF
--- a/.github/workflows/cli-test.yml
+++ b/.github/workflows/cli-test.yml
@@ -17,6 +17,8 @@ jobs:
   find-nightly:
     name: Find Nightly Release
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     outputs:
       url: ${{ steps.get-url.outputs.nightly-url }}
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/cli-test.yml
```diff
@@ -17,6 +17,8 @@ jobs:
   find-nightly:
```
koesie10, line 17: Should we set `permissions: {}` on the top-level so any new jobs that are defined won't get permission by default? Alternatively, we could also set `permissions: { contents: read }` at the workflow-level and keep the `permissions` block on the `report-failure` job.

koesie10 (approved):
Thanks!

.github/workflows/cli-test.yml
aeisenberg: Sure.

PR #3930, opened 2025-02-19
Fix package-lock.json to have `acorn` as a direct dependency

koesie10:
This fixes `npm run storybook` by ensuring that `node_modules/acorn` is installed on the same level as `node_modules/acorn-jsx`. I'm not sure why we have this discrepancy.

Screenshot 2025-02-19 at 09 55 07


robertbrignull:
What commands did you run to make these `package-lock.json` changes?

robertbrignull (approved):
LGTM

I was mostly worried that future changes might accidentally undo this, perhaps if one of these other packages that depends on `acorn` gets updated. Hopefully that's unlikely unless the `package-lock.json` file gets completely discarded and recreated. 🤞🏼

PR #3931, opened 2025-02-19
Fix Storybook build

koesie10:
This fixes `npm run build-storybook` by ensuring we're not importing anything from the `path` module in stories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ artifacts/
 # E2E Reports
 **/playwright-report/**
 **/test-results/**
+
+# Storybook artifacts
+**/storybook-static/**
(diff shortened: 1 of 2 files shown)


robertbrignull (approved):
LGTM

Should `build-storybook` be added to `.github/workflows/main.yml` with the other build/linting CI jobs so we don't break it again in future?

koesie10:
> Should `build-storybook` be added to `.github/workflows/main.yml` with the other build/linting CI jobs so we don't break it again in future?

That's a good idea. I can definitely add that, we can even publish the stories to GitHub Pages. I'll do that in a follow-up PR though since it's unrelated to this.

PR #3961, opened 2025-03-19
Add "actions" as an analysis language

aeisenberg:
Allows interacting with qlpacks that use actions as a language.


koesie10 (approved):
Changes LGTM! Should we also add an example database to [this list](https://github.com/github/vscode-codeql/blob/d23c3c5b57f31d813cd3cad9e790cd5525285a53/extensions/ql-vscode/src/local-queries/skeleton-query-wizard.ts#L40-L50)?

PR #3970, opened 2025-03-26
Add "rust" as an analysis language

redsun82:
This allows to interact with qlpacks that use rust as a language, which is currently experimental.

### DIFF
--- a/extensions/ql-vscode/src/common/query-language.ts
+++ b/extensions/ql-vscode/src/common/query-language.ts
@@ -6,6 +6,7 @@ export enum QueryLanguage {
   Javascript = "javascript",
   Python = "python",
   Ruby = "ruby",
+  Rust = "rust",
   Swift = "swift",
 }
 
@@ -25,6 +26,8 @@ export function getLanguageDisplayName(language: string): string {
       return "Python";
     case QueryLanguage.Ruby:
       return "Ruby";
+    case QueryLanguage.Rust:
+      return "Rust";
     case QueryLanguage.Swift:
       return "Swift";
     default:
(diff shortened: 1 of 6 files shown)


koesie10 (commented):
Thanks, this makes sense to me.

The only thing I can find that might be missing is [this constant](https://github.com/github/vscode-codeql/blob/9fa6d99f09414c6eb9be55b067c77fae9701b89c/extensions/ql-vscode/src/local-queries/skeleton-query-wizard.ts#L40-L49). While we may not have a good repo yet, we also don't really handle it being empty. This is probably fine for now since the CLI server will not return Rust as a supported language, but if we do have a repo that we can add here, that would be nice. Perhaps just adding `github/stack-graphs` for now would be fine?

koesie10:
> @koesie10 Thanks for spotting that. What should be the qualities of a repo to be added to that mapping? Long term, I'd probably go for https://github.com/denoland/deno, it's a top-rated rust repository and we know we extract it pretty well already. Should I go for that already, or should I stick to GitHub owned repositories while we only staff shipped?

I think generally we go for somewhat well-known repositories that also aren't too large. I think it could make sense to go for `denoland/deno`, unless it has a very large database. It seems like the largest database currently in that list it the Swift database at 12MB, so much more than that is probably too large and then it might make sense to choose another prominent Rust project that has a smaller database.

redsun82:
> > @koesie10 Thanks for spotting that. What should be the qualities of a repo to be added to that mapping? Long term, I'd probably go for https://github.com/denoland/deno, it's a top-rated rust repository and we know we extract it pretty well already. Should I go for that already, or should I stick to GitHub owned repositories while we only staff shipped?
> 
> I think generally we go for somewhat well-known repositories that also aren't too large. I think it could make sense to go for `denoland/deno`, unless it has a very large database. It seems like the largest database currently in that list it the Swift database at 12MB, so much more than that is probably too large and then it might make sense to choose another prominent Rust project that has a smaller database.

Ok, `denoland/deno` seems like is too large. I'll go for https://github.com/sharkdp/bat, also quite successful and the (zipped) DB is currently at 7.7 MB

redsun82:
I could test that VSCode is succeeding in starting out a MRVA run on rust :tada:
image
All queries are currently failing, but that is due to still missing upgrade and downgrade scripts (we'll get to that hopefully soon), so that's expected.

koesie10 (approved):
Thanks!

PR #3972, opened 2025-03-27
Fix legacy label regex

koesie10:
When the `legacyLabelRegex` is tested multiple times against the same string (as happens most of the time, unless a user-specified label is added), it would start searching from the end of the previous match instead of from the start of the string:

[“JavaScript RegExp objects are stateful when they have the global or sticky flags set (e.g., /foo/g or /foo/y). They store a lastIndex from the previous match. Using this internally, test() can be used to iterate over multiple matches in a string of text (with capture groups).”](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test)

This fixes it by resetting `lastIndex` before running the `test` method.


starcke (approved):
:+1: Nice find!

PR #3979, opened 2025-03-28
Use newer version of JSDOM

koesie10:
To prepare for upgrading to `@vscode-elements/elements`, we need a newer version of JSDOM that better supports web components. This upgrades to a newer version of JSDOM by using a custom Jest environment that is available in Jest 30.


extensions/ql-vscode/package.json
```diff
@@ -2079,9 +2081,9 @@
     "gulp-replace": "^1.1.3",
     "gulp-typescript": "^5.0.1",
     "husky": "^9.1.5",
-    "jest": "^29.0.3",
-    "jest-environment-jsdom": "^29.0.3",
+    "jest": "^30.0.0-alpha.7",
     "jest-runner-vscode": "^3.0.1",
+    "jsdom": "^26.0.0",
```
robertbrignull, line 2086: Looking at [`jest-environment-jsdom`](https://www.npmjs.com/package/jest-environment-jsdom?activeTab=code) it seems we're going `jsdom` version 20 to 26.

Looking in the [changelog](https://github.com/jsdom/jsdom/blob/main/Changelog.md) I can't immediately see any references to the fields that weren't defined and were causing errors, but it could be I missed it or it's from another updated dependency 🤷🏼

robertbrignull (approved):
LGTM if this helps us move forwards with the vscode/elements migration. Maybe it's not ideal being on an alpha version but given we only use it for testing it can hopefully only cause stability issues for us and not for users 🤷🏼

extensions/ql-vscode/package.json
koesie10: I believe what we're looking for is in [version 23.1.0](https://github.com/jsdom/jsdom/blob/main/Changelog.md#2310): "Added an initial implementation of `ElementInternals`, including the `shadowRoot` getter and the string-valued ARIA properties." This is what was giving errors when using some `@vscode-elements/elements` components (after fixing a missing `CSSProperties.replaceSync` definition).

PR #3983, opened 2025-04-01
Remove the use of `VSCodeTag`

tuan-nguen:
This is part of the work to remove the use of `@vscode/webview-ui-toolkit`. The new library `@vscode-elements/elements` does not have a direct replacement for `VSCodeTag`, this implements a new `Tag` component for which the styles have been mostly copied from [here](https://github.com/microsoft/vscode-webview-ui-toolkit/blob/main/src/tag/tag.styles.ts). Some of the css attributes are hard-coded (e.g. `background-color` and `color`) because using the corresponding CSS variables makes the styles different for different themes. I have attached comparison screenshots of the old and new component.

Dark theme
Screenshot 2025-04-01 at 10 13 32

Light theme
Screenshot 2025-04-01 at 10 13 38

### DIFF
--- a/extensions/ql-vscode/src/view/common/Tag.tsx
+++ b/extensions/ql-vscode/src/view/common/Tag.tsx
@@ -1,0 +1,14 @@
+import { styled } from "styled-components";
+
+export const Tag = styled.div`
+  background-color: var(--vscode-badge-background);
+  border: 1px solid var(--vscode-button-border, transparent);
+  border-radius: 2px;
+  color: var(--vscode-badge-foreground);
+  padding: 2px 4px;
+  text-transform: uppercase;
+  box-sizing: border-box;
+  font-family: var(--vscode-font-family);
+  font-size: 11px;
+  line-height: 16px;
+`;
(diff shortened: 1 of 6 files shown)


extensions/ql-vscode/src/view/common/Tag.tsx
```diff
@@ -0,0 +1,14 @@
+import { styled } from "styled-components";
+
+export const Tag = styled.span`
+  background-color: #4d4d4d;
+  border: 1px solid transparent;
+  border-radius: 2px;
+  color: #ffffff;
```
robertbrignull, line 7: > Some of the css attributes are hard-coded (e.g. background-color and color) because using the corresponding CSS variables makes the styles different for different themes. 

Could you explain why we wouldn't want the styles to change?

tuan-nguen: I thought it makes more sense the keep the appearance the same as the old component. I am happy to use the CSS variables if you think that's the better approach. Here are some screenshots of the differences where left is the old component and the right one is the new component.

Github dark theme
Screenshot 2025-04-01 at 12 26 37

Dark high contrast theme
Screenshot 2025-04-01 at 12 26 02

robertbrignull (approved):
LGTM

I've given it another quick local test and the colours all look good to me and match with the other components.

PR #3987, opened 2025-04-03
Compare perf: Fix missing unit in grouped predicate entry

asgerf:
Fixes a minor bug in the compare performance view, where the "ms" unit was not displayed on a grouped predicate entry.

Since the optional parameter `unit` is now provided at all call sites, and its absence was the cause of the bug, I also decided to make the parameter mandatory.

# Before: missing unit
The unit was missing next to the first `2` in this screenshot:
Screenshot 2025-04-03 at 14 52 45

# After: unit is there
(sorry, I didn't have the same query history item available for both screenshots)

Screenshot 2025-04-03 at 14 52 16


koesie10 (approved):
LGTM, thanks!

PR #4000, opened 2025-04-11
Set up CSS module support

koesie10:
This sets up support for CSS modules such that the following would now work:

`ActionButton.module.css`:

```css
.button {
  display: block;
}
```

```tsx
import styles from "./ActionButton.module.css";

export const ActionButton = () => {
  return <button className={styles.button} />
}
```


tuan-nguen (approved):
Thanks!

PR #4007, opened 2025-04-17
Make fields in PipelineRun optional

robertbrignull:
Hopes to fix a crash when using the performance comparison viewer. But I'm not very confident if this is correct or not so will need review from someone with domain knowledge.

> Remember to update the [changelog](https://github.com/github/vscode-codeql/blob/main/extensions/ql-vscode/CHANGELOG.md) if there have been user-facing changes!


robertbrignull:
Closing as this isn't the solution we're going to go with.

PR #4013, opened 2025-04-23
Pin azure/login Action

koesie10:
Closes https://github.com/github/vscode-codeql/security/code-scanning/646

Dependabot is configured to update actions, so when a new version of this action is released Dependabot should open a PR to update the SHA and comment.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -147,7 +147,7 @@ jobs:
           name: vscode-codeql-extension
 
       - name: Azure User-assigned managed identity login
-        uses: azure/login@v2
+        uses: azure/login@a457da9ea143d694b1b9c7c869ebb04ebe844ef5 # v2.3.0
         with:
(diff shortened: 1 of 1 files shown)


robertbrignull (approved):

.github/workflows/release.yml
```diff
@@ -147,7 +147,7 @@ jobs:
           name: vscode-codeql-extension
 
       - name: Azure User-assigned managed identity login
-        uses: azure/login@v2
+        uses: azure/login@a457da9ea143d694b1b9c7c869ebb04ebe844ef5 # v2.3.0
```
robertbrignull, line 150: Will dependabot update the "v2.3.0" comment too?

koesie10: Yes, it should: https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/

PR #4018, opened 2025-04-25
Compare performance: Detect 'shadowed' cache hits and a few other fixes

asgerf:
Adds detection of 'shadowed' cache hits in the compare performance view, greatly improving its usefulness when comparing two query runs where the cache was not cleared in-between. Previously, one side of the comparison would get blamed for the cost of evaluating a lot of predicates that the other side could omit because of the cache. Detecting this is harder than it sounds, see explanation comment in 06fcd6fc47454092a249396c23d03c4199eb6a9b.

Also fixes a few other things, such as the `NAMED_LOCAL` not being handled, and the `millis` field referring to the entire SCC cost on `COMPUTE_RECURSIVE` events, which is not what we want.

Commit-by-commit strongly recommended.


extensions/ql-vscode/src/view/compare-performance/ComparePerformance.tsx
```diff
@@ -860,3 +881,14 @@ function collatePipelines(
 function samePipeline(a: string[], b: string[]) {
   return a.length === b.length && a.every((x, i) => x === b[i]);
 }
+
+function getPipelineSummaryHash(pipelines: Record<string, PipelineSummary>) {
+  // Note: we can't import "crypto" here because it is not available in the browser,
+  // so we just concatenate the hashes of the individual pipelines.
```
aeisenberg, line 978: How long will this be? Can we possibly have 100s of keys? It might be safer to do some checks to ensure that the `result` is never larger than a fixed amount. And if it is, then you can sample the first `n` chars in each key to keep it under the limit, retain a reasonable amount of stochasticity, and ensure the result easy to reproduce given the same input.

extensions/ql-vscode/src/view/compare-performance/ComparePerformance.tsx
```diff
@@ -97,6 +97,75 @@ class ComparisonDataset {
...
+        if (ownIndex != null) {
+          visit(ownIndex, raHash);
+        }
+      }
+      {
+        // Look up by name
+        const name = other.data.names[otherCacheHit];
+        const ownIndex = this.nameToIndex.get(name);
+        if (ownIndex != null) {
+          visit(ownIndex, this.data.raHashes[ownIndex]);
```
aeisenberg, line 162: nit: This should work.

```suggestion
          visit(ownIndex, raHashes[ownIndex]);
```

aeisenberg (commented):
Looks good. I have one real comment and a couple of copy-edit nits.

I haven't tried this locally.

extensions/ql-vscode/src/view/compare-performance/ComparePerformance.tsx
asgerf: Back when the "standard order" existed there could be multiple pipelines, but nowadays there are 2 for recursive predicates and 1 for all other predicates. So it's really small.

aeisenberg (approved):

PR #4056, opened 2025-06-18
Convert to eslint 9

robertbrignull:
Upgrades eslint to version 9, as well as upgrading a few plugins.

I've tried to convert the old config as closely as I could, but I've removed a few plugins that weren't working well because they haven't received updates in a couple of years. It's also very possible I've misunderstood bits of the config!

I can now run `npm run lint` locally and hopefully it'll pass in CI too. I have fixed a few rules where it was simple or obvious to do, but for most rules I have disabled them as there were thousands of errors. After this is merged we'll want a secondary effort to try to re-enable rules as much as feasible.

I've tried to keep the list of commits as clear and understandable as possible, so hopefully you can review this commit-by-commit.

### DIFF
--- a/extensions/ql-vscode/src/view/common/SuggestBox/SuggestBox.tsx
+++ b/extensions/ql-vscode/src/view/common/SuggestBox/SuggestBox.tsx
@@ -200,7 +200,6 @@ export const SuggestBox = <
   }, [disabled]);
 
   return (
-
     <div onClick={stopClickPropagation}>
       {renderInputComponent(
         getReferenceProps({


extensions/ql-vscode/src/common/mock-gh-api/request-handlers.ts
```diff
@@ -73,6 +73,7 @@ async function readRequestFiles(
     ) {
       request.response.body = await readFile(
         join(scenarioDirPath, request.response.body.substring(5)),
+        "utf8",
```
koesie10: I don't think this is equivalent/correct. All scenarios that have a `file:` response body have a ZIP file as their response and should not be decoded as UTF-8, but should be returned as-is.

extensions/ql-vscode/src/view/common/SuggestBox/SuggestBox.tsx
```diff
@@ -201,7 +201,7 @@ export const SuggestBox = <
 
   return (
     // Disabled because the div is used to stop click propagation, it's not a button
```
koesie10:
```suggestion
```

extensions/ql-vscode/test/vscode-tests/test-dispose-handler.ts
```diff
@@ -1,11 +1,11 @@
...
-export function testDisposeHandler(disposable: any & Disposable) {
+export function testDisposeHandler(disposable: Disposable) {
   if (
-    disposable.onDidExpandElement &&
-    disposable.onDidCollapseElement &&
-    disposable.reveal
+    (disposable as any).onDidExpandElement &&
+    (disposable as any).onDidCollapseElement &&
+    (disposable as any).reveal
   ) {
```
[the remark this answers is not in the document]
koesie10: I think this suggestion from Copilot makes sense, assuming the `in` checks are also `true` for an actual TreeViewer.

extensions/ql-vscode/src/local-queries/query-contents.ts
```diff
@@ -15,7 +15,7 @@ import { dbSchemeToLanguage } from "../common/query-language";
  */
 export function getInitialQueryContents(language: string, dbscheme: string) {
   if (!language) {
-    const dbschemeBase = basename(dbscheme) as keyof typeof dbSchemeToLanguage;
+    const dbschemeBase = basename(dbscheme);
     language = dbSchemeToLanguage[dbschemeBase];
```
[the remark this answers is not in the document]
robertbrignull: > could cause a runtime error

I think the code being removed is entirely compile-time, so I'm going to ignore this copilot suggestion about runtime errors.

extensions/ql-vscode/src/common/mock-gh-api/request-handlers.ts
robertbrignull: Ah sorry. I thought this was textual data. I can revert this bit and find another way around the type issues.

Issue #3739, opened 2024-10-03
Show path provenance in result comparison


aschackmull:
Would it be possible to get the provenance information from flow paths shown in the compare view as well? Currently provenance shows up nicely in the paths in the alerts view, but when paths are shown in a comparison between two path-problem result sets, then there's no provenance shown.

dbartol:
Potential related issue, from @owenmc on Slack:
>More feedback on the same feature (I'm comparing lots of results right now): sometimes it seems to think that two alerts are different when I think they're the same. Everything that is shown in the comparison is identical. I know that I've changed the model for the sink from one defined in ql to one defined in models-as-data. Is that what is causing them to be considered different? I would suggest that we should be robust to this.

While we're working on displaying the comparison results, we should also determine whether the provenance info was affecting the comparison even though it was not being displayed.

PR #4065, opened 2025-06-29
Add "View Autofixes" feature for variant analysis results

jcogs33:
## Description

Adds a "View Autofixes" feature for variant analysis results. This is a canary-only feature.
Generates autofixes for selected variant analysis results and opens the resulting autofix outputs in a markdown file.
Activated via the "View Autofixes" button in the variant analysis webview or via right-clicking the variant analysis query history item. 

AutofixInMrvaScreenshot

</br>
</br>
At a high-level, the implementation:

- Checks for a local checkout of autofix
- Sets up to run autofix by:
  - Overriding the relevant query help
  - Getting the relevant SARIF(s)
  - Downloading source root(s)
- Runs autofix locally
- Outputs autofix results in a markdown file

Commit-by-commit review recommended. Let me know if you want me to split anything onto a separate PR.

## Consideration

- See in-line comments
- Change from my original implementation: Instead of downloading databases and then extracting the source root from the database, I directly download the source root since it seemed a bit faster. Let me know if you think the database approach is better.
- I will make a follow-up PR to convert this to support Go autofix instead of cocofix, and I'll update the `queryHelpOverrideDirectory` to the new `go/pkg/autofix/prompt/qhelps` directory in that PR.

## Testing
- I've manually tested this feature but have not added unit/view/integration tests since it seems [another canary feature](https://github.com/github/vscode-codeql/pull/3843) did not add tests. Let me know if I should add tests after all.
- Note: There is a known issue with the E2E CI test failing. It is not caused by changes in this PR.

### DIFF
--- a/extensions/ql-vscode/src/common/interface-types.ts
+++ b/extensions/ql-vscode/src/common/interface-types.ts
@@ -151,10 +151,13 @@ interface SetStateMsg {
 export interface UserSettings {
   /** Whether to display links to the dataflow models that generated particular nodes in a flow path. */
   shouldShowProvenance: boolean;
+  /** Whether to display the "View Autofixes" button. */
+  shouldShowViewAutofixesBtn: boolean;
 }
 
 export const DEFAULT_USER_SETTINGS: UserSettings = {
   shouldShowProvenance: false,
+  shouldShowViewAutofixesBtn: false,
 };
 
 /** Message indicating that the user's configuration settings have changed. */
@@ -559,7 +562,8 @@ export type ToVariantAnalysisMessage =
   | SetVariantAnalysisMessage
   | SetFilterSortStateMessage
   | SetRepoResultsMessage
-  | SetRepoStatesMessage;
+  | SetRepoStatesMessage
+  | SetUserSettingsMsg;
 
 export type FromVariantAnalysisMessage =
   | CommonFromViewMessages
--- a/extensions/ql-vscode/src/common/interface-types.ts
+++ b/extensions/ql-vscode/src/common/interface-types.ts
@@ -151,19 +151,35 @@ interface SetStateMsg {
 export interface UserSettings {
   /** Whether to display links to the dataflow models that generated particular nodes in a flow path. */
   shouldShowProvenance: boolean;
-  /** Whether to display the "View Autofixes" button. */
-  shouldShowViewAutofixesBtn: boolean;
 }
 
 export const DEFAULT_USER_SETTINGS: UserSettings = {
   shouldShowProvenance: false,
-  shouldShowViewAutofixesBtn: false,
 };
 
 /** Message indicating that the user's configuration settings have changed. */
 interface SetUserSettingsMsg {
   t: "setUserSettings";
   userSettings: UserSettings;
+}
+
+export interface VariantAnalysisUserSettings {
+  /** Whether to display the "View Autofixes" button. */
+  shouldShowViewAutofixesButton: boolean;
+}
+
+export const DEFAULT_VARIANT_ANALYSIS_USER_SETTINGS: VariantAnalysisUserSettings =
+  {
+    shouldShowViewAutofixesButton: false,
+  };
+
+/**
+ * Message indicating that the user's variant analysis configuration
+ * settings have changed.
+ */
+interface SetVariantAnalysisUserSettingsMsg {
+  t: "setVariantAnalysisUserSettings";
+  variantAnalysisUserSettings: VariantAnalysisUserSettings;
 }
 
 /**
@@ -563,7 +579,7 @@ export type ToVariantAnalysisMessage =
   | SetFilterSortStateMessage
   | SetRepoResultsMessage
   | SetRepoStatesMessage
-  | SetUserSettingsMsg;
+  | SetVariantAnalysisUserSettingsMsg;
 
 export type FromVariantAnalysisMessage =
   | CommonFromViewMessages
--- a/extensions/ql-vscode/src/common/interface-types.ts
+++ b/extensions/ql-vscode/src/common/interface-types.ts
@@ -527,6 +527,11 @@ interface OpenQueryTextMessage {
   t: "openQueryText";
 }
 
+interface ViewAutofixesMessage {
+  t: "viewAutofixes";
+  filterSort?: RepositoriesFilterSortStateWithIds;
+}
+
 interface CopyRepositoryListMessage {
   t: "copyRepositoryList";
   filterSort?: RepositoriesFilterSortStateWithIds;
@@ -561,6 +566,7 @@ export type FromVariantAnalysisMessage =
   | RequestRepositoryResultsMessage
   | OpenQueryFileMessage
   | OpenQueryTextMessage
+  | ViewAutofixesMessage
   | CopyRepositoryListMessage
   | ExportResultsMessage
   | OpenLogsMessage
--- a/extensions/ql-vscode/src/compare/compare-view.ts
+++ b/extensions/ql-vscode/src/compare/compare-view.ts
@@ -119,6 +119,8 @@ export class CompareView extends AbstractWebview<
       t: "setUserSettings",
       userSettings: {
         shouldShowProvenance: isCanary(),
+        // "View Autofixes" button is not supported in compare view
+        shouldShowViewAutofixesBtn: false,
       },
     });
 
--- a/extensions/ql-vscode/src/compare/compare-view.ts
+++ b/extensions/ql-vscode/src/compare/compare-view.ts
@@ -119,8 +119,6 @@ export class CompareView extends AbstractWebview<
       t: "setUserSettings",
       userSettings: {
         shouldShowProvenance: isCanary(),
-        // "View Autofixes" button is not supported in compare view
-        shouldShowViewAutofixesBtn: false,
       },
     });
 
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
@@ -135,6 +135,12 @@ export class VariantAnalysisView
       case "openQueryText":
         await this.manager.openQueryText(this.variantAnalysisId);
         break;
+      case "viewAutofixes":
+        await this.manager.viewAutofixes(
+          this.variantAnalysisId,
+          msg.filterSort,
+        );
+        break;
       case "copyRepositoryList":
         await this.manager.copyRepoListToClipboard(
           this.variantAnalysisId,
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
@@ -1,4 +1,5 @@
-import { ViewColumn } from "vscode";
+import type { ConfigurationChangeEvent } from "vscode";
+import { ViewColumn, workspace } from "vscode";
 import type { WebviewPanelConfig } from "../common/vscode/abstract-webview";
 import { AbstractWebview } from "../common/vscode/abstract-webview";
 import {
@@ -27,6 +28,7 @@ import type { App } from "../common/app";
 import {
   getVariantAnalysisDefaultResultsFilter,
   getVariantAnalysisDefaultResultsSort,
+  isCanary,
 } from "../config";
 
 export class VariantAnalysisView
@@ -46,6 +48,39 @@ export class VariantAnalysisView
     manager.registerView(this);
 
     this.dataFlowPathsView = new DataFlowPathsView(app);
+
+    // Set up configuration change listener
+    this.push(
+      workspace.onDidChangeConfiguration(
+        this.onConfigurationChanged.bind(this),
+      ),
+    );
+  }
+
+  /**
+   * Handler for configuration changes
+   */
+  private onConfigurationChanged(e: ConfigurationChangeEvent): void {
+    // Check if the canary setting has changed
+    if (e.affectsConfiguration("codeQL.canary")) {
+      void this.updateUserSettings();
+    }
+  }
+
+  private async updateUserSettings(): Promise<void> {
+    if (!this.isShowingPanel) {
+      return;
+    }
+
+    await this.postMessage({
+      t: "setUserSettings",
+      userSettings: {
+        // Provenance is not supported in variant analysis view
+        shouldShowProvenance: false,
+        // Only show "View Autofixes" button in canary mode.
+        shouldShowViewAutofixesBtn: isCanary(),
+      },
+    });
   }
 
   public async openView() {
@@ -53,6 +88,8 @@ export class VariantAnalysisView
     panel.reveal(undefined, true);
 
     await this.waitForPanelLoaded();
+
+    await this.updateUserSettings();
   }
 
   public async updateView(variantAnalysis: VariantAnalysis): Promise<void> {
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
@@ -49,29 +49,18 @@ export class VariantAnalysisView
     this.dataFlowPathsView = new DataFlowPathsView(app);
   }
 
-  private async updateUserSettings(): Promise<void> {
-    if (!this.isShowingPanel) {
-      return;
-    }
-
-    await this.postMessage({
-      t: "setUserSettings",
-      userSettings: {
-        // Provenance is not supported in variant analysis view
-        shouldShowProvenance: false,
-        // Only show "View Autofixes" button in canary mode.
-        shouldShowViewAutofixesBtn: isCanary(),
-      },
-    });
-  }
-
   public async openView() {
     const panel = await this.getPanel();
     panel.reveal(undefined, true);
 
     await this.waitForPanelLoaded();
 
-    await this.updateUserSettings();
+    await this.postMessage({
+      t: "setVariantAnalysisUserSettings",
+      variantAnalysisUserSettings: {
+        shouldShowViewAutofixesButton: isCanary(),
+      },
+    });
   }
 
   public async updateView(variantAnalysis: VariantAnalysis): Promise<void> {
--- a/extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
+++ b/extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
@@ -1,5 +1,4 @@
-import type { ConfigurationChangeEvent } from "vscode";
-import { ViewColumn, workspace } from "vscode";
+import { ViewColumn } from "vscode";
 import type { WebviewPanelConfig } from "../common/vscode/abstract-webview";
 import { AbstractWebview } from "../common/vscode/abstract-webview";
 import {
@@ -48,23 +47,6 @@ export class VariantAnalysisView
     manager.registerView(this);
 
     this.dataFlowPathsView = new DataFlowPathsView(app);
-
-    // Set up configuration change listener
-    this.push(
-      workspace.onDidChangeConfiguration(
-        this.onConfigurationChanged.bind(this),
-      ),
-    );
-  }
-
-  /**
-   * Handler for configuration changes
-   */
-  private onConfigurationChanged(e: ConfigurationChangeEvent): void {
-    // Check if the canary setting has changed
-    if (e.affectsConfiguration("codeQL.canary")) {
-      void this.updateUserSettings();
-    }
   }
 
   private async updateUserSettings(): Promise<void> {
--- a/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
+++ b/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
@@ -561,6 +561,9 @@ async function runAutofixForRepository(
     );
   }
 
+  // Format the output text file with markdown.
+  await formatWithMarkdown(outputTextFilePath, `${nwo}`);
+
   // Save output text files from each repo to later merge
   // into a single markdown file containing all results.
   outputTextFiles.push(outputTextFilePath);
@@ -747,6 +750,42 @@ async function mergeFiles(
     }
   } catch (error) {
     console.error("Error merging files:", error);
+    throw error;
+  }
+}
+
+/**
+ * Formats the given input file with the specified header.
+ * @param inputFile The path to the input file to format.
+ * @param header The header to include in the formatted output.
+ */
+async function formatWithMarkdown(
+  inputFile: string,
+  header: string,
+): Promise<void> {
+  try {
+    // Check if the input file exists
+    const exists = await pathExists(inputFile);
+    if (!exists) {
+      return;
+    }
+
+    // Read the input file content
+    const content = await readFile(inputFile, "utf8");
+
+    const frontFormatting: string =
+      "<details><summary>Fix suggestion details</summary>\n\n```diff\n";
+
+    const backFormatting: string =
+      "```\n\n</details>\n\n ### Notes\n - notes placeholder\n\n";
+
+    // Format the content with Markdown
+    const formattedContent = `## ${header}\n\n${frontFormatting}${content}${backFormatting}`;
+
+    // Write the formatted content back to the file
+    await writeFile(inputFile, formattedContent);
+  } catch (error) {
+    console.error("Error formatting file:", error);
     throw error;
   }
 }
--- a/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
+++ b/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
@@ -753,9 +753,45 @@ function execAutofix(
       if (showCommand) {
         void extLogger.log(`Spawning '${bin} ${args.join(" ")}' in ${cwd}`);
       }
-      const p = spawn(bin, args, { stdio: [0, 1, 2], ...options });
+
+      let stdoutBuffer = "";
+      let stderrBuffer = "";
+
+      const p = spawn(bin, args, {
+        stdio: ["ignore", "pipe", "pipe"],
+        ...options,
+      });
+
+      // Listen for stdout
+      p.stdout?.on("data", (data) => {
+        stdoutBuffer += data.toString();
+      });
+
+      // Listen for stderr
+      p.stderr?.on("data", (data) => {
+        stderrBuffer += data.toString();
+      });
+
+      // Listen for errors
       p.on("error", reject);
-      p.on("exit", (code) => (code === 0 ? resolve() : reject(code)));
+
+      // Listen for process exit
+      p.on("exit", (code) => {
+        // Log collected output
+        if (stdoutBuffer.trim()) {
+          void extLogger.log(`Autofix stdout:\n${stdoutBuffer.trim()}`);
+        }
+
+        if (stderrBuffer.trim()) {
+          void extLogger.log(`Autofix stderr:\n${stderrBuffer.trim()}`);
+        }
+
+        if (code === 0) {
+          resolve();
+        } else {
+          reject(new Error(`Autofix process exited with code ${code}.`));
+        }
+      });
     } catch (e) {
       reject(e);
     }
--- a/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
+++ b/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
@@ -8,6 +8,8 @@ import type { CodeQLCliServer } from "../codeql-cli/cli";
 import { pathExists } from "fs-extra";
 import { withProgress, progressUpdate } from "../common/vscode/progress";
 import type { ProgressCallback } from "../common/vscode/progress";
+import { join, dirname, parse } from "path";
+import { tryGetQueryMetadata } from "../codeql-cli/query-metadata";
 
 /**
  * Generates autofixes for the results of a variant analysis.
@@ -27,6 +29,16 @@ export async function viewAutofixesForVariantAnalysisResults(
       // Get the path to the local autofix installation.
       progress(progressUpdate(1, 4, `Checking for local autofix installation`));
       const localAutofixPath = findLocalAutofix();
+
+      // Get the variant analysis with the given id.
+      const variantAnalysis = variantAnalyses.get(variantAnalysisId);
+      if (!variantAnalysis) {
+        throw new Error(`No variant analysis with id: ${variantAnalysisId}`);
+      }
+
+      // Generate the query help and output it to the override directory.
+      progress(progressUpdate(2, 4, `Generating query help override`));
+      await overrideQueryHelp(variantAnalysis, cliServer, localAutofixPath);
 
       // TODO
     },
@@ -52,4 +64,69 @@ function findLocalAutofix(): string {
     throw new Error(`Local autofix path ${localAutofixPath} does not exist.`);
   }
   return localAutofixPath;
+}
+
+/**
+ * Overrides the query help from a given variant analysis
+ * at a location within the `localAutofixPath` directory .
+ */
+async function overrideQueryHelp(
+  variantAnalysis: VariantAnalysis,
+  cliServer: CodeQLCliServer,
+  localAutofixPath: string,
+): Promise<void> {
+  // Get path to the query used by the variant analysis.
+  const queryFilePath = variantAnalysis.query.filePath;
+  if (!(await pathExists(queryFilePath))) {
+    throw new Error(`Query file used by variant analysis not found.`);
+  }
+  const queryFilePathNoExt = join(
+    dirname(queryFilePath),
+    parse(queryFilePath).name,
+  );
+
+  // Get the path to the query help, which may be either a `.qhelp` or a `.md` file.
+  // Note: we assume that the name of the query file is the same as the name of the query help file.
+  const queryHelpFilePathQhelp = `${queryFilePathNoExt}.qhelp`;
+  const queryHelpFilePathMarkdown = `${queryFilePathNoExt}.md`;
+
+  // Set `queryHelpFilePath` to the existing extension type.
+  let queryHelpFilePath: string;
+  if (await pathExists(queryHelpFilePathQhelp)) {
+    queryHelpFilePath = queryHelpFilePathQhelp;
+  } else if (await pathExists(queryHelpFilePathMarkdown)) {
+    queryHelpFilePath = queryHelpFilePathMarkdown;
+  } else {
+    throw new Error(
+      `Could not find query help file at either ${queryHelpFilePathQhelp} or ${queryHelpFilePathMarkdown}. Check that the query help file exists and is named correctly.`,
+    );
+  }
+
+  // Get the query metadata.
+  const metadata = await tryGetQueryMetadata(cliServer, queryFilePath);
+  if (!metadata) {
+    throw new Error(`Could not get query metadata for ${queryFilePath}.`);
+  }
+  // Get the query ID (used for the overridden query help's filename).
+  const queryId = metadata.id;
+  if (!queryId) {
+    throw new Error(`Query metadata for ${queryFilePath} is missing an ID.`);
+  }
+  // Replace `/` with `-` for use with the overridden query help's filename.
+  // Use `replaceAll` since some query IDs have multiple slashes.
+  const queryIdWithDash = queryId.replaceAll("/", "-");
+
+  // Get the path to the output directory for overriding the query help.
+  // Note: the path to this directory may change in the future.
+  const queryHelpOverrideDirectory = join(
+    localAutofixPath,
+    "prompt-templates",
+    "qhelps",
+    `${queryIdWithDash}.md`,
+  );
+
+  await cliServer.generateQueryHelp(
+    queryHelpFilePath,
+    queryHelpOverrideDirectory,
+  );
 }
--- a/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
+++ b/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
@@ -14,28 +14,33 @@ import type { CodeQLCliServer } from "../codeql-cli/cli";
 import {
   pathExists,
   ensureDir,
-  readdir,
-  move,
   remove,
   unlink,
-  mkdtemp,
   readFile,
   writeFile,
+  createWriteStream,
 } from "fs-extra";
-import { withProgress, progressUpdate } from "../common/vscode/progress";
+import {
+  withProgress,
+  progressUpdate,
+  reportStreamProgress,
+} from "../common/vscode/progress";
 import type { ProgressCallback } from "../common/vscode/progress";
-import { join, dirname, parse } from "path";
+import { join, parse } from "path";
 import { tryGetQueryMetadata } from "../codeql-cli/query-metadata";
 import { pluralize } from "../common/word";
 import { readRepoTask } from "./repo-tasks-store";
-import { tmpdir } from "os";
+import { tmpDir } from "../tmp-dir";
 import { spawn } from "child_process";
 import type { execFileSync } from "child_process";
 import { tryOpenExternalFile } from "../common/vscode/external-files";
 import type { VariantAnalysisManager } from "./variant-analysis-manager";
 import type { VariantAnalysisResultsManager } from "./variant-analysis-results-manager";
-import { getAutofixPath, getAutofixModel } from "../config";
+import { getAutofixPath, getAutofixModel, downloadTimeout } from "../config";
 import { getErrorMessage } from "../common/helpers-pure";
+import { createTimeoutSignal } from "../common/fetch-stream";
+import { unzipToDirectoryConcurrently } from "../common/unzip-concurrently";
+import { reportUnzipProgress } from "../common/vscode/unzip-progress";
 
 // Limit to three repos when generating autofixes so not sending
 // too many requests to autofix. Since we only need to validate
@@ -308,7 +313,7 @@ async function processSelectedRepositories(
       withProgress(
         async (progressForRepo: ProgressCallback) => {
           // Get the sarif file.
-          progressForRepo(progressUpdate(1, 3, `Getting sarif`));
+          progressForRepo(progressUpdate(1, 3, `Getting sarif...`));
           const sarifFile = await getRepoSarifFile(
             variantAnalysisResultsManager,
             variantAnalysisIdStoragePath,
@@ -332,17 +337,17 @@ async function processSelectedRepositories(
 
           // Download the source root.
           // Using `0` as the progress step to force a dynamic vs static progress bar.
-          // Consider using `reportStreamProgress` as a future enhancement.
-          progressForRepo(progressUpdate(0, 3, `Downloading source root`));
+          progressForRepo(progressUpdate(0, 3, `Fetching source root...`));
           const srcRootPath = await downloadPublicCommitSource(
             nwo,
             repoTask.databaseCommitSha,
             sourceRootsStoragePath,
             credentials,
+            progressForRepo,
           );
 
           // Run autofix.
-          progressForRepo(progressUpdate(2, 3, `Running autofix`));
+          progressForRepo(progressUpdate(2, 3, `Running autofix...`));
           await runAutofixForRepository(
             nwo,
             sarifFile,
@@ -354,7 +359,7 @@ async function processSelectedRepositories(
           );
         },
         {
-          title: `Processing ${nwo}`,
+          title: `${nwo}`,
           cancellable: false,
         },
       ),
@@ -399,6 +404,7 @@ async function downloadPublicCommitSource(
   sha: string,
   outputPath: string,
   credentials: Credentials,
+  progressCallback: ProgressCallback,
 ): Promise<string> {
   const [owner, repo] = nwo.split("/");
   if (!owner || !repo) {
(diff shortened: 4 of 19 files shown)


extensions/ql-vscode/src/common/interface-types.ts
```diff
@@ -151,10 +151,13 @@ interface SetStateMsg {
 export interface UserSettings {
   /** Whether to display links to the dataflow models that generated particular nodes in a flow path. */
   shouldShowProvenance: boolean;
+  /** Whether to display the "View Autofixes" button. */
+  shouldShowViewAutofixesBtn: boolean;
```
jcogs33: Let me know if I should make `shouldShowViewAutofixesBtn` and `shouldShowProvenance` optional. It feels a bit ugly that I am setting these values for views where they are not relevant.

extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
```diff
@@ -46,13 +48,48 @@ export class VariantAnalysisView
...
+  }
+
+  /**
+   * Handler for configuration changes
+   */
+  private onConfigurationChanged(e: ConfigurationChangeEvent): void {
+    // Check if the canary setting has changed
+    if (e.affectsConfiguration("codeQL.canary")) {
+      void this.updateUserSettings();
+    }
```
jcogs33: This configuration listener code was suggested by Copilot. I'm wondering if I should be using a pre-existing config listener like `VariantAnalysisConfigListener` instead? Or not bother with a config listener at all and just require closing and re-opening the variant analysis panel for the canary setting update to take affect (which is how the `shouldShowProvenance` setting works)?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+
+/**
+ * Generates autofixes for the results of a variant analysis.
+ */
+export async function viewAutofixesForVariantAnalysisResults(
+  variantAnalysisId: number,
+  filterSort: RepositoriesFilterSortStateWithIds = defaultFilterSortState,
+  variantAnalyses: Map<number, VariantAnalysis>,
+  credentials: Credentials,
+  logger: NotificationLogger,
```
jcogs33: In retrospect, I'm not sure if I should be using `NotificationLogger`. 🤔 Let me know if I should use `extLogger` instead of or in addition to `NotificationLogger`?
Also let me know if I should add more logging in general?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+}
+
+/**
+ * Finds the local autofix installation path from the AUTOFIX_PATH environment variable.
+ * Throws an error if the path is not set or does not exist.
+ * @returns An object containing the local autofix path.
+ * @throws Error if the AUTOFIX_PATH environment variable is not set or the path does not exist.
+ */
+function findLocalAutofix(): string {
+  const localAutofixPath = process.env.AUTOFIX_PATH;
```
jcogs33: I'm currently requiring that users set this environment variable (see corresponding documentation). Let me know if there is something better I should do (config file, etc.) instead of using this environment variable.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+  if (!localAutofixPath) {
+    throw new Error("Path to local autofix installation not found.");
```
jcogs33: External users who have the canary enabled should run into this error if they try to view autofixes. Let me know if this error message should be different/clearer for external users.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+            throw new Error(`Missing database commit SHA for ${nwo}`);
+          }
+          if (!repoTask.resultCount) {
+            throw new Error(`Missing variant analysis result count for ${nwo}`);
+          }
+
+          // Download the source root.
+          // Using `0` as the progress step to force a dynamic vs static progress bar.
+          // Consider using `reportStreamProgress` as a future enhancement.
+          progressForRepo(progressUpdate(0, 3, `Downloading source root`));
```
jcogs33: If there is a better way to handle this (and other) progress updates, please let me know. 🙈

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+      `Expected to find exactly one \`*.sarif\` file for ${nwo}, but found ${sarifFiles.length}.`,
+    );
+  }
+  return sarifFiles[0];
+}
+
+/**
+ * Downloads the source code of a public commit from a GitHub repository.
+ */
+async function downloadPublicCommitSource(
```
jcogs33, line 413: Note: this function was mostly written by Copilot with DCA's `downloadPublicCommitSource` function as context.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+    args.push("--only-alert-number", alertNumber.toString());
+  }
+
+  return args;
+}
+
+/**
+ * Executes the autofix command.
+ */
+function execAutofix(
```
jcogs33, line 770: Note: this function is mostly the same as DCA's `execFollow` function which DCA uses to execute autofix.

extensions/ql-vscode/src/common/interface-types.ts
koesie10: I agree with you that it feels ugly to set values we don't need. I think it could make sense to create a separate message for the user settings for the variant analysis view so that you don't need to set the `shouldShowProvenance` setting for the variant analysis view, and don't need to set `shouldShowViewAutofixesBtn` for the other views.

I also think we don't need to shorten `Button` to `Btn`, so `shouldShowViewAutofixesButton` is fine.

extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
koesie10: I think in general we don't bother with a config listener for the canary setting, but if you do want to add one, I'd recommend creating a new config listener in `config.ts` that listens to the `canary` setting and using that here instead of subscribing to the VS Code event inline.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisActions.tsx
```diff
@@ -9,13 +10,17 @@ export type VariantAnalysisActionsProps = {
...
   onCopyRepositoryListClick: () => void;
   onExportResultsClick: () => void;
+  viewAutofixesDisabled?: boolean;
   copyRepositoryListDisabled?: boolean;
   exportResultsDisabled?: boolean;
 
   hasSelectedRepositories?: boolean;
   hasFilteredRepositories?: boolean;
+
+  userSettings: UserSettings;
```
koesie10: Could this just be something like `showViewAutofixesButton` to make this slightly easier to test (since you won't need to construct the full user settings)?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
koesie10: If you aren't showing notifications, I think you can use the `BaseLogger` type here.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+  credentials: Credentials,
+  logger: NotificationLogger,
+  storagePath: string,
+  app: App,
+  cliServer: CodeQLCliServer,
+): Promise<void> {
+  await withProgress(
+    async (progress: ProgressCallback) => {
+      // Get the path to the local autofix installation.
+      progress(progressUpdate(1, 4, `Checking for local autofix installation`));
```
koesie10: Since the `findLocalAutofix` function should be very fast, does this actually show up in the progress window?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+ * Throws an error if the path is not set or does not exist.
+ * @returns An object containing the local autofix path.
+ * @throws Error if the AUTOFIX_PATH environment variable is not set or the path does not exist.
+ */
+function findLocalAutofix(): string {
+  const localAutofixPath = process.env.AUTOFIX_PATH;
+  if (!localAutofixPath) {
+    throw new Error("Path to local autofix installation not found.");
+  }
+  if (!pathExists(localAutofixPath)) {
```
koesie10: The `pathExists` function returns a promise, so this should be:


```suggestion
  if (!(await pathExists(localAutofixPath))) {
```

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
koesie10: I think using a config setting would make sense (something like `codeQL.autofix.path`). I'm not sure how VS Code picks up environment variables, but it's a lot easier to set config settings than environment variables.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+  localAutofixPath: string,
+): Promise<void> {
+  // Get path to the query used by the variant analysis.
+  const queryFilePath = variantAnalysis.query.filePath;
+  if (!(await pathExists(queryFilePath))) {
+    throw new Error(`Query file used by variant analysis not found.`);
+  }
+  const queryFilePathNoExt = join(
+    dirname(queryFilePath),
+    parse(queryFilePath).name,
```
koesie10: It seems like `parse` also returns a `dir`, so to avoid calling multiple `path` functions, could we store the result of `parse` in a variable and then do `join(parsedQueryFilePath.dir, parsedQueryFilePath.name)`?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+async function getStoragePaths(
+  variantAnalysisId: number,
+  storagePath: string,
+): Promise<{
+  variantAnalysisIdStoragePath: string;
+  sourceRootsStoragePath: string;
+  autofixOutputStoragePath: string;
+}> {
+  // Confirm storage path for the variant analysis ID exists.
+  const variantAnalysisIdStoragePath = join(
```
koesie10: Can we use `variantAnalysisManager.getVariantAnalysisStorageLocation` to ensure it always uses the same path as the variant analysis manager?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+ * Gets the path to a SARIF file in a given `repoStoragePath`.
+ */
+async function getSarifFile(
+  repoStoragePath: string,
+  nwo: string,
+): Promise<string> {
+  // Get results directory path.
+  const repoResultsStoragePath = join(repoStoragePath, "results");
+  // Find sarif file.
+  const sarifFiles = await glob(`${repoResultsStoragePath}/**/*.sarif`);
```
koesie10: I believe this should always live in `results/results.sarif`. However, can we add a method to `VariantAnalysisManager`/`VariantAnalysisResultsManager` that returns this path since those should be responsible for the result storage?

Perhaps it could even make sense to first call `variantAnalysisResultsManager.isVariantAnalysisRepoDownloaded` to ensure the SARIF file is downloaded so you can give a better error message.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+        async (progressForRepo: ProgressCallback) => {
+          // Get the sarif file.
+          progressForRepo(progressUpdate(1, 3, `Getting sarif`));
+          const repoStoragePath = join(variantAnalysisIdStoragePath, nwo);
+          const sarifFile = await getSarifFile(repoStoragePath, nwo);
+
+          // Read the contents of the variant analysis' `repo_task.json` file,
+          // and confirm that the `databaseCommitSha` and `resultCount` exist.
+          const repoTask: VariantAnalysisRepositoryTask =
+            await readRepoTask(repoStoragePath);
```
koesie10: Same here, could we add a method to the `VariantAnalysisManager`/`VariantAnalysisResultsManager` to ensure we aren't duplicating logic in regards to where the file lives etc.?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+        "Accept: application/octet-stream",
+        "--user-agent",
+        "GitHub-CodeQL-Extension",
+        "-L", // Follow redirects
+        "-o",
+        tarballPath,
+        url,
+      ];
+
+      const process = spawn("curl", curlArgs, { cwd: downloadDir });
```
koesie10: Can we use the `fetch` API instead of spawning curl and tar processes? For a good example, you can probably take a look at [the `download` method](https://github.com/github/vscode-codeql/blob/f184b213d36d224b4477d82f972b6acd171ccb0c/extensions/ql-vscode/src/variant-analysis/variant-analysis-results-manager.ts#L72-L131) for the variant analysis results manager or [the `downloadDistribution` method](https://github.com/github/vscode-codeql/blob/afdef0029e5fc32af713f6e1a381fc586ae54818/extensions/ql-vscode/src/codeql-cli/distribution.ts#L499-L563) for downloading the CLI which includes a progress indicator. I believe both extract a ZIP file, so you might need to switch to `downloadZipballArchive` to use the same unzip functionality.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+
+    // Move the extracted source to the final location
+    await move(extractedSourcePath, checkoutDir);
+
+    // Clean up the temporary directory
+    await remove(downloadDir);
+
+    return checkoutDir;
+  } catch (error) {
+    const errorMessage = error instanceof Error ? error.message : String(error);
```
koesie10: You should be able to use `getErrorMessage` for this.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+      `Source for ${nwo} at ${sha} already exists at ${checkoutDir}.`,
+    );
+    return checkoutDir;
+  }
+
+  void logger.log(`Fetching source of repository ${nwo} at ${sha}...`);
+
+  try {
+    // Create a temporary directory for downloading
+    const downloadDir = await mkdtemp(join(tmpdir(), "download-source-"));
```
koesie10: I'd recommend using the `tmp-promise` package for this since that also includes cleanup.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+import { pathExists, ensureDir, readdir, move, remove } from "fs-extra";
+import { withProgress, progressUpdate } from "../common/vscode/progress";
+import type { ProgressCallback } from "../common/vscode/progress";
+import { join, dirname, parse } from "path";
+import { tryGetQueryMetadata } from "../codeql-cli/query-metadata";
+import { window as Window } from "vscode";
+import { pluralize } from "../common/word";
+import { glob } from "glob";
+import { readRepoTask } from "./repo-tasks-store";
+import { unlink, mkdtemp, readFile, writeFile } from "fs/promises";
```
koesie10: We generally use `fs-extras` instead of `fs/promises`.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+
+    // Write merged content
+    await writeFile(outputFile, contents.join("\n"));
+
+    // Delete original files
+    if (deleteOriginalFiles) {
+      await Promise.all(inputFiles.map((file) => unlink(file)));
+    }
+  } catch (error) {
+    console.error("Error merging files:", error);
```
koesie10: I don't think `console.error` shows up in the VS Code logs, if we want to show this we should use the logger.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+    const backFormatting: string =
+      "```\n\n</details>\n\n ### Notes\n - notes placeholder\n\n";
+
+    // Format the content with Markdown
+    const formattedContent = `## ${header}\n\n${frontFormatting}${content}${backFormatting}`;
+
+    // Write the formatted content back to the file
+    await writeFile(inputFile, formattedContent);
+  } catch (error) {
+    console.error("Error formatting file:", error);
```
koesie10: Same here, I think `console.error` messages do not show up.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+  alertNumber?: number, // Optional parameter for specific alert
+): Promise<void> {
+  // Set up args for autofix command.
+  const fixedAutofixArgs = [
+    "--sarif",
+    sarifFile,
+    "--source-root",
+    srcRootPath,
+    "--model",
+    "capi-dev-4o", // may fail with older versions of cocofix
```
koesie10: Should this be configurable (something like `codeQL.autofix.model`)?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+    "--dev",
+    "--no-cache",
+    "--format",
+    "text",
+    "--diff-style",
+    "diff", // could do "text" instead if want line of "=" between fixes
+  ];
+  const varAutofixArgs = createVarAutofixArgs(
```
koesie10: I don't really think we need a separate function for this, it would be fine to just add this to the args in this function.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+  await execAutofix(
+    logger,
+    cocofixBin,
+    autofixArgs,
+    {
+      cwd: workDir,
+      env: {
+        CAPI_DEV_KEY: process.env.CAPI_DEV_KEY,
+        PATH: process.env.PATH,
+      },
```
koesie10, line 761: Would it make sense to allow specifying the environment variables in the config (in something like `codeQL.autofix.env`)?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+  options: Parameters<typeof execFileSync>[2],
+  showCommand?: boolean,
+): Promise<void> {
+  return new Promise((resolve, reject) => {
+    try {
+      const cwd = options?.cwd || process.cwd();
+      if (showCommand) {
+        void logger.log(`Spawning '${bin} ${args.join(" ")}' in ${cwd}`);
+      }
+      if (args.some((a) => a === undefined || a === "")) {
```
koesie10: I don't think we need this check, it's fine for the command to fail if this is the case.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+        throw new Error(
+          `Invalid empty or undefined arguments: ${args.join(" ")}`,
+        );
+      }
+      const p = spawn(bin, args, { stdio: [0, 1, 2], ...options });
```
koesie10: I think `stdio: [0, 1, 2]` will ensure that the autofix CLI logs to the stdout/stderr of VS Code, but I don't think you can view those. I think it would make sense to log those to the extension logs. A good example of this can be found in [`runCodeQlCliInNewProcess`](https://github.com/github/vscode-codeql/blob/49839a1a529a0092ed37c1331edac95c59627f65/extensions/ql-vscode/src/codeql-cli/cli.ts#L443), specifically in [`handleProcessOutput`](https://github.com/github/vscode-codeql/blob/49839a1a529a0092ed37c1331edac95c59627f65/extensions/ql-vscode/src/codeql-cli/cli.ts#L519).

koesie10 (commented):
Thanks! I've not tested this locally, but the demo you shared looks good!

extensions/ql-vscode/src/compare/compare-view.ts
```diff
@@ -119,6 +119,8 @@ export class CompareView extends AbstractWebview<
       t: "setUserSettings",
       userSettings: {
         shouldShowProvenance: isCanary(),
```
aeisenberg: Do we support provenance in the compare view either? Should this also be set to `false` always? (not needed for this PR)

Also, below.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
aeisenberg: I don't think we can distinguish internal and external users (except for the canary flag).

It might be good to expand on this error message. Maybe something like this:

```suggestion
    throw new Error("Path to local autofix installation not found. Internal GitHub access required.");
```

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+  }
+  // Replace `/` with `-` for use with the overridden query help's filename.
+  // Use `replaceAll` since some query IDs have multiple slashes.
+  const queryIdWithDash = queryId.replaceAll("/", "-");
+
+  // Get the path to the output directory for overriding the query help.
+  // Note: the path to this directory may change in the future.
+  const queryHelpOverrideDirectory = join(
+    localAutofixPath,
+    "prompt-templates",
```
aeisenberg, line 217: Does `localAutofixPath` need to be a checkout of the autofix repo? Or does it point to a binary?

The new location for the qhelp overrides is here: `go/pkg/autofix/prompt/qhelps`.  But, this is only available in a full checkout of the repo. It's not available with the go binary. The cocofix prompt-templates will go away.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
aeisenberg: Not sure.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+          ? resolve()
+          : reject(new Error(`curl exited with code ${code}`)),
+      );
+    });
+
+    void logger.log(`Download complete, extracting source...`);
+
+    // Extract the tarball
+    await new Promise<void>((resolve, reject) => {
+      const process = spawn("tar", ["-xzf", tarballPath], { cwd: downloadDir });
```
aeisenberg: We're using `tar-stream` elsewhere for extracting tarballs. It would be better to use something like that instead of spawning a new process.

Just be aware that `tar-stream` is currently a dev dependency and would need to be moved to a dependency.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+  // Get storage paths for the autofix results for this repository.
+  const {
+    repoAutofixOutputStoragePath,
+    outputTextFilePath,
+    transcriptFilePath,
+    fixDescriptionFilePath,
+  } = await getRepoStoragePaths(autofixOutputStoragePath, nwo);
+
+  // Get autofix binary.
+  // Switch to Go binary in the future and have user pass full path
```
aeisenberg, line 610: Yes.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
aeisenberg: Yes. gpt 4o is going away soon. And we are trying various models to replace it.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,801 @@
...
+    const content = await readFile(inputFile, "utf8");
+
+    const frontFormatting: string =
+      "<details><summary>Fix suggestion details</summary>\n\n```diff\n";
+
+    const backFormatting: string =
+      "```\n\n</details>\n\n ### Notes\n - notes placeholder\n\n";
+
+    // Format the content with Markdown
+    const formattedContent = `## ${header}\n\n${frontFormatting}${content}${backFormatting}`;
```
aeisenberg: If `content` itself has three backticks, then this formatting will be broken. I don't know how likely that is. In this case, probably just doing a find/replace of three ` with \`\`\`.

extensions/ql-vscode/src/variant-analysis/variant-analysis-view.ts
jcogs33: > I think in general we don't bother with a config listener for the canary setting

In that case I'll just remove this code for simplicity.

Removed in https://github.com/github/vscode-codeql/pull/4065/commits/4e089e73545b00a402c0b9fdb2f06992b8ff9266.

extensions/ql-vscode/src/view/variant-analysis/VariantAnalysisActions.tsx
jcogs33: Updated in https://github.com/github/vscode-codeql/pull/4065/commits/04a858affa7fc4b6fe49362b8b7bfa4895fa3794.

extensions/ql-vscode/src/common/interface-types.ts
jcogs33: I created a separate message for the user settings for the variant analysis view in https://github.com/github/vscode-codeql/pull/4065/commits/04a858affa7fc4b6fe49362b8b7bfa4895fa3794. Let me know if I didn't interpret what you had in mind correctly.

> I also think we don't need to shorten Button to Btn, so shouldShowViewAutofixesButton is fine.

👍

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Good point. I don't think I've seen it show up. Same for the `overrideQueryHelp` progress update. I've removed them in https://github.com/github/vscode-codeql/pull/4065/commits/7d3e403e0b2b158fa4e8732123b58600b360f295.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Fixed in https://github.com/github/vscode-codeql/pull/4065/commits/161c5de55e8815bef6263272f25ba0a1c0aeb79d.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Updated in https://github.com/github/vscode-codeql/pull/4065/commits/e23699f54b784d08029a38b659515741133c6987.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Updated in https://github.com/github/vscode-codeql/pull/4065/commits/22d642945fa16a55c34e6f456493ecc067a68b19.

extensions/ql-vscode/src/compare/compare-view.ts
jcogs33: It's not currently supported, but there's an [issue](https://github.com/github/vscode-codeql/issues/3739) to eventually support it.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Yes, it needs to be a checkout of the autofix repo. Will that be a problem for any reason? 
I've included instructions about checking out the repo in the documentation for this feature (see documentation links beneath my original commits). Note that DCA also expects a checkout of the autofix repo when running autofix locally with DCA, and I mimicked that expectation.

I'll update the prompt-template directory when I do a follow-up PR to update from cocofix to Go.

jcogs33:
Thanks for the reviews @koesie10 and @aeisenberg! :bow:
I've addressed some of your comments. I'll re-request your reviews once I'm finished with them all.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Update in https://github.com/github/vscode-codeql/pull/4065/commits/1d203a4bd20c7878355fccc7f54307e4f16b32da.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Updated in https://github.com/github/vscode-codeql/pull/4065/commits/1d203a4bd20c7878355fccc7f54307e4f16b32da.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Used the existing `variantAnalysisResultsManager.getRepoStorageDirectory` method. Updated in https://github.com/github/vscode-codeql/pull/4065/commits/1d203a4bd20c7878355fccc7f54307e4f16b32da.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Updated in https://github.com/github/vscode-codeql/pull/4065/commits/561cbc4b37f6a41fbf94b58e1899653bc25654e8.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Updated in https://github.com/github/vscode-codeql/pull/4065/commits/561cbc4b37f6a41fbf94b58e1899653bc25654e8.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
aeisenberg: > Will that be a problem for any reason?

I don't think it's a fundamental problem. It just puts a bit of extra burden on the users, but I think it's fine for now.

If you wanted to be really snazzy, you could automatically try to download the autofix binary from as a github package or release. That's a big lift, though. Then the extension would need to do autoupdates and manage the state of the binary (exactly how it works for the codeql cli right now, and that's quite complicated). So, what you have now is fine.

aeisenberg (commented):
There's one more comment from @koesie10 that should be addressed, but after that I think it's good to go.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Updated in https://github.com/github/vscode-codeql/pull/4065/commits/d887ca505adf9a5177645c33cfc887a717e7452b.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Updated in https://github.com/github/vscode-codeql/commit/d887ca505adf9a5177645c33cfc887a717e7452b. Throwing an error instead of using the logger.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Updated in https://github.com/github/vscode-codeql/commit/d887ca505adf9a5177645c33cfc887a717e7452b. Throwing an error instead of using the logger.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Removed in https://github.com/github/vscode-codeql/pull/4065/commits/bbc741319520cb57c6ddb64ab9109ddabc5d9eb4.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: I doubt it's likely, but I added a replace for three backticks in https://github.com/github/vscode-codeql/pull/4065/commits/b7f84fca9ef88061279dd5ab6001b6006b6ecd0e.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Switched to fs-extras in https://github.com/github/vscode-codeql/pull/4065/commits/41a3612f59a1e70d99a57bfd9129d1bf33b6ae65. (I shouldn't have let Copilot handle these imports apparently 🙁)

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: I ended up switching to `extLogger` in https://github.com/github/vscode-codeql/pull/4065/commits/492b805e6677df45ba7d6b8d1e75100047e8bdc8. Let me know if you disagree with that choice. It seemed simpler than passing around a `logger` variable.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: I'm not sure I'm understanding how to specify these in the config. i.e. Are you suggesting specifying the API key for `CAPI_DEV_KEY` directly in the settings.json file (is that safe)? Or specifying the path to a local file where the API key is stored? Or something else?

jcogs33:
> There's one more comment from `@koesie10` that should be addressed

@aeisenberg Which comment are you referring to? There were four things I still was planning to address: three comments related to editing parts of `downloadPublicCommitSource`, and one comment about getting the autofix CLI logs into the extension logs.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
aeisenberg: I don't think we want to add the secret as a config setting. There are ways of adding secrets securely, but I think just keeping it as an env var is fine. Much simpler. Just please be sure to document.

There are some complications if you are using a codespace. I'm not sure how it would work. Can you inject env vars into it?

As for the `PATH` variable. I'm not sure if it makes sense either.

aeisenberg:
Oh...there might be more things that I missed. I'm out tomorrow. I think it's also a holiday in the US, so I can take another look on Monday.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
koesie10: I agree that we probably don't want to add the secret in the config file, but allowing you to set custom environment variables might still be useful, for example an `ENVIRONMENT` environment variable. We could also just pass through all of `process.env` to set custom environment variables though.

I'm not sure exactly how VS Code picks up environment variables, but it seems like it's possible to specify secrets when using Codespaces: [Using secrets](https://docs.github.com/en/enterprise-cloud@latest/codespaces/managing-your-codespaces/managing-your-account-specific-secrets-for-github-codespaces).

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Done in https://github.com/github/vscode-codeql/pull/4065/commits/561cd6c34b8d74056e201c19f360b8904ffe78bf.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Done in https://github.com/github/vscode-codeql/pull/4065/commits/561cd6c34b8d74056e201c19f360b8904ffe78bf.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Switched to using a zipball and existing unzip functionality in https://github.com/github/vscode-codeql/pull/4065/commits/561cd6c34b8d74056e201c19f360b8904ffe78bf, as [suggested by `@koesie10`](https://github.com/github/vscode-codeql/pull/4065#discussion_r2179443466).

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Updated in https://github.com/github/vscode-codeql/pull/4065/commits/8ed9e6a2be0b4bd80af25a1e5ed9b05d15ad4c8e. I may have oversimplified the logic 🙈, so let me know if you want any changes. I've confirmed that stderr/stdout from the autofix process are appearing in the extension logs now.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: > I think just keeping it as an env var is fine. Much simpler. Just please be sure to document.

Sounds good. I've documented it.

> There are some complications if you are using a codespace. I'm not sure how it would work. Can you inject env vars into it?

I don't use codespaces, so I'm not sure. 🤔 I can look into this further if needed?

>  but allowing you to set custom environment variables might still be useful, for example an `ENVIRONMENT` environment variable. We could also just pass through all of `process.env` to set custom environment variables though.

@koesie10 Sorry, I'm still not picturing what you have in mind. Can you elaborate on what you mean by this? i.e. What would the `ENVIRONMENT` environment variable be set to?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,886 @@
...
+  } else if (await pathExists(queryHelpFilePathMarkdown)) {
+    queryHelpFilePath = queryHelpFilePathMarkdown;
+  } else {
+    throw new Error(
+      `Could not find query help file at either ${queryHelpFilePathQhelp} or ${queryHelpFilePathMarkdown}. Check that the query help file exists and is named correctly.`,
+    );
+  }
+
+  // Get the query metadata.
+  const metadata = await tryGetQueryMetadata(cliServer, queryFilePath);
```
koesie10: Since the metadata is required, would it make sense to just call `cliServer.resolveMetadata` directly and let it throw an error if the metadata is not available? It seems like the `tryGetQueryMetadata` was created for cases where the query metadata is optional.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,886 @@
...
+    ?.filter((a) => a.resultCount && a.resultCount > 0)
+    .map((a) => a.repository.fullName);
+  if (!fullNames || fullNames.length === 0) {
+    throw new Error("No repositories with results found.");
+  }
+
+  // Limit to MAX_NUM_REPOS by slicing the array.
+  if (fullNames.length > MAX_NUM_REPOS) {
+    fullNames = fullNames.slice(0, MAX_NUM_REPOS);
+    void extLogger.log(
```
koesie10: Should this perhaps be a `window.showWarningMessage` call so that the user is notified more directly?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,886 @@
...
+
+    void extLogger.log(`Download complete, extracting source...`);
+
+    // Extract the downloaded zip file
+    await unzipToDirectoryConcurrently(
+      archivePath,
+      outputPath,
+      progressCallback
+        ? reportUnzipProgress(`Unzipping source root...`, progressCallback)
+        : undefined,
```
koesie10: I think the `progressCallback` is always defined here:

```suggestion
      reportUnzipProgress(`Unzipping source root...`, progressCallback),
```

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,886 @@
...
+  const [owner, repo] = nwo.split("/");
+  if (!owner || !repo) {
+    throw new Error(`Invalid repository name: ${nwo}`);
+  }
+
+  // Create output directory if it doesn't exist
+  await ensureDir(outputPath);
+
+  // Define the final checkout directory
+  const checkoutDir = join(
```
koesie10, line 429: I think I'm missing something, but where do we define that the archive is extracted to this directory?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,886 @@
...
+      outputPath,
+      progressCallback
+        ? reportUnzipProgress(`Unzipping source root...`, progressCallback)
+        : undefined,
+    );
+    await remove(archivePath);
+
+    return checkoutDir;
+  } catch (error) {
+    throw new Error(
```
koesie10, line 583: Should we try to remove the `checkoutDir` when something in this method fails to ensure it's re-downloaded on the next try?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,886 @@
...
+    repoAutofixOutputStoragePath,
+    outputTextFilePath,
+    transcriptFilePath,
+    fixDescriptionFilePath,
+  } = await getRepoStoragePaths(autofixOutputStoragePath, nwo);
+
+  // Get autofix binary.
+  // Switch to Go binary in the future and have user pass full path
+  // in an environment variable instead of hardcoding part here.
+  const cocofixBin = join(process.cwd(), localAutofixPath, "bin", "cocofix.js");
```
koesie10, line 612: If the Go binary accepts the same parameters as cocofix, would it makes sense to already add a config setting for this and use a default of `bin/cocofix.js`?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,886 @@
...
+ * Throws an error if the path is not set or does not exist.
+ * @returns An object containing the local autofix path.
+ * @throws Error if the `codeQL.autofix.path` setting is not set or the path does not exist.
+ */
+async function findLocalAutofix(): Promise<string> {
+  const localAutofixPath = getAutofixPath();
+  if (!localAutofixPath) {
+    throw new Error(
+      "Path to local autofix installation not found. Internal GitHub access required.",
+    );
```
koesie10: Would it be useful to add the name of the config setting here (either hard-coded or using `AUTOFIX_PATH.qualifiedName`)?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -0,0 +1,886 @@
...
+  outputTextFilePath: string,
+  fixDescriptionFilePath: string,
+  transcriptFilePath: string,
+  workDir: string,
+  alertNumber?: number, // Optional parameter for specific alert
+): Promise<void> {
+  // Get autofix model from user settings.
+  const autofixModel = getAutofixModel();
+  if (!autofixModel) {
+    throw new Error("Autofix model not found.");
```
koesie10: Same here, would it be useful to add the name of the config setting?

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
koesie10: I don't think this is really necessary if it's unlikely that this will be used. I haven't run autofix myself locally, so I'm not sure if it depends on any environment variables.

koesie10 (approved):
Thanks, this looks good to me. I have some optional comments, but I don't think any of those are necessary.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -395,91 +422,143 @@ async function downloadPublicCommitSource(
...
+
+        await new Promise((resolve, reject) => {
+          archiveFileStream.write(value, (err) => {
+            if (err) {
+              reject(err);
+            }
+            resolve(undefined);
+          });
+        });
+      }
```
aeisenberg: This seems overly complicated. Can you just pipe the body stream into the output stream?

Something like this:

```
body.getReader().pipe(archiveFileStream)
```
You can still attach event handlers to the readable stream to report progress. Eg- https://nodejs.org/api/stream.html#event-data.

aeisenberg (commented):
Looks generally good. I see one area where I think you can simplify things.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Done in https://github.com/github/vscode-codeql/pull/4065/commits/7f9d7ae9e173e717619d0895ba3ccc4fc338eae5.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: I've gone back-and-forth on whether to display notifications to the user for the repo/alert limits since doing so somewhat clutters the progress updates, but I think I like it for the repo limit case. Added in https://github.com/github/vscode-codeql/pull/4065/commits/4e76a0dd8a7e0b879b7ccde78438aaf626c7b944.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Ah, thanks, copy-paste issue. Updated in https://github.com/github/vscode-codeql/pull/4065/commits/0e6aa813734acc575b118799fde32794325fddfd.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Ah, thanks for catching this. This got affected by my refactor. The `checkoutDir` here ends up being the same directory name that `unzipToDirConcurrently` extracts to so it was still working, but it would break if the extracted directory name ever changed to a different pattern. Fixed in https://github.com/github/vscode-codeql/pull/4065/commits/cb4bfbc517accded65ee458452cec61257a5ce0a.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Done in https://github.com/github/vscode-codeql/pull/4065/commits/6333c9e70d7ee8c78bf510586773d4b1ae09e6e2.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: I need to experiment with the Go binary manually. I'm not sure yet if it accepts all of the same parameters. I'll likely edit this into a config setting when I create a follow-up PR for switching to Go.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Done in https://github.com/github/vscode-codeql/pull/4065/commits/bf6d8e72b871f3068c1c89465d2ad527389d90de.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: Done in https://github.com/github/vscode-codeql/commit/bf6d8e72b871f3068c1c89465d2ad527389d90de.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: I'll leave as-is for now then, and maybe re-visit this idea in my follow-up PR for Go.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: This code mimics what is done elsewhere in the codebase, but simplifying sounds good to me. I _think_ I've done what you're suggesting in https://github.com/github/vscode-codeql/pull/4065/commits/2c04019532395b9893ddc818bd4e4327c7e74dcc (and fixed in https://github.com/github/vscode-codeql/pull/4065/commits/1f5cc9084a643d7259b669141c9319f5f3c71180).

jcogs33:
@koesie10 @aeisenberg I've responded to all new review comments. Thanks for the thorough reviews! 🙇

aeisenberg:
I'm fine with that.

Issue #4059, opened 2025-06-26
CodeQL Extension submits empty block to multi-repository variant analysis (MRVA)


mrigankpawagi:
**Describe the bug**
When I select a query for MRVA, the extension submits an empty bundle. Every file that should have been included in the bundle, like my query, the qlpack.yml file, and the files related to all dependencies, get "ignored" because they "are outside the pack root" &mdash; as logged by the extension when it runs `codeql pack bundle`. As a result, and as one would expect, the MRVA jobs fail.

I have tried this with the example queries (no matter the language) in the [vscode-codeql-starter](https://github.com/github/vscode-codeql-starter) repository too (after launching VSCode from their workspace) but even that gave me the same issue.

Any help will be highly appreciated.

**Version**
CodeQL CLI: 2.22.1
CodeQL VSCode Extension version: 1.17.3
VSCode version: 1.101.2

koesie10:
@mrigankpawagi Thank you, this is very useful information! Does the `USERNAME` in your logs show up as two different strings, or are they all the same string? It seems like [Windows short paths](https://en.wikipedia.org/wiki/8.3_filename) might be enabled on your system and we aren't resolving that correctly. If they do all show up as the same string, do they contain a `~` character or are they your full username?

mrigankpawagi:
@koesie10 I have updated the logs [at the same link](https://gist.github.com/mrigankpawagi/546b605d18ba629bdf33c18d2684ea9b) with the usernames.

koesie10:
@mrigankpawagi Thanks for the updated logs! Those confirm that the issue is with 8.3 short filenames. I'm trying to investigate whether there's a solution we can use for this, and it seems like we should be able to call the Windows function [`GetLongPathNameW`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getlongpathnamew) from the extension to determine the full path name, which should fix the issue. I'll test whether this works and let you know.

koesie10:
[...]

PR #4068, opened 2025-07-02
Add expansion of short paths using native Windows call

koesie10:
- Closes https://github.com/github/vscode-codeql/issues/4059

It seems like in certain cases we cannot resolve Windows short paths and fail with an error message in the CodeQL CLI. As a fallback mechanism, this adds a native call to [the `GetLongPathNameW` win32 API function](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getlongpathnamew), which will correctly resolve the path if it exists. To reduce the cost of calling this function, we are now also only expanding the short paths once instead of twice per variant analysis run.

To call this function, we're using [koffi](https://koffi.dev/). koffi includes a native Node.js extension for calling into the C functions and bundles this extension for a lot of platforms by default:

Screenshot 2025-07-02 at 14 30 16

This would increase the size of our extension by tens of megabytes (all native extensions together are about 66 MB), so I've worked around this by only bundling Windows x64 by default. The call is only made on Windows, and I expect x64 to be the only common Windows platform. In total, this adds about 2.2MB to the extension.

It's been tested by the user that reported the issue and this seems to have fixed the issue (https://github.com/github/vscode-codeql/issues/4059#issuecomment-3027646631), and I've also tested it on one of my own Windows machines where it resolves the path correctly (although the original workaround also works there, so I've artificially made it use the native method).


LWSimpkins:
I was investigating this issue for the forked version we use and was going to submit a PR, but this was already in progress!

For some background about what caused this:
- There was a bug in Node.js. More specifically, in https://github.com/libuv/libuv which implements the lstats functionality
  - I believe it was introduced in this PR: https://github.com/libuv/libuv/pull/4327 in `v1.49.1` which was used starting in Node `v22.12.0`
- It was fixed in Node.js `22.17.0` (latest 2.x) which upgraded libuv to `1.51.0` and I think was fixed by these PRs
  - https://github.com/libuv/libuv/pull/4696
  - https://github.com/libuv/libuv/pull/4698
- VS Code upgraded the bundled version of Node.js from `20.19.0` to `22.15.1` in their 2025 May release (1.101.x) on June 12th, so they are using a version with the bug
  - https://github.com/microsoft/vscode/releases/tag/1.101.0


The bug does not seem to affect directories with spaces in the name, only ones without spaces. That's why the unit test for `C:\\PROGRA~1` / `C:\\Program Files` did not catch this in my local testing. Example from local testing with `C:\\Users\\LSIMPK~1` with `22.15.1` on my `PATH`:

stats:
|Directory|stats.dev|stats.ino|
|--|--|--|
|C:\\Users\\LSIMPK~1| 0n|6473924466378324n|

childStats:
|Directory|stats.dev|stats.ino|
|--|--|--|
|C:\\Users\\All Users|3705393598n|3096224745883570n|
|C:\\Users\\Default|0n|2251799815715884n|
|C:\\Users\\Default User|3705393598n|3659174699303827n|
|C:\\Users\\desktop.ini|0n|2251799815738381n|
|C:\\Users\\lsimpkins|0n|6473924466378324n|

charisk (approved):
LGTM

fluffy123dizzy-max:
> * Closes [CodeQL Extension submits empty block to multi-repository variant analysis (MRVA) #4059](https://github.com/github/vscode-codeql/issues/4059)
> 
> It seems like in certain cases we cannot resolve Windows short paths and fail with an error message in the CodeQL CLI. As a fallback mechanism, this adds a native call to [the `GetLongPathNameW` win32 API function](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getlongpathnamew), which will correctly resolve the path if it exists. To reduce the cost of calling this function, we are now also only expanding the short paths once instead of twice per variant analysis run.
> 
> To call this function, we're using [koffi](https://koffi.dev/). koffi includes a native Node.js extension for calling into the C functions and bundles this extension for a lot of platforms by default:
> 
> Screenshot 2025-07-02 at 14 30 16
> 
> This would increase the size of our extension by tens of megabytes (all native extensions together are about 66 MB), so I've worked around this by only bundling Windows x64 by default. The call is only made on Windows, and I expect x64 to be the only common Windows platform. In total, this adds about 2.2MB to the extension.
> 
> It's been tested by the user that reported the issue and this seems to have fixed the issue ([#4059 (comment)](https://github.com/github/vscode-codeql/issues/4059#issuecomment-3027646631)), and I've also tested it on one of my own Windows machines where it resolves the path correctly (although the original workaround also works there, so I've artificially made it use the native method).

PR #4082, opened 2025-07-21
New command: expose `--cache-cleanup=overlay` base cache cleaning

d10c:
Add new command "CodeQL: Trim Overlay Base Cache" that returns a database to the state prior to overlay evaluation, leaving only base predicates and types that may later be referenced during overlay evaluation.

* Requires a recent enough query server. So the command is hidden behind a version check.
* Tested locally and with unit tests.

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Add new command "CodeQL: Trim Overlay Base Cache" that returns a database to the state prior to overlay evaluation, leaving only base predicates and types that may later be referenced during overlay evaluation. [#4082](https://github.com/github/vscode-codeql/pull/4082)
+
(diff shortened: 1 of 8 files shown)


d10c:
I've added a VSCode context variable for all the CLI features, but if that's too complicated, I can change it back to just the new CLI feature that we use.

d10c:
I'm fine with merging the query server changes first, so that we can include an integration test in this PR.

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Add new command "CodeQL: Trim Overlay Base Cache" that returns a database to the state prior to overlay evaluation, leaving only base predicates and types that may later be referenced during overlay evaluation.
```
koesie10: Can you add the link to the PR here?

koesie10 (approved):

PR #4111, opened 2025-08-14
Enable the @typescript-eslint/restrict-template-expressions rule

robertbrignull:
Enable the `@typescript-eslint/restrict-template-expressions` rule. This was "disabled" in the eslint 9 migration, though actually I don't think this rule was enabled before so really this is a new rule.

This PR will make behavioural changes to the extension, but I think all the changes are for the better 🤞🏼 

The purpose of this rule is to protect us from trying to stringify something that won't look nice. It should stop the classic `[Object object]` problem by forcing us to manually convert everything to a string first so we're being deliberate about how that transformation happens. This can be a little bit tedious but enabling it for our repo didn't turn out to be too hard.

To be clear, when you try to include a value in a string it'll call `.toString()` for you which is defined on all objects. Often this is fine, and many values (e.g. the URI class) have implemented a good `toString` method, but generic objects and arrays have no and they do not stringify well.

The most common cases we have are errors and URIs and both of those are easy to fix and shouldn't change the behaviour, even if calling `.toString()` on all URIs is a little annoying. But this rule did end up surfacing quite a few cases where I think we did have a bug and if that code had been hit (often it's an error message) then it wouldn't have looked pretty and wouldn't have been useful to us.

I haven't split this up into separate commits, but each code change is very localised and you can consider each of them separately when reviewing.


extensions/ql-vscode/src/codeql-cli/query-language.ts
```diff
@@ -89,7 +89,7 @@ export async function askForLanguage(
   if (!isQueryLanguage(language)) {
     void showAndLogErrorMessage(
       extLogger,
-      `Language '${language}' is not supported. Only languages ${Object.values(
+      `Language '${language as QueryLanguage}' is not supported. Only languages ${Object.values(
```
robertbrignull: This is an interesting case because the `language` variable actually has type `never` here. An alternative would be to cast it to `string` perhaps, or maybe `any` and then call `.toString()` manually, but I think we need to cast it to something.

extensions/ql-vscode/src/language-support/ast-viewer/ast-viewer.ts
```diff
@@ -80,11 +80,11 @@ class AstViewerDataProvider
     const treeItem = new TreeItem(item.label || "", state);
     treeItem.description = line ? `Line ${line}` : "";
     treeItem.id = String(item.id);
-    treeItem.tooltip = `${treeItem.description} ${treeItem.label}`;
+    treeItem.tooltip = `${treeItem.description} ${typeof treeItem.label === "string" ? treeItem.label : (treeItem.label?.label ?? "")}`;
```
robertbrignull, line 83: This is a bit ugly but the type of `treeItem.label` is `string | TreeItemLabel | undefined` and it doesn't seem that `TreeItemLabel` implements `toString()`.

extensions/ql-vscode/src/databases/local-databases/database-manager.ts
```diff
@@ -68,9 +68,7 @@ function eventFired<T>(
 ): Promise<T | undefined> {
   return new Promise((res, _rej) => {
     const timeout = setTimeout(() => {
-      void extLogger.log(
-        `Waiting for event ${event} timed out after ${timeoutMs}ms`,
-      );
+      void extLogger.log(`Waiting for event timed out after ${timeoutMs}ms`);
```
robertbrignull: The `vscode.Event<T>` type is just an interface with a "listen" method. So I don't know if there is any good way we could stringify the event to find out what it was. 🫤

extensions/ql-vscode/src/codeql-cli/query-language.ts
koesie10: I think casting it to `string` would make sense here. We know it's not a `QueryLanguage`, so I don't think casting it to `QueryLanguage` makes sense here.

extensions/ql-vscode/src/databases/local-databases/database-manager.ts
koesie10: Could we just pass a `string` to this function that gives a "friendly name" for the event? It seems like there's only 1 caller anyway.

extensions/ql-vscode/src/language-support/ast-viewer/ast-viewer.ts
```diff
@@ -80,11 +80,11 @@ class AstViewerDataProvider
...
     treeItem.command = {
       command: "codeQLAstViewer.gotoCode",
       title: "Go To Code",
-      tooltip: `Go To ${item.location}`,
+      tooltip: `Go To ${item.location?.toString() ?? "Code"}`,
```
koesie10: What will `item.location.toString()` return? It looks like `item.location` is an object, so would that just give `[Object object]`?

extensions/ql-vscode/src/codeql-cli/query-language.ts
robertbrignull: I'm fine with casting to `string`. Theoretically it's all just types here so it won't matter too much what we cast to.

extensions/ql-vscode/src/databases/local-databases/database-manager.ts
robertbrignull: Good idea. I thought it would be hard to do but you're right it's only called from one place and it's really easy to pass in an event name arg.

extensions/ql-vscode/src/language-support/ast-viewer/ast-viewer.ts
robertbrignull: Ah good spot. I just saw `UrlValue` as the type and assumed it would stringify nicely. 

Also I've been trying to test this to see what it looks like and I can't work out how to get this value to show up. It's not the hover text of the tree item but it's the tooltip of the on-click command. I'm not convinced it shows up anywhere so I've just changed it to a static "Go to Code" like the title.

PR #4170, opened 2025-10-21
Compare performance: Show names of query runs being compared

asgerf:
Adds a list of the query runs being compared, and their relative order. The UI for opening the comparison doesn't make it clear which is run will appear on the left or on the right.

-----

Screenshot 2025-10-21 at 16 44 20


d10c (approved):
LGTM

PR #4178, opened 2025-10-27
Bump jest version to 30.2.0

shati-patel:
Updates the jest version, which should unblock https://github.com/github/vscode-codeql/pull/4174. (I'm not sure why dependabot hasn't suggested this update yet—possibly because we were on a weird alpha version of jest previously.)

Anyway, see https://github.com/github/vscode-codeql/pull/4174#issuecomment-3451087109 for the details! 

----

(ℹ️ Note: I don't plan to merge this until after the GH Universe code freeze is lifted 🧊)

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -2113,7 +2113,7 @@
     "gulp-replace": "^1.1.3",
     "gulp-typescript": "^5.0.1",
     "husky": "^9.1.7",
-    "jest": "^30.0.0-alpha.7",
+    "jest": "^30.2.0",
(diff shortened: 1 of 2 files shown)


extensions/ql-vscode/package.json
koesie10: Do we also need to update this version?


```suggestion
    "jest": "^30.2.0",
```

shati-patel: Woops! Good point 🤦🏽‍♀️

koesie10 (approved):
Thanks!

PR #4195, opened 2025-11-12
add commands for warming the overlay-base cache

ginsbach:
- This fixes https://github.com/github/codeql-core/issues/5026 and is an important part of https://github.com/github/codeql-core/issues/5005.

This PR adds a second query server that is configured with
```
--no-evaluate-as-overlay --cache-at-frontier --warm-cache-only
```
and several commands:
- CodeQL: Warm Overlay-Base Cache for Query,
- CodeQL: Warm Overlay-Base Cache for Queries in Selected Files,
- CodeQL: Warm Overlay-Base Cache for Query Suite.

These commands are based on 
- CodeQL: Run Query on Selected Database,
- CodeQL: Run Queries in Selected Files,
- CodeQL: Run Selected Query Suite,

but they run on the second query server and their results are not loaded into the "CodeQL Query Results" view.

One complication of this approach with two query servers is that only one of them can register the database at any given time. This is handled via `withDatabaseInQsForWarmingOverlayBaseCache` and seems to work well in practice. It does mean that running a query while a cache warming is ongoing will result in an error message. That seems better than silently running on the wrong kind of query server, though, which would be the effect on an alternative implementation with a single query server that gets killed and restarted.

### DIFF
--- a/extensions/ql-vscode/src/extension.ts
+++ b/extensions/ql-vscode/src/extension.ts
@@ -86,7 +86,7 @@ import {
   extLogger,
   languageServerLogger,
   queryServerLogger,
-  queryServerForWarmingOverlayBaseCacheLogger,
+  getQueryServerForWarmingOverlayBaseCacheLogger,
 } from "./common/logging/vscode";
 import { QueryHistoryManager } from "./query-history/query-history-manager";
 import type { CompletedLocalQueryInfo } from "./query-results";
@@ -173,7 +173,7 @@ function getCommands(
   app: App,
   cliServer: CodeQLCliServer,
   queryRunner: QueryRunner,
-  queryRunnerForWarmingOverlayBaseCache: QueryRunner,
+  queryRunnerForWarmingOverlayBaseCache: QueryRunner | undefined,
   languageClient: LanguageClient,
 ): BaseCommands {
   const getCliVersion = async () => {
@@ -191,7 +191,9 @@ function getCommands(
         cliServer.restartCliServer();
         await Promise.all([
           queryRunner.restartQueryServer(progress),
-          queryRunnerForWarmingOverlayBaseCache.restartQueryServer(progress),
+          queryRunnerForWarmingOverlayBaseCache
+            ? queryRunnerForWarmingOverlayBaseCache.restartQueryServer(progress)
+            : {},
           async () => {
             if (languageClient.isRunning()) {
               await languageClient.restart();
@@ -204,10 +206,12 @@ function getCommands(
           queryServerLogger,
           "CodeQL Query Server restarted.",
         );
-        void showAndLogErrorMessage(
-          queryServerForWarmingOverlayBaseCacheLogger,
-          "CodeQL Query Server for warming overlay-base cache restarted.",
-        );
+        if (queryRunnerForWarmingOverlayBaseCache) {
+          void showAndLogErrorMessage(
+            getQueryServerForWarmingOverlayBaseCacheLogger(),
+            "CodeQL Query Server for warming overlay-base cache restarted.",
+          );
+        }
       },
       {
         title: "Restarting Query Server",
@@ -288,7 +292,7 @@ export interface CodeQLExtensionInterface {
   readonly ctx: ExtensionContext;
   readonly cliServer: CodeQLCliServer;
   readonly qs: QueryRunner;
-  readonly qsForWarmingOverlayBaseCache: QueryRunner;
+  readonly qsForWarmingOverlayBaseCache: QueryRunner | undefined;
   readonly distributionManager: DistributionManager;
   readonly databaseManager: DatabaseManager;
   readonly databaseUI: DatabaseUI;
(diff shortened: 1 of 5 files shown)


extensions/ql-vscode/package.json
```diff
@@ -1378,11 +1394,21 @@
           "group": "9_qlCommands",
           "when": "resourceScheme != codeql-zip-archive"
         },
+        {
+          "command": "codeQL.runWarmOverlayBaseCacheForQueries",
+          "group": "9_qlCommands",
+          "when": "resourceScheme != codeql-zip-archive"
```
nickrolfe: I think we should probably make this, and the other new commands, visible only in canary mode. At least for now, we're targeting this at internal users, and it'll let us gain some confidence with the approach of spinning up a second query server.

```suggestion
          "when": "resourceScheme != codeql-zip-archive && config.codeQL.canary"
```

extensions/ql-vscode/package.json
```diff
@@ -1408,6 +1434,10 @@
           "command": "codeQL.runQuery",
           "when": "resourceLangId == ql && resourceExtname == .ql"
         },
+        {
+          "command": "codeQL.runWarmOverlayBaseCacheForQuery",
+          "when": "resourceLangId == ql && resourceExtname == .ql"
```
nickrolfe: Same here:
```suggestion
          "when": "resourceLangId == ql && resourceExtname == .ql && config.codeQL.canary"
```

extensions/ql-vscode/package.json
```diff
@@ -1832,6 +1874,10 @@
           "command": "codeQL.runQueryContextEditor",
           "when": "editorLangId == ql && resourceExtname == .ql && !inDebugMode"
         },
+        {
+          "command": "codeQL.runWarmOverlayBaseCacheForQueryContextEditor",
+          "when": "editorLangId == ql && resourceExtname == .ql && !inDebugMode"
```
nickrolfe: Same here:
```suggestion
          "when": "editorLangId == ql && resourceExtname == .ql && !inDebugMode && config.codeql.canary"
```

extensions/ql-vscode/src/local-queries/local-queries.ts
```diff
@@ -160,10 +184,13 @@ export class LocalQueries extends DisposableObject {
           progress,
           token,
           undefined,
+          warmOverlayBaseCache,
         );
       },
       {
-        title: "Running query",
+        title: warmOverlayBaseCache
+          ? "Warm overlay-base cache for query"
```
nickrolfe, line 194: How about removing "for query", to shorten it slightly?

extensions/ql-vscode/src/query-server/query-server-client.ts
```diff
@@ -214,9 +215,21 @@ export class QueryServerClient extends DisposableObject {
...
+    }
+
+    const queryServerSuffix = this.warmOverlayBaseCache
+      ? " for warming overlay-base cache"
+      : "";
+
     const child = spawnServer(
       this.config.codeQlPath,
-      "CodeQL query server",
+      `CodeQL query server${queryServerSuffix}`,
```
nickrolfe, line 232: I notice that we spawn this second query server on startup. I imagine the vast majority of users will never need it. Can we find a way to start it lazily?

extensions/ql-vscode/src/extension.ts
```diff
@@ -201,6 +204,10 @@ function getCommands(
           queryServerLogger,
           "CodeQL Query Server restarted.",
         );
+        void showAndLogErrorMessage(
```
nickrolfe: I think you probably meant to write:
```suggestion
        void showAndLogInformationMessage(
```

extensions/ql-vscode/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [UNRELEASED]
 
+- Add new commands "CodeQL: Warm Overlay-Base Cache for Query", "CodeQL: Warm Overlay-Base Cache for Queries in Selected Files", and "CodeQL: Warm Overlay-Base Cache for Query Suite". These commands populate the overlay-base cache of an overlay database to prepare for incremental evaluation. They are primarily intended for overlay-specific performance debugging. [#4195](https://github.com/github/vscode-codeql/pull/4195)
```
nickrolfe: If we do put all the new commands behind the canary flag, we might not want to add a changelog entry.

ginsbach: :+1: commit 6 removes the changelog entry.

extensions/ql-vscode/src/extension.ts
ginsbach: Indeed

extensions/ql-vscode/src/query-server/query-server-client.ts
ginsbach: Commit 7 starts the second query server on demand, and also hides the logger until needed.

extensions/ql-vscode/src/local-queries/local-queries.ts
ginsbach: I'm not strictly against this, but on first impression, I think that "for query" actually does carry information here.
The cache will be set up so that this individual query will afterwards be evaluated incrementally, but that only holds for that one query.
Also, I don't think the command name is overly long.

nickrolfe (approved):
Looks great.

PR #4209, opened 2025-11-25
Fix rendering of settings by adding ids to package.json

nickrolfe:
This appears to fix a problem I noticed, where all the settings defined by this extension appeared in the settings UI under "Extensions->CLI" (the name of our first config category), rather than "Extensions->CodeQL->CLI", "Extensions->CodeQL->Running queries", etc.

I don't know exactly why this works, but I was just looking for differences between our `package.json` and that of a different open-source extension.

### Before
Screenshot 2025-11-25 at 16 43 51

### After
Screenshot 2025-11-25 at 16 43 38

### DIFF
--- a/extensions/ql-vscode/package.json
+++ b/extensions/ql-vscode/package.json
@@ -164,6 +164,7 @@
       {
         "type": "object",
         "title": "CLI",
+        "id": "codeQL.cli",
         "order": 0,
         "properties": {
           "codeQL.cli.executablePath": {
@@ -182,6 +183,7 @@
       {
         "type": "object",
         "title": "Running queries",
+        "id": "codeQL.runningQueries",
         "order": 1,
         "properties": {
           "codeQL.runningQueries.numberOfThreads": {
@@ -279,6 +281,7 @@
       {
         "type": "object",
         "title": "Results",
+        "id": "codeQL.resultsDisplay",
         "order": 2,
         "properties": {
           "codeQL.resultsDisplay.pageSize": {
@@ -291,6 +294,7 @@
       {
         "type": "object",
         "title": "Query history",
+        "id": "codeQL.queryHistory",
         "order": 3,
         "properties": {
           "codeQL.queryHistory.format": {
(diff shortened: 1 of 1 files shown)


d10c (commented):
Also, according to your screenshot, there remains a difference in the display of the title between `codeQL.cli.downloadTimeout` (id: `codeql-cli`, displayed as "Download Timeout" within the "CLI" category) and `codeQL.runningQueries.cacheSize` (id: `codeql-running-queries`, displayed as "Code QL > Running Queries: Cache Size" within the "Running queries" category), which makes me suspect that `-` is treated similar to `.` in stripping off the prefix from the setting ID. So I would suggest renaming the category `id` to always be a prefix of the setting IDs beneath it, and seeing if that gives a more consistent look to all the settings.

extensions/ql-vscode/package.json
```diff
@@ -457,6 +466,7 @@
       {
         "type": "object",
         "title": "Model Editor",
+        "id": "codeQL.modelEditor",
         "order": 9,
         "properties": {
           "codeQL.model.packLocation": {
```
d10c, line 472: Here we have a discrepancy between `id` and prefix.

extensions/ql-vscode/package.json
```diff
@@ -279,6 +281,7 @@
       {
         "type": "object",
         "title": "Results",
+        "id": "codeQL.results",
         "order": 2,
         "properties": {
           "codeQL.resultsDisplay.pageSize": {
```
d10c, line 287: Discrepancy between `id` and prefix.

d10c (changes requested):
Nice! Still a couple of discrepancies left, which if my theory is correct should display differently from the other settings.

extensions/ql-vscode/package.json
nickrolfe: Yes, your theory holds for this discrepancy. I'll fix.

d10c (approved):

PR #4210, opened 2025-11-25
Deprecate `codeQL.runningQueries.saveCache` setting

nickrolfe:
This setting was a bit of a footgun for users, and generally only caused performance to get worse.

By adding a `deprecationMessage` to `package.json`, the [docs](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration) suggest, and my testing confirms, that this will cause the setting to be hidden for users who haven't set it. For those who have explicitly set it (to either true or false), the settings UI will show the deprecation message.
 
Screenshot 2025-11-25 at 16 58 19

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Rename command "CodeQL: Trim Overlay Base Cache" to "CodeQL: Trim Cache to Overlay-Base" for consistency with "CodeQL: Warm Overlay-Base Cache for [...]" commands. [#4204](https://github.com/github/vscode-codeql/pull/4204)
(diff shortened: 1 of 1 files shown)


extensions/ql-vscode/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [UNRELEASED]
 
 - Rename command "CodeQL: Trim Overlay Base Cache" to "CodeQL: Trim Cache to Overlay-Base" for consistency with "CodeQL: Warm Overlay-Base Cache for [...]" commands. [#4204](https://github.com/github/vscode-codeql/pull/4204)
+- Deprecate the setting (`codeQL.runningQueries.saveCache`) to aggressively save intermediate results to the disk cache. [#4210](https://github.com/github/vscode-codeql/pull/4210)
```
cklin: (nit) There is some ambiguity, as the sentence can be parsed in two ways:
- Deprecate (the setting to aggressively save ...)
- (Deprecate the setting) to aggressively save ...

Please consider rewording it for greater clarity.  Thanks!

nickrolfe: Good point; fixed.

PR #4215, opened 2025-11-28
MRVA: use Go Autofix instead of Cocofix

d10c:
Updates the (still internal, therefore no change note) MRVA Autofix feature to use the latest Go-based binary.

Can be configured to use either an env var (`env:<ENV_VAR_NAME>`) or a 1Password secret reference (`op://<vault>/<item>/<field>`; requires `op` command) for the Copilot API key. By default, it uses the `CAPI_DEV_KEY` env var, as before.

- In the .md output, I'm seeing unescaped terminal escape codes like `�[31m-             }.run();�[0m`. Was it like that before?
- I'll invite discussion as to whether calling out to 1Password CLI is the best option, or if passing a path to a file containing the secret would be preferable.

<details>
<summary>Example autofix output</summary>

## mysql/mysql-connector-j

<details><summary>Fix suggestion details</summary>

```diff
Changes:

Will patch: src/test/java/testsuite/regression/StatementRegressionTest.java
...
                  }
  
[31m-             }.run();[0m
[32m+             }.start();[0m
  

Description:

To fix this issue, replace the direct call to `run()` with a call to `start()`. The `start()` method initializes a new thread and then calls the `run()` method internally, ensuring that the code in `run()` executes asynchronously on a separate thread. This approach preserves the intended concurrency.

Steps:
1. Locate the erroneous line (`.run()`).
2. Replace `.run()` with `.start()` to ensure the `Thread` runs in its own thread.
3. Validate that this change aligns with the intended behavior of the code.

---


Fix is valid.
```

</details>

 ### Notes
 - notes placeholder

## chinabugotech/hutool

<details><summary>Fix suggestion details</summary>

```diff
Changes:

Will patch: hutool-core/src/test/java/cn/hutool/core/io/WatchMonitorTest.java
...
  	public void testDir() {
  		monitor = WatchMonitor.createAll("d:/", new DelayWatcher(watcher, 500));
[31m- 		monitor.run();[0m
[32m+ 		monitor.start();[0m
  	}

Description:

To fix the problem, we should properly start the `WatchMonitor` instance on a new thread instead of calling its `run()` method directly. This typically involves using the `start()` method, which ensures the `run()` method executes on a separate thread as intended.

In this case, the following changes are needed:
1. Replace `monitor.run()` with `monitor.start()` in the affected test methods: `testDir` and `testDelay`.
2. No other modifications appear necessary since the `monitor` object is already correctly set up.

---


Fix is valid.
Changes:

Will patch: hutool-core/src/test/java/cn/hutool/core/io/WatchMonitorTest.java
...
  	public void testDir() {
  		monitor = WatchMonitor.createAll("d:/", new DelayWatcher(watcher, 500));
[31m- 		monitor.run();[0m
[32m+ 		new Thread(monitor).start();[0m
  	}

Description:

To fix this issue, we need to ensure that the `run()` method is called on a separate thread instead of the current thread. If `WatchMonitor` is a `Thread` or similar, then `start()` should be invoked instead of `run()`. If `WatchMonitor` is not directly a thread but implements `Runnable`, a `Thread` instance wrapping the `WatchMonitor` object should be created and started. This approach preserves the expected behavior of starting the monitoring task in a separate thread.

There are two locations in the code where this issue appears: line 68 (`monitor.run()`) and line 84 (`monitor.run()`). Both should be modified to ensure a new thread is started.

---


Fix is valid.
```

</details>

 ### Notes
 - notes placeholder

</details>

### DIFF
--- a/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
+++ b/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
@@ -1015,9 +1015,11 @@ async function formatWithMarkdown(
     const backFormatting: string =
       "```\n\n</details>\n\n ### Notes\n - notes placeholder\n\n";
 
-    // Format the content with Markdown
-    // Replace ``` in the content with \``` to avoid breaking the Markdown code block
-    const formattedContent = `## ${header}\n\n${frontFormatting}${content.replaceAll("```", "\\```")}${backFormatting}`;
+    // Format the content with Markdown:
+    // - Replace ``` in the content with \``` to avoid breaking the Markdown code block
+    // - Remove raw terminal escape sequences if any (workaround until `--diff-style plain` is handled by autofix)
+    // eslint-disable-next-line no-control-regex
+    const formattedContent = `## ${header}\n\n${frontFormatting}${content.replaceAll("```", "\\```").replaceAll(/\x1b\[[0-9;]*m/g, "")}${backFormatting}`;
 
     // Write the formatted content back to the file
     await writeFile(inputFile, formattedContent);


extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
```diff
@@ -155,6 +157,39 @@ async function findLocalAutofix(): Promise<string> {
...
+    // Don't allow literal keys in config.json for security reasons
+    throw new Error(
+      `Invalid CAPI dev key format. Use 'env:<ENV_VAR_NAME>' or 'op://<1PASSWORD_SECRET_REFERENCE>'.`,
+    );
+  }
+  if (capiDevKey.startsWith("env:")) {
+    const envVarName = capiDevKey.substring("env:".length);
+    capiDevKey = process.env[envVarName] || "";
+  }
+  if (capiDevKey.startsWith("op://")) {
```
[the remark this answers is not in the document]
d10c: I wanted to allow the option of the environment variable containing an `op://` reference.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
jcogs33: > In the .md output, I'm seeing unescaped terminal escape codes like `�[31m-             }.run();�[0m`. Was it like that before?

No, it wasn't like that before. Something changed with how Go Autofix handles the diff style. There's no longer a `diff` option. I tried all the [diff-style options](https://github.com/github/codeml-autofix/blob/d9426ce86844bfd24ce75a68e8c133d1fb6bf6e2/go/pkg/autofix/replacement/diff.go#L18-L29) for Go Autofix, but they all give this same output with the terminal codes. I'm not sure if that's expected or if there's a bug with how Go Autofix is handling diff styles. It's not a big deal, but it does make the output less readable. 

I've recommended using the `git` style option for now (to at least use an option that's supported for Go Autofix), but as mentioned, it won't currently fix the terminal code issue.
```suggestion
"--diff-style",
"git",
```

jcogs33 (commented):
Thanks! Looks reasonable to me. I tested locally and it works for me.
One suggestion to update the diff-style to an option supported by Go Autofix.

> I'll invite discussion as to whether calling out to 1Password CLI is the best option, or if passing a path to a file containing the secret would be preferable.

I don't have a strong opinion about the best option for this.

extensions/ql-vscode/src/variant-analysis/view-autofixes.ts
d10c: Bummer, but thanks, I'll apply your suggestion.

I'll also create an issue to track this regression in Go autofix relative to cocofix.

d10c: Strangely enough, I'm not getting any difference in output when setting a different `--diff-style` (e.g. git vs unified vs plain). They're all giving the same kind of diff:

```diff
[31m-             }.run();[0m
[32m+             }.start();[0m
```

jcogs33 (commented):
I double-checked that the qhelp is still being used correctly and found a problem while checking that: The Go binary needs to be re-built if the qhelp file changes. Otherwise it will keep using the older version of the qhelp. Can we add a function to run `make build` sometime after running [`overrideQueryHelp`](https://github.com/d10c/vscode-codeql/blob/35d636ee192c74c05f54b61b8950604e40962992/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts#L87) and before running [`execAutofix`](https://github.com/d10c/vscode-codeql/blob/35d636ee192c74c05f54b61b8950604e40962992/extensions/ql-vscode/src/variant-analysis/view-autofixes.ts#L790)? If we can check whether the qhelp changed during `overrideQueryHelp` and only run `make build` if it changed, that would probably be the most efficient approach.

Noting some other things that aren't working quite as expected. These would need to be resolved in Go Autofix, not in this PR.
- I don't think the [`--no-cache` option](https://github.com/github/codeml-autofix/blob/c68c3d420130812520f6442317c37aa1361a53ab/go/cmd/autofix/main.go#L565) for Go Autofix works. I'm seeing [`cocofix/.cache`](https://github.com/github/codeml-autofix/blob/c68c3d420130812520f6442317c37aa1361a53ab/go/cmd/autofix/main.go#L564C39-L564C55) directories every time I run Go Autofix despite using `--no-cache`. I'm not sure whether or not this is a problem. I used `--no-cache` just to be sure that the model response was always based on the current sarif/qhelp, but maybe it's fine without that. Maybe worth an issue in `codeml-autofix`.
- Go Autofix does not output a [fix-description](https://github.com/github/codeml-autofix/blob/c68c3d420130812520f6442317c37aa1361a53ab/go/cmd/autofix/main.go#L576). It's not a big deal for this feature since nothing breaks and the fix-description output is not particularly important. Maybe worth an issue in `codeml-autofix`.

jcogs33 (approved):
Looks reasonable to me. Thanks!

PR #4362, opened 2026-04-09
Add support for selection-based result filtering in result viewer

asgerf:
Adds support for selection-based result filtering via a checkbox in the result viewer.
- When enabled, only results from the currently-viewed file are shown.
- Additionally, if the editor selection is non-empty, only results within the selection range are shown.

A tuple matches the filter if any of its cells contain an entity whose location matches the filter.

https://github.com/user-attachments/assets/b84b7c21-8ca0-4e3b-9fe6-29759e98e0a0

## Implementation notes
The result viewer shows results in pages and the webview only has access to its current page. But if we simply apply the filtering on a page-by-page basis, the filtered results could end being scattered across thousands of mostly-empty pages.

The ideal might have been to support for filtering in `codeql bqrs` like we do for sorting, although that wouldn't work for interpreted results from SARIF files. This is the approach I took:
- When the currently-viewed file changes, read the whole BQRS/SARIF in the extension and send the results for that file to the webview.
- The webview then post-filters those to the current selection range (if the selection is non-empty).
- There is no support for pagination, but the filtering will in practice reduce the number of results so they fit in the UI just fine.

The most complex part of the change is getting the extension <-> webview communication to work well and making sure we don't show stale results regardless of what order UI changes occur in.

The `selectedTable` state had to be lifted one component up in the hierarchy (from `ResultTable` to `ResultApp`) so it is available at the point where we request file-filtered results from the extension.

### DIFF
--- a/extensions/ql-vscode/src/local-queries/results-view.ts
+++ b/extensions/ql-vscode/src/local-queries/results-view.ts
@@ -296,12 +296,23 @@ export class ResultsView extends AbstractWebview<
           this.onWebViewLoaded();
           break;
         case "viewSourceFile": {
-          await jumpToLocation(
+          const jumpTarget = await jumpToLocation(
             msg.databaseUri,
             msg.loc,
             this.databaseManager,
             this.logger,
           );
+          if (jumpTarget != null) {
+            // For selection-filtering purposes, we want to notify the webview that a new file is being looked at.
+            await this.postMessage({
+              t: "setEditorSelection",
+              selection: this.rangeToEditorSelection(
+                jumpTarget.uri,
+                jumpTarget.range,
+              ),
+              wasFromUserInteraction: false,
+            });
+          }
           break;
         }
         case "toggleDiagnostics": {
--- a/extensions/ql-vscode/src/local-queries/results-view.ts
+++ b/extensions/ql-vscode/src/local-queries/results-view.ts
@@ -51,6 +51,7 @@ import {
   GRAPH_TABLE_NAME,
   NavigationDirection,
   getDefaultResultSetName,
+  RAW_RESULTS_LIMIT,
 } from "../common/interface-types";
 import { extLogger } from "../common/logging/vscode";
 import type { Logger } from "../common/logging";
@@ -62,6 +63,8 @@ import type {
 import { interpretResultsSarif, interpretGraphResults } from "../query-results";
 import type { QueryEvaluationInfo } from "../run-queries-shared";
 import {
+  getLocationsFromSarifResult,
+  normalizeFileUri,
   parseSarifLocation,
   parseSarifPlainTextMessage,
 } from "../common/sarif-utils";
@@ -82,7 +85,7 @@ import { redactableError } from "../common/errors";
 import type { ResultsViewCommands } from "../common/commands";
 import type { App } from "../common/app";
 import type { Disposable } from "../common/disposable-object";
-import type { RawResultSet } from "../common/raw-result-types";
+import type { RawResultSet, Row } from "../common/raw-result-types";
 import type { BqrsResultSetSchema } from "../common/bqrs-cli-types";
 import { CachedOperation } from "../language-support/contextual/cached-operation";
 
@@ -349,6 +352,7 @@ export class ResultsView extends AbstractWebview<
           await this.openFile(msg.filePath);
           break;
         case "requestFileFilteredResults":
+          void this.loadFileFilteredResults(msg.fileUri, msg.selectedTable);
           break;
         case "telemetry":
           telemetryListener?.sendUIInteraction(msg.action);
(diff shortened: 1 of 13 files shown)


extensions/ql-vscode/src/view/results/result-table-utils.ts
```diff
@@ -107,3 +117,110 @@ export function nextSortDirection(
...
+  }
+
+  // If the result location has no line info, it's a whole-file location — include it
+  if (loc.startLine === undefined) {
+    return true;
+  }
+
+  // Only include results whose starting line falls within the selection range
+  return (
+    loc.startLine >= selection.startLine && loc.startLine <= selection.endLine
```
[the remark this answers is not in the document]
asgerf: It is the intended behaviour

extensions/ql-vscode/src/view/results/ResultsApp.tsx
```diff
@@ -180,7 +253,43 @@ export function ResultsApp() {
...
+                results.fileUri === prev.editorSelection?.fileUri &&
+                results.selectedTable === prev.selectedTable &&
+                prev.fileFilteredResults === undefined
+              ) {
+                return { ...prev, fileFilteredResults: results };
+              }
+              return prev;
+            });
+          }
           break;
```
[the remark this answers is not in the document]
asgerf: Fixed in [7fe9983](https://github.com/github/vscode-codeql/pull/4362/commits/7fe99836a93b63d800b4d862a7990d95aa075cbb)

extensions/ql-vscode/src/view/results/ResultTables.tsx
```diff
@@ -223,15 +196,85 @@ export function ResultTables(props: ResultTablesProps) {
 
   const resultSetName = resultSet ? getResultSetName(resultSet) : undefined;
 
+  // True if file-filtered results are still loading from the extension
+  const isLoadingFilteredResults =
+    selectionFilter != null && fileFilteredResults == null;
```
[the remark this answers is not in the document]
asgerf: Fixed in [7fe9983](https://github.com/github/vscode-codeql/pull/4362/commits/7fe99836a93b63d800b4d862a7990d95aa075cbb)

cklin:
Some initial feedback:

1. Please restructure the PR into smaller, self-contained commits so that they are easier to understand and to review.  Ideally the PR would separate into different commits (a) structural changes that are obviously behavior-preserving from (b) behavior changes, with most of the change in the first category, so that I can devote more attention to the parts that need to be reviewed in more detail.
2. Can you elaborate how you arrived at the design of performing the filtering on the (webview) client side?  I would have expected this design:
   1. The client side informs the (extension) server side when the state of the filtering checkbox changes
   2. The server side applies the filtering (or not) to the results
   3. The server side informs the client side that the results have changed
   4. The client side fetches the updated results from the server side

   The client and the server probably already handle server-side result changes, so steps (iii) and (iv) can use that existing infrastructure.  There would be minimal changes on the client side, and existing features (such as result pagination) would "just work" also for filtered results.

   I am sure you have reasons for doing things your way, and I want to hear them.

asgerf:
Thanks for looking into this. I'll look into restructuring the commit history when I have time.

As for the way filtering work: It started out being more server-side but state management became too error-prone. There's more state to keep track of than is obvious at first, and you end up with a bunch of fields and patterns of "whenever X changes/happens remember to also do Y". I eventually refactored into the tried and true "React + stateless server" pattern which is less error-prone.

It's hard to remember _exactly_ why it was so hard, other than the fact that there was always one more nuance to deal with, and it felt like the architecture wasn't sound. This is my best attempt at bringing up from memory what the issues were:

**Selection and active editor**
There are some subtleties around when the editor selection or active-editor change should update the effective selection filter.
- Clicking a link in the results view can change the selection and/or active editor, but should not change the selection filter. (Otherwise it's hard to fully explore a set of results that you're focusing on)
- But if the user clicks a link and _then_ enables filtering we want the filter to apply to the currently-viewed file. 
- VSCode doesn't have an "active" editor when the webview has focus so we end up having to remember the last-active editor, yet another thing to track.

**Race conditions**
There is potential for race conditions between UI events and messages from the extension. The client might request a state change, but then the user quickly clicks a checkbox, then results arrive but now they are stale. If the client doesn't "own" enough state it gets very hard to validate what parts of a `setState` message is stale.

**Pagination**
Pagination for BQRS-based results still would not work without also changing `codeql bqrs decode` to do the file-filtering. We rely on `bqrs diff` for pagination, but as mentioned, we can't just filter page-by-page as the filtered results can get scattered across thousands of mostly-empty pages.

We'd also have to re-paginate the results whenever the selection changes, preferrably starting using the cached file-filtered results for efficiency. Not impossible to implement or anything, but I wouldn't say we get pagination "for free".

cklin:
Thanks for the additional context!  Sounds like server-side filtering is simpler only in theory; in that case client-side filtering makes sense.

Please @-mention me when you are done with the commit restructuring!

extensions/ql-vscode/src/local-queries/results-view.ts
```diff
@@ -1031,6 +1052,77 @@ export class ResultsView extends AbstractWebview<
...
+   * Does not send when there is no active text editor (e.g. when the webview
+   * gains focus), so the webview retains the last known selection.
+   */
+  private sendEditorSelectionToWebview(wasFromUserInteraction = false): void {
+    if (!this.isShowingPanel) {
+      return;
+    }
+    const selection = this.computeEditorSelection();
+    if (selection === undefined) {
+      return;
```
cklin, line 1066: I am surprised by the early return here.  How would the user revert from selection filtering to whole-file filtering?

extensions/ql-vscode/src/local-queries/results-view.ts
```diff
@@ -296,12 +296,23 @@ export class ResultsView extends AbstractWebview<
...
           );
+          if (jumpTarget != null) {
+            // For selection-filtering purposes, we want to notify the webview that a new file is being looked at.
+            await this.postMessage({
+              t: "setEditorSelection",
+              selection: this.rangeToEditorSelection(
+                jumpTarget.uri,
+                jumpTarget.range,
+              ),
+              wasFromUserInteraction: false,
```
cklin, line 313: Should `wasFromUserInteraction` be true here?

cklin (commented):
Thank you for breaking the PR down to smaller commits!  Just two minor questions.

extensions/ql-vscode/src/local-queries/results-view.ts
asgerf: `selection` being `undefined` means there is no active editor. An empty selection (just a cursor position) is represented as an selection object with `.isEmpty = true` and will reset to whole-file filtering.

extensions/ql-vscode/src/local-queries/results-view.ts
asgerf: Perhaps this was a poorly named concept, but no, `wasFromUserInteraction` means the user manually changed the selection in their editor (with mouse or keyboard), as opposed to being triggered by an extension.

cklin (approved):

asgerf:
@cklin Thanks for taking the time to review!

PR #4390, opened 2026-05-08
Add new command: 'Go to File in Selected Database'

hvitved:
@asgerf recently introduced [selection-based result filtering](https://github.com/github/vscode-codeql/pull/4362), which is a really great feature. However, getting to file in the source archive where you want to do the filtering currently requires running _some_ query and then being able to interpret the result of that query to match the file where filtering is intended.

This PR adds a new command `CodeQL: Go to File in Selected Database`, which works like VS Code's built-in `Go to File`, except it does so on the source archive of the selected database.

### DIFF
--- a/extensions/ql-vscode/src/databases/source-archive-file-search.ts
+++ b/extensions/ql-vscode/src/databases/source-archive-file-search.ts
@@ -1,0 +1,109 @@
+import type { QuickPickItem, Uri } from "vscode";
+import { FileType, window, workspace } from "vscode";
+import type { DatabaseItem } from "./local-databases";
+import {
+  encodeSourceArchiveUri,
+  decodeSourceArchiveUri,
+} from "../common/vscode/archive-filesystem-provider";
+
+interface SourceArchiveFileQuickPickItem extends QuickPickItem {
+  uri: Uri;
+}
+
+/**
+ * Recursively collects all file URIs from a source archive directory.
+ */
+async function collectFiles(
+  dirUri: Uri,
+  sourceArchiveZipPath: string,
+  prefix: string,
+): Promise<SourceArchiveFileQuickPickItem[]> {
+  const entries = await workspace.fs.readDirectory(dirUri);
+  const items: SourceArchiveFileQuickPickItem[] = [];
+
+  for (const [name, type] of entries) {
+    const childPath = prefix ? `${prefix}/${name}` : name;
+    const childUri = encodeSourceArchiveUri({
+      sourceArchiveZipPath,
+      pathWithinSourceArchive: `${decodeSourceArchiveUri(dirUri).pathWithinSourceArchive}/${name}`,
+    });
+
+    if (type === FileType.File) {
+      items.push({
+        label: name,
+        description: prefix,
+        uri: childUri,
+      });
+    } else if (type === FileType.Directory) {
+      const subItems = await collectFiles(
+        childUri,
+        sourceArchiveZipPath,
+        childPath,
+      );
+      items.push(...subItems);
+    }
+  }
+
+  return items;
+}
+
+/**
+ * Shows a Quick Pick to search for and open a file from the source archive
+ * of the given database.
+ */
+export async function searchSourceArchiveFiles(
+  databaseItem: DatabaseItem,
+): Promise<void> {
+  let explorerUri: Uri;
+  try {
+    explorerUri = databaseItem.getSourceArchiveExplorerUri();
+  } catch (e) {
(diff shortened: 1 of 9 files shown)


extensions/ql-vscode/src/databases/source-archive-file-search.ts
```diff
@@ -0,0 +1,109 @@
...
+  uri: Uri;
+}
+
+/**
+ * Recursively collects all file URIs from a source archive directory.
+ */
+async function collectFiles(
+  dirUri: Uri,
+  sourceArchiveZipPath: string,
+  prefix: string,
```
asgerf:
```suggestion
  prefix: string,
  items: SourceArchiveFileQuickPickItem[] = [],
```
I'd move `items` to an accumulator parameter to avoid the N^2 cost of building up intermediate lists

extensions/ql-vscode/src/databases/source-archive-file-search.ts
```diff
@@ -0,0 +1,109 @@
...
+): Promise<SourceArchiveFileQuickPickItem[]> {
+  const entries = await workspace.fs.readDirectory(dirUri);
+  const items: SourceArchiveFileQuickPickItem[] = [];
```
asgerf:
```suggestion
````

File: extensions/ql-vscode/src/databases/source-archive-file-search.ts [hunk 3]
```diff
@@ -0,0 +1,109 @@
...
+        description: prefix,
+        uri: childUri,
+      });
+    } else if (type === FileType.Directory) {
+      const subItems = await collectFiles(
+        childUri,
+        sourceArchiveZipPath,
+        childPath,
+      );
+      items.push(...subItems);
```
asgerf:
```suggestion
        childPath,
        items,
      );
```

extensions/ql-vscode/src/databases/source-archive-file-search.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  const quickPick = window.createQuickPick<SourceArchiveFileQuickPickItem>();
+  quickPick.placeholder = "Go to File in Selected Database...";
+  quickPick.matchOnDescription = true;
+  quickPick.busy = true;
+  quickPick.show();
+
+  try {
+    const items = await collectFiles(explorerUri, sourceArchiveZipPath, "");
+    // Sort items by file name, then by path
```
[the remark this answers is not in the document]
asgerf: I believe this is correct, although it only matters in extremely niche situations.

The returned promise is ultimately only used by the command handlers, which doesn't matter for UI-triggered commands, but it can matter if the command is invoked programmatically through `executeCommand`.

The fix should be straightforward though, just move the `Promise` creation up before the `show()` call and store the promise in a variable.

extensions/ql-vscode/src/databases/source-archive-file-search.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  return new Promise<void>((resolve) => {
+    quickPick.onDidAccept(async () => {
+      const selected = quickPick.selectedItems[0];
+      quickPick.dispose();
+      if (selected) {
+        const doc = await workspace.openTextDocument(selected.uri);
+        await window.showTextDocument(doc);
+      }
+      resolve();
```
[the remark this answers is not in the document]
asgerf: The suggestion here looks good to me.

hvitved:
@copilot: please address all review comments.

asgerf (commented):
Thanks! Looks good from my side. Someone from the core team will need to approve though.

extensions/ql-vscode/src/databases/source-archive-file-search.ts
```diff
@@ -0,0 +1,113 @@
...
+  try {
+    explorerUri = databaseItem.getSourceArchiveExplorerUri();
+  } catch (e) {
+    void window.showErrorMessage(e instanceof Error ? e.message : String(e));
+    return;
+  }
+  const sourceArchiveZipPath =
+    decodeSourceArchiveUri(explorerUri).sourceArchiveZipPath;
+
+  const filesPromise = collectFiles(explorerUri, sourceArchiveZipPath, "");
```
cklin: It looks like we are walking the entire source archive each time the user invokes the go-to-file command.  Is that the case?  If so, have you considered moving the tree walk outside the hot path?

hvitved: Good point, I'll update.

cklin (approved):
Thanks for the change!

PR #4415, opened 2026-06-04
Fix flaky Windows tests: dispose file watchers and use retried cleanup

cklin:
The root cause was twofold:

1. `MultiFileSystemWatcher.addWatcher()` never disposed the `FileSystemWatcher` object itself — only its event subscriptions. This leaked OS file handles on Windows, preventing temp directory deletion.
2. The tests used tmp's synchronous `rimrafSync` for cleanup, which has no retry logic for Windows' asynchronous handle release.

Fix:
- Dispose the `FileSystemWatcher` itself in `MultiFileSystemWatcher`
- Replace `removeCallback` with async `fs.rm` with `maxRetries`, giving Windows time to release handles between retries


nickrolfe (approved):
That seems plausible.

PR #4443, opened 2026-07-08
Fix CI failures

mbg:
The main change in this PR is that `yauzl` is bumped from `3.3.0` to `3.4.0`. That seems to resolve issues with zip extraction getting "stuck" and tests timing out as a result.

The other commits are misc. changes to fix issues I encountered while working on the repo locally.


josefs (approved):
LGTM

PR #4455, opened 2026-07-14
Refactor unzip functionality to limit concurrency and improve error handling in `copyStream`

### LINKED ISSUE #4453: Updating CodeQL CLI stalls EVERY TIME visual studio code is opened
**Describe the bug**
Every time I open Visual Studio Code a pop-up appears:

> Updating CodeQL CLI: Extracting CodeQL CLI v2.26.o... [658.2MB of 658.8 MB]
> Source: CodeQL

But the download/extract never seems to finish.

From the Output log, it seems that the download did finish and the CodeQL CLI is being extracted:

> Starting GitHub.vscode-codeql extension
> INFO: Could not find CodeQL on path.
> INFO: Could not find CodeQL on path.
> Searching for latest release including codeql-win64.zip.
> Extracting CodeQL CLI to c:\Users\JesseHouwing\AppData\Roaming\Code\User\globalStorage\github.vscode-codeql\distribution15

**Version**
Identifier
github.vscode-codeql
Version
1.17.7
Last Updated
4 days ago

Version: 1.128.0 (user setup)
Commit: fc3def6774c76082adf699d366f31a557ce5573f
Date: 2026-07-07T15:14:24-07:00
Electron: 42.5.0
ElectronBuildId: 14525058
Chromium: 148.0.7778.271
Node.js: 24.17.0
V8: 14.8.178.33-electron.0
OS: Windows_NT x64 10.0.26200

**To reproduce**
Open any visual studio code window, a notification toast appears and the download starts. But never seems to finish.

**Expected behavior**
Either the download completes correctly and installation completes
Or I get a clear message helping me figure out what's wrong.

**Additional context**

shati-patel:
Fixes https://github.com/github/vscode-codeql/issues/4453

Discussed with @josefs and tried various other fixes locally (like limiting the number of available cores). Eventually settled on this change to `copyStream`.

I installed the dev VSIX build locally and can confirm that the extraction now works smoothly 🎉 

--- 

Note: this still needs a changelog note and potentially some tidy-up

### DIFF
--- a/extensions/ql-vscode/src/common/unzip-concurrently.ts
+++ b/extensions/ql-vscode/src/common/unzip-concurrently.ts
@@ -9,7 +9,7 @@ export async function unzipToDirectoryConcurrently(
   progress?: UnzipProgressCallback,
 ): Promise<void> {
   const queue = new PQueue({
-    concurrency: availableParallelism(),
+    concurrency: Math.min(availableParallelism(), 4),
   });
 
   return unzipToDirectory(
--- a/extensions/ql-vscode/src/common/unzip-concurrently.ts
+++ b/extensions/ql-vscode/src/common/unzip-concurrently.ts
@@ -3,13 +3,22 @@ import type { UnzipProgressCallback } from "./unzip";
 import { unzipToDirectory } from "./unzip";
 import PQueue from "p-queue";
 
+/**
+ * Maximum number of files to extract concurrently. We cap this rather than
+ * using the full core count so that we don't open an excessive number of
+ * simultaneous file writes, which can overwhelm slower or contended storage
+ * during extraction.
+ */
+const MAX_UNZIP_CONCURRENCY = 4;
+
 export async function unzipToDirectoryConcurrently(
   archivePath: string,
   destinationPath: string,
   progress?: UnzipProgressCallback,
+  timeoutSeconds?: number,
 ): Promise<void> {
   const queue = new PQueue({
-    concurrency: Math.min(availableParallelism(), 4),
+    concurrency: Math.min(availableParallelism(), MAX_UNZIP_CONCURRENCY),
   });
 
   return unzipToDirectory(
@@ -19,5 +28,6 @@ export async function unzipToDirectoryConcurrently(
     async (tasks) => {
       await queue.addAll(tasks);
     },
+    timeoutSeconds,
   );
 }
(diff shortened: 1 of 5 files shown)


josefs:
Great start @shati-patel . I've pushed a commit to also add timeouts, in case something goes wrong with the writing. That will give us an extra layer or protection.

josefs:
I've added a test for the timeout and reworked the other test a bit. This is ready for review now.

extensions/ql-vscode/src/common/unzip.ts
```diff
@@ -88,31 +101,23 @@ export async function openZipBuffer(
...
-  return new Promise((resolve, reject) => {
-    readable.on("error", (err) => {
-      reject(err);
-    });
-    readable.on("end", () => {
-      resolve();
-    });
-
-    readable
-      .pipe(
```
kaspersv, line 105: Do I understand it correctly that the primary issue with the old code was that the use of `.pipe()` requires explicitly attaching error handlers to every stream which wasn't done here?

extensions/ql-vscode/src/common/unzip-concurrently.ts
```diff
@@ -3,13 +3,22 @@ import type { UnzipProgressCallback } from "./unzip";
 import { unzipToDirectory } from "./unzip";
 import PQueue from "p-queue";
 
+/**
+ * Maximum number of files to extract concurrently. We cap this rather than
+ * using the full core count so that we don't open an excessive number of
+ * simultaneous file writes, which can overwhelm slower or contended storage
+ * during extraction.
+ */
+const MAX_UNZIP_CONCURRENCY = 4;
```
kaspersv, line 12: Is this part of the actual bug-fix or just an optimisation to avoid contention due to excessive parallelism?

extensions/ql-vscode/src/common/unzip.ts
josefs: That's one of two problems. The other problem was that the old version attached the callback on the read-part of the stream, not the write part. The situation we seem to run into is that the read part finishes but the write part is blocked. So the callback is never invoked.

extensions/ql-vscode/src/common/unzip-concurrently.ts
josefs: This is just to reduce the likelihood of contention on the file lock. So it's not strictly necessary, just an extra precaution. 
We have actually replicated the bug on a Windows machine with just 1 thread.

extensions/ql-vscode/src/common/unzip.ts
kaspersv: Thanks! That makes sense.

kaspersv (approved):

PR #4465, opened 2026-07-17
When restarting the query server wait for the old one to stop properly

josefs:
Some of the Windows tests have been very flaky. I particularly experience that on  https://github.com/github/vscode-codeql/pull/4455

The problem is restarting the query server. The current code doesn't wait for the old server to stop properly. On Windows, the old process will hold locks on the file and if the new server starts up too quickly it will fail to open those files.

This PR makes sure to wait for the old server process to close properly (with a timeout, just to be sure) before opening the new one.

I don't know if people are experiencing any problems with this in practice, but it's good to have less flaky tests.

### DIFF
--- a/extensions/ql-vscode/CHANGELOG.md
+++ b/extensions/ql-vscode/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [UNRELEASED]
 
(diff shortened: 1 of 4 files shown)


josefs:
Closed in favor of https://github.com/github/vscode-codeql/pull/4464

PR #4484, opened 2026-07-29
Bump @vscode-elements/react-elements and fix resulting test failures

nickrolfe:
This contains the dependabot bump from https://github.com/github/vscode-codeql/pull/4420 plus some test fixes generated by Copilot.

I asked it to check whether anything in the implementation needed to change as well, and it claims not:

> Confirmed — nothing in the implementation code needed changes. Evidence:
> 
> 1. Both test failures were test-environment-only issues, not runtime bugs:
> 
> • ResizeObserver — jsdom doesn't implement it, but the real VS Code webview (a browser) provides it natively. The 2.x components (`vscode-table`, etc.) work fine at runtime; only the test env needed the polyfill.
> • Checkbox `role — 2.x removed the `role="checkbox"` attribute that 1.x reflected onto the host element, but the `<input type="checkbox">` inside its shadow DOM still carries the implicit `checkbox` role, so accessibility is preserved. Only the test's DOM query needed updating.
> 
> 2. No implementation code uses any removed/rewritten 2.0 API:
> 
> • The 2.0 breaking changes were: Tree rewrite, minimized `:host` styles, removed deprecated APIs, and Tabs accessibility. Grepping non-test code found no usage of `VscodeTree`, `VscodeScrollable` (deprecated `title`), or `VscodeContextMenu`, and no reliance on the host `role`.
> • The components actually used (`VscodeButton`, `VscodeTextfield`, `VscodeCheckbox`, `VscodeDivider`, `VscodeProgressRing`, `VscodeBadge`, `VscodeSingleSelect`/`VscodeOption`, `VscodeTabs`/`VscodeTabHeader`/`VscodeTabPanel`) keep the same props.
> 
> 3. `npx tsc --noEmit` on `src/view` passes cleanly against the new 2.x typings — no prop/type changes broke any implementation file.
> 
> So the version bump is safe; only the two test-only adjustments were required.


josefs (approved):
Looks plausible to me

